PR #227, opened 2017-05-24
fixed update-all driver (ae)

Paul-St-Young:
corrected handling of species with different mass in update-all driver
added tests for update-all driver on bcc hydrogen
added more observables to bcc test

### DIFF
--- a/tests/system/CMakeLists.txt
+++ b/tests/system/CMakeLists.txt
@@ -353,7 +353,6 @@ ENDIF()
                     BCC_H_SCALARS
                     0 # VMC
                     TRUE)
-  LIST(APPEND BCC_H_RMC_SCALARS "totenergy" "-1.8420 0.0011")
 
 # Check DMCUpdateAllWithRejection driver using all-particle updates
   QMC_RUN_AND_CHECK(short-bccH_1x1x1_ae-dmc-all_sdj
@@ -365,6 +364,7 @@ ENDIF()
                     1 # DMC
                     TRUE)
 
+  LIST(APPEND BCC_H_RMC_SCALARS "totenergy" "-1.8420 0.0011")
   #check RMC on 16 cores.  1 MPI
   QMC_RUN_AND_CHECK(short-bccH_1x1x1_ae-rmc_sdj
                     "${CMAKE_SOURCE_DIR}/tests/solids/bccH_1x1x1_ae"
--- a/tests/system/CMakeLists.txt
+++ b/tests/system/CMakeLists.txt
@@ -327,15 +327,6 @@ ENDIF()
                     0 # VMC
                     TRUE)
 
-  QMC_RUN_AND_CHECK(short-bccH_1x1x1_ae-vmc-all_sdj
-                    "${CMAKE_SOURCE_DIR}/tests/solids/bccH_1x1x1_ae"
-                    qmc_short_vmc_all
-                    qmc_short_vmc_all.in.xml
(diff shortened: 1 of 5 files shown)


tests/system/CMakeLists.txt
```diff
@@ -323,9 +327,21 @@ ENDIF()
                     0 # VMC
                     TRUE)
 
+  QMC_RUN_AND_CHECK(short-bccH_1x1x1_ae-vmc-all_sdj
```
markdewing: Could you add a comment here (and to the other added test) about the focus of the tests (all-electron moves).   The test names are necessarily short, but cryptic.   Plus there's a double use of 'all electron' - 'ae' is the all-electron potential (vs a pseudopotential) and 'all' for all-electron moves (vs. particle-by-particle moves)

tests/system/CMakeLists.txt
```diff
@@ -335,6 +351,15 @@ ENDIF()
...
+                    "${CMAKE_SOURCE_DIR}/tests/solids/bccH_1x1x1_ae"
+                    qmc_short_vmc_dmc_all
+                    qmc_short_vmc_dmc_all.in.xml
+                    1 16
+                    BCC_H_DMC_SCALARS
+                    1 # DMC
+                    TRUE)
+
   IF(NOT QMC_CUDA)
   LIST(APPEND BCC_H_RMC_SCALARS "totenergy" "-1.8420 0.0011")
```
ye-luo: Move the two QMC_RUN_AND_CHECK after line 363     IF(NOT QMC_CUDA) .
GPU has no all-electron move support.

tests/system/CMakeLists.txt
```diff
@@ -336,7 +343,28 @@ ENDIF()
...
+# Check VMCUpdateAllWithDrift driver using all-particle updates
+  QMC_RUN_AND_CHECK(short-bccH_1x1x1_ae-vmc-all_sdj
+                    "${CMAKE_SOURCE_DIR}/tests/solids/bccH_1x1x1_ae"
+                    qmc_short_vmc_all
+                    qmc_short_vmc_all.in.xml
+                    1 16
+                    BCC_H_SCALARS
+                    0 # VMC
+                    TRUE)
   LIST(APPEND BCC_H_RMC_SCALARS "totenergy" "-1.8420 0.0011")
```
markdewing: It looks like the DMC test got stuck between the RMC energy definition and the RMC test.

ye-luo:
To build up the matrix, we need a list of specification for the tests we have.
We can then filter out the unnecessary redundant tests and add the missing ones.

prckent:
@Paul-St-Young Any thoughts on what is likely working or near working in terms of update-all methods?

Paul-St-Young:
I just added a new test for the update-all VMC driver without drift. My initial pull request included a test for the update-all DMC driver. Currently the update-all drivers work in the case of no pseudopotential. I am not sure about cases with pseudopotential.

Driver selection is currently not very robust. In QMCApp/QMCDriverFactory.cpp, a bit is set to 1 if move="pbyp"
`WhatToDo[UPDATE_MODE]= (update_mode == "pbyp");`
otherwise the bit is set to 0 and the update-all drivers are used. Someone could inadvertently invoke an update-all driver if "pbyp" was misspelled for example.

prckent:
I suspect we'll need to error trap the GPU version, but lets see the overnight tests first.

Merging.

PR #231, opened 2017-05-26
Fix allp move with MP. No need for recomputePsi.

ye-luo:
Address the nightly test failure.

### DIFF
--- a/src/QMCDrivers/DMC/DMCOMP.cpp
+++ b/src/QMCDrivers/DMC/DMCOMP.cpp
@@ -299,7 +299,7 @@ bool DMCOMP::run()
         if(QMCDriverMode[QMC_UPDATE_MODE] && now%updatePeriod == 0)
           Movers[ip]->updateWalkers(wit, wit_end);
         // recompute the accuracy critical part of Psi at the end of the last step.
-        if ( step+1 == nSteps && nBlocksBetweenRecompute && (1+block)%nBlocksBetweenRecompute == 0 )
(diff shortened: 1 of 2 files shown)


prckent:
For future reference, what is the sensitivity to mixed-precision here? What should we be on the lookout for? Why is the call to recomputePsi() dangerous? Would it not be better long term to fix this function instead?

ye-luo:
In all electron move, every proposed move only calls evaluateLog which compute WF from scratch and doesn't need the walker buffer (size 0 as print out). recomputePsi recomputes SD inversion by checking out the buffer, redoing inversion and updating the buffer. So there is a segfault.

To me, fixing this function seem hiding developer mistakes to me. Better to have a crash and identify the real problem.

prckent:
So recomputePsi should check is the buffer is allocated, or simply abort. Presumably other functions have the same problem.

I think the higher level problem here is that the contract / responsibilities between the driver and walker classes is not clear. Agree?

ye-luo:
I agree. I guess at the beginning, QMCPACK didn't have so many variants in methods and didn't define the responsibilities so clear.

When I added recomputePsi, having it working as I need is the first priority. I was even not very happy to add recomputePsi because I had thought I only need to add WF routines instead of driver routines.
As we keep cleaning up the code on the ANL branch, we changed the way to handle recompute because recomputePsi has too much overhead. It has been deprecated.

prckent:
I'll merge this pragmatically: this is by far the fastest way to get the allp driver working. The proviso is that we we need to fix the logic / contract between the drivers and walker classes.

PR #233, opened 2017-05-27
Add output to DMC CPU driver.  Indicates status of t-moves

rcclay:
Oddly, DMCOMP doesn't display status of non-local moves, whereas DMC_CUDA does.  This writes a line of output to standard out.

### DIFF
--- a/src/QMCDrivers/DMC/DMCOMP.cpp
+++ b/src/QMCDrivers/DMC/DMCOMP.cpp
@@ -164,6 +164,12 @@ void DMCOMP::resetUpdateEngines()
         o << "\n  Walkers are killed when a node crossing is detected";
       else
         o << "\n  DMC moves are rejected when a node crossing is detected";
+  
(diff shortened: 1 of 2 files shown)


prckent:
Changed DMC_CUDA.cpp to match.

ye-luo:
The print must be in run().
The flag used to distinguish T-move is different from the one actually used in the mover.
This place only print once when the movers are created but the code actually allows switch on and off T-moves in different DMC blocks.

ye-luo:
@rcclay please resubmit a PR.
1, remove NonLocalMove in DMCOMP.h and DMCOMP.cpp
2, add a function in QMCUpdateBase which return the value of UseTMove
3, print from the end of resetUpdateEngines after "std::ostringstream o"
4, apply the same change on DMC_CUDA

rcclay:
@ye-luo Thanks for the heads up.  I'll try to resubmit over the weekend.

PR #234, opened 2017-05-27
Revert "Add output to DMC CPU driver.  Indicates status of t-moves"

ye-luo:
Reverts QMCPACK/qmcpack#233


prckent:
I assume there is an issue, so we'll revert

PR #237, opened 2017-05-30
do not reset 'feedback' parameter

Paul-St-Young:
The DMC population feedback parameter should be controlled by the input
as stated by the manual. It is currently forced to be 1.0.


ye-luo:
Could you give more detail how this change impacts your calculation? You think there is a bug or something else. I need solid justification that this change is necessary.

ye-luo:
For what I understand. This overwrite happens when you switch from fix-pop to fluc-pop. During fix-pop the default value is 0.0 and after switching the value should be non-zero that is why it is set 1.0 since users don't set feedback in most cases. I can understand what you need but I can't accept your change as it breaks the code for other users unless you can prove that after removing this line the value is 1.0 even if a user don't specify feedback in his DMC block.

Paul-St-Young:
This parameter allows me to study population bias. I see how removing the line would not be correct. I guess the correct way to do this is to propagate the user input into the reset function?

ye-luo:
I have an earlier investigation on population control and population bias. Probably you will help me to complete that investigation. I will communicate with you via email.

PR #238, opened 2017-05-31
Manual updates cmake flags

jptowns:
Added CMake optional arguments MPIEXEC and MPIEXEC_NUMPROCS_FLAG to the manual under the installation section.

### DIFF
--- a/manual/installation.tex
+++ b/manual/installation.tex
@@ -248,6 +248,7 @@ FFTW_LIBRARY_DIRS   Specify library directories for FFTW
 \begin{verbatim}
 MPIEXEC   Specify the mpi wrapper, e.g. srun, aprun, mpirun, etc.
 MPIEXEC_NUMPROC_FLAG   Specify the number of mpi processes flag, e.g. "-n", "-np", etc.
+
 \end{verbatim}
 \end{itemize}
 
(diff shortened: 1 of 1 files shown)


prckent:
Are you planning on adding the CMakeLists.txt update to this PR?

grahamlopez:
okay to test

ye-luo:
I clicked details but I can't access the page. Is it only accessible from ORNL?

grahamlopez:
@ye-luo For right now, it is only available at ORNL. This will hopefully change in the future. In the meantime, failing build  logs will get mailed to Mark and Paul (and me so I can debug Jenkins).

Everyone else, I'm still getting the final CI configuration tweaks settled, so ignore the CI messages on this PR for right now.

grahamlopez:
test this please

grahamlopez:
test this please

ye-luo:
@grahamlopez Could add me as a build log receiver?

atillack:
I am fine with the addition to installation.tex.

At the same time, changing CMakeLists.txt to tell the user about these options when the defaults fail may be more useful.

jptowns:
I did not include the updated CMakeLists.txt because specifying MPIEXEC solves the problem. I had assumed that people would much rather add to their build scripts than dig around in CMakeLists.txt.
Also, I'm not sure how general the problem I was experiencing is, so I was reticent to change the main CMakeLists.txt

prckent:
I see now that the FindMPI module in CMake handles this https://cmake.org/cmake/help/latest/module/FindMPI.html

Updates to CMakeLists.txt are therefore not needed. 

+1

PR #239, opened 2017-05-31
Jenkins on Rhea

grahamlopez:
These scripts will be called by Jenkins at OLCF to run CI on Rhea


grahamlopez:
blocking_qsub.py lives in the jenkins workspace right now. There are quite a few hacks right now that can eventually be cleaned up (I have a list)

PR #240, opened 2017-05-31
Updates for smoother compilation on ppc64le (SummitDev) architecture.

atillack:
This pull request updates the compiler detection and settings for the newer IBM XL compiler on the ppc64le architecture as used on OLCF SummitDev. Care was taken that legacy settings for AIX are kept around.

Additionally, the build recipe for OLCF SummitDev is updated to reflect the current software stack and this change.


prckent:
ok to test

CMake/IBMCompilers.cmake
```diff
@@ -54,3 +55,43 @@ IF(QMC_BITS MATCHES 64)
...
+ENDIF()
+
+# Suppress compile warnings
+SET(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -Wno-deprecated -Wno-unused-value")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated -Wno-unused-value")
+
+# Set standard and other optimization settings
+SET(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -std=gnu11 -O3")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11 -qnoxlcompatmacros -O3")
+
```
ye-luo: Do we need to add -std=gnu11 by hand? Does the upper level cmake add -std=c++11 and does it work?
Doesn't the build type release add -O3?

CMake/IBMCompilers.cmake
```diff
@@ -54,3 +55,43 @@ IF(QMC_BITS MATCHES 64)
...
+# Clean up flags
+
+IF(CMAKE_C_FLAGS MATCHES "-qhalt=e")
+  SET(CMAKE_C_FLAGS "")
+ENDIF()
+IF(CMAKE_CXX_FLAGS MATCHES "-qhalt=e")
+  SET(CMAKE_CXX_FLAGS "")
+ENDIF()
+
```
ye-luo: This seems like a hack. Is this needed because cmake doesn't play well with XL?

CMakeLists.txt
```diff
@@ -393,7 +393,12 @@ ELSE(CMAKE_TOOLCHAIN_FILE)
     MESSAGE(WARNING "No compiler identified")
   ENDIF()
   MESSAGE(STATUS "Compiler: ${COMPILER}")
-  MESSAGE(STATUS "Compiler Version: ${CMAKE_CXX_COMPILER_VERSION}")
+  # Workaround for not properly identified CXX compiler in case of IBM XL compiler
+  IF( ${COMPILER} MATCHES "IBM" )
+    MESSAGE(STATUS "Compiler Version: ${CMAKE_C_COMPILER_VERSION}")
+  ELSE()
+    MESSAGE(STATUS "Compiler Version: ${CMAKE_CXX_COMPILER_VERSION}")
+  ENDIF()
```
ye-luo: Is it possible to print inside  CMake/IBMCompilers.cmake

ye-luo: Is it possible to print inside CMake/IBMCompilers.cmake

manual/installation.tex
```diff
@@ -632,10 +610,9 @@ \subsubsection{Building QMCPACK}
 module load cuda
 mkdir build_summitdev
 cd build_summitdev
-cmake -DCMAKE_C_COMPILER="mpicc" \
-      -DCMAKE_C_FLAGS="-std=gnu11 -qsmp -O3" \
-      -DCMAKE_CXX_COMPILER="mpicxx" \
-      -DCMAKE_CXX_FLAGS="-std=gnu++11 -qnoxlcompatmacros -qsmp -O3" \
+cmake -DCMAKE_C_COMPILER="mpixlc" \
+      -DCMAKE_CXX_COMPILER="mpixlC" \
+      -DBUILD_LMYENGINE_INTERFACE=0 \
```
ye-luo: BUILD_LMYENGINE_INTERFACE=0 is not necessary. QMC_CUDA=1 will disable it.

CMake/IBMCompilers.cmake
atillack: Yes, cmake populates those flags with "-qthreaded -qhalt=e" where -qthreaded is not doing anything and -qhalt=e is not supported anymore on the newer architecture. The other work-around would be to pass -DCMAKE_C_FLAGS="" and -DCMAKE_CXX_FLAGS="" to the cmake command which is not overly user-friendly.

CMakeLists.txt
prckent: If CMAKE_C_COMPILER_VERSION not defined with cmake on summitdev and IBM compilers?

Why not print both C and CXX cmake compiler versions all the time? (Note that cmake does not guarantee these will be set, but it is good to see)

manual/installation.tex
atillack: You are technically correct (the best kind of correct) ;-)

I left it in for those folks who want to compile the CPU version in which case the CXX11 test now passes and formic/utils MPI implementation causes a hard compile error as the C++ bindings are no longer available on SummitDev.

CMakeLists.txt
atillack: Sure, but this message is originally output at this location and that's why I left it there. If I leave the original line untouched and add some output to IBMCompilers.cmake we would have two contradicting lines of output...

manual/installation.tex
prckent: BUILD_LMYENGINE_INTERFACE=0 needed for CPU builds

atillack:
@ye-luo That's a great question (easier CUDA arch passing) and is probably easy to implement - it does seem hard-coded to sm_35 CUDA version >= 6.5 and sm_20 below.

manual/installation.tex
ye-luo: @prckent Is this because of the MPI in the optimizer library?

CMakeLists.txt
ye-luo: I was wrong. DO NOT move it in CMake/IBMCompilers.cmake.
Somehow we already have printout in the very beginning, for example.
 -- The C compiler identification is GNU 6.2.0
-- The CXX compiler identification is GNU 6.2.0
Here we print again.
-- Compiler: GNU
-- Compiler Version: 6.2.0
Unless they can be different, we can just remove the printout here.

CMake/IBMCompilers.cmake
ye-luo: Adding CMAKE_C_FLAGS and CMAKE_CXX_FLAGS is definitely not desired.

manual/installation.tex
ye-luo: @atillack Clear. Thanks.

CMake/IBMCompilers.cmake
atillack: Interestingly, gnu11 and gnu++11 used to be needed but this does not seem to be the case anymore - c++11 works fine and gets added automatically when needed, unlike -O3 which is not automatically added for me. I'll make another commit after some more testing.

atillack: Well, never mind, the flags are needed as they are in this commit. While it compiles fine with c++11, I could not successfully run the created binary (tested with the GPU version) ...

ye-luo: Could you paste
-- CXX compiler flags:
-- C compiler flags:
reported by cmake?

ye-luo:
Can we drop sm_20(fermi) completely? We don't even have any machine to test this case.

prckent:
Yes, we can drop Fermi. Large production machines have Kepler or Pascal.

CMake/IBMCompilers.cmake
atillack: Both C and CXX compiler flags going into IBMCompilers.cmake are "-qthreaded -qhalt=e".

prckent:
Any objections to merge this version?

ye-luo:
wait

CMakeLists.txt
ye-luo: On a Cray, I got
-- The C compiler identification is Intel 17.0.0.20170213
-- The CXX compiler identification is Intel 17.0.0.20170213
and later
-- Compiler: Intel
-- Compiler Version: 17.0.0.20170213
It is basically redundant info and better just remove.

ye-luo: Any objection to remove?

atillack: Not always though - the first lines of output come from CMake's compiler tests which we have no influence on, while the later lines are our output. I agree that the version info may be redundant but I would urge to at least keep the compiler info as this is our only line of output to verify that our own tests chose wisely.

ye-luo: Please remove the version printout.

ye-luo:
I have added the switch allowing changing CUDA architecture. @atillack Please check and update your build recipe for SummitDev. use -D CUDA_ARCH=sm_60 to overwrite the default sm_35.

atillack:
@ye-luo Thanks! The build still works fine and I'm updating the build recipe and the compiler version output with the next (and hopefully final) commit.

ye-luo:
+1 OK to merge

PR #243, opened 2017-06-01
Add check for python modules for estimator tests.

markdewing:
Check a list of python modules to ensure they are installed before adding the estimator tests.
The test for module installation is based on whether the module successully imports or not.
If needed, additional tests of module functionality could be added.


jtkrogel:
This looks useful and I think it is implemented well.  +1

PR #244, opened 2017-06-01
Make short-bccH_1x1x1_ae-dmc_sdj-1-16 test longer

ye-luo:
See issue #241


markdewing:
okay to test

prckent:
Shouldn't we update the expected standard error in CMakeLists.txt for this test (/2) ?

ye-luo:
0.005 was still not very stable. Use 0.01 in both.

prckent:
I would prefer to keep the timestep at 0.001 and lengthen the tests as originally suggested. This way we will still be able to refer to the test history.

ye-luo:
Do we really need to refer to the test history?
Issue #241, the test was failing >20%.

ye-luo:
@prckent All the dmc references are updated.

prckent:
This update also removes the not needed or used parts of the Jastrow functions.

PR #246, opened 2017-06-01
nexus: fix for net_spin="low" in generate_physical_system

jtkrogel:
Very small change to correct a bug in tiled systems w/ requested low spin.  Useful for systems such as Al, which have a single atom in the primitive cell (necessarily spin polarized with integer electrons), but are spin unpolarized in supercells.


markdewing:
okay to test

Issue #232, opened 2017-05-26
Testing: Test of restart code needed


prckent:
Clear tests (and therefore examples) of working restarts with VMC and DMC are needed.

Based on these tests, the restart code can be updated/cleaned/rewritten as needed. No need for backward compatibility.

prckent:
@markdewing Delegate/ask for helpers as needed

prckent:
This is also needed to interface with the ECP VeloC project.

ye-luo:
This part of my project for the summer student. Deal with the restart and paralell IO.

prckent:
Tests are due by end of June. Fixes later.

prckent:
Provisionally added as a requirement to the next release (3.1) particularly in light of bug referenced in #2 . Would prefer to have fixed unless it looks to be overly complicated.

ye-luo:
I have prepared the test case for checking restart. Not integrated into ctest yet.
The electron coordinates and random numbers seem loaded correctly.
1) The extra VMC block (no electron move) and the restart run with the same VMC block agree on KE, ElecElec, LocalECP but NonLocalECP is slightly different. I can't trace all the places where random numbers are used.
2) Run the restart run many times and I got exactly the same results even the initial random seed is not fixed.

So I will set up the checking based on the energy components except NonLocalECP.

prckent:
I'll close this since the VMC restart tests are incorporated ( #249 ). DMC restart currently also looks clean ( #267 ) so we'll need a specific reproducer/bug report with inputs to make progress, if still needed.

PR #249, opened 2017-06-04
Add a restart test

ye-luo:
For issue #232


markdewing:
okay to test

prckent:
Will merge Monday afternoon, after allowing others to comment but still ahead of the nightlies running.

Note that these are VMC tests only at this point, an easier case than DMC.

ye-luo:
If we define restart as loading the random number and electron configurations. There is no difference between VMC and DMC. If we need to fully check every restart behaviour, it will be hard.

prckent:
We need to specifically check DMC because restart problems have been observed  ( #2 ). We also need to create tests for any functionality that is likely to be refactored or that we will have collaborators work on. Ideally we would have tests for all functionality, but for now we prioritize.

prckent:
Merging, DMC tests still needed.

Issue #253, opened 2017-06-06
ReadFileBuffer needs fix


ye-luo:
The unit test crashes in MPI when pp file is missing. This is introduced in #251
Error message cannot be printed correctly when a PP file is missing.
The reason is that is_open is not initialized and always set in open_file.
is_open must be initialized to false and set to false if an open failed.
open_file should check if a file is open or not before try to open anther file.

ye-luo:
Please also avoid pseudopotentials/BFD/C.BFD.xml. This a bad file with nan in it. Though it doesn't really matter for this test, it is better to choose another one considering more tests may need this file.
See issue #43

ye-luo:
Or just overwrite the bad one by a good one from the diamond tests.
@prckent can we just overwrite the bad one instead of waiting for issue #43.

ye-luo:
APP_ABORT  message is not consistent with the actually routine.

ye-luo:
Can get_file_length be a member function of ReadFileBuffer?
Can we move ReadFileBuffer to individual files instead of mixing with ECPComponentBuilder class?
Have it in src/Utilities and we can probably use it in other cases.

markdewing:
The C.BFD.upf file as a 'nan' in it, but the C.BFD.xml file does not.

ReadFileBuffer could be moved to it's own file - something to do later, though.

I'm uncertain about get_file_length between OO design and functional design. 
  In OO design, the method operates on member variables (fin, length) and changes them.  I find this couples the function to the class more tightly and makes it harder to understand.
  In functional design, the function should be understandable in terms of inputs and outputs.  This function works well for this.
I can make it it a member function of the class, but I'll make it more functional (ie, it will not operate on the member variables) - that should keep it easier to understand w/o referencing the whole class.

prckent:
How often do we need get_file_length or similar functionality?

prckent:
Closed by #254

PR #254, opened 2017-06-07
Initialize is_open value in ReadFileBuffer.

markdewing:
Address comments in issue #253.
 - initialize the value of is_open to false by default
 - Update the filename in the APP_ABORT messages
 - change get_file_length to be a member function.

### DIFF
--- a/src/QMCHamiltonians/ECPComponentBuilder.h
+++ b/src/QMCHamiltonians/ECPComponentBuilder.h
@@ -86,10 +86,12 @@ class ReadFileBuffer
   char *cbuffer;
   std::ifstream *fin;
   Communicate *myComm;
+  int get_file_length(std::ifstream *f);
+
 public:
   bool is_open;
   int length;
-  ReadFileBuffer(Communicate *c) : cbuffer(NULL), fin(NULL), myComm(c), length(0) {}
+  ReadFileBuffer(Communicate *c) : cbuffer(NULL), fin(NULL), myComm(c), is_open(false), length(0) {}
   bool open_file(const std::string &fname);
   bool read_contents();
   char *contents() { return cbuffer; }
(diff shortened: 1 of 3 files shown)


ye-luo:
In open_file, set is_open to false if an open attempt fails.
Also check if fin has been allocated, free it first and new after. Otherwise, memory leaks.

src/QMCHamiltonians/ECPComponentBuilder.h
```diff
@@ -86,10 +86,12 @@ class ReadFileBuffer
   char *cbuffer;
   std::ifstream *fin;
   Communicate *myComm;
+  int get_file_length(std::ifstream *f);
```
ye-luo: Add const if this routine doesn't touch member variables.

markdewing:
The behavior for re-using this class is undefined, since it's currently not used that way.
Options
1. Allow only once - immutable.   Attempting to call open_file if another file is open will return false.
2. Allow reopening a file.   Then the open_file method close/delete fin and delete the buffer if allocated.

I would be inclined to go for (1), because it would simplify the usage.

src/QMCHamiltonians/ECPComponentBuilder.h
ye-luo: How about remove the argument and directly operates on fin?

ye-luo:
The fact that open_file operates differently from other standard "open" provided by languages is a potential trap.
1) if return false, the meaning of false is different from a failed false.
2) Just a few more lines of code and it is consistent with common knowledge.

src/QMCHamiltonians/ECPComponentBuilder.h
markdewing: How to write get_file_length was mentioned in the bug report on OO vs. functional.   Yes, it could operate directly on fin and write to length.     However, I think that makes it harder to understand in isolation vs. using a more functional style.

PR #255, opened 2017-06-07
Using Intel MKL with gcc in manual

prckent:
Added example to manual. This is the recipe used on oxygen every night for gcc_mkl.


manual/installation.tex
```diff
@@ -333,6 +333,31 @@ \subsection{Build scripts}
...
+in finding these libraries, but specific combinations can require
+additional hints, as described below:
+
+\subsubsection{Using Intel MKL with non-Intel compilers}
+
+To use Intel MKL with, e.g. an MPICH wrapped gcc:
+\begin{verbatim}
+cmake \
+ -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx \
+ -DBLA_VENDOR=Intel10_64lp_seq -DCMAKE_PREFIX_PATH=$MKLROOT/lib \
```
ye-luo: May be better to use threaded MKL Intel10_64lp

prckent: There is not a clear story for threaded MKL. Threading helps the optimizer but hurts the Slater determinant update and coefficient application in molecular calculations. Since we use the above every night, lets keep with this version unless you have data indicating that we should do otherwise.

ye-luo: "Threading hurts the Slater determinant update" This is not ture.
Inside OMP region, threading is off by default.
With intel compiler -mkl has the threaded one by default.

prckent: We have to check what happens with gcc. Do you know?

ye-luo: You are right. The behavior of threaded MKL with gcc is unknown to me. Leaving the sequential MKL is safer.

PR #260, opened 2017-06-09
Output more information for estimator failures

markdewing:
Print some information about estimator failures - see if it's a tolerance issue, or something more serious.


markdewing:
The error message in the output would be different if pandas were not available - there would be an import error.  The current error message indicates the closeness check is failing, but gives no details.

Issue #219, opened 2017-05-20
Possible bug in DMC of the complex version on GPUs


maxamsler:
I believe I found a bug in QMCPack when using the complex version compiled for GPUs. The energies obtained with a DMC simulation in a periodic system with twists do not agree with the values obtained with the CPU version. This issue arises both with the double and mixed precision version of the GPU code.  I have attached the nexus file for a test case of a Si64 supercell. The twist averaged DMC energy for the CPU version results in:

qmca  -a -q ev -e 50  silicon/dmccpu_64_444/*001.scalar.dat
                            LocalEnergy               Variance           ratio 
avg  series 1  -252.106622 +/- 0.000573   1.820444 +/- 0.000643   0.0072 

This energy value is significantly lower than the value obtained from the GPU version:
qmca  -a -q ev -e 50  silicon/dmcgpu_64_444/*001.scalar.dat
                            LocalEnergy               Variance           ratio 
avg  series 1  -252.084855 +/- 0.000572   1.792653 +/- 0.000634   0.0071 

The bug does NOT seem to affect VMC calculations, since both CPU and GPU versions give results within the statistical error from each other.

The calculations were performed on Daint at the Swiss National Supercomputing center CSCS, a Cray XC50 with a NVIDIA Tesla P100 with 16GB per node. The following modules were loaded to compile QMCPack:
module switch  PrgEnv-cray/6.0.3 PrgEnv-intel
module switch intel/17.0.1.132 intel/16.0.3.210
module load cray-hdf5
module load daint-gpu
module load cudatoolkit

I configured the make as follows:
cmake -DCMAKE_C_COMPILER=cc -DCMAKE_CXX_COMPILER=CC -DQE_BIN=/project/s700/QMCPack/CPU/qmcpack/external_codes/quantum_espresso/espresso-5.3.0/bin -DQMC_COMPLEX=1  -DLibxml2_INCLUDE_DIRS=/project/s700/local/include/libxml2/ -DLIBXML2_LIBRARIES=/project/s700/local/ -DBOOST_INCLUDEDIR=/project/s700/boost_1_61_0/ -DQMC_INCLUDE="-I/project/s700/local/include/libxml2" -DQMC_EXTRA_LIBS="/project/s700/local/lib/libxml2.a /project/s700/local/lib/liblzma.a" -DQMC_CUDA=1 -DQMC_MIXED_PRECISION=0

[...]

PR #262, opened 2017-06-12
nexus: better guard for structure file and internal consistency

jtkrogel:
Add a better guard to alert the user against missing structure files (per prckent's comment on #219 ).  Also put checks in place to ensure the internal consistency of physical system objects (checked when physical system is passed to a simulation).  This prevents Nexus from running a simulation when the user has corrupted a physical system object.  

Other changes: 
1) Small fix for homonuclear reduced mass for a Morse potential
2) Added complete pwscf functional list.  Warning generated for unrecognized functional, vdW_kernel_table file demanded for vdW functionals.


prckent:
ok to test

nexus/library/structure.py
```diff
@@ -766,6 +766,26 @@ def __init__(self,axes=None,scale=1.,elem=None,pos=None,mag=None,
     #end def __init__
 
 
+    def check_consistent(self,tol=1e-8,exit=True,message=False):
```
markdewing: The 'message' flag doesn't seem necessary, and could be removed to simplify the function.   The function can always return (consistent,msg).  If the caller doesn't want the message, it can ignore it (a single underscore is a valid variable name, so it's often used to represent an unwanted return value, giving something like "consistent, _ = self.systemcheck_consistent(exit=False)"

Similarly for check_folded_system.

jtkrogel: This is true, but it can help for the UI.  The "message" part is only really relevant to a message reporting system; most of the time someone writing a nexus user script will use s.check_consistent(), or "if s.check_consistent(exit=False): ...".  There are a few user facing functions elsewhere in nexus that allow variable output lists depending on user request.

markdewing: I see.  I didn't realize this was a user-facing function, but that makes sense.

PR #265, opened 2017-06-15
Use full precision in ForceCeperley::InitMatrix

ye-luo:
Solving force unit test failure
  REQUIRE( force.c[i] == Approx(coeff[i]) )
with expansion:
  4369.5f == Approx( 4375.0 )


markdewing:
okay to test

Issue #258, opened 2017-06-08
Restart-8-2 failing due to numerical tolerance issues


prckent:
https://cdash.qmcpack.org/CDash/testDetails.php?test=485831&build=7444

> Test output
> 2c2
> < 1.1318030733e+01 -2.9875893977e+00 -1.2775667611e+01 -6.5545628484e+00
> --- 
> 1.1318030733e+01 -2.9875893981e+00 -1.2775667611e+01 -6.5545628487e+00
>
It would help to have a more verbose error on this failure. The VMC restart is clearly not exact in this case.

ye-luo:
Such errors appear sometimes on mixed precision codes both CPU and GPU.
The difference is mainly due to single precision round error.
The checking script needs to be a bit more tolerant.

prckent:
Good that this is a tolerance problem. This test has to be nearly 100% reliable since it is the kind of test that could become part of our CI tests or will be used by non-specialists.

PR #268, opened 2017-06-16
Increase restart check tolerance

ye-luo:
Printout is more verbose.
Check up to 7th digit after the decimal.

Issue #258

Because I was not able to reproduce the failure on my platforms, I only simulated the case for checking the fix.


markdewing:
okay to test

prckent:
Should this print a FAIL if there is a difference?

prckent:
I verified this new restart checker, including using a modified check.sh to ensure failures are caught.

ye-luo:
@prckent This checker doesn't print Fail but returns a failure value which is picked up by ctest.

PR #270, opened 2017-06-17
Test VMC with drift at non-gamma twist that real code can still run

prckent:
Missing from our testing and added due to the issue found while checking #219. Currently we test gamma (too easy) and an arbitrary twist that only the complex code can run. This test should run successfully with the real code as well as the complex code.

### DIFF
--- a/tests/solids/LiH_solid_1x1x1_pp/hf_vmc_LiH-x-drift.xml
+++ b/tests/solids/LiH_solid_1x1x1_pp/hf_vmc_LiH-x-drift.xml
@@ -1,0 +1,45 @@
+<?xml version="1.0"?>
+<simulation>
+  <project id="hf_vmc_LiH-x-drift" series="0">
+  </project>
+  <!--random seed="49154"/-->
+
+  <qmcsystem>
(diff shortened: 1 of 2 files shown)


ye-luo:
I thought about the same VMC test with drift and did the run manually and the result was correct.
I noticed that the LiH VMC runs are much longer than the DMC runs on GPU.
I'm considering taking the chance to revise all the LiH VMC tests. Reduce blocks to 100 but use 10x walkers. So it can run faster on GPU and write less lines to the disk.
Now each VMC>5m and we have gamma, x, arb.
https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=short-LiH_solid_1x1x1_pp-gamma-drift-vmc_hf_noj-1-16&date=2017-06-16

prckent:
@ye-luo I was dissappointed that this test did not surface the problem with DMC too, but the test is still worth having

Improving the statistical reliability of the short test should be our priority. This may need more equilibration, more blocks, perhaps longer runs. When/if we revise the tests to improve the reliability we can consider extra walkers for the reasons you state.

PR #271, opened 2017-06-17
Coverage - mark lines from files not in unit tests correctly

markdewing:
For files that are completely untouched in the unit tests, but have some execution
in the the base tests, the behavior was to mark all the executed lines as uncovered.
This would incorrectly handle lines that are not executed (uncovered) in the base.
These lines should be marked as 'no-code' (not counted in coverage) in the output
directory.

This was most obvious in the case of formic/utils/lmyengine/engine_timing.cpp.
Even though the new optimizer is not called by the base test, there are some static
initializers that get called.   All the executable lines of the file were marked
as uncovered in the output, which was incorrect.  After the fix, only the lines
called by the static initializers are marked as uncovered.

With this fix the percentage coverage rises to 67% locally (probably will vary
some from that on the official test system)

### DIFF
--- a/tests/coverage/compare_gcov.py
+++ b/tests/coverage/compare_gcov.py
@@ -124,13 +124,16 @@ def mark_as_uncovered(gcov_base):
   for line in gcov_base.line_info.iterkeys():
     base_line = gcov_base.line_info[line]
     Uncov_norm= '#####'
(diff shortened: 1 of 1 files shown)


prckent:
+2

PR #275, opened 2017-06-19
nexus: add Cori at NERSC

jtkrogel:
Add support for Cori at NERSC.  Differentiate between KNL and Haswell partitions.  Leave finer details, e.g. "OMP_PROC_BIND=spread", to user input.


prckent:
OK to test

PR #285, opened 2017-06-28
Add merging of gcov files.

markdewing:
Fixes because include files do not get coverage handled correctly for unit tests.  Only the coverage information corresponding to one executable is included, the coverage information from all the others is overwritten.  The coverage information should be combined.

There are two types of files produced for coverage:
  .gcno - produced at compile-time, contains notes for translating basic blocks to source lines
  .gcda - produced at run-time, contains the counts for basic blocks and arcs between then (for branch coverage)

The .gcda files can be merged, but only for multiple runs of the same executable.
(The .gcda files get automatically merged at run time, and newer versions of gcc have a 'gcov-tool' that can merge .gcda files)
This works fine for the system tests, since they all run 'qmcpack'.
For the unit tests, they cannot be merged because they come from different executables.

CMake puts object files for different executables in different directories (even for the same source file), and the .gcno and .gcda files follow, so nothing gets overwritten during collection.

However, while running 'gcov', these files get processed into files with the '.gcov' suffix, and these can get overwritten.  It's particularly bad for include files, because they don't even have a separate .gcno/.gcda file. (The collected data for one object file can generate multiple .gcov files, usually for included files)

Gcov has a -l (--long-file-names) switch to help with this.  The .gcov filename has two parts, separated by '##'.  The first part contains the cpp.gcda file the gcov file originated from, and the second part contains the usual source file name.
The -p (--preserve-paths) option keeps all the path components, separated by '#'.
Combining the -l and -p options with the CMake directory layout, the (long!) gcov file name is unique.

The next step is to combine the data at the .gcov file level.  This is the purpose of merge_gcov.py.   Lines are combined based on the priority: covered > uncovered > no code.

The collection script is modified to first put the .gcov files in a 'raw' directory where files have the two-part name.  Then those files are merged and put into the previously used directory (with a one-part name), and the processing proceeds as before.

This commit only fixes the standalone coverage script.  The coverage processing for the nightly runs will be a future commit.

### DIFF
--- a/tests/coverage/compare_gcov.py
+++ b/tests/coverage/compare_gcov.py
@@ -6,6 +6,7 @@ import os
 import sys
 import demangle
 import read_gcov
+import merge_gcov
 
 def get_gcov_files(dir_name):
   """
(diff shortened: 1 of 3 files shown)


prckent:
+1

Hopefully a variant of this merging process can be used for MPI

PR #286, opened 2017-07-05
Add merging of gcov files to the nightly tests

markdewing:
Some additional changes from the previous commit for merging gcov files:

With the --long-file-names (-l) and --preserve-paths (-p) options to gcov,
some of the file names would exceed the filesystem limits (255 characters).
One solution for shortening the names is to use the -source-prefix (-s) option
which removes some of the path elements in the name.  The shortened path is
used in the file name and in the Source: tag in the file.
This causes problems with gcovr, so the path prefix is restored to the Source:
tag when merging gcov files.

The processing pipeline for gcov files has an extra step - the long file names
are written to the tgcov_*_raw directories, and then merged to the corresponding
tgcov_* directory.

Many files have some executed code related to static initializers. It clutters
the output to have those included in the output, even though they only have one
line in them and it's considered covered.  Add files of this type to the files
to be removed in 'remove_unwanted_file'.

The gcov file merge and write does not yet handle branch information, which also
includes function headers.  These function headers are necessary for filtering the
static initializers.  Add an explicit step to filter the gcov files in the *_raw
directories.


prckent:
Thanks Mark. Good to have these improvements.

PR #304, opened 2017-07-26
compatiple with ipython & local machine setting

zoowe:
"exit_call = exit" does not work in ipython. Make it become sys,exit
Add an option for not modifying the code while adding new machine to nexus.


jtkrogel:
This looks OK to me.  +1

markdewing:
okay to test

prckent:
Thank you!

Issue #290, opened 2017-07-12
unit_test_estimators numerical issue with Intel compiler


ye-luo:
https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=unit_test_estimators&date=2017-07-12

All the GNU builds are fine but the Intel builds are not. I think we need to make the checking a bit more flexible to the numerics.

prckent:
Is a full precision comparison used? The error messages show full agreement up to the number of printed digits.

/data/ci/scratch/QMCPACK_CI_BUILDS_DO_NOT_REMOVE/qmcpack/src/Estimators/tests/test_accumulator.cpp:153: FAILED:
  REQUIRE( mv.second == a.variance() )
with expansion:
  5.1909181302 == 5.1909181302

ye-luo:
Showing all pass
https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=unit_test_estimators&date=2017-08-04

PR #309, opened 2017-08-01
Add tolerance to compares for estimators unit test

markdewing:
Addresses #290

mean() and mean_and_variance() duplicate the formulas and can have slight differences as a result (divide vs. reciprocal and multiply).   An better eventual fix would be to use identical formulas in both methods.


prckent:
(Waiting for CI to reactivate after OLCF maintenance)

grahamlopez:
test this please

PR #311, opened 2017-08-02
nexus: better support for density file postprocessing, support for QE…

jtkrogel:
… angle keywords

Add functions to compute norms and make line plots of density data to XsfFile class.

Support QE keywords "angle1" and "angle2" with values assigned by species keyword/label.


markdewing:
okay to test

PR #289, opened 2017-07-12
Incorporate SoA changes

ye-luo:
This PR is huge but it is better to merge it now than later.

SoA code paths are not activated by default.
No changes to the known issues and no new issues spotted so far.

To access SoA optimization, add -D ENABLE_SOA=1 in cmake.
Improvements:
Significant reduction of memory usage.
2X or more speed up on most CPU architectures.
Better mixed precision support.
Known issues:
Localized basis set support is not ready in SoA.

ToDo before next release:
1) Improve GNU compiler configurations and Cray handling.
2) Add the description of SoA in manual.
3) Enforce C++11 in CMake.

Issue #229, opened 2017-05-25
Testing: Status of all-particle move QMC methods not clear; tests needed


prckent:
QMCPACK supports all-particle move based algorithms as well as the conventional particle-by-particle algorithms. The state of these is unclear as identified by #227 . VMC all electron is now working.

- Which DMC algorithms work?
- Is there pseudopotential support?

A first step would be to try all particle move versions of the current tests. For unimplemented methods we should add error traps and update the manual. e.g. GPU support.

rcclay:
I briefly checked the VMCUpdateAll w/ and w/o drift a few weeks ago, and found them to be operational and consistent with the pbyp numbers.  Did PR #227 turn up any numerical issues when using electrons?

prckent:
@rcclay Please do

rcclay:
Preliminary tests are in.  I considered the 1x1x1 Diamond example used in the ctest runs.  512 walkers, time step of 0.005 Ha^-1.  Still trying to beat down error bars on my workstation, but:
1.)  DMC Alle driver w/ pseudopotentials and w/o T-moves is good.  Agrees with PbyP for every observable.
2.)  DMC Alle driver w/ pseudopotentials and w/ T-moves is uncertain.  Total energy agrees with PbyP, but all other quantities are off.  Example with the potential energy:  PbyP says <V>=-22.0265(19), but Alle says <V>=-22.0584(42) (Alle is still running, hence the larger error bar).  Either way, right now it's a 6 sigma difference between the two.  The reason I say the status of this driver is uncertain is because I haven't ruled out that it might be a timestep error.

rcclay:
Update:
2.)  DMC Alle driver w/ pseudopotentials and w/ T-moves seems to work, however, the timestep error is large.  Specifically, I ran all electron calculations at t=0.005, 0.0025, and 0.001.  Performing a linear time step extrapolation, I find the all electron results estimate <V>=-22.021(4), which is within error bars of the pbyp estimate.  Discrepancies in all other observables likewise vanish.

Can't tell if the large timestep error is a bug or a feature.  Whatever it is, the DMC Alle driver (with T-moves) gives the correct results in the tau->0 limit.  Perhaps its use should be discouraged unless there is a compelling reason to do otherwise.

prckent:
Current status: VMC and DMC all electron tests for bccH now included. PP tests still required.

ye-luo:
@rcclay What system exactly were you using?
[...]

PR #312, opened 2017-08-07
Fix Tmove

ye-luo:
1) A bug was introduced in PR #289 affecting Tmove in both AoS and SoA cases.
This fix is validated against v3.1.1.
The current tests labelled dmcnl are all short tests. Need long ones.

2) Add ENABLE_SOA to ctest_script.cmake


markdewing:
okay to test

CMake/ctest_script.cmake
```diff
@@ -74,6 +74,12 @@ ELSEIF( ${HOSTNAME} MATCHES "billmp1" )
     ELSE()
         SET( CTEST_BUILD_NAME Clang-Release-Mac )
     ENDIF()
+ELSEIF( ${HOSTNAME} MATCHES "ryzen-box" )
```
markdewing: Did you intend to check in this code?

ye-luo: Yes. Maintaining a local change is painful.

prckent: OK if it makes development+testing easier. It does raise the question about why these settings are needed, but that could be asked of any of the machine settings in the ctest script...

src/QMCDrivers/DMC/DMCUpdatePbyPFast.cpp
```diff
@@ -192,22 +189,27 @@ void DMCUpdatePbyPWithRejectionFast::advanceWalker(Walker_t& thisWalker, bool re
...
-      Psi.acceptTMove(W,iat,w_buffer);
-#if 0
-      RealType ratio=Psi.ratio(W,iat,dG,dL);
-      W.G += dG;
-      W.L += dL;
-      //thisWalker.R[iat]=W.R[iat];
-      //w_buffer.rewind();
-      //W.copyToBuffer(w_buffer);
-      RealType logpsi = Psi.evaluateLog(W,w_buffer);
+#ifdef ENABLE_SOA
```
prckent: Please check the reptation implementation also. See @rcclay 's comment about earlier tests.

ye-luo: RMC doesn't have Tmove enabled. The UseTMove was disabled by @rcclay in 2014.
pbyp and allp moves of AoS and SoA builds are both checked against v3.1.1. See my comment on #229

prckent: Good. So Tmoves will need investigating by a future reptation developer/user, and they can't be run accidentally.

PR #313, opened 2017-08-07
Clean up git-rev-tmp.h after each build.

markdewing:
This should ensure the commands to build git-rev.h are triggered on the next build.
This might help with some of the Spack issues.


naromero77:
Thanks Mark.

PR #314, opened 2017-08-07
Optimizable determinants, finite-difference linear response, nodeless guiding functions

nsblunt:
This request is for three features developed in Eric's (@eneuscamman) group.

The main branch is called "nodeless", and implements nodeless guiding functions, in order to give a more honest (more Gaussian) distribution of samples for the variance of the energy. I was not involved in this, this is entirely Eric's (@eneuscamman) work. Perhaps he can give more input if needed about this feature?

We wished to use this nodeless guiding function with the finite-difference linear response (FDLR) wave function that I'd developing (in a separate older version of QMCPACK). I therefore created new branches off of nodeless.

A paper describing the FDLR wave function is on arXiv: https://arxiv.org/abs/1707.09439

The implementation of FDLR uses optimizable determinants (determinants with orbital optimization). This had also been coded up by Eric in the "older" version of QMCPACK with which the group was working previously. I started by copying this across (commits 0345c04 to f82303f), and also fixing a bug I made previously, d8fed21.

The rest of the commits are mostly copying across the FDLR code.

In commit 923d755 I've touched a few bits of code away from the self-contained work. This was to replace NULL and 0 by nullptr, as I found necessary to compile QMCPACK on NERSC's Edison machine - presumably a C++11 requirement?

I also added some new member functions to the VariableSet class, to allow two parameter sets to be added or subtracted. This is required for the FDLR wave function, where the determinants are evaluated at orbital rotations "x+d" and "x-d", where "d" is a small orbital rotation, for the finite difference approximation to be valid.

There are changes to WaveFunctionFactory.cpp, to deal with initialization and input files.

Most other additions should be self contained within FDLRWfn.cpp and FDLRWfn.h.

Best,
Nick


markdewing:
Hi Nick, 
 Looks like some good additions here.  However, it would be easier to review and understand if it were broken up into several smaller pull requests.   Perhaps the nodeless guiding function as one PR, one for optimizable determinants, and one for the FDLR wavefunction?

For each of these features, there should be some tests and documentation on how to use them.   An example input file for each feature would be great start.  If you have some input files, I can help you put them into the right place and integrate them into the test system.

What is the consequence of using the nodeless guiding function?  You mentioned it makes the variance more Gaussian, but then what?  Does it make the optimization process better or more stable? Or something else?

Are the optimizable determinants general, or does it only work for the FDLR wavefunction?

Other items:
 - there are 'tags' and 'typedef' files that should not be checked in (ideally, rewrite history to remove the files from the commits they occur in)
 - the changes to nullptr to get it to compile.  The older NULL (and 0) should be backwards compatible in C++11. I would like to understand more about why these aren't compiling - what compiler are you using, and what is the error message?

ye-luo:
Hi Nick,
Nice to see new features appearing in QMCPACK and welcome.
As Mark pointed out, I would also like to see PRs for individual topics. And it seems feasible.
For the moment, I saw quite a few hacks to get things working. This was happening probably we don't have a good documentation for developers.
I think we can work it out with an iterative process.

nsblunt:
Hi Mark,

I agree it would be easier and cleaner as three PRs. Given that the optimizable determinant and FDLR code is branching from this nodeless branch, how would you recommend I best do this? Should I cherry pick the nodeless commits to a new branch and send a PR for that. Then once that is merged, rebase the opt-det and FDLR commits on new branches from there? I'm new to the github PR system, and unsure of best practice.

I'll add some tests for these options. I'll get some input files ready first, then may ask for help on doing this.

I believe the main motivation for the nodeless guiding function comes from the new target function used to optimize excited states (http://pubs.acs.org/doi/abs/10.1021/acs.jctc.6b00508). In practice, for a non-exact ansatz, we initially set the shift, \omega, to E -\sigma (energy minus standard deviation). The error on this needs to be smaller than the gap between excited states to choose it appropriately, so it's important to have a reliable estimate of the error on the variance. Previously this was hard to estimate because the variance profile had very large spikes (very skewed distribution). With this approach the error bars looks larger, but are more honest, because these spikes aren't present and so closer to Gaussian. I'll ask Eric for input on this, since I haven't worked on it myself.

The optimizable determinants are general, they're not just for the FDLR wave functions. You can take a Slater-Jastrow wave function for any system and optimize the orbitals.

I'll correct the commits to remove the unwanted files, sorry about that.

nsblunt:
Hi again Mark,

I tried to reproduce the compilation error that I was having on NERSC's Edison, but am unable to now. Edison had a big upgrade recently, with the OS having been replaced.

I went back to the commit before I made this pointer change, and it now compiles fine.

I was using Intel compilers, and from my memory I had the problem for every version that I tried, available on Edison. It occurred to me that this error was very odd... If it was a real error then it seemed unlikely that it wouldn't have been caught already.

I'm happy to either leave or remove this commit, as you like.

nsblunt:
Hi Ye,

Thanks very much for your comments. If you point out to me anything that is a hack then I'm happy to fix it.

ye-luo:
The nodeless driver add ion particle set and even hard-coded ion0 as the name.
The distance tables are computed from scratch but only open bondary contions.
There is walker counter and restricting only one walker per thread?
It probably needs many changes.
Not look at the other parts yet.

jtkrogel:
Side comment here: its nice to see the option for a nodeless guiding funciton.  I would expect benefits from having the guiding and trial wavefunction separate, expecially for optimizing multideterminant wavefunctions as this could be used to reduce undersampling when introducing excited state determinants.

ye-luo:
@jtkrogel The nodeless guiding wavefuction sampling sounds like a special case of the reweighting method similar to correlated sampling. I thought we already have it or something close.

jtkrogel:
@ye-luo Don't know if we already have alternate guiding functions.  As far as I know reweighting can't really fix undersampling, at least the type I am thinking of (e.g. you have a guiding wavefunction with only 1s and you are trying to add/sample a configuration w/ 2s which has exponentially small number of samples according to 1s function).

prckent:
@nsblunt Excellent. Indeed this would be easier as a few separate PRs. However as mentioned by @markdewing it is absolutely critical that we have some tests of this functionality. i.e. Small inputs and reference outputs that we can run quickly (see all the nightlies run on https://cdash.qmcpack.org/CDash/index.php?project=QMCPACK ). Do you have anything suitable?

Tests (as well as docs) are crucial to keep your contributions alive in the future and prove that no one else has broken them. Our integration tests currently all use 16 cores (as 16 threads or 16 mpi, and combinations thereof) and last no more than a few minutes, just enough to get reliable statistics on a few electron test case.

It is very likely that untested code -- essentially any code where we don't have trusted reference inputs and outputs -- will not make it to the next major version of QMCPACK.

Since I am on travel I'll let @markdewing and co determine how to best merge your contributions. Thanks again.

markdewing:
@nsblunt , your idea for arranging the commits and PR's sounds good (nodeless first, then the others)

And I think it would be better to remove the C++11 nullptr fixes.  Eventually we will be using C+11 everywhere, but it would be nice to be more systematic about it, and removing this commit means the PR will touch less code.

prckent:
C++11 should be a required feature now, after the SOA merge, even when SOA is not activated. We should enforce this in the CMakeLists.

prckent:
@nsblunt Let me know the current plans. Will it be possible to separate this PR into separate pieces as described above? Should I close this PR?

eneuscamman:
Regarding nodeless guiding functions, or alternative guiding functions more
generally, there are many features one might want that are not covered by
the current code base.  What is contained in the code Nick is working with
is a pretty simple hack, and not necessarily essential going forward.  If
it can be ring-fenced from the optimizable determinant and his FDLR code,
it might make sense to leave it out.  Another student of mine is exploring
making a more general and responsible system for alternative guiding
functions, and if that goes somewhere we could include it later.

nsblunt:
@prckent I'm sorry for the slow response, I am going to send a new PR this morning. As Eric says, he and I discussed it and decided not to merge the nodeless guiding work for now. But I will send a PR for the optimizable determinant code today.

I think that this PR can be closed.

prckent:
@nsblunt No problems with response speed. This is a tricky issue to work through.

Closed.

PR #315, opened 2017-08-08
nexus: add makov payne correction, make plotting robust

jtkrogel:
Add ability to calculate the first order Makov-Payne correction for any cell.  Supported in new function 'makov_payne' in structure.py.

Also improve robustness of plotting generally, including tools like qmca.  Previously plots would fail in some environments (e.g. Titan at OLCF) due to default backend issues.  This is now fixed.  Fix contributed by Kayahan Saritas.

### DIFF
--- a/nexus/executables/qmca
+++ b/nexus/executables/qmca
@@ -1372,15 +1372,31 @@ def convert(value,source_unit,target_unit):
 
 
 
+success = False
 try:
+    import matplotlib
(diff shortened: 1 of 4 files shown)


prckent:
Are there any plans to add more recent schemes?

prckent:
ok to test

jtkrogel:
I don't have direct plans, but I am open to suggestions.

prckent:
e.g. The Feysoldt defect alignment scheme or its various extensions. Ganesh, Ram must have opinions.

PR #317, opened 2017-08-11
Lattice analyzer multi-precision bugfix 

rcclay:
This pull request addresses bugs in LatticeAnalyzer::find_shorter_base() and secondarily removes the do...while loop in LaticeAnalyzer::find_reduced_basis(), which in conjunction with the bugs in find_shorter_base(), can lead to an infinite loop.  This infinite loop was found using multi-precision and the following unit cell:

`<parameter name="lattice" units="bohr">
        7.37924194        2.76721573        2.76721573
        2.76721573        7.37924194        2.76721573
        2.76721573        2.76721573        7.37924194
  </parameter>` 

**Bug fixes in LatticeAnalyzer::find_shorter_base()**
1.)  Changes the definition of eps tolerance to reference the floating point type used in the definition of the lattice vectors, and not "float".  
2.)  For the lattice vector replacement condition ((r2-r2max) < -eps), replace eps with tol=2*sqrt(r2max)*eps, which is the correct propagation of errors in the expression x^2.  

**Improvement in LatticeAnalyzer::find_reduced_basis()**
a.)  I changed the do...while(1) loop to a loop over some hardcoded maximum iteration (maxIter=10000) for now.  I figure if QMCPACK can't find a reduced cell by then, the user should check their unit cell or contact a developer.

### DIFF
--- a/src/Lattice/LatticeAnalyzer.h
+++ b/src/Lattice/LatticeAnalyzer.h
@@ -314,7 +314,9 @@ inline bool found_shorter_base(TinyVector<TinyVector<T,3>,3>& rb)
 template<typename T>
 inline void find_reduced_basis(TinyVector<TinyVector<T,3>,3>& rb)
 {
-  do
+  int maxIter=10000;
+ 
+  for(int count=0; count<maxIter; count++)
   {
     TinyVector<TinyVector<T,3>,3> saved(rb);
     bool changed=false;
@@ -329,7 +331,7 @@ inline void find_reduced_basis(TinyVector<TinyVector<T,3>,3>& rb)
     if(!changed && !found_shorter_base(rb))
       return;
   }
-  while(1);
+  
 }
 
 }
--- a/src/Lattice/LatticeAnalyzer.h
+++ b/src/Lattice/LatticeAnalyzer.h
@@ -279,7 +279,7 @@ struct LatticeAnalyzer<T,1>
 template<typename T>
 inline bool found_shorter_base(TinyVector<TinyVector<T,3>,3>& rb)
 {
-  const T eps=10.0*std::numeric_limits<float>::epsilon();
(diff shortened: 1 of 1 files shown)


ye-luo:
**Bug fixes in LatticeAnalyzer::find_shorter_base()**
Does this bug affect the results of calculation or just causing infinite loop?

**Improvement in LatticeAnalyzer::find_reduced_basis()**
If maxIter is reached, is it better to abort the code and issue an error instead of keep running silently?

rcclay:
1.)  I don't know yet.  In the case described above, the round-off errors made it so that (for example) the vectors  **a** and **-a** had different magnitudes, and so it would cycle between those two in what it thought was vector with the smallest magnitude.  That would just cause an infinite loop.  A more scary scenario I could imagine would be if distance table AA initialization ended up picking one set of reduced lattice vectors, and distance table AB picked a different set.  I don't know this part of the code well enough to know what would happen in this case, but we probably need to look into it.  

2.)  I feel like its far better to abort the code with a helpful error message.  I had originally tried to slap an APP_ABORT() after the for loop, but this introduced a linker error into the test_lattice.cpp unit test (but qmcpack built and worked according to plan).  I'll create an issue for it and tackle it when I have a bit more bandwidth.

ye-luo:
1) As long as you spot the issue by encountering an infinite loop, I'm not worried. Distance tables AA AB all use the ParticleSet lattice which is taken from the global cell. It is safe.
2) Please create an issue as you described.

rcclay:
Created issue #318 in response to 2.).

prckent:
@ye-luo No infinite loops please! This could be very costly in a parallel run. The code should abort and not hang.

prckent:
OK to test

ye-luo:
@prckent "Not worried" because it is not causing false positive results. Definitely, infinite loop is bad.
+1

PR #319, opened 2017-08-11
Optimizable determinants

nsblunt:
Here is a PR for just the optimizable determinant code, mostly by Eric. I used 'git cherry-pick' to take the appropriate commits from the previous nodeless PR.

It failed to compile after trying to cherry-pick the first commit, b747d55. This was because of a change to the simd::transpose routine in d7dcb60, so that the interface no longer matched the call from MatrixOperators::transpose. The first commit (3d2d1ee) therefore updates the latter function call.

Because this code was written a while back and copied across, it's almost all in one commit, sorry about that. I should point out that it was only ever implemented for a single OMP thread. If you try and run with more then it will crash. I imagine you'd like this corrected before merging, though I don't understand the reasons for this not working. Is this likely a simple fix?

I have a couple of tests to add, are there instructions for doing so? I can probably do it mostly through analogy with existing tests in system/CMakeLists.txt, but I can't see how the script extracts benchmarks. It looks like it's using check_scalars.py, but I'm unsure...?


nsblunt:
Sorry, was this a message to me to perform testing? I assumed this would be an automated thing? (I did run the short tests on my workstation and all passed).

markdewing:
No, that was a message to the CI system to perform testing.   We don't have it set up just yet to run automatically when a PR is created.

markdewing:
Following the existing tests in system/CMakeLists.txt should work.
The steps to add a test would be
1. Find a place for the input files (most likely in tests/solids or tests/molecules)
2. Determine which values to check against and create a list in system/CMakeLists.txt
3. Add a call to QMC_RUN_AND_CHECK in system/CMakeLists.txt

QMC_RUN_AND_CHECK is defined in CMake/macros.cmake and there is some description of the parameters there.  This function adds multiple tests at the CMake level - one to run qmcpack, and then one for each scalar value to check (these tests call check_scalars.py)
The check_scalars.py script reads the output values from the *.scalar.dat files produced by the qmcpack run and the expected value (and error bar) is given on the command line to check_scalars.py.

Regarding the OMP issue - once you get an input file in the PR we can look at why it is crashing.

nsblunt:
Thanks for the instructions. I'm currently doing a study with DMC and need to use T-moves. There were a few functions required for this that weren't implemented in this PR, so I am dealing with those currently, and will add them to the PR when done. I've implemented these functions so that jobs run without crashing, but am not convinced of the results yet. I might need advice depending on how this goes.

In the OrbitalBase member function "ratio", with the following interface:

virtual ValueType ratio(ParticleSet& P, int iat,       
                          ParticleSet::ParticleGradient_t& dG,
                          ParticleSet::ParticleLaplacian_t& dL)

dG is described as a "differential" gradient. Can I please check that this simply means the value of G after the update, minus the value before the update (and similarly for dL)?

src/QMCDrivers/QMCCostFunctionOMP.cpp
```diff
@@ -709,6 +709,36 @@ void QMCCostFunctionOMP::resetPsi(bool final_reset)
...
 }
 
+///////////////////////////////////////////////////////////////////////////////////////////////////
+/// \brief  For each component of the trial function, copy that component's optimizable parameters
+///         into it, convert the parameters to a standard form (for objects that need this, most
+///         don't) and then optionally copy the standard form parameters back.
+///
+///////////////////////////////////////////////////////////////////////////////////////////////////
+void QMCCostFunctionOMP::putParametersInStandardForm() {
+
```
ye-luo: The commit log shows this is a hack and can be fixed with the usual way of QMCPACK checking in/out paramaters.

src/QMCWaveFunctions/EinsplineSetBuilder_createSPOs.cpp
```diff
@@ -72,6 +75,7 @@ EinsplineSetBuilder::createSPOSetFromXML(xmlNodePtr cur)
   std::string spo_prec("double");
   std::string truncate("no");
   std::string hybrid_rep("no");
+  std::string optimizable("no");
 #if defined(QMC_CUDA)
```
ye-luo: Since the orbital rotation scheme doesn't touch the einspline SPO. Optimization switch should be introduced outside, probably in the slater determinant builder.

src/QMCWaveFunctions/EinsplineSetBuilder_createSPOs.cpp
```diff
@@ -190,11 +195,22 @@ EinsplineSetBuilder::createSPOSetFromXML(xmlNodePtr cur)
   if ((iter != SPOSetMap.end() ) && (!NewOcc) && (qafm==0))
   {
     qafm=0;
+
+    // if requested, use an optimizable orbital set that has the spline orbital set as its basis
+    if ( optimizable == "yes" ) {
+      app_log() << "using an existing EinsplineSet object (not a clone) for the basis of an optimizable SPO set.\n";
+      return new LCOrbitalSetOpt<LocalizedBasisSet<NGOBuilder::CenteredOrbitalType> >(SPOSetMap.find(aset)->second);
+    }
+
```
ye-luo: I think this can be move to the slater determinant builder.

ye-luo:
Two additional questions:
1) Does the code work with up/down determinants sharing the same optimizable SPO set?
2) How to save the optimized MO coefficients at the moment? In principle, in the LCAO case, the MO coefficients and be dumped at its original place. In the spline case, we can apply the rotation on the spline coefficients and dump it.

ye-luo:
@nsblunt If you are following the most recent develop branch, 
ratio(ParticleSet& P, int iat,
ParticleSet::ParticleGradient_t& dG,
ParticleSet::ParticleLaplacian_t& dL)
has been removed from the DMC driver DMCUpdatePbyPFast.cpp.
The routine in TrialWavefunction has been commented out (to be removed).
I don't understand why you need to reintroduce it.

nsblunt:
The code I was running is a private fork, on a branch which didn't contain the recent changes to develop. I was implementing it because it was previously not implemented and throwing an error when running with T-moves. I didn't realize about this change. I'll swap to running on the updated branch and see if that fixes things.

Thanks for the review, these changes should be okay. Although I'm unsure what you would like regarding the SlaterDetOpt changes. The DiracDeterminantOpt class didn't exist when this was done originally, I don't know what it does. Is it exactly the same thing as this - orbital rotations that can be optimized via the linear method?

To answer your questions:
1. Yes they can share the same SPO set.
2. Currently the orbital rotation parameters are printed out each linear method iteration, and these can be manually copied and pasted into the input file if needed for subsequent calculations. The values printed out are not actually MO coefficients but the optimizable parameters themselves, i.e. those in the matrix in the exponential, which is applied for orbital rotations.

ye-luo:
Don't worry about DiracDeterminantOpt, its status is not clear. I only wanted to point out that the SlaterDetOpt can better placed at the same level as DiracDeterminantOpt.
What are the extra routines implemented in SlaterDetOpt comparing to DiracDeterminantBase? 
evaluateDerivatives must be one.
I'm thinking to have SlaterDetOpt renamed to DiracDeterminantOpt (deprecate the old one) which is a derived class of DiracDeterminantBase. In this way, we can avoid duplicated codes.

nsblunt:
I added a test for H2, optimizing both the orbitals and Jastrow with the adaptive linear method algorithm. I ran the short tests and this new test seemed to pass for me, but please check and let me know. If this is all okay then I'll add another for more code coverage.

markdewing:
The test passes on my machine.
Also, I can reproduce the segfault with OpenMP, and I'll look into it more.

markdewing:
The OpenMP crash is caused by the makeClone method of LCOrbitalSetOpt not copying or setting up the m_tf member.
The makeClone method is used to make copies for each thread.

It the m_tf member is copied, then the crash disappears.  I don't know if this is the correct way to fix it or not.  It depends if m_tf is the same between threads or no.
```
    SPOSetBase * makeClone() const {

      // create a clone that contains a cloned spo set or basis set
      LCOrbitalSetOpt * retval;
      if ( m_spo_set )
        retval = new LCOrbitalSetOpt(m_spo_set->makeClone(), m_report_level);
      else
        retval = new LCOrbitalSetOpt(m_basis_set->makeClone(), m_report_level);

      // set the number of molecular orbitals
      retval->setOrbitalSetSize(this->OrbitalSetSize);
      retval->m_tf = m_tf;

      // return the clone
      return retval;

    }
```

prckent:
Checking on the status: @nsblunt do you have what you need based on Mark's and Ye's feedback? Do you need any extra help/time?

nsblunt:
Hi Paul, sorry for the delay, I'm having some discussion off list with Ye and Eric. I'm currently away at the ACS meeting and back on Friday. The main concern is the extent of changes required to combine SlaterDetOpt with DiracDeterminantBase. I think these ultimately have different data structures, particularly because of how LCOrbitalSetOpt is used by SlaterDetOpt, and so am worried that this requires a significant/total rewrite. My situation is that once I return from ACS, my postdoc position here finishes in two weeks, and during my remaining time I'm performing a DMC study with the trial functions we've been developing, for a paper. I'm also not familiar with the existing DiracDeterminantBase code, so this would take some time to learn. I'm happy to work on this, including after I have left, but it would be most time efficient for me to at least get big DMC calculations running, and then do coding while they're going. I'm also still somewhat worried that there's a bug for SlaterDetOpt with T-moves, which I could do with fixing first to allow DMC calculations to start (could run most systems without T-moves, but would still good to know what's going on in case something serious is wrong).

I might be misunderstanding the extent of the changes desired to combine SlaterDetOpt with DiracDeterminantBase, Ye would like the former to be a child of the latter, and we're discussing this.

eneuscamman:
Hi Nick,

I *think* the actual required changes are not so bad. We can discuss when
we're both back in Berkeley. I would agree though that if there is
unexpected behavior when using t-moves, this needs to be resolved first.

Best,
Eric

grahamlopez:
CI: test this please

grahamlopez:
Rhea is still in a weird state after maintenance. I'm looking into it

grahamlopez:
test this please

(retrying again after Rhea has righted itself)

grahamlopez:
test this please

(still working on Rhea)

grahamlopez:
test this please

grahamlopez:
test this please

grahamlopez:
CI won't be working until at least tomorrow morning

nsblunt:
Hi @markdewing, thanks for the fix for the OMP problem, sorry for not getting back earlier. I imagine that m_tf should be the same between threads, but can check that it behaves correctly when set up like this.

ye-luo:
@nsblunt m_tf holds data calculated for a snapshot of electronic configuration. It doesn't make sense to copy the pointer. In fact, I think LCOrbitalSetOptTrialFunc can be completely removed. Its only role is adding the derivatives with respect to parameters. There are only a few functions connected to TrialWavefunction and all the others are basically empty. I think all the data object of LCOrbitalSetOptTrialFunc can be moved inside LCOrbitalSetOpt and all the routines with real contents (resetParameters, checkInVariables, checkOutVariables, evaluateDerivatives...) can be merged into LCOrbitalSetOpt and SlaterDetOpt calls all these routines in corresponding routines. There is no need to add extra components directly into TrialWavefunction.

nsblunt:
@ye-luo this seems sensible and I'm happy to do this. Although I spoke to Eric previously and there was a reason for splitting it into a separate component of TrialWaveFunction, I believe to make the orbital rotation more general and applicable to other data types. For example, add_derivatives currently takes in data from the SlaterDetOpt class and creates m_pder_mat and m_hder_mat for use in evaluateDerivatives. If I understand correctly, this class should work with other wave function types (i.e., AGPs rather than determinants) if passed the appropriate data.

But I think it's a simple change to make and don't mind doing this. Eric (@eneuscamman), would you be happy with this?

eneuscamman:
I am OK with this.  If we need to generalize later e.g. for AGP then we can
do that then.

ye-luo:
Since AGP holds two instance of LCOrbitalSetOpt, I think AGP can access everything inside LCOrbitalSetOpt without directly touching TrialWaveFunction.
add_derivatives currently is only used by evaluateDerivatives of LCOrbitalSetOpt. If you need to access from outside (AGP class), you can make it a virtual function in the SPOSetBase.

nsblunt:
OK sure will do then.

grahamlopez:
CI: test this please

grahamlopez:
test this please

grahamlopez:
test this please

grahamlopez:
Thanks for your patience everybody. CI should be working now, so errors from here forward are actual build errors. Let me know if it looks like there is anything fishy going on with the testing.

test this please

ye-luo:
This is from the error log. The compiler used by the CI is not consistent with the module.
```
Currently Loaded Modulefiles:
  1) gcc/5.3.0             5) xalt/0.5.3            9) fftw/3.3.4
  2) openmpi/1.10.2        6) hsi/5.0.2.p1         10) szip/2.1
  3) PE-gnu/5.3.0-1.10.2   7) bbcp/15.02.03.00.1   11) hdf5/1.8.11
  4) lustredu/1.4          8) DefApps              12) git/2.9.3
-- The C compiler identification is GNU 4.4.7
-- The CXX compiler identification is GNU 4.4.7
```

grahamlopez:
test this please

nsblunt:
Hi all,

I'm looking for the bug with t-moves, and could do with some help.

I'm comparing optimizable determinants (SlaterDetOpt.cpp) with the standard determinants (DiracDeterminantBase.cpp). If I set orbital rotations to zero in the former then these should be equivalent. With t-moves there is a definite error in optimizable determinants. This seems to be coming from the function call to TrialWaveFunction::acceptTMove in DMCUpdatePbyPWithRejectionFast::advanceWalker. If I comment out this call then both sets of results are numerically identical, when starting from the same RNG seeds, but different when included.

acceptTMove calls three functions on the OrbitalBase objects: ratioGrad, acceptMove and updateBuffer. I believe that the call to updateBuffer is the cause of the problem: for SlaterDetOpt, updateBuffer *always* recomputes everything from scratch, even if the logical is set to false. If I understand correctly, this should make the function less efficient, but shouldn't lead to incorrect results.

However, if I recompute everything from scratch for the standard determinants (using the calls in DiracDeterminantBase.cpp) then the results are different to if I use the functions calls in acceptTMove, which suggests that computing everything from scratch is fundamentally the wrong thing to do in this situation. I don't see why this should be, unless some of the particle positions in P are wrong (i.e., we should only be using particle iat, not the others). Presumably the situation here is sensitive to the order that functions are called and data is updated.

What is acceptTMove actually supposed to do? How is it different to the standard acceptMove method of TrialWaveFunction? I see that is zeroes some data at the start, does this affect things?

Any help would be very appreciated.

Also, I noticed that the LogValue is added to the buffer before it's actually calculated - the convert function from logpsi to LogValue is only performed after LogValue is added to the buffer. Is this an error?

Best,
Nick

ye-luo:
Probably you are not following the develop branch. The T-move bug has been fixed #312 about two weeks ago. After the fix, acceptTMove is no more called and is to be removed.

nsblunt:
I'm working on this branch from this PR, which is based off commit 44e9aca from 2017-08-07, it's seem that that commit isn't present at that point. How would you like me to update this PR to include those newer commits? Should I merge the upstream/develop branch into this one, or rebase?

ye-luo:
I'm OK with merge.

nsblunt:
Thanks very much, T-moves seem to be working now based on a quick test, but I'll test more thoroughly.

prckent:
I have built this and run the new test successfully on oxygen with Intel2017. Thanks @nsblunt .

>Start 48: short-H4-orb-opt-16-1-totenergy
>48: Test command: /home/pk7/projects/qmc/git_QMCPACK_nsblunt/qmcpack/utils/check_scalars.py "--ns" "3" "--series" "14" "-p" "H4" "-e" "2" "--le" "-1.93045653 0.00837934"
>48: Test timeout computed to be: 3600
>48: Tests for series 14
>48:   Testing quantity: LocalEnergy
>48:     reference mean value     :  -1.93045653
>48:     reference error bar      :   0.00837934
>48:     computed  mean value     :  -1.93001757
>48:     computed  error bar      :   0.00844232
>48:     pass tolerance           :   0.02513802  (  3.00000000 sigma)
>48:     deviation from reference :   0.00043896  (  0.05238613 sigma)
>48:     error bar of deviation   :   0.01189479
>48:     significance probability :   0.04177911  (gaussian statistics)
>48:     status of this test      :   pass
>48:
>48: Test status: pass
> 2/12 Test #48: short-H4-orb-opt-16-1-totenergy ............   Passed    0.05 sec


Nick, can you confirm for our records where the reference test data "-1.93045653 0.00837934" came from? I am assuming a single run of this test. Do you have any sense of how reliable this will be? i.e. Is this "short" test long enough that we are getting valid statistics? 

I am OK to merge once @ye-luo  updates his "requested changes" status.

nsblunt:
Hi Paul,

The data is being extracted from the final file (H4.s014.scalar.dat) so I simply ran the calculation (with 16 MPI processes) and ran the check_scalars.py script on that file to extract benchmarks. Looking at the linear method iterations, they're mostly converged but not entirely. So if there are changes to the Markov chain then this might trigger an error.

In my previous work on FCIQMC we set up tests to ensure that benchmarks were met exactly (allowing some precision error of 10^{-10} or something), and I was following this approach here. If you want benchmarks to only be met within statistical error bars here then I should probably update this test with better convergence, and maybe bigger error bars.

I still need to make some of the changes requested by Ye. T-move errors are fixed for these optimizable determinants but not for my FDLR wave functions (not in this PR). I'm fixing that bug so I can start running big calculations, and then will make these changes. Apologies again for the delay.

Best,
Nick

prckent:
Nick - Since this is the development version, we can (and should) do incremental development, particularly where there is no risk to other code, such as here. I think FDLR can be fixed later, which means that this code can be merged now. OK?

nsblunt:
The FDLR code isn't on this branch, that bug is a separate issue. There were still two or three things that @ye-luo wanted changing with this PR before merging. I'm happy to merge it and make those changes as separate pull request soon, if you would prefer? I believe that this code is all stable and correct, just a few things need tidying.

prckent:
Will wait to hear from @ye-luo

ye-luo:
I still would like to have the cleaning up work done before merging. Considering the cleaning up work is not very complicated.
If someone would like to try the new feature, this branch on the fork can be accessed directly.
1) merge LCOrbitalSetOptTrialFunc into LCOrbitalSetOpt
2) move SlaterDetOpt to src/QMCWaveFunctions/Fermion and make it as a derived class of DiracDeterminantBase. Some cleaning up in the builders.
3) Need input example using spline orbitals.

grahamlopez:
All, I'm mucking about with Jenkins trying to address issues 323 and 340

grahamlopez:
CI: test this please

nsblunt:
Sorry for the very long delay. I've made SlaterDetOpt a child of DiracDeterminantBase, and put the up and down spin objects within the SlaterDet structure as requested.

This is giving different results compared to before. The reason is that, with SlaterDet objects, the correlatedSampling function in QMCCostFunctionOMP.cpp doesn't recompute anything if StoreDerivInfo is true. However, this isn't the case for the previous code, and for most OrbitalBase objects I've looked at. What is the correct behaviour here? Is it correct that nothing should be recomputed by the correlatedSampling function if StoreDerivInfo is true? This doesn't seem correct to me, as correlatedSampling is called after the wave function is updated in our adaptive linear method.

The reason for the above behaviour comes from the function:

TrialWaveFunction::evaluateDeltaLog(ParticleSet& P, PooledData<RealType>& buf)

The above functions calls:

logpsi += (*it)->evaluateLog(P, P.G, P.L,buf,false);

Because the bool provided here is false, the subsequent function call in SlaterDet.cpp uses evaluateLogFromDerivativeBuffer instead of evaluateLogForDerivativeBuffer. The former function doesn't recalculate anything, and therefore evaluateDeltaLog never recalculates anything. But for many other OrbitalBase objects (including the one used before this change), things are recomputed.

Let me know if the above isn't clear. I'm basically wondering what the correct behaviour is, so that I can decide is I should override SlaterDet's behaviour, and if anything is buggy.

ye-luo:
@nsblunt Thank for updating the code. I think we only need to focus on the StoreDerivInfo=false (input parameter usebuffer=no) code path. Buffer is only needed by the old quartic optimizer which will be deprecated in the near future. Just don't implement any routine requiring buffer used by CostFunction, the default dummy fallback can be enough to get the code built.

nsblunt:
I'm happy to do this, although doesn't usebuffer=no make the code run slower, since it then recomputes things instead of storing them?

Also the two tests I added have usebuffer=yes, should they be updated?

ye-luo:
The buffer saves computing when correlated sampling are applied on a few proposed parameter sets and NLPP is accounted in CostFunction. The OneShiftOnly only does it once and adaptive may do three correlated sampling (I don't know the detail very well). The overall time saving using a buffer is not significant but the memory requirement is significant for large systems. There is an assumption that everything in buffer doesn't change when WF parameter changes. This is mainly for static Slater determinants and Jastrow only optimization. Since you are updating determinants, buffer is no more relevant as you discovered.

prckent:
@ye-luo Then what happens when all parameters are optimized? i.e. determinants and a jastrow. This would not be unreasonable to try for a small system and has been done in the literature for molecules.

nsblunt:
I’ve checked that turning the buffer off does give consistent results after the latest commit. It could probably do with more careful checking (and a DMC test) before it’s ready for merging. There was also still a point about spline-based orbitals to be looked at.

So, should we not be using the buffer when optimizing determinants? What about multi-determinant expansions? As you suggest, the adaptive linear method runs correlated sampling with three different sets of parameters. It’s important that they are resampled, which it seems isn’t happening if the buffer is used, except for specific wave functions.

tests/system/CMakeLists.txt
```diff
@@ -141,6 +141,23 @@ IF(BUILD_LMYENGINE_INTERFACE)
...
+# spo set for both spin-up and spin-down determinants) and then doing a DMC calculation
+#
+  LIST(APPEND H4_ORB_OPT_DMC_SCALARS "totenergy" "-2.191461 0.0048") # total energy
+
+IF(BUILD_LMYENGINE_INTERFACE)
+  QMC_RUN_AND_CHECK(short-H4-orb-opt-dmc
+                    "${CMAKE_SOURCE_DIR}/tests/molecules/H4_orb_opt_dmc"
+                    H4
+                    h4_orb_opt.xml
+                    16 1
```
prckent: Better to have a test with 16 threads so that this test can run without MPI. i.e. "1 16" here, not "16 1"

nsblunt:
The optimizable determinants still don't seem to work with more than 1 thread. @markdewing previously mentioned that the problem was the m_tf member not being copied, and that it went away when copied. This m_tf member has now been removed entirely, so I don't know what the current problem is. I'll take a look. If it's obvious to anyone then I'd appreciate suggestions, I'm unfamiliar with the OMP side of the code.

prckent:
OK. I suggest to *add* the "1 16" test in addition to the "16 1" version. If the MPI one is totally reliable and the threaded one is not, it will be clear where to focus. We are also now adding tests for features that are intended to work shortly in the future, so this is a good candidate.

nsblunt:
I think the requested changes are done now. The only remaining point was about einspline orbitals. I've tried using these and it crashes. I have no experience with solid state calculations so am afraid i don't know what the desired behavior is or what's going wrong.

I have asked someone who knows this code to take a look, but they can't for a few days. Can this be merged now, and einspline orbitals fixed in a separate PR? Or could we remove the optimizable einspline code and add it back in this later PR (it's only a few lines)?

eneuscamman:
As Luning is fully focused on NiO at the moment and thus unavailable to
assist in the short term, I think it would be prudent to leave support for
solid state orbital optimization for a future pull request.  This way we
can tie up Nick's code now and solidify the important new features it
brings to the chemistry side.

prckent:
@eneuscamman Agree. 
@ye-luo Lets fix and merge. Let me know when you are satisfied.

prckent:
@ye-luo Please check you are OK with my conflict fix

ye-luo:
This is a good strategy.
We should leave the solid part aside until Luning has time for creating tests and improving the code.
I will look through the code again this week.

jtkrogel:
@eneuscamman Once this PR is in, will we have what we need to begin forming the Nexus interface discussed earlier?

eneuscamman:
Yes I think the tests that Nick has included should provide what is needed
for the Nexus setup.  Nick, let us know if you agree.

nsblunt:
Yes, for the optimizable determinants we could do the Nexus step. We still have the finite-diiference linear response (FDLR) code to merge, which is a separate PR. I hope this will be relatively quick to merge, it's quite self contained.

nsblunt:
@ye-luo Have you had chance to look through the code yet? Is it okay to merge?

ye-luo:
I went through the first edit and have fixed the mixed precision on my local repo. I need a bit more reading and thinking for a certain part. That is why I have not pushed yet. Expect to get it done before thanksgiving. I may still need to bother you for some changes if I think it is very necessary.

eneuscamman:
We have other merges waiting for this one, such as Nick's primary FDLR
code, so waiting until Thanksgiving will slow down other commits
significantly...

ye-luo:
This is an iterative process in the following weeks but fully settled before Thanksgiving. If everything goes quickly, the timeline will be accelerated. For the moment, any one who wants to try the code should use the branch instead of waiting for the merge. 

The code is still invasive to the old classes. Since the code is not written in QMCPACK types and standards, I would like keep it as separate as possible. If it finally becomes a core functionality, the current code must be rewritten (not now).

FDLR is in a similar situation. IF FDLR is just used for a few cases, it is not worth maintaining it in the main code just like we are just deprecating rarely used features.

Of the other hand, if a contribution is already very aligned with QMCPACK standard, we will probably accept the feature even if it is not used very often.

In general, we don't have time to rewrite everything and much of the responsibility is with the contributors. We need to start the training by having a dialog about expectation, arrange for whatever updates are really needed made now, and have later updates improve the situation.

ye-luo:
This is also a good chance for use to formalize requirements and standardize workflow and codes. It will be a bit pain at the moment but long term beneficial. The contribution from Berkeley is very important to us and we would like to ensure its hight quality.

jtkrogel:
If it doesn't break core functionality I don't see why this PR can't be merged as is with tidying up to follow (certainly best done once an integration test is in place anyway).

prckent:
@ye-luo My suggestions would be to queue up issues for requested improvements, just as I have in #456 following #451 . Then we can get these PR merged.

If your requests are surfaced as issues we can all see what is requested and work out who would be best to implement them, assuming we agree that they are really needed. For anything stylistic I suggest we expand the developer part of the manual so that we can refer to it in future.

eneuscamman:
I agree with Paul.  This will prevent us from getting into a situation
where the longer we wait the more the development branch will change and
the more will be required of Nick or someone else to get the merge done.
I'd rather have a merged code that works but with tracked issues than an
unmerged code.

ye-luo:
@jtkrogel There is also entanglement with other undergoing development. Minimal interference is preferred.

I will disentangle further the part must be fixed before merge and leave the left with issues as @prckent suggested.

ye-luo:
Is the DMC failure expected? I got them with real+full precision build.
```
      Start 120: short-H2-orb-opt-16-1
 1/22 Test #120: short-H2-orb-opt-16-1 ..............................   Passed   27.97 sec
      Start 121: short-H4-orb-opt-16-1
 2/22 Test #121: short-H4-orb-opt-16-1 ..............................   Passed   44.77 sec
      Start 122: short-H4-orb-opt-dmc-16-1
 3/22 Test #122: short-H4-orb-opt-dmc-16-1 ..........................***Failed    0.33 sec
      Start 123: short-H4-orb-opt-dmc-1-16
 4/22 Test #123: short-H4-orb-opt-dmc-1-16 ..........................***Failed    0.18 sec
```

eneuscamman:
Not sure. Nick?

nsblunt:
@ye-luo No that's not expected. I did test them and they all passed for me, but I was using the default build. It looks like they fail instantly, so presumably they're dying in initialization. Is there any error message or hint of what's happening?

ye-luo:
@nsblunt debugger shows that it is in the two body Jastrow. I opened your input and noticed that there is only ud J2. A few months ago this is allowed and uu and dd is populated as ud. Now the convention has been changed to align AoS SoA and GPU, if you want to populated all the pairs, uu is the entry instead of ud.

Some action items,
1, Please update the pair id and check if all the old reference results are still good.
2, In qmcpack, we usually have optimize="yes" instead of optimizable="yes". Unless you really want to distinguish using the optimizable class and controlling on and off the optimization, we can just use optimize as input name. Please update the code and test cases.
3, The matrices m_B and m_init_B need to be moved from the base class to the LCOrbitalSetOpt.h. The matrix multiplications in SlaterDetOpt.cpp need to be moved to LCOrbitalSetOpt as a member function and the rotation matrix is passed in as an argument. Before calling that member function, dynamically cast the pointer to LCOrbitalSetOpt pointer. I can do this quickly if you prefer, just let me know a function name you like.

nsblunt:
I can deal with points 1, 2 and 4 now. If you can do 3 quickly then that would be a big help. I'm not sure exactly which matrix multiplications you mean, there are several in that file. But I'm happy for you to make the change and name it as you like.

nsblunt:
OK those points are dealt with, please check they're okay. You might want to check that everyone required has been added to the license headers.

@eneuscamman are you happy with how you have been listed in the licence headers? Are there others who need adding as contributors?

eneuscamman:
I'm happy with them, and no I don't think we need any other names.

prckent:
Looks like we are OK to merge after the CI on the updated branch has passed :-)

nsblunt:
That's great, good news!

prckent:
Merging.

Big thanks to @nsblunt the updates and endurance and @ye-luo for the improvements.

Issue #162, opened 2017-03-28
Core dump at checkpoint


anbenali:
I will not have time to investigate this but I have all the files and core dump. It goes as follow:

- Run VMC correctly.
- Run DMC with a checkpoint every 20 blocks. Runs nicely the first 20 blocks and dies with the following error message:
2017-03-28 01:30:14.206 (INFO ) [0x400016334e0] 1395:tatu.runjob.monitor: tracklib completed
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc

The stack points at:

000000000108a3c0
00000012.long_branch_r2off._ZNSt8ios_base4InitD1Ev+0
/soft/applications/qmcpack/github/src/QMCDrivers/QMCDriver.cpp:446

Which corresponds to :
RandomNumberControl::write(RootName,myComm);

ye-luo:
std::bad_alloc means not enough memory.

prckent:
When reporting crashes, please give details of the run. e.g. electron count, size of spline buffers, machine etc. Then we will be able to tell if this is a small reproducible run or will need to be cut down conveniently debug.

prckent:
The checkpointing implementation should consider memory usage - it is better to be slow and reliable than run out of memory and crash.

ye-luo:
This issue is confirmed. Impact both read and write.
Random number IO is handled on MPI rank 0. mt19937 save 624 4-byte integer per thread.
Assuming using 16384 node with 32 threads each, need 1.2 GB memory.
It is fixable and has nothing to do with hdf5.
Currently the RN are store in *.random.xml. I don't like xml for such large files.
Since there is an h5 implementation, I only need to parallelize it.

The electron configuration is already handled in parallel. No issue in principle.

prckent:
Fixed by #320 ?

ye-luo:
I would like to close these issues after the nightly tests tonight.

ye-luo:
PHDF5 is working well in the nightly tests.

PR #320, opened 2017-08-16
Fix restart IO

ye-luo:
The memory and I/O issues of restart have been fixed via parallel collective I/O hdf5
by Bradley Protano @bprotano from Northern Illinois University.

Parallel hdf5 is turned on by default whenever the library supports it.
On/Off by -D ENABLE_PHDF5=1/0.
Performance is much better with hdf5 >=1.10.
Should fix issues #2 and #162.

Details:
1) RNG seeds and Walkers I/O are rerouted.
2) Seeds are no more stored in random.xml but in random.h5.
3) When the number of nodes in the restart run matches the previous run. Seeds and Walkers are reloaded as they were.
4) When the number of nodes doesn't match, seeds are ignored but walkers are redistributed among the nodes near equally.

screen shot 2017-08-11 at 11 00 43 am
screen shot 2017-08-11 at 10 57 21 am
screen shot 2017-08-11 at 10 57 45 am


prckent:
OK to test

prckent:
Thanks @Bprotano ! Looks to be a big improvement.

prckent:
How many walkers were there and what was the electron count/system size in the above graphs?

ye-luo:
The system is diamond 256 electrons.  There are 32 walkers on average per node. My largest run has totally 524286 walkers on 16384 nodes. The file is about 1.6G in single precision.

src/OhmmsApp/RandomNumberControl.cpp
```diff
@@ -375,4 +373,238 @@ void RandomNumberControl::write(const std::string& fname, Communicate* comm)
...
+  for(int i=0; i<nthreads; i++, vt_it += shape[1])
+  {
+    std::vector<uint_type> c(vt_it,vt_it+shape[1]);
+    Children[i]->load(c); //read seeds for each thread from buffer back into object
+  }
+  Random.load(mt); //read seeds back into object
+}
+
+//scatter write
+void RandomNumberControl::write_scatter(hdf_archive& hout, Communicate* comm)
```
markdewing: To match the underlying operation, this should be write_gather.
But I wonder if it would be better to use something different than 'scatter/gather' for this name - something that could be the same suffix on both read and write.
The operation is to read or write on MPI rank 0, so maybe 'write_rank_0 / read_rank_0' or 'write_head_node / read_head_node' or something else?

src/io/hdf_archive.h
```diff
@@ -61,17 +63,17 @@ struct hdf_archive
    * @param c communicator
    * @param use_collective turn on/off collective
```
markdewing: Update the parameter name in the documentation description.  Something like "request_pio  Request parallel I/O" (so that the short 'pio' gets defined somewhere).
Also it might be good to describe what happens if parallel I/O is not available (and under what conditions parallel I/O is or is not available).

jtkrogel:
@ye-luo This looks really great.  I've had checkpointing bring down a run before (but hopefully no more!).

Quick question: does this work for the gpu code as well?

prckent:
@ye-luo Does the current restart test exercise this adequately? We can+should expand the test if needed.

We have restart 1-16 and 8-2 running even with the CUDA build. Probably we need to enable parallel HDF5 on oxygen though.

src/OhmmsApp/RandomNumberControl.cpp
prckent: something like head_node is more descriptive

ye-luo:
@jtkrogel The CPU/GPU builds use the same lines of codes to checkpoint and restart. Our restart tests are running well on both platforms.
@prckent Serial or Parallel is chosen at build time. There is no need to add more tests. Do you have parallel enabled HDF5 on oxygen? If you have PHDF5, it is used by default automatically. If not, we can have the tests running for one night and then you switch to PHDF5.

I manually checked a few cases, including MP and also ensemble run. The PHDF5 works correctly in all cases. The serial walker reading ran without stop but was not reporting correct number of loaded walkers. Since It is less important and I have no time to investigate, I decide to leave it.

prckent:
@ye-luo OK.

Will wait for solutions to Mark's requests before merging.

ye-luo:
Yes. Please wait me to address the requested changes.

src/OhmmsApp/RandomNumberControl.cpp
ye-luo: head_node is not very precise when there are several MPI ranks on a node. I will do _rank_0.

prckent:
@markdewing Are you happy with these updates?

markdewing:
Yes, the changes look good.

Issue #297, opened 2017-07-24
Improve GNU compiler configurations and Cray handling.


ye-luo:
GNU compiler configuration now checks the CPU flags (SSE,AVX...) on the host building the code.
This is not friendly to cross-compiling and usually not accurate enough.
Proposed changes:
Align with Intel compiler setup.
1) -march=native as -xhost by default.
2) when -march is specified by users, respect it for cross-compiling.
3) In cray environment, do nothing as CC/cc already have the tuning flag beneath. (both Intel and GNU)

ye-luo:
@prckent The nightly tests build well. Do we need extra manual checks? Nersc(see my PR note)?
This issue can be closed if no issue is spotted.

prckent:
On Cori/Haswell with Intel compilers this yields
>-std=c++11  -qopenmp -Wno-deprecated -qopt-prefetch -ftz -mkl -O3 -DNDEBUG -restrict -unroll -ip
and most importantly the code still builds.

On Cori/KNL with Intel compilers and *following the instructions in the manual* we get
>-std=c++11 -xMIC-AVX512 -qopenmp -Wno-deprecated -qopt-prefetch -ftz -mkl -O3 -DNDEBUG -restrict -unroll -ip
whereas if we ignore the manual and don't put
>-DCMAKE_CXX_FLAGS="-xMIC-AVX512" -DCMAKE_C_FLAGS="-xMIC-AVX512"
then we don't get the AVX512 flags as expected.

I think that this is correct. Close this issue?

ye-luo:
I would recommend not to put ``-DCMAKE_CXX_FLAGS="-xMIC-AVX512" -DCMAKE_C_FLAGS="-xMIC-AVX512"`` for KNL. CC wrapper is doing everything for you. Could you type make in this case? if you don't seen tons of -xHost warning, we can close this ticket.

prckent:
1. There are no explicit -xHost on Cori shown by make VERBOSE=1 in any configuration.
2. Indeed, without the CXX_FLAGS override, the code does build. Hopefully the Cray environment is working.
3. The manual entry for Cori needs to be updated.
4. Theta is missing from the manual.

PR #325, opened 2017-08-17
Deprecate SSE and improve cross-compiling

ye-luo:
1) Historically we have two sets of einspline implementation, intrinsic version (up to SSE4.1) and standard version. The selection is controlled via flags like 
HAVE_SSE HAVE_SSE2 HAVE_SSE3 HAVE_SSSE3 HAVE_SSE41 HAVE_FMA4
Which are set by either detecting the flags on the compiling host or user input.
They are very unfriendly to cross-compiling and often not set accurate enough.
My recent observation shows on both AMD(interlago) and Intel(haswell) CPUs the std3 version outperform the legacy SSE versions. There is no reason to keep the intrinsic versions. In addition, they are no more used in any critical part after the SoA change and they are needed to build the GPU code only. For this reason only std3 versions are kept and sse versions are moved to legacy.

2) More friendly cross-compiling #297 
All the detecting of instruction set on the compiling host is removed.
a) If building on a Cray, do nothing and respect the compiler wrapper.
b) if not on a Cray and user specified -march (GNU and Clang) or -x/ax (Intel) in CMake compiler flags, respect user choice
c) if not on a Cray and no user input, use -march=native (GNU and Clang) and -xhost (Intel)

Note: building recipes need to be updated. For example, NERSC cori phase 2 no more needs to specify -xMIC-AVX512 (no harm if you keep it).

3) removed a few cuda code files in einspline folder, they are never used.


markdewing:
okay to test

prckent:
Confirmation needed from people with cross-compile machines, e.g. @lshulen that this is OK

prckent:
OK with deprecation, more than timely

ye-luo:
I manually build the gpu code on rhea and it was successful.
GPU build needs twice make as the CI does.
In the first time, make will stop very quickly and issue some error.
In the second time, make will trigger cmake again and then build binaries only with warnings.
The CI fails seem like the first make didn't stop properly but the second started and then everything messed up.

@grahamlopez I'm wondering whether changing the first "make -j 8" to "make -j 1" helps.

lshulen:
In general I think the cross-compilation part is reasonable.  I will try to get on one of our older crays and test this to see if it is working properly.  A question though.  Why should I not be able to override the compiler wrapper's choice of architecture flags on a cray?

ye-luo:
@lshulen You can still override like -DCMAKE_CXX_FLAGS="-xMIC-AVX512" but no more necessary.
In the past, we have to override because our CMake setting puts -xhost but CC wrapper has -xMIC-AVX512 and you get tons of warnings. When our CMake setting detects -x from CMake command line, it drops -xhost and no more warnings.
Now, you fully trust CC wrapper to put the flag for you.

lshulen:
That sounds reasonable.  I wonder if someone with access to NERSC could double check that this would work properly on Cori?  In particular, by changing the craype module can you properly produce executables for both phase 1 and phase 2 of the machine?  (Trinity will have the same thing to worry about but I don't have access at the moment)

ye-luo:
Not much idea about NERSC. Based on the section for Cori in our manual, You need to switch modules for phase 1 and 2. I think the underneath flags of CC are changed. In this new setting, I think the manual can even been merged because the only difference is "module swap craype-haswell craype-mic-knl"

To build a single binary for both KNL and BDW, I found some info on NERSC website
http://www.nersc.gov/users/computational-systems/cori/programming/compiling-codes-on-cori/
There is also a compiler flag "-ax MIC-AVX-512,AVX" for Intel compilers that will build a fat binary for both Haswell and KNL, and at run time will use the binary for the correct architecture.
You can still pass it through CMAKE_CXX_FLAGS.

ye-luo:
test this please

grahamlopez:
test this please

lshulen:
I can build it on my local cray machines.  It looks fine to me.

ye-luo:
The CI Rhea GPU build is resolved. The GPU node has HSW but the software is way behind.
Workaround by disabling AVX2 and BMI2.

PR #333, opened 2017-08-25
qmcpack manual entry for qmca

jtkrogel:
Provide detailed instructions regarding the use of qmca.  The added section should be sufficient for new practitioners (pedantic).  The aim of the section is to outline what is needed from a data analysis/statistical perspective to obtain production quality results while illustrating the various features of the qmca tool. 

Covered topics include: trace plotting, obtaining the mean and error bar of the local energy and other quantities, the exclusion of equilibration data, plotting data vs. series as is useful for optimization, processing multiple files, performing twist averaging, setting output units, and turning off autocorrelation estimation for faster trace plots.

The final two sections provide a listing of example qmca commands and a checklist of recommendations for obtaining production quality results.

A new appendix has been added to the manual that contains a derivation of the relative efficiency of twist averaging with uniform vs. irreducible weighted twists for any weight distribution.


grahamlopez:
Okay to test

grahamlopez:
test this please

PR #337, opened 2017-08-25
Add tolerance to einspline 1D unit tests.

markdewing:
Addresses failure of unit tests on some systems.

### DIFF
--- a/src/einspline/tests/test_one.cpp
+++ b/src/einspline/tests/test_one.cpp
@@ -44,13 +44,13 @@ TEST_CASE("double_1d_natural", "[einspline]")
 
   double val;
   eval_UBspline_1d_d(s, 1.0, &val);
-  REQUIRE(val == 2.0);
+  REQUIRE(val == Approx(2.0));
 
   eval_UBspline_1d_d(s, 10.0, &val);
-  REQUIRE(val == 3.0);
+  REQUIRE(val == Approx(3.0));
 
   eval_UBspline_1d_d(s, 5.5, &val);
-  REQUIRE(val == 2.5);
+  REQUIRE(val == Approx(2.5));
 
   destroy_Bspline(s);
 }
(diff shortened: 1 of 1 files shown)


ye-luo:
Which systems?

markdewing:
Kayahan (from ORNL) emailed me about these unit test failures on the OIC cluster, after compiling with the OIC5.cmake toolchain file.

ye-luo:
@markdewing I see.
@jtkrogel Are OIC5.cmake and PsiIntelMPI.cmake still necessary?

prckent:
1. Odd that this has not turned up on any other system. Perhaps there are some interesting optimization settings?

2. I would rather these be replaced by a build script, as we have for OLCF machines. I think the only odd thing about the OIC and PSI machines is the library locations. If a toolchain is still needed it is a sign that we need to work on CMakeLists.txt

ye-luo:
@markdewing Did you checked this on OIC5 or someone at ORNL confirmed this patch solving the problem?

kayahans:
Yes, after the fix that Mark suggested, all the tests ran fine. I had only these modules loaded during compilation: 
composerxe/2013.5.192    mpi/openmpi-1.6.4-gcc4

ye-luo:
@kayahans Everything seems old. I don't think we should compromise the rigorous testing for legacy.
The toolchain indicates this is a machine with AMD processor but you are using intel compiler.
Do you have any newer compilers on this machine? gcc 4.8 or newer?

prckent:
OIC5 has gcc4.9.3, openmpi 1.6.5, cmake 3.2.2. The issue is that no one has touched the toolchain in years. Less critically, it should also use a recent HDF5 and libxml2.

markdewing:
I'm pretty sure all the floating point comparisons in the unit tests should
use Approx.  I don't think there's any place where we need to demand
bitwise reproducibility for floating point values.

Some checks do not have Approx because they were overlooked and haven't
caused problems (until now).

ye-luo:
@markdewing True in general but seems not true in this case. 1.0+1.0 should be 2.0 instead of approx(2.0) even if they are floats.

ye-luo:
@kayahans Could you check if a build with the most recent compiler on OIC5 without using toolchain still require this patch? You can find the way to get AMD libm passed to cmake in build_olcf_titan.sh.
If it is still needed, I will approve this PR.

markdewing:
I see this failure on a regular workstation (x86_64) with clang 5.0 in release mode.

Issue #299, opened 2017-07-24
Add the description of SoA in manual.


ye-luo:
Need to describe how to access SoA in the user manual.

ye-luo:
Also add a paragraph about how to cross compile.

ye-luo:
Completed.

PR #338, opened 2017-08-25
Update documentation.

ye-luo:
1) SOA switch
2) cross compiling
3) parallel HDF5
4) update ALCF Mira Theta build recipe
5) restart miniapp
6) Remove "Compiling the GPU code" in the running chapter. All the info are the same as the installation chapter.

address #299

I need a bit help
1) I can't make the \cite working installation.tex
2) I would like to add a list of available features in SoA but I didn't find a good entry.

### DIFF
--- a/manual/installation.tex
+++ b/manual/installation.tex
@@ -260,7 +260,7 @@ MPIEXEC_NUMPROC_FLAG   Specify the number of mpi processes flag, e.g. "-n", "-np
 \end{itemize}
 
 \subsection{Configure and build using cmake and make}
-To configure and build QMPACK, move to build directory, run cmake and make
+To configure and build QMCPACK, move to build directory, run cmake and make
 \begin{verbatim}
 cd build
 cmake ..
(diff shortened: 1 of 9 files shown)


prckent:
OK to test

manual/installation.tex
```diff
@@ -87,16 +87,16 @@ \section{Prerequisites}
 Section \ref{sec:buildperformance} for performance suggestions.
 
 \begin{itemize}
-\item C/C++ compilers such as GCC, Intel, IBM XLC. LLVM-based compilers
-  are supported only on a preliminary basis.
+\item C/C++ compilers such as GNU, Intel, IBM XL. LLVM-based compilers
```
markdewing: Is the statement about LLVM-based compilers still true?  Or do we need to have clang in the regular testing to say it's supported?

ye-luo: It is probably no more true. I'm using Clang 4.0 on BGQ and my laptop without any problem.

prckent:
For the SoA features I suggest we actually add a new chapter "Beta test features" with the introduction. "This chapter describes developmental features in QMCPACK that might be ready for production, but require additional testing, features or documentation to be ready for general use. We describe them here because they offer significant benefits and are well tested in specific cases."

For SoA I would add the comment that users should check what we specifically test on cdash/ctest and to make specific requests to expand the functionality that they need.

ye-luo:
add a chapter "Features of QMCPACK"
section "features in production", basically copy paste the qmcpack.org and move GPU features from running chapter here.
section "features in beta test"
add SoA

ye-luo:
test this please

grahamlopez:
test this please

ye-luo:
@prckent let me know if there are any other changes needed.

prckent:
@ye-luo Yes. I will edit when I have time.

PR #339, opened 2017-08-28
Minor changes related to JeeI.

ye-luo:
1) add a unit test for OhmmsMatrix row/col swap.
2) allow eeI Jastrow defined only on partial atoms.

### DIFF
--- a/src/OhmmsPETE/tests/test_matrix.cpp
+++ b/src/OhmmsPETE/tests/test_matrix.cpp
@@ -13,6 +13,7 @@
 #include "catch.hpp"
 
 #include "OhmmsPETE/OhmmsMatrix.h"
+#include "config.h"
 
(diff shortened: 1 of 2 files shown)


markdewing:
okay to test

prckent:
@ye-luo Please state specifically what you mean by partial atoms and what is enabled by the JeeI patch, e.g. by giving an example here of an input configuration that would now work (for records purposes).

ye-luo:
For example, you in a H2O molecule. You can only specify O-u-u and O-u-d correlation but don't add H-u-u and H-u-d correlation.
The AoS supports this but SoA was not because I was not protecting the nullptr case.

prckent:
Matching capability of SoA code with AoS code. 
+1

PR #345, opened 2017-08-29
QMCPACK manual entry for timestep fitting tool

jtkrogel:
Add instructions on performing statistical timestep extrapolation with the qfit command line tool.  

Since the tool relies on jack-knife resampling to obtain error bars of fitted parameters (including the zero timestep intercept) a subsection on general purpose jack-knife resampling is included.

Future developments will expand the range of curve fitting options, e.g. to EOS curves.


grahamlopez:
okay to test

prckent:
Please remove the new qmcpack_manual.pdf from this PR. This reduces growth of the git repo.

[ Soon the old PDF will be permanently removed from the git and autogenerated nightly at some *.qmcpack.org URL.  ]

jtkrogel:
I reconstructed the commit without the pdf.  It should be fine now.

PR #347, opened 2017-08-29
Route HDF5_ROOT in ctest script

prckent:
Needed for testing


CMake/ctest_script.cmake
```diff
@@ -249,6 +249,10 @@ IF ( QMC_OPTIONS )
     SET( CTEST_OPTIONS "${CTEST_OPTIONS};${QMC_OPTIONS}" )
 ENDIF()
 
+IF ( HDF5_ROOT )
+    SET( CTEST_OPTIONS "${HDF5_ROOT};${QMC_OPTIONS}" )
+ENDIF()
+
 MESSAGE("Configure options:")
```
ye-luo: Based on CMake module description, HDF5_ROOT is not a CMake variable but an environmental variable.
Is it possbile to just add
export HDF5_ROOT=...
in nightly_ornl_oxygen.sh?
or it didn't work?

ye-luo: If you still need to pass something in, QMC_OPTIONS can be a choice. I don't feel a need to add this ambiguous HDF5_ROOT which seems not consistent with the usual use described in README.

prckent: Just being consistent with current usage.

The nightly script on oxygen had the necessary flags.

ye-luo: Could you explain more how you use HDF5_ROOT in ctest? I think it is not effective at all and no need to add it here.
```
cmake -DHDF5_ROOT=/opt/fake ..
...
CMake Warning:
  Manually-specified variables were not used by the project:

    HDF5_ROOT
```

prckent: HDF5_ROOT is used by the findhdf5 https://cmake.org/cmake/help/latest/module/FindHDF5.html

ye-luo: Yes. It is via environmental variable. But you pass it through cmake command line. That is why I think this line is not effective. You need to add an export in nightly_ornl_oxygen.sh. If you already have it, then it should work out of box without this patch.

ye-luo: The syntax is also very strange to me. A usually reroute looks like
```
SET( CTEST_OPTIONS "${CTEST_OPTIONS}; -DHDF5_ROOT=${HDF5_ROOT}" )
```
If you do `cmake HDF5_ROOT=/opt/fake ..` without -D, I feel HDF5_ROOT is just ignored.
Could you check if the reroute is really necessary?

prckent: I have updated the patch previously to be consistent with the usual reroute. Cut and paste error on my part.
We need to use command line arguments for ctest. 

I have run this successfully via ctest, see https://cdash.qmcpack.org/CDash/buildSummary.php?buildid=10796
>-- Found HDF5: /home/pk7/apps/hdf5-1.10.1-intel-mpi/lib/libhdf5.so (found version "1.10.1") found components:  C 
>-- Using HDF5 parallel collective I/O
>-- Boost version: 1.53.0
>-- Setting Boost_INCLUDE_DIRS=/usr/include
>-- Disabling CUDA
>-- QMC_DATA is /data/pk7/QMC_DATA
>-- Ready to build qmcpack
>Git branch: ctest_hdf5_fix
>Git commit hash: 9d47fbf9d270f28ce6fde74e79ec9551b9a14a9b

ye-luo: I see would like to know the answer whether adding
export HDF5_ROOT=/home/pk7/apps/hdf5-1.10.1-intel-mpi
in the beginning of nightly_ornl_oxygen.sh works or not?
It is better not to get ctest command line longer and longer.

prckent: I just checked and happily I can report that export HDF5_ROOT=/home/pk7/apps/hdf5-1.10.1-intel-mpi works.
We use the lot of arguments route because in the past we have found that many variables are not reliably passed, at least on oxygen and with the software versions currently installed.

ye-luo:
I saw the error. It seems that ctest does not pass in environmental variables by default.

Issue #294, opened 2017-07-21
Reenable tests currently disabled in SoA implementation


prckent:
After the SoA version #289 has been determined to be stable, we should add e.g. the gaussian molecular tests back to track the missing functionality.

ye-luo:
gaussian tests are reactivated.

PR #351, opened 2017-08-29
Update oxygen test script for parallel HDF5 and additional SoA testing

prckent:
Add use of parallel HDF5 to many configurations
Add gcc_mkl_soa and gcc_cuda_soa configurations to SoA testing

Oxygen is essentially full at this point so to add more tests we will have to either remove unneeded configurations (Intel2015?), trim the tests, or buy/find/use extra hardware.


prckent:
As requested by @ye-luo , update also reactivates gaussian tests as per #294

PR #353, opened 2017-08-29
Fix OpenBC PP with SoA.

ye-luo:
The PP local channel computation was not correct with SoA. Reported by H. Shin.


prckent:
OK to test

prckent:
Why did not our tests catch this?

ye-luo:
We only have one OpenBC test in the QE test atomO. In that case, the local channel contribution is small. This was caught on a H PP with only a local channel.

prckent:
Suggestion: make some "edge case" tests, e.g. pseudopotential hydrogen atom, splines and gaussians. This will only have one occupied spin channel and might catch some edge cases that we don't normally test. It would also catch this case.

PR #354, opened 2017-08-30
Fix variance problem in CSVMC and add CMakeTests

rcclay:
Main branch seemed to have an unusually large variance when computing differences in correlated sampling.  This commit fixes this (changes how differences are accumulated), and adds some CMakeTests for the all-electron CSVMC drivers with and without drift.


prckent:
OK to test

prckent:
@eneuscamman Is this relevant to any of your work?

grahamlopez:
test this please

rcclay:
I guess pending no serious objections, it should be good to merge.

QMCPACK separately implements some correlated sampling routines in the optimization part of the code.  However, I think maintaining the CSVMC driver is important  because:
1.)  Sampling is done via Umbrella sampling, instead of via sampling |Psi_0|^2 and reweighting.  Also, it supports Umbrella sampling with more than 2 wavefunctions/hamiltonians/configurations.  
2.)  Generating the trajectory and computing observables is done on the fly, and not in post processing like in the optimization ddriver.  In this way, its more like how we currently do VMC, DMC, and RMC than the optimization driver.  
3.)  CSVMC should work with general sets of [ParticleSet,Hamiltonian,Wavefunction].  Recently, I've been looking at finite differences of ion positions, although in the past I have done things like computing differences between unit cells with different strains.

prckent:
Thanks Ray. A useful summary. We should have a note in the manual :-)

@eneuscamman Has been busy paper writing. I am hoping he can comment soon. Then we'll merge.

prckent:
No criticism after a week, merging.

PR #356, opened 2017-09-01
Minor correction in doc.

ye-luo:
Though both CPU SoA algorithms and GPU algorithms have a similar memory footprint, the algorithms are very different. In principle, GPU implementation can be improved to match the efficiency of SoA ones but not very critical at the moment.


grahamlopez:
okay to test

prckent:
The language was actually ambiguous. It was meant to refer to the memory usage, not the actual algorithm.

PR #357, opened 2017-09-01
Require C++11 and increase minimum compiler versions

### LINKED ISSUE #298: Enforce C++11 in CMake
After the SoA merge, C++11 is required. The CMake configuration needs to be updated.

prckent:
Closes #298 

Breaking change for users with outdated compilers


ye-luo:
@prckent Is it a good idea to raise the compiler minimum requirement?
GNU 4.8, Intel 15, Clang 3.3

prckent:
Where do we state the minimum versions?

PR #366, opened 2017-09-08
clean up nofk input block, add nexus support

jtkrogel:
Momentum distribution estimator xml contained a single xml sub-element (<kpoints/>) which contained only a single attribute (grid).  Remove the kpoints element and move the grid attribute to the <estimator/> element.

Add support to Nexus to generate the momentum distribution xml element in the input file.

### DIFF
--- a/nexus/library/qmcpack_input.py
+++ b/nexus/library/qmcpack_input.py
@@ -2205,6 +2205,13 @@ class flux(QIxml):
     attributes = ['type','name']
     identifier = 'name'
 #end class flux
+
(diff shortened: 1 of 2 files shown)


prckent:
OK to test

ye-luo:
Can we have an estimator test?

jtkrogel:
Ilkka is checking that the estimator gives the expected results vs historical HEG and Na metal studies.  We will post a test and add a manual section once we are sure everything is working.

Issue #365, opened 2017-09-07
Disallow environment variable in CMake


ye-luo:
I mean the QMCPACK specific ones not the libraries.

I saw the commend in CMakeLists.txt
\# todo: disallow environment
QMC_MPI
QMC_OMP
QMC_BITS
QMC_COMPLEX
QMC_SK_RECURSIVE
CXX
They are used with $ENV{}

I specified CMAKE_C_COMPILER as Clang but CMake thought it is XL because I have CXX specified in my .bashrc.

CXX is very convenient but can cause hidden problems.
They can be made either cached variables or just removed.

Any thoughts?

prckent:
@naromero77 What are the SPACK requirements for environment variable handling? Do they consider environment variables a good thing?

Our current trajectory has been to add configure options (cmake command line etc.) for everything. Our environment variable handling is certainly inconsistent and not comprehensive. I favor stripping the handling as per @ye-luo since the side effects can be confusing. This would also require updates to the manual and possibly some build scripts.

naromero77:
@prckent OK, I talked to Ye and I understand the question. In the context of the Spack, what we have in their now for passing the compiler to cmake is a bit of hack. The preference is to avoid environment variables and pass the options in through the CMake command line. 

I am fine with getting rid of those environment variables. Not even sure what some of those mean (e.g. QMC_BITS?). Also, it would be wonderful if the CMake build log contained clearly which M^N combinations of options were turned on in a localized place in plain English. A while ago, the only way to verify that certain options were selected where to dig around in a header file. Not very user friendly.

PR #367, opened 2017-09-10
Remove using environment variables in CMakeLists.

ye-luo:
Following #365 
QMCPACK specific ones are removed.
CC, CXX and library specific ones are managed by CMake.

CXX, C_FLAGS, CXX_FLAGS are no more used explicitly by our CMakeLists.

Since CC, CXX is still supported by CMake. There is no need to update the README and the manual.


ye-luo:
@prckent In the ctest_script.cmake, ENV{N_PROCS} is still kept.

ye-luo:
OK to test

CMakeLists.txt
```diff
@@ -361,13 +338,11 @@ ELSE(CMAKE_TOOLCHAIN_FILE)
   #------------------------------------
   # Identify the compiler
   #------------------------------------
-  IF( $ENV{CXX} MATCHES "icpc" )
-    SET( COMPILER Intel )
-  ELSEIF($ENV{CXX} MATCHES "xlC" OR CMAKE_C_COMPILER_ID MATCHES "XL")
-    SET( COMPILER IBM )
-  ELSEIF ( CMAKE_C_COMPILER_WORKS OR CMAKE_C_COMPILER_WORKS )
+  IF ( CMAKE_C_COMPILER_WORKS OR CMAKE_C_COMPILER_WORKS )
     IF( CMAKE_COMPILE_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX )
```
prckent: While we are updating: I suggest updating this section to 

1.  Use CMAKE_C_COMPILER_ID and CMAKE_CXX_COMPILER_ID consistently, as per the Intel check, for GNU and IBM. See https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html
2. Remove the microsoft check. We know of no one using this in years and it is a totally unchecked configure path. We have an "unknown compiler" route.

prckent:
OK to test

ye-luo:
Include a change for #368

PR #372, opened 2017-09-12
Include optimization tips in Backflow wavefunction section.

rcclay:
Incorporate optimization tips into Backflow section from document found on Assembla.


prckent:
OK to test

PR #373, opened 2017-09-13
Resurrect Backflow Optimization

Paul-St-Young:
Enabled and tested backflow wavefunction optimization. 

I had to pull 2 legacy spline implementations back in and use EinsplineSet.cpp instead of the SplineAdoptors when backflow optimization is needed. However, everything is coded up such that regular code behavior is unchanged. Namely, all ctests should still pass.

To optimize backflow, one must manually set <sposet_builder use_old_spine="yes"> or <determinantset use_old_spline="yes"> and use walker-by-walker move. @rcclay it may be worth checking if stress estimators work using use_old_spine="yes".

This is probably not the best way to resurrect backflow, but it is A way that works. I figured it would be better to start the conversation instead of trying to find the perfect way by myself.


ye-luo:
Why do you need einspline/**legacy**/multi_bspline_eval_std3_d_impl.h?
Is what you need missing in einspline/multi_bspline_eval_d.h and multi_bspline_eval_d_std3.cpp?

Paul-St-Young:
I get seg. faults if I try to call the new spline routines

ye-luo:
multi_bspline_eval_d_std3.cpp should be very close to legacy/multi_bspline_eval_std3_d_impl.h. If you run gdb, the fix can be quick.

ye-luo:
I think create_einspline_extended can be avoided. In the gpu code path, EinsplineSetExtended has already been created. You even enabled "#if 0" I put.

Paul-St-Young:
I did run gdb. Sometimes the first call to the new spline functions work, but later calls cannot access the spline coefs. The bug was random which made me worried that memory was being shuffled around inconsistently by EinsplineSet.cpp and multi_bspline_eval_d_std3.cpp. Thus I went back to the legacy implementations.

Paul-St-Young:
I wasn't sure how best to do this. That's why I started this PR. I will be happy with any implementation that can pass the two short tests I added: short-bccH_3x3x3*

ye-luo:
You add an h5 binary in the repo. How large is it?

Paul-St-Young:
440KB

prckent:
OK to test

prckent:
@Paul-St-Young Thanks for this update. I would  prefer to solve the bug and have no "legacy" code usage. Do you know what the problem is/was?

prckent:
We should be cautious about adding binaries since it permanently enlarges the size of the git repo. 440KB is not that large, but since backflow is an advanced feature, how about we generate the wavefunction.h5 using QE? We already have the recipe from the QE tests.

grahamlopez:
Here is the build.log for the mixed-precision CI.

grahamlopez:
I’ve attached the build.log for the mixed-precision CI. Sorry for the duplicate emails (needed reply-all).

Paul-St-Young:
@prckent I would love to have squashed the bug, but I have no idea what was causing segmentation faults when I tried to call the new spline routines as I discussed with @ye-luo.

In terms of the test binary, I would prefer to keep it as a file. I generated it using a custom version of QE which removes the cusps from the DFT orbitals. Given that we already have 60MB worth of pwscf.pwscf.h5 files in the test folder, I do not think another 440KB (0.73% increase) is so bad. In comparison, the wavefunction file for bccH_1x1x1 is 3.1 MB.

Paul-St-Young:
@prckent , to answer your first questions directly:
The problem was that backflow optimization needed a version of evaluate_notranspose that evaluates the gradients of the hessian matrix, namely a variant of evaluate_notranspose that fills grad_grad_grad_logdet in EinsplineAdoptor.h. 

A "real" fix should implement BsplineSet<SplineAdoptor>::evaluate_notranspose for grad_grad_grad_logdet.

However the new spline routines are quite different from the old ones in EinspleinSet.cpp, which makes it very difficult for me to figure out what exactly needs to be implemented.

Therefore the fastest way for me to fix backflow optimization was to switch code route back to EinsplineSet when a particular user flag is turned on.

prckent:
@Paul-St-Young Unless this is urgent, lets try to fix the new spline routines when people have bandwidth (there is a two week crunch for some of us).

Valid point on the file size. This is more an opportunity to set an example for future tests added by others.

Can you explain your "custom version of QE"?

Paul-St-Young:
OK. This can wait a bit. My custom version of QE is in the repository: https://github.com/Paul-St-Young/qmcpack-espresso-5.3.0

prckent:
General comment to all: The difficulty in figuring out what is needed here points to a need for a clearer structure, cleaner code, and better documentation in general.

rcclay:
@ye-luo I'm looking at fixing this within the current SplineAdoptor framework.  Do you have any objections if I allocate grad_hess_matrix data containers in in the SplineC2XAdoptor and SplineR2RAdoptors?

ye-luo:
The classes are at QMCWaveFunctions/BsplineFactory. They are SplineC2CAdoptor.h, SplineC2RAdoptor.h and SplineR2RAdoptor.h.

What is the dimension of these matrices? Where were they in the old implementation?All the SplineX2XAdoptor classes holds data only for one electron. Any object with a size equal and larger than Nelec^2 must be handled carefully (minimal use and allocated by the feature).

Long term, SplineX2XAdoptor should be able to handle this. For this PR, using EinsplineSet is OK but not the legacy einspline routines and builders.

rcclay:
Correct me if I'm wrong, but it looks like the role of StorageValueVector in the old code is now played by myV in the SplineX2XAdoptor, likewise for the gradients (StorageGradientVector -> myG), laplacians, and hessians.  If that's right, then backflow needs StorageGradHessVector, which is Norbs*OHHMS_DIM^3 in size.

rcclay:
Also, why are there SplineX2XAdoptor.h files in both src/QMCWavefunction and src/QMCWavefunction/BsplineFactory?

ye-luo:
@rcclay You are correct.
Norbs*OHHMS_DIM^3 is OK. The scaling is the same as myV/G/H
src/QMCWavefunction has the old ones.
src/QMCWavefunction/BsplineFactory has the new ones but only part of of the features are covered.

rcclay:
@ye-luo Sounds good.  I'll add a myGH to the adoptor classes, resize these at the same time as myV, myG, etc get resized.

prckent:
Once these updates are done, can we remove the legacy spline code src/einspline/legacy from the current git version? I believe the derivatives were the only concern.

ye-luo:
I think once the use of legacy routines in this PR is fixed. I will remove both the einspline/legacy and the old builder routines.

Paul-St-Young:
I wanted to clarify the problem in this PR in relation to that in #390.

The quasi-particle transforms handled by BackflowTransformation.h are correct in the current develop branch. Bugs reported in this PR are purely due to the spline orbitals.

Paul-St-Young:
I should also mention three addition problems I ran into: 
1. backflow is broken if `UseRealOrbitals` is set to true in `EinsplineSetBuilderCommon.cpp::AnalyzeTwists2()`.
2. backflow optimization is broken if simulation cell is not orthorhombic because `StorageGradHessVector` is not correctly transformed in `EinsplineSetExtended<StorageType>::evaluate_notranspose` for `grad_grad_grad_logdet`
3. backflow optimization is broken if target particles have a mass different from 1.0 because `BackflowTransformation::Ymat` is not allocated in a particle-resolved manner. Instead, there is an implicit sum over particles assuming mass=1.0 in `BackflowTransformation::evaluateWithDerivatives`

ye-luo:
@Paul-St-Young Please
1. Could you update the status after my changes?
2. Revert all changes in src/einspline/legacy
3. grad_lap test was checking Testing/Temporary/LastTest.log. This is not reliable. Is it possible to check wftest.000

Paul-St-Young:
@ye-luo Thanks a lot for linking up the new splines!

I will:
1. re-evaluate what is working and what is broken
2. revert all changes in src/einspline/legacy
3. rewrite check_grad_lap.py to read wftest.000 directly

ye-luo:
@Paul-St-Young My pleasure. The most recent splines are still not linked and @rcclay will fix it. Recover the EinsplineSet for the moment is still useful for checking new implementation.

Paul-St-Young:
@ye-luo I have 
1. modified python scripts to read wftest.000 only. 
2. added some documentation regarding backflow in the manual
3. confirmed that backflow optimization is still broken for the three cases I listed above

I have committed the failing examples in that last commit. This last commit contains binary files, which is not ideal. However, my plan is to watch the last commit fail on CI, commit my fixes, then watch the failing examples pass. Once all the tests pass, we can perhaps remove the commit containing the binary files before merging.

It is weird that that CI thinks the last commit is passing. How do I trigger the tests properly?

ye-luo:
@Paul-St-Young 
In one of your original post, `UseRealOrbitals` is not working even just in VMC. Now VMC is fine but optimization is not. Right?

CI only checks unit tests.

Please resolve the conflict. First merge develop to your branch and then fix, commit and push.

Paul-St-Young:
@ye-luo I do not remember if VMC worked or not, but optimization never worked when `UseRealOrbitals=true`

Paul-St-Young:
@ye-luo I merged the current develop branch, but my new commits are not showing up in this PR. Should I close this and submit another one?

ye-luo:
What is the name of your new commits? Did you push? I saw your merge commit and one 'fixed failing tests' commit.

Paul-St-Young:
Oh yeah, nvm, my browser did not update. My commits are here now.

Paul-St-Young:
I just added another example that breaks backflow optimization.

This one is more severe. Both overlap and hamiltonian derivatives are wrong. None of the previous fixes I committed can resolve this problem.

Any suggestion is greatly appreciated!

ye-luo:
@Paul-St-Young Your last 'fix' is not a fix but a workaround. Please revert it in this PR. You can keep in your own branch if you need it for your work. Tests are meant to alert problems.
I need a explanation this workaround. In which case the code is broken and what the workaround gets right.

The broken optimization test is useful. Does it take a long time to run on a workstation?

Paul-St-Young:
@ye-luo fairpoint. I have removed the workaround flag use_complex_orb="yes". The parameter derivative runs that use `UseRealOrbitals=true` are now broken again.

All the tests I added take on the order of 10s on 1 core because they run the wavefunction tester instead of an actualy VMC/DMC driver.

Paul-St-Young:
@ye-luo I was able to fix the breaking example I added. The problem is either due to the reading of wavefunction hdf5 file or due to the tiling keyword. I will investigate further.

ye-luo:
@Paul-St-Young How long do you need to create the fix for the test failurue? If it takes a bit time, we can leave it to the next PR. For the moment, I still need "revert all changes in src/einspline/legacy"

Paul-St-Young:
@ye-luo I am at a lost as to where the bug is. Let's leave it to the next PR. I will revert the legacy einsplein changes ASAP.

src/QMCWaveFunctions/EinsplineSetBuilderOld.cpp
```diff
@@ -275,7 +275,7 @@ EinsplineSetBuilder::MuffinTinPath(int ti, int bi, int tin)
   return groupPath.str();
 }
 
-#ifdef QMC_CUDA
```
prckent: What is the logic behind keeping this commented out code?

@ye-luo  Keep or delete?

ye-luo: A decision is needed to fully remove all the unused routines in the builder. The builder has the old I/O code for APW(MuffinTin). If we delete, better to purge it everywhere not just the builder in one short.

prckent: OK. Perhaps do this when the new APW work is published and productionized.

Issue #300, opened 2017-07-24
Addition of max_seconds to input in order to avoid corrupted output data.


kylanpaa:
Currently, if QMCPACK exceeds wall time limit, the hdf5 files with observables such as density will not be closed and they become unusable. If the wall time limit would be given in the input QMCPACK could estimate during the simulation when to end the run without loosing all the data, that is, whether to continue to next block or not. If it would not continue further then the data as well as walkers should be saved. Mainly a feature for more user friendly QMCPACK, and conveniently utilized with NEXUS also.

markdewing:
The DMC driver has the 'maxcpusecs' parameter.   At the end of each block, it checks to see if this time has been exceeded, and quits if so.   This isn't quite what is requested as it requires the user to know about how long each DMC block takes and set parameter to the wall time limit minus the time for one block (and minus start up time and other time occurring before the DMC driver)

It would be better for a time limit parameter work as described.

prckent:
Agree that we should predict reaching the time limit and shut down sooner and more safely. Are the HDF5 files (observables) closed correctly in the current maxcpusecs implementation?

markdewing:
The current maxcpusecs code exits the loop the same as reaching the target number of blocks.   So all the files should be closed correctly as long as the regular code implementation closes everything correctly.

prckent:
@markdewing Can you get to this? If it is "just" checking that the next block can be completed with some safety factor, this seems straightforward and worthwhile.

markdewing:
I will look into what this would take for implementation.

prckent:
(Mark has a prototype of this feature)

PR #377, opened 2017-09-14
Change behavior for the DMC CPU time limit.

markdewing:
The time limit is set with the MaxCPUSecs parameter.

Previously, the time limit was measured from the beginning of the
DMC portion of the run, and continued until the set time runs out.
To effectively set this parameter, the user would need to know how much
time is spent prior to the DMC section, and the duration of each block,
in order to leave enough time for the simulation to shut down cleanly.

The new behavior measures the time elapsed from the beginning of the
run (captured during static initialization of the RunTimeManager object).
The time per DMC block is tracked and used to estimate if there is sufficient
time to run another block.

The time reserved to finish the next block is computed from the existing
average with some margin (to handle variations in block to block time) and
a fixed amount (to cover the time to end the simulation cleanly).

This addresses issue #300 
Potential issues
- It silently changes the behavior of MaxCPUSecs.   I'm not sure if anyone depends on the current behavior?
- the margin time (10%) and fixed padding (10s) are arbitrarily chosen, and most likely incorrect.
- The decision to stop is made on every MPI node independently.  There is a chance that one node may decide to exit and other nodes continue, and this would cause problems (especially if the exit path need global reductions or parallel HDF IO).    However, the new code is no worse than the old code in this regard.

### DIFF
--- a/src/QMCDrivers/DMC/DMCOMP.cpp
+++ b/src/QMCDrivers/DMC/DMCOMP.cpp
@@ -27,6 +27,7 @@
 #include "Message/Communicate.h"
 #include "Message/OpenMP.h"
 #include "Utilities/Timer.h"
+#include "Utilities/RunTimeManager.h"
 #include "OhmmsApp/RandomNumberControl.h"
 #include "Utilities/ProgressReportEngine.h"
 #include <qmc_common.h>
@@ -234,6 +235,7 @@ bool DMCOMP::run()
 {
 
   Profile prof("DMC","run",QMC_DMC_0_EVENT);
+  LoopTimer dmc_loop;
 
   bool variablePop = (Reconfiguration == "no");
   resetUpdateEngines();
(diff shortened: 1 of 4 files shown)


ye-luo:
1) Some part can be moved into QMCDriver to be a feature also For VMC, OPT, RMC.
2) Define the stop control flag false by default and do MPI reduce at every block to stop all the MPI within this calculation(not all, in ensemble job), For each block, I/O is anyway synchronized.

src/QMCDrivers/DMC/DMCOMP.cpp
```diff
@@ -327,7 +332,21 @@ bool DMCOMP::run()
         *(RandomNumberControl::Children[ip])=*(Rng[ip]);
     }
     recordBlock(block);
-  } while(block<nBlocks && myclock.elapsed()<MaxCPUSecs);
```
ye-luo: Is myclock still used by others?

prckent:
I believe that there are no users of maxcpusecs currently.

I agree with @ye-luo that this would be good to generalize in future, with maxcpusecs being a property of the run and not a specific driver call.

src/QMCDrivers/DMC/DMCOMP.cpp
markdewing: myclock is not used elsewhere - it's local to this file.  I should remove it.

ye-luo:
@markdewing is the manual description consistent?

markdewing:
The manual description seems vague enough that it could describe both the previous and current behavior.  It might be better to defer updating the manual to when the feature is generalized to other QMC methods.

PR #378, opened 2017-09-14
Add N_PROCS_BUILD variable to ctest script

prckent:
Overrides N_PROCS if defined. Needed to avoid process limits and perform parallel testing. Updates to OLCF titan and eos scripts will follow. Should not affect any other systems.

### DIFF
--- a/CMake/ctest_script.cmake
+++ b/CMake/ctest_script.cmake
@@ -40,18 +40,20 @@ ELSEIF( ${HOSTNAME} MATCHES "cori" )
     SET( N_PROCS 32)
 ELSEIF( ${HOSTNAME} MATCHES "eos" )
 # Setup for eos.ccs.ornl.gov Cray XC30 Intel E5-2670 Aries interconnect.
+# N_NPROCS and N_PROCS_BUILD should be set to respect current OLCF limits
     SET( CC ${CMAKE_C_COMPILER} )
     SET( CXX ${CMAKE_CXX_COMPILER} )
     SET( CTEST_CMAKE_GENERATOR "Unix Makefiles")
     SET( CTEST_SITE "eos.ccs.ornl.gov" )
-    SET( N_PROCS 16)
+    SET( N_PROCS_BUILD 8 )
 ELSEIF( ${HOSTNAME} MATCHES "titan" )
 # Setup for titan.ccs.ornl.gov Cray XK7
+# N_NPROCS and N_PROCS_BUILD should be set to respect current OLCF limits
     SET( CC ${CMAKE_C_COMPILER} )
     SET( CXX ${CMAKE_CXX_COMPILER} )
     SET( CTEST_CMAKE_GENERATOR "Unix Makefiles")
     SET( CTEST_SITE "titan.ccs.ornl.gov" )
-    SET( N_PROCS 16)
+    SET( N_PROCS_BUILD 8 )
 ELSEIF( ${HOSTNAME} MATCHES "cetus" )
 # Setup for cetus.alcf.anl.gov BlueGene/Q
     SET( CTEST_CMAKE_GENERATOR "Unix Makefiles")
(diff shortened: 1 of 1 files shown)


markdewing:
I assume the updates to the titan/eos scripts will define N_PROCS?

prckent:
Yes.
>export N_PROCS=`expr $PBS_NUM_NODES \* 16`

CMake/ctest_script.cmake
```diff
@@ -157,6 +158,12 @@ IF( NOT DEFINED N_PROCS )
     ENDIF()
 ENDIF()
 
+# Set the number of processors
+IF( NOT DEFINED N_PROCS_BUILD )
+     IF ( DEFINED $ENV{N_PROCS_BUILD} )
+        SET( N_PROCS_BUILD $ENV{N_PROCS_BUILD} )
+     ENDIF()
```
ye-luo: else SET( N_PROCS_BUILD ${N_PROCS} )
Just make sure N_PROCS_BUILD is always defined.

prckent: We would have to set it for every machine. This is intentionally on the safe side.

CMake/ctest_script.cmake
```diff
@@ -173,10 +180,14 @@ SET( CTEST_NIGHTLY_START_TIME "22:00:00 EST" )
 SET( CTEST_COMMAND "\"${CTEST_EXECUTABLE_NAME}\" -D ${CTEST_DASHBOARD}" )
 IF ( BUILD_SERIAL )
     SET( CTEST_BUILD_COMMAND "${MAKE_CMD} -i" )
+ELSEIF ( DEFINED N_PROCS_BUILD )
+    SET( CTEST_BUILD_COMMAND "${MAKE_CMD} -i -j ${N_PROCS_BUILD}" )
```
ye-luo: No more need to test N_PROCS_BUILD. always use it.

prckent: N_PROCS_BUILD is a non-default override.

CMake/ctest_script.cmake
```diff
@@ -40,18 +40,20 @@ ELSEIF( ${HOSTNAME} MATCHES "cori" )
...
 ELSEIF( ${HOSTNAME} MATCHES "eos" )
 # Setup for eos.ccs.ornl.gov Cray XC30 Intel E5-2670 Aries interconnect.
+# N_NPROCS and N_PROCS_BUILD should be set to respect current OLCF limits
     SET( CC ${CMAKE_C_COMPILER} )
     SET( CXX ${CMAKE_CXX_COMPILER} )
     SET( CTEST_CMAKE_GENERATOR "Unix Makefiles")
     SET( CTEST_SITE "eos.ccs.ornl.gov" )
-    SET( N_PROCS 16)
+    SET( N_PROCS_BUILD 8 )
 ELSEIF( ${HOSTNAME} MATCHES "titan" )
```
ye-luo: In principle, this can be removed since you added the environment variable.

prckent: Yes it could, but we are not going to. OLCF has process limits that cause weird build errors if this is not set. If someone wants to experiment, the environment variable takes precedence.

CMake/ctest_script.cmake
ye-luo: Not fully understand your intention. I mean if both N_PROCS_BUILD and a corresponding environment variable is not defined, set N_PROCS_BUILD = N_PROCS

CMake/ctest_script.cmake
ye-luo: fair enough.

CMake/ctest_script.cmake
ye-luo: If we have always defined N_PROCS_BUILD properly. It is cleaner to have a single
``SET( CTEST_BUILD_COMMAND "${MAKE_CMD} -i -j ${N_PROCS_BUILD}" )``

CMake/ctest_script.cmake
prckent: Please merge - I need this merged before I can finish testing the job scripts.

ye-luo: Could you fix the previous two places. Add an else defining N_PROCS_BUILD and then always set make with N_PROCS_BUILD.

CMake/ctest_script.cmake
markdewing: I think Ye's point is that if his suggested 'else' branch is added here, then N_PROCS_BUILD is defined through all code paths, and and be used directly to simplify the next section that he commented on.

markdewing:
Merging.  I think the logic works well enough as is.

PR #388, opened 2017-09-25
Fix divided by zero issue in CUDA noMPI build.

ye-luo:
During the MPI and device matching step, allgather is called. In no MPI build, all gather was not implemented and caused a float point exception (divided by zero).


markdewing:
okay to test

PR #389, opened 2017-09-25
update build_olcf scripts

jtkrogel:
Add parallel hdf5, include double cmake call for titan gpu builds.


prckent:
OK to test

yingwaili:
Tested scripts on EOS and Titan.  It comes to my attention that two "make -j 32"s are needed to compile the GPU codes (both real and complex) successfully.  Otherwise ok.

prckent:
Thanks. What breaks on the first make?

PR #393, opened 2017-09-28
Add tests for Nexus examples.

markdewing:
Nexus scripts are run with 'generate_only' set to True.
This will at least test the Nexus API and catch breaking changes.

Currently the LiH test fails:
 QmcpackInput error:
    pseudo does not have the following attributes/elements:
        l-local

The 'generate_only' setting is changed in the script by setting a particular value in the globals prior to exec'ing the Nexus script.    It does require a change to the Nexus script being tested.
It would be nice to be able to check this inside of Nexus, rather than requiring each example script to check it.    The modification of the globals only affects the exec'ed file, and does not propagate to the files imported from that file.
Another option would be to set an environment variable, and check that from inside Nexus, but I'm not sure where to insert that check.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -861,6 +861,7 @@ SUBDIRS(src)
 IF(NOT(QMC_BUILD_LEVEL GREATER 4))
   SUBDIRS(tests)
   SUBDIRS(examples)
+  SUBDIRS(nexus/examples)
 ENDIF()
 
 #ADD_CUSTOM_TARGET(print_settings ALL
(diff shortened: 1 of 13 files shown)


jtkrogel:
The example fails because Nexus is now more fastidious about writing xml attributes.  For example to avoid propagating user typo errors to qmcpack,  the attributes now have to be known to Nexus.  When the example was written this check was not enforced and so any attribute that was assigned to the xml was simply written.

In terms of driving existing scripts without adding global checks, probably the right thing to do is to enable command line inputs into nexus to control/modify the options to settings.  

I will add the command line capability in a later PR along with a fix for the unrecognized xml attribute (l-local).

jtkrogel:
+1 for merge

PR #398, opened 2017-09-29
Add approximate testing for complex values.

markdewing:
Create a ComplexApprox class that acts similar to the Approx class in Catch.
It will compare the real and imaginary parts separately. It can also test just
the real part if compare_real_only is set.
The epsilon and scale values settings are not exposed, but it would be easy
to add them.
Also create a ValueApprox typedef for testing ValueType with the appropriate
real or complex setting.   It's not used in this PR, but it will be used in a future PR.


markdewing:
test this please

markdewing:
test this please

Issue #394, opened 2017-09-29
Tests of STO needed


jtkrogel:
Slater-type orbital functionality needs integration tests.  A good target is Li2 from Umrigar, Nightingale, & Runge (1993), which also includes a simple Pade Jastrow factor.

This system was used as a test in the early development of QMCPACK and likely was run throughout its informal testing period.  Given the amount of benchmark data available in UNR, integration of this test also provides a good opportunity to reverify the correctness and efficiency of the QMCPACK projector/Green's function.

ye-luo:
It is good to have it.
At the moment we only have STO in one of the example He.

prckent:
Agree good to have, particularly since we can compare with historical UNR data.

jtkrogel:
Brief update (full detail coming in a PR soon):  VMC and DMC energies with QMCPACK reproduce UNR results, at least at small enough timesteps for Li2 STO:
```
        VMC          DMC (tau=0.01)
6259  -14.9471(2)  -14.9894(2)
v300  -14.9473(2)  -14.9893(2)
v311  -14.9469(2)  -14.9894(2)
v320  -14.9470(2)  -14.9895(2)
UNR   -14.9472(2)  -14.9895(4) 
6259  -14.94721(5) -14.98944(6)  (very long run)
```

For timesteps larger than ~0.06/Ha, QMCPACK departs strongly from UNR, though still with a quadratic timestep dependence.  The largest timestep we can use for a 1 mHa accuracy is ~0.08/Ha, while for UNR it is ~0.15/Ha.  Total energy differences are likely still accurate for UNR up to 0.25/Ha (5 mHa deviation vs. 38 mHa deviation for QMCPACK).

I am now adding long/short vmc/dmc tests to ctest.

prckent:
We don't have the all electron fixes from UNR ( http://aip.scitation.org/doi/10.1063/1.465195 ) around Eqn 35+. Presumably this is the difference at large time step?

jtkrogel:
Possibly.  I'm not familiar enough with UNR or our partial implementation of it.  

From my brief reading of the paper I don't immediately see why these parts/formulae are not implementable in single particle moves, but if we don't have them then that is the most likely explanation.

I am also checking Ye's other branching schemes (one of which he terms UNR) alongside our original one.

prckent:
Probably sociology and an historic focus on PP calculations.

jtkrogel:
[...]

PR #403, opened 2017-10-03
Add Li2 STO tests

jtkrogel:
Addresses #394.  VMC and DMC tests are provided for the Li dimer system used in Umrigar, Nightingale, & Runge's 1993 JCP.  This is a Slater (STO) Jastrow (Pade) wavefunction.  

The VMC energy matches UNR to 0.2 mHa.  A timestep study was performed to compare the Green's function in QMCPACK with UNR.  QMCPACK and UNR agree for moderately small timesteps, with a small difference (~1 mHa) apparently present in the t=0 limit.  Please read the README file included in this PR as part of the review.

The QMC_RUN_AND_CHECK function was modified to handle an arbitrary number of series checks (using the catch-all ARGN input list), so that e.g. VMC and DMC results generated from a single run can be checked without redoing the run.  This might be also be of use for future tests as a function of timestep.  Series checks added beyond the first result in a series number being added to the full test name; the first series test does not contain the series number in the name for backwards compatibility.

A final point worth noting is that v3.0.0, v3.1.1, and v3.2.0 (AoS) are all of a similar speed, while SVN revision 6259 is about 16% faster.


markdewing:
okay to test

jtkrogel:
I'm not sure of the minimum version, but this worked with CMake 2.8.10

CMake/macros.cmake
```diff
@@ -290,7 +290,7 @@ ENDFUNCTION()
 # SHOULD_SUCCEED - whether the test is expected to pass or fail.  Expected failing tests will not have
 #                  the scalar tests added.
 
-FUNCTION(QMC_RUN_AND_CHECK BASE_NAME BASE_DIR PREFIX INPUT_FILE PROCS THREADS SCALAR_VALUES SERIES SHOULD_SUCCEED)
+FUNCTION(QMC_RUN_AND_CHECK BASE_NAME BASE_DIR PREFIX INPUT_FILE PROCS THREADS SHOULD_SUCCEED)
```
markdewing: Could you update the documentation of the function arguments above?   Some changes in the order, plus the added optional argument pairs for series and scalar values.

tests/system/CMakeLists.txt
```diff
@@ -326,6 +326,66 @@ ELSE()
...
+  LIST(APPEND LI2STOSJS_VMC_SCALARS "totenergy" "-14.947214  0.000502") # total energy
+  LIST(APPEND LI2STOSJS_VMC_SCALARS "variance"  "  0.170618  0.001035") # energy variance
+  LIST(APPEND LI2STOSJS_DMC_SCALARS "totenergy" "-14.989438  0.000643") # total energy
+  LIST(APPEND LI2STOSJS_DMC_SCALARS "variance"  "  0.151118  0.000312") # energy variance
+
+  QMC_RUN_AND_CHECK(short-li2_sto-sj_dmc
+                    "${CMAKE_SOURCE_DIR}/tests/molecules/Li2_STO_ae"
+                    Li2.STO.short
+                    Li2.STO.short.in.xml
+                    16 1
```
markdewing: These tests are only added for MPI builds - is it preferable to have 16 procs and 1 thread or 1 proc and 16 threads (the latter would at least work on non-MPI builds)?

prckent:
Good catch. An new / updated PR is needed. Unless we are seeking to test MPI functionality (which we do in other tests) 16 threads and 1 (MPI) process is preferred.

markdewing:
The tests don't work with SOA builds.  But maybe we want to leave that failure in place to motivate fixing the code.

prckent:
@markdewing Exactly. The SoA work is far from being done and this is an easy way to identify what does not yet work. The idea is to put tests for all part of the code that we intend to preserve moving forward. This is needed for SoA in the short term and Kokkos etc. in the long term.

jtkrogel:
OK, I can put in a new PR switching to 1x16 and updating the docs.

Another set of fixes that might be worth putting in place are actual aborts when someone tries to use e.g. the complex/gpu code on gaussian/sto inputs and have tests catching the expected aborts.

ye-luo:
LCAO with SoA is being worked. Please leave the tests always on.

prckent:
Skip with GPU and Complex. Run with SoA.

Issue #349, opened 2017-08-29
OIC5 toolchain is out of date and should be replaced by configure script


prckent:
Discovered as part of #337 . The toolchain uses old libraries and might use old compilers depending on which modules the user has loaded. A script would reduce ambiguity and should be able to replace the toolchain.

jtkrogel:
Configure script added, issue resolved.

PR #405, opened 2017-10-06
Add build script for OIC

jtkrogel:
Addresses #349.  A build script is added for OIC5 for builds with C++11 support.  The old toolchain file is still operational for older builds.  Both the build script and toolchain will be removed when OIC5 is decommissioned at the end of FY2018.

### DIFF
--- a/config/build_oic.sh
+++ b/config/build_oic.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+module () 
+{ 
+    eval `/opt/modules/3.1.6/bin/modulecmd bash $*`
+}
+module purge
+module load gcc/4.9.3
+module load mpi/openmpi-1.4.5-gcc4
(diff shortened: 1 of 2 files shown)


markdewing:
okay to test

prckent:
Did you try to get AMD libm included, or are we just going to note that this is a decent but not great build? e.g. Intel compilers and libraries on AMD processors. (Note that I have not checked how much of a problem this is in practice, but we have all seen the Intel optimization warning.)

jtkrogel:
I did not try with AMD libm.  The build script can be a starting point to be built upon down the road as needed.  I did not see a performance hit moving from 3.1.1 AMD libm to this build style for small systems, but I haven't tried anything very large.

I will add a descriptive header and rename.

prckent:
OK. libm (and vectorized sin/cos in general) is less relevant for large systems. It is only for the phase factor evaluation.

PR #408, opened 2017-10-06
Remove old toolchain for OIC. Use new build_ornl_oic.sh script

prckent:


prckent:
Also confirming that the build_ornl_oic.sh script works

PR #410, opened 2017-10-09
Fix a bug in PairCorrEstimator.

ye-luo:
No more crash in short-heg_14_gamma-hf-1-16


prckent:
OK to test

prckent:
One more reason for routine memory checks

PR #414, opened 2017-10-12
Remove unused PbyP Hamiltonian routines.

ye-luo:
PbyP in Hamiltonian is not used and no more maintained.
In practice, it is not useful.
@rcclay If I missed something to remove, let me know.


prckent:
OK to test

PR #417, opened 2017-10-13
Clean up wavefunction

ye-luo:
Apart from some internal interface name changes in J1 SoA, all the rest is only removing no more needed routines.
ratio(P,iat,G,L), logRatio
evaluateLog(P,buf)
dumpToBuffer, dumpFromBuffer


markdewing:
okay to test

prckent:
Are any of these functions used at the higher build levels? Called by other historical code?

I very much agree with this cleanup. We should note here if some reimplementation might be needed in future to reactivate any features.

ye-luo:
We are building at level 3 by default. 3 has the most features. 5 has been checked.
Most of these are only partially implemented. They are deprecated because either the name doesn't reflect the actual implementation and their role has been replaced by better routines.
evaluateLog(P,buf) is implemented everywhere but it is more flexible to use the way we have now
by calling evaluateLog(P) + updateBuffer(P,buf,false)

prckent:
I am happy to merge but will wait until next week to allow others to comment.

PR #418, opened 2017-10-15
Fix double free pointer crashing in heg test input.

ye-luo:
With this fix, processPWH return true properly.


markdewing:
okay to test

prckent:
Please could you add a comment here about what in the HEG inputs trips this bug? The bug is not HEG specific.

ye-luo:
The HEG tests with the old input style with 3 include and href inside trigger the issue.
The bug is not HEG related but the include related.

PR #422, opened 2017-10-17
Fix reconfiguration="yes" by updating UnitZeta each generation.

adbacze:
Minimal fix needed to enable reconfiguration="yes". UnitZeta and DeltaStep (the position of the comb 'teeth') needs to be randomized at each generation to fix the bias error reported in Issue #18.


markdewing:
okay to test

prckent:
Please can you add a small note in the manual about this option.

prckent:
Lets merge now and update docs when the other options implemented and explored.

Issue #419, opened 2017-10-17
SkEstimator in h5 file is incorrect using all-particle move VMC


Paul-St-Young:
The SkEstimator values in hdf5 file are incorrect in VMC if all-particle move is used. The problem is virtually undetectable in all-particle DMC. I produced this problem using the current development branch (commit af960d3d6bde89956998d61d36ed4c9995a58e96) in the attachment. The results can be plotted using the included script "split_sk.py". The resultant plot is also attached here.

I tried playing with the calls `W.update()`, `H.auxHevaluate()` and `thisWalker.Age` in `VMCUpdateAllWithDrift::advanceWalkers` with no success. These are the only obvious differences compared to the all-particle DMC driver that I noticed.

Does anyone have an idea on why S(k) might be broken in VMC but not DMC?
Any suggestion is highly appreciated!

![broken-sk](https://user-images.githubusercontent.com/8060441/31642737-24d378b4-b2b2-11e7-9385-780aa42ca443.png)

[sk-wbyw-vmc.zip](https://github.com/QMCPACK/qmcpack/files/1389492/sk-wbyw-vmc.zip)

jngkim:
It is how the exit condition is handled in wbyw not S(k).

S(k) is not stored and has to be computed.
Needs to rewrite update so that auxH stuff is called even though a move is
rejected. Just rearranging if conditions.

This is happening in DMC as well but not noticeable because of the
acceptance rate.

I proposed Ye to get rid of G in Walker but now I know why we have to keep
them.

You can communicate with me directly until the bug is fixed.
Jeongnim

Paul-St-Young:
Hi Jeongnim,

Thanks for the explanation! Adding another `H.auxHevaluate` before `H.rejectedMove` seem to have resolved this problem. I am almost certain I tried this before. I probably put the fix in the wrong variant of advanceWalker ...

Thanks,
Paul

jngkim:
You need to call P.update or something to compute S(k) with the old
position.

I will send you the proposed changes.

jngkim:
This is just to give you an idea how to fix VMC wbyw.
Need to debug but this will show what I mean.

You should consider merging wbyw with/without drift. This can be done
nicely with lambda function.

Jeongnim

PR #423, opened 2017-10-18
fixed hdf5 observables in all-particle driver

Paul-St-Young:
also enabled substep in all-particle VMC

### DIFF
--- a/src/QMCDrivers/VMC/VMCUpdateAll.cpp
+++ b/src/QMCDrivers/VMC/VMCUpdateAll.cpp
@@ -35,139 +35,55 @@ VMCUpdateAll::~VMCUpdateAll()
 
 void VMCUpdateAll::advanceWalker(Walker_t& thisWalker, bool recompute)
 {
-    makeGaussRandomWithEngine(deltaR,RandomGen);
-    //if (!W.makeMove(thisWalker,deltaR, m_sqrttau))
-    if (!W.makeMove(thisWalker,deltaR,SqrtTauOverMass))
-    {
-      H.rejectedMove(W,thisWalker);
-      return;
+  /* thisWalker.R will track the last accepted configuration
+   * W.R will track the proposed configuration 
+   *
+   * upon Call:
+   *   thisWalker.R,G,L,DistTables,SK,Properties must be consistent
+   *   recompute flag is for pbyp driver and is not used here
+   *
+   * upon Return:
+   *   thisWalker.R,G,L,DistTables,SK,Properties must be kept consistent
+   * */
+  bool updated=false;
+  W.loadWalker(thisWalker,false); // W.R,G,L = thisWalker.R,G,L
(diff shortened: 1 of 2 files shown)


Paul-St-Young:
fixed bug from #419

Paul-St-Young:
I cleaned the VMCUpdateAll.cpp file. @jngkim and I revamped the `advanceWalker` functions to allow substep. I was a bit heavy-handed with the comments, please let me know if some of it needs to be trimmed.

I did not test `advanceWalkerForEE` because I do not have an example for it. I hope the CI has tests for this method.

markdewing:
okay to test

src/QMCDrivers/VMC/VMCUpdateAll.cpp
```diff
@@ -35,139 +35,55 @@ VMCUpdateAll::~VMCUpdateAll()
...
-    if (RandomGen() > g)
-    {
-      thisWalker.Age++;
-      ++nReject;
-      H.rejectedMove(W,thisWalker);
+  /* thisWalker.R will track the last accepted configuration
+   * W.R will track the proposed configuration 
+   *
+   * upon Call:
+   *   thisWalker.R,G,L,DistTables,SK,Properties must be consistent
```
ye-luo: I think thisWalker doesn't carry DistanceTables and SK.

Paul-St-Young: @ye-luo that is a good point, I was confused. 

It seems that ParticleSet::loadWalker transfers R,G,L from thisWalker to W, but only update W.DistTables or SK for pbyp driver. I think this means the internal states of W become inconsistent after calling saveWalker in the wbyw driver. 

If there a reason why the `if (pbyp)` has to be in `ParticleSet::loadWalker`?

ye-luo: This is necessary.
In pbyp moves, e-I distance table need to be up-to-date only when JeeI is used. SK need to be up-to-date if SK is needed at every pbyp move.
In allp moves, nothing is needed. Only update distance table and SK for proposed moves, if the last move is rejected, update everything with the old configuration before Hamiltonian and estimators.

At this line, please say `thisWalker.R,G,Properties must be consistent`

src/QMCDrivers/VMC/VMCUpdateAll.cpp
```diff
@@ -35,139 +35,55 @@ VMCUpdateAll::~VMCUpdateAll()
...
+   *   recompute flag is for pbyp driver and is not used here
+   *
+   * upon Return:
+   *   thisWalker.R,G,L,DistTables,SK,Properties must be kept consistent
```
ye-luo: At this line, please say `thisWalker.R,G,Properties must be kept consistent`

ye-luo:
@prckent if no more changes are needed, this is good to merge.

prckent:
Thanks all.

PR #424, opened 2017-10-20
Add msd tests

ye-luo:
C2 molecule. 20 sec on my desktop. VMC only


grahamlopez:
OK to test

ye-luo:
@prckent How much longer is good? 20mins(60x)?

jtkrogel:
I've typically made the long test 10x longer than short, aiming for about 5min for long.

prckent:
After discussing with @ye-luo , we'll merge now and add the long test next week.

jtkrogel:
Possible to add a reasonable length DMC test as well?

ye-luo:
@jtkrogel Yes. That is the next step.

prckent:
What would a DMC run test that the VMC test does not? Why is there a concern that a DMC run is needed? i.e. In what way is a VMC test not sufficient?

ye-luo:
VMC with Drift is largely identical to DMC. I think VMC is sufficient. We need to leave DMC tests on lighter test cases.

Issue #415, opened 2017-10-12
Three body Jastrow factor description needed in manual


prckent:
This is currently an empty section.

prckent:
Partially covered by #426

prckent:
This could be based on the text in our citation paper.

PR #426, opened 2017-10-20
Add feature documentation by K. Esler and S. Chiesa

jtkrogel:
Add feature documentation for optimized long-range breakup (Ewald), cubic splines, B-spline orbital tiling, hybrid representation, e-e-I Jastrow, and k-space Jastrow.

Notes are taken from QMCPACK and wfconvert files with permission from K. Esler.  These notes should be revised/corrected and expanded on in the future.


markdewing:
okay to test

ye-luo:
Please DO NOT move the eeI Jastrow to the user manual. Apart from the first formula which may be useful to users, all the rest is meant only for developers. I will take care the user part in #415.

ye-luo:
Since the files are incorporated into the manual, shall we remove individual files in the repo and we just update the manual?

prckent:
@ye-luo eeI description is needed soon - it is an everyday feature with no docs!

I agree on removing the individual files. Not needed once included in manual.

jtkrogel:
Look OK now?

PR #427, opened 2017-10-21
Fix MSD on BGQ

ye-luo:
1, Workaround the MSD bug on BGQ with Clang.
2, Add quantum package input/output of C2.
3, Add a C2 long test


markdewing:
okay to test

prckent:
Why is this bug triggered on BGQ/BGClang? Is there a fix in the works? What about regular CLANG? Can the code be written a different way to avoid the #ifdefs?

ye-luo:
1, Need to look at regular Clang.
2, ifdef=true should always work. But I prefer the cleaner ifdef=false code path.

prckent:
Please check CLANG. I would expect a subtle bug on our side if that fails.

markdewing:
What is the failure mode?  The added unit test failing?  I tried clang 4 and 5 on x86_64 and neither failed the unit test.

ye-luo:
@markdewing Neither BGClang and x86 clang failed the unit test.
I had thought I could catch something and wrote the unit test but it didn't help. I just leave it there.
I still don't understand why these lines failed on BGQ. I caught them by printf.

ye-luo:
x86 Clang 3.8 was giving the right answer without the workaround.

prckent:
@markdewing Are you happy if we merge as is? i.e. With an ifdef code path for bgq?

markdewing:
Yes, I'm okay with an ifdef code path for bgq.   It might be nice if the comment said more about the failure mode (in this case I think it is giving wrong answers?)

ye-luo:
I added more detail about the failure.
@markdewing has verified that x86 with Clang 4.0 is also correct.

PR #428, opened 2017-10-21
Eliminate unroll warning from Intel compilers.

### LINKED ISSUE #376: SoA code and Intel compiler gives many mismatched unroll warnings
Should be tidied. These flood the cmake output.
>remark: unroll pragma will be ignored due to unroll factor exists
>remark: unroll pragma will be ignored due to unrolling factor mismatched

ye-luo:
closes #376


markdewing:
okay to test

Issue #326, opened 2017-08-17
LMY engine MPI interface in formic/utils fails to compile due to missing C++ MPI wrappers


atillack:
I am getting multiple compile failures like these:

In file included from /ccs/home/andreas/codework/QMCPACK_MAIN_DEV/src/formic/utils/numeric.cpp:9:
/ccs/home/andreas/codework/QMCPACK_MAIN_DEV/build_summitdev_cpu_xl/src/formic/utils/mpi_interface.h:66:16: error: use of undeclared identifier 'MPI'
  inline const MPI::Comm & world() {
               ^

The error is due to the mpi_interface.h header using C++ bindings for MPI which have been deprecated for a long time and at least on SummitDev are no longer present at all.

Ironically, the error is triggered now because our C++11 detection that used to fail on SD and switched of the LMY engine now works wonderfully, leaving the LMY code on.

A work-around is to switch off LMY for that scenario (a cmake switch would be nice, but I can live with manually editing CMakeLists.txt)

The real solution is to get rid of the C++ bindings in mpi_interface.h and replace them with the C standard ones which is probably the preferred one as more and more installations will want to use newer MPI versions (w/o C++ bindings).

ye-luo:
You can use -D BUILD_LMYENGINE_INTERFACE=0 to switch off the LMY part instead of editing CMakeLists.txt

atillack:
@ye-luo Thanks, I tried that first, but it gets overwritten in CMakeLists.txt line 191 where it gets set back to 1.

ye-luo:
@atillack It should not be. The command line variable is cached and that overwrite only happen when you don't specify anything.

atillack:
@ye-luo -DBUILD_LMYENGINE_INTERFACE=0 is working now. Maybe my build script didn't update in time.

The issue with the C++ bindings for formic/utils/mpi_interface.h remains though.

ye-luo:
Good. I remember I checked the SoA PR manually on summitdev using the instructions you wrote in the manual and at that time everything was fine.

prckent:
In some months, the MPI interface will be redone to solve this problem. In the interim, disabling the LMYENGINE via -DBUILD_LMYENGINE_INTERFACE=0 is a workaround for incompatible MPI versions. One possibility would be to do a test compile as part of the CMake configuration and disable the incompatible code automatically.

atillack:
[...]

PR #432, opened 2017-10-24
Use the C MPI API in the formic library.

markdewing:
The C++ bindings to the MPI library were deprecated and are removed in MPI-3.
This change should enable QMCPACK to compile with newer versions of MPI
implementations.

Addresses issue #326

Some issues:

There is no MPI_* type corresponding to MPI::Bool.  Hopefully translating to
MPI_INT should work okay.

MPI-3 added 'const' to a number of parameters in the C interface.
These 'const' qualifiers had already been present in the C++ interface.
When compiling with MPI-2.2 (or older), the formic wrappers use a
'const T*' to call a function with a non-const 'T *'.
Newer GCC versions (7.1) give an error about converting from 'const' to non-'const'.

The fix is to add const_cast's to remove the 'const' qualifier.

### DIFF
--- a/src/formic/utils/lmyengine/block_mat.cpp
+++ b/src/formic/utils/lmyengine/block_mat.cpp
@@ -313,7 +313,7 @@ void cqmc::engine::LMBlockerMatData::mpi_finalize(const double total_weight) {
 
   // <wfn|wfn>
   double m_ww_tot = 0.0;
-  formic::mpi::reduce(&m_ww[0], &m_ww_tot, 1, MPI::SUM);
+  formic::mpi::reduce(&m_ww[0], &m_ww_tot, 1, MPI_SUM);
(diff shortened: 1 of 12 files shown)


ye-luo:
Has the comm_world been solved?

prckent:
@ye-luo No. That is a totally separate issue.

prckent:
@markdewing Do you plan to add const_casts? Perhaps the overlap of old MPI and new compilers is small.

markdewing:
@prckent The necessary const casts are already there (in mpi_interface.cpp.in)

prckent:
OpenMPI 1.10.6 with gcc 4.8.5, OpenMPI 2.0.2 with gcc 4.8.5, and Intel MPI 2018 with Intel 2018 compilers checked manually on oxygen. Built successfully and adaptive optimizer tests run.

( @markdewing Sorry, missed the casts in a quick scan of the changes )

prckent:
I checked that @eneuscamman is OK for us to merge this.

ye-luo:
MPICH 3.2 Intel 17.0.2, I got tons of const warning.

PR #435, opened 2017-10-25
Update default optimizer to OneShiftOnly.

### LINKED ISSUE #19: Defaults for linear optimization are required
Reported by: jtkrogel

Set sensible defaults for as many input parameters as possible for the "linear" optimizer.  Preferable to complete prior to the school as newcomers (or veterans!) shouldn't have to turn many knobs.  If this is not simple to do then the optimizer should either be fixed or very detailed explanations/examples of driving the optimizer should be included in the manual.

ye-luo:
Manual is updated as well.
closes #19

### DIFF
--- a/manual/opt.tex
+++ b/manual/opt.tex
@@ -293,10 +293,12 @@ Recommendations:
     <parameter name="max_param_change"> 0.3 </parameter>
     <parameter name="chase_lowest"> yes </parameter>
     <parameter name="chase_closest"> yes </parameter>
-    <parameter name="block_lm"> yes </parameter>
-    <parameter name="nblocks"> 100 </parameter>
-    <parameter name="nolds"> 5 </parameter>
-    <parameter name="nkept"> 3 </parameter>
+    <parameter name="block_lm"> no </parameter>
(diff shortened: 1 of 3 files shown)


jtkrogel:
It doesn't have to be part of this PR, but it would be nice if a complete/functional '<qmc> ... </qmc>' listing was provided for each of quartic, adaptive, and OneShiftOnly in the manual.  Right now the user has to assemble one based on the instructions, but it would be more convenient to have a copy/paste-able example for each.

prckent:
@jtkrogel Agreed

I will mandate full & functional sections for oneshift and adaptive shift before merging this PR.

prckent:
OK to test

ye-luo:
@eneuscamman I added the VMC part in for the adaptive optimizer. The optimizer aborts with some error. Could you check what is missing? The optimizer should be fine since the nightly H4_ae is good.

ye-luo:
Full optimization sections have been added in the manual.

prckent:
Thanks Ye and Eric.

Issue #437, opened 2017-10-25
Updated formic MPI generates many warnings due to const return types


prckent:
Both Intel compilers generate warnings about meaningless const return types:

```

/.../build_intel2017/src/formic/utils/mpi_interface.h(64): warning #858: type qualifier on return type is meaningless

    inline const MPI_Comm world() {
           ^
```

The Intel compilers consistently generate 272 warnings per build ( https://cdash.qmcpack.org/CDash/index.php?project=QMCPACK ). I think we simply don't need the const here.

PR #438, opened 2017-10-25
Remove const from return in formic MPI wrappers.

markdewing:
Depending on how the MPI implementation defines these values, the Intel compiler
may or may not warn about "type qualifier on return type is meaningless".
(If the return value is a class, no warning.  If the return value is a plain type,
then there is a warning.)

Addresses issue #437.


prckent:
Thanks. It is an interesting question how we should test for this if it turns out to be needed with certain MPIs. Lets see what happens in cdash.

PR #439, opened 2017-10-25
Update Cray build setting.

ye-luo:
export CRAYPE_LINK_TYPE=dynamic
is a more robust way to set dynamic linking on Cray.

Solves the ALCF Theta CMake issue after UP04 update.

Need to check OLCF Cray machines.
NERSC should be good already due to the similarity to ALCF Theta.


ye-luo:
I have checked OLCF Cray machines but have no access to NERSC.

ye-luo:
Edison and Cori on both HSW and KNL partitions are good.

prckent:
+1

PR #445, opened 2017-10-26
Update DMC timers.

markdewing:
Use new registration function and reference the timer list by an enum name rather than a number.


ye-luo:
Does the Stack timer profile print time in branching now?
Last night, it didn't.
https://cdash.qmcpack.org/CDash/testDetails.php?test=1029622&build=13608

markdewing:
That run has only 1 MPI process.
It does print the branching time for runs with > 1 MPI process (and the timers enabled on medium or fine)

ye-luo:
This will be very helpful for my following study on time spend on MPI.

markdewing:
This doesn't add any new timers to the code, it only cleans up some that were already there.

Issue #446, opened 2017-10-26
Need timer for checkpoint IO


ye-luo:
QMCDriver::recordBlock generates checkpoints.
Timers are needed there.

ye-luo:
@markdewing Thanks, it works well.

PR #447, opened 2017-10-27
Add checkpoint timer

markdewing:
Addresses #446

### DIFF
--- a/src/QMCDrivers/QMCDriver.h
+++ b/src/QMCDrivers/QMCDriver.h
@@ -26,6 +26,7 @@
 #include "Configuration.h"
 #include "OhmmsData/ParameterSet.h"
 #include "Utilities/PooledData.h"
+#include "Utilities/NewTimer.h"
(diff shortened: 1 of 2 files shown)


src/QMCDrivers/QMCDriver.h
```diff
@@ -399,6 +400,8 @@ class QMCDriver: public QMCTraits, public MPIObjectBase
   std::string getRotationName( std::string RootName);
   std::string getLastRotationName( std::string RootName);
 
+  NewTimer *checkpointTimer;
+
```
ye-luo: I saw above
std::vector<NewTimer*> myTimers;
It is it beneficial to engage the above one?
Probably this will introduce some entanglement with the enumeration in the VMC/DMC drivers.

markdewing: Probably not, since it's not used in this class.
What should probably be done is to move that declaration of myTimers to the individual classes that use it.

ye-luo: I agree.

PR #449, opened 2017-10-28
Fix a bug in AoS JeeI, SoA was correct.

ye-luo:
During optimization, evaluateDerivatives pollutes internal data and causes wrong NLPP evaluation in checkConfigurations. Then the proposed pathological large change of parameters is rejected. Only affects AoS NLPP case.

### DIFF
--- a/src/QMCWaveFunctions/Jastrow/eeI_JastrowOrbital.h
+++ b/src/QMCWaveFunctions/Jastrow/eeI_JastrowOrbital.h
@@ -1230,9 +1230,6 @@ public:
       dLogPsi=0.0;
       gradLogPsi = PosType();
       lapLogPsi = 0.0;
-      RealType u;
-      PosType gradF;
-      Tensor<RealType,3> hessF;
       // Now, evaluate three-body term for each ion
       for (int i=0; i<Nion; i++)
       {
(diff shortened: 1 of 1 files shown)


markdewing:
okay to test

prckent:
@ye-luo Please send a note to Berkeley about this. I am surprised no one has caught this before, but it is clearly very important.

prckent:
How was this problem identified? Can we check for it?

ye-luo:
Hyeondeok by accident run the AoS binary and the optimization of JeeI blows up at the second step when loose controlling OneShiftOnly. I ran the SoA, it was fine because I rewrote evaluateDerivatives. By comparing these two code path, I noticed this bug. The derivatives are all correct but this bug affects the following NLPP evaluation.

You can check your VMC scalar.dat with "qmca -e 0" and grep the output file "VMC Eavg" (recomputed by the Optimization driver). If they don't agree well, there is a problem. The should not agree exactly in principle since NLPP contains random numbers but they should not disagree largely. The case from Hyeondeok, qmca gives about -1300 Ha but VMC Eavg is -1440.

ye-luo:
@prckent note sent.

PR #451, opened 2017-10-31
Merge vmc drivers

ye-luo:
Merge VMC P-by-P/All-P drift and nodrift.
Refine timer tags.


markdewing:
okay to test

jngkim:
It does not have wbyw to begin with.

prckent:
Fusing run / runwithdrift and advancewalkers / advancewalkerswithdrift will save a few pages of source code and - more importantly - mean that there are fewer instances of repeated code and repeated algorithm specification.

jngkim:
VMC_CUDA is pretty compact. My dislike of if-statement is rather extreme,
but the saving in this case is very little, once you have ifs and macros
all over the places.

ye-luo:
@prckent In the VMC_CUDA, w/ drift and w/o drift are not very identical. w/o drift has some WF optimization routine inside. So more attention is needed and it will be solved in a different PR.

prckent:
@ye-luo OK, please address the CUDA drivers in a new PR.

Your comment totally proves my point about needing single implementations :-)

Happily the optimization "forOpt==true" parts seem identical, but there are various silly differences in the driver code. e.g.  VMCcuda::run() has Esum as a local variable, but after initialization it is unused, so should be deleted. runWithDrift does not have it. The forOpt parts are triggered by put() in correlatedSamplingLinearOptimize via setOpt(bool opt). 

@markdewing Should review next.

ye-luo:
@markdewing I started the merge of drivers first and got pbyp done but had to wait all-p PR last week. Meanwhile, to confirm my modification is effective, I modified the timer and they landed in one PR. I will separate them next time.

PR #457, opened 2017-11-02
RMC returns pure estimates of Collectables.

rcclay:
Previously, RMC returned mixed estimates for collectables (density, S(k), etc).  Now I'm forcing RMC to return pure estimates for these quantities.


markdewing:
okay to test

rcclay:
Huh.  What's the error with the build tests?

markdewing:
I think there's some parenthesis nesting errors
Excerpt from CPU build log:


[ 72%] Building CXX object src/QMCHamiltonians/CMakeFiles/qmcham.dir/DensityMatrices1B.cpp.o
/autofs/nccs-svm1_home1/bgl/continuous_integration/qmcpack-agent/workspace/qmcpack-rhea-cpu/src/QMCDrivers/RMC/RMCSingleOMP.cpp: In member function 'void qmcplusplus::RMCSingleOMP::resetRun()':
/autofs/nccs-svm1_home1/bgl/continuous_integration/qmcpack-agent/workspace/qmcpack-rhea-cpu/src/QMCDrivers/RMC/RMCSingleOMP.cpp:316:26: error: qualified-id in declaration before '(' token
   bool RMCSingleOMP::put (xmlNodePtr q)
                          ^
/autofs/nccs-svm1_home1/bgl/continuous_integration/qmcpack-agent/workspace/qmcpack-rhea-cpu/src/QMCDrivers/RMC/RMCSingleOMP.cpp:323:36: error: qualified-id in declaration before '(' token
   void RMCSingleOMP::resetReptiles (int nReptiles_in, int nbeads_in,
                                    ^
/autofs/nccs-svm1_home1/bgl/continuous_integration/qmcpack-agent/workspace/qmcpack-rhea-cpu/src/QMCDrivers/RMC/RMCSingleOMP.cpp:391:2: error: expected '}' at end of input
 };
  ^



/autofs/nccs-svm1_home1/bgl/continuous_integration/qmcpack-agent/workspace/qmcpack-rhea-cpu/src/QMCDrivers/RMC/RMCUpdateAll.cpp: In member function 'void qmcplusplus::RMCUpdateAllWithDrift::advanceWalkersRMC()':
/autofs/nccs-svm1_home1/bgl/continuous_integration/qmcpack-agent/workspace/qmcpack-rhea-cpu/src/QMCDrivers/RMC/RMCUpdateAll.cpp:597:17: error: redeclaration of 'qmcplusplus::QMCUpdateBase::Walker_t& centerbead'
       Walker_t& centerbead = W.reptile->getCenter();
                 ^
/autofs/nccs-svm1_home1/bgl/continuous_integration/qmcpack-agent/workspace/qmcpack-rhea-cpu/src/QMCDrivers/RMC/RMCUpdateAll.cpp:368:16: note: 'qmcplusplus::QMCUpdateBase::Walker_t& centerbead' previously declared here
     Walker_t & centerbead = W.reptile->getCenter ();
                ^
[ 75%] Building CXX object src/QMCHamiltonians/CMakeFiles/qmcham.dir/model/Lennard

ye-luo:
@rcclay Could you call my office?

rcclay:
@ye-luo Can I call you after lunch?  Can you send your phone number to my email?

rcclay:
After talking with Ye, we both agree that collectables depending explicitly on the trial wavefunction will not have proper pure estimate accumulation, but mixed & pure estimates for properties, and pure estimates for other collectables are fine.  @ye-luo I can't remember:  can we leave a note to this issue, or should I fix it now?

prckent:
Please give some examples of which observables will be pure and which mixed in plain English. i.e. For people not familiar with how QMCPACK works internally.

jtkrogel:
So this mainly effects density-matrix (-like) estimators (density matrix, momentum distribution, non-local pseudopotential energy contribution)?  Anything else?

prckent:
Perhaps there should be switch for different behavior? I can imagine having a mix of contaminated (mixed) and pure estimates to be very confusing. At least some very clear labeling is needed. i.e. Different labels for mixed and pure, because they are not the same thing.

rcclay:
Pretty much everything that defaults to dumping to scalar.dat will have mixed and pure estimates accumulated, and these will be clearly marked with the suffix _p or _m for pure and mixed estimates respectively.  As Jaron said, observables like g(r), S(k), density, etc., will now be collected only over the pure distribution (after all, why are we running RMC to begin with?).  The current implementation will give garbage results if you try to evaluate the 1-RDM, which is not stored on the walker and has an explicit dependence on the trial wavefunction.  Of course, I'm not too concerned, because even with a perfect implementation, we can't get pure estimates of these quantities correctly anyways in RMC without significant algorithmic improvement.

jtkrogel:
Can we write up in the manual what our evaluation formulas are for "pure" and "mixed" observables in RMC?  I guess form there it would be pretty easy to see that e.g. the non-local pseudopotential is not evaluated purely and it will help users reason about what they are/are not calculating.

rcclay:
I don’t know how much detail I should go into on the manual—whether its worth explaining the code and theory or whether its better to just add some “Do’s” and “Don’ts”.  Basically, all observable evaluation happens on the mixed distribution like in DMC, since this just requires a walker and a trial wavefunction evaluated for that walker.  Transitioning to the pure estimates of observables ***diagonal in the position basis*** is done by either explicitly (forward walking) or implicitly (RMC) reweighting the observable...  so most of the trickiness happens in observable accumulation rather than in computing the observable itself.

prckent:
@rcclay This area won't be obvious to people who have not been thinking of reptation as much as you, so we need to keep them out of trouble. I also agree with @jtkrogel . I suggest a list of what is computed and some do's and don't . My concern is that we don't leave subtleties that you know about undocumented, nor do we change the meaning of output data (pure/mixed), if that is the case.

rcclay:
@prckent Sounds good.  I'll push an updated manual after lunch.  Once that's done, is there anything else this pull request needs?

src/QMCHamiltonians/QMCHamiltonian.cpp
```diff
@@ -537,6 +537,49 @@ void QMCHamiltonian::auxHevaluate(ParticleSet& P, Walker_t& ThisWalker)
     auxH[i]->setParticlePropertyList(P.PropertyList,myIndex);
   }
 }
+///Evaluate properties only.
```
ye-luo: @jtkrogel Do we have Hamiltonians or estimators contributing to both properties and collectables? if this is true, @rcclay calling auxHevaluateProperties is not safe.
@rcclay I feel that the code duplication can be avoided by introducing an extra optional argument for picking the three mode, both(default), properties, collectables.

src/QMCDrivers/RMC/RMCUpdatePbyP.cpp
```diff
@@ -527,16 +386,13 @@ namespace qmcplusplus
...
 
 	prophead.Properties (LOCALENERGY) = eloc;
 	prophead.Properties (R2ACCEPTED) = rr_accepted;
 	prophead.Properties (R2PROPOSED) = rr_proposed;
-	H.auxHevaluate (W, prophead);
-	//H.saveProperty(overwriteWalker.getPropertyBase());
+	H.auxHevaluateProperties (W, prophead);
 	H.saveProperty (prophead.getPropertyBase ());
-	//overwriteWalker.Age=0;
+
```
ye-luo: Remove white space line.

src/QMCDrivers/RMC/RMCUpdatePbyP.cpp
```diff
@@ -552,6 +408,11 @@ namespace qmcplusplus
 	W.reptile->flip ();
 	// return;
       }
+      //Collectables should be evaluated on center bead.  Here we go.
+      Walker_t& centerbead = W.reptile->getCenter();
+      W.loadWalker(centerbead,true);
+      W.update(false);  //Called to recompute S(k) and distance tables.  Recompute S(k)?  False.
```
ye-luo: @rcclay false means skipSK=false. update your comment and remove false(default).

src/QMCHamiltonians/QMCHamiltonian.cpp
jtkrogel: I thought all propertylist contributing QMCHamiltonianBase objects were explicitly required to not be collectable.  See e.g. QMCHamiltonianBase::addValue.

jtkrogel: @ye-luo Something looks odd with setParticlePropertyList in this regard.  In the regular code, setParticlePropertList is called for all auxH, but I don't see any guard against collectables.  In the case of collectables, it looks like myIndex remains set to -1; is there a sneaky index trick to have all collectables write to the same (safe) place by using index of -1?

jtkrogel: In any case this does not look any worse than what is there already...

src/QMCDrivers/RMC/RMCUpdatePbyP.cpp
```diff
@@ -302,6 +254,9 @@ namespace qmcplusplus
 	gf_acc = 1.0;
 	nReject++;
       }
+      Walker_t& centerbead = W.reptile->getCenter();
+      W.loadWalker(centerbead,true);
+      H.auxHevaluateCollectables(W,centerbead);
```
ye-luo: need W.update() before auxHevaluate

src/QMCHamiltonians/QMCHamiltonian.cpp
jtkrogel: OK, now I see (remembering).  Collectables always need to define the virtual setParticlePropertyList function to be empty.  Will have to check if some do not do this...

jtkrogel: ... only true if using hdf5 exclusively.

jtkrogel: Some collectable estimators use either one, e.g. MomentumEstimator, while others, e.g. SkAllEstimator just use the propertylist.

jtkrogel: Why is the call to auxHevaluateProperties not safe in this case?

ye-luo: I really feel the handling of data collection needs to be looked at.
PropertyList takes care of the scalar.dat for small dataset.
Collectables takes care of h5 for large dataset.

jtkrogel: It is a bit odd.  I expect writing of collectables to scalar.dat has been done for the relative convenience/accessibility/robustness of post processing ascii files rather than hdf5.  I'm OK with it so long as both paths are provided correctly.

ye-luo: @jtkrogel auxHevaluateProperties checks the QMCHamiltonianBase::COLLECTABLE flag. If true, skip the evaluation.

jtkrogel: OK.  In that case there is something to address.

src/QMCDrivers/RMC/RMCUpdatePbyP.cpp
```diff
@@ -411,7 +410,7 @@ namespace qmcplusplus
       //Collectables should be evaluated on center bead.  Here we go.
       Walker_t& centerbead = W.reptile->getCenter();
       W.loadWalker(centerbead,true);
-      W.update(false);  //Called to recompute S(k) and distance tables.  Recompute S(k)?  False.
+      W.update();  //Called to recompute S(k) and distance tables.  Recompute S(k)?  False.
```
ye-luo: Recompute is actually true. skipSK is false.

src/QMCHamiltonians/QMCHamiltonian.cpp
ye-luo: @jtkrogel You are right. How to IO (text or binary) is a separated issue. What I need to be clear is what should go to propertylist and what should be in collectables. propertylist should be minimized since it is carried by the walker.

src/QMCDrivers/RMC/RMCUpdatePbyP.cpp
rcclay: I agree.  I don't know why I said "recompute" in the comment.  Fixed.

src/QMCHamiltonians/QMCHamiltonian.cpp
jtkrogel: @ye-luo Good point.  At the very least all collectables should support hdf5 and, IMO, use it by default.

ye-luo: @rcclay once you merge all these routines, I will approve this PR.

rcclay: @ye-luo How does it look now?  At the moment, I prefer to keep the auxHevaluate() used by the rest of the code, and my overloaded auxHevaluate(do_properties,do_collectables) separate.  I can merge them even more if need be.

ye-luo: @rcclay Good for now.

rcclay:
Is there anything else this needs to be merged?

ye-luo:
@prckent your review is pending. Do you need more time to review?

PR #458, opened 2017-11-02
Remove no more used TrialWaveFunction::acceptTMove.

ye-luo:


prckent:
OK to test

ye-luo:
This is not used anywhere now.

PR #459, opened 2017-11-03
nexus: tiling fix when net_spin="low", gaussian basis plotting

jtkrogel:
Fix for when supercell has no spin imbalance, but primitive cell technically does (e.g. bcc Na).

Add ability to plot gaussian basis set primitives by width on a log scale.  Useful for comparing/truncating/merging basis sets.


markdewing:
okay to test

jtkrogel:
@prckent I highly agree.  It is really just a matter of relative prioritization w/ testing of native qmcpack features.

PR #460, opened 2017-11-03
Generalize time limit code

markdewing:
Extract the cpu time limit code from DMCOMP.cpp to a new class, RunTimeControl.
The remaining code in DMCOMP.cpp is a pattern for the code to be added to the other drivers.

The one change beyond refactoring is to add the MPI broadcast of the loop decision.  This prevents the case where different processors make different decisions about continuing the loop, which would cause QMCPACK to hang.


prckent:
Is this correct for ensemble runs? The  members of the ensemble should quit individually.

markdewing:
It should work for ensemble runs, but I haven't tested it.
The communicator for the broadcast should be the one local to the individual run.   All members of the ensemble need to measure the time from the beginning of the run, so a single RunTimeManager should be sufficient.

PR #463, opened 2017-11-03
nexus: fix failing LiH test

jtkrogel:
LiH example/test was failing because "l-local" was not a known/supported attribute of `<pseudo/>`.  It has now been added.

Also the example (which I did not write!) generated a truly large number of files for twist averaging (17^3=4913 files), which was rather slow.  I reduced this to a 7x7x7 twist grid.


prckent:
OK to test

Issue #433, opened 2017-10-24
Removing useBuffer option in WF optimization


ye-luo:
useBuffer was introduced before the analytic derivatives are implemented. Nowadays, this code path is no more useful and doubles work when new features are introduced. I'm going to remove this code path. Let me know if there is any concern by Friday Oct. 27.

jtkrogel:
Could you explain a little more about what functionality will be lost by removing this option?

ye-luo:
Basically nothing. I think all the optimization executes well as no buffer is the default.

prckent:
How was usebuffer used in the past?

ye-luo:
usebuffer saves static components during optimization. For example, it saves the determinants of each sample (not just each walker) in a Jastrow-only optimization. When a new set of parameters is proposed, the correlated sampling (CS) can use the saved determinants to update NLPP if nonlocalpp=yes.
if nonlocalpp=no, only wasting memory.
if nonlocalpp=yes, oneshiftonly does only 1 CS, adaptive does 3 CS, quartic does 7 CS.
In a large simulation, you can not use it because of the memory. In a small simulation, recomputing is so cheap.
Even in nonlocalpp=yes, we can avoid using determinants by saving just the ratios used for NLPP.
if determinants are also optimized, it is not useful at all.

prckent:
This is a decision to always recompute the wavefunction and gradients and to not buffer constant parts. Given the trend towards greater optimzation and techniques to evaluate derivatives efficiently, I am OK with removing the buffering for the reasons you give. For now at least it is vastly more important that we have good Jastrows and robust optimizers. 

Q. How do varying the availability of usebuffer affect runtime of the optimizer for simple cases that someone new to QMC might encounter? e.g.  one and two body Jastrow only with single Slater determinant, 54 atoms of bulk silicon, or running one-two-three body Jastrow multi determinant small molecule?

ye-luo:
NiO S8 32 atom 384 electron J12:
OneShiftOnly:
```
OSO-buffer-nlpp/opt-NiO-fcc-S8.out-  QMC Execution time = 7.1629e+01 secs 
OSO-buffer/opt-NiO-fcc-S8.out-  QMC Execution time = 6.1625e+01 secs 
OSO-nlpp/opt-NiO-fcc-S8.out-  QMC Execution time = 7.7717e+01 secs 
OSO/opt-NiO-fcc-S8.out-  QMC Execution time = 6.1545e+01 secs 
[...]

PR #470, opened 2017-11-06
Remove opt buffer

ye-luo:
deprecate usebuffer as discussed in #433.


markdewing:
okay to test

markdewing:
Can the DataSetForDerivatives member in Walker.h be removed, or not yet?

ye-luo:
Thanks for the reminder. Updated.

ye-luo:
@jtkrogel Does nexus have usebuffer settings?

jtkrogel:
There is no default use of usebuffer in Nexus that will affect general users.  Users can provide usebuffer to Nexus (just as to QMCPACK), and the user requested value will show up in the generated input file.  

There are defaults for usebuffer in qmcpack_workflows, which essentially only I use at this point, and they should remain for now.

I assume that after this change, usebuffer input will simply be ignored within QMCPACK.  Probably better to issue a non-aborting warning w.r.t. the deprecated parameter.

ye-luo:
I would like to use the general mechanism for detecting unused input param/arguments when we deal with XML input validation.

PR #472, opened 2017-11-07
nexus: guard against save fail on analysis step of workflows

jtkrogel:
Nexus could fail at the analyze step for a QMCPACK simulation in cases where the amount of output data is large (e.g. large charge densities in stat.h5).  The analyzer by current default reads all of this data in, post-processes it, and then stores it for potential later access by the user.  The protocol used for the "store" operation (cPickle) fails for data sizes in excess of 4 GB (yes it is a 32 bit integer overflow error).  

The change in this PR is to not automatically postprocess stat.h5 and traces files when the analyzer is used in a workflow setting.  Direct user interaction will result in the files being read and analyzed.


prckent:
OK to test

PR #473, opened 2017-11-07
Enable HDF5 output for SkAllEstimator

jtkrogel:
This PR enables HDF5 (stat.h5) output for SkAllEstimator.  It also adds the walker weight to the estimator for consistency with other estimators in the context of DMC.

Outputted S(k) values from SkAllEstimator were checked for agreement against StaticStructureFactor and kpoint output was added to the latter class as well.


markdewing:
okay to test

jtkrogel:
The modulus needs to be included.  Will add shortly.

jtkrogel:
Ready for review.

jtkrogel:
I have only checked that VMC values match to high precision between the estimators.  I can run a short DMC if desired.

prckent:
OK. So long as we know that the weight was not being subtly added elsewhere.

prckent:
Do you need @rcclay to review this?

jtkrogel:
As far as I know the weight is not used consistently throughout the code.  Weighted and unweighted values should converge in the large population limit.  This way is more correct though.

I would prefer to have his review since he is the original author of this code, but it may not be strictly necessary.

prckent:
Will wait for @rcclay then.

prckent:
If you know of anything unweighted, please make an issue since it is almost certainly a bug (IMO).

prckent:
Thanks Ray. Merging.

PR #474, opened 2017-11-08
LCAO SoA

ye-luo:
In PR
1, SoA implmenation of LCAO
2, Reduce MSD memory footprint by avoiding per thread copy
3, Change OhmmsVector implementation (already used in miniQMC). operator () is removed. Using [] consistently everywhere.
4, Build OhmmsMatrix on OhmmsVector
5, Remove ParticleBase and build ParticleAttrib on OhmmsVector, reducing code base.
6, Displacement AA table only stores lower triangle, saving memory.
7, Improve J2 and JeeI vectorization
8, Print process memory
9, Change #pragma simd to #pragma omp simd, eliminate warnings.
10, Remove SPOSetBase::t_logpsi. The memory is managed by upper level classes.
11, Support Pade Jastrow with SoA
12, Update enclosed miniapps to support NiO as miniQMC.

TODO:
update the manual.

Known issue:
libmvec from glibc 2.23 is buggy. GNU build will crash on ubuntun 16.04.


prckent:
OK to test

src/QMCWaveFunctions/EinsplineAdoptor.h
```diff
@@ -248,6 +248,11 @@ struct BsplineSet: public SPOSetBase, public SplineAdoptor
     SplineAdoptor::set_spline(spline_r,spline_i,twist,ispline,level);
   }
 
+  inline RealType RATIO(const ParticleSet& P, int iat, const ValueType* restrict arow)
```
markdewing: A method name that is all capital letters seems confusing because it's normally used for pre-processor defines and constants.

CMakeLists.txt
```diff
@@ -864,7 +864,9 @@ SUBDIRS(src)
 IF(NOT(QMC_BUILD_LEVEL GREATER 4))
   SUBDIRS(tests)
   SUBDIRS(examples)
-  SUBDIRS(nexus/examples)
+  IF(IS_DIRECTORY nexus)
```
markdewing: Why is this check needed?  Is the nexus directory not present sometimes?

jngkim:
It is a burden for bare-metal development and is good to skip it if nexus/python is not available.

prckent:
I am seeing of hints of a double free or some sort of corruption problem in tests on oxygen. li2_sto dmc seems to trigger it. I won't be able to look further today.

prckent:
I am seeing of hints of a double free or some sort of corruption problem in tests on oxygen. li2_sto dmc seems to trigger it. I won't be able to look further today.

src/QMCWaveFunctions/Fermion/DiracDeterminantWithBackflow.cpp
```diff
@@ -2,7 +2,7 @@
 // This file is distributed under the University of Illinois/NCSA Open Source License.
 // See LICENSE file in top directory for details.
 //
-// Copyright (c) 2016 Jeongnim Kim and QMCPACK developers.
+// Copyright [c] 2016 Jeongnim Kim and QMCPACK developers.
```
markdewing: Misguided search and replace for () -> []?

markdewing:
Large PR's from long-lived branches get to be very difficult to review.
In the future it would be better if changes that are independent of the feature under development were added to the main QMCPACK repo separately.

In particular,  the change of () -> [], adding std:: prefixes, and  changing int -> size_t are pervasive and dominate the diffs, making it hard to find the more substantial changes.
Would it be possible to split this PR into multiple pieces? From a quick look at the diffs, many of the files contain only changes listed above - if those could be placed into their own commits/PR, it would make the remaining changes easier to review.

prckent:
I agree with Mark.

prckent:
Please also indicate which tests are still expected to fail after these updates. Should the orbital optimization work?

CMakeLists.txt
ye-luo: Yes. I do not always checkout it.

ye-luo:
() -> [] is due to the replacement of low level OhmmsVector
std:: changes are probably due to fixing failing builds.
int -> size_t is processed during MSD memory saving changes.
Unfortunately, they get entangled especially when I was doing a lot of development during weekend.
Sorry about the heavy review.
This PR is already a preparation of a follow-up.

ye-luo:
Orbital rotation determinants doesn't need change. It should work out of box. SPO need to be rerouted in the new SoA builder for LCAO.
Backflow is still broken. [Here](https://cdash.qmcpack.org/CDash/viewTest.php?onlyfailed&buildid=14183) is what fails with this PR

prckent:
@ye-luo Would you be able to rebuild this PR in a few pieces that are more easily reviewable?

Important to keep separate developments separate.

ye-luo:
@prckent It is time consuming and hard to keep accuracy if I rewrite the history. For example, when I changed the code at low level, all the tests pass. After a few months, more tests are added and bugs are exposed and fixes come in. Then the development is entangled. Rewriting the history also breaks the consistency with other development built on top of this branch.

I tried to review it myself and I went it through once. It is long but a lot of changes can just be scanned through without much attention needed.
files under QMCWavefunctions/lcao are probably used for the first time, only one file is used by hybrid already.
files under OhmmsPETE are replacement and simplification.
Removing ParticleBase is a critical change and files under Particle and ParticleBase are touched.
The Jastrow changes affects the numerics, all the reset should be technical changes.

jngkim:
Ye consulted me before making any change and I approved and reviewed code
changes. Vice versa.

lcao directory is *added* just like SoA classes to coexist with the
original implementation. Numerical issues are very carefully tested. I
don't know anyone else knows what is in MO implementations and who can
review the code.

markdewing:
Large PR's from a long-lived feature branch have a number of problems, both during development and at PR time.  Do @ye-luo or @jngkim (or anyone else?) have ideas on how to mitigate these problems in the future?

I would really like move the development process more towards thinking about how to make PR's as small or simple as possible (while still retaining a unity of purpose - some might affect many files, but still be conceptually simple, like the ()->[] change).   This makes the changes easier to reason about and review, easier to track down breaking changes, and makes sections of the code more testable.

My personal preference would be feature branches be treated as prototypes, and then rewritten into a more organized fashion for the actual PR (or series of PR's). (Graphical diff/merge tools can make this much easier.)  However, I realize this might be too time-intensive and slow down development too much.

jngkim:
I don't have any idea or plan  for PR and leave it to Ye and you guys.

ye-luo:
A few months is a typical life time a sophisticated feature branch. There are two ways during the development. 1, ignore the changes on the develop and just focus on the branch. 2, merge frequently.

1, will allow you to rewrite and reorganize changes.
2, terrible to reorganize commits due to the entanglements.
I'm doing the second way because I do as much as possible (code cleaning and adding tests) on the develop and then merge to my feature branch. However, I can't do everything on the develop but a feature branch is not ready to merge to develop. Then a feature branch start to accumulate things.

I understand your concern. I think what I can do in the future is trying to PR as early as possible and delay changes like ()->[] to the develop.

prckent:
@ye-luo 

Are you planning any more updates to this PR?

ye-luo:
@prckent resolved the conflict and updated the manual

prckent:
General comment - please split up future developments as per https://github.com/QMCPACK/qmcpack/wiki/Development-workflow and my email to the ECP developers.

PR #475, opened 2017-11-08
nexus: fix k-point add for tiling corner case

jtkrogel:
When a structure object was provided directly to generate_structure via generate_physical_system, the k-point grid was erroneously added prior to tiling.  This PR enforces the expected behavior: the k-point grid is first added to the supercell and and then the k-point grid is tiled for the primitive cell.


prckent:
OK to test

PR #478, opened 2017-11-08
updates to get real/complex X full/mixed CI on rhea

grahamlopez:
This PR will enable the following builds/tests on Rhea, for each of CPU and GPU builds:
- full real
- full complex
- mixed real
- mixed complex

As a result, I am also disabling the previous 'qmcpack-rhea-cpu-mixed' Jenkins robot, and the two existing 'qmcpack-rhea-[cpu,gpu]' robots will each handle all four variants.


tests/test_automation/jenkins_rhea_cpu.sh
```diff
@@ -43,7 +44,32 @@ fi
 make -j 24
 
 ctest -L unit
-#ctest -R short-LiH_dimer_ae-vmc_hf_noj-16-1
+
+
+echo "starting new test for real mixed precision"
+rm -rf ./*
```
ye-luo: This typically dangerous. Can we create build_XXX folders to hold different builds?

tests/test_automation/jenkins_rhea_cpu.sh
```diff
@@ -30,7 +30,7 @@ mkdir -p build
 
 cd build
 
-# real, full precision
+echo "starting new test for real full precision"
```
ye-luo: Could you please add a few empty lines by echo if there is no blank lines between this message and the previous tests?

tests/test_automation/jenkins_rhea_cpu.sh
grahamlopez: What are you worried about here?

tests/test_automation/jenkins_rhea_cpu.sh
grahamlopez: There are blank lines after the final unit tests run.

tests/test_automation/jenkins_rhea_cpu.sh
grahamlopez: Note that the CI builds have never persisted, so creating new folders and then deleting them is the same as the current strategy.

grahamlopez: We do need to stay cognizant of disk issues, so that we don't run up against quotas, VM issues, etc. This will only get worse as time goes on and requirements expand. For troubleshooting, a human needs to manually build and troubleshoot the offending patches to post a fix for them anyway.
As for `rm -rf ./*`, it is indeed a dangerous tool that can burn the user. But similarly to lighting a fire, there are times when it is the best way to accomplish the goal. I don't think that keeping old CI builds around is a winning strategy, and so I need a way to clear out this directory. I'm open to other suggestions.

jtkrogel: What about adding a "cd .." followed by "rm -rf ./specific_directory".  This should greatly narrow the possible range of undesirable outcomes.

ye-luo:
@grahamlopez Thank you a lot for addressing my requests.

PR #482, opened 2017-11-09
Add MP guards and SoA warning.

ye-luo:
CPU MP is forbidden with multi determinants
CPU MP is forbidden with WF optimization.
Though the above two cases pass tests, they are not safe to use in general at the moment.

Warning is issued for Chiesa correction with SoA build.


prckent:
How do you reconcile the use of mixed precision in the GPU code (default, long used) and disabling mixed precision optimization?

ye-luo:
QMC_MIXED_PRECISION is for CMake only. MIXED_PRECISION is the internal flag for CPU. GPU code always have FULL_PRECISION on the host.

prckent:
Q. Should the KECorr warning be an error? If we only print a warning, will people see it? ( #35 )

prckent:
OK to test

jtkrogel:
IMO if a user requests a feature in the input that works in at least some cases, then the code should abort with an error if the feature is not supported with the used code configuration (e.g. SoA).  

It might also be a good idea for the error message to include a link to the github issues page along with instructions/encouragement to submit an issue if the missing feature is needed/useful.

prckent:
@jtkrogel This is a valid point. @ye-luo Please can you add aborts to the KECorr in the canonical jastrows where the SoA implementation is missing?

ye-luo:
I don't feel necessary to abort the code for KECorr. Actually, l first implemented aborting and then changed it to print just a warning. Unlike other estimators, KECorr is computed at the initialization instead of sampling. For the moment rerunning AoS for 1 block will give you the value. I prefer to keep the same input file for SoA and AoS. If the code aborts, I must prepare two input files, one with and one without. I have many old input files and I don't want to change them to run SoA.

prckent:
Kayahan was caught out by this, and I am confident that if I was running the code I would miss the warning.

ye-luo:
When a user runs analysis, he/she immediately notice it is zero in the statistics and the next step he should do is read the error file and output file and then understand the situation. I can reroute the message to error file if needed. It is a user responsibility to read the output files if he/she feels something is strange.

prckent:
If we get a response from Kayahan at https://groups.google.com/forum/#!topic/qmcpack/2bYIQ6shTMs we can consider modifying the current KECorr handling in a later PR.

Will merge once the CI has passed.

rcclay:
For the record, we should really remove KECorr.  It's formally incorrect without a long-range jastrow.

jtkrogel:
We should do a close comparison sometime between the performance of KECorr+MPC, recent Holzmann corrections (when finished), KZK functional and MZK functional all with the use of short or long ranged Jastrows (which need to be re-enabled).

Issue #409, opened 2017-10-09
Need to break up tests/system/CMakeLists.txt


ye-luo:
tests/system/CMakeLists.txt is getting too big.
Solid, Molecule, HEG should generate tests in corresponding folder.
We already have tests/heg, tests/solid, tests/molecules.

PR #483, opened 2017-11-09
Break up tests/system/CMakeLists.txt

markdewing:
The tests/system/CMakeLists.txt file is getting large and it's time to break it up
This change removes that file, and moves the system tests to CMakeLists.txt files in each test directory.
Addresses #409

The test output is now in individual directories in the build directories( eg, build/tests/molecules/H4_orb_opt_dmc/) , rather than all gathered in tests/system. Is this change okay?

The checks for QMC_COMPLEX and QMC_CUDA occur in every individual test CMakeFiles.txt, and there is no message about excluded tests for gaussian basis tests.   Maybe there should be a message in each file?  Alternately, the check for QMC_COMPLEX and QMC_CUDA could move to the tests/CMakeLists.txt file, and issue the message there.

This is an initial PR for feedback, don't commit yet.   The manual needs to be updated to match, and the changes from #482 need to be incorporated (I can do that after that PR is merged).


ye-luo:
Probably we can create CMakeLists.txt under molecules and solid for better hierarchy and precise control.

prckent:
I prefer to test output to be in the new proposed location e.g. build/tests/molecules/H4_orb_opt_dmc/ . This is more logical that the current situation.

ye-luo:
build/tests/molecules/H4_orb_opt_dmc is good. I just mean add another layer of CMakeLists.txt in molecules/solids instead of having a long list in tests/CMakeLists.txt

prckent:
If each leaf cmakelist prints a message to say why it is not adding a test (e.g. cuda, complex), would that add greatly to our cmake output? Provided the output volume is not too different I think it would be an improvement since we can add a future requirement that each set of tests should declare if and why it is being skipped, making maintenance and future contributions easier.

prckent:
I agree with the molecules/solids break.

Mark: please could you update the cmake configuration and also update the manual consistently as you noted ?

ye-luo:
@markdewing Can we move QMC_COMPLEX and QMC_CUDA checking to molecule/solid cmakelists.txt and print the warning?

markdewing:
In some cases that will not work (solids/bccH_1x1x1_ae).  For consistency, I would like to keep the tests and messages at the same level  and not have some in solids/CMakeLists.txt , and some in the lower level CMakeListst.txt

ye-luo:
@markdewing fair enough.

PR #485, opened 2017-11-09
Fdlr wave function

nsblunt:
Dear All,

Here is the code for the finite-difference linear response wave function class. I have cherry-picked the appropriate commits on to a branch off the latest develop branch (with opt_det_new merged in). This branch relies on the optimizable determinants. Because of the changes there, the cherry-picks were non-trivial, and optimizing the wave function parameters doesn't seem to work currently. So it needs more work, but I thought I could create this PR to get the ball rolling (in case it takes a while to review)?

The theory of the approach is described in our paper (should be online in JCP shortly):

https://arxiv.org/abs/1707.09439

This idea is that a CIS wave function can be written as a linear combination of first derivatives of a determinant with orbital rotations, and these first derivatives can be approximated with a finite-difference approach - the difference of two determinants at slightly different rotation values. However, the idea (and the code) is more general and can be used for the linear response of other WF ansatz (although not really tested).

In the code the FDLR wave function is an OrbitalBase child, and the two wave functions that form the FD approximation (within the FDLR object) are TrialWaveFunction objects. This might seem strange, but means that the WFs in the finite-difference can be any product of OrbitalBase objects, so it should be general. Doing this also simplified several things, i.e., input of the two wave functions is nearly trivial, exactly the same as existing wave functions by XML. I did put some thought into this and decided it was the simplest and most general way, I hope this is OK.

Let me know if you have questions about any of this. There is an example set of input files in tests/molecules/H4_FDLR_CIS_opt. I haven't added benchmarks yet because it's still buggy, but the idea is there and VMC energies should be correct, and the job completes.

Best,
Nick


markdewing:
okay to test

nsblunt:
The bug is just since the changes to the optimizable determinants from the last PR. FDLR depends on optimizable determinants and seems to need updating, as optimization of the wave function no longer reproduces old results. Though it's not obvious to me why it's not working now, so this needs investigating.

I'll add a manual entry.

nsblunt:
I think this is fixed now. The problem was actually in acceptMove for the standard SlaterDet class. This doesn't update its own LogValue variable. This then means that when TrialWaveFunction::acceptMove tries to update LogValue, it does so incorrectly. I assume that for other wave functions this doesn't cause a problem since it isn't used, but for FDLR this updated LogValue is used in evalGrad, so was causing errors that weren't there before. See the recent commit to check if you are happy with the change.

Needs a bit more testings still but certainly ready to be reviewed when some one has time. There is an example in the tests directory.

ye-luo:
@nsblunt LogValue is not used during PbyP move. No need to update the value. It is only updated when all the single particle moves are settled. Is your finite difference code doing additional fake single electron moves?

nsblunt:
There are no additional fake single electron moves. But due to its form as a difference of two wave functions, it turns out that the values of those two underlying wave functions are needed to calculate the gradient. I can go through the math if required or helpful.

ye-luo:
I see. You need up-to-date LogValue during PbyP moves.

nsblunt:
I've added two tests, and added a manual entry for FDLR. After trying quickly, I couldn't compile the manual so haven't checked what the entry looks like, and that all the tex is correct. Would you mind checking this, or letting me know if there's a simple way for me to compile the manual? Feel free to make any changes you like to the way I've formatted it, it might be quicker if you know what you want.

I think I've made all changes requested by @prckent now, and fixed the bug. Any other changes required?

prckent:
I linked the chapter and made some LaTeX fixes to get it to compile.

I am happy to merge this when the CI has passed.  @ye-luo , @markdewing : OK ?

tests/molecules/H2_FDLR_opt/H2.info.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
```
markdewing: This is an output file, it should not be checked-in (unless output is included for reference)

src/QMCWaveFunctions/TrialWaveFunction.h
```diff
@@ -124,6 +124,9 @@ class TrialWaveFunction: public MPIObjectBase
   ///differential laplacians
   ParticleSet::ParticleLaplacian_t L;
 
+  ///the size of gradient component (QMCTraits::DIM)*the number of particles
+  int TotalDim;
```
ye-luo: Please move these three lines back to its original place and replace all your use of TotalDim with G.size()*DIM. TotalDim is not safe to use and to be removed.

ye-luo:
Usually we put all types of tests of the same system in one folder. I just feel too many folders at the moment.
Is it possible to merge 
H2_FDLR_opt  H2_orb_opt in to H2_ae
also
H4_ae  H4_FDLR_CIS_opt  H4_orb_opt  H4_orb_opt_dmc into H4_ae

nsblunt:
Apologies for the errors. Yes I will merge those tests into the existing folders. And yes my last commit only fixed the non-complex builds, the complex ones seem a little more tricky, but as you say probably correcting some types. I'll try and fix this tomorrow.

nsblunt:
I could do with some help regarding which QMCPACK types I should be using. I just want to multiply a scalar with a Laplacian object, i.e. in the line:

L += scaling_fac_1*L_temp_1 + ...

L_temp_1 is of type ParticleSet::ParticleLaplacian_t, which I think it should be. If I set scaling_fac_1 to be either RealType or ValueType then I get a compiler error about the * operator not be supported for these pairs of types (for a complex build). Sorry if this is a basic question, what is the intended QMCPACK type for scaling_fac_1 in this operation?

ye-luo:
@nsblunt Could you paste the whole error message? Could you also try 
`L += L_temp_1 * scaling_fac_1 + ...`

nsblunt:
Swapping the order didn't correct it unfortunately. The full error is very long, but starts:

/home/nsb37/qmcpack_public/src/QMCWaveFunctions/FDLRWfn.h:1019:14:   required from here
/home/nsb37/qmcpack_public/src/PETE/OperatorTags.h:334:15: error: no match for ‘operator*’ (operand types are ‘const std::complex<double>’ and ‘const float’)
In file included from /home/nsb37/qmcpack_public/src/OhmmsPETE/TinyVector.h:43:0,
                 from /home/nsb37/qmcpack_public/src/Configuration.h:25,
                 from /home/nsb37/qmcpack_public/src/Particle/VirtualParticleSet.h:20,
                 from /home/nsb37/qmcpack_public/src/QMCWaveFunctions/TrialWaveFunction.h:26,
                 from /home/nsb37/qmcpack_public/src/QMCWaveFunctions/WaveFunctionFactory.h:22,
                 from /home/nsb37/qmcpack_public/src/QMCWaveFunctions/WaveFunctionFactory.cpp:22:
/home/nsb37/qmcpack_public/src/OhmmsPETE/TinyVector.h:254:40: note: candidate: template<class T1, class T2, unsigned int D> typename qmcplusplus::BinaryReturn<qmcplusplus::TinyVector<T1, D>, qmcplusplus::TinyVector<T2, D>, qmcplusplus::OpMultiply>::Type_t qmcplusplus::operator*(const qmcplusplus::TinyVector<T1, D>&, const qmcplusplus::TinyVector<T2, D>&)
 OHMMS_META_BINARY_OPERATORS(TinyVector,operator*,OpMultiply)
                                        ^
/home/nsb37/qmcpack_public/src/OhmmsPETE/OhmmsTinyMeta.h:74:1: note: in definition of macro ‘OHMMS_META_BINARY_OPERATORS’
 FUNC( const TENT<T1,D>& v1, const TENT<T2,D>& v2 )                            \
 ^
/home/nsb37/qmcpack_public/src/OhmmsPETE/TinyVector.h:254:40: note:   template argument deduction/substitution failed:
 OHMMS_META_BINARY_OPERATORS(TinyVector,operator*,OpMultiply)
                                        ^
/home/nsb37/qmcpack_public/src/OhmmsPETE/OhmmsTinyMeta.h:74:1: note: in definition of macro ‘OHMMS_META_BINARY_OPERATORS’
 FUNC( const TENT<T1,D>& v1, const TENT<T2,D>& v2 )                            \

In file included from /home/nsb37/qmcpack_public/src/PETE/PETE.h:66:0,
                 from /home/nsb37/qmcpack_public/src/OhmmsPETE/TinyVector.h:42,
                 from /home/nsb37/qmcpack_public/src/Configuration.h:25,
                 from /home/nsb37/qmcpack_public/src/Particle/VirtualParticleSet.h:20,
                 from /home/nsb37/qmcpack_public/src/QMCWaveFunctions/TrialWaveFunction.h:26,
                 from /home/nsb37/qmcpack_public/src/QMCWaveFunctions/WaveFunctionFactory.h:22,
                 from /home/nsb37/qmcpack_public/src/QMCWaveFunctions/WaveFunctionFactory.cpp:22:
/home/nsb37/qmcpack_public/src/PETE/OperatorTags.h:334:15: note:   ‘const std::complex<double>’ is not derived from ‘const qmcplusplus::TinyVector<T1, D>’
     return (a * b);

ye-luo:
@nsblunt This happened with mixed precision complex build, right? Does the full precision complex build go through? What compiler do you use?

nsblunt:
I just tried full precision complex, it still fails to build but with a  different set of errors. The one I just mentioned isn’t there.

The compiler is GCC 5.4.0.

nsblunt:
The difficulties seem to be because FDLRWfn is an OrbitalBase object, which expect to work with ValueType, but depends on two TrialWaveFunctions, which already convert everything into RealType. I could convert everything carefully for the complex case, but there are a very large number of statements to deal with and FDLR isn't intended for use with complex numbers. Is it OK to just guard it with QMC_COMLEX (and an appropriate abort statement if used with QMC_COMPLEX) as you suggested before?

prckent:
I agree with guarding against complex wavefunctions and giving an appropriate abort.

ye-luo:
@nsblunt Sure, let us do the pragmatic way.

src/QMCWaveFunctions/FDLRWfn.h
```diff
@@ -12,6 +12,8 @@
 #ifndef QMCPLUSPLUS_FDLRWFN_H
 #define QMCPLUSPLUS_FDLRWFN_H
 
+#ifndef QMC_COMPLEX
+
```
ye-luo: Please do not guard complex here but at the place where this file is included.

nsblunt:
Sure, apologies for that. Is it okay to remove the last commit and force push or will that cause problems with GitHub/bad practice?

ye-luo:
You can just amend your last commit.
git add your-fix
git ci --amend
git push with --force
It should not be a problem for the PR. I did it before.

prckent:
You can simply commit and push again. Should be no need to force or use any tricks.

nsblunt:
@ye-luo Thanks. I just have a few bits of tidying I wanted to add before merging. I'll push something now and again in the morning if that's okay.

ye-luo:
@nsblunt I guess you may also need to guard your cpp file in CMakeList.txt for complex.
I actually like splitting since it will benefit parallel compilation.

Let us know when it is ready to merge.

nsblunt:
I think this is ready now. OK to merge?

prckent:
Thanks Nick.

I am rerunning all the tests with freshly built real and complex codes and will merge once they have finished successfully.

prckent:
Thanks again for all the effort on this.

PR #487, opened 2017-11-10
Skip Nexus example tests if numpy is not present

### LINKED ISSUE #484: NEXUS example tests run and fail when numpy is not available
Numpy availability test or bypass needed  

```
    Start 274: nexus_example_c20

274: Test command: /opt/local/bin/python "nexus_example_test.py" "/Users/ .... /c20" "--nexus-file" "c20.py" "--nexus-path" "/Users/ .... /examples/../library"
274: Test timeout computed to be: 240
274: ('Error ', <type 'exceptions.ImportError'>, ImportError('No module named numpy',))
274:   FAIL
1/8 Test #274: nexus_example_c20 .....................***Failed    0.08 sec
```

markdewing:
Fixes #484

### DIFF
--- a/CMake/python.cmake
+++ b/CMake/python.cmake
@@ -15,13 +15,16 @@ ENDFUNCTION()
 # Test python module prerequisites for a particular test script
 #   module_list - input - list of module names
 #   test_name - input - name of test (used for missing module message)
+#                     - use empty string to silence output
 #   add_test - output - true if all modules are present, false otherwise
 FUNCTION(CHECK_PYTHON_REQS module_list test_name add_test)
   set(${add_test} true PARENT_SCOPE)
   foreach(python_module IN LISTS ${module_list})
     TEST_PYTHON_MODULE(${python_module} has_python_module)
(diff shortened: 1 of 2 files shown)


jtkrogel:
This looks good to check for numpy.  Do we need a check for presence of python in general?  I didn't see one, but may have missed it.

markdewing:
Are there systems with no python at all?
If that is an issue we could add a check for python.

jtkrogel:
At least some developers (e.g. Jeongnim) do develop on systems w/o Python and sometimes remove nexus from the repo as a result.  It is conceivable that other users will run into this as well so a check/guard is probably warranted.

markdewing:
PR #474 adds a check for the presence of the nexus directory, which should take care of some of it.
Testing the CMake configuration with the python executable temporarily removed, the checks seemed to behave the same as if all the modules were not found.  So it doesn't seem that a separate check for python itself is needed.

jtkrogel:
OK, sounds good. +1

PR #488, opened 2017-11-10
Extract walker assignment code from walker swaps.

markdewing:
The walker swapping code first determines the new number of walkers each
processor should have, what exchanges are needed to get there, and then
communicates the walkers to be exchanged.
This change extracts the first two parts into a separate function, which
separates it from the MPI calls and makes it easier to test.

This commit also introduces a property-based test for this function, but
it is off by default.

This *should* be purely refactoring the code, with no change in functionality.  It looks like the diff algorithm is really mangling the change - it should be moving code from swapWalkersSimple to determineNewWalkers.


markdewing:
Changed to a fixed seed.  Eventually there should be some infrastructure to manage the random inputs, so every test doesn't need to handle that itself.

prckent:
Thanks Mark.

PR #489, opened 2017-11-10
Improve Bspline Jastrow test.

markdewing:
Compare with a series of values obtained from a python script (gen_bspline_jastrow.py)
The script generates the comparsion using the B-splines built into Sympy.
The first and second derivatives are also tested now.

The testing is not as extensive as code generating the expression to compare against, but is better than what was there.


prckent:
I have nominated Jaron to review this, since he is more pythonic than I am. However I am happy to see it. sympy + bspline basis + autodiff = excellent.

markdewing:
Some of the code was taken from this Jupyter notebook
https://github.com/markdewing/qmc_algorithms/blob/master/Wavefunctions/Explain_Bspline.ipynb

jtkrogel:
Looks OK to me.  +1

Issue #491, opened 2017-11-13
Intel 2017 fails to build SoA version of QMCPACK


prckent:
Intel 2017  version 17.0.1.20161005 fails with the error given below. This is "update 1" to the 2017 compiler. @ye-luo has said that the "update 2" of 2017 is OK. I have verified that 2018 is OK (e.g. this is how I tested the LCAO SoA PR), however it is not very widely deployed.  I verified today that 2016.3 is not OK. 

Unfortunately the error message looks like we could be to blame.

I suggest we either error trap the <=2017.0.1... compiler versions on configure,  i.e. Set a minimum Intel compiler version when SoA is enabled, or build a workaround if possible and small.  Lets decide today. Folks with Intel compiler bug/feature experience, please comment.

```
/home/pk7/projects/qmc/git_QMCPACK_qmcpack/qmcpack/src/QMCWaveFunctions/Jastrow/J2OrbitalSoA.h(466): error: no matching declaration for this reduction
      #pragma omp simd reduction(+:cur_g) aligned(old_dX,new_dX,save_g,cur_du_pt,old_du_pt)
                                   ^
          detected during:
            instantiation of "void qmcplusplus::J2OrbitalSoA<FT>::acceptMove(qmcplusplus::ParticleSet &, int) [with FT=qmcplusplus::PadeFunctor<qmcplusplus::OrbitalBuilderBase::RealType={qmcplusplus::TrialWaveFunction::RealType={qmcplusplus::QMCTraits::RealType={double}}}>]" at line 255
            instantiation of "qmcplusplus::J2OrbitalSoA<FT>::J2OrbitalSoA(qmcplusplus::ParticleSet &, int) [with FT=qmcplusplus::PadeFunctor<qmcplusplus::OrbitalBuilderBase::RealType={qmcplusplus::TrialWaveFunction::RealType={qmcplusplus::QMCTraits::RealType={double}}}>]" at line 64 of "/home/pk7/projects/qmc/git_QMCPACK_qmcpack/qmcpack/src/QMCWaveFunctions/Jastrow/PadeJastrowBuilder.cpp"

compilation aborted for /home/pk7/projects/qmc/git_QMCPACK_qmcpack/qmcpack/src/QMCWaveFunctions/Jastrow/PadeJastrowBuilder.cpp (code 2)
```

ye-luo:
It is not our fault. Just the compiler didn't know how to handle reference type.
I think building a workaround is more useful. I can do it quickly.

[...]

PR #492, opened 2017-11-13
Workaround issue with Intel compiler <= 17u1.

ye-luo:
addresses  #491


markdewing:
okay to test

prckent:
@markdewing Please can you check this? Do you have access to a suitably old compiler?

markdewing:
I'll check - I'm sure there are systems around with the different compiler versions

PR #493, opened 2017-11-13
Change walker offset IO handling.

ye-luo:
Hit some issue on Mira when using >= 32768 ranks.

walker offsets (an array) of all the nodes are recorded in the config.h5 for restart.
It was written by every node to the file even though the file is open in collective IO mode.
It is not a problem when the node count was small (<=16384). With a large node count, hdf5 library crashed inside.
Changes to the offset IO:
write: collectively by every node, each contribute its own offset. The last node contributes 2 (for the end)
read: since the next run can be with a different node count, offset now is loaded by the master node and broadcast to all the nodes.
performance:
Both IO operations are measured faster.

Restart miniapp has passed all the checks with 65536 ranks.
restart tests also pass.

Wait for a big run (in a few days) on Mira before merge.

### DIFF
--- a/src/Particle/HDFWalkerInput_0_4.cpp
+++ b/src/Particle/HDFWalkerInput_0_4.cpp
@@ -293,44 +293,68 @@ bool HDFWalkerInput_0_4::read_phdf5( std::string h5name)
 
   int nw_in=0;
   h5name.append(hdf::config_ext);
+  std::vector<int> woffsets;
+  int woffsets_size;
+  bool success;
+
+  { // handle small dataset with master rank
+    hdf_archive hin(myComm,false); //everone reads this
+    if(!myComm->rank())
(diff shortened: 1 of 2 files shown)


markdewing:
okay to test

ye-luo:
Updated the PR comments and pasted the story into the commit log. Also updated in-source comments.

prckent:
@ye-luo Will merge when you give the OK based on your test run (ECP ALCC time hopefully)

ye-luo:
My jobs are near the top of the queue. I will comment when it passes through.

ye-luo:
My big run confirms that the IO library crash is resolved. This PR is safe to merge.
Sadly, the restart still have some other issue at scale. Will PR when the issue is resolved.

PR #495, opened 2017-11-13
Improve optimization text

prckent:
Slightly more detailed guidance. Remove brick wall reference.

### DIFF
--- a/manual/opt.tex
+++ b/manual/opt.tex
@@ -1,4 +1,6 @@
 \section{Wavefunction Optimization}
+\label{sec:optimization}
+
 Optimizing wavefunction is critical in all kinds of real-space quantum Monte Carlo calculations
 because it significantly improves both the accuracy and efficiency of computation.
 However, it is very difficult to directly adopt deterministic minimization approaches due to the stochastic nature of evaluating quantities with Monte Carlo.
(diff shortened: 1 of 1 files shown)


manual/opt.tex
```diff
@@ -386,15 +388,14 @@ \subsubsection{quartic}
...
-\item One-body spline Jastrow from old calculations can be good starting point.
-\item Three-body polynomial Jastrow should start from zero.
+\item All electron wavefunctions are typically more difficult to optimize than pseudopotential ones due to the importance of the wavefunction near the nucleus.
+\item Two body Jastrow contributes the largest portion of correlation energy from bare Slater determinants. It's also the easiest part for optimizer and gains most. For this reason, the recommend order optimizing wavefunction components is two-body, two-body, three-body Jastrow factors and MSD coefficients.
+\item For two-body spline Jastrows, always start from a reasonable one. The lack of physically-motivated constraints in the functional form at large distances can cause slow convergence if starting from zero. 
+\item One-body spline Jastrow from old calculations can be a good starting point.
+\item Three-body polynomial Jastrow can start from zero.
+\item When adding higher-body terms, it can help to perform an initial optimization with the lower body terms held fixed. e.g. Hold the one and two body fixed when initially adding three-body terms.
```
ye-luo: No need to fix the one and two body Jastrow. 3B is very easy to optimize.

prckent: I prefer to have the safer text in the manual.

manual/opt.tex
```diff
@@ -386,15 +388,14 @@ \subsubsection{quartic}
...
-\item Two body Jastrow contributes the largest portion of correlation energy from bare Slater determinants. It's also the easiest part for optimizer and gains most. For this reason, the recommend order optimizing wavefunction components is two-body, one-body, three-body Jastrow factors and MSD coefficients.
-\item Never use all-zero two-body spline Jastrow and always start from a reasonable one. 
-      You might want to bang your head against a brick wall for a bad two-body Jastrow.
-\item One-body spline Jastrow from old calculations can be good starting point.
-\item Three-body polynomial Jastrow should start from zero.
+\item All electron wavefunctions are typically more difficult to optimize than pseudopotential ones due to the importance of the wavefunction near the nucleus.
+\item Two body Jastrow contributes the largest portion of correlation energy from bare Slater determinants. It's also the easiest part for optimizer and gains most. For this reason, the recommend order optimizing wavefunction components is two-body, two-body, three-body Jastrow factors and MSD coefficients.
```
markdewing: easiest part for the optimizer and gains the most

recommend -> recommended

two-body is repeated

manual/opt.tex
ye-luo: When J1 and J2 is already optimized, adding J3 and optimizing is very easy. Usually 3-5 iterations will converge. It is very safe and robust. Adding an extra step of J1&2 fixed with J3 only complicates the workflow and takes extra time.

prckent: This is not published anywhere, particularly for d electron systems. I think the text is fine - "can help".

ye-luo: J3 is well tested with d electrons. Most of my experience is with d electrons. Ni, Ti, Cu.
It doesn't help. If one find difficulty optimizing Jastrows, it is typically a problem in VMC or SPOs. Trying to optimize with this suggestion is in vain.

PR #496, opened 2017-11-13
Modify relative path to derivative check scripts

markdewing:
For the bccH_2x2x2_ae test, the check scripts are given relative to bccH_3x3x3_ae.
Separating the test files changed the relative layout of these directories. Change
it to match the new layout.

Also fix the test skipping message in bccH_3x3x3_ae to match the test.


prckent:
Not merging on Ye's request.

Issue #379, opened 2017-09-14
Make MaxCPUSecs available for all the QMC drivers and stop consistently.


ye-luo:
From my earlier comment in PR #377
1. Make MaxCPUSecs a feature also For VMC, OPT, RMC.
2. Define the stop control flag false by default and do MPI reduce at every block to stop all the MPI within this calculation(not all, in ensemble job), For each block, I/O is anyway synchronized.

prckent:
I think that this will be very useful when implemented widely.

We need at least:

The various VMC drivers
The newer optimizers (all would be ideal)
Reptation MC

[ I look forward to when we have ~1 VMC driver etc. ]

markdewing:
Questions
 - The current default is 100 hours.  Should that be changed to unlimited?  I'm concerned about someone hitting it by accident, especially now that it will be present in more drivers?  It seems to be much larger than normal queue limits, so someone running that long is likely intending to a very long run.    We could change the default to -1, and that would mean unlimited.
 - Should there be an error message if a cpu limit is specified for a driver, but not implemented? (I'm not sure I can get it in all the drivers)
 - It's going to be harder to implement in the optimizers - the "loop" element is handled by executeLoop in QMCMain.  Either the time limit needs to be added to the loop element, or the drivers need to set a flag that they stopped due to exceeding a time limit, and executeLoop would check for that flag.  And more - if the sample collection is stopped due to the time limit, those samples should not be used for a next optimization step (I haven't looked at the code to see how to do this).   Is a time limit on optimization valuable enough for the extra implementation work?

ye-luo:
For the last question,
1) Any time limit present in the driver section should refer to the whole run instead of a section. So I don't feel a loop matters.
2) sample collection can handle cases when the actually number of collected samples is smaller than the request and only the collected ones are used for OPT.

prckent:
Fixed by Mark in #497 except for optimization cases.

ye-luo:
Optimization drives VMC internally, hence implicitly inherit the limit.

ye-luo:
One more question. When I have multiple sections and the execution of the second last section is stopped due to time limit, will the code proper skip the last section?

markdewing:
With the current fix, the execution of each section is independent - so if the second to last section stops due to the time limit, the last section will start executing, and then stop after one block because time has expired.
I expect the change to handle the optimization case will work on this case as well (need to propagate the   'stop' message from the drivers to the code that is executing the drivers)

[...]

PR #497, opened 2017-11-14
Extend CPU limit to new drivers and add tests

markdewing:
Add CPU time limit code to VMC CPU and CUDA, DMC CUDA, RMC.
Also add tests.  The time limit is set to 60 seconds, and the corresponding test configuration has a time-out set for 120 seconds.
The test names start with 'cpu_limit'.  They don't seem important enough to include with the 'short' tests.

Partially addresses #379


ye-luo:
Are the tests sensitive to machines? I mean run on a dual core node and 16 core Xeon.

markdewing:
I shouldn't think so.  The test only checks the output for the 'time limit reached' message.
The number of blocks should be large enough so that no current machine should finish before the time limit is reached.   There might be an issue with machines that are too slow to even get through the first block before the test times out.

PR #499, opened 2017-11-15
Fix WalkerOffset bug due to code extraction change.

ye-luo:
Fix bug introduced in PR #488.
FairOffset  is used by the caller of determineNewWalkerPopulation.
So FairOffset needs to be returned from determineNewWalkerPopulation

### DIFF
--- a/src/QMCDrivers/DMC/WalkerControlMPI.cpp
+++ b/src/QMCDrivers/DMC/WalkerControlMPI.cpp
@@ -148,16 +148,16 @@ WalkerControlMPI::branch(int iter, MCWalkerConfiguration& W, RealType trigger)
 }
 
 // determine new walker population on each node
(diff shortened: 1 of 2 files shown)


markdewing:
okay to test

markdewing:
What is the bug?  I think i figured it out from looking at the code, but the commit or PR message should say what it is.

ye-luo:
Updated the PR message. I didn't check the unit test since I was debugging on BGQ. Expect CI to do the check.

markdewing:
FairOffSet is not used by the immediate caller of determineNewWalkerPopulation (swapWalkersSimple).  It's used by "branch" (the caller of swapWalkersSimple).  More importantly, FairOffSet is a member variable, and so not listed as a parameter to these functions.  This is also the reason why a naming convention is often used to signify member variables (prefix with m_ or suffix with _) - to make it more obvious when they are referenced.

PR #505, opened 2017-11-16
fix test FeCO6 test

anbenali:
Fix Gamess output that decided to not Write some of the orbitals making the output wrong and feeding garbage to the converter


prckent:
OK to test

prckent:
@anbenali Please update your branch next PR. This will save a step in the CI.

PR #506, opened 2017-11-17
Refine timers.

ye-luo:
Add more in ParticleSet.
Remove ratio timer in evaluateLog.
Change JeeI timer name from eeI to JeeI.

### DIFF
--- a/src/Particle/ParticleSet.cpp
+++ b/src/Particle/ParticleSet.cpp
@@ -43,10 +43,12 @@ template<> int ParticleSet::Walker_t::cuda_DataSize = 0;
 
 void add_p_timer(std::vector<NewTimer*>& timers)
 {
-  timers.push_back(new NewTimer("ParticleSet::makeMove",timer_level_fine)); //timer for MC, ratio etc
-  timers.push_back(new NewTimer("ParticleSet::makeMoveOnSphere",timer_level_fine)); //timer for the walker loop
+  timers.push_back(new NewTimer("ParticleSet::makeMove",timer_level_fine)); //timer for moves
+  timers.push_back(new NewTimer("ParticleSet::makeMoveOnSphere",timer_level_fine)); //timer for NLPP moves
+  timers.push_back(new NewTimer("ParticleSet::donePbyP",timer_level_fine)); //timer for donePbyP
   TimerManager.addTimer(timers[0]);
   TimerManager.addTimer(timers[1]);
+  TimerManager.addTimer(timers[2]);
 }
 
 ParticleSet::ParticleSet()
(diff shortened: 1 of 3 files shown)


src/Particle/ParticleSet.cpp
```diff
@@ -43,10 +43,12 @@ template<> int ParticleSet::Walker_t::cuda_DataSize = 0;
 
 void add_p_timer(std::vector<NewTimer*>& timers)
 {
-  timers.push_back(new NewTimer("ParticleSet::makeMove",timer_level_fine)); //timer for MC, ratio etc
-  timers.push_back(new NewTimer("ParticleSet::makeMoveOnSphere",timer_level_fine)); //timer for the walker loop
+  timers.push_back(new NewTimer("ParticleSet::makeMove",timer_level_fine)); //timer for moves
```
markdewing: The createTimer method on TimerManager will create and add a timer in one call.  So this could be simplified to
timers.push_back(TimerManager.createTimer("ParticleSet::makeMove", timer_level_fine)); // timer for moves
Then there is no need for TimerManager.addTimer(timers[0])

markdewing:
okay to test

src/Particle/ParticleSet.cpp
```diff
@@ -43,10 +43,12 @@ template<> int ParticleSet::Walker_t::cuda_DataSize = 0;
...
+  timers.push_back(new NewTimer("ParticleSet::makeMoveOnSphere",timer_level_fine)); //timer for NLPP moves
+  timers.push_back(new NewTimer("ParticleSet::donePbyP",timer_level_fine)); //timer for donePbyP
   TimerManager.addTimer(timers[0]);
```
prckent: Please explain the stack-like creation of the timers followed by fixed indexing [0], [1], [2]. I would prefer to avoid magic indices e.g. myTimers[1]

ye-luo:
@prckent The timers followed by fixed indexing was the original way to use timers. I'm not changing it. @markdewing has used enumeration instead of indices in DMCUpdatePbyPFast. I think we should update all the myTimers in many places from a dffierent PR. This PR is to improve the correctness and accuracy instead of readability.

markdewing:
There's also a way of setting up the timers using an enum for the timer array indices.

See the last unit test in src/Utilities/tests/test_timer.cpp, or in the DMC code in DMCUpdatePbyP.h and DMCUpdatePbyPFast.cpp

ye-luo:
Please merge this PR today. I would like to use it on my big runs which will probably start during weekend.

markdewing:
I'm okay with splitting the changes - adding new timers now, and moving to enums later.

PR #508, opened 2017-11-17
Fix GNU 7.2 warning about char array fixed length.

ye-luo:
Replace fixed length char array with std::string.


markdewing:
okay to test

Issue #518, opened 2017-11-24
WFtester grad-lap check script issue


ye-luo:
https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=short-bccH_2x2x2_ae-grad-lap-1-1-check&date=2017-11-24
SoA fails can be ignore but some failures in the AoS are due to script issue. For example, Intel2017-Complex-Release.

Paul-St-Young:
Ah, I see. I need to parse complex numbers of the form (a,b). Will commit a fix ASAP

ye-luo:
resolved.

PR #519, opened 2017-11-24
fix grad_lap wftest printing and parsing

Paul-St-Young:
WaveFunctionTester output was tweak to make real and complex branches
consistent


Paul-St-Young:
response to issue #518

markdewing:
okay to test

PR #520, opened 2017-11-26
Specify Kepler GPU in nightlies

prckent:
Update oxygen nightly reference script to currently run version. GPU tests are re-enabled specifying Kepler GPU. Note that since oxygen is nearly full, a reevaluation of the test matrix is needed.


prckent:
[Fix the thing](https://trello.com/c/fq5bi6xZ/2-fix-the-thing)

PR #521, opened 2017-11-27
Pooled memory

ye-luo:
In this PR, the memory buffer carried by each walker is replaced by a byte addressable PooledMemory implementation and a few optimizations are introduced to take advantage of the new facility. The mixed type support in PooledData has been reverted to avoid type mixing and enhance robustness.

The biggest advantage of this implementation is to eliminate the thread and walker memory replication in PbyP moves. However, this is not automatic. Old AoS classes still have the duplication but SoA ones engage the new memory attaching mechanism. @lshulen we will revise our projection of memory usage for ECP when needed.

The second advantage is the removal of serialization during the preparation of send/recv walkers since it has already been done in the memory pool. So it becomes much faster and there is no more need of MPI buffer memory.

Checked with realistic CPU and GPU runs. MPI handling change affects both CPU and GPU code paths.

@markdewing I only engage the dynamic population control Simple code path and reconfiguration code path. The rest schemes in dynamic population control are not touched since you will probably do some cleaning. We may need to implement the non-blocking case, contact me when you plan to touch it. There is a trick to handle all-particle move case.

@nsblunt Unfortunately I break the FDLRWfn tests. I have to consistently adopt the new registerData assumption but my change make the code wrong. Please take a look when you have time.


src/ParticleIO/ESHDFParticleParser.cpp
```diff
@@ -86,7 +86,7 @@ bool ESHDFIonsParser::put(xmlNodePtr cur)
   }
   delete [] species_names;
   ParticleSet::Tensor_t lat(ref_.Lattice.R);
-  ParticleSet::Buffer_t pbuffer;
+  PooledData<double> pbuffer;
```
prckent: Why double and not a parameterized type?

src/QMCWaveFunctions/AGPDeterminant.cpp
```diff
@@ -248,37 +248,27 @@ AGPDeterminant::evaluateLogAndStore(ParticleSet& P)
   d2Y = GeminalBasis->d2Y;
 }
 
-AGPDeterminant::ValueType
-AGPDeterminant::registerData(ParticleSet& P, PooledData<RealType>& buf)
+void
```
prckent: Do we have any tests to indicate that AGPDeterminant is functional pre/post modification?

src/QMCWaveFunctions/ConstantOrbital.h
```diff
@@ -79,11 +79,11 @@ class ConstantOrbital: public OrbitalBase
     return FakeGradRatio;
   }
 
-  virtual RealType registerData(ParticleSet& P, BufferType& buf) {return 0.0;}
+  virtual void registerData(ParticleSet& P, WFBufferType& buf) {}
```
prckent: Do we have any tests to indicate that ConstantOrbital is functional pre/post modification?

src/QMCWaveFunctions/IonOrbital.cpp
```diff
@@ -195,19 +195,18 @@ IonOrbital::evaluateLogAndStore(ParticleSet& P,
 }
 
 /** equivalent to evalaute with additional data management */
-IonOrbital::RealType
-IonOrbital::registerData(ParticleSet& P, PooledData<RealType>& buf)
+void
+IonOrbital::registerData(ParticleSet& P, WFBufferType& buf)
```
prckent: Do we have any tests to indicate the IonOrbital is working pre/post modification? When is it needed?

src/QMCWaveFunctions/einspline_helper.hpp
```diff
@@ -361,7 +361,7 @@ namespace qmcplusplus
     mpi::bcast(*myComm,nbands);
 
     //buffer to serialize BandInfo
-    PooledData<RealType> misc(nbands[0]*5);
+    PooledData<double> misc(nbands[0]*5);
```
prckent: Why double and not a parameterized type?

ye-luo:
registerData used to handle the memory pool request and initial WF component computing from scratch. In this PR, registerData no more takes care of the initial computing but an evaluateLog was added after registerData. For this reason, the return value of registerData may not be correct and thus removed. For classes not having tests, I did not remove the recompute inside registerData and the change should be low risk.

src/QMCWaveFunctions/IonOrbital.cpp
ye-luo: No idea even pre modification.

src/QMCWaveFunctions/ConstantOrbital.h
ye-luo: This is introduced only for unit tests. My modification must be correct as the unit tests pass.

src/QMCWaveFunctions/AGPDeterminant.cpp
ye-luo: We don't have tests. But my change should be low risk see my explanation comment below.

src/ParticleIO/ESHDFParticleParser.cpp
ye-luo: This is to pack send/recv data for the builder. Safe to use highest precision. I will change it to the full precision macro.

src/QMCWaveFunctions/einspline_helper.hpp
ye-luo: This is to pack send/recv data for the builder. Safe to use highest precision. I will change it to the full precision macro.

src/QMCWaveFunctions/TrialWaveFunction.h
```diff
@@ -282,9 +283,9 @@ class TrialWaveFunction: public MPIObjectBase
   void rejectMove(int iat);
   void acceptMove(ParticleSet& P, int iat);
 
-  RealType registerData(ParticleSet& P, BufferType& buf);
-  RealType updateBuffer(ParticleSet& P, BufferType& buf, bool fromscratch=false);
-  void copyFromBuffer(ParticleSet& P, BufferType& buf);
+  void registerData(ParticleSet& P, WFBufferType& buf);
```
markdewing: It would be good to add some description of these buffer methods - what they do and when should they be called.
My current understanding:
registerData allocates space in the buffer 'buf' based on the data type sizes in the ParticleSet objects.   It should be called before the buffer is used in updateBuffer/copyFromBuffer (?)

Is the buffer only used to for sending walkers across processors?  Or is it also the temporary storage for cached values?

updateBuffer - copies data from the TrialWaveFunction tree to the buffer (?).  The code also appears to update some values in ParticleSet (P.G and P.L), but maybe those values are scratch space used to put the gradient and laplacian in the buffer?

copyFromBuffer - copy data from 'buf' to the tree describing the wavefunction rooted in the TrialWaveFunction object.

src/Utilities/PooledMemory.h
```diff
@@ -0,0 +1,244 @@
...
+ * @tparam PageSize : page size in bytes, default=4096 (4K)
+ * @tparam Alloc : allocator, Mallocator<T, PageSize>
+ *
+ * The bulk part is accessed directly by address
+ * The scalar part works as PooledData, all the values are static_cast to T_scalar.
+ */
+template<typename T_scalar=OHMMS_PRECISION_FULL,
+         size_t PageSize=4096,
+         typename Alloc=Mallocator<char, PageSize> >
+struct PooledMemory
```
markdewing: There should be a unit test added for this class.

src/QMCWaveFunctions/TrialWaveFunction.h
ye-luo: The basic description is in OrbitalBase class.
registerData requests space in the buffer 'buf' based on the data type sizes of the objects in this class.
After all the registration in walker initialization, buf.allocate() is called to actually allocate the space.
copyFromBuffer copy data or attach memory from 'buf' to the class and updateBuffer put the data or forward the memory cursor.

The buffer is used not only for MPI but more for the case when the number of walkers different from the number of threads.

ParticleSet also registerData, copyFromBuffer, updateBuffer. This is purely for MPI. P.G and P.L are not scratch space but used everywhere in the code.

I agree with you that it will be healthier to have more in-source documentation. I will update them.

src/Utilities/PooledMemory.h
ye-luo: I agree. Do you have time to help me on making the unit tests?

src/QMCWaveFunctions/TrialWaveFunction.h
markdewing: These comments will be a good addition the documentation for these functions - the context of how a function is used is sometimes the hardest to capture, but also the most valuable for someone trying to understand the code.

It would also be good to add a comment here pointing to OrbitalBase as the location of the documentation.  TrialWaveFunction doesn't inherit from OrbitalBase, so it's not obvious where to find it.

src/Utilities/PooledMemory.h
markdewing: Yes, I can help with making some unit tests.

src/Utilities/PooledMemory.h
```diff
@@ -0,0 +1,244 @@
...
+
+namespace qmcplusplus
+{
+
+/** Memory pool to manage arrays+scalars with alignment
+ * @tparam T_scalar : type of the scalar data type, typically OHMMS_PRECISION_FULL
+ * @tparam PageSize : page size in bytes, default=4096 (4K)
+ * @tparam Alloc : allocator, Mallocator<T, PageSize>
+ *
+ * The bulk part is accessed directly by address
```
markdewing: I assuming these design points:
 - Reuse sections of the buffer directly as storage in some wavefunction classes (saves memory and copy overhead).  The attach methods were added to some container classes to enable this.
 - Because the buffer storage is used directly, maintaining alignment is important
 - Data should be stored in a single contiguous container, making it easy to pass the data as a single MPI message (otherwise it would be simpler to use separate structures for block and scalar data)

Because the new reuse of buffer data, I am concerned about issues due to object lifetimes - that the wavefunction classes could access buffer locations after the buffer has been deallocated or reused.
What is the lifetime?  It appears copyFromBuffer attaches the storage from the buffer.  But when is the buffer data no longer needed?  When copyFromBuffer is called again on a new walker.
It might be possible to implement some sort of reference counting scheme to detect this case.

I'm also concerned about the variability in how the attach and resetByRef methods handle freeing existing memory.  resetByRef will free existing memory, but Vector::attach will throw an exception if previously allocated memory is present (meaning the user needs to free the memory before calling attach)

prckent:
Has this been run through memory leak checking?

src/Utilities/PooledMemory.h
ye-luo: 1) Yes.
2) Yes.
3) PooledData was the old implementation mixing block and scalar data. I decide to keep supporting block and scalar for backward compatibility.

In any QMC driver, copyFromBuffer is always called before any pbyp moves on wavefunction classes. So the buffer location is always correct. updateBuffer can be called as many as you want.

The buffer is persistent with the walker. Once created, always valid until the walker is destroyed.
In DMC, load balancing is the last operation in an iteration, the next cycle will redo the copyFromBuffer ensuring correct assignment.

resetByRef was introduced earlier than attach. I think we should change its name as well as behaviour to throwing an exception.

ye-luo:
This PR can not leak memory because the memory allocation and deallocation of PooledMemory is handled by std::vector. The only risk is pointer double free which doesn't happen because the container classes know whether it holds actual memory or not.

ye-luo:
Checked a DMC run with valgrind. No leaking is related to PooledMemory.

ye-luo:
@prckent moving forward?

lshulen:
I'd like to see this merged sooner rather than later.

prckent:
Remove the [WIP] when you think this is ready to merge.

@ye-luo I would prefer not to break FDLR. 
1. How tricky is the fix?
2. What feature of FDLR causes it to break (vs any other functionality)?
3. Does this imply there are other parts of the code where there might be breakage that we have not caught?
4. Does Nick have any bandwidth? (We can't assume so for changes like this, which are more "on us")

prckent:
@markdewing Have your comments been satisfied?

ye-luo:
@prckent 
1. No idea about how tricky is the fix. In principle, it should be simple. FDLR is not like our other wavefunction components that following certain assumptions. This one is more like a hack to the code. Certain portion of computation can be added in several places and will work but when we change the code, it may break.
2. Sadly, no idea. I can only answer this when the issue is fixed.
3. This change was designed to provide backward compatibility. In principle no.
4. No idea. Not seeing any response.

I would prefer to merge it first and resolve the issue later since it is not affecting the core functionality.

nsblunt:
I'm sorry for not seeing this and responding sooner. I'll try and spend some time tomorrow looking at this, but can't guarantee that I'll be able to fix it quickly.

markdewing:
The name of the 'attach' method on PooledMemory should be changed.  It's confusing to have two methods with complementary actions with the same name.
More possibilities:
  getReference, getMemorySegment, borrowReference, borrowMemorySegment, checkOutMemorySegment (but the code already uses checkOut for optimization variables)

I think my other concerns could be addressed in future PR's.  (Most of them have to do with making it easier to catch incorrect usage of PooledMemory when changing code that uses it)

ye-luo:
@markdewing "two methods with complementary actions with the same name". What did you mean two methods? For the name, borrowMemorySegment is a better name.

src/QMCWaveFunctions/EinsplineSetBuilderCommon.cpp
```diff
@@ -138,7 +138,7 @@ EinsplineSetBuilder::BroadcastOrbitalInfo()
   int numIons = IonTypes.size();
   int numAtomicOrbitals = AtomicOrbitals.size();
   int numDensityGvecs = TargetPtcl.DensityReducedGvecs.size();
-  PooledData<RealType> abuffer;
+  PooledData<double> abuffer;
```
prckent: I think this buffer type should be parameterized. Same for the other declarations. Why was it changed?

ye-luo: This is in builder to send receive meta data info. Some infos are floats and some are doubles. In order not to loose precision, using double type PooledData is safe. This is local and I don't want to entangle it with others. Instead in wavefunction, typedef WFBufferType is used to have a consistent control.

ye-luo: I created an issue for the documentation task. I need to learn and practice a bit doxygen style before doing it.

markdewing:
By "two methods", I mean that there is PooledMemory::attach and Vector::attach (and Matrix) - they have the same name, but different meanings.

ye-luo:
@markdewing I was just wondering how to change these two cases. Do you have suggestion how to distinguish them?

markdewing:
Change PooledMemory::attach to borrowReference.    The name 'attach' is fine for  the Matrix and Vector classes.  Unless you think it would be clearer to call it attachReference?

ye-luo:
After a quick chat with @markdewing, I will change PooledMemory::attach to lendReference and Matrix, Vector attach to attachReference.

markdewing:
PooledMemory is really writing a memory allocator (which is fine - we need the specialized functionality).
But for a memory allocator, it feels like it has a strange interface.
 - 'add' will allocate the space (well, not quite, but it reserves the space to be allocated).  It does not return a pointer right away, like malloc does.
 - 'allocate' performs the internal allocation (malloc would do this on demand for every allocation)
 - 'put' and 'get' will write/read to the allocated memory (like memcpy)
 - 'lendReference' (formerly 'attach') will return the memory address (this is like the return value from malloc)

What I don't like is that the all the access operations (put/get/forward/lendReference/etc) all need to occur in the same order as the add operations in order to be correct.

The current interface seems modeled after streaming I/O, but it might be better to make it act more like a memory allocator:
 - 'add' would return an index value (similar to a pointer returned by malloc.  It would not be a raw memory pointer, but an index into the space of a particular memory pool).  One tricky part is that there are two memory spaces for each pool - one for scalars and one for arrays.   I don't know if keeping them separated would be an issue or not?
 - 'allocate' - reallocation of underlying storage could be performed at each 'add' for absolute correctness, but seems a bit wasteful, since the sizes are only set up once.  This could stay the same.
- put/get/lendReference would all take the index value as an additional parameter.   This would ensure the correct segment of the memory pool is accessed, regardless of order.   Accessing the memory segments in order might be necessary for performance, but would not be correctness issue.
 - there is no analog of 'free'.   While not strictly necessary for correct operation, it might be nice to have in order to detect certain problems.

I'm not suggesting these changes for this PR, but something in the future.

prckent:
@markdewing Please can you setup an issue describing these changes? I think they should be immediately worked on. However since we know the functionality is needed I am open to merging this PR now. However the issue *must be* prioritized otherwise this will never get fixed.

Hidden state machines and dependencies such as you list are dangerous and should be avoided if at all possible. In this case I think they are.

jngkim:
I admit PooledMemory naming is not the best choice.


PooledMemory is not an allocator, since it does not meet the requirements,
and thus there is no reason to make the names sound like allocators.
Pooled memory allocators have existed in boost but their design goals are
very different from this.

It is a memory manager of walker for QMCPACK. Maybe, at some point, one can
make it more generally useful. Then, changing these names to standards
should be easy.

ye-luo:
PooledMemory is more focused on the specialized needs of walker memory management in QMCPACK.

Let me give the story.
PooledData was the old one and the underlying container is std::vector. I noticed two issues with it.
1. we are adding data like scalars, vectors of size Nelec, and matrices of size Nelec^2. When Nelec^2 is added and then a scalar is checked in, the std::vector will reserve Nelec^2 memory and causing huge waste.
2. I tried resizing to the desired size before adding the data. The memory can not be recycled fully.
So I started PooledMemory. Which should provide the maximal backward compatibility and solve the previous two issue and also incorporate the needs from communication.

Now I comment on the points given by Mark.
1. 'add' only register the size but do not allocate memory. This is intended to avoid frequent inefficient reallocation. Regarding separating buffers, historically, two buffer was owned by PooledData to support mixed precision, one for float and one for double. The main pain of having two is always do twice which requires specializing utilities. For example, when you send it, you need a specialized MPI wrapper to do two internal sends. Since PooledMemory is by byte, adding the scalar in one buffer is easy and simplifies coding a lot.
2. 'allocate' only when every thing is registered. This is intended to avoid any inefficient memory recycling from the OS and std::vector.
3. 'put/get/lendReference'. I think it safer to record and check the index. However, it requires touch every WF component. Since PooledData has been used for about a decade and didn't show much issue, it don't see this urgent.
4. Your are right, we only have clear but not free. It is trivial to just add it.

Here is a use case.
For a determinant matrix, it holds its own memory when the class is built. Now if you use walker buffer,
It first registers the memory with add and then frees its memory. When the registration of the whole WF is done. The only once allocate recycles the memory. This seems working well.

prckent:
Can we delete the old pooleddata.*?

ye-luo:
Unfortunately not now.
PooledData is used by collectables and several misc. places for convenience.

PooledData convert the input data element by element.
PooledMemory takes bulk data by memory but scalar data by element.
They do not fully overlap.

We may switch collectables to PooledData but more care is needed. I saw codes register with one object but put with another object. They have the same vector length but different precision. So PooledData is fine but PooledMemory will go wrong.

prckent:
So we need an issue to deal with this, in addition to Mark's interface update. We need a strategy to remove the old code / refresh it / keep things synced / share common functionality etc. Impractical to keep increasing the surface area of the code. We need to improve the navigation for people new to the code.

ye-luo:
@prckent absolutely. Since the whole estimator use needs a redesign, the PooledData will be taken into account at that time.

ye-luo:
Can we get this merged today?
Then my other PR can be reviewed and merge hopefully before the holiday season starts.
I can submit an issue to track FDLRWfn fix.

prckent:
@ye-luo Please make an issue noting the cleanup issues around the two Pooled approaches.


@nsblunt Do you mind if we break FDLR?

nsblunt:
@prckent yes that's fine, apologies for not fixing this before the merge. Create an issue and I'll look into it

prckent:
Thanks Nick.

Issue #404, opened 2017-10-05
Verbosity level of output should be settable


prckent:
At least half the current output should not be printed for "normal" users with well functioning calculations, while "debug" or "verbose" output could potentially require more than we currently print. 

It would be helpful if the verbosity level can be set in the input XML as well as on the command line.

I suggest we discuss implementation here before someone does it. e.g. Could we have a per XML section verbosity parameter? i.e. set DMC to verbose but leave the wavefunction setup at a default, or would that be too fiddly/not help any problems? Are two settings "normal", "high" enough?

jtkrogel:
I think if information is potentially useful, then it should be printed somewhere.  Perhaps we have a stripped down and very clean set of info printed to stdout, while a much more detailed ("verbose") set of information is printed to a different output/log file with a standardized name?

prckent:
Doing I/O is extremely problematic on HPC systems, so I am not keen to log extensively by default.

jtkrogel:
A single verbosity parameter is probably sufficient.  Starting with normal/high as options and expanding if ever necessary sounds reasonable to me.

QMCPACK currently lacks an input section to communicate global information.  We could co-opt <project/> for this as it does contain some global info already (id and series), and just add <parameter/>'s.  This would avoid the proliferation of xml elements, but we could alternatively add e.g. <controls/> or <global/> for this purpose.

prckent:
So the suggestion is:
```
<simulation>
   <project id="qmc_short" series="0">
      <application name="qmcapp" role="molecu" class="serial" version="1.0" verbosity="high"/>
   </project>
```
This would also be a good place to put the command line options

jtkrogel:
I was actually thinking:

```
<simulation>
   <project id="qmc_short" series="0">
      <parameter name="verbosity">   high  <parameter/>
   </project>
```

The `<application/>` element doesn't add much, and I think we can stop using it.  I agree that this is a good place to add all the other command line options.  Perhaps we move "id" and "series" to parameters too, but leave the option to read them as attributes for backwards compatibility.

prckent:
Better.

Is the application element actually needed? The code screams "legacy".

markdewing:
[...]

PR #527, opened 2017-11-29
Initial addition of command line verbosity flag

markdewing:
This PR adds the classes InfoStream and OutputManager (which will replace OhmmsInform and OhmmsInfo)
The '-verbose' command line option takes low,high, and debug for options, eg 'qmcpack -verbose=high'
The default value is high for now.  Eventually the default verbosity will be low.

Also some conversion to using app_summary in the initial output.
Partially addresses #404

This code has not been heavily tested, and may not be ready to check in just yet.  It's posted to get some feedback on the code, and for discussion on some remaining questions.

1. Adding tags for debug output in the code:  app_debug will take an optional tag name.

       app_debug("here") << "some output"
2. Specifying debug tags on the command line. Options:
      A.  One option is to add to the -verbose option  

            -verbose=debug=tag1 or -verbose=debug,tag1
            
           But these seem awkward

      B. Another option is a separate command line option, like -debug=tag.  This should probably also imply -verbose=debug.  Having two ways to turn on debug level of output seems a little ugly.


src/Utilities/InfoStream.h
```diff
@@ -0,0 +1,93 @@
...
+
+// Bit bucket to discard output
+// From https://stackoverflow.com/questions/7818371/printing-to-nowhere-with-ostream
+class NullStreamBuf : public std::streambuf
+{
+protected:
+    virtual int overflow(int c) { return c; }
+};
+
+class NullStream : public NullStreamBuf, public std::ostream
```
ye-luo: In the old code, we have std::ostream(0) as a null stream. This was because std::ostream() is not thread safe and crashing the code. I feel that we need to use the old way since NullStream is not safe.

ye-luo:
Can we just have `-verbose=tag1` and if you need multiple debug tag, do `-verbose=tag1,tag2`

jtkrogel:
What are the values the "tag*" take?  Are these targeted sections of output, e.g. orbitals, jastrows, etc?

I'm OK with list inputs as Ye suggests.  Still in favor of moving all input to the input file.

prckent:
I think the tag should be "verbosity".
 
I suggest we support global settings on the command line only. Global and per section settings can be specified in the input XML. I don't think it is worth going to the trouble (creation, docs, maintenance) of defining how to address individual sections on the command line. If someone needs more than a global switch - likely a very rare case - they can edit the input file.

markdewing:
I like @prckent 's suggestion - it would greatly simplify the implementation.

For putting the flag on the input file, the simplest implementation is to turn on debugging when the tag starts, and turn it off when processing the tag ends.  (More generally, this would change verbosity levels at these points)

For example
```
<wavefunction verbosity="debug">
   <!-- wavefunction elements -->
</wavefunction>
```
This would turn on debug output for all the wavefunction setup and for processing all the tags under wavefunction.

ye-luo:
I think we should first focus simple things and then improve if really needed.
Using tags or piece wise verbosity is more complicated.
For users, the simply the better. I remember QE had low, medium, high levels but now medium has been removed since very few people use it and hard to maintain.
For developers, I think we should use both verbosity levels or using facilities provided by C++ like assertion which cost nothing when built as release.

markdewing:
I think this PR is ready.  It completes the switch of logging infrastructure to the new version.
It does not add setting of verbosity to XML tags, but that is independent and can be added later.

prckent:
I agree. 

If there aren't any objections I'll merge at 5pm EST today.

ye-luo:
Did you resolve the thread-safety issue of Nullstream?

ye-luo:
Please also check build level 5.

markdewing:
Latest updates switch to using ostream(NULL) for the null output stream.  This puts the stream into a error state, which skips most of the formatting and output.
Also fixed the build for build level 5.

ye-luo:
Thank you @markdewing. All my concerns are addressed.

ye-luo:
Just a question,
@markdewing with this PR, do we still print the initial available node memory at the beginning of the run?

markdewing:
@ye-luo Yes, it still prints the node memory at the beginning.   In general the output should be unchanged by this PR (except for one place in Lattice where some cleanup slipped in)

PR #532, opened 2017-12-04
Updating gaussian converter

anbenali:
Fix small bug in converter for  All electron orbital correction when … using HDF5. Behavior was duplicating CuspInfo line


markdewing:
okay to test

prckent:
After discussing with @anbenali , merge deferred for additional updates.

anbenali:
I am done with this part. Ready to merge. Rest of commits will be part of different branch.

src/QMCTools/convert4qmc.cpp
```diff
@@ -55,10 +55,23 @@ int main(int argc, char **argv)
...
+
+  
   std::string punch_file;
   std::string psi_tag("psi0");
   std::string ion_tag("ion0");
-  std::string prefix("sample");
+  std::string prefix(argv[2]);
+
+
+  std::string delimiter =".out";
```
prckent: I think we could improve the handling here. Would it be simpler to add the requirement that the suffix actually is *.out ? Is this completely standard for GAMESS output?

src/QMCTools/convert4qmc.cpp
```diff
@@ -55,10 +55,23 @@ int main(int argc, char **argv)
...
+  int pos = 0;
+  std::string token;
+  pos = prefix.find(delimiter);
+  token = prefix.substr(0, pos);
+  prefix.erase(0, pos + delimiter.length());
+  prefix=token;
+  std::cout << "PREFIX="<<prefix << std::endl;
```
prckent: Friendlier formatting "Using " << prefix << " to name output files"

src/QMCTools/QMCGaussianParserBase.cpp
```diff
@@ -1566,6 +1575,7 @@ void QMCGaussianParserBase::createGridNode(int argc, char** argv)
             if(a == "-nojastrow")
```
prckent: Remove the nojastrow option and *always* create both files .wfj and .wfnoj

I don't think space can be a concern due to the hdf5 options.

ye-luo:
Can we have just one prefix.wfs.xml instead of two files with different names? I just feel it unnecessary and harder for scripting or making workflow. If the xml file is large, we make things even worse now.
For the particle set file, it is describing the partcle sets (elecs, ions, cell) not just the structure. I would prefer not changing it.

prckent:
@ye-luo  

Wavefunctions: I have decided that we are going to strongly encourage people to run checks with no jastrow. As the inputs currently work I think this requires two files. We will also eventually be producing QMCPACK inputs to handle these files so that we always produce 

Structure: The important contents of this file is the atom positions or structure. Other ideas are welcome, but the cryptic, non-user friendly, and code specific "particle set" terminology is not going to be used.

prckent:
General comment: This converter is the first thing that new users of QMCPACK are likely to encounter. e.g. "Can I run a water molecule?". Suggestions for updates should be viewed through this lens.

prckent:
@ye-luo Misunderstood your comment. Good idea. @anbenali Do you think these will be too big in the large molecule case or can the HDF5 input also get the structure? If the size/complexity is no unreasonable, why not a single .wfj.xml for the wavefunction with jastrow and a .wfnoj.xml for the determinant(s) only file?

src/QMCTools/convert4qmc.cpp
anbenali: this will pick anything that has .out* (so out or output will be taken)

src/QMCTools/convert4qmc.cpp
anbenali: Done

ye-luo:
The only difference between .wfj.xml and .wfnoj.xml is only a few Jastorw blocks. The major contents (determinants dominant) are identical. I have seen files as large as 60MB for large molecules.
I feel it unnecessary to generate both files and users can very easily copy the file and delete the jastrow blocks if they need a noJ wavefunction for testing. It is a no pain step for user and they should verify the file contents rather than blindly trust the file name.

tests/converter/Testing/Temporary/CTestCostData.txt
```diff
@@ -0,0 +1 @@
+---
```
markdewing: This file should not be checked in.

tests/converter/Testing/Temporary/LastTest.log
```diff
@@ -0,0 +1,3 @@
+Start testing: Dec 04 22:13 CST
```
markdewing: This file should not be checked in.

tests/converter/Testing/Temporary/CTestCostData.txt
anbenali: Done

tests/converter/Testing/Temporary/LastTest.log
anbenali: Done

src/QMCTools/convert4qmc.cpp
```diff
@@ -55,10 +55,23 @@ int main(int argc, char **argv)
   QMCGaussianParserBase *parser=0;
   int iargc=0;
   std::string in_file(argv[1]);
+
+  
   std::string punch_file;
   std::string psi_tag("psi0");
   std::string ion_tag("ion0");
-  std::string prefix("sample");
+  std::string prefix(argv[2]);
```
ye-luo: This is not safe, based on the help info. The first argument can be omitted and argv[2] is null.

anbenali: The first argument omitted never works. 
Maybe it is time to remove that option and make a clean fail if you don't provide the first argument.

ye-luo: I tried it. It works on our test cases.

anbenali: In all cases, I "secured" it for now.

anbenali:
Remaining issue to discuss:
- wfs VS wfj +wfnoj

I personally prefer only one file but I do not care that much about it to spend time on the issue. I will go with majority.

src/QMCTools/QMCGaussianParserBase.cpp
anbenali: Any Update on this? @ye-luo @prckent ? 
Since you are the only two who cared, what do you prefer? On file? or two files. If two files, which one goes in the Input.

ye-luo: I prefer it simple and better if backward compatiable. Just one file. noj can be created manually by users in 2 secs.

prckent:
Please don't print the basis functions on stdout.

anbenali:
From @prckent : I addressed all your requests:

>>    Support -gamess , deprecate -gamessAscii.
  Done
>>    Don't print the basis functions on stdout.
  Since this is useful when introducing a new converter, I added a tag -debug that will print it. By default it is not printed anymore.

>>    Don't write the .combo files
only printed with the -debug
 >>  Presumably these can become huge? Does anyone use them?
No idea who uses them but they can be useful so accessible through -debug
   
>> Write wfj and wjfnoj
  all set of combinations are written by default.

>>  We can add an option to only write one later
I do need this one as I really do not want to have multiple files laying around. by using the tag "-writewfjonly" only the requested files will be created.
    
>> Remove all the gpu="no" tags. This clutters the file
        The code will stop if anything non-supported is reached. Sadly, the code does not support cpu fallbacks.

Commented for now.

prckent:
@anbenali Thanks. I am going to merge this so that the nightly tests can start seeing these updates.  Revisions to the example inputs can be addressed in a later issue/PR.

PR #533, opened 2017-12-04
Fix n(k) for walker weight and twist averaging

kylanpaa:
Walker weight was missing in the n(k) estimator. Changed sign in the dot product for dot(k,dr), which makes n(k) correct with twist averaging.

### DIFF
--- a/src/QMCHamiltonians/MomentumEstimator.cpp
+++ b/src/QMCHamiltonians/MomentumEstimator.cpp
@@ -21,7 +21,7 @@
 #include <Utilities/SimpleParser.h>
 #include <Particle/DistanceTableData.h>
 #include <Numerics/DeterminantOperators.h>
-
+//#include <cstdint>
 #include <set>
 
 namespace qmcplusplus
@@ -42,7 +42,8 @@ void MomentumEstimator::resetTargetParticleSet(ParticleSet& P)
 }
 
 MomentumEstimator::Return_t MomentumEstimator::evaluate(ParticleSet& P)
-{
+{ 
+  RealType w=tWalker->Weight;
   const int np=P.getTotalNum();
   nofK=0.0;
   compQ=0.0;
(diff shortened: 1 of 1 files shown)


prckent:
OK to test

src/QMCHamiltonians/MomentumEstimator.cpp
```diff
@@ -21,7 +21,7 @@
 #include <Utilities/SimpleParser.h>
 #include <Particle/DistanceTableData.h>
 #include <Numerics/DeterminantOperators.h>
-
+//#include <cstdint>
```
ye-luo: Please remove this line.

src/QMCHamiltonians/MomentumEstimator.cpp
```diff
@@ -213,7 +218,7 @@ bool MomentumEstimator::putSpecial(xmlNodePtr cur, ParticleSet& elns, bool rootN
           kpt[2]=k-twist[2];
           //convert to Cartesian: note that 2Pi is multiplied
           kpt=Lattice.k_cart(kpt);
-          kPoints.push_back(kpt);
+	  kPoints.push_back(kpt);
```
ye-luo: Please revert the tab to white spaces.

ye-luo:
@kylanpaa Could you comment which cases are verified? Both real and complex? Non high symmetry twists?

kylanpaa:
Real and complex codes have been verified. Test sets have used, e.g., twist grids of 2x1x1, 3x1x1, 2x2x2, 3x3x3, and 4x4x4. All cases have been tested with the complex code, and real code has been tested in the cases where it has been possible (2x1x1 and 2x2x2).

prckent:
@kylanpaa Thanks. For future reference, is the HEG code capable enough for these tests or did you do all the testing with a metal oxide and splines?

ye-luo:
@markdewing Any comments? I need to propagate this change to the actually needed code for production as soon as possible.

kylanpaa:
@prckent I believe the HEG is only capable for Gamma point calculations at present. Therefore, the tests described above were done with VO2, and some additional tests also with Na.

Issue #87, opened 2017-01-22
 hierarchical timers are inconsistent with ensemble runs


ye-luo:
When trying an ensemble run, I just noticed that the info printed in *info.xml is for the first group of ranks dealing with the first job. However the name of the file does not carry g000 like other files in a ensemble run. This signals an inconsistency in communicator and also design.
I think we probably need to move the timer entity inside QMCMain (per QMC run) rather than placing it in qmcapp (the ensemble run). Another reason, if the ensemble job is killed by the walltime, the finished calculations can still spit out the timer info.

PR #537, opened 2017-12-05
Switch timer summary output to use app_log

markdewing:
The code used 'printf' previously and would not redirect timer output appropriately with ensemble runs.
I started to try converting the formatting to C++ style, but gave up.   The existing formatting works fine. 
Now it prints to a temporary buffer which is then output to app_log.  Feels like a hack, but it works.

Partially addresses #87.  The name of the info file still needs to be fixed.

### DIFF
--- a/src/Utilities/NewTimer.cpp
+++ b/src/Utilities/NewTimer.cpp
@@ -240,13 +240,13 @@ TimerManagerClass::print(Communicate* comm)
 #ifdef USE_STACK_TIMERS
   if(comm == NULL || comm->rank() == 0)
   {
-    printf("Stack timer profile\n");
+    app_log() << "Stack timer profile" << std::endl;
   }
   print_stack(comm);
 #else
   if(comm == NULL || comm->rank() == 0)
   {
-    printf("\nFlat profile\n");
+    app_log() << "\nFlat profile" << std::endl;
   }
   print_flat(comm);
 #endif
@@ -265,16 +265,18 @@ TimerManagerClass::print_flat(Communicate* comm)
   {
     #pragma omp master
     {
+      char tmpout[256];
       std::map<std::string,int>::iterator it(p.nameList.begin()), it_end(p.nameList.end());
       while(it != it_end)
       {
         int i=(*it).second;
         //if(callList[i]) //skip zeros
-        printf ("%-40s  %9.4f  %13ld  %16.9f  %12.6f TIMER\n"
+        snprintf (tmpout, 255, "%-40s  %9.4f  %13ld  %16.9f  %12.6f TIMER\n"
         , (*it).first.c_str()
         , p.timeList[i], p.callList[i]
         , p.timeList[i]/(static_cast<double>(p.callList[i])+std::numeric_limits<double>::epsilon())
         , p.timeList[i]/static_cast<double>(omp_get_max_threads()*comm->size()));
+        app_log() << tmpout;
         ++it;
       }
     }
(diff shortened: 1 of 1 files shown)


src/Utilities/NewTimer.cpp
```diff
@@ -318,9 +320,14 @@ TimerManagerClass::print_stack(Communicate* comm)
       max_name_len = std::max(name_len, max_name_len);
     }
 
+    char tmpout[256];
     std::string timer_name;
     pad_string("Timer", timer_name, max_name_len);
-    printf("%s  %-9s  %-9s  %-10s  %-13s\n",timer_name.c_str(),"Inclusive_time","Exclusive_time","Calls","Time_per_call");
+
+    snprintf(tmpout, 255, "%s  %-9s  %-9s  %-10s  %-13s\n",timer_name.c_str(),"Inclusive_time","Exclusive_time","Calls","Time_per_call");
+    tmpout[255] = '\0';
```
ye-luo: A terminating null character is automatically appended after the content written.
Please change all the 255 to 256 because snprintf only writes n-1+"\0" characters.
Found in http://www.cplusplus.com/reference/cstdio/snprintf/

src/Utilities/NewTimer.cpp
```diff
@@ -265,16 +265,20 @@ TimerManagerClass::print_flat(Communicate* comm)
...
       {
         int i=(*it).second;
         //if(callList[i]) //skip zeros
-        printf ("%-40s  %9.4f  %13ld  %16.9f  %12.6f TIMER\n"
+        snprintf (tmpout, bufsize, "%-40s  %9.4f  %13ld  %16.9f  %12.6f TIMER\n"
         , (*it).first.c_str()
         , p.timeList[i], p.callList[i]
         , p.timeList[i]/(static_cast<double>(p.callList[i])+std::numeric_limits<double>::epsilon())
         , p.timeList[i]/static_cast<double>(omp_get_max_threads()*comm->size()));
+        tmpout[bufsize-1] = '\0';
```
ye-luo: I mean no need for this line since snprintf does this.

src/Utilities/NewTimer.cpp
```diff
@@ -318,9 +322,15 @@ TimerManagerClass::print_stack(Communicate* comm)
...
     }
 
+    const int bufsize = 256;
+    char tmpout[bufsize];
     std::string timer_name;
     pad_string("Timer", timer_name, max_name_len);
-    printf("%s  %-9s  %-9s  %-10s  %-13s\n",timer_name.c_str(),"Inclusive_time","Exclusive_time","Calls","Time_per_call");
+
+    snprintf(tmpout, bufsize, "%s  %-9s  %-9s  %-10s  %-13s\n",timer_name.c_str(),"Inclusive_time","Exclusive_time","Calls","Time_per_call");
+    tmpout[bufsize-1] = '\0';
```
ye-luo: The same here since snprintf does this.

src/Utilities/NewTimer.cpp
```diff
@@ -330,12 +340,14 @@ TimerManagerClass::print_stack(Communicate* comm)
...
       std::string padded_name_str;
       pad_string(indented_str, padded_name_str, max_name_len);
-      printf ("%s  %9.4f  %9.4f  %13ld  %16.9f\n"
+      snprintf (tmpout, bufsize, "%s  %9.4f  %9.4f  %13ld  %16.9f\n"
               , padded_name_str.c_str()
               , p.timeList[i]
               , p.timeExclList[i]
               , p.callList[i]
               , p.timeList[i]/(static_cast<double>(p.callList[i])+std::numeric_limits<double>::epsilon()));
+      tmpout[bufsize-1] = '\0';
```
ye-luo: The same here since snprintf does this.

src/Utilities/NewTimer.cpp
markdewing: Ah, you're right.  I was thinking of the _snprintf on Windows, which doesn't properly null-terminate if it overwrites the buffer.

Issue #442, opened 2017-10-26
bccH 2x2x2 and 3x3x3 runs and derivative checks are failing


prckent:
```
Super twist #0:  [   0.00000   0.00000   0.00000 ]
Using supercell twist 0:  [   0.00000   0.00000   0.00000]
Using 1 copies of twist angle [-0.000, -0.000, -0.000]
Using real orbitals.
Fatal Error. Aborting at backflow optimization is broken with UseRealOrbitals
```
Lots of breakage on cdash. 

I suspect a simply incomplete merge or something wrong with the test definition. Hopefully an easy fix.

Paul-St-Young:
@prckent, my current workarounds for these failures are to re-route code path. For example, to fix the above test, I simply force `UseRealOrbitals=true` at the current abort. After discussion with @ye-luo in #373, we have decided revert the workaround and leave the broken tests in the PR to alert problems.

@prckent and @ye-luo, please let me know how to proceed. It is not immediately clear to me where the bug is. The current tests will result in a red cdash. For a green cdash, I can either:
1. re-apply workarounds OR
2. remove broken tests

ye-luo:
@Paul-St-Young I think your workaround force `UseRealOrbitals=true` If you did a typo in the previous post, please edit it and remove this comment.

ye-luo:
In principle, if complex is good, fixing the real case is fairly close. Find the correspondent code path in EinsplineSet.cpp and compare it with the complex working codepath.

ye-luo:
There are three code path Spline to SPO, R2R and C2R and C2C. The old EinsplineSet also has a similar setting.
R2R: HalfG is either 0 or 1 for distinguishing Gamma and X.
C2R: When you use tiling and your twist angle is 0 or 0.5, use exp(k,r) to apply the phase in the supercell. This k is not the twist angle but the k point of the primitive cell. MakeTwoCopies is because of symmetry. If true, each complex spline contributes two real SPO.
C2C: Any twist angle. Each complex spline always contributes a complex SPO.

ye-luo:
resolved by #1204

PR #539, opened 2017-12-05
MNC to fix backflow optimization for QMC_COMPLEX

Paul-St-Young:
MNC stands for Minimum Necessary Change

all backflow tests now pass when QMC_COMPLEX=1

partially address #442

### DIFF
--- a/src/QMCWaveFunctions/EinsplineSet.cpp
+++ b/src/QMCWaveFunctions/EinsplineSet.cpp
@@ -1860,14 +1860,6 @@ EinsplineSetExtended<StorageType>::evaluate_notranspose(const ParticleSet& P, in
     int psiIndex(0);
     for (int j=0; j<NumValenceOrbs; j++)
     {
-      bool trs(true);
-      for(unsigned a0(0); a0<OHMMS_DIM; a0++)
-      {
-        double olp= std::abs((PrimLattice.R[a0]*kPoints[j][a0])/M_PI +1e-6);
-        olp -= (int)std::round(olp);
-        if (olp>1e-4)
-          trs=false;
-      }
       if (MakeTwoCopies[j])
       {
         psi(i,psiIndex)=imag(StorageValueVector[j]);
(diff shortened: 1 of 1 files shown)


markdewing:
okay to test

src/QMCWaveFunctions/EinsplineSet.cpp
```diff
@@ -1864,7 +1864,7 @@ EinsplineSetExtended<StorageType>::evaluate_notranspose(const ParticleSet& P, in
       for(unsigned a0(0); a0<OHMMS_DIM; a0++)
       {
         double olp= std::abs((PrimLattice.R[a0]*kPoints[j][a0])/M_PI +1e-6);
-        olp -= (int)olp;
+        olp -= (int)std::round(olp);
```
ye-luo: This part of code seems suspicious to me. PrimLattice.R is a tensor 3x3. The loop only uses the first 3 elements.
For the checking, I would prefer to remove 1e-6 and abs in the first line and change the second line to
`olp = std:abs(olp-std:round(olp));

Paul-St-Young: @ye-luo I think you are right to suspect this loop. It tries to determine if the desired orbital `j` can be obtained from a special k-point in the primitive cell and sets the `trs` flag. The `trs` flag, in turn, determines how the derivatives are assigned. The current implementation passed all the tests I provided, but I see now that it will fail in general.

How is k-point classification done in the current spline implementation? Is there a routine I can steal instead of this somewhat ad hoc for loop?

ye-luo: trs=true seems like the very very old code path to UseRealOrbitals (R2R) case which you don't need to use.
The contents should have been already in
```
EinsplineSetExtended<double>::evaluate_notranspose(const ParticleSet& P, int first, int last,
    RealValueMatrix_t& psi, RealGradMatrix_t& dpsi,
    RealHessMatrix_t& grad_grad_psi,
    RealGGGMatrix_t& grad_grad_grad_logdet)
```
So just remove the trs=true code path.

ye-luo:
@Paul-St-Young Just do the deletion as I suggested since that part of code is anyway not useful and should not be accessed any more. Since we won't have time to check the whole real code path. Just get the obvious bug removed in this PR.

Paul-St-Young:
@ye-luo the reason I am hesitate to remove the `trs=true` path is that I found it necessary to correct the real code path.

More specifically, I can make the QMC_COMPLEX=0 code path pass all tests if:
1. I keep the `trs=true` path as it is in this PR and
2. force `UseRealOrbitals=false`

If I force `trs=false`, then the QMC_COMPLEX=0 code path fails.

ye-luo:
I read the code again. It seem the trs=true is the right code path for C2R case. So remove the trs=false code path, it should pass the tests in the case you listed above.

Paul-St-Young:
@ye-luo yes `trs=false` fixed all tests. Thanks a lot for looking this over!

I will eliminate the `trs=true` code path and update this PR.

Paul-St-Young:
Oops, got it backwards in the above comment. `trs=true` fixed all tests

ye-luo:
Please update the branch and merge after PR #533.

PR #543, opened 2017-12-07
Modified configuration files for titan, eos, and oic5

kylanpaa:
The build scripts for titan, eos, and oic5 have been updated in the config folder. On titan the script compiles real and complex codes for AoS, SoA and GPU. On eos and oic5 the scripts compile real and complex codes for AoS and SoA.

### DIFF
--- a/config/build_olcf_eos.sh
+++ b/config/build_olcf_eos.sh
@@ -11,7 +11,7 @@
 ## Settings should be consistent with test scripts e.g.       ##
 ##  ./tests/test_automation/nightly_ornl_olcf_eos.job         ##
 ##                                                            ##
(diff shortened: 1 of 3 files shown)


ye-luo:
Please change the build_XXXX based on largest to smallest divergence rule.
build_machine_cpu/gpu_real/complex_AoS/SoA

prckent:
My preference is that only the soa builds have an soa-like tag applied. aos builds (== old code) should not. This is consistent with the oxygen build naming. I prefer that we do NOT label *both* the aos and soa builds with the data layout because they are too similar in name and confusions may result. i.e. Use qmcpack_cpu_real, qmcpack_cpu_soa_real.

prckent:
OK to test

Issue #540, opened 2017-12-05
restrict types for accumulators and iszero()


grahamlopez:
Newer versions of glibc provide an external `iszero()`, which causes the estimators unit test to (correctly) fail on integers. This test was previously passing because the internal `iszero()` at `src/config/stdlib/limits.h:19:inline bool iszero(T a)` happily accepts integers, but doesn't do the right thing with them.

Proposed solution is to restrict the templates on the accumulators and the internal `iszero()` to only accept float types, to match the behavior of glibc (see [the glibc manual](https://www.gnu.org/software/libc/manual/html_node/Floating-Point-Classes.html#index-iszero))

A tiny detail before I submit a patch: do we prefer to use `enable_if` or SFINAE to restrict the template types? The only preference I can think of is that `enable_if` with give compile-time errors that will likely be easier to understand than the link-time SFINAE errors.

ye-luo:
Is this related to issue #429 ?

grahamlopez:
No, I think the checking is doing the right thing. In my situation, CMake correctly sees that glibc's `iszero()` is available, and that's what gets called. The problem is that the estimator unit test for the estimators is written to accommodate `iszero()` at `src/config/stdlib/limits.h:19` which does the wrong thing when passed an integer. This unit test fails when using glibc's `iszero()`.

Since glibc's `iszero()` implementation doesn't even support integers, the proposed fix is to disallow integers in our `iszero()` and accumulators, and to disable the integer version of the estimators unit test.

ye-luo:
@markdewing Can we just remove `accumulator basic int` in the unit test? I don't think the main code uses int as the template type of accumulator.

markdewing:
The accumulator basic int test should be removed.  
It would be nice to have, but not essential, to add the type restrictions on the accumulator and is_zero.  Mostly to avoid any surprises in the future if anyone does try to use them with an int type.

ye-luo:
OK. Since the glibc has been engaged for iszero. The removal must be conducted before v3.3 release.

grahamlopez:
resolved via #545

PR #545, opened 2017-12-07
type_trait guards for accumulator_set/iszero()

grahamlopez:
This is an incomplete PR to address issue #540 that I need another pair of eyes on. The type_trait guards here should cause a build error when src/Estimators/tests/test_accumulator.cpp:30 is encountered, but everything still seems to build fine for me (except that this integer test still fails). 

Once I see the `accumulator_set<int> a1;` cause a failure, then I'll follow up with another commit to completely remove that test.

The following minimum example properly doesn't compile, so I can't tell what I'm doing wrong in this patch. Probably something silly.

```
#include <type_traits>  
                                                                                  
template<typename T, typename = std::enable_if_t<std::is_floating_point<T>::value>>
struct A {       
  void afoo(T t)  {   std::cout << "in afoo()" << '\n';  }
};

int main() {         
   int i; 
   double d;     
   A<int> a;   
   a.afoo(i);  
}  
```

### DIFF
--- a/src/Estimators/accumulators.h
+++ b/src/Estimators/accumulators.h
@@ -32,7 +32,8 @@
  *
  * To simplify i/o, the values are storged in contens
  */
-template<typename T, typename = std::enable_if<std::is_floating_point<T>::value>>
+template<typename T,
+  typename = typename std::enable_if<std::is_floating_point<T>::value>::type>
 struct accumulator_set
 {
   typedef T value_type;
--- a/src/Estimators/accumulators.h
+++ b/src/Estimators/accumulators.h
@@ -10,8 +10,8 @@
 //
 // File created by: Jeongnim Kim, jeongnim.kim@gmail.com, University of Illinois at Urbana-Champaign
 //////////////////////////////////////////////////////////////////////////////////////
-    
-    
+
+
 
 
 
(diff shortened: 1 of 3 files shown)


markdewing:
Looking at some examples online, this seems to work for iszero
```
template<typename T, class = typename std::enable_if<std::is_floating_point<T>::value>::type>
inline bool iszero(T a)
```

It adds 'class = typename ...' to the beginning, and references ::type of enable_if.
I don't know enough to know why this works.

grahamlopez:
Thanks. That's pretty much equivalent to what I'm doing. Briefly, your example is defining a typename based on the boolean result, and then setting that to a template parameter (class =). My way is just setting the template parameter directly to the boolean result. They are both correct as far as I can tell in my reading and independent test code snippets.

I dug a little more, and changing any declaration of `accumulator_set<T>` to `accumulator_set<int> in **non-test** code resulted in the desired build failures with the patch as it stands here. So I think I'm misunderstanding something about how Catch works (?)

markdewing:
They don't seem to quite be equivalent, at least with gcc 4.8.4.   Using the wordier version results in an error when compiling the accumulator test, but the shorter version doesn't.

grahamlopez:
You are absolutely right, @markdewing. I was in too much of a hurry and didn't notice I was using `enable_if_t<>`. This is c++14. Since we're on c++11 right now, we've got to be a little bit more verbose. Thanks for the sanity check. 

Still not sure why I saw different behavior for the test code, but I'm willing to let that one lie for now. I've updated the PR, and it should be ready to go.

Issue #544, opened 2017-12-07
drift implementation potentially deviate from intention


Paul-St-Young:
The scaling factor of the drift vector, used in importance sampling, may deviate from the intended implementation of "eq. (34) *tau in Umrigar, JCP 99, 2865 (1993)". The deviation happens whenever the wavefunction gradient has a non-zero complex component. Example code snippet from `DriftOperators.h`:

```c++
template<class T, class TG, unsigned D>
inline T getDriftScale(T tau, const ParticleAttrib<TinyVector<std::complex<TG>,D> >& ga)
{ 
  T vsq=Dot(ga,ga);
  return (vsq<std::numeric_limits<T>::epsilon())? tau:((-1.0+std::sqrt(1.0+2.0*tau*vsq))/vsq);
}
```
In the above, `vsq` may become negative if `ga` has a significant complex component. I think the `vsq` in Umrigar, JCP 99, 2865 (1993) is intended to be a non-negative real number. Thus implantation deviates from intention.

A realistic occurrence of the aforementioned situation can be reproduced using the attached nexus script (solidh.zip). In that example, the symptom of the bug is as follows. The variance in  the "vmc" folder will be > 2 ha^2 near the configuration I have chosen (using a random seed). This is a lot higher than the optimized variance ~0.4 ha^2 in the "opt" folder.

The cause of the bug is two-fold:
1.  The initial random arrangement of electrons is near a node of the many-body wavefunction causing large complex component in the wavefunction gradient.
2. drift vector implementation deviates from intention at this configuration.

Cause 1 can be verified by running a wavefunction test (wftest.zip) at the initial configuration. The resultant laplacian values are shown in the figure below. 
![laplacian](https://user-images.githubusercontent.com/8060441/33735396-c38f88f6-db54-11e7-9d16-0423a5724e4b.png)

The same symptom was observed in #514, albeit the cause was mis-assigned to `rcut`.

[solidh.zip](https://github.com/QMCPACK/qmcpack/files/1540429/solidh.zip)
[wftest.zip](https://github.com/QMCPACK/qmcpack/files/1540375/wftest.zip)

rcclay:
After some chats, here's what's happening:
VMC/DMC/RMC should be using |Psi_T| as the guiding function, which means that the wavefunction force used to compute the drift should be calculated as the real part of grad( log(Psi_T) ).  Umrigar drift scaling should be computed with respect to this object, basically by computing the square magnitude of the REAL wavefunction force.  

[...]

PR #546, opened 2017-12-08
Fix Umrigar drift diffusion for complex wavefunctions

rcclay:
This is a minimally invasive fix of issue #544. It redefines getScaledDrift to make the wavefunction force real before computing the drift scaling.  For DMCNonLocalMoves, I introduce a temporary PosType called "real_grad_iat", which is used to hold the real wavefunction force while getDriftScale is called.  

PLEASE DO NOT MERGE UNTIL CUDA IMPLEMENTATION IS COMPLETED
This is done in exact analogy to DMCNonLocalMoves.  It just took longer than expected to get a working GPU copy for debugging.  Will push when tested.


ye-luo:
@rcclay please revert any change in DMCNonLocalUpdate.cpp.
DMCUpdatePbyPFast.cpp is the pbyp move file.
DMCUpdateAll.cpp is the all-e move file.
DMC_CUDA.cpp DMCcuda::run is the CUDA code path.

Paul-St-Young:
@rcclay Is it possible to remove the deprecated functions instead of adding aborts?

The deprecated functions have been carried in git history for a while, so it should be easy for us to dig them back up if needed. It would be great to make `DriftOperators.h` more readable and minimize the chance of programming error in the future.

rcclay:
OK.  That should do it.  The short tests for CPU+Complex will be done in a few minutes.  It would be nice to verify that the numbers coming out of the GPU versions are still OK.  It will build, but there's a gremlin in the GPU build on my workstation unrelated to this commit.  

@Paul-St-Young They can almost assuredly be removed.  I'm just being overly cautious at the moment.  If it's OK with everyone, I'll remove them.

rcclay:
Status update:  with recent commits, ctest -R short is passing everything (except for usual suspects like backflow opt in the real code path) with the real & complex CPU versions, as well as the real & complex GPU versions.

Paul-St-Young:
I added some unit tests for `DriftOperators.h`. The coverage is not great right now. I am only testing setScaledDrift() and setScaledDriftPbyPandNodeCorr(), which are used by the pbyp and wbyw drivers, respectively.

More tests should be added for other drivers.

All tests are passing with @rcclay's fixes.

ye-luo:
I cleaned the GPU driver to minimize ifdefs. Now GPU also uses the same getScaledDrift as CPU.
All the size inconsistent version (not applied per particle) of scaling is risky.
DMC/VMC allp and pbyp are fine. RMC drivers are not safe in this regard.

rcclay:
@ye-luo The changes look good.  

RMC uses the same getScaledDrift as VMC and DMC pbyp, so I would expect it to be fine.  If there's a size consistency issue with computing vsq on the all-electron gradient versus the particle-by-particle gradients, then I think RMC all-particle moves will suffer from size consistency issues due to its reliance on setScaledDrift.  This should be corrected in a later pull request, since its a slightly different issue.

src/QMCDrivers/DriftOperators.h
```diff
@@ -157,22 +139,45 @@ inline T setScaledDriftPbyPandNodeCorr(T tau,
...
-    T sc=(vsq<std::numeric_limits<T>::epsilon())? tau:((-1.0+std::sqrt(1.0+2.0*tau*vsq))/vsq);
+    // effectively set the Umrigar "a" parameter to particle mass
+    T tau_over_mass = tau_au*massinv[iat]; // !!!! assume timestep is scaled by mass
+    // save real part of wf log derivative in drift
+    convert(qf[iat],drift[iat]);
+    vsq = dot(drift[iat],drift[iat]);
+    // calculate drift scalar "sc" of Umrigar, JCP 99, 2865 (1993); eq. (34) * tau
+    // use naive drift if vsq may cause numerical instability in the denominator
+    vsq = (vsq < std::numeric_limits<T>::epsilon()) ? tau_over_mass : vsq;
+    sc  = (-1.0+std::sqrt(1.0+2.0*tau_over_mass*vsq))/vsq;
```
ye-luo: @Paul-St-Young 
line 175, 176: I don't see a reason to rewrite them since the result is basically the same.

Paul-St-Young: @ye-luo do you mean the comment "The naive drift diverges at a node..."?

src/QMCDrivers/DriftOperators.h
```diff
@@ -157,22 +139,45 @@ inline T setScaledDriftPbyPandNodeCorr(T tau,
...
+ * The naive drift diverges at a node, causing persistent configurations.
+ * The norm of the drift vector should be limited in two ways:
+ *  1. Umrigar: suppress drift divergence to mimic wf divergence
+ *  2. Ceperley: limit max drift rate to diffusion rate -> set Umrigar "a" parameter
+ * The choice of drift vector does not affect VMC correctness
+ *  so long as the proposal probabilities are correctly calculated.
+ * The choice of drift vector changes the DMC Green's function. BE CAREFUL!
+ *
+ * T should be either float or double
+ * T1 may be real or complex
```
ye-luo: T1 may be real or complex of float or double.

src/QMCDrivers/DriftOperators.h
ye-luo: I mean the source line not the comment.

ye-luo:
@rcclay In RMC, I saw some assignDrift and setScaledDrift based on all-electron gradient. Yes, fix in a separate PR.

src/QMCDrivers/DriftOperators.h
Paul-St-Young: @ye-luo indeed the functionality is unchanged. I added the `sc` variable so that the meaning of `vsq` is definitively "the gradient squared". I believe the extra `sc` variable improves the readability of the code. The separation of `sc` and `vsq` also elucidates the purpose of the ternary.

src/QMCDrivers/DriftOperators.h
Paul-St-Young: agreed. I will edit the comment.

Paul-St-Young:
sorry I kept pushing the wrong button. It was not my intention to start the previous two reviews.

src/QMCDrivers/DriftOperators.h
Paul-St-Young: @ye-luo I'm sorry, I was confused. Yes the two lines you mentioned should be merged. It has been a long day ...

I will change the code according to your suggestions ASAP

ye-luo: The problem is that you only changed one routine but all the others are still written in the old way. I prefer them consistent.

Paul-St-Young: @ye-luo yes, I agree consistency is important. I have updated all the routines in my last commit. Please take a look.

ye-luo:
ok to test

Paul-St-Young:
@ye-luo the changes look good.

I still run into persistent configurations (rarely) in all-particle driver. I think the problem is with warmup, not the drift. The all-particle driver's Initialization issues should probably be addressed in another PR.

ye-luo:
@rcclay @Paul-St-Young anything else to add on this PR?

rcclay:
@ye-luo Nope.  Looks good.  Ready to merge whenever.

prckent:
I'll merge this once the CI has passed unless anyone would prefer we hold off.

Paul-St-Young:
@ye-luo @prckent Can we remove the unused routines? (the ones with APP_ABORT)

Paul-St-Young:
I cleaned `DriftOperators.h` and reran the unit and short tests (CPU) just in case. 

No failure beyond the usual suspects in the real code path:

The following tests FAILED:
  222 - short-bccH_3x3x3_ae-gamma-deriv-1-1 (Failed)
  223 - short-bccH_3x3x3_ae-gamma-deriv-1-1-check (Failed)
  224 - short-bccH_3x3x3_ae-not_orth-deriv-1-1 (Failed)
  225 - short-bccH_3x3x3_ae-not_orth-deriv-1-1-check (Failed)
  230 - short-bccH_2x2x2_ae-gamma-deriv-1-1 (Failed)
  231 - short-bccH_2x2x2_ae-gamma-deriv-1-1-check (Failed)
  282 - short-bccH_1x1x1_ae-csvmc-all-nodrift_sdj-16-1-deeenergy_AB (Failed)

prckent:
Thanks Paul. Very happy to reduce the source line count.

prckent:
Thanks to all!

PR #547, opened 2017-12-08
Break up the cartesian tensor test.

markdewing:
Some compilers are having trouble with the single large test in test_cartesian_tensor.
Break it into multiple smaller test cases.

For hessian and third derivatives, the laplacian is not computed.  Remove the checks in those cases (it worked before because the previously computed laplacian values were still present)


ye-luo:
@prckent Just merge. We will get that fixed together on several tests in a future PR.

markdewing:
This PR doesn't fix the issue - I will need to test on more compilers.
(But this change is a necessary first step)

PR #548, opened 2017-12-08
Remove unused DMC driver sources

prckent:
These files 

- Are unused
- Do not appear to implement any special/needed functionality that we should preserve
- Make it confusing to see where the production DMC algorithms are actually implemented (#546)

For easy reference the following are removed:

```
DMC.h
DMC.cpp
DMCNonLocalUpdate.h
DMCNonLocalUpdate.cpp
DMCOMP.master.cpp
DMCOMPOPT.h
DMCOMPOPT.cpp
DMCPeta.h  
DMCPeta.cpp  
DMCWOS.h
DMCWOS.cpp
```


prckent:
@ye-luo Cool. Size consistent t-moves?

ye-luo:
@prckent Yes. it is on my fork Tmove branch.

PR #550, opened 2017-12-08
nexus: fix bad default for pw2qmcpack input

jtkrogel:
Very small PR to turn off write_psir by default.  Generally this has been provided by the user as "False", overriding the poor/outdated default.


markdewing:
okay to test

jtkrogel:
@ye-luo Out of curiosity, why the "merge develop into nxdev"?  

I generally pull with rebase from develop immediately prior to changes and then PR.  Is there something more I should be doing?

ye-luo:
Not your problem. Approvers handle this. Since I merged another PR (build script update), your PR is no more up-to-date with the develop. I need to update your branch before merging.

PR #554, opened 2017-12-11
Check clang version for undefined-var-template.

markdewing:
The -Wno-undefined-var-template option was added in clang 3.9.
Add a version check so this flag is not added for older versions of clang.

Clang 3.8 and earlier issue a warning about the flag which then causes the C++ 11 check to fail.


ye-luo:
Current c++11 checking is not robust. We need to improve it by asking CMake to take over.

prckent:
@ye-luo Agree about relying more on CMake checking. Happily no-one else has hit this bug yet, or at least they did not complain about it. It is a step in the correct direction.

Issue #425, opened 2017-10-20
CMake Deprecation Warning


ye-luo:
cmake 3.9.1 complains
```
CMake Deprecation Warning at CMake/macros.cmake:2 (CMAKE_POLICY):
  The OLD behavior for policy CMP0026 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
Call Stack (most recent call first):
  tests/estimator/CMakeLists.txt:23 (INCLUDE)
```

ye-luo:
Resolved.

PR #556, opened 2017-12-11
Removed unused CMake testing functions.

markdewing:
Also clears up the cmake deprecation warning about policy CMP0026 (issue #425)
(warning present since cmake 3.9). 

I don't know if anyone plans to use these functions, but they are currently unused.


prckent:
These functions were setup based on of of Mark B.'s templates back when ctest was first implemented. They are safe to remove.

Issue #504, opened 2017-11-16
Collectables weighting and "value_squared"


jtkrogel:
"Collectables" type estimators currently write both "values" and "values_squared" to the stat.h5 files.  Currently the only way to obtain correctly weighted mean values in DMC is to apply the weight by hand during Collectables accumulation.  This results in "values_squared" incorrectly containing a squared weight.

Two options: 1) delete value_squared and save 1/2 of the stat.h5 disk space (often GB's), 2) change the accumulation mechanism so that appropriate weights are applied to values and values_squared for each estimator (even on values, weights may be applied patchily).  

The potential value of keeping value_squared around is to have obtain better estimates of the variance of each quantity than can be obtained from the blocked sample variance.

prckent:
Two issues: correctness and affordability.

Since the values_squared is obviously wrong currently, can we safely conclude that no-one is using it?

jtkrogel:
I am not using it at present; I doubt if anyone else is.  It was really an attempt to use it that revealed that it is wrong, though it may be wrong in more ways than just the weighting.

@Paul-St-Young do you want to comment on what you are seeing w.r.t. the S(k) estimators?

ye-luo:
If we need to report an error bar on a quantity measured by estimators the weighted value_square is needed, I think we need to have the estimator manager properly handling the properly weighted value square. Dumping or not can be optional.

Paul-St-Young:
I recently experimented with obtaining `S(k)=<rho(k)rho(-k)>` from `rho(k)=a+ib`. Since `rho(-k)=a-ib`, `S(k)=<a^2>+<b^2>`. Namely, by recording `<a>` and `<b>` in value, value_squared will hold `<a^2>` and `<b^2>`. Thus `S(k)` can be reconstructed from `rho(k)`. This strategy is enabled by the SkAllEstimator dumping to hdf5 file.

I have created an example adapted from `tests/solids/bccH_3x3x3_ae/gamma_deriv.xml`. In particular, I have removed `<backflow>` and switched to the `pbyp` driver. I ran the example using the most recent develop branch (see `out`) and the example runs in a second, so please feel free to rerun.

[...]

PR #560, opened 2017-12-12
disable value_squared in h5

Paul-St-Young:
Addressing #504, do not print wrong values in `value_squared`. This will roughly cut the size of the stat.h5 file in half. User can code `<a^2>` as a separate column in the observable that calculates `<a>`.

This is NOT a full removal of `value_squared`. Other candidates to consider for removal:
1. `EstimatorManager::SquaredAverageCache` is mostly unused now. However the first entry for local energy is still used to calculate variance. 
2. nexus has some dependencies on `value_squared` in "qmcpack_quantity_analyzers.py". I do not have examples for most of the functions that use `value_squared`. @jtkrogel please take a look.
I did fix `ScalarsHDFAnalyzer.analyze_local`, because it is used to analyze optimized wavefunction.


jtkrogel:
Nexus will need other modifications.  A fair amount of the touches to value_squared are done in the context of twist averaging both collectables and scalar data.  I will download your QMCPACK modifications and make the required changes to Nexus.  Cases like this add priority to having more explicit tests for Nexus.

prckent:
@jtkrogel Do you plan to make a separate PR in the next day or so?

jtkrogel:
I can make a separate PR.  It should probably go in before this one to avoid breakage.

prckent:
Before merging I would like to summarize the new situation and correct if needed:

- These changes disable the I/O of (incorrect) squared values
- value^2 are still tracked in memory via EstimatorManageBase.h/.cpp
- The internal weighting of value^2 is wrong/inconsistent and needs to be rethought/reimplemented.

(Will edit the above based on comments received)

grahamlopez:
okay to test

jtkrogel:
Working on companion PR for Nexus now.

jtkrogel:
@Paul-St-Young please revert commits relating to Nexus in this PR.

jtkrogel:
See companion PR #567

PR #562, opened 2017-12-12
Unit test for Bspline one-body Jastrow

markdewing:
Mostly just tests the underlying Bspline functor. 
Checks at the level of calling 'evaluate' need to be added.

### DIFF
--- a/src/QMCWaveFunctions/tests/test_bspline_jastrow.cpp
+++ b/src/QMCWaveFunctions/tests/test_bspline_jastrow.cpp
@@ -23,11 +23,13 @@
 #include "QMCWaveFunctions/OrbitalBase.h"
 #include "QMCWaveFunctions/TrialWaveFunction.h"
 #include "QMCWaveFunctions/Jastrow/TwoBodyJastrowOrbital.h"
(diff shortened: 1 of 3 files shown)


src/QMCWaveFunctions/tests/test_bspline_jastrow.cpp
```diff
@@ -214,6 +216,170 @@ const char *particles = \
...
+  for (int i = 0; i < N; i++) {
+    OrbitalBase::RealType dv = 0.0;
+    OrbitalBase::RealType ddv = 0.0;
+    OrbitalBase::RealType val = bf->evaluate(Vals[i].r,dv,ddv);
+    REQUIRE(Vals[i].u == Approx(val));
+    REQUIRE(Vals[i].du == Approx(dv));
+    REQUIRE(Vals[i].ddu == Approx(ddv));
+  }
+
+#if 0
```
prckent: I think we need a policy for #ifdef regions. I would prefer that we distinguish functional but protected code behind named #ifs to distinguish them from everything else. e.g. here #ifdef DEBUG_PRINT. The J2 test code also has this issue.

markdewing: Yes, good idea.

PR #563, opened 2017-12-13
Enable hybrid (APW-like) representation testing on GPU

prckent:
In view of recent developments regarding the effectiveness of this basis, this functionality has to work again, so we begin testing it.


ye-luo:
What is the current testing behavior if enabled on CUDA builds? Run through by ignoring the input tag or crash?

prckent:
See #564 . It aborts "correctly". I am not sure if it is preferred that we fix the AoS-GPU code or finally port the GPU code over to SoA. What do you think? - put comments in #564

ye-luo:
These input files are intended only for CPU SoA implementation of hybrid representation not the old GPU implementation. Unless enabling these tests on GPU is to remind us porting this feature to the GPU code, I don't feel the need for this PR.

prckent:
I had not realized that the inputs were different, but these tests are indeed intended to remind us that the hybrid representation is meant to work everywhere. The same procedure as we followed for gaussians with SoA etc. 

Hybrid representation is certainly core functionality that should work everywhere, otherwise we wouldn't be investing time in it.

ye-luo:
Then, I need to add a stop and printout indicating new hybrid is currently unavailable for GPU.
@prckent do you prefer me to check the fail status and let the test pass or just let the test fail?

prckent:
No one else is going to run the hybrid. OK to have this crash and fail. I hope that the "new" hybrid code is not so different from the "old" hybrid code that the GPU code can be activated easily.

ye-luo:
test this please

ye-luo:
The first CI failed in GPU build. Triggered a second CI to see if it is just a glitch.

tests/solids/LiH_solid_1x1x1_pp/CMakeLists.txt
```diff
@@ -50,7 +50,7 @@
                     0 LIH_X_SCALARS # VMC
                     )
 
-IF(ENABLE_SOA)
+IF(ENABLE_SOA OR QMC_CUDA)
```
ye-luo: The logic is still strange here. If we decide to port the new hybrid to CPU, then ENABLE_SOA is always needed and it should be OR.

tests/solids/LiH_solid_1x1x1_pp/CMakeLists.txt
```diff
@@ -60,9 +60,9 @@ IF(ENABLE_SOA)
                     0 LIH_X_SCALARS # VMC
                     )
 ELSE()
-  MESSAGE("Skipping LiH_solid_1x1x1_hybridrep_pp-x-vmc_hf_noj test because SOA is not enabled (ENABLE_SOA=0)")
+  MESSAGE("Skipping LiH_solid_1x1x1_hybridrep_pp-x-vmc_hf_noj test because SOA and/or GPU is not enabled (ENABLE_SOA=0 or QMC_CUDA=0)")
```
ye-luo: The logic in the message seems not consistent with the if statement.

tests/solids/LiH_solid_1x1x1_pp/CMakeLists.txt
prckent: Since I want to see that the  hybrid rep code is not working at all with CUDA, the logic is correct. I wrote these when I thought the hybrid rep code had not been rewritten  and the CUDA implementation broken, but that does not change the logic.

tests/solids/LiH_solid_1x1x1_pp/CMakeLists.txt
prckent: The logic is fine.

ye-luo: For example, when I build SoA without GPU. ENABLE_SOA=1 and QMC_CUDA=0. This sounds like the case skipping the test as the message described.

prckent: Got it, will update.

PR #567, opened 2017-12-14
nexus: fix qmcpack analyzer to handle stat.h5 without value_squared

jtkrogel:
This is a companion PR to #560 for Nexus handling of stat.h5 data with the value_squared field removed.  

To cover the full range of functionality runs were performed with v3.1.1 and #560 including density, spin density, energy density, and MPC estimators with and without twist averaging and analyzed with Nexus.

The updated code is backwards compatible with old data (value and value_squared present) and also works with new data (only value present).

### DIFF
--- a/nexus/library/hdfreader.py
+++ b/nexus/library/hdfreader.py
@@ -144,6 +144,15 @@ class HDFgroup(DevBase):
     #end def read_arrays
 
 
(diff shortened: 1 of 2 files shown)


ye-luo:
ok to test

PR #568, opened 2017-12-14
Clean up random number and unit cell output

markdewing:
The random number info is not present in low-verbosity output - it doesn't seem useful to the user.

The handling of wire/slab geometries resulted in double output of the unit cell info.
Now only do this printing for those boundary conditions.  A future enhancement would be to also detect whether the vacuum multiplier setting is different from 1.0, but that will require testing to ensure it works.

The message in ParticleSetPool about creating or overwriting a global supercell has been moved to debug-level output.  Unless there's something the user needs to know from this information?

Partially addresses #35

Low verbosity output:
```
                    QMCPACK 3.2.0

  (c) Copyright 2003-  QMCPACK developers

  Git branch: clean_rng_output
  Last git commit: c18957440c91c25dd2496d28d50ee082d2428d49-dirty
  Last commit date: Tue Dec 12 16:08:51 2017 -0500

  MPI Nodes            = 1
  MPI Nodes per group  = 1
  MPI Group ID         = 0
  OMP_NUM_THREADS      = 16

  Base precision      = double
  Full precision      = double

  Input XML = qmc_short.in.xml

  Project = qmc_short
  date    = 2017-12-13 23:27:12 CST
  host    = sven

 Unit Cell
 ---------
  Lattice:          6.7463223         6.7463223                 0
                            0        3.37316115        3.37316115
                   3.37316115                 0        3.37316115

  Boundary Conditions:  p  p  p

```

High verbosity output:
```
  Input file(s): qmc_short.in.xml

                    QMCPACK 3.2.0

  (c) Copyright 2003-  QMCPACK developers

  Git branch: clean_rng_output
  Last git commit: c18957440c91c25dd2496d28d50ee082d2428d49-dirty
  Last commit date: Tue Dec 12 16:08:51 2017 -0500
  Global options
  async_swap=0 : using blocking send/recv for walker swaps

  MPI Nodes            = 1
  MPI Nodes per group  = 1
  MPI Group ID         = 0
  OMP_NUM_THREADS      = 16

  Base precision      = double
  Full precision      = double

  Input XML = qmc_short.in.xml

  Project = qmc_short
  date    = 2017-12-13 23:31:57 CST
  host    = sven

  DO NOT READ DENSITY
 Random Number
 -------------
  Offset for the random number seeds based on time 205
  Random number offset = 205  seeds = 1279-1409

        1279        1283        1289        1291        1297        1301        1303        1307
        1319        1321        1327        1361        1367        1373        1381        1399
        1409        1423
  Random seeds Node = 0:        1283        1289        1291        1297        1301        1303        1307        1319        1321        1327        1361        1367        1373        1381        1399        1409

 Lattice
 -------
  Simulation cell radius = 1.947495
  Wigner-Seitz    radius = 2.385185

 Unit Cell
 ---------
  Lattice:       6.7463223000      6.7463223000      0.0000000000
                 0.0000000000      3.3731611500      3.3731611500
                 3.3731611500      0.0000000000      3.3731611500

  Boundary Conditions:  p  p  p

  Volume (A^3) = 153.5222266112

  Reciprocal vectors without 2*pi.
    g_1 =       0.0741144549      0.0741144549     -0.0741144549
    g_2 =      -0.1482289098      0.1482289098      0.1482289098
    g_3 =       0.1482289098     -0.1482289098      0.1482289098

  Metric tensor in real-space.
    h_1 = 91.0257291510 22.7564322877 22.7564322877
    h_2 = 22.7564322877 22.7564322877 11.3782161439
    h_3 = 22.7564322877 11.3782161439 22.7564322877

  Metric tensor in g-space.
    h_1 = 0.6505592096 -0.4337061397 -0.4337061397
    h_2 = -0.4337061397 2.6022368383 -0.8674122794
    h_3 = -0.4337061397 -0.8674122794 2.6022368383

```


ye-luo:
1) There are two sections of "nprocs<4" printing RNG seeds and they are annoying. I think we just remove.
2) Always print the RNG info since there are only 2 lines and it gives user information whether their run takes the seed from input or generated based on time. It is important for reproducibility and cross-validation.

src/OhmmsApp/RandomNumberControl.cpp
```diff
@@ -117,7 +117,7 @@ void RandomNumberControl::make_children()
     o << "  Random seeds Node = " << rank << ":";
```
ye-luo: First nprocs<4

src/OhmmsApp/RandomNumberControl.cpp
```diff
@@ -215,6 +217,7 @@ bool RandomNumberControl::put(xmlNodePtr cur)
     Random.init(pid,nprocs,mySeeds[pid],Offset+pid);
     app_log() << "  Random number offset = " << Offset
               <<  "  seeds = " << mySeeds[0] <<"-" << mySeeds[nprocs*omp_get_max_threads()] <<  std::endl;
+    app_log() << std::endl;
     if(nprocs<4)
```
ye-luo: second nprocs<4

prckent:
- [low,high] Add units (A) to the lattice vector print out
- [high] Add units to the simulation cell and Wigner Seitz cell radii
- [low] It would be good to print *how* the random numbers were initialized, clock or file. A future improvement for random number handling would be to write the initial random number state to an XML file in the same way that we write the final XML. There are no docs on how to restart the random number generator from what is printed currently. -> Separate issue to be addressed later?

prckent:
- [high] Why do we print "DO NOT READ DENSITY"? Why not skip this and only print something if we actually take action? "Reading density XYZ for ABC"

ye-luo:
MPC needs to load the density from h5.
The actual loading happens when the orbitals are initialized from h5 before the Hamiltonian is constructed.
So the code first survey the xml and identify whether density is needed or not at the very beginning.

Paul-St-Young:
Now that output is being worked on, may I make two trivial requests?
1. Make `Volume` unit depend on lattice unit.
I think the unit of `Volume` is dependent on the units of the simulation cell. The current output makes it seem like `Volume (A^3)` is always in angstroms.
2. Change `Wigner-Seitz    radius` to `Wigner-Seitz cell radius`
The Wikipedia definition of [Wigner-Seitz radius](https://en.wikipedia.org/wiki/Wigner%E2%80%93Seitz_radius) is the density parameter `r_s`, namely the radius of the sphere of volume per particle. What we actually print is: the radius of the inscribed sphere inside the real-space Wigner-Seitz cell.

prckent:
Thanks @Paul-St-Young . Highlights why it is worth making these updates. Lets print the volume both ways and always with accurate units.

Re: Densities. Thanks @ye-luo . -> "Did not find density in XYZ.h5 to evaluate MPC interaction". Where practical, the output should explain what and why things are done and not rely on additional knowledge.

src/OhmmsApp/RandomNumberControl.cpp
markdewing: This one seems okay to remove.  The app_log output is only printed for rank 0 anyway.

src/OhmmsApp/RandomNumberControl.cpp
markdewing: This test seems more difficult to remove.  Without it, the list of random seeds keeps growing with number of processors.  Eventually it seems like it will be too large to be useful.

One the initial seed offset is known, all the rest of processor and thread seeds should be deterministic?

src/OhmmsApp/RandomNumberControl.cpp
ye-luo: I don't mean remove just the if line but the whole if section.

src/OhmmsApp/RandomNumberControl.cpp
ye-luo: I also remove completely. Bad on KNL with a large thread count.
Once a seed is given in the input. It is deterministic.

markdewing:
The units on the lattice seem like it's always in atomic units.  Many of the examples and tests have a units="bohr" attribute, but this attribute doesn't seem to be handled in ParticleLayoutIO.cpp::put

Is the "A" in the output supposed to mean Angstroms or Atomic units?   It doesn't look like there's any unit  conversions on the lattice.

prckent:
A=Angstrom, always. a.u.=atomic units.

I thought we were a.u. internally and made conversions on read/write.

prckent:
@Paul-St-Young Are you OK with these changes?

Paul-St-Young:
@prckent sorry I didn't see this. Yes the changes look good

Paul-St-Young:
question: if lattice unit is set to `A` in the xml, is there an internal unit conversion to `bohr`?

markdewing:
There does not appear to be any unit conversion for the lattice.  I couldn't find any part of the code to read the 'units' attribute.

There is some unit conversion code in Utilities/unit_conversion.cpp, so it might not be too hard to actually implement.

Paul-St-Young:
If there is not unit conversion, then I think it is best to abort if lattice unit is anything but `bohr`. 
I think it is better to inform, rather than ignore, the user.

markdewing:
That's good idea.  I'll add it in a future PR (trying to keep formatting changes separate from other changes as much as possible)

Paul-St-Young:
@markdewing great! The current formatting changes look good.

Another thought: unit conversion, if implemented, needs to be consistent throughout the code.
I am not sure how thorough one needs to be. At least, lattice and particle positions must have the same unit. Perhaps there are estimator or Hamiltonian elements that also rely on spatial units.

markdewing:
I agree that unit conversion needs to be handled carefully and consistently.

prckent:
I will merge this as soon (and assuming) the CI passes again. Improvements identified can be subject to future issues, PRs.

Issue #578, opened 2017-12-15
Integration test: quasi-2D


jtkrogel:
Combined test of vacuum option and quasi-2D ~adaptive B-spline mesh.

ye-luo:
What is adaptive B-spline mesh? We will probably deprecate it.

prckent:
These should be split.

vacuum options are apparently useful and used, but need to be eventually moved inside the input xml specification.

adaptive mesh is worth testing provided it is not much work to setup. Indeed it might be deprecated but we should not go out of our way to do so. It has some champions but considering that there is no published data and we already have multiple bets on different basis functions to save memory it is very much a research fall back.

jtkrogel:
I can add two tests instead of a combined single one.

lshulen:
I have found truncate useful in a number of instances.  @ye-luo why do you say we will deprecate it?  Are you thinking the memory savings from hybrid are enough to offset it's usefulness?

ye-luo:
@jtkrogel Could you clarify what exact option you mean "adaptive B-spline mesh"?

ye-luo:
@lshulen Using hybrid + vacuum option are enough to offset its usefulness. This way also reduces the pain of DFT. The "truncate" two grid scheme has not been updated to SoA.

prckent:
Before taking a decision about future support or deprecation we will examine actual data for problems such as surface catalysis.

jtkrogel:
By "adaptive" I mean the two grid scheme enabled by truncate (+buffer).

lshulen:
Actual data is a bit tricky, but I can give a couple of back of the envelope calculations for reference.  The first case is the aforementioned catalysis.  I'm imagining a slab of material, let's say platinum with a molecule landing on top, let's say CO.  For these purposes, I would imagine needing ~6 layers of Pt, which would come to ~22 bohr deep.  Then add about 4 bohr for the CO.  To prevent interactions with the periodic image in espresso, I would probably add about 20 bohr of vacuum on the top.  I would estimate a buffer layer of 2 bohr would be sufficient, so using truncate, I would expect 30 bohr of a fully dense grid in a box that is ~46 bohr tall.  The fully dense grid costs 8x in memory for a given volume than the coarse one currently used, so the total memory cost without truncate is:
46x
[...]

Issue #542, opened 2017-12-07
What does the vacuum option do, and does anyone use it?


markdewing:
The vacuum setting is used in Particle/ParticleSet.BC.cpp, in createSK and it modifies the lattice (or at least a copy of the lattice) for the slab and wire geometries. 

The question comes up because createSK prints another copy of the lattice information to the output.  It seems like this should only be printed if the boundary conditions and vacuum setting change the lattice.  And I'm wondering what the significance of this is to the user and their calculation?

And as mentioned in #146, the --vacuum option should be moved to the XML input file.

ye-luo:
I think we only need to print the lattice when --vacumm != 0 case.

jngkim:
It is a way to expand the vacuum layer (z) even though PW calculation is
done with a smaller vacuum.

It is better to do the quasi-2d correctly.

prckent:
This is potentially useful functionality for slabs but untested and little used. I think we reexamine if/when we have users of it.

jngkim:
Correct. Not for general use.

prckent:
So it is useful and used today but untested.

@anbenali Could you arrange for someone to create an issue proposing what the vacuum option should look like inside the .xml to aid planning removing this option.

 (@markdewing ) To test this I think we would need a unit test to checking particle positions and periodic boundary conditions?

prckent:
I think this issue has now been adequately addressed: option relocated, docs improved, test added.

Issue #146, opened 2017-03-08
Relocate command line flags to XML input


jtkrogel:
Currently the code accepts several command line flags (e.g. --dryrun, --save_wfs, --async_swap, --vacuum, --noprint) that should be placed in the input file.  It is desirable for the input file to be capable of reflecting the totality of information inputted to the code to give the user a uniform interface to control QMCPACK's behavior, and because input files are more easily saved and disseminated.

ye-luo:
I will replace --save_wfs and put it in the input file.

prckent:
Other than --dryrun, I agree. 

These flags should go in different places though. save_wfs is clearly spline related. async_swap is DMC driver related, but we need to avoid duplicating the input handling in the plethora of DMC drivers. 

Note that none of these flags are currently tested. save_wfs and vacuum definitely should be.

markdewing:
I have opinions on which items should be command line arguments and which should go into the input file, but the harder part is trying to make this intuitive sense more concrete.
Options like 'help' and 'version' should be command line arguments, and don't have much value as input file options.

Much this comes down to usage models in different environments and which way best to support users and minimize human error. I spend a lot of time running directly from the command line, but that's for development. Production use (and ramping to production use) uses more batch scripts and Nexus.

One point with usability - dryrun could go in the input file, but seems like a potential trap - it's too easy to make an input file that doesn't perform the run.

As to reproducibility, writing out the options used in the output file does provide a record of what was used, but may make it harder to correctly reproduce (need to remember to add the command line switches).


Another design goal is separation of concerns - the problem should be specified by the input file, but not more.  Output control (debug, noprint, maybe dryrun, etc) seems like it should not be part of the problem specification.  Although this could be addressed by a separate section in the input file.

Nothing precludes having both a command line option and XML file input option (though these should be limited to prevent confusion, and multiple paths to the same result can be a source of bugs.).  The timer control flags might fall into this category.

Moving some of these to the XML input file may require some thought on parsing the XML (see #147).

There are two sources of command line flags - qmc_common.cpp and QMCApp/qmcapp.cpp.  This should be changed so the command line is processed in one place.

jtkrogel:
I would favor allowing all command line options in the input file (but not necessarily eliminating all command line ones).  Having a single (global?) class (QMCCommon/QMCControl?) handle both a new xml input block and the command line processing could avoid duplication/bug concerns.  

[...]

PR #582, opened 2017-12-16
Move command line --vacuum into xml

ye-luo:
As discussed in #146 #542 #578 
Deprecate --vacuum
In xml, use vac_scale. Since I often mistype 'vacuum', use only 'vac'
```
<simulationcell>
  <parameter name="bconds">
    p p n
  </parameter>
  <parameter name="vac_scale" > 1.2 </parameter>
</simulationcell>
```
Some additional change.
in PPN case,
The old code does `R(2,2)*=Scale`
The PR rescale the whole R(2) vector, namely
```
R(2,0)*=Scale;
R(2,1)*=Scale;
R(2,2)*=Scale;
```
in PNN case,
Both R(1) and R(2) are rescaled instead of only R(1,1) and R(2,2)

This should provide better support for the "N" directions not orthogonal to P direction.

I check the existence of --vacuum and the code is stopped to prevent wasting hours.


ye-luo:
I don't know the historic reason but with the current choice, the rescaling is applied on R(2) not the z axis of R(2). So a pre-factor is more convenient than 3 distances for the vector.
in a wire case, vacuum is applied on R(1) and R(2) so put zaxis in the name is misleading.

ye-luo:
1, abort has been handled properly.
2, input tag name changed to vacuum.
3, unit test added.
4, manual updated.
Ready to go.

markdewing:
Should deprecation be a little more gradual? 
First step would be to add the new keyword handling and print a warning message if the command line option is used (but not stop)
Second step (in a future release) would be to print an error and stop of the command line option is used.

Maybe for this feature the cost of forcing the users to change their input files is not so high (versus the development cost of doing it more gradually)?

manual/simulationcell.tex
```diff
@@ -26,17 +27,18 @@ \section{Specifying the simulation cell}
...
+  <simulationcell>
+    <parameter name="lattice">
+      3.80000000       0.00000000       0.00000000
+      0.00000000       3.80000000       0.00000000
+      0.00000000       0.00000000       3.80000000
+    </parameter>
+    <parameter name="bconds">
+       p p p
+    </parameter>
+    <!--parameter name="vacuum"> 1.0 </parameter-->
```
markdewing: I don't think this is clear, being put in here with an example that won't use the 'vacuum' option even if it is uncommented.   It should be put in a separate example with p p n or p n n boundary conditions.

prckent:
@markdewing This is a good point about deprecation, but I think only the core developers have been using this feature. Sadly, it was not e.g. described in any tutorials.

manual/simulationcell.tex
ye-luo: You are right. I will add an example in the vacuum section.

ye-luo:
I feel no one reads (or reads carefully) the standard output file unless the code aborts.
If we do slow deprecation, users will not notice this until the actual deprecation happens.

markdewing:
I agree that realistically no one will likely read the message and act on it until forced.   However, it might be nice to have a version of the that accepts the option both ways (?)

ye-luo:
In principle, we should provide a version accepting both options and it is not difficult but messy.
In practice, I don't see enough needs and prefer just have this hard switch. We can test it cleanly and I don't need to revisit it.

PR #583, opened 2017-12-16
Input File from convert4qmc

anbenali:
- Fix Generic Input File for begginers mode according to conversation with P. Kent.
- Fix bug in Input file: 
      -No optimization block when No Jastrow present; 
      -Fix Comment about ECP if no ECP is present; 

-Removed tag "-writewfJonly" for writing only one set of files. 
- Added Tag "-production" that will use more efficient parameters for production runs;
         -Only one file is generated depending if "-noJastrow" tag is added.


markdewing:
Okay to test

prckent:
- Renamed example QMCPACK input files to \*.qmc.in\*.xml to distinguish from \*.in\* files used by other applications
- J2 is not created if there is only 1 electron
- J2 u-u terms are only created if there are at least two electrons of one spin 
- J3 is not added unless there are at least 2 atoms
- Commented out the forced logging in src/Numerics/GaussianBasisSet.h::addGaussian and putBasisGroupH5. Separate log functions should be created and used if needed.

prckent:
- Fixed J3 creation rule
- Increased print threshold for HDF5 suggestion to 4000 orbitals * basis functions (in earlier commit)

prckent:
@ye-luo Make a note in the PR reviewers wiki

PR #584, opened 2017-12-16
Hyperion update nightly test script

naromero77:
- Increase the number of procs uses on hyperion to 64 to make test run faster.
- Also, added some checks for flat versus cache mode to prevent from creating a mess of failures on cdash.
- QE configure does not detect optimized libraries automatically, use a manual customized make.sys instead.
- Also, update script to make more robust; based on ideas borrow from the nightly build script on oxygen.


markdewing:
okay to test

prckent:
I note that the custom make.sys for KNL is not included. When we move to QE 6.2+ we should include it, but hopefully the autoconfigure will work better and at worst only a small edit will be needed to the automatically created version.

PR #591, opened 2017-12-19
Sent output of ReportEngine to debug stream

markdewing:
Except for any error output, the default output stream for the ReportEngine is now the debug stream.

Partially addresses #35


ye-luo:
Does this change affect the default output? Or just something internal.

markdewing:
The output is mostly from the constructor, which outputs the class and function name.   Useful for developers to see the code's progress, not so useful for everyone else.
There are a few places that use the error function.   These should eventually be changed to use APP_ABORT.

prckent:
Thanks Mark.

Issue #465, opened 2017-11-03
Testing jastrow functionality


prckent:
#464 and #462 make clear that we need to be more thorough in checking/identifying that the full set of functionality is implemented for different jastrows. This is  relevant for the "extended" functionality used by estimators such as the momentum distribution that need functionality not touched upon by a standard VMC run.

It would be good to (1) at least know the matrix of functionality that is supported, (2) [harder] that this functionality is correct.

Is this straightforward to add to our test sets?

ye-luo:
Do we have a bspline J1 unit test?

jtkrogel:
More needed here or done?

ye-luo:
We have
```
test_J1OrbitalSoA.cpp
test_J1Spin.cpp
test_J2_bspline.cpp
test_kspace_jastrow.cpp
test_polynomial_eeI_jastrow.cpp
```
if the functionality in a specific jastrow is not  covered by unit test, we will use new issues to track that.

Issue #464, opened 2017-11-03
Three-body Jastrow is not supported with the momentum distribution


kylanpaa:
Runs with three-body Jastrow and the momentum distribution estimator result in "Fatal Error. Aborting at OrbitalBase::get_ratios is not implemented by OrbitalBase".

Inputs and outputs are given in the zip-file. The file pwscf.pwscf.h5 can be either recreated with the scf.in or copied from eos (location in the readme.txt).

[J3_nofk_issue.zip](https://github.com/QMCPACK/qmcpack/files/1442087/J3_nofk_issue.zip)

prckent:
Is anyone already familiar with how this functionality differs from the already implemented functionality? One part of the difference should be that for this kind of estimator we need ratios with a least one set of coordinates that are not electron positions, "virtual particles".

jngkim:
It uses 0-th electron as an active electron and compute ratios.
All the elements are in place and just have to be put together.

ye-luo:
@prckent You might noticed that there is a initial VirtualParticleSet implementation but not connected. I have reworked it in order to be used by NLPP evaluation and estimator needs. For the moment, I only cleaned up the class for minimal creation cost. The computation part is not started yet and it will not appear soon, at least after new year and probably I will target SoA first. For the current needs, the optimal way is just implement the missing function get_ratios with minimal effort.

prckent:
Lets discuss prioritization offline. Ilkka needs this for science soon.

ye-luo:
Addressed by #611

PR #595, opened 2017-12-20
[WIP] Consistent active ptcl

### LINKED ISSUE #462: Momentum distribution estimator not working with SoA implementation
SoA implementation (developer version) results in "Fatal Error. Aborting at OrbitalBase::get_ratios is not implemented by J1OrbitalSoA" if the momentum distribution estimator used. No problems with calculations not having n(k) estimator.

Other note: the AoS implementation in the developer version performed roughly 12% slower than the cpu version of qmcpack-3.1.1 for the same test case. In this case the n(k) does work.

Find attached the zip file containing inputs and outputs, apart from the pwscf.pwscf.h5, which can be either recreated using the scf.in file or copied from eos (path is given in the readme.txt).

ye-luo:
This PR should be reviewed and merged after #521. The amount of change to review will reduce.

The is PR changes the behaviour of proposed moves and virtual moves in PbyP moves.
Previously, when a move on electron iat is proposed. R[iat] is overwritten and the old position was save in activePos(fortunately not used outside ParticleSet). When a move is rejected, R[iat] is rewritten as activePos. This choice simplifies the evaluation of single particle orbitals, always accessing R[iat]. In fact for SPOs relying on both R and distances, current setting is not safe and becomes inconvenient. In addition, it makes other implementation difficult like virtual moves.

With this PR, R[iat] is overwritten only when a move is accepted and the proposed position is stored at activePos before accept. In SPOs evaluation routines, R[iat] is no more directly accessed but query P.activePos(iat) by checking iat with activePtcl, the internal state updated during PbyP moves.
All the SPO routines are checked by changing API and finding all the spot by the compiler.

This change also fixes the PbyP drivers relying on Walker.R[iat] for the old position. Now only ParticleSet is needed during moves.

With this change, implementing ratio computing with virtual moves becomes extremely simple and efficient. Within a virtual move to activePos, all the ratios by moving each particles can be computed at once with minimal coding. Some wavefunction classes and distance tables are updated to support this change.

routine get_ratios has been renamed to evaluateRatiosAlltoOne for virtual moves and a fall back implementation by calling ratio() is added. If a wavefunction component is compliant with the assumption, the fall back routine should work out of the box. AoS and SoA Jastrows all works fine.
Determinants are special. Multideterminant has been guarded with a error message.
Unit tests have been added for this routine.

Fix momentum estimator with SoA.

covers #465 
resolves #462 #464


ye-luo:
Will resubmit.

PR #598, opened 2017-12-20
increase g(r) cutoff to Wigner-Seitz cell radius

Paul-St-Young:
g(r) max distance used to default to simulation cell radius. The past default is not optimal for non-cubic cell where Wigner-Seitz cell radius > simulation cell radius.


markdewing:
okay to test

Issue #569, opened 2017-12-15
qmca does not print analyses in same order as datasets are specified


prckent:
A minor issue, but at least sometimes the output order is not consistent with the order of datasets specified on the command line:

```
$ qmca -q e -u ev H2O_noj.s000.scalar.dat  H2O.s000.scalar.dat
H2O  series 0  LocalEnergy           =  -462.539950 +/- 0.148661
H2O_noj  series 0  LocalEnergy           =  -460.908017 +/- 0.336092
$ qmca -q e -u ev H2O.s000.scalar.dat  H2O_noj.s000.scalar.dat
H2O  series 0  LocalEnergy           =  -462.539950 +/- 0.148661
H2O_noj  series 0  LocalEnergy           =  -460.908017 +/- 0.336092
```

jtkrogel:
Currently qmca prints out all files alphabetically.  I can change this easily if desired.

prckent:
I prefer the order of specification - here I was going to write "qmca good better best". Order of specification also means the order does not change as new data files are added.

jtkrogel:
@prckent This issue is addressed by PR #600

PR #600, opened 2017-12-20
nexus: modify qmca to print output in same order as inputted files

jtkrogel:
Addresses #569.  By default, lines of output and plots are now produced following the order of the files provided as input to qmca.  The original behavior (sorting alphabetically) can be recovered with the new "--sort" option.


prckent:
OK to test

PR #601, opened 2017-12-20
correct mass output

Paul-St-Young:
Found and corrected a typo in the `SpeciesKineticEnergy` estimator.
There is no functionality change, only the printed output (particle masses) is corrected.


prckent:
OK to test

PR #603, opened 2017-12-20
Consistent active ptcl

### LINKED ISSUE #462: Momentum distribution estimator not working with SoA implementation

ye-luo:
The is PR changes the behaviour of proposed moves and virtual moves in PbyP moves.
All electron moves are not affected.

Previously, when a move on electron iat is proposed. R[iat] is overwritten and the old position was save in activePos(fortunately not used outside ParticleSet). When a move is rejected, R[iat] is rewritten as activePos. This choice simplifies the evaluation of single particle orbitals, always accessing R[iat]. In fact for SPOs relying on both R and distances, current setting is not safe and becomes inconvenient. In addition, it makes other implementation difficult like virtual moves.

With this PR, R[iat] is overwritten only when a move is accepted and the proposed position is stored at activePos before accept. In SPOs evaluation routines, R[iat] is no more directly accessed but query P.activeR(iat) by checking iat with activePtcl, the internal state updated during PbyP moves.
All the SPO routines are checked by changing API and finding all the spot by the compiler.
Single or multi determinant does not matter as their call to SPO are updated with P.activeR(iat).

This change also fixes the PbyP drivers relying on Walker.R[iat] for the old position. Now only ParticleSet is needed during moves.

Up-to-here, all the changes are internal changes not affecting any users.

With this change, implementing ratio computing with virtual moves becomes extremely simple and efficient. Within a virtual move to activePos, all the ratios by moving each particles can be computed at once with minimal coding. Some wavefunction classes and distance tables are updated to support this change.

evaluateRatiosAlltoOne, previously called get_ratios, is the routine for calculating WF ratios for virtual moves and a fall back implementation by calling ratio() is added. If a wavefunction component is compliant with the assumption, the fall back routine (computing ratios one by one) should work out of the box. AoS and SoA Jastrows all works fine. Unit tests have been added for this routine.
The fallback routine may probably work for multideterminant but the status is unclear due to the lack of test examples. For this reason, safeguards are added to avoid risky use.

Fix momentum estimator with SoA and enables using JeeI in WF.

covers #465 
addresses #462 #464


prckent:
What is the situation with (1) multideterminants and (2) all electron moves? Please edit your PR message to make this clear.

ye-luo:
@prckent update the PR message to reflect all electron moves and multi determinant.

markdewing:
Are the bug fixes necessarily part of the core change in this PR?  And the expanded Jastrow tests?
I would really prefer it if these PR's that touch a large number of files were kept to a minimum of conceptual changes - it would make review and understanding the changes much easier.

prckent:
These mixed + large PRs are really a problem. Very tempted to close...

ye-luo:
I have already tried to separate things but this is already quite stand alone thing.
I need to enable the ratio computing for momentum estimator. What I found is the build blocks not optimal. So I changed them and then enabled ratio computing and the estimator. However, I need to handle code to people who needs to run it. So I didn't fix every piece at first place. When the needed code path is completed, I go back to address the residual needed fixes. Now it is done and I PR to the develop. In the intermediate step, the code is partly broken and not valid to be PRed.
All the bug fixes in this PR is to fixed the bugs I introduced during my changes earlier.

markdewing:
So it sounds like the historical development of the code does not match the best logical progression of the code (at least the way I understand it).

jngkim:
I don't think there is any "historical" development.

The original design was based on the patterns (hardware, software and QMC)
10 years ago and did not make the best choice for today. It is not a
surprise to have big jumps like this is a sign of progress.

Many of these changes were planned but the code was too fragile to make
broad changes. All the work the team has done, making unit tests and CI,
have such changes very easy.
Kudos to Paul's leadership and all the developers.

markdewing:
I meant the historical development of the branch, not of QMCPACK in general.

Looking at the commits in the branch, the first several are all related to the activePtcl change in different parts of the code.
Then there is a change to rename get_ratios to evaluateRatiosAlltoOne, which doesn't seem related to the activePtcl change (?)
Then a commit to fix the momentum estimator for SoA, which does not appear to involve the activePtcl change or the get_ratios rename.
Then back to a commit for evaluateRatiosAllToOne
Then back to the momentum estimator. This commit might depend on the activePtcl fixes - I can't tell?
Then back to activePtcl fixes
One commit is to change the SPO evaluate APIs and apparently a few commits later it is reverted (the message indicates the change was intended to be temporary to catch all the places that needed changing).
Then more changes related to activePtcl
Then additional unit tests for J2 and J3 - it appears these use evaluateRatiosAlltoOne, but it's not clear if they depend on the activePtcl fixes or not.
Then back to the momentum estimator, but these changes look like they might depend on the activePtcl changes?

It looks like there are at least 3 or 4 changes that could be made independently (or at least rearranged into separate logical PR's)
 - activePtcl changes
 - rename get_ratios to evaluateRatiosAllToOne
 - fixes for momentum estimator (these might depend on the active ptcl changes)?
 - additional J2 and J3 tests (which might depend on activePtcl changes and get_ratio renaming)

In my opinion, branches like this should be treated as prototypes, and then rearranged into more logical groups in new branches for the actual PR(s).

ye-luo:
activePtcl changes
This was the initial part of change. You see basically P.R[iat] to R.activePos() changes in src/Wavefunction and some other changes in src/ParticleSet.
    rename get_ratios to evaluateRatiosAllToOne
I also changed the implementation of evaluateRatiosAllToOne in various places and fixed distance tables and jastrow factors to work with ratio computing.
    fixes for momentum estimator (these might depend on the active ptcl changes)?
Call interface is changed. I did the initial fix but need the patch from develop branch from Ilkka and Jaron to complete, they tested it and reported a bug and then I had a second fix.
    additional J2 and J3 tests (which might depend on activePtcl changes and get_ratio renaming)
They are.

I have explained that I am not just developing a separate functionality and also needs update from develop. It naturally ends like this.
Since the breaking up will still give 1 big and 3 very small PRs. I don't see much different from the current PR. To prepare the 4 separate PRs, I need a lot of work to resolve intermediate issue just to get the code build and pass CI.

markdewing:
Breaking it up into 1 large PR and 3 smaller ones would be an improvement.

The large PR would be focused on a single conceptual change.  Even if it's pervasive and touches a lot of files, it's much easier to understand and review.  And changes like this usually follow a pretty regular pattern, so it's fairly easy to skim the diffs and look for anything that deviates.

The first problem with mixing smaller changes into a larger PR is simply finding the changes in a sea of diffs.  The second problem is then trying to understand if any particular change is a necessary consequence of the large change or not.

In my opinion, changes to the code need to be planned and presented with as much care, attention, and planning as we might give to the code itself.

ye-luo:
I prefer not to rewrite the history and redo PRs.
I have other branches based on this one.
Because I do advance the code very quickly, and now I have
0 starting-> 1 PooledMemory -> 2 ActivePtcl -> 3 LAPW -> 4 VritualParticleSet.
The size of changes from 1 to 4 gradually decrease.
Clearly we don't want directly 4.
So once 1 is finished, I PRed 1.
Meanwhile I got 2 in development, once 1 is PRed, I merged 1 to 2 and keep working on 2.
When 2 is done, I noticed that I need to fix 3 and then I merge 2 to 3 and get it fixed.
I started 4 from 0 a while ago but didn't finish. When 3 is done, I merge 3 to 4 and then finished 4 for extending 3's features.

I want only one version to maintain since it carries all the features and optimizations so I can always just maintain one version. Since my changes are often complicated, it often took time to merge. I clearly can't wait the completion of the first one to start the next.

prckent:
> In my opinion, changes to the code need to be planned and presented with as much care, attention, and planning as we might give to the code itself.

Yes. Doing this carefully will improve the quality and health of the code. Speaking generally, as developers we need to be prepared to spend more time upfront planning the development efforts.

As someone who worries a lot about the health of the code and its accessibility I would have liked  to know about these developments and the likely trickiness to review them (as carried out) well ahead of time. 

As for this current PR, I am unable to review it and test it properly before January. So @markdewing someone like @lshulen or @jtkrogel will be needed if they are available.

ye-luo:
I suggest @jtkrogel to review this PR and @lshulen to review the LAPW one.

src/QMCWaveFunctions/ElectronGas/ElectronGasOrbitalBuilder.h
```diff
@@ -158,45 +163,16 @@ struct RealEGOSet: public SPOSetBase
...
   {
-    for(int i=0,iat=first; iat<last; i++,iat++)
-      evaluate_p(P.R[iat],logdet[i],dlogdet[i],d2logdet[i]);
+    for(int iat=first, i=0; iat<last; ++iat,++i)
+    {
+      ValueVector_t v(logdet[i],OrbitalSetSize);
+      GradVector_t g(dlogdet[i],OrbitalSetSize);
+      ValueVector_t l(d2logdet[i],OrbitalSetSize);
+      evaluate(P,iat,v,g,l);
+    }
```
jtkrogel: Are these lightweight views of data or deep copies?

src/QMCWaveFunctions/Jastrow/BsplineFunctor.h
```diff
@@ -658,7 +658,7 @@ BsplineFunctor<T>::evaluateV(const int iat, const int iStart, const int iEnd,
 #pragma vector always 
   for ( int jat = 0; jat < iLimit; jat++ ) {
     real_type r = distArray[jat];
-    if ( r < cutoff_radius )
+    if ( r < cutoff_radius && iStart+jat != iat )
```
jtkrogel: What is the additional logic for here?  Probably worth an in code comment.

src/QMCWaveFunctions/Jastrow/kSpaceJastrow.cpp
```diff
@@ -602,10 +602,10 @@ kSpaceJastrow::ratio(ParticleSet& P, int iat)
 
 /** evaluate the ratio
 */
-inline void kSpaceJastrow::get_ratios(ParticleSet& P, std::vector<kSpaceJastrow::ValueType>& ratios)
+inline void kSpaceJastrow::evaluateRatiosAlltoOne(ParticleSet& P, std::vector<kSpaceJastrow::ValueType>& ratios)
 {
   RealType J1new(0.0);
-  PosType rnew(P.R[0]);
+  const PosType &rnew(P.activePos);
```
jtkrogel: Does switch to reference gain anything or is it just stylistic?  Local copy of 3 doubles is probably quite light.

src/QMCWaveFunctions/OrbitalBase.cpp
```diff
@@ -113,11 +113,11 @@ OrbitalBase::RealType OrbitalBase::KECorrection()
...
 
-void OrbitalBase::get_ratios(ParticleSet& P, std::vector<ValueType>& ratios)
+void OrbitalBase::evaluateRatiosAlltoOne(ParticleSet& P, std::vector<ValueType>& ratios)
 {
-  std::ostringstream o;
-  o << "OrbitalBase::get_ratios is not implemented by " << OrbitalName;
-  APP_ABORT(o.str());
+  assert(P.getTotalNum()==ratios.size());
+  for (int i=0; i<P.getTotalNum(); ++i)
+    ratios[i]=ratio(P,i);
```
jtkrogel: Is this intended to be a primary implementation or just a (slow) fallback?  If the latter, would it be better to have something available just for testing rather than production (or at least a warning that the code is slow)?

What wavefunction components will currently call this code?

src/QMCWaveFunctions/PlaneWave/PWBasis.h
```diff
@@ -300,8 +301,9 @@ class PWBasis: public QMCTraits
   inline void
   evaluateAll(const ParticleSet& P, int iat)
   {
+    const PosType &r(P.activeR(iat));
+    evaluate(r);
```
jtkrogel: What is the advantage here?

jtkrogel:
See in code comments/questions.

A full test/review probably needs n(k) test for HEG and Na.  Also should check that TM/LA total energies are OK for system including J3.  Not sure if I can get to all of these today (last day until Jan).

prckent:
Perhaps add issues, or update existing ones: n(k) integration tests, TM and LA runs with J3?

jtkrogel:
I'm OK with deferring the check given time constraints.  Possible downside is heavier reversion commitment down the line if there is an issue.

@ye-luo response to code-level comments?

src/QMCWaveFunctions/ElectronGas/ElectronGasOrbitalBuilder.h
ye-luo: These are all data views just like other evaluate_notranspose.

src/QMCWaveFunctions/Jastrow/BsplineFunctor.h
ye-luo: This a very important change to facilitate virtual moves. Previously, the exclusion of self iteration term is done at the distance table level. During a pbyp move, the iel-iel distance was assigned a very big distance. However, in a virtual move, you will compute the ratio with respect to every electron. So the distance table must contain intact distances (see the changes in distance table SoA classes). So the removal of self interation term must be handled at evaluation.

src/QMCWaveFunctions/Jastrow/kSpaceJastrow.cpp
ye-luo: This is just stylistic.

src/QMCWaveFunctions/OrbitalBase.cpp
ye-luo: In general, this is a fallback solution but it doesn't mean it is slow in all cases.
For determinants, it is slow because it is evaluating SPO nelec times and So we have a specialization to just do once for spin up and once for spin down.
This is the primary implementation for all the Jastrow factors. Since there is little data reuse. You can compare the old implementation in AoS. Since there is no benefit having specialization in Jastrow factors, I removed all the old specialized AoS code for having less code to maintain.

src/QMCWaveFunctions/PlaneWave/PWBasis.h
ye-luo: A stylish thing and I did copy paste.

ye-luo:
@jtkrogel I've replied all the code-level comments.

jtkrogel:
I have no further questions.  Approved.  

Would have been (or someday would be) nice to see any change in speed for the non-local updates.

PR #606, opened 2017-12-21
nexus: fix qfit for variable length input data

jtkrogel:
Bug in the code caused reblocking lengths to be computed incorrectly.  Simple fix.


markdewing:
okay to test

PR #608, opened 2017-12-22
Use local uninitialized headers instead of thrust library implementation

### LINKED ISSUE #565: Complex CUDA build fails with CUDA 9 due to thrust uninitialized missing file
With CUDA 9.1, determinant_update.cu will fail to build due to a missing file:
```
#include <thrust/system/cuda/detail/bulk/uninitialized.hpp>
```

I tried a quick replacement with thrust/device_vector.h, but this results in a series of declaration errors later on.

prckent:
Allows complex CUDA version to compile with CUDA 9. The local implementation from https://github.com/jaredhoberock/uninitialized uses a different namespace. Hence should be safe with older or newer CUDA versions. Licensing is safe for incorporating in QMCPACK.

Closes #565 

Changes during review:

- [x] Move header files to src/CUDA


prckent:
Done

prckent:
Thanks @jefflarkin , @jaredhoberock

Issue #609, opened 2017-12-30
Improve the implementation of population control


ye-luo:
There are a few performance bottlenecks spotted in the population control implementation.
All these bottlenecks amplify the load imbalance among MPI ranks.

This issue lists the upcoming PRs to address this issue
1, The container std::vector with aligned allocator used by PooledMemory is slower than std::vector with default allocator. This is a performance regression comparing to v3.3.0. Solution: replace std::vector with our Vector class.
2, The cost in 1 was mainly due to frequent allocation and deallocation. A new algorithm will be used to recycle bad walkers. It also avoids sudden temporal high memory needs.
3, A new algorithm will be added to send/recv identical walkers in batches. isend/recv will also help reducing serialization. The old implementation mentioned as async in #461 can be removed. This part of improvement can also be used for reconfiguration.

more details will be given in each PR.

A NiO S64 run on Theta 1024 KNL nodes. 1024 MPI ranks with 64 threads. Using ynamic population control. On average, 63 walkers per node. Intel compiler 18, Cray MPI.
version DMC block CPU
v3.3.0 179.8 sec 
develop about 2x slower than v3.3.0. I don't have the exact run.
new 69.4 sec
The new run was a survey run with extra barriers. The final version has no barriers and saves a few seconds. With all the changes, the branching cost goes down to 7% level.

prckent:
@ye-luo Thanks for catching these. Please can you update your Issue to give full details of the runs for which you give timings? e.g. walker count, MPI task count, which run (NiO performance, ideally), which computer system/mpi, where tested, which versions of code etc.

I would like to understand the history of these problems so that we can catch them sooner in future.

ye-luo:
@prckent run details updated.

lshulen:
This is quite interesting.  The factor of 2 performance difference for a calculation so close to 1 walker per thread suggests the interpretation that in the first case some thread is forced to propagate 2 walkers whereas in the second case all threads stay below 1 walker per thread.  

Are there enough steps / blocks in your run for you to be sure that this is a persistent problem rather than a statistical fluke?  Another way of asking the question, do you still see a large population control cost if you have ~50 walkers / node instead of ~63?

prckent:
The performance history should be trackable in cdash...

ye-luo:
@lshulen I set the warm up steps longer than the run so the tight population control is always effective and the number of walkers never goes beyond 64 per node, confirmed by dmc.dat.
[...]

PR #611, opened 2018-01-02
Opt momentum estimator

ye-luo:
The computation inside momentum estimator has been optimized by rearranging the calculation and reusing data. This is effective for both AoS and SoA.
The evaluateRatiosAlltoOne of J1,J2,J3 with SoA has also been optimized by using specialized implementation.

The test case I used is VO2 48 atom cell. 400 electron. On titan, 2 MPI per node and 8 thread per MPI.
When using 128 samples, I saw huge speed up.
No estimator, 1.564921 sec
With 128 samples, 28.948585 sec
Optimized code with 128 samples, 4.872676 sec.

@kylanpaa Please verify that
- [x] real is good.
- [x] complex is good.

### DIFF
--- a/src/Particle/DistanceTableData.h
+++ b/src/Particle/DistanceTableData.h
@@ -80,7 +80,6 @@ struct TempDisplacement
   T rinv1;
   ///new displacement
   TinyVector<T,N> dr1;
-  TinyVector<T,N> dr1_nobox;
   inline TempDisplacement():r1(0.0),rinv1(0.0) {}
   inline void reset()
   {
--- a/src/Particle/SymmetricDistanceTableData.h
+++ b/src/Particle/SymmetricDistanceTableData.h
@@ -147,7 +147,6 @@ struct SymmetricDTD
     for(int iat=0; iat<N[SourceIndex]; ++iat)
     {
       PosType drij(rnew - P.R[iat]);
-      Temp[iat].dr1_nobox=drij;
       RealType sep=std::sqrt(DTD_BConds<T,D,SC>::apply_bc(drij));
       Temp[iat].r1=sep;
       Temp[iat].rinv1=1.0/sep;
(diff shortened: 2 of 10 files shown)


prckent:
@ye-luo , @kylanpaa please work together to add a test for this estimator.

@ye-luo  please describe in detail for which system these timings come from (same comment as for #609 )

kylanpaa:
The tests at eos indicate that both real and complex are good and significant speed-up is evident. The test case was R phase of VO2 with 50 electrons and 2x1x1 twist grid for both real and complex. Additionally, long VMC runs were performed with 3x1x1 twist grid using the complex SoA code. Comparison of the 3x1x1 results with those from the older implementation coincide properly. The 3x1x1 results are shown in the attachment, where on the left one has the spherically averaged momentum distributions (new and old in the same plot) and on the right the deviation of the results in units of 1sigma (the red dashed line is a normalized Gaussian).

[nk_tests2.pdf](https://github.com/QMCPACK/qmcpack/files/1604475/nk_tests2.pdf)

ye-luo:
@kylanpaa What is the size of the h5 orbital file of VO2?
@prckent I have no experience dealing with h5 output from estimators. Can we have extra hands to help with the test?

prckent:
First we need a small/quick reliable test. Unfortunately custom check code may be needed for the h5.  Perhaps we should add the small NiO cell as we discussed to the performance tests and use that wavefunction?

jtkrogel:
Why not just add the n(k) estimator to existing real/complex diamond test runs?  Then we just need to write an additional post-processing script.

I am planning to operate in this way for coming estimator tests in general (i.e. add to existing integration tests to avoid eventually very long test times).  

I can do this for n(k) now and later for others.  It will require a different kind of test (a formalized version of what Ilkka is showing in his attached pdf) to judge significance for a large number of values simultaneously.

kylanpaa:
I think it would be good to merge the n(k) tests with the diamond case as @jtkrogel suggests. 

@ye-luo In my tests the wave function files are 56Mb for 2x1x1 and 79Mb for 3x1x1.

prckent:
OK, lets add the test based on the existing carbon diamond wavefunctions. Since this will take a little while to prepare, I propose to merge this PR and that we add the test in a later PR very shortly. Note that the set of estimators that are not supported by the CUDA code will have to have their own dedicated run.

PR #612, opened 2018-01-02
Switch pooledmemory container

ye-luo:
This is the first PR described by #609.
1, The container std::vector with aligned allocator used by PooledMemory is slower than std::vector with default allocator. This is a performance regression comparing to v3.3.0. Solution: replace std::vector with our Vector class.

@markdewing I used valgrind on the unit test and noticed two minor memory leaking not from pooledmemory. One from the timer manager and one from the infostream.


src/Utilities/tests/test_pooled_memory.cpp
```diff
@@ -91,6 +94,34 @@ TEST_CASE("pack scalar", "[utilities]")
   REQUIRE(i6[0]==i6[0]);
   REQUIRE(i6[1]==i6[1]);
   REQUIRE(i6[2]==i6[2]);
+
+  // up to 512 MB.
+  for(size_t size_scale=15; size_scale<30; size_scale++)
```
markdewing: This timing code should be moved to it's own unit test, and ifdef-ed out by default, since it doesn't check anything.

markdewing:
I ran the timing code in the unit test with std::vector and Vector.  
On a Xeon, for the smaller allocations, Vector is faster and on larger allocations, std::vector is faster.
What behavior is expected?  And is it different on KNL?

ye-luo:
When both using our aligned allocator, Vector should be faster than std::vector.
Vector with aligned allocator may be slower than std::vector with default allocator in small allocation but the difference goes to zero as the allocation gets large. I also noticed Vector faster than std::vector at very large allocation.
In the next PR, the algorithm change will minimized the allocator/deallocation, so the performance impact due to memory allocation will be negligible.

markdewing:
My test runs were faulty, apparently.  Now the Vector results are uniformly faster than the std::vector results.

PR #613, opened 2018-01-02
Branching memory

ye-luo:
This PR changes the memory handling of walker creating and annihilation.
This is the second PR for #609 
In the old algorithm,
1. sortWalkers makes decision of walker creating and annihilation. good_w and bad_w are assigned and good walkers have counters (ncopy_w) recording how many copies it should have. bad_w are freed.
2. copyWalkers first copy good walkers based on ncopy_w.
3. With MPI loadbalancing, walkers are sent from nodes with more walkers to nodes with less walkers. When walkers are transferred, their memory is freed.

In the new algorithm,
1. largely unchanged but the bad walkers are not freed.
2. First do the old step 3, send walkers from good_w, and decrease copy counter if a copy is sent. On the receiving side, bad_w are used to accept walkers from other nodes until bad_w is depleted. Received walkers are removed from bad_w registered in good_w.
3. copyWalkers based on the copy counter, first recycle bad_w and then allocating new walkers. If there is any bad_w left, free them. The copy procedure is also threaded.

With this new algorithm, actual walker memory allocation and deallocation is around zero each iteration and there is no sudden high memory needs in the intermediate steps.

### DIFF
--- a/src/QMCDrivers/DMC/WalkerControlMPI.cpp
+++ b/src/QMCDrivers/DMC/WalkerControlMPI.cpp
@@ -24,35 +24,27 @@
 namespace qmcplusplus
 {
 
-//#if defined(PRINT_DEBUG)
-//#define DMC_BRANCH_START(NOW) NOW
-//#define DMC_BRANCH_STOP(TID,TM) TID=TM
-//#define DMC_BRANCH_DUMP(IT,T1,T2,T3)  \
-//  OhmmsInfo::Debug->getStream()  << "BRANCH " \
-//<< std::setw(8) << IT \
-//<< " SORT" << std::setw(16) << T1 \
-//<< " GSUM" << std::setw(16) << T2 \
-//<< " SWAP" << std::setw(16) << T3 << std::endl
-//#else
-#define DMC_BRANCH_START(NOW)
-#define DMC_BRANCH_STOP(TID,TM)
-#define DMC_BRANCH_DUMP(IT,T1,T2,T3)
-//#endif
-
-
 //#define MCWALKERSET_MPI_DEBUG
 
 enum DMC_MPI_Timers
 {
   DMC_MPI_branch,
+  DMC_MPI_imbalance,
   DMC_MPI_prebalance,
+  DMC_MPI_sortWalkers,
+  DMC_MPI_copyWalkers,
+  DMC_MPI_allreduce,
   DMC_MPI_loadbalance
 };
 
 TimerNameList_t<DMC_MPI_Timers> DMCMPITimerNames =
 {
   {DMC_MPI_branch, "WalkerControlMPI::branch"},
+  {DMC_MPI_imbalance, "WalkerControlMPI::imbalance"},
   {DMC_MPI_prebalance, "WalkerControlMPI::pre-loadbalance"},
+  {DMC_MPI_sortWalkers, "WalkerControlMPI::sortWalkers"},
+  {DMC_MPI_copyWalkers, "WalkerControlMPI::copyWalkers"},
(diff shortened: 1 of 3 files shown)


src/QMCDrivers/DMC/WalkerControlMPI.cpp
```diff
@@ -76,60 +64,49 @@ WalkerControlMPI::WalkerControlMPI(Communicate* c): WalkerControlBase(c)
 int
 WalkerControlMPI::branch(int iter, MCWalkerConfiguration& W, RealType trigger)
```
prckent: Description of new algorithm needed in doxygen comment.

src/QMCDrivers/DMC/WalkerControlMPI.cpp
```diff
@@ -186,9 +161,8 @@ void WalkerControlMPI::swapWalkersSimple(MCWalkerConfiguration& W)
   std::vector<int> minus, plus;
```
prckent: Description of algorithm needed in doxygen comment

src/QMCDrivers/WalkerControlBase.cpp
```diff
@@ -443,23 +440,40 @@ int WalkerControlBase::sortWalkers(MCWalkerConfiguration& W)
 
 int WalkerControlBase::copyWalkers(MCWalkerConfiguration& W)
```
prckent: Description of algorithm needed in doxygen comment

src/QMCDrivers/DMC/WalkerControlMPI.cpp
ye-luo: The doxygen comment for this routine will be added in the next PR since the algorithm will further change.

src/QMCDrivers/DMC/WalkerControlMPI.cpp
```diff
@@ -73,63 +61,59 @@ WalkerControlMPI::WalkerControlMPI(Communicate* c): WalkerControlBase(c)
...
   {
     Cur_pop+= NumPerNode[i]=static_cast<int>(curData[j]);
   }
   myTimers[DMC_MPI_prebalance]->stop();
-  myTimers[DMC_MPI_loadbalance]->start();
   if(qmc_common.async_swap)
+  {
+    myTimers[DMC_MPI_copyWalkers]->start();
+    copyWalkers(W);
+    //myComm->barrier();
```
markdewing: What are these barriers for?

ye-luo: When I investigate the source of imbalance, I turn on these barrier one at a time.

prckent: Any need to keep them? We are trying to minimize / remove sections of commented out code.

ye-luo: Not really. I will clean them up in the next PR which also touches this part. If I remove it here, it will be more complicated for me to do the rebase.

src/QMCDrivers/WalkerControlBase.cpp
prckent: Please add a note to say that this algorithm attempts to minimize large allocations/deallocations. This was in the PR. The new comments say "what" but not "why".

prckent:
@markdewing OK to merge?

@ye-luo In your next PR please expand the comments to say *why* the algorithm is as it is (to avoid memory fluctuations, as you describe in your PR comment). Your current comments describe what is done but not why. The latter is much more important to write down.

ye-luo:
@prckent I will add 'why' in the next PR.

PR #615, opened 2018-01-03
Resurrect Ionwf

Paul-St-Young:
Ratio test is broken at the current commit a7047891a7191ffa3b772905d5043862748d167f
However ratio test passes as of Dec. 23 commit 8194d94148ec6a2bb60e44c349b9b1e6ca152c09
 
update: Ratio test is still broken at commit 8194d94148ec6a2bb60e44c349b9b1e6ca152c09

### DIFF
--- a/manual/ionwf.tex
+++ b/manual/ionwf.tex
@@ -8,6 +8,8 @@ The Gaussian Product wavefunction implements eq.~\ref{eq:gauss_prod_wf}
 \end{equation}
 where $\vec{R}_i$ is the position of the $i^{\text{th}}$ quantum particle and $\vec{R}_i^o$ is its center. $\sigma_i$ is the width of the Gaussian orbital around center $i$.
 
+This variational wavefunction enhances single-particle density at chosen spatial locations with adjustable strengths. It is useful whenever such localization is physically relevant yet not captured by other parts of the trial wavefunction. For example, in an electron-ion simulation of a solid, the ions are localized around their crystal lattice sites. This single-particle localization is not captured by the ion-ion Jastrow. Therefore the addition of this localization term will improve the wavefunction. The simplest use case of this wavefunction is perhaps the quantum harmonic oscillator (please see the ``tests/models/sho'' folder for examples).
+
 \subsubsection{Input Specification}
 
 \begin{table}[h]
--- a/manual/ionwf.tex
+++ b/manual/ionwf.tex
@@ -1,0 +1,71 @@
+\section{Gaussian Product Wavefunction}
+\label{sec:ionwf}
+
+The Gaussian Product wavefunction implements eq.~\ref{eq:gauss_prod_wf}
+\begin{equation}
+\Psi(\vec{R}) = \prod_{i=1}^N \exp\left[ -\frac{(\vec{R}_i-\vec{R}_i^o)^2}{2\sigma_i^2} \right]
+\label{eq:gauss_prod_wf},
+\end{equation}
+where $\vec{R}_i$ is the position of the $i^{\text{th}}$ quantum particle and $\vec{R}_i^o$ is its center. $\sigma_i$ is the width of the Gaussian orbital around center $i$.
+
+\subsubsection{Input Specification}
+
+\begin{table}[h]
+\begin{center}
+\begin{tabular}{l c c c l }
+\hline
+\multicolumn{5}{l}{Gaussian Product Wavefunction (ionwf)} \\
+\hline
+\bfseries name & \bfseries datatype & \bfseries values & \bfseries defaults  & \bfseries description \\
+\hline
+name & text & ionwf & (required) & Unique name for this wavefunction \\
+width & floats & 1.0 -1 & (required) & Widths of Gaussian orbitals.\\ 
+source & text & ion0 & (required) & Name of classical particle set.\\ 
+\hline
+\end{tabular}
+\end{center}
+\end{table}
+
+\FloatBarrier
+
+Additional information:
+\begin{itemize}
+\item \texttt{width} There must be one width provided for each quantum particle. If a negative width is given, then its corresponding Gaussian orbital is removed. Negative width is useful if one wants to use Gaussian wavefunction for a subset of the quantum particles.
+\item \texttt{source} The Gaussian centers must be specified in the form of a classical particle set. This classical particle set is likely the ion positions ``ion0'', hence the name ``ionwf''. However, one may define arbitrary centers using a different particle set. Please refer to examples in `tests/models/sho'.
+\end{itemize}
+
+\subsection{Example Use Case}
+\begin{lstlisting}
+  <qmcsystem>
+    <simulationcell>
+      <parameter name="bconds">
+            n n n
+      </parameter>
+    </simulationcell>
+    <particleset name="e">
+      <group name="u" size="1">
+        <parameter name="mass">5.0</parameter>
+        <attrib name="position" datatype="posArray" condition="0">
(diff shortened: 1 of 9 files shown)


prckent:
OK to test

src/QMCWaveFunctions/IonOrbitalBuilder.cpp
```diff
@@ -28,6 +28,8 @@ bool
 IonOrbitalBuilder:: put(xmlNodePtr cur)
 {
   ParticleSet &p = targetPtcl;
+  widthOpt.resize(targetPtcl.getTotalNum());
+  widthOpt = -1;
```
ye-luo: Why initialize the value to -1? The later non-zero checking seems assume initialized value 0 and requires full user input.

ye-luo:
@Paul-St-Young I would like to understand why my change on develop break the IonOrbital builder. Since you provide a fix, could you explain your logic?

manual/ionwf.tex
```diff
@@ -0,0 +1,71 @@
+\section{Gaussian Product Wavefunction}
+\label{sec:ionwf}
+
+The Gaussian Product wavefunction implements eq.~\ref{eq:gauss_prod_wf}
```
prckent: Thanks for adding the manual section. Please can you add a citation to any published work using this wavefunction (QMCPACK or not) and a couple of sentences on when/where the wavefunction might be used.

src/QMCWaveFunctions/IonOrbitalBuilder.cpp
Paul-St-Young: @ye-luo the width parameter is required. The intended behavior is abort when no user input is given.

Paul-St-Young:
@ye-luo I actually have no idea what broke IonOrbital and provided no fix. I noticed that my test is broken on the current develop HEAD. I was wondering if it was my test or the code, so I went back to a previous commit.

manual/ionwf.tex
Paul-St-Young: @prckent unfortunately I have no knowledge of the history of this wavefunction. The file banner suggests it was first developed by Ken Esler. Could we contact him?

Paul-St-Young: I will add a few sentences describing potential use of the wavefunction.

src/QMCWaveFunctions/IonOrbitalBuilder.cpp
ye-luo: I agree that the intended behaviour is abort when no user input is given. But I'm not convinced that it justifies your change. The old code relies on oAttrib to do the resize of widthOpt if a user input is give. If no input is given. widthOpt has zero size and abort. In your change, you resize and initialize the value to -1. The following num_nonzero will not be able to abort if no user input is given. If you set the initial value to 0, I may agree.

Paul-St-Young: @ye-luo "The old code relies on oAttrib to do the resize of widthOpt" is exactly why IonOrbital has been broken. For some reason `oAttrib` no longer automatically does the resize. I see your point about the change in initial default now. I will change the defaults to 0.

ye-luo:
After my investigation, this is not a recent breakage. v3.1.1 also has the size issue. @Paul-St-Young your magic resize is the correct fix. I will manually do the merge to make your branch up-to-date after PR #622 and make the two lines in 1.

ye-luo:
@Paul-St-Young 
```
Finite difference test: PASS
Wavefunction ratio exceeds tolerance 1.0000000000e-03) for particle 0
  Internally computed ratio = 2.1626500948e-01
  Separately computed ratio = 2.1626476619e-01
Ratio test: FAIL
```
Is this failure normal or just a false alarm?

Paul-St-Young:
@ye-luo thanks for looking into this. I am not sure if what you are seeing is a false alarm or not. Do you see failure using the `check_grad_lap.py` script?

Please also move the quantum particle around and rerun the wavefunction test to see if the error persists.

ye-luo:
@Paul-St-Young I ran the ctest and got a failure. Then I opened the log and saw the error.

Paul-St-Young:
@ye-luo the ctest should pass. Could you try an older binary?

ye-luo:
@Paul-St-Young what do you mean by older binary?

Paul-St-Young:
Sorry, I meant an older commit, like 8194d94148ec6a2bb60e44c349b9b1e6ca152c09

Paul-St-Young:
No need to port the entire ctest over. I just slap on the MNC, run `grad_lap.xml`, then `check_grad_lap.py`.

ye-luo:
I tried the old commit and got
```
Wavefunction ratio exceeds tolerance 1.0000000000e-03) for particle 0
  Internally computed ratio = 8.5525616287e-02
  Separately computed ratio = 8.5525038991e-02
Ratio test: FAIL
```
I just look at the number and it seems that the relative error is only 0.00067, smaller than 0.001.

Paul-St-Young:
@ye-luo Does the error persist when you move the quantum particle around?

Paul-St-Young:
Oh I am sorry, the old commit is also broken. I must have used my own binary when I did my test.
Please try move the quantum particle to `0.1 -0.1 0.2`. It will make the error more obvious.

ye-luo:
I move the particle in particleset name="e"
```
Finite difference test: PASS
Wavefunction ratio exceeds tolerance 1.0000000000e-03) for particle 0
  Internally computed ratio = 4.1884301184e+01
  Separately computed ratio = 8.4155461651e-02
Ratio test: FAIL
```
So?

Paul-St-Young:
The internal ratio 48.9 differs from finite-difference ratio 0.084, so the wavefunction ratio is wrong.
This is the ratio test failure I mentioned at the beginning of this PR. Do you have an idea what caused it?

Paul-St-Young:
The code is still buggy. The ratio test passes on my branch.

ye-luo:
Which is 'my branch'?
A quick look to the code indicates it is not correct in many places. The orbital should be like the one body Jastrow. The distance table handling seems not right. The Orbital class should only hold the index and then get the table from the argument P.

Paul-St-Young:
https://github.com/Paul-St-Young/qmcpack/tree/kylin

Paul-St-Young:
Oh shoot, I forgot a piece of fix when I made this PR. The ratio test should be fixed now. Sorry about the confusion @ye-luo

src/QMCWaveFunctions/IonOrbital.cpp
```diff
@@ -155,7 +156,12 @@ IonOrbital::ratioGrad(ParticleSet& P, int iat, GradType& grad_iat)
 }
 
 void
-IonOrbital::restore(int iat) {}
+IonOrbital::restore(int iat)
```
ye-luo: Restore should not be left empty. evaluateLog should directly assign zero.

Paul-St-Young: Yes, that was the problem. Please see "fix ratio test". I accidentally left out this piece of the fix when I made this PR. Sorry about that.

ye-luo: I saw "fix ratio test". I mean revert the change of restore and assign zero directly in evaluateLog. One goal is to remove all the restore at least in CPU code.

Paul-St-Young: Oh I see. I was not aware of that goal. Will clean now

prckent:
Many thanks @Paul-St-Young

Issue #602, opened 2017-12-20
FDLRWfn is broken by PR #521


ye-luo:
The breakage from #521 will be fixed ASAP by me or @nsblunt

ye-luo:
Some updates from #605 that at least one FDLRWfn test works.
@nsblunt There is an action item for you. Since you introduced the test short-H4-orb-opt-16-1, could you verify that it was actually not working and now works and the reference value needs to be updated?

nsblunt:
Yes I looked at it and it was indeed broken before and seems fixed now. It's a little strange because the values being printed to standard output are unchanged, and were all correct before this fix. But the values in scalar.dat files were very wrong. But I agree that's it's fixed now anyway. Thanks for this.

ye-luo:
@nsblunt Thanks for checking. Could you please update the reference value for that test and also upload reference calculation (namely the long run with optimized WF which gives the reference energy). No hurry.

nsblunt:
I can update the reference, but I didn't perform a long run to create the reference value. I simply took the final value and error bars from the test output. I agree it's better to obtain a more accurate value, but hopefully this will catch significant deviations. What would you like me to do, should I instead now perform a bigger calculation to update the benchmark value now?

ye-luo:
Very simple, in your original input file, add a section of very long VMC after the last optimization block.
Use the VMC block energy and last optimization block error bar to update the reference value.

nsblunt:
Should I update the reference on a new branch with a pull request?

prckent:
The improved tests would be a "new feature" deserving their own short lived branch. Please make a new PR from a fresh branch.

nsblunt:
Sure, could I please be given write access to push my branch? (I can do it from my fork if not but currently not working from there).

ye-luo:
You can directly push your working branch to your fork (needed to be added as a remote) even if you didn't not start from there.

PR #616, opened 2018-01-03
Update short-H4-orb-opt test

nsblunt:
This is updating the reference for the short-H4-orb-opt test, as requested in #602. An optimized Jastrow was added, since without it the variations between runs were always huge, and so difficult to get consistency.


prckent:
OK to test

PR #618, opened 2018-01-03
Add bibliography to manual listing all qmcpack papers

prckent:
Using @jtkrogel 's .bib file, add a new Appendix listing all the papers in this file.

The goal is to not require any novel scripts or preprocessing outside the normally available LaTeX tools and packages, and to be able to cite every entry in a .bib file automatically. After surveying those available, using the bibtopic package seemed the most appropriate. As noted in the new README.md, bibtex must be invoked twice on newly created files.

Note that the references are generated in the order listed, which is currently not alphabetical. Sorting will likely lose the comments scattered through this file, unless done by hand.


manual/README.md
```diff
@@ -0,0 +1,15 @@
+# QMCPACK Manual
+
+This directory contains the LaTeX source for the QMCPACK manual. The
+PDF manual is not currently created by the build process and must be
+created manually. Note that due to the use of the bibtopic package to
+support two bibliographies, bibtex must be invoked twice:
+
+```
+pdflatex qmcpack_manual.tex
```
markdewing: Could this sequence of commands also be put into a shell script ('build_manual.sh')?

prckent: Added.

manual/qmcpack_papers.bib
```diff
@@ -0,0 +1,1676 @@
...
+  year = {2005},
+  month = {Jan},
+  publisher = {American Physical Society},
+  doi = {10.1103/PhysRevLett.94.036404},
+  url = {http://link.aps.org/doi/10.1103/PhysRevLett.94.036404}
+}
+
+
+% (app) spherical jellium qwalk (used orbitals from SQD code, packaged/part of qmcpack)
+@article{bajdich2011,
```
jtkrogel: I really don't know if QMCPACK was used in these cases.  Any reason not to comment these out pending further investigation?  (Only necessary due to auto-include)

manual/qmcpack_papers.bib
```diff
@@ -0,0 +1,1676 @@
...
+  year={2016},
+  school={University of Illinois at Urbana-Champaign},
+  url={http://hdl.handle.net/2142/90737},
+}
+
+
+
+
+
+% cite qmcpack only
```
jtkrogel: Anything below this point either only cites QMCPACK or is a derivative work that does not use QMCPACK.  These should be excluded from the citation list (though tracking them separately is possibly a matter of internal interest).

manual/qmcpack_papers.bib
prckent: According to the paper, QMCPACK was used for the orbitals, so I think this counts.

manual/qmcpack_papers.bib
prckent: Removed

manual/qmcpack_papers.bib
jtkrogel: Agreed.  SQD was used in bajdich2011.  

What about the other two (chiesa2005 and swingle2013)?

prckent: I emailed Norm to check Swingle2013, but considering we gained lots of Renyi code in the same time period, I think the answer is yes. Chiesa2005 is presumably no.

jtkrogel:
I think we are safe with these assumptions and clarifications.  Approved.

PR #620, opened 2018-01-03
Nonblocking send/recv

ye-luo:
This PR introduce two optimization in the walker send/recv. This is the third PR for #609.
1, multiple copy of one walker is sent in one shot. This reduces the amount of data transferred.
2, Using non-blocking send/recv to avoid blocking send/recv serialization. Turned on by default.
Both algorithms will be transferred to reconfiguration in the near future.

Command line option async_swap has been removed as well as the old implementation.
input parameter use_nonblocking is introduced to toggle on/off non-blocking send/recv.
manual has been updated.

### DIFF
--- a/src/QMCDrivers/DMC/WalkerControlMPI.cpp
+++ b/src/QMCDrivers/DMC/WalkerControlMPI.cpp
@@ -198,22 +198,30 @@ void WalkerControlMPI::swapWalkersSimple(MCWalkerConfiguration& W)
   fout << std::endl;
 #endif
   int nswap=std::min(plus.size(), minus.size());
+
+  // sort good walkers by the number of copies
+  assert(good_w.size()==ncopy_w.size());
+  std::vector<std::pair<int,int> > ncopy_pairs;
+  for(int iw=0; iw<ncopy_w.size(); iw++)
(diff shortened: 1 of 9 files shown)


src/QMCDrivers/DMC/WalkerControlMPI.cpp
```diff
@@ -190,183 +191,203 @@ void WalkerControlMPI::swapWalkersSimple(MCWalkerConfiguration& W)
...
   fout << std::endl;
 #endif
- //  From the class
-  // MyContext (read)
-  // myComm (read)
-  // NumWalkersSent (write)
-  int nswap=std::min(plus.size(), minus.size());
+  if(plus.size()!=minus.size())
+  {
+    app_error() << "Walker send/recv pattern doesn't match." << std::endl;
```
markdewing: Could you put the plus and minus sizes in the error message.

prckent:
(I expect to review this later today)

ye-luo:
OpenMP is not in send/recv. It is only in walker copy. This is checked by a DMC with no MPI build.
send/recv part is MPI only but MPI probably uses threads to handle non blocking send/recv.

prckent:
Yes, we have several tests that check this. But I felt it was worth noting that a pure MPI run would no longer catch everything.

PR #622, opened 2018-01-03
Initialize Vector value similarly to std::vector.

ye-luo:


ye-luo:
@markdewing Could you review this one?

PR #623, opened 2018-01-04
Fix FindMKL

ye-luo:
FindMKL was not behaving properly on my desktop.
The reason is the following
CMAKE_FLAGS only accept -DVAR:TYPE=VALUE form.
COMPILE_DEFINITIONS is the right one to use. It will invoke add_definitions and allows any flag.
https://cmake.org/cmake/help/v3.2/command/try_compile.html

I actually investigated the problem a year ago #11 but now I have a solution instead of a workaround.

The change has been checked on a linux workstation as well as Cray X40.
Hope this also resolves the LLNL pain.


naromero77:
@ye-luo Can you clarify what the COMPILE_DEFINITIONS flag is actually doing? The CMake manual is not 100% clear. 

Is it taking the original definitions from the initial cmake invocation, e.g. -DSOA_ENABLE=1, -DCOMPLEX=0, etc. and passing to the try_compile command so that the proper (temporary) CMakeLists.txt is generated?

PR #625, opened 2018-01-04
Start cleaning up wavefunction setup output and distance table output.

markdewing:
Also start on some of the SPO output.   I've been trying to move through the code in order the output appears, but that doesn't always correspond neatly to isolated sections of code.

It's not clear what the summary output should be for the wavefunction setup - that will need to wait until the verbose output is cleaned up.

The distance table output is interspersed with the wavefunction setup because the distance tables are created on demand when the tables are added to the wavefunctions.

 The exact distance table function called is relegated to debug output.

Example verbose output:
```
 Wavefunction setup:
 -------------------
  Name: psi0
  Created basis set builder named 'einspline' of type einspline
  Building SPOset '' with '' basis set.

  Distance table for similar particles (A-A):
    source/target: e
    Using structure-of-arrays (SoA) data layout
    PBC: bulk   Orthorhombic: no
    Setting Rmax = 1.9475


  Distance table for dissimilar particles (A-B):
    source: ion0  target: e
    Using structure-of-arrays (SoA) data layout
    PBC: bulk   Orthorhombic: no
    Setting Rmax = 1.9475

```


src/Particle/DistanceTableAA.cpp
```diff
@@ -37,53 +37,63 @@ DistanceTableData* createDistanceTable(ParticleSet& s, int dt_type)
...
+
   if(sc == SUPERCELL_BULK)
   {
     if(s.Lattice.DiagonalOnly)
     {
       if(useSoA)
       {
-        o << "    PBC=bulk Orthorhombic=yes  Using SoaDistanceTableAA<T,D,PPPO> of SoA layout " << PPPO << std::endl;
+        o << "    PBC: bulk   Orthorhombic: yes" << std::endl;
+        dbg << "    Using SoaDistanceTableAA<T,D,PPPO> " << PPPO << std::endl;
```
ye-luo: The actually useful info in this line is only PPPO. SoA and particle pair have been printed earlier.
Can we just always print "PBC: bulk   Orthorhombic: yes  Type: PPPO"?

src/Particle/DistanceTableAA.cpp
```diff
@@ -37,53 +37,63 @@ DistanceTableData* createDistanceTable(ParticleSet& s, int dt_type)
...
         dt = new SoaDistanceTableAA<RealType,DIM,PPPO+SOA_OFFSET>(s);
       }
       else
       {
-        o << "    PBC=bulk Orthorhombic=yes  Using SymmetricDTD<T,DIM,PPPO> " << PPPO << std::endl; 
+        o << "    PBC: bulk   Orthorhombic: yes" << std::endl;
```
ye-luo: This now redudant as the previous case, just move it to the upper level and keep one copy.

src/Particle/DistanceTableAA.cpp
markdewing: Removed the debug output and added the 'Type:' information to each line.

prckent:
Is there a mention of the meaning of codes such as "PPNX" anywhere?

ye-luo:
Not yet. PPN is the same as your boundary condition input, the last letter is for the specific implementation.

src/Particle/DistanceTableAA.cpp
```diff
@@ -38,140 +38,136 @@ DistanceTableData* createDistanceTable(ParticleSet& s, int dt_type)
...
+    o << "    Using structure-of-arrays (SoA) data layout" << std::endl;
+  } else {
+    o << "    Using array-of-structure (AoS) data layout (less efficient than SoA)" << std::endl;
+  }
+
   if(sc == SUPERCELL_BULK)
   {
     if(s.Lattice.DiagonalOnly)
     {
+      o << "    PBC: bulk   Orthorhombic: yes   Type: PPPO" << std::endl;
```
prckent: Please split all of these codes "PPPO" etc. over at least two lines: (1) "PPP" is related to the system simulated and input file, while (2) the additional letters relate to the implementation. I prefer a less cryptic long form description "Using optimized orthorhombic implementation". This would also be consistent with the SoA text.

ye-luo: I still prefer it to be the same line and short. PPP info is already reflected as "PBC: bluk". O, the actual implementation, doesn't not need to be explained in the output file. It is better to mention it in the manual. Keep them in the same line make it easy for grep.

prckent: I don't want to ever have to explain these abbreviations e.g. at a school or in the manual. I want them to be separate and that is final.

markdewing: The boundary conditions are already printed out in the 'Simulation Cell' section.  What about a purely text line, like
```
Distance computations use orthorhombic periodic cell in 3D.
```

ye-luo: That is good. Just be sure to print DIM-D not just 3D.

markdewing:
Output now looks like
```
 Wavefunction setup:
 -------------------
  Name: psi0
  Created basis set builder named 'einspline' of type einspline
  Building SPOset '' with '' basis set.

  Distance table for similar particles (A-A):
    source/target: e
    Using structure-of-arrays (SoA) data layout
    Distance computations use general periodic cell in 3D with corner image checks.
    Setting Rmax = 1.9475


  Distance table for dissimilar particles (A-B):
    source: ion0  target: e
    Using structure-of-arrays (SoA) data layout
    Distance computations use general periodic cell in 3D with corner image checks.
    Setting Rmax = 1.9475

```

prckent:
Thanks Mark

PR #628, opened 2018-01-07
Size consistent Tmove

ye-luo:
Implements size consistent Tmove v1 from M.Casula 2010 paper. Issue #22.
Tested on water molecules sitting far away. A large time step 0.2 is used. Noticed 10-20% variance increase with v1.
compare

Manual is updated but I can't  build the bibliography part. @prckent what is the current recipe?
Three tests based on 2 water molecules are added. LA, TM0, TM1 have distinguishable energy values.
Some extra cleaning removing unused DMC drivers.


prckent:
@ye-luo Re: Manual. Does the script work for you? Due to #618 are use of bibtopic,  two bibtex calls are needed.

prckent:
Looks convincing. Is there any published data we can use for validation?

prckent:
I think the nonlocalmoves option is appropriate. We might want to change the default in future.

ye-luo:
[2010 paper]http://aip.scitation.org/doi/10.1063/1.3380831) has figure 2(a) which is an isolated O atoms test. The figure is similar to my H2O. I also checked acceptance like the fig 1.
@prckent I can't change the default silently since the result is different. Also GPU implementation is not done yet.

prckent:
OK on the validation.

I wasn't suggesting to change the default yet. Indeed we need the GPU version, and people also need to experiment in solid-state and gain confidence before we make a switch. This will take some months.

src/QMCHamiltonians/NonLocalTOperator.cpp
```diff
@@ -44,7 +45,30 @@ bool NonLocalTOperator::put(xmlNodePtr cur)
...
+    o << "  Using Non-local T-moves v0, M. Casula, PRB 74, 161102(R) (2006)";
+  }
+  else if(use_tmove=="v1")
+  {
+    v_tmove=TMOVE_V1;
+    o << "  Using Non-local T-moves v1, M. Casula et al., JCP 132, 154113 (2010)";
+  }
+  else
+  {
+    APP_ABORT("NonLocalTOperator::put uknown nonlocalmove option");
```
markdewing: spelling of 'unknown'.   Also it would be nice to print the given option in the error message.

manual/dmc.tex
```diff
@@ -37,7 +37,7 @@ \section{Diffusion Monte Carlo}
    &   \texttt{reconfiguration     } &  string  & yes/pure/other & no   & fixed population technique  \\
    &   \texttt{branchInterval      } &  integer  & $\ge 0$ & 1   & branching interval \\
    &   \texttt{substeps            } &  integer  & $\ge 0$ & 1   & branching interval \\
-   &   \texttt{nonlocalmoves       } &  string  & yes/other & no   & run with tmoves  \\
+   &   \texttt{nonlocalmoves       } &  string  & yes/v0/v1/other & no   & run with tmoves  \\
```
markdewing: It appears that 'other' is longer an option (I assume here that 'other' means 'a string other than yes or no or v0 or v1', and not the exact string 'other').   The new code gives an error for a string that is not yes,v0,v1, or no.

manual/dmc.tex
```diff
@@ -99,7 +99,7 @@ \section{Diffusion Monte Carlo}
 \item \texttt{substeps}.  Same as BranchInterval.
 
 
-\item \texttt{nonlocalmoves}. DMC driver for running Hamiltonians with non-local moves.  An typical usage is to simulate Hamiltonians with non-local psuedopotentials with T-Moves.  Setting this equal to false will impose the locality approximation.
+\item \texttt{nonlocalmoves}. DMC driver for running Hamiltonians with non-local moves.  An typical usage is to simulate Hamiltonians with non-local psuedopotentials with T-Moves.  Setting this equal to false will impose the locality approximation. `yes/v0' impmements the algorithm in the 2006 paper~\cite{Casula2006} and `v1' impmements the v1 algorithm in the 2010 paper~\cite{Casula2010}.
```
markdewing: spelling of 'implements'

manual/dmc.tex
prckent: Based on the manual description I am happy with the new error code and deprecating "other". I don't believe that it was widely used and find the vagueness dangerous. In my experience the nonlocalmoves parameter is only specified to enable it, "yes", so the number of input file affected by this change will be tiny if not zero. Reasonable?

ye-luo: ''other'' means anything not ''yes''. After the change, the code will complain. We only used ''yes'' to access the Tmoves. Sometimes when I need to switch it off without deleting the input line, I put ''no'' as a natural response. So the number of input files affected should be zero or it is a user mistake.

jtkrogel:
Someday it would be nice to quantitatively reproduce Fig. 2b in Casula's 2010 paper, but they are using a Jastrow factor that we don't have and do not record the wavefunction used in enough detail for quantitative reproduction.  Possibly Michele still has it.

This eventuality should not affect this PR though.

manual/dmc.tex
prckent: Add "The v1 algorithm is size-consistent and an important advance over the previous v0 non-size consistent algorithm. Investigating the importance of size consistency is highly recommended."

ye-luo:
@jtkrogel v1 is working pretty well. S16 no Tmove, t=0.0125 is a bit wired.
tstep

jtkrogel:
@ye-luo That is nice to see.  The LA behavior is indeed a bit weird looking.  Can you plot S16-S1 (size consistency error) as a function of timestep?  I want to see how v0 is doing for t<0.03/Ha.

ye-luo:
The error is small at t<0.03Ha region but this H2O.
tstep_diff

ye-luo:
@prckent 
1. I don't think we need long tests. I made the short tests able to distinguish three cases, no/v0/v1. In fact, the short tests on LiH are not long enough to tell the difference of k points and Tmoves.
2. This is not a H2O_dimer, it is two H2O molecules sitting very far.
3. I will add more description there
4. Will do.

prckent:
1. In practice, the role of the short runs is to see if anything is strongly broken since the error bars are usually too big to determine subtle changes. The role of the long runs is for a non-expert to be able to see if something has changed consequentially. Please add a long test. 
2.  "H2O_dimer_sep"
3. OK.
4. Thanks.

ye-luo:
1. Will do, I will make the long ones generating 10x statistics.

ye-luo:
@prckent all the requested changes are made.

Issue #631, opened 2018-01-09
data race in nonblocking send/recv implementation


ye-luo:
I realized there is a data race bug in the nonblocking send/recv implementation.
When the same walker is sent to multiple ranks, encoding the number of sent copies in the message has a data race.
This will cause MPI hanging on the receiving side.
expect to get it fixed in about a week.

ye-luo:
This is bug doesn't show up with the default maxCopy=2. Only when you use maxCopy>2, it causes a problem. In the case of maxCopy=2, you can only send 1 walker + 1copy to 1 node or  1 walker + 0 copy to 2 nodes. No data race.

prckent:
Q. Relating to the previous as well as current implementation. Why is maxCopy not infinity, or at least a much larger value? (Say targetwalkers/no_mpi_tasks)  Would the improvement in load balancing not always win out over the reduced messaging/copying? We need to think about atomic runs as well as large solid state runs when setting default parameters...

ye-luo:
maxCopy is per walker not per node. So on a node by default you can get maximum 2x walkers before the load balancing. I think it is mainly to maintain a stable population and avoid the ensemble being destroyed by bad walkers.

prckent:
Just set to 2 if >2 requested for now. This area needs to be looked at sometime and sensibly studied/documented.

PR #632, opened 2018-01-09
Limit maxCopy due to nonblocking send/recv bug

ye-luo:
For #631 
The default maxCopy=2 doesn't hit the bug. But I think it is better to have code on the safe side.
In this PR, I just disabled the nonblocking.
if disabling only maxCopy>2 + nonblocking is preferred, I can change this PR.


ye-luo:
I'm thinking to change the behaviour as
keep nonblocking = yes as default. When the code sees maxCopy>2, switch the nonblocking off and print a warning. So people can still get the benefit of the nonblocking during this period period.
What is your preference @prckent ?

prckent:
Set maxcopy=2 if maxCopy is requested >2 and print a warning. The code will run and the performance will be good in almost all circumstances.

PR #640, opened 2018-01-12
Use h5diff to compare HDF output files

markdewing:
The FindHDF module sets the variable HDF5_DIFF_EXECUTABLE.
This should the make the tests independent of variations in the output of h5dump.


anbenali:
Fantastic!!

prckent:
@anbenali 
This does a full machine precision diff. Is that appropriate? If the coefficients are only copied, I think yes. If they are transformed/modified in any way, no.

anbenali:
The coefficients are only copied. No transformation is done which is left
either inside the codes (gamess, qp or pyscf) or inside qmcpack. The
converter should never touch them. Therefore I think it is safe to move
forward!

prckent:
OK, merging. Lets see how the nightlies fare.

prckent:
Thanks @markdewing

PR #643, opened 2018-01-17
Skip git-rev files for non-git builds.

markdewing:
For builds that do not come from a git repository, the git-rev.h file is empty.  Change the code so git-rev.h and git-rev-tmp.h are not referenced, and are not even created.

Hopefully this can address a Spack issue where sometimes either file is not found.


ye-luo:
I'm thinking of some simplification.
In the main CMakeList.txt, if git is not found, set IS_GIT_PROJECT false.
Then in the src/CMakeList.txt, there is no need to introduce another variable, always use IS_GIT_PROJECT

markdewing:
That's a good idea - I'll make the change.

PR #646, opened 2018-01-22
Modified the k-grid for the momentum estimator and removed compQ

kylanpaa:
1. The compQ was removed: it can be post-processed rather straightforwardly from n(k).
2. Modified the input options for n(k):
  (a) kmax: desired k-points reside within the sphere of radius kmax
  (b) kmax0, kmax1, kmax2: maximum k-values for a parallelepiped in which the k-points are located
  (c) both (a) and (b): includes k-points within the sphere and those of the parallelepiped outside the sphere
  (d) default behaviour (i.e., not defining any kmax values): uses (a) with kmax=2.0 x k_F, where k_F is the Fermi momentum of polarized non-interacting electron gas at corresponding electron density

### DIFF
--- a/src/QMCHamiltonians/MomentumEstimator.cpp
+++ b/src/QMCHamiltonians/MomentumEstimator.cpp
@@ -27,7 +27,7 @@ namespace qmcplusplus
 {
 
 MomentumEstimator::MomentumEstimator(ParticleSet& elns, TrialWaveFunction& psi)
-  :M(1), refPsi(psi), kgrid(4), Lattice(elns.Lattice), norm_nofK(1), hdf5_out(false)
+  :M(40), refPsi(psi), Lattice(elns.Lattice), norm_nofK(1), hdf5_out(false)
 {
   UpdateMode.set(COLLECTABLE,1);
   psi_ratios.resize(elns.getTotalNum());
@@ -81,14 +81,6 @@ MomentumEstimator::Return_t MomentumEstimator::evaluate(ParticleSet& P)
                        - ( phases_s[ik]*phases_vPos_c[ik] + phases_c[ik]*phases_vPos_s[ik] ) * ratio_s ;
     }
   }
-
-  compQ=0.0;
-  for (int iq=0; iq<compQ.size(); ++iq)
-  {
-    for (int i=0; i<mappedQtonofK[iq].size(); ++i)
-      compQ[iq] += nofK[mappedQtonofK[iq][i]];
-    compQ[iq] *= mappedQnorms[iq];
-  }
   for (int ik=0; ik<nofK.size(); ++ik)
     nofK[ik] *= norm_nofK;
   if (hdf5_out)
@@ -97,8 +89,6 @@ MomentumEstimator::Return_t MomentumEstimator::evaluate(ParticleSet& P)
     int j=myIndex;
     for (int ik=0; ik<nofK.size(); ++ik,++j)
       P.Collectables[j]+= w*nofK[ik];
-    for (int iq=0; iq<compQ.size(); ++iq,++j)
-      P.Collectables[j]+= w*compQ[iq];
   }
   return 0.0;
 }
@@ -118,13 +108,6 @@ void MomentumEstimator::registerCollectables(std::vector<observable_helper*>& h5
     h5o->addProperty(const_cast<std::vector<PosType>&>(kPoints),"kpoints");
     h5o->addProperty(const_cast<std::vector<int>&>(kWeights),"kweights");
     h5desc.push_back(h5o);
-    //add compQ
-    ng[0]=Q.size();
-    h5o=new observable_helper("compQ");
-    h5o->set_dimensions(ng,myIndex+nofK.size());
-    h5o->open(gid);
-    h5o->addProperty(const_cast<std::vector<RealType>&>(Q),"q");
-    h5desc.push_back(h5o);
   }
 }
 
@@ -135,7 +118,6 @@ void MomentumEstimator::addObservables(PropertySetType& plist, BufferType& colle
   {
     myIndex=collectables.size();
     collectables.add(nofK.begin(),nofK.end());
-    collectables.add(compQ.begin(),compQ.end());
   }
   else
   {
@@ -144,12 +126,6 @@ void MomentumEstimator::addObservables(PropertySetType& plist, BufferType& colle
     {
       std::stringstream sstr;
       sstr << "nofk_" <<i;
-      int id=plist.add(sstr.str());
-    }
-    for (int i=0; i<Q.size(); i++)
-    {
-      std::stringstream sstr;
-      sstr << "Q_" <<i;
       int id=plist.add(sstr.str());
     }
   }
@@ -162,7 +138,6 @@ void MomentumEstimator::setObservables(PropertySetType& plist)
   if (!hdf5_out)
   {
     copy(nofK.begin(),nofK.end(),plist.begin()+myIndex);
-    copy(compQ.begin(),compQ.end(),plist.begin()+myIndex+nofK.size());
   }
 }
 
@@ -172,7 +147,6 @@ void MomentumEstimator::setParticlePropertyList(PropertySetType& plist
   if (!hdf5_out)
   {
     copy(nofK.begin(),nofK.end(),plist.begin()+myIndex+offset);
-    copy(compQ.begin(),compQ.end(),plist.begin()+myIndex+nofK.size()+offset);
   }
 }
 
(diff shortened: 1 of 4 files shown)


markdewing:
okay to test

src/QMCHamiltonians/MomentumEstimator.cpp
```diff
@@ -172,115 +147,271 @@ void MomentumEstimator::setParticlePropertyList(PropertySetType& plist
...
+  pAttrib.add(M,"samples"); // default value is 40 (in the constructor)
   pAttrib.put(cur);
   hdf5_out = (hdf5_flag=="yes");
+  RealType min_Length=1.0e10;
+  PosType vec_length(0);
+  for (int i=0; i<OHMMS_DIM; i++)
+  {
+    PosType a(0);
+    for (int j=0; j<OHMMS_DIM; j++)
+      a[j]=elns.Lattice.G(j,i);
```
ye-luo: G=inverse(R). Not understanding why you take the columns instead of rows.

src/QMCHamiltonians/MomentumEstimator.cpp
```diff
@@ -172,115 +147,271 @@ void MomentumEstimator::setParticlePropertyList(PropertySetType& plist
...
+    //length of reciprocal lattice vector
+    vec_length[i]=2.0*M_PI*std::sqrt(dot(a,a));
+    if (min_Length>vec_length[i])
+      min_Length=vec_length[i];
+  }
```
ye-luo: minimal G vector length may not be a good choice. I feel more safe to use WS radius in G space.

src/QMCHamiltonians/MomentumEstimator.cpp
```diff
@@ -172,115 +147,271 @@ void MomentumEstimator::setParticlePropertyList(PropertySetType& plist
...
-  if (twist[2]==0)
-    mappedQnorms[5*kgrid+2]=qn/RealType(M);
-  Q.resize(numqtwists);
-  for (int i=-kgrid; i<(kgrid+1); i++)
+  PosType kmaxs(0);
+  kmaxs[0]=kmax0;
+  kmaxs[1]=kmax1;
+  kmaxs[2]=kmax2;
+  bool sphere_only=true;
+  bool directional_only=true;
```
ye-luo: All the XXX_only logics are messy. _only implicitly impose the state of the other. The I saw later !sphere_only was doing direction. !direction_only is doing sphere.

src/QMCHamiltonians/MomentumEstimator.cpp
```diff
@@ -172,115 +147,271 @@ void MomentumEstimator::setParticlePropertyList(PropertySetType& plist
...
+    kmax = 2.0*std::pow(6.0*M_PI*M_PI*elns.getTotalNum()/elns.Lattice.Volume,1.0/3);
+    kgrid = int(kmax/min_Length)+1;
+    sphere_kmax=kmax;
+    sphere_only=true;
+    directional_only=false;
+  }
+  else
+  {
+    sphere_kmax=kmax;
+    if (sum_kmaxs>0.0 && kmax>0.0)
```
ye-luo: This part of logic can be simplified.

src/QMCHamiltonians/MomentumEstimator.cpp
```diff
@@ -172,115 +147,271 @@ void MomentumEstimator::setParticlePropertyList(PropertySetType& plist
...
+    }
+    kgrid = int(kmax/min_Length)+1;
   }
-  app_log()<<" Using all k-space points with (nx^2+ny^2+nz^2)^0.5 < "<< kgrid <<" for Momentum Distribution."<< std::endl;
-  app_log()<<"  My twist is:"<<twist[0]<<"  "<<twist[1]<<"  "<<twist[2]<< std::endl;
-  int indx(0);
-  int kgrid_squared=kgrid*kgrid;
+  int kgrid_squared[OHMMS_DIM];
+  for (int i=0; i<OHMMS_DIM; i++)
+    kgrid_squared[i]=int(std::round(kmaxs[i]*kmaxs[i]/vec_length[i]/vec_length[i]));
```
ye-luo: should be float and no int(round())

src/QMCHamiltonians/MomentumEstimator.cpp
```diff
@@ -172,115 +147,271 @@ void MomentumEstimator::setParticlePropertyList(PropertySetType& plist
...
-        if (i*i+j*j+k*k<=kgrid_squared) //if (std::sqrt(i*i+j*j+k*k)<=kgrid)
+        PosType kpt;
+        kpt[0]=i-twist[0];
+        kpt[1]=j-twist[1];
+        kpt[2]=k-twist[2];
+        //convert to Cartesian: note that 2Pi is multiplied
+        kpt=Lattice.k_cart(kpt);
+        bool not_recorded=true;
+        // This collects the k-points within the parallelepiped (if not in sphere_only mode)
+        if (!sphere_only && i*i<=kgrid_squared[0] && j*j<=kgrid_squared[1] && k*k<=kgrid_squared[2])
```
ye-luo: Need to include the twist

src/QMCHamiltonians/MomentumEstimator.h
```diff
@@ -77,19 +75,16 @@ class MomentumEstimator: public QMCHamiltonianBase
   std::vector<int> kWeights;
   ///dims of a grid for k points
   int kgrid;
+  ///maximum k-value in the k-grid in cartesian coordinates
+  RealType kmax;
```
ye-luo: If they are only used inside putSpecial, move them there.

src/QMCHamiltonians/MomentumEstimator.cpp
kylanpaa: In CrystalLattice.h "///Reciprocal unit vectors. G(j,i) i=vector and j=x,y,z".

src/QMCHamiltonians/MomentumEstimator.cpp
kylanpaa: This is only used for getting an overestimation for the for-loop bounds such that definitely all desired k-points will be collected. I suspect that WS radius would not be adequate here.

src/QMCHamiltonians/MomentumEstimator.cpp
kylanpaa: I agree that the kgrid* variables could be better named as ngrid*, but I followed the old convention in which these are actually integers.

src/QMCHamiltonians/MomentumEstimator.cpp
kylanpaa: The twist is included in the actual k-point, and therefore, for the parallelepiped the grid will be a twist shifted parallelepiped when done this way. This is what Jaron and I thought would be best in general, and also, it is convenient for our next simulations. To this end, this way the values kmax0, kmax1, and kmax2 are not absolute limiting values, but e.g. kmax0+|twist0| is possible.

src/QMCHamiltonians/MomentumEstimator.h
kylanpaa: Currently this is true. For post-processing purposes it could be good to include more information in the h5-files, at some point.

src/QMCHamiltonians/MomentumEstimator.cpp
jtkrogel: G=inverse(transpose(R))

src/QMCHamiltonians/MomentumEstimator.cpp
jtkrogel: Shortest length k-space axis => largest needed grid size

src/QMCHamiltonians/MomentumEstimator.cpp
jtkrogel: Attempt here is to do one of three things: 1) k-points in a sphere, 2) k-points in a parallelepiped, 3) both at the same time (better than making a second estimator).

Fusing into a single loop+logic is somewhat ugly but obviates the need to track (e.g. in a set/hash table) which k-points have been selected already.

jtkrogel: Agree that naming+logic can be improved.

src/QMCHamiltonians/MomentumEstimator.cpp
ye-luo: Example,
```
g0=[0.5  0.0  0.0]
g1=[0.5  0.3  0.0]
g2=[0.0  0.0  1.0]
```
Shortest k-space axis is 0.5 but the WS_radius=0.3, the shortest vector is actually g1-g0.

src/QMCHamiltonians/MomentumEstimator.cpp
ye-luo: @jtkrogel where in the code?
I found CrystalLattice.cpp:113
```
  G = inverse(R); //G = transpose(Inverse(R));
```
transpose is commented

src/QMCHamiltonians/MomentumEstimator.cpp
ye-luo: I don't mean fusing into 1 logic. The problem is ! XXX_only double negative. I think you can directly use sphere and directional instead of the opposite meaning.

src/QMCHamiltonians/MomentumEstimator.h
ye-luo: I think even when you write them to h5, these meta data need to be written in putSpecial.

src/QMCHamiltonians/MomentumEstimator.cpp
kylanpaa: Theoretically transpose(G)R=2pi x identity_matrix. In QMCPACK the 2pi is omitted. Therefore, inverse(R)=transpose(G)=[a,b,c], where the column vectors a, b, and c are the reciprocal space vectors.

src/QMCHamiltonians/MomentumEstimator.cpp
ye-luo: The sphere kmax checking includes twist so it need to be consistent.
I think the checking should be (i-twist[0])*(i-twist[0])<=kgrid_squared[0]

src/QMCHamiltonians/MomentumEstimator.cpp
ye-luo: I was thinking of twists. Since the twists will be added later. rounding here will cause bugs.

src/QMCHamiltonians/MomentumEstimator.cpp
ye-luo: @kylanpaa You are right. The definition requires a transpose.

ye-luo: Then please use Gv to simplify your code.

src/QMCHamiltonians/MomentumEstimator.cpp
kylanpaa: If it always leads to a smaller value, then it should be ok to use that also, but not necessary. Since kmax0 refers to the maximum k-value along the g0 direction, then the nmax0 would be less than int(kmax0/|g0|)+1. The same applies to all directions. Now, if one chooses the maximum of those nmax0, nmax1, and nmax2, then one should definitely go through all possible k-points... also the shortest vector you showed will be included, since the loop goes from -nmax to nmax.

src/QMCHamiltonians/MomentumEstimator.cpp
kylanpaa: In the (i-twist[0]) case the kgrid_squared does need to be float, which is what you referred to in your previous comment.

src/QMCHamiltonians/MomentumEstimator.cpp
ye-luo: In principle, I don't think -nmax to nmax is sufficient if nmax is defined by the min_length.
```
g0=[0.5  0.0  0.0]
g1=[0.5  0.1  0.0]
g2=[0.0  0.0  1.0]
```
g1-g0=0.1 and you need 5x nmax to include all the points in the sphere.
However, in QMCPACK we use translation symmetry only up to 27 corners. So probably if such extreme case happens, many parts of the code already break down. So probably using the min_Length is OK within our assumption.

kylanpaa: Yes, you are right. In general it will be better to use the WS_radius.

src/QMCHamiltonians/MomentumEstimator.cpp
```diff
@@ -229,36 +218,37 @@ bool MomentumEstimator::putSpecial(xmlNodePtr cur, ParticleSet& elns, bool rootN
...
       {
-        PosType kpt;
-        kpt[0]=i-twist[0];
-        kpt[1]=j-twist[1];
-        kpt[2]=k-twist[2];
+        PosType ikpt,kpt;
+        ikpt[0]=i-twist[0];
+        ikpt[1]=j-twist[1];
+        ikpt[2]=k-twist[2];
+        kpt=ikpt;
```
ye-luo: No need this line. Overwritten 2 lines later.

src/QMCHamiltonians/MomentumEstimator.h
```diff
@@ -69,18 +69,12 @@ class MomentumEstimator: public QMCHamiltonianBase
   VectorSoaContainer<RealType,2> phases;
   ///phases of vPos
   std::vector<VectorSoaContainer<RealType,2> > phases_vPos;
+  ///dims of a grid for k points
+  int kgrid;
```
ye-luo: Can we move this also to putSpecial if is not used anywhere else?

src/QMCHamiltonians/MomentumEstimator.cpp
kylanpaa: ikpt is used in the directional. So, instead of (i-twist[0])*(i-twist[0]) in the if I use ikpt[0]*ikpt[0]. kpt is the ikpt in Lattice.k_cart with 2*pi included, which is used in the push_pack and in the sphere case.

src/QMCHamiltonians/MomentumEstimator.h
kylanpaa: It's used in the makeClone function, which I did not change. Therefore, if it is still desired to be there then we cannot move it.

src/QMCHamiltonians/MomentumEstimator.cpp
ye-luo: Yes.

src/QMCHamiltonians/MomentumEstimator.cpp
```diff
@@ -172,115 +151,236 @@ void MomentumEstimator::setParticlePropertyList(PropertySetType& plist
...
   {
     for (int j=-kgrid; j<(kgrid+1); j++)
     {
-      if (i*i+j*j<=kgrid_squared) //if (std::sqrt(i*i+j*j+k*k)<=kgrid)
+      PosType ikpt,kpt;
+      ikpt[0]=i-twist[0];
+      ikpt[1]=j-twist[1];
+      kpt=ikpt;
+      //convert to Cartesian: note that 2Pi is multiplied
+      kpt=Lattice.k_cart(kpt);
```
ye-luo: The same here. Thanks.

src/QMCHamiltonians/MomentumEstimator.h
ye-luo: Remove also the line in makeClone.

kylanpaa: Yes, I don't think there is any reason why the clone would need to have that information, so it can be put in the putSpecial and removed from the makeClone.

ye-luo:
I can not approve because of my pushed changes.
But all the changes from @kylanpaa are good.

src/QMCHamiltonians/MomentumEstimator.cpp
jtkrogel: We will probably revert this change in a subsequent PR.  It is better to have a uniform grid size on all twist groups and this is better for postprocessing also.

ye-luo: Apart from postprocessing convenience, is there any other reason for physics?

PR #651, opened 2018-01-25
Improve the shift_s adjustment scheme in OneShiftOnly

ye-luo:
Based on the accept/reject pattern, update the shift_s adjustment base.
This accelerates the optimization in hard cases (multi determinants).


markdewing:
Could you describe 'accept_history' in more detail? What is the allowed range? (From the code, it looks like 0-3?)  What do the values mean? (It looks like the initial value is 3.  The if-test for the value does something different if the value is 1)

prckent:
What systems were tested? Is the optimization unchanged for small/trivial systems?

ye-luo:
@markdewing the last 2 iterations, value 0(00), 1(01), 2(11), 3(10)

ye-luo:
@prckent This only kicks in when you have reject,accept,reject sequence. No change to the trial system like Jastrow only optimization which has almost zero rejection.

ye-luo:
Can we get this merged?

prckent:
Not before updated release. This is a potentially significant change.

markdewing:
The values of 'accept_history' - is 1 accept and 0 reject?
if the last two bits are the history, then '10' is 2 and '11' is 3?

Using an integer as a bit vector of length 2 is something that requires more documentation.  Using a std::bitset of size 2 might be a little more clear (no need for the %4 to truncate the leftmost bits).
Acceptance could be written as
```
accept_history = accept_history << 1;
accept_history.set(0, true);
```
or if you really want it on one line
```
accept_history = (accept_history << 1)  | std::bitset<2>(1);
```



Backing up further, the OneShiftOnly optimizer bases accept or reject on whether the cost function is valid (IsValid), which is based on the effective number of samples (or if the new cost is NaN). (I finally read the manual.)

The decision to update the shift base is determined by the history of 3 iterations - the current one, and the two past ones (stored in accept_history)
Why does it help in this particular case (reject,accept,reject)?  It appears to reduce the shift base, which is used to increase the shift on rejection, and decrease the shift upon acceptance.

ye-luo:
shift_s needs to be as small as possible to let the optimization go fast. But if too small, you get a rejection. When you get rejection,accept,rejection, your shift_s_base resolution is not enough. If you keep shift_s_base unchanged, you get alternating accept/reject.

PR #657, opened 2018-01-29
Hybrid representation

ye-luo:
Add a smooth region in computing.
Significantly improved builder parallelization and efficiency.
Some bug and compatibility fixes.


lshulen:
Thanks for this commit Ye.  I've started reading through the commits, but as this is rather large it is taking some time.  As I do this, is there anything in particular that you would like me to look for?

ye-luo:
SplineHybridAdoptorReaderP.h is the one containing large modification in which I changed the parallelization and computing scheme.

prckent:
Thanks.

How should we check that this "works" and does not break in future? Do we have a sensitive test?

Is the smoothing consistent in CPU and GPU versions? Are there parameters that need to be set in an updated manual entry?

ye-luo:
There are test cases labelled as hybridrep which you can try. They failed once when I touched the active particle and I consider them sensitive enough.
Unfortunately, there is not smooth consistent between CPU and GPU, it needs to be reworked.
A complete manual update will be provided when I got the paper submitted.

lshulen:
I'm looking forward to getting this in the main code base.  I'm a little unfamiliar with the details of the PR process.  It seems like everything has happened to get it approved, but it is not merged.  Is there any chance this could get merged today?

Issue #579, opened 2017-12-15
Integration test: k-space Jastrow


jtkrogel:

jtkrogel:
See #652 for broader context.

jtkrogel:
Some examples containing kspace Jastrow inputs exist (Si2 Abinit 2008, Be8 QE 2011), but the parameters are identically zero.

PR #662, opened 2018-01-30
Add test for k-space Jastrow

jtkrogel:
Addresses #579.  Long and short tests are added for the k-space Jastrow along with manual input entry, README file, and a small fix to Nexus I/O for this case.  

For the small (2 atom) diamond cell in the test, addition of the k-space Jastrow significantly reduces the variance and also lowers the total energy by about 4 mHa/atom.  See README for more detail.

The current version of QMCPACK passes the tests.


ye-luo:
Do you know if the optimization works with current code? Just to know if the evaluateDerivative routine is correct or not.

rcclay:
The e-e k-space jastrow passed the wavefunction tester the last time I checked.

jtkrogel:
@ye-luo I just reran the optimization with 3.40 AoS and SoA.  Both work well.

Anything else needed for this PR?

ye-luo:
No. Everything seems good and the testing time is reasonable. I will leave mark to handle the merge sequence.

PR #663, opened 2018-01-31
Create densityestimator.hpp

kencat:
This is a new estimator of spin density.


ye-luo:
I will handle this locally.

Issue #574, opened 2017-12-15
Integration test: cslinear optimizer


jtkrogel:

ye-luo:
Do we still need it?

prckent:
If we want to consider dropping something (== removing the code, manual, tutorial entries etc) we need a wider discussing in a dedicated issue. We also need to verify that the replacement works better, has better documentation, that the tutorials are updated etc.

PR #664, opened 2018-02-01
Add tests for cslinear

jtkrogel:
Addresses #574.

Add quartic, rescale, and linemin tests for cslinear.  Also add missing rescale and linemin tests for linear.

New tests are based off of existing H4 optimization tests (identical requirement for final optimized energy).


prckent:
OK to test

ye-luo:
@jtkrogel for your knowledge. What is the main difference between cslinear and linear? Is it useful?
The current class hierarchy in optimization classes are quite messy and I would really like to remove the parts no more useful in the near future.

jtkrogel:
My impression is that it dealt with how correlated sampling was handled for the linemin.  This part may have been removed in May 2012:

https://github.com/QMCPACK/qmcpack/commit/33fe815c2925e4a746940ea55c48710cde862433#diff-a0316c23075984c74ad219f0dd0fd6d8

Both linear and cslinear were actively developed up to May 2012.  Probably @mmorale3 can say more about the actual differences.

From a practial perspective, I would like to see all optimizers run on a common set of test problems prior to deciding what to remove, etc.

mmorale3:
cslinear was meant to be an alternative to linear that generated the structures used in the linear method on the fly directly in the VMC calculation, rather than saving samples and post-processing. The point was to be able to run as long as needed to achieve a given statistical error in the matrix elements, which is very hard to control with samples. Jeremy did most of the work on this. I don't know if this still works. If no one in the current team has worked on it, it probably doesn't work.

jtkrogel:
cslinear still works in the limited case I tried it for; it reproduces the energies currently used to test linear and one-shift-only.

ye-luo:
I see the point of @mmorale3. I think the philosophy is right. But the current implementation is largely duplicating codes and then not well maintained. For example, the cslinear is using VMCLinearOptOMP under VMC. We should minimize the lines of code but unfortunately, this is not high priority.

PR #667, opened 2018-02-03
Generate code for H and I shells in CartesianTensor.h

markdewing:
Add H and I shells to CartesianTensor.h.
Now the computations in the evaluate* functions in the entire file are generated by the codegen/gen_cartesian_tensor.py script.

The tests in test_cartesian_tensor.cpp test a subset of values.
A more complete test can be done with test_full_cartesian_tensor.cpp.  To enable this test, generate it with the gen_gto.py script.  Then uncomment a line in the CMakeLists.txt file adding test_full_cartesian_tensor.cpp to the unit test sources.
This file can take quite a long time to compile, especially if optimization is enabled.

This is the first PR to add code generation to the main code (vs. using to create unit test values).  (Though comments in CartesianTensor.h indicated some of the higher derivatives were auto-generated in the past)

Comments on adding code generation:
1. The code generation uses a Python script that uses Sympy for the symbolic differentiation.
2. The proposed structure is to create a codegen/ subdirectory to hold the scripts and input files to re-created the targets (CartesianTensor.h)
3. It is important that others understand (and could maintain) the code in codegen/gen_cartesian_tensor.py.    Please review this file.
4. Code generation can easily become messy - one consequence (or symptom) is that it is hard to understand how changes in the input script flow through to changes in the output file.

This file actually uses two steps of code generation, but only the second is automated.
The first step is to read the order (from order.txt, which is snippet from a Gamess output file) with read_order.py, which can output a python version, get_ijk, which is pasted into gen_cartesian_tensor.py and ../tests/gen_gto.py.  The read_order.py script also produces a C++ version, get_ABC, which is pasted into CartesianTensor.h.in.

The second steps fills in the bodies of the evalute* functions, and gen_cartesian_tensor.py does that automatically.   It produces the CartesianTensor.h in codegen/, and it manually needs to be moved up on directory.


markdewing:
See this Jupyter notebook to see more about Gaussian Type Orbitals expressions used in the code generation scripts (and to see the math pretty-printed)
https://github.com/QMCPACK/qmc_algorithms/blob/master/Wavefunctions/GaussianOrbitals.ipynb

prckent:
@anbenali is requesting an SoA implementation.

prckent:
@anbenali Is this a safe change to the GAMESS converter?

grahamlopez:
test this please

grahamlopez:
test this please

prckent:
@anbenali ping.

Issue #571, opened 2017-12-15
Integration test: RPA Jastrow


jtkrogel:

jtkrogel:
See #652 for broader context.

jtkrogel:
Old QMCPACK builds (2010-2014, i.e. SVN revisions 4648-6259) segfault during the initialization of the LR part (HEG 54 rs=5 example from 2008) in RPAJastrow::buildOrbital.

Some reference data does exist in this case (VMC/RMC/RMC):
HEG.rs5.rpa  series 1  LocalEnergy           =  -4.216503 +/- 0.014584 
HEG.rs5.rpa  series 2  LocalEnergy           =  -4.258910 +/- 0.005513 
HEG.rs5.rpa  series 3  LocalEnergy           =  -4.255253 +/- 0.003184

jtkrogel:
In another old example (Si2 Abinit+J2 RPA, short range only), builds starting in 2012 (SVN 5488, older ones not built with internal einspline) get past the initialization and then abort due to missing interface functions in VMC:

"Fatal Error. Aborting at OrbitalBase::evalGradient is not implemented"

Reference data does not exist for Si2 with RPA.  The bare input (no J2 RPA, Bspline only) works fine with revisions 5488 and 5594, producing output in agreement for existing B-spline reference data.

jtkrogel:
I recommend using the existing HEG reference data to make a developer test (not run during regular testing).

PR #669, opened 2018-02-08
RPA Jastrow manual description and test

jtkrogel:
Addresses #571. PR includes documentation for the Gaskell RPA and Yukawa Jastrows, including literature references.  A "developer" test is included to aid the restoration of this Jastrow.

Description below is from the README file:

This test is to be used to resurrect the two-body RPA Jastrow.
The RPA Jastrow was used in the original test set for QMCPACK (circa
2008) and it has been broken in QMCPACK since at least 2010.  Recent
efforts were made (2016) to resurrect the RPA Jastrow, but the current
test case results in an unphysically high energy and variance.  The
current test is intended for developer use only and not as a test of
the correctness of the current code.  As such it should not be run
in the e.g. nightly tests until the issue is fixed.

The test is for a 54 electron HEG in 3D at rs=5.0.  A two-body
Gaskell RPA Jastrow factor is used, including both short and long
ranged parts.  The original output data includes results for VMC
and RMC, though only the VMC inputs are used for the test at this
time.  All of the original test files are included in ./qmc-ref.

The total energy used in this test was obtained from the original
data with qmca:

>qmca -q e -e 2 qmc-ref/*s001*scalar*
qmc-ref/HEG.rs5.rpa  series 1  LocalEnergy = -4.209725 +/- 0.012941

The original scalar files do not contain the square of the local
energy, precluding a test of the variance.


prckent:
OK to test

PR #670, opened 2018-02-12
[WIP] Correctly read GAMESS output with merged numbers in eigenvector coefficients

### LINKED ISSUE #666: Error reading molecular orbitals from GAMESS
I'm using the convert4qmc script to generate a wave function xml file from GAMESS. In one case, some of the virtual molecular orbitals have very large coefficients. In cases where these are negative and have magnitude greater than 100.0, there is a quirk in the printing where orbital coefficient are merged together - here's an example, see the "-103.88..." on the final line:

  102  C  5  Z    0.000000   0.000000  -0.215939   0.071102   0.000706
  103  C  5  X    0.000000   0.000000  -8.413034 -73.890437  10.713047
  104  C  5  Y   -1.957422   0.366991  -0.000000  -0.000000   0.000000
  105  C  5  Z    0.000000   0.000000  -0.567800-103.884284  -2.142253

The convert4qmc script doesn't deal with this properly, and the corresponding coefficients for that MO look like this in the QMCPack xml file:

7.11020000000000e-02 -7.38904370000000e+01 -0.00000000000000e+00 -5.67800000000000e-01

i.e., the -103.88 is replaced by the -0.5678 before it.

For multi-determinant calculations that make use of the virtual orbitals, this gives some badly incorrect energies. I'm not sure if you're aware of this, but I thought I should at least point it out. It could cause problems if a user doesn't spot it. Although the energy was clearly wrong in my current case, perhaps it may be more subtle in others.

I can send the GAMESS file if helpful.

prckent:
Closes #666 . This bug is triggered when the eigenvector output of GAMESS has "merged" numbers due to use of a fixed Fortran format 123.45-123.45.

When reading eigenvector coefficients, parse line using our existing code, then use regex to split words with pattern .\*[0-9]-[0-9].\*   .  Checked with benzonitrile from #666 .

Must be reviewed by @anbenali and checked with other GAMESS outputs. Could the regex match  any high-angular momentum basis set labels and cause problems?

converter_HDF5_FeCO6 test fails on my dev box, but this also fails with mainline develop (?).

### DIFF
--- a/src/Utilities/SimpleParser.cpp
+++ b/src/Utilities/SimpleParser.cpp
@@ -146,9 +146,9 @@ int getwordsWithMergedNumbers(std::vector<std::string>& slist, std::istream &fp,
 	  {
 	    for (size_t i = 1; i < ftn_match.size(); ++i) {
 	      std::ssub_match sub_match = ftn_match[i];
-	      std::string piece = sub_match.str();
+	      //	      std::string piece = sub_match.str();
 	      //	      std::cout << "  submatch " << i << ": " << piece << '\n';
-	      //	      splitlist.push_back(sub_match);
+	      splitlist.push_back(sub_match);
 	    }   
 	  }
 	else


ye-luo:
converter_HDF5_FeCO6 was failing only on oxygen but not other machines. Could you manually do a h5diff -r and see what is actually different?

anbenali:
@ye-luo  I am testing it now on all Gamess cases..

ye-luo:
@anbenali That is a specific problem on oxygen box and only @prckent can provide info.

markdewing:
This seems like a temporary solution.  If GAMESS outputs fixed-width fields, it seems the correct solution is   to read fixed-width fields.

jtkrogel:
I've manually changed the field width of GAMESS at compile time before to get higher precision.  If we set a fixed width then that approach will no longer work.

Still, fixed width read could be faster if there are a lot of merged columns.  Having a fixed width read with this as a fallback would be OK down the road.

src/Utilities/SimpleParser.cpp
```diff
@@ -130,6 +129,39 @@ int getwords(std::vector<std::string>& slist, std::istream &fp, int dummy /* = 0
...
+      std::regex ftn_regex("(.*[0-9])(-[0-9].*)");
+      std::smatch ftn_match;
+      for (const auto &item : slist) {
+	if (std::regex_match(item, ftn_match, ftn_regex))
+	  {
+	    for (size_t i = 1; i < ftn_match.size(); ++i) {
+	      std::ssub_match sub_match = ftn_match[i];
+	      std::string piece = sub_match.str();
+	      //	      std::cout << "  submatch " << i << ": " << piece << '\n';
+	      //	      splitlist.push_back(sub_match);
```
markdewing: Isn't this line necessary for this code to work? (And shouldn't it be `splitlist.push_back(piece)`?)  Otherwise, how do the matched items end up in the list of split strings?

prckent:
Indeed this is a temporary solution. Now I have read the code I think there is considerable additional fragility that should also be improved when the converter is updated more thoroughly.  Clearly it was not written with current demands in mind.

To improve robustness, the reader should check where it is (e.g. check the orbital index in the output vs where it thinks it is), check the number of items read is as expected, and should error check all the conversions. We convert to float via atof() which does not have the level of error handling we might hope for. The converter was also clearly not written with I/O performance in mind. A slicker solution would deduce the format and parse the input as fixed format fortran output, which is what it actually is.

src/Utilities/SimpleParser.cpp
prckent: Correct. Comments are off by one line. Fixed in new version.

ye-luo:
Does the current solution work when it meets 123-456-789?

prckent:
Hold merging. Was intended to but does not.

markdewing:
The new function is the kind of code that cries out for a unit test - the functionality is fairly simple and is isolated from the rest of the code, and it should be tested on a range of input strings.  See #671 for updates to the parser testing code.

prckent:
Agreed. Update won't happen for a while due to multiple time sinks.

grahamlopez:
test this please

prckent:
(I accidentally closed this by being too thorough in my cleaning of old branches in my forked copy)

PR #674, opened 2018-02-15
Save the particle sizes in the info.xml file

markdewing:
It's useful to know the problem size when analyzing the performance.


src/Particle/ParticleSet.cpp
```diff
@@ -313,6 +313,13 @@ bool ParticleSet::get(std::ostream& os) const
   return true;
 }
 
+void ParticleSet::output_particle_info(Libxml2Document &doc, xmlNodePtr root)
```
ye-luo: Can we avoid adding this routine in ParticleSet? I'd like to keep the class as light as possible.
The number of particles and name can be accessed via getName() and getTotalNum()

prckent:
Agreed. Ideally we would employ a design where these core classes know nothing about I/O. Input is of course the biggest headache.

prckent:
@ye-luo OK now?

ye-luo:
@prckent Yes. It is good now.
@markdewing github doesn't send notification when you force push. Please leave a note when you did it.

PR #676, opened 2018-02-20
Enable GAMESS HDF5 workflow in Nexus

jtkrogel:
Update interface to convert4qmc to include all current inputs.  Enable GAMESS->QMCPACK workflows that use the new HDF5 file for the orbitals.


markdewing:
okay to test

nexus/library/qmcpack_converters.py
```diff
@@ -686,16 +826,42 @@ class Convert4qmc(Simulation):
...
+    #end def get_prefix
+
+
+    def list_output_files(self):
+        # try to support both pre and post v3.3.0 convert4qmc
+        prefix = self.get_prefix()
+        wfn_file  = prefix+'.Gaussian-G2.xml'
+        ptcl_file = prefix+'.Gaussian-G2.ptcl.xml'
+        if not os.path.exists(os.path.join(self.locdir,ptcl_file)):
+            wfn_file  = prefix+'.wfnoj.xml'
```
markdewing: The new converter produces both .wfnoj.xml and .wfj.xml files by default.  If only one is selected by Nexus, I would think the .wfj.xml is more useful as part of a pipeline.

jtkrogel: Nexus generally writes its own Jastrow at user request.  The workflow dependency that is being processed by the code in question involves only the orbitals (i.e. qmc.depends(gms,'orbitals')) so the wavefunction without Jastrow is what is desired.

PR #678, opened 2018-02-22
Add QE 6.2.1 pw2qmcpack converter.

### LINKED ISSUE #266: Discussion: Converter for latest Quantum Espresso (v6.1)
Currently we have a patch for QE5.3 to produce the pw2qmcpack executable. This works well but newer QE versions are improved, particularly in the area of exact exchange and threading for memory usage. 

Updating the converter is sometimes not trivial due to changes in the QE fortran. We had been planning to use the QE HDF5 implementation so that we can write a converter "once and for all", but the format does not yet appear stable.

I suggest we update the converter for the old .save format used by QE6.1, so that we are up to date with QE. We will benefit from the QE workflow tests developed by @ye-luo .

Comments on this route?

ye-luo:
Yes. It finally arrives.
The patched reference code is hosted on [github](https://github.com/ye-luo/q-e/releases/tag/v6.2.1%2Bpw2qmcpack).
I know little about autotools. So I rely on QE to handle HDF5 search. This is different from the past.
close #266

### DIFF
--- a/external_codes/quantum_espresso/download_and_patch_qe6.2.1.sh
+++ b/external_codes/quantum_espresso/download_and_patch_qe6.2.1.sh
@@ -38,7 +38,8 @@ if [ -e $codename/PP/src/pw2qmcpack.f90 ]; then
 echo --- SUCCESS: ${codename} patched for pw2qmcpack converter
 echo There are two ways to build
 echo "1) if your system already has HDF5 installed with Fortran, use the --with-hdf5 configuration option."
-echo "   Current hdf5 support in QE is preliminary. If it is intended to avoid, replace '-D__HDF5' with '-D__HDF5_C' in make.inc."
+echo "   Currently HDF5 support in QE itself is preliminary. To enable use of pw2qmcpack"
+echo "   but use the old non-HDF5 I/O within QE, replace '-D__HDF5' with '-D__HDF5_C' in make.inc."
 echo "2) if your system has HDF5 with C only, manually edit make.inc by adding '-D__HDF5_C -DH5_USE_16_API'"
 echo "   in 'DFLAGS' and provide include and library path in 'IFLAGS' and 'HDF5_LIB'"
 else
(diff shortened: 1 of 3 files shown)


external_codes/quantum_espresso/download_and_patch_qe6.2.1.sh
```diff
@@ -0,0 +1,52 @@
...
+echo --- Please advise QMCPACK Developers via Google Groups if this problem persists
+else
+cd ${codename}
+patch -f -p1 -i ../add_pw2qmcpack_to_${codename}.diff
+cd ..
+if [ -e $codename/PP/src/pw2qmcpack.f90 ]; then
+echo --- SUCCESS: ${codename} patched for pw2qmcpack converter
+echo There are two ways to build
+echo "1) if your system already has HDF5 installed with Fortran, use the --with-hdf5 configuration option."
+echo "   Current hdf5 support in QE is preliminary. If it is intended to avoid, replace '-D__HDF5' with '-D__HDF5_C' in make.inc."
```
prckent: >Current hdf5 support in QE is preliminary. If it is intended to avoid, replace '-D__HDF5' with '-D__HDF5_C' in make.inc."

Did you mean:

Currently HDF5 support in QE itself is preliminary. To enable use of pw2qmcpack but use the old non-HDF5 I/O within QE, replace '-D__HDF5' with '-D__HDF5_C'

ye-luo: Correct. Let me update the words.

prckent:
The autotools produced configure script for QE6.2.1 appears to not have an HDF5 search ("grep -n -i hdf configure"). To use HDF5 we therefore have to modify make.inc. Agree? If so, we'll probably want to make some instructions/install examples.

ye-luo:
Incorrect. There is hdf5 detection from QE (`grep -n -i hdf install/configure`). `./configure` is a wrapper. However, if it can't find fortran module, you have to manually modify make.inc.

prckent:
Successfully built and tested (ctest -R qe) with QE6.2.1 and HDF5 1.10.1.

PR #682, opened 2018-03-02
Release candidate for 3.0.1

markdewing:
Fix the GPU 'nan' issue on parallel runs.
Also updated qmcpack.settings with compiler version and cuda version.


markdewing:
I ran the short tests on Titan for a CUDA build, and on my laptop with a CPU build.  All the tests passed.

This branch is not currently protected, so it could be merged w/o CI.   I suspect trying to backport the CI scripts would be more work that it's worth.

PR #683, opened 2018-03-07
Using VirtualParticleSet in non-local pseudo potential evaluating

ye-luo:
I would like to merge this PR to avoid further divergence.
The main changes in this PR.
1. VectorView class has been replaced by Vector which now supports using preallocated memory. This is conceptually simple but involving many files. Unfortunately, I realized that this change needed when I was half way developing 2.
2. Use VirtualParticleSet in NLPP evaluation. This is added on the regular code-path and the old NonLocalECPComponent.vp.cpp(not used) is removed. On the wavefunction side, necessary functions are implemented or adopted to use VP.
3. Hybrid representation is optimized when using VP in NLPP. Using this optimization will be less accurate when two spherical regions overlap. So a safety check needs to be added. 
4. When calculating NLPP, the quadrature grid is no more randomized per ion but per ion species as the GPU code path, there should be no change in the final results.
5. There is no need to save the grid for each electron in ParticleSet. So ParticleSet.Sphere is removed and randomize_grid() is simplified.

VP now works only with splines not LCAO and the implementation is still not matured yet. This is a necessary step for splitting a single walker in the future.


src/QMCHamiltonians/ECPotentialBuilder.cpp
```diff
@@ -50,10 +50,12 @@ bool ECPotentialBuilder::put(xmlNodePtr cur)
     nonLocalPot.resize(ng,0);
   }
   std::string ecpFormat("table");
+  std::string NLPP_algo("default");
   std::string pbc("yes");
   std::string forces("no");
   OhmmsAttributeSet pAttrib;
   pAttrib.add(ecpFormat,"format");
+  pAttrib.add(NLPP_algo,"algo");
```
markdewing: This should be expanded to 'algorithm'.  I don't see an advantage in using the abbreviated name in the input file.

markdewing:
Some higher-level questions about the code:
- What is the Virtual Particle algorithm?  What is different for the user if they specify it - faster?  more accurate?
- How does the VirtualParticle class differ from the regular Particle class (how does the implementation reflect the previous question)
- What changes are needed to the wavefunction classes to support the VirtualParticle algorithm?

prckent:
I would like changes -- in general, not only this area -- reduce code complexity and code volume while increasing accessibility. Additionally, can we get rid of old implementations at the same time?

ye-luo:
1. Algorithm. Currently when computing WF ratios for NLPP, all the quadrature points are computed one at a time. First point, propose the move, compute every WF component ratios, reject the move and then move on to the next point. This is basically serial per walker. In order to gain concurrency, the computation over all the points per WF component are put together. However, this requires extra memory and I handle it via VP.
2. When this algorithm is in effect, it doesn't directly speed up calculation but save a bit by removing the side effect of current algorithm. In the current algorithm, all the SPO will be recomputed when a T-move is accepted due to an unresolvable internal flag handling.
3. VirtualParticleSet is a derived class of ParticleSet to handle virtual moves (no accept). It contains virtual particle positions and distance tables to the real particles.
4. For each WF component, evaluateRatios(VP,...) is needed. For each SPO class, evaluateValues is added.
5. @prckent This PR in general reduces the lines of source code. The added lines are a bit for hybrid optimization and a lot in tests. All the existing VP evaluation in AoS are recycled and refined. We gain very very little by removing the old implementation. The unused routines in the old implementation have been removed in this PR.

markdewing:
Is the choice of algorithm going to stay, or is it temporary until all the code is changed to use the virtual particle algorithm?  I'm wondering because it affects how hard to work on the name and the input parameter.

I don't like the 'virtual particle' name for this algorithm - it refers to an implementation detail.  And it's not even a key detail - it refers to a change in a support class and the change is only necessary because the original code precisely targets the Monte Carlo "evaluate trial move, then accept/reject" strategy.   Would something like 'batch ratios' or 'ratio vector' be a more descriptive name?

ye-luo:
I think it is not a problem to keep both algorithms since the old and new diverge very little. Removing the old one saves very little. All the single move ratio methods are needed by VMC without drift.
In principle, once I add a fallback implementation in the WF base class, we can switch the new algorithm on by default. But I prefer to do the change when the batching strategy is matured.

Better name is always welcome. I will change the input tag to algorithm="batched".

PR #684, opened 2018-03-07
Ppconvert cmake

### LINKED ISSUE #679: ppconvert segfaults in release with modern clang and mkl
Seg faults in release on OSX with llvm 5.0.1. on lab 2's 
```shell
ppconvert --gamess_pot O.BFD.gamess --s_ref "1s(2)2p(4)" --p_ref "1s(2)2p(4)" --d_ref "1s(2)2p(4)" --xml O.BFD.xml
```
Also trips so so many warnings even with the warning suppression the rest of the codebase requires on.

PDoakORNL:
Integration of ppconvert with the main build, seems to fix #679 as well.


CTestConfig.cmake
```diff
@@ -0,0 +1,14 @@
+## This file should be placed in the root directory of your project.
```
markdewing: This file should not be checked-in.   Are you building from the qmcpack/ directory?  You should be building in the build/ directory to keep the directory tree cleaner.

manual/installation.tex
```diff
@@ -561,6 +561,46 @@ \subsection{Installing on Mac OS X using Homebrew (brew)}
 \end{verbatim}
```
markdewing: Could you put this manual section (and associated changes) in a separate pull request? It's not related to getting ppconvert into the build system

src/QMCTools/CMakeLists.txt
```diff
@@ -71,3 +71,6 @@ FOREACH(p ${QTOOLS})
 
 ENDFOREACH(p ${QTOOLS})
 
+#ppconvert
+SUBDIRS(ppconvert)
```
markdewing: Could you put this under a flag, such as BUILD_PPCONVERT.  I'm not sure we want to start building ppconvert on every platform by default.

CMake/FindMKL.cmake
```diff
@@ -1,43 +1,90 @@
 # Simple file to find MKL (if availible)
 # This needs a lot of work to make it robust
 INCLUDE( CheckCXXSourceCompiles )
+CMAKE_MINIMUM_REQUIRED(VERSION 3.2.0)
```
markdewing: What features of cmake need version 3.2.0?   We've been supporting back as far as 2.8.10, at the cost of some cmake version if-tests.

CTestConfig.cmake
PDoakORNL: This is used by the ci tool I'm using locally to store settings about the project, I'll see about moving it somewhere else.  I'll remove it from here.

src/QMCTools/CMakeLists.txt
PDoakORNL: See ppconvert/src/CMakeLists.txt  EXCLUDE_FROM_ALL means it is not built by default on any platform.

prckent:
Thanks.  ppconvert is actually required to run the tutorials. Adding to the build would be an improvement.

Can we assess how risky building ppconvert is? If we think it is likely to be reliable then we should build it by default and perhaps add an option to disable the build. For a reliable build I think we need to be able to compile a serial C++ application and successfully link with BLAS et al. This does not seem risky. The main historical problem I am aware of was that it used its own makefile and not CMake.

ye-luo:
Is your development based on MacOS? Did you check linux?

PDoakORNL:
I've build it with this on Centos7 with intel 17.0.0

CMake/ClangCompilers.cmake
```diff
@@ -12,27 +12,46 @@ SET(ENABLE_OPENMP 1)
...
+IF (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.0 )
+  SET( HAVE_POSIX_MEMALIGN 1 )    # Clang doesn't support -malign-double it does in 5.0.1
+ELSE ()
+  SET( HAVE_POSIX_MEMALIGN 0 )
+ENDIF()
 
 # Suppress compile warnings
 SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -Wno-deprecated -Wno-unused-value")
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated -Wno-unused-value")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated -Wno-unused-value -Wno-undefined-var-template")
```
ye-luo: Need a version check.
https://github.com/QMCPACK/miniqmc/blob/105f43c4ab1a1a2f7005b6314cbd85a3fcc2f195/CMake/ClangCompilers.cmake#L27

CMake/ClangCompilers.cmake
```diff
@@ -12,27 +12,46 @@ SET(ENABLE_OPENMP 1)
...
+    SET(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fopenmp")
+    SET(CMAKE_C_FLAGS_RELWITHDEBINFO     "${CMAKE_C_FLAGS_RELWITHDEBINFO} -fopenmp")
+    SET(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -fopenmp")
 ENDIF()
 
 # Set clang specfic flags (which we always want)
 ADD_DEFINITIONS( -Drestrict=__restrict__ )
 
-SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -fomit-frame-pointer -fstrict-aliasing")
+SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -fomit-frame-pointer -fstrict-aliasing -D__forceinline=inline")
```
ye-luo: I don't think we have __forceinline in c code.

CMake/FindMKL.cmake
```diff
@@ -1,43 +1,90 @@
...
+IF(COMMAND cmake_policy)
+  cmake_policy(SET CMP0056 NEW)
+ENDIF(COMMAND cmake_policy)
 
+# if MKL_ROOT is set, use that
+if ( MKL_ROOT )
```
ye-luo: I think it should be $ENV{MKL_ROOT}

CMakeLists.txt
```diff
@@ -477,13 +476,27 @@ IF (ENABLE_GCOV)
...
 IF(CXX11_NEEDED)
   SET(CXX11_FLAG "-std=c++11")
   #check if the CXX compiler supports -std=c++11 option
-  include(CheckCXXCompilerFlag)
-  CHECK_CXX_COMPILER_FLAG(${CXX11_FLAG} CXX_COMPILER_SUPPORT_CXX11)
+  # Old method
+  # include(CheckCXXCompilerFlag) found this problematics with clang 5.0.1
+  # CHECK_CXX_COMPILER_FLAG(${CXX11_FLAG} CXX_COMPILER_SUPPORT_CXX11)
+  
+  FILE( WRITE "${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp/std11.cxx"
```
ye-luo: Since CXX11 is now required. We probably need to let CMake handle standard than checking by us. In the past, I found our checking is not robust.

ye-luo:
I'm not confortable with the changes in FIND_MKL. More checking is needed before merge.

prckent:
Does this build on Cray?

PDoakORNL:
Ok I think that should address the concerns here. I do not think it is necessary to pick apart the changes to clang compilers and find mkl into separate PR's. Branching is a tool not a religion. I appreciate the desire to keep the docs separate so I removed them.

If someone else following develop is really building on clang or has gotten mkl to work on a compiler other than the intel compiler, they are most qualified to fix anything this has broken in their build. And since they haven't weighed in on this PR the easiest way to get their attention is to break their build.

PDoakORNL:
Those __forceinline are necessary for llvm 5.0 build. If we need to support ancient clang (why?) it should be added to CI.

CMake/ClangCompilers.cmake
```diff
@@ -29,10 +31,14 @@ IF ( CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 3.8 )
 ENDIF()
 
 # Set extra optimization specific flags
-SET( CMAKE_C_FLAGS_RELEASE     "${CMAKE_C_FLAGS_RELEASE} -ffast-math" )
-SET( CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -ffast-math" )
-SET( CMAKE_C_FLAGS_RELWITHDEBINFO     "${CMAKE_C_FLAGS_RELWITHDEBINFO} -ffast-math" )
-SET( CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -ffast-math" )
+SET( CMAKE_C_FLAGS_RELEASE     "${CMAKE_C_FLAGS_RELEASE} -ffast-math -D__forceinline=inline" )
```
ye-luo: Why don't you put forceline in CMAKE_CXX_FALGS at line 28

PDoakORNL:
Don't want it in debug, but I'm about to push the answer to that. You pushed me to a cleaner solution.

CMakeLists.txt
```diff
@@ -338,9 +369,10 @@ ELSE(CMAKE_TOOLCHAIN_FILE)
   ENDIF($ENV{CRAYPE_VERSION} MATCHES ".")
 
   #------------------------------------
-  # Identify the compiler
+  # Identify the compiler -- I don't get why this COMPILER variable is added --PD
```
ye-luo: It is used to pick compiler vendor specific cmake configuration later. Perhaps, it can be simplified.

ye-luo:
The MKL detection failed on my workstation ubuntu 14.04 + intel 18.
I really want the MKL change out of this PR such that we can finish review/validate the ppconvert integration.

PDoakORNL:
I needed the MKL detection to work to do the ppconvert integration.

ye-luo:
If you work with gnu/intel compiler on a linux box, why you need to touch MKL detection to integrate?

PDoakORNL:
That's not where I did the initial integration.

ye-luo:
You can just overwrite you find_mkl with the old one + put back the intel guard. Then PR the changes later.
To be honest, your bunch of guards for Clang 5.0 is not solid to me. I feel it is due to macOS not really clang.
MacOS is the least platform we care about. It is good to support it well but the baseline is always linux.

PDoakORNL:
It is stock clang, that's the point llvm 5.0.1 != apple clang 9.0.
That's why the CXX_EXE_LINKER flags need to work.

I just pushed the fix for linux. Once this is merged I can run my own CI on linux, with this PR in I know it works. I can't find the link to the main repo CI anywhere. And I don't see any checks coming in here.

ye-luo:
find_mkl still failed on my workstation. In the past, it took us a long time to cover many corner cases.
Even stock Clang can be different on different platforms. For example -march doesn't work on power.

PDoakORNL:
Still works on mine
What does you cmake invocation look like?

CMake/ClangCompilers.cmake
```diff
@@ -17,9 +17,14 @@ ENDIF()
 # Set clang specfic flags (which we always want)
 ADD_DEFINITIONS( -Drestrict=__restrict__ )
 
-SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -fomit-frame-pointer -fstrict-aliasing")
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fomit-frame-pointer -fstrict-aliasing -D__forceinline=inline")
-SET( HAVE_POSIX_MEMALIGN 0 )    # Clang doesn't support -malign-double
+IF (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.0 )
+  SET( HAVE_POSIX_MEMALIGN 1 )
+  SET( CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -D__forceinline=inline" )
```
ye-luo: forceinline is needed in *any* case for clang. Please move it to CMAKE_CXX_FLAGS.

PDoakORNL: My debug works fine without it.

markdewing:
General comment:  Changes to the build system are going to receive a lot more scrutiny than other changes because the build is most users first experience with the code.   Also the build system has to work on many more cases, which have to be tested.   For the code changes we can target standard C++11.  For compilers, libraries, and environments, there is no standard.   Which makes understanding the impact of build system changes harder.

CMake/ClangCompilers.cmake
ye-luo: Even without `-D__forceinline=''`it works?

ye-luo: I prefer a clean solution as I explained above. It should also not inside version check. It is needed on PowerPC Clang 3.8.

markdewing:
For building using gcc and clang with MKL, I use this on the CMake line:
-DBLA_VENDOR=Intel10_64lp_seq -DCMAKE_PREFIX_PATH=$MKLROOT/lib

This uses the CMake BLAS/LAPACK detection.

CMake/ClangCompilers.cmake
PDoakORNL: No it doesn't __forceinline is a compiler specific tag that shouldn't be in a portable codebase. However this sort of intense inlining is also something I wouldn't expect enabled in a debug build.

PDoakORNL:
@markdewing  So you use MKL but avoid the FindMKL all together. I think you should modify FindMKL so it also covers your case. I wouln't have anticipated this as a build strategy.

markdewing:
I would highly recommend creating a new branch and new PR with just the ppconvert changes.

* pp_convert changes: it doesn't build with CMake previously so anything is an improvement.
* other build system changes: already breaks a number of known configurations.  Will require lots of testing on many different platforms to ensure it works.

The pp_convert changes can be even more isolated from the existing code if an 'if (BUILD_PPCONVERT)' test is added before the SUBDIRS(ppconvert) line, and BUILD_PPCONVERT defaults to false.  This would also isolate the minimum cmake version change.

I see the cmake version needs to be raised to support 'target_sources' (and some others).  Making the cmake files more target-oriented is a more modern style, and probably where we would like to move in the future. However, I'm not ready to raise the minimum cmake version for the whole project yet.

I would be okay with requiring a higher minimum cmake for building ppconvert (but maybe others have a different view?)

If you need the other build system change for your environment, that's fine to have them there (but not in the commit).  ppconvert *should* build with the existing build system from any particular environment, and if that's not true we can fix in the future.

CMake/ClangCompilers.cmake
ye-luo: You are right. It is compiler specific. There are two ways to protect it, 1 put macro in source, 2 add option in the compiler configuration. For the moment, 2 is simpler.

prckent:
Note that fully optimizing the performance of ppconvert is not necessary - any decent compile will be fast enough. No need to have the fanciest optimization settings, particularly if we have to special case them for compiler version etc.

markdewing:
Comment on 'inline' and '__forceinline'.

The 'inline' keyword in C++ is mostly used to specify a linkage convention to avoid multiple definition errors when the same code is used in multiple libraries (templated code, usually).   I'm not sure it has much value as a suggestion to the compiler for inlining anymore.

On a simple test, gcc, clang, and icpc did not inline a function at -O0, whether or not the 'inline' keyword was present.  At -O2, the function was inlined, regardless of the 'inline' keyword.
If the __forceinline keyword was present (for icpc), or __attribute__((always_inline)) (for gcc and clang), the function was inlined even at -O0.

There doesn't seem to be a need for a different definition for __forceinline for release vs. debug builds, at least when setting it to 'inline' or blank.

(I do agree that the code should use a more generic macro name rather than the compiler-specific __forceinline.  But __forceinline is already there, and so we might as well use that as the macro name for now.)

grahamlopez:
Apologies for not noticing the problems with the CI earlier - working on it.

grahamlopez:
test this please

PDoakORNL:
Closing this, I will offer as at least two separate PR's when I have time.

ye-luo:
Could you please move only_ppconvert branch from qmcpack repo to your fork?

PR #686, opened 2018-03-14
Add ppconvert to CMake build system

PDoakORNL:
Well this builds with Centos intel 17.0.0 split from the FindMKL changes that's what i can reasonably look at. It is not part of the default build.  You must both -DBUILD_PPCONVERT=ON and make ppconvert. 

My approach to blitz is to spend as little effort working around it as possible. If ppconvert survives long term I would advocate completely removing blitz.

* [x]  Update README to say where ppconvert is known to build. If we know that one class of systems is risky/broken, OK to just say so at this point. ppconvert is going to run on a local workstation in all likelihood.
* [x] Update README to remove "great chance to learn" comment. Put a friendlier "ask a developer or post on google groups" message as we do elsewhere.
* [ ] Verify "default" builds are not affected on main production systems, e.g. Cray. i.e. This patch is benign unless activated. (@markdewing perhaps? Should not be an ORNL person by our reviewing rules)
* [x] Fix release crash
* [x] Allay any serious cmakelist.txt concerns, if any.

I am hopeful that this can be merged soon.


grahamlopez:
okay to test

src/QMCTools/ppconvert/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+#####################################################
+CMAKE_MINIMUM_REQUIRED(VERSION 3.2.0)
```
markdewing: Seems to work with CMake 3.1.0 (at least on Linux with clang compiler)

src/QMCTools/ppconvert/src/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+SUBDIRS(common)
+
+add_executable(ppconvert EXCLUDE_FROM_ALL "")
+
+target_sources(ppconvert
+  PRIVATE
+    "${CMAKE_CURRENT_SOURCE_DIR}/CubicSpline.cc"
```
markdewing: I don't think the ${CMAKE_CURRENT_SOURCE_DIR} is necessary.  Just the source file name w/o quotes should work.

src/QMCTools/ppconvert/src/common/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+add_library(common EXCLUDE_FROM_ALL "")
+
+target_sources(common
+  PRIVATE
+  "${CMAKE_CURRENT_SOURCE_DIR}/GKIntegration.cc"
```
markdewing: Similar to src/CMakeLists.txt, the ${CMAKE_CURRENT_SOURCE_DIR} should not be needed.

src/QMCTools/ppconvert/src/common/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+  "${CMAKE_CURRENT_SOURCE_DIR}/Functionals.cc"
+  )
+
+# get_property( COMMON_COMPILE_DEFS TARGET common PROPERTY COMPILE_DEFINITIONS )
+# set(COMMON_COMPILE_DEFS ${COMMON_COMPILE_DEFS} $<$<NOT:$<CXX_COMPILER_ID:Clang>>:__GNUC__=4>)
+# MESSAGE( "$<CXX_COMPILER_ID>" )
+
+  target_compile_definitions(common
+    PRIVATE
+    $<$<CXX_COMPILER_ID:Intel>:__ICC=1>
```
markdewing: ICC should already define this variable (at least it appears to on Linux).

src/QMCTools/ppconvert/src/common/CMakeLists.txt
PDoakORNL: Specifying this is just part of the idiom as I learned it, perhaps because if the sources are moved to another directory and brought in via an include() it is changed to ${CMAKE_CURRENT_LIST_DIR} and this emphasizes these sources are not in a separate location.

PDoakORNL: As well as making regular expression update trivial.

markdewing: Okay. Looking this up I see the add_library call could be in a different directory from the target_sources call, and the directories searched are relative to the location of the add_library call - adding the CMAKE_CURRENT_SOURCE_DIR allows them to still be found.

prckent:
Since this is a major upgrade over the previous situation with ppconvert I think the only requirements for a merge should be:

* [ ]  Update README to say where ppconvert is known to build. If we know that one class of systems is risky/broken, OK to just say so at this point. ppconvert is going to run on a local workstation in all likelihood.
* [ ] Update README to remove "great chance to learn" comment. Put a friendlier "ask a developer or post on google groups" message as we do elsewhere.
* [ ] Verify "default" builds are not affected on main production systems, e.g. Cray. i.e. This patch is benign unless activated. (@markdewing perhaps? Should not be an ORNL person by our reviewing rules)
* [ ] Allay any serious cmakelist.txt concerns, if any.

I am hopeful that this can be merged soon.

@PDoakORNL  I suggest copying the checklist to your original PR message and updating as items are checked off.

prckent:
As regards the future of ppconvert, first note that the tutorials currently describe its use and we do need a way of generating orbitals/projectors for potentials. ppconvert is needed for users but it is certainly not a friendly or easy to use code. Once the future of potential generation is determined, e.g. in the context of https://pseudopotentiallibrary.org/ , we can decide what to do with ppconvert. Perhaps we upgrade it, or perhaps we use some python to drive a single or multiple projector capable atomic code.

ye-luo:
@PDoakORNL could you answer my question? Since the source code is changed. I’m going to check a simple case.

PDoakORNL:
Where ppconvert using isnan imported from blitz that was ambiguous with std::isnan I changed source to use std::isnan (which on most platforms was what was being used anyway). This hasn't caused issues in gcc, intel 17.0.0 or llvm 5.0.1

prckent:
@ye-luo I see:

* Delete unused files not part of ppconvert
* Definitions and namespace fixes for compilation (std::, Mac problems etc.)
* Cmakelist updates

That is it.

PDoakORNL:
Spoke too soon on gcc, have not tested that without the FindMKL changes

markdewing:
test this please

src/QMCTools/ppconvert/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
...
+IF(COMMAND cmake_policy)
```
ye-luo: Please remove these three lines of policy.

ye-luo:
Unfortunately, ppconvert failed to convert pp.
From develop,
```
Atomic symbol = Si
PseudoCharge = 4
lmax = 2
Finished reading GAMESS pseudopotential.
```
From this branch,
```
Atomic symbol = 
PseudoCharge = 0
lmax = 0
Finished reading GAMESS pseudopotential.
```
Something goes WRONG.

PDoakORNL:
What platform?  I can only attest to Centos7 intel 17.0.0 using the MKL detection as is i.e. the -mkl flag. It does not crash for me.

PDoakORNL:
I no longer have a release mode crash building with my full modifications i.e. the previously closed PR on Centos icc17 or OSX llvm 5.0.1. I am using the example from the lab2
```shell
ppconvert --gamess_pot O.BFD.gamess --s_ref "1s(2)2p(4)"  --p_ref "1s(2)2p(4)" --d_ref "1s(2)2p(4)" --xml O.BFD.xml
```

PDoakORNL:
If you want me too look at your Si example make a gist or put input in slack. Or 
gdb ppconvert ...
and bt and tell us where it blew up.

ye-luo:
Ubuntu 14. Intel 18u1 icpc mkl.
```
(gdb) bt
#0  0x000000000046fc22 in CubicSplineCommon::Update() ()
#1  0x000000000045a807 in DFTAtom::SetGrid(Grid*) ()
#2  0x000000000042ca98 in PseudoClass::CalcProjector (this=0x7fffffffc678, refstate=<error reading variable: Cannot access memory at address 0x0>, lchannel=0)
    at /sandbox/opt/cleanup/PDoak/src/QMCTools/ppconvert/src/NLPPClass.cc:2119
#3  0x0000000000429f09 in main (argc=-14728, argv=0x0) at /sandbox/opt/cleanup/PDoak/src/QMCTools/ppconvert/src/NLPPClass.cc:1941
```
I don't think the problem is here but more from parsing the text of gamess format.

prckent:
@ye-luo Is anything strange about the build options used? Perhaps try with less optimization? Peter's goal is not to fix ppconvert here, which would be a different and potentially much larger project. These can be tackled in later PRs. I think our experience here shows that (1) not many people are using ppconvert (2) unfortunately it is somewhat flaky and holey code. If it can't be stablized without much work the decision for the fate of this utility is easy.

ye-luo:
I found the problem. `-DNDEBUG` inherited from main QMCPACK will make all the assert() empty. However, ppconvert have a lot of calls inside assertion like `assert (parser.OpenFile (fileName))` basically, no file operation open/read/close is called.

prckent:
Ouch. By eye there are about 100 asserts that do something important (parser.read()) vs check a pure assertion (a==b).

PDoakORNL:
Yes I'd clobbered my CXX_FLAGS leading me to think ppconvert release was well when it wasn't.  Explains the disappearing release crash.  Thanks for the catch @ye-luo

PDoakORNL:
So with removal of the -DNDEBUG flag for Release builds of ppconvert it produces results matching debug for Intel 17&18 with no command line CMAKE_CXX_FLAG tricks.

src/QMCTools/ppconvert/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# this new link path policy has fixed ldpath problems
+# experienced with out of dyld_load_path libraries on
+# OSX without bad effect on other platforms.
+cmake_policy(SET CMP0003 NEW)
```
ye-luo: Do we really need cmake_policy? What happends if this is removed completely.

src/QMCTools/ppconvert/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
...
+
+# in this directory and below remove the -DNDEBUG flag
+string(REPLACE "-DNDEBUG" "" CMAKE_CXX_FLAGS_RELEASE ${CMAKE_CXX_FLAGS_RELEASE})
```
ye-luo: If this CMAKE_CXX_FLAGS change only affects locally, why not apply it on CMAKE_CXX_FLAGS instead of just release. Someone like me, maybe build release with debug case.

src/QMCTools/ppconvert/README
```diff
@@ -12,6 +12,12 @@
 
 ppconvert supports conversion of pseudopotentials between different formats
 
-ppconvert is currently not built automatically, but only requires a C++ compiler and BLAS/LAPACK
+ppconvert is currently not built automatically, but can be built on limited platforms
+(Centos7 intel 17.0.0 as of now) it may work on other platforms. Clang and GCC support
+with MKL are planned.
 
-Edit the Makefile to include an appropriate compiler and BLAS source. Examples using MKL are included.
+cmake -DBUILD_PPCONVERT
```
ye-luo: cmake -DBUILD_PPCONVERT=1
If you don't assign a value, cmake will complain. Because QMCPACK doesn't use on/off, use 1.

src/QMCTools/ppconvert/CMakeLists.txt
PDoakORNL: Good point

ye-luo:
I checked the generated UPF and xml in QE and QMCPACK. They are all good.
A converter test is desired to automate the checking but I will ask for it later, not now.

ye-luo:
Since the cmake part is not turned on by default. There is no issue with regular build.
We don't run ppconvert on HPCs since they usually use cross-compiling and binaries usually can't run on login nodes. So there is no such need for Cray or BGQ. On the other hand, I prefer to have a converter test first. Once we have it, we can go back to check Cray and BGQ and the checking will be also simpler.

Once all my CMake questions are addressed, I will approve and merge this PR.

PDoakORNL:
After all this I am writing a test.  At this point just to catch crashes. The issue with comparing the output is between different build configurations you get lots of 1E-13 sort of diffs in the output.  It's consistent within the same build.  So a simple diff test is out.

PDoakORNL:
Ok stand alone test added.  Since this is an isolated subproject I thought this was the way to do this for now. Removed cmp0003 really belongs in the llvm 5.0.1 changes. Fixed readme.

src/QMCTools/ppconvert/CMakeLists.txt
PDoakORNL: It's gone for this PR

src/QMCTools/ppconvert/CMakeLists.txt
ye-luo: I think this is the last thing.

PDoakORNL: Except it doesn't work as CMAKE_CXX_FLAGS doesn't contain -DNDEBUG only release has that by default.  In build config release you get CMAKE_CXX_FLAGS + CMAKE_CXX_FLAGS_RELEASE. You don't get CMAKE_CXX_FLAGS += CMAKE_CXX_FLAGS_RELEASE.  So it can't be removed from there. I think just protecting against the default config flags is sufficient. Someone who is messing with flags by hand and has ppconvert break is likely to take a look in here.

ye-luo: I see.
Could you also do the same removal for CMAKE_CXX_FLAGS_RELWITHDEBINFO which also has -DNDEBUG.
Thank you.

PR #687, opened 2018-03-14
Handle the negative count coverage case.

markdewing:
Occasionally GCC produces negative coverage counts (should be impossible!), and this
causes problems wih the coverage scripts.  The processing stops at the point the
file with the negative count is found.  Which may be why files seem to be missing from
the coverage reports.

Fix is to detect negative counts when reading the file and mark them as uncovered (all the
cases I found with negative counts should have been uncovered).


prckent:
Please add a note about when/how you saw this.

Issue #695, opened 2018-03-19
sed warning at make


ye-luo:
I got the following warning on Titan.
Coming from `src/CMakeLists.txt`
```
sed: invalid option -- 'E'
Usage: sed [OPTION]... {script-only-if-no-other-script} [input-file]...
...
```

PDoakORNL:
https://stackoverflow.com/questions/3139126/whats-the-difference-between-sed-e-and-sed-e

Its apparently been support by gnu sed for some time but not always.
the newest gnu-sed source code supports -E and implies it will become part of the POSIX standard.
For now i think we should just branch for Darwin since -r will work everywhere that isn't BSD.

prckent:
@grahamlopez more reason to get CI working on Titan et al.

markdewing:
The consequence will be there is no git commit subject in the git-rev.h file on Titan.

PR #696, opened 2018-03-19
fix for BSD's sed

PDoakORNL:
While fixing #695 This leaves OSX able to get git rev and I assume other BSD's as well. It works with  cmake 2.8.10.2 on titan.

### DIFF
--- a/manual/installation.tex
+++ b/manual/installation.tex
@@ -614,7 +614,7 @@ module load cudatoolkit           # CUDA for GPU build
 module load cray-hdf5
 module load cmake
 module load fftw
-export FFTW_HOME=$(echo $FFTW_DIR | sed -e 's/\/lib//')
+export FFTW_HOME=$FFTW_DIR/..
 module load boost
 mkdir build_titan_gpu
 cd build_titan_gpu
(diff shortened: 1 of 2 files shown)


prckent:
OK to test

prckent:
I think it would be more robust to check for -E and/or -r support and then use the appropriate flags. A hardwired list is less future proof and for sure there are non-GNU variants not there.

PDoakORNL:
Ok this should last until sed -r only machines all die. Although I'm sure there is a way to implement that custom command without extended regexp in sed if someone wants to go for it.

manual/installation.tex
```diff
@@ -614,7 +614,7 @@ \subsection{Installing on ORNL OLCF Titan Cray XK7 (NVIDIA GPU
 module load cray-hdf5
 module load cmake
 module load fftw
-export FFTW_HOME=$FFTW_DIR/..
+export FFTW_HOME=$(echo $FFTW_DIR | sed -e 's/\/lib//')
```
ye-luo: Please explain this change. Is this causing problems? I used the build script yesterday on titan and it was fine.

Issue #429, opened 2017-10-23
Compilation error with iszero with glibc (>2.24)


fdmalone:
I ran into the following when trying to compile qmcpack on my laptop (linux, gcc v7.2.0):

```
[ 15%] Building CXX object src/CMakeFiles/qmcbase.dir/Lattice/Uniform3DGridLayout.cpp.o
[ 15%] Building CXX object src/CMakeFiles/qmcbase.dir/Particle/ParticleSet.cpp.o
[ 15%] Building CXX object src/CMakeFiles/qmcbase.dir/Particle/ParticleSet.BC.cpp.o
[ 16%] Building CXX object src/CMakeFiles/qmcbase.dir/Particle/MCWalkerConfiguration.cpp.o
In file included from /home/fionn/projects/qmcpack_public/src/Estimators/accumulators.h:27:0,
                 from /home/fionn/projects/qmcpack_public/src/Estimators/ScalarEstimatorBase.h:24,
                 from /home/fionn/projects/qmcpack_public/src/Estimators/EstimatorManagerBase.h:30,
                 from /home/fionn/projects/qmcpack_public/src/QMCDrivers/QMCDriver.h:32,
                 from /home/fionn/projects/qmcpack_public/src/Particle/MCWalkerConfiguration.cpp:33:
/home/fionn/projects/qmcpack_public/src/config/stdlib/limits.h:17:13: error: redefinition of ‘template<class T> bool iszero(T)’
 inline bool iszero(T a)
             ^~~~~~
In file included from /usr/include/c++/7.2.0/cmath:45:0,
                 from /usr/include/c++/7.2.0/complex:44,
                 from /home/fionn/projects/qmcpack_public/src/Configuration.h:24,
                 from /home/fionn/projects/qmcpack_public/src/Particle/ParticleSet.h:25,
                 from /home/fionn/projects/qmcpack_public/src/Particle/MCWalkerConfiguration.h:28,
                 from /home/fionn/projects/qmcpack_public/src/Particle/MCWalkerConfiguration.cpp:24:
/usr/include/math.h:551:1: note: ‘template<class __T> bool iszero(__T)’ previously declared here
 iszero (__T __val)
 ^~~~~~
make[2]: *** [src/CMakeFiles/qmcbase.dir/build.make:135: src/CMakeFiles/qmcbase.dir/Particle/MCWalkerConfiguration.cpp.o] Error 1
[...]

PR #697, opened 2018-03-20
Llvm 5 mkl

### LINKED ISSUE #685: cmake halts on iszero test that should just fail for intel and clang on linux and clang on osx and when bad cmake policy is set on cray.
It intermittently fails the iszero test for clang and intel setup in the obvious way on linux and (llvm not apple) clang on osx.

PDoakORNL:
This makes building with llvm 5 possible.  Also fixes #685 
Documentation of this for os x to follow.
Build for linux llvm 5 I'm still working on the support libraries.

Build is protected from new FindMKL logic by
-DENABLE_OUTSIDE_MKL
which is false by default.

Since cmp0056 is needed for using any non system compiler and using rpath instead of LD_LIBRARY_PATH I have made that generally invoked except on cray, where while it should have had no effect with 2.8 cmake's, it still broke the build.

I've tested this PR with intel 16 on Centos 7, llvm 5 on OS X, gcc 4.9.3 on titan.


markdewing:
okay to test

CMake/ClangCompilers.cmake
```diff
@@ -19,7 +19,11 @@ ADD_DEFINITIONS( -Drestrict=__restrict__ )
 
 SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -fomit-frame-pointer -fstrict-aliasing")
 SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fomit-frame-pointer -fstrict-aliasing -D__forceinline=inline")
-SET( HAVE_POSIX_MEMALIGN 0 )    # Clang doesn't support -malign-double
+IF (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.0 )
```
markdewing: Shouldn't posix_memalign depend on the libc being using, not the compiler version?

That said, the old line doesn't match the comment.

PDoakORNL:
test this please

CMake/ClangCompilers.cmake
PDoakORNL: I have to say I took the scripts word for it.  It seems that in some cases clang needed changes even when posix_memalign existed on the system. However its generally something provided or not at the system level.

There is lots of discussion of posix_memalign on different plarforms in the clang 3.x time period so I image this is real and we should still set it to 0 for old enough clang.  I've fixed this so if we're on a system where we earlier failed to detect posix_memalign we don't set it to 1 here.

CMakeLists.txt
```diff
@@ -5,19 +5,30 @@ PROJECT(qmcpack)
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8.10)
 IF(COMMAND cmake_policy)
   cmake_policy(SET CMP0003 NEW)
+  IF(NOT($ENV{CRAYPE_VERSION} MATCHES "."))
```
markdewing: This test is backwards - the 'NOT' should be removed.

grahamlopez:
test this please

CMake/ClangCompilers.cmake
ye-luo: If the alignement is not guaranteed. The calculation may go wrong.
I also feel this is not a compiler thing but glibc.
@PDoakORNL which platform you use requires this change?

CMake/FindMKL.cmake
```diff
@@ -2,42 +2,105 @@
 # This needs a lot of work to make it robust
 INCLUDE( CheckCXXSourceCompiles )
 
+if ( NOT CMAKE_CXX_COMPILER_ID MATCHES "Intel" )
+  # if not intel and MKL_ROOT not set
+  if (NOT MKL_ROOT)
+    find_path(MKL_ROOT "mkl.h"
+      HINTS "${MKLROOT} ${MKL_HOME} $ENV{MKLROOT} $ENV{MKL_ROOT} $ENV{MKL_HOME}")
```
markdewing: I think HINTS expects multiple separated arguments following it, not a single argument (no quotes around all of the variables).
Also, it only seemed to work when the /include was explicitly added (${MKLROOT}/include)

CMake/FindMKL.cmake
```diff
@@ -2,42 +2,105 @@
...
+    string(REPLACE "/include" "" MKL_ROOT ${MKL_ROOT})
+  endif (NOT MKL_ROOT)
+  if (NOT MKL_ROOT)
+    message (FATAL_ERROR "ENABLE_OUTSIDE_MKL is TRUE and mkl not found. Set MKL_ROOT." )
+  endif (NOT MKL_ROOT)
+
+  #Do we support ilp?
```
markdewing: No, we don't use ILP64.  I don't think there are any really long vectors that need 64 integers for indices.

CMakeLists.txt
PDoakORNL: Message was backwards.  The policy needs to not be attempted on CRAY, at least with cmake 2.8.10 it causes a failure.

CMakeLists.txt
```diff
@@ -5,19 +5,30 @@ PROJECT(qmcpack)
...
+    MESSAGE(STATUS "Running on a Cray machine.")
+    cmake_policy(SET CMP0056 NEW)
+    # This policy insures that the CMAKE_EXE_LINKER_FLAGS of the calling project
+    # are used in the try_compile test cmake project.  Without this
+    # clean rpath builds with mkl aren't possible. While it is unknown by
+    # cmake 2.8.10.2 on titan it still breaks the iszero test there --PWD
```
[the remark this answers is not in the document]
PDoakORNL: I think the warning is something you want to see.  If you have a cmake that old the iszero and c++11 test can fail if you are trying to use a compiler without polluting the Library Paths.  Unfortunately somehow on cray with cmake 2.8.10 you get the warning and a failure of the iszero test.

grahamlopez:
test this please

CMake/FindMKL.cmake
PDoakORNL: Right on the expansion. And the PATH_SUFFIXES dropped out.  Which is why that replace is there.
I don't need this but this should cover some of the mkl modules I've seen, thought it would be nice to start moving in that direction

CMake/FindMKL.cmake
PDoakORNL: Ok well then I think these libraries should be fairly portable.  They work on Centos and Darwin at least.

PDoakORNL:
test this please

CMakeLists.txt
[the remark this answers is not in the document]
markdewing: The test should use the CMake version, and not whether it's a Cray system.

It's okay to have degraded functionality if building with older versions of CMake.   (In this case, the ability to use rpath consistently. )   Seeing the CMake warning about CMP0056 is not very helpful - it would be better to print a more user-friendly warning message.

The iszero test is a compile-only test - runtime library paths should not enter into it.

grahamlopez:
test this please

PDoakORNL:
test this please

CMakeLists.txt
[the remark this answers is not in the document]
PDoakORNL: it does now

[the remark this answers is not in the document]
PDoakORNL: If the linker flags are not set correctly this fails because -fopenmp and other flags are added but they can't be linked.  So you don't just get a test failure you get a cmake fail.

grahamlopez:
test this please

CMakeLists.txt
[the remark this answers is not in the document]
markdewing: Do you have an example on how to reproduce this failure (#685 is light on details)?

The iszero function only appears in fairly new glibc versions (>=2.25).  I would expect the test to even fail to compile on most systems.  And even if the system has it, and it fails to link, it should still not cause the cmake configuration step to fail.  The failure mode will be a 'redefinition of iszero' error during compilation. See #429 for more about the iszero test.

The c++11 flag test seems more likely to be a problem, where link flag problems will lead to the CMake configuration file stopping.

grahamlopez:
test this please

CMakeLists.txt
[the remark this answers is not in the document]
PDoakORNL: Updated #685, I should say the iszero fails, but it also halts configuration.

CMakeLists.txt
```diff
@@ -477,14 +489,15 @@ IF (ENABLE_GCOV)
   ENDIF()
 ENDIF(ENABLE_GCOV)
 
-# Require C++11 while leaving possibility to fall back by modifying this file
+# Check for the compiler C++11 flag, if we go to requiring >= cmake 3.1.3 we could drop this
+# and just write
+# set(CMAKE_CXX_STANDARD_REQUIRED 11 CACHE INTEGER "C++ stangard must be at least 11")
```
ye-luo: CXX11_NEEDED was introduced when we support compilers with and without C++11.
Now C++11 is mandatory.
For this reason, we should clean it up. I confess that I didn't place it in a good spot.
I think now it is time to solve it. We need to remove CXX11_NEEDED and then move the whole section before the MPI detection.

CMake/FindMKL.cmake
```diff
@@ -2,42 +2,106 @@
...
 INCLUDE( CheckCXXSourceCompiles )
 
+if ( NOT CMAKE_CXX_COMPILER_ID MATCHES "Intel" )
+  # if not intel and MKL_ROOT not set
+  if (NOT MKL_ROOT)
+    find_path(MKL_ROOT "mkl.h"
+      HINTS ${MKLROOT} ${MKL_HOME} $ENV{MKLROOT} $ENV{MKL_ROOT} $ENV{MKL_HOME}
+      PATH_SUFFIXES include)
+    string(REPLACE "/include" "" MKL_ROOT ${MKL_ROOT})
+  endif (NOT MKL_ROOT)
```
ye-luo: Open an ELSE and move the later MKL_INCLUDE_DIRECTORIES here. I don't think we need to do it twice.

CMake/FindMKL.cmake
```diff
@@ -2,42 +2,106 @@
...
+  endif (MKL_LINK_DIRECTORIES-NOTFOUND)
+  message("MKL_LINK_DIRECTORIES: ${MKL_LINK_DIRECTORIES}")
+  set(SUFFIXES include)
+  find_path(MKL_INCLUDE_DIRECTORIES name "mkl.h" HINTS ${MKL_ROOT}
+    PATH_SUFFIXES ${SUFFIXES})
+  if (MKL_INCLUDE_DIRECTORIES-NOTFOUND)
+    message(FATAL_ERROR "MKL_INCLUDE_DIRECTORIES not set. \"mkl.h\" not found in MKL_ROOT/(${SUFFIXES})")
+  endif (MKL_INCLUDE_DIRECTORIES-NOTFOUND)
+  message("MKL_INCLUDE_DIRECTORIES: ${MKL_INCLUDE_DIRECTORIES}")
+  set(MKL_FLAGS "-m64")
```
ye-luo: What is this flag? it seems compiler specific not MKL related.

CMake/FindMKL.cmake
```diff
@@ -2,42 +2,106 @@
...
+  set(MKL_LINKER_FLAGS "-L${MKL_LINK_DIRECTORIES} -Wl,-rpath,${MKL_LINK_DIRECTORIES}")
+  set(MKL_FLAGS "-I${MKL_INCLUDE_DIRECTORIES}")
+  set(MKL_LIBRARIES "-lmkl_intel_lp64 -lmkl_sequential -lmkl_core -lpthread -lm -ldl")
```
ye-luo: We need threaded MKL.

CMake/FindMKL.cmake
ye-luo: Actually this is overwritten two lines later.

CMake/FindMKL.cmake
PDoakORNL: My prefered use case is that MKL_ROOT is simply set and we don't depend on any environment variables.  However this section will support some mkl module setups I've seen.  I am not interested in making the main execution path more convoluted for to support this.

CMake/FindMKL.cmake
PDoakORNL: really?  Generally openmp codes run into performance problems is they also let mkl thread it self.  This just means there is not threading within mkl itself.

CMake/FindMKL.cmake
PDoakORNL: The overwrite is a mistake but it should really be in clang compilers.

CMakeLists.txt
PDoakORNL: I think this is outside of the scope of this PR which really is just so I and others can use llvm 5 and AST based tools while we work on qmcpack. I it in there but removed it because I expected pushback that would delay the PR.

CMake/FindMKL.cmake
ye-luo: Trust us.
Remember -mkl is default -mkl=parallel
https://software.intel.com/en-us/mkl-linux-developer-guide-using-the-mkl-compiler-option

CMake/FindMKL.cmake
ye-luo: After this PR, FindMKL should be general for GNU, Clang, PGI...
I don't see a reason to add this. From Intel linker advisor, there is no -m64 listed.

CMake/FindMKL.cmake
PDoakORNL: Ok, I trust you for the intel build, but it requires (as far as I can see) intels openmp and I haven't tested that at all on clang or gcc. I'm not using intels openmp or mixing openmp libraries anywhere else. I'd rather file that as an issue with this PR and deal with it as a separate concern.

markdewing: I'm also skeptical of linking with the threaded MKL.    It seems safer and more general to link with the sequential version.

CMake/FindMKL.cmake
PDoakORNL: So I or anyone else can use llvm 5 Abstract Syntax Tree based tools like rtags. Over time it could expand into full featured not intel compilers and MKL support but it doesn't need to for my goal to be accomplished.  It's not a full "feature" in one PR it is something I'm implementing so I can do more important work.

CMake/FindMKL.cmake
ye-luo: Only the matrix inversion used by the optimizer and FFT needs threaded MKL. All the rest MKL calls are inside OMP region and MKL will not turn on nested threading by default. I personally don't have any concern but if you guys prefer sequential. Just do sequential at the moment.

CMake/FindMKL.cmake
ye-luo: I see. Then leave it this way. Not here but in the later lines, could you add a bit more comment saying "detecting library" "detecting header files" "setting variables"? This will help readability.

CMakeLists.txt
ye-luo: OK. Let me handle that with a different PR. Unfortunately, that requires another checking on multiple platforms.

CMakeLists.txt
```diff
@@ -5,19 +5,29 @@ PROJECT(qmcpack)
...
-# Version 
+# Version
 ######################################################################
 SET(QMCPACK_VERSION_MAJOR 3)
 SET(QMCPACK_VERSION_MINOR 4)
 SET(QMCPACK_VERSION_PATCH 0)
 SET(QMCPACK_VERSION "${QMCPACK_VERSION_MAJOR}.${QMCPACK_VERSION_MINOR}.${QMCPACK_VERSION_PATCH}")
 
+MESSAGE( "INITIAL CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS}" )
+MESSAGE( "INITIAL CMAKE_CXX_FLAGS_RELEASE ${CMAKE_CXX_FLAGS_RELEASE}" )
```
ye-luo: If I'm not building release, why print RELEASE. I think users should not specifically touch RELEASE.
We actually have a problem. In later lines, we have
```
MESSAGE(STATUS "CXX compiler flags: ${CMAKE_CXX_FLAGS}")
MESSAGE(STATUS "CC  compiler flags: ${CMAKE_C_FLAGS}")
```
However, this is not the full flag. RELEASE and other flags are not printed together. Do you know if there is a way to print the full flag?

CMake/FindMKL.cmake
ye-luo: At the moment, you can always pass it in via CMAKE_CXX_FLAGS.

ye-luo:
```
CMake Warning (dev) in CMakeLists.txt:
  A logical block opening on the line

    /sandbox/opt/cleanup/PDoak/CMakeLists.txt:8 (IF)

  closes on the line

    /sandbox/opt/cleanup/PDoak/CMakeLists.txt:14 (endif)

  with mis-matching arguments.
```

CMakeLists.txt
ye-luo: I found the layout is not elegant. If it was added for debugging. Just remove both.

ye-luo:
I thinking about extend ENABLE_OUTSIDE_MKL to ENABLE_MKL which is set true with Intel compiler and false in other cases. This also allows us to turn off MKL even using Intel compiler. I remember someone asked about Intel+FFTW3 library.

CMakeLists.txt
```diff
@@ -539,9 +548,6 @@ IF ( HAVE_MPI )
   MESSAGE(STATUS "MPI runner is ${MPIEXEC}")
 ENDIF(HAVE_MPI)
 
-MESSAGE(STATUS "CXX compiler flags: ${CMAKE_CXX_FLAGS}")
```
ye-luo: Sorry, I was not asking to remove these ones but only those INITIAL ones. Please recover this.

grahamlopez:
test this please

CMakeLists.txt
PDoakORNL: I put them back with better output. Target properties can still change these.

grahamlopez:
test this please

ye-luo:
It gets harder to review the code in place. Some questions left.
1) HAVE_POSIX_MEMALIGN setting doesn't seems correct in Clang configuration. I can build with HAVE_POSIX_MEMALIGN=1 with clang 3.8 on linux (ubuntu with clang built by system admin)
2) Is org_CMAKE_CXX_FLAGS really needed? I didn't see temporal modification of CMAKE_CXX_FLAGS.

Some other things TODO but not in this PR.
1, Move CXX11 check before all the library detection.
2, Change ENABLE_OUTSIDE_MKL to ENABLE_MKL and make it general for all compilers and update the README and manual.

PDoakORNL:
@ye-luo I'm not sure when Clang patched for this but it was well before 5.0. Your build indicates we can at least detect 3.8.  So I'll change that.  I'll look again probably the CMAKE_CXX_FLAGS protector is no longer necessary.

I agree ENABLE_MKL would be more useful.

grahamlopez:
test this please

ye-luo:
@PDoakORNL I feel that it is better to remove that check. If you don't respect the alignment, the result can be wrong. It is worse than not being able to build the code.

PDoakORNL:
That's fine with me, it was in clang compilers when I started so I assumed it was important somewhere.  Dates back to @berrill  in 2016.

ye-luo:
@PDoakORNL we didn't seriously looked at the Clang configuration in the past and you helped us surface the potential risk. @berrill was doing AppleClang at that time but I think we care the standard clang only.

ye-luo:
@markdewing All my concern has been addressed. If no one asks for extra checks, we can merge this PR.

markdewing:
test this please

PR #698, opened 2018-03-20
updating scripts for new ci

grahamlopez:
This updates our current scripts for OLCF CI to use the new infrastructure.

### DIFF
--- a/tests/test_automation/jenkins_rhea_cpu.sh
+++ b/tests/test_automation/jenkins_rhea_cpu.sh
@@ -99,14 +99,9 @@ ctest -L unit
 
 EOF
 
-cp $BUILD_TAG.pbs $BUILD_DIR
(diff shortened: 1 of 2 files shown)


grahamlopez:
Old CI is expected to fail here, new CI expected to succeed. After this is merged, this will be the case for all following CI. At that point, I'll turn off the old (failing) robots.

markdewing:
It appears the old CI failures prevent merging.  Do we need to turn of the protections temporarily, or is there some other way to merge this?

grahamlopez:
I updated the scripts in place. It seems we have two choices:
1. The other choice would be to copy the scripts to have both the old and the new, merge the new scripts, then follow with another PR to remove the old scripts.
2. Disable protections to merge.

I don't mind either way.

prckent:
I'll temporarily switch the protections.

PR #702, opened 2018-03-22
Update T-move size-consistent algorithm for consistency with published version

ye-luo:
The previously added v1 size-consistent Tmove is not exactly the same as the M.Casula et al 2010 paper. Now the exact v1 is added. The previous version is renamed as v3. v3 gives the same result as v1 but v3 needs much less computation.


ye-luo:
No. Too slow and very invasive, it needs local energy every single particle move.

manual/dmc.tex
```diff
@@ -37,7 +37,7 @@ \section{Diffusion Monte Carlo}
    &   \texttt{reconfiguration     } &  string  & yes/pure/other & no   & fixed population technique  \\
    &   \texttt{branchInterval      } &  integer  & $\ge 0$ & 1   & branching interval \\
    &   \texttt{substeps            } &  integer  & $\ge 0$ & 1   & branching interval \\
-   &   \texttt{nonlocalmoves       } &  string  & yes/v0/v1/no & no   & run with tmoves  \\
+   &   \texttt{nonlocalmoves       } &  string  & see below & no   & run with tmoves  \\
```
prckent: Please update with full list of options. yes/v0/v1/v2/no

manual/dmc.tex
```diff
@@ -99,8 +99,8 @@ \section{Diffusion Monte Carlo}
 \item \texttt{substeps}.  Same as BranchInterval.
 
 
-\item \texttt{nonlocalmoves}. Evaluate pseudopotentials using one of the nonlocal move algorithms such as t-moves.  Setting nonlocalmoves to `no' will impose the locality approximation. `yes/v0' implements the algorithm in the 2006 Casula paper~\cite{Casula2006} and `v1' implements the v1 algorithm in the 2010 Casula paper~\cite{Casula2010}.
-The v1 algorithm is size-consistent and an important advance over the previous v0 non-size-consistent algorithm. Investigating the importance of size consistency is highly recommended.
+\item \texttt{nonlocalmoves}. Evaluate pseudopotentials using one of the nonlocal move algorithms such as t-moves.  Setting nonlocalmoves to `no' will impose the locality approximation. `yes/v0' implements the algorithm in the 2006 Casula paper~\cite{Casula2006} and `v1' implements the v1 algorithm in the 2010 Casula paper~\cite{Casula2010} and `v3' implements an algorithm similar to v1 but much faster.
+The v1 and v3 algorithms are size-consistent and an important advance over the previous v0 non-size-consistent algorithm. Investigating the importance of size consistency is highly recommended.
```
prckent: Update the language here to be more precise and give more details: is v3 exactly the same in terms of answer as v1 or is it an approximation? Why is it faster?

prckent: v3 should be renamed v2. Options should be numbered sequentially.

tests/molecules/H2O_dimer_sep_pp/long-H2O-TM3.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0"?>
```
prckent: TM3 -> TM2 consistent with version renaming.

manual/dmc.tex
ye-luo: Unfortunately, that row goes too wide. I prefer not to list all options.

prckent: TeX harder. This is a common option. We can have two lines, smaller fonts...

prckent:
Ignore my v2 comments. There is a v2 algorithm in the Casula paper. We should avoid using v2 to minimize confusion. Add a note that v2 from that paper is not implemented.

PR #703, opened 2018-03-23
Add unit test for VMCSingleOMP

markdewing:
For ConstantOrbital.h - added one 'override', but then clang complained about all the
 other methods without it.  Using 'override' pointed out a virtual method that was no
 longer overriding anything - remove it.

Add diffuse.py for computing unit test values (the values were slightly different than
 the existing ones in test_vmc.cpp - update those to match.)
(This follows a general trend of putting the generic derivation/explanantion in
qmc_algorithms, and putting the particular python script for computing the unit test
values along side the unit tests.)

Starting point for tests for SimpleFixedNodeBranch, QMCUpdateBase, CloneManager.  The classes are
set up, but there are no tests there yet.

Added a constructor for initializing CloneManager to a number of threads specified by the test,
not just the number of OpenMP threads.

### DIFF
--- a/src/QMCDrivers/CloneManager.cpp
+++ b/src/QMCDrivers/CloneManager.cpp
@@ -51,10 +51,12 @@ std::vector<std::vector<TrialWaveFunction*> > CloneManager::PsiPoolClones;
 std::vector<std::vector<QMCHamiltonian*> > CloneManager::HPoolClones;
 
 
-// Clear the static clones.  For now only clear wClones so makeClones will work.
-// Used in unit tests.
+// Clear the static clones so makeClones will work as expected.
+// For now only clearing wClones is strictly necessary.  The storage is not freed,
+// and this will leak memory.
+// Only for use in unit tests.
 void
-CloneManager::clear()
+CloneManager::clear_for_unit_tests()
 {
   wClones.clear();
 }
--- a/src/QMCDrivers/CloneManager.cpp
+++ b/src/QMCDrivers/CloneManager.cpp
@@ -50,6 +50,15 @@ std::vector<std::vector<MCWalkerConfiguration*> > CloneManager::WPoolClones;
 std::vector<std::vector<TrialWaveFunction*> > CloneManager::PsiPoolClones;
 std::vector<std::vector<QMCHamiltonian*> > CloneManager::HPoolClones;
 
+
+// Clear the static clones.  For now only clear wClones so makeClones will work.
+// Used in unit tests.
+void
+CloneManager::clear()
+{
+  wClones.clear();
+}
+
 /// Constructor.
 CloneManager::CloneManager(HamiltonianPool& hpool): cloneEngine(hpool)
 {
--- a/src/QMCDrivers/CloneManager.cpp
+++ b/src/QMCDrivers/CloneManager.cpp
@@ -57,6 +57,14 @@ CloneManager::CloneManager(HamiltonianPool& hpool): cloneEngine(hpool)
   wPerNode.resize(NumThreads+1,0);
 }
 
+void
+CloneManager::setup(int numThreads)
+{
+  NumThreads = numThreads;
+  wPerNode.resize(NumThreads+1, 0);
+}
+
+
 ///cleanup non-static data members
 CloneManager::~CloneManager()
 {
--- a/src/QMCDrivers/tests/test_dmc_omp.cpp
+++ b/src/QMCDrivers/tests/test_dmc_omp.cpp
@@ -85,6 +85,7 @@ TEST_CASE("DMCOMP", "[drivers][dmc]")
   elec.addTable(ions,DT_AOS);
   elec.update();
 
+  CloneManager::clear();
 
   TrialWaveFunction psi = TrialWaveFunction(c);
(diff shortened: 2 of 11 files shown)


ye-luo:
question on the override. I thought only when you use this specifier, compliers do the check. You are saying clang is doing more than that?

markdewing:
Clang goes beyond the standard override check and issues a warning (-Winconsistent-missing-override), where if one method in a class has the 'override' keyword, it will issue warnings for other methods that should have it, but don't.  Presumably, if a programmer uses 'override' on one method, it signals intent to use the feature, and the warning is helpful.   I like the override keyword and the checking it provides (it's very easy to not get the signature exactly right in the derived class, and then be mystified as to why the method in the derived class is not getting called at runtime.)

ye-luo:
USE_FAKE_RNG in the VMCSingleOMP doesn't seem clean solution. It is fine for the current PR but in long term, more cleaning is needed in the drivers and I prefer to have as many as possible operations of clones within the CloneManager.
Why the RNG manipulation didn't occur in DMC driver unit test?

markdewing:
There is no change in DMC because I haven't written that test yet.  The same change will be needed in DMCOMP.cpp.
The issue is passing the fake random number generator to  the Hamiltonian.   The way the fake RNG is currently constructed, it would require making a separate unit-test version of the qmcham library.  A separate unit-test specific version of qmcdriver is already being built, but not for qmcham.

Two ways to do the RNG classes is to either make it a base class with actual implementations as derived classes (performance concerns?), or to template everything on the RNG (will wind up with extra code, but the mangled names will keep the implementations for different RNG's separate)

src/QMCDrivers/CloneManager.cpp
```diff
@@ -50,13 +50,30 @@ std::vector<std::vector<MCWalkerConfiguration*> > CloneManager::WPoolClones;
 std::vector<std::vector<TrialWaveFunction*> > CloneManager::PsiPoolClones;
 std::vector<std::vector<QMCHamiltonian*> > CloneManager::HPoolClones;
 
+
+// Clear the static clones.  For now only clear wClones so makeClones will work.
+// Used in unit tests.
```
ye-luo: Please emphasize more that it is not for general use and should be used only by unit tests. Please also add the same comment in header file.

ye-luo: In general, I don't think the clear method is safe since it erases the vector of pointers without freeing the pointers.

src/QMCDrivers/tests/test_dmc_omp.cpp
```diff
@@ -0,0 +1,146 @@
...
+  int massIdx = tspecies.addAttribute("mass");
+  tspecies(chargeIdx, upIdx) = -1;
+  tspecies(chargeIdx, downIdx) = -1;
+  tspecies(massIdx, upIdx) = 1.0;
+  tspecies(massIdx, downIdx) = 1.0;
+
+  elec.addTable(ions,DT_AOS);
+  elec.update();
+
+  CloneManager::clear();
```
ye-luo: I prefer to put this at the end if possible.

src/QMCDrivers/tests/test_dmc_omp.cpp
```diff
@@ -0,0 +1,146 @@
...
+  int upIdx = tspecies.addSpecies("u");
+  int downIdx = tspecies.addSpecies("d");
+  int chargeIdx = tspecies.addAttribute("charge");
+  int massIdx = tspecies.addAttribute("mass");
+  tspecies(chargeIdx, upIdx) = -1;
+  tspecies(chargeIdx, downIdx) = -1;
+  tspecies(massIdx, upIdx) = 1.0;
+  tspecies(massIdx, downIdx) = 1.0;
+
+  elec.addTable(ions,DT_AOS);
```
ye-luo: Please update all DT_AOS to DT_SOA or put ifdef for both cases.

src/QMCDrivers/CloneManager.cpp
markdewing: I'll add more comments and change the name to 'clear_for_unit_tests'

Not freeing the pointers is safe - it will leak memory, but that's okay for now.   The alternative is to free all the pointers, and then we have to be very careful put only dynamically allocated objects in there (that is, it will crash the unit tests as currently written).   Ultimately, the correct solution is to use some form of smart pointer.

src/QMCDrivers/tests/test_dmc_omp.cpp
markdewing: With putting it at the end, I'm concerned that if one test exits early due to a failure, the 'clear' won't get called, and it could lead to a cascading failure in a different test.

src/QMCDrivers/CloneManager.cpp
ye-luo: Renaming the function is a very good solution. As long as it is sufficient for unit tests, no need to spend time investigating how to make it fully safe.

src/QMCDrivers/tests/test_dmc_omp.cpp
prckent: We should aupport both cases

PR #712, opened 2018-03-26
Add test for TraceManager.

markdewing:
It appears there exists test code to check compilation in TraceManager.cpp.
Move that to a unit test and fix it so it compiles and runs.

The operations in the first test may not be called in the correct order and may not even be consistent.   They do run without crashing, though.  
I would appreciate input on what the correct calling sequence should be.


src/Estimators/tests/CMakeLists.txt
```diff
@@ -20,6 +20,10 @@ SET(UTEST_NAME unit_test_${SRC_DIR})
 
 SET(SRCS test_accumulator.cpp test_local_energy_est.cpp test_manager.cpp)
 
+IF (NOT REMOVE_TRACEMANAGE)
```
prckent: What is the concern about this dependency? Also, REMOVE_TRACEMANAGER not TRACEMANAGE,

markdewing: It was to match some conditionals in the code. 
Now i tested compilation with REMOVE_TRACEMANAGER, and the unit test compiles and runs okay.  I'll remove the conditional.

grahamlopez:
test this please

grahamlopez:
test this please

grahamlopez:
test this please

grahamlopez:
test this please

grahamlopez:
test this please

PR #713, opened 2018-03-27
fix for unsupported -march=native on summit

grahamlopez:
we could replace with -mcpu=native, but when I did this I got lots of warnings of the sort:
```
/usr/bin/ld: LPQHIBasis.cpp:(.text+0x1bb8): toc optimization is not supported for 0xe6a90002 instruction.
```
so for now, just disabling completely.


ye-luo:
Can we put the protection in the upper level as the miniqmc? The reason is any -march is invalid for ppc64.

grahamlopez:
yes I see your point. I did this to let the user still use custom flags, but maybe we should protect them from themselves...

prckent:
@grahamlopez We should be consistent with miniqmc

@ye-luo , others: Are there any other systems identifying as ppc64 that we need to guard for?

grahamlopez:
@prckent, yes, sorry it wasn't clear, I'm agreeing with @ye-luo. 
I think this should guard anything that identifies as ppc64, not just summit

prckent:
What do the BG head nodes identify as?

ye-luo:
@grahamlopez In this [PR](https://github.com/QMCPACK/miniqmc/pull/87) on miniQMC the clang config is also changed like GNU accordingly. Share we do the same?
@prckent XL flags have significant changes since power8 (first little endian). Our XL configuration is targeting that. BGQ uses toolchain file.

grahamlopez:
@ye-luo interestingly, clang on summit doesn't seem to have a problem with `-march=native` If you want me to add the guard anyway, I don't mind

grahamlopez:
test this please

grahamlopez:
test this please

ye-luo:
I tried the clang module and got
```
clang -march=native hello.cpp
clang-3.8: warning: argument unused during compilation: '-march=native' [-Wunused-command-line-argument]
```
So the guard seems necessary.

grahamlopez:
test this please

PR #717, opened 2018-03-28
Fix unit test linking issue

ye-luo:
Happens in certain scenario.
Adjust the linking order to qmcham before qmcwfs because qmcham depends on qmcwfs.

### DIFF
--- a/src/Estimators/tests/CMakeLists.txt
+++ b/src/Estimators/tests/CMakeLists.txt
@@ -27,7 +27,7 @@ SET(SRCS test_accumulator.cpp test_local_energy_est.cpp test_manager.cpp test_tr
 
 
 ADD_EXECUTABLE(${UTEST_EXE} ${SRCS})
-TARGET_LINK_LIBRARIES(${UTEST_EXE} qmc qmcwfs qmcham qmcbase qmcutil qmcdriver ${QMC_UTIL_LIBS} ${MPI_LIBRARY})
+TARGET_LINK_LIBRARIES(${UTEST_EXE} qmc qmcham qmcwfs qmcbase qmcutil qmcdriver ${QMC_UTIL_LIBS} ${MPI_LIBRARY})
 
 #ADD_TEST(NAME ${UTEST_NAME} COMMAND "${QMCPACK_UNIT_TEST_DIR}/${UTEST_EXE}")
 ADD_UNIT_TEST(${UTEST_NAME} "${QMCPACK_UNIT_TEST_DIR}/${UTEST_EXE}")


prckent:
This seems like a workaround for current cmake behavior that might not be guaranteed in future.
Is not the bug that qmcham does not specify qmcwfs as a dependency?

PDoakORNL:
Why not?
```cmake
#src/QMCHamiltonians/CMakeLists.txt
...
IF(QMC_CUDA)
  CUDA_ADD_LIBRARY(qmcham ${HAMSRCS} qmcwfs)
ELSE(QMC_CUDA)
  ADD_LIBRARY(qmcham ${HAMSRCS})
ENDIF(QMC_CUDA)
target_link_libraries(qmcham qmcwfs)
...
```

markdewing:
What scenario causes this error?

ye-luo:
@prckent I don't think this is a workaround. Source file dependency is imposed by CMake but CMake basically follows how you link libraries like command line and preserve the order written in cmakelists.txt. Maybe the word 'depend' is not very accurate. It is like lapack needs blas but you don't want blas object inside lapack library. To have lapack/blas linked properly, lapack is placed before blas.

@PDoakORNL The same as lapack/blas. We don't want to include qmcwfs inside qmcham.

@markdewing When I made a modification in QMCHamiltonian.cpp, I got the following
```
/soft/apps/packages/mpich-3.2-intel17-initial/bin/mpicxx   -std=c++11  -qopenmp -Wno-deprecated -qopt-prefetch -ftz -xHost -mkl -O3 -DNDEBUG -restrict -unroll -ip   CMakeFiles/test_estimators.dir/test_accumulator.cpp.o CMakeFiles/test_estimators.dir/test_local_energy_est.cpp.o CMakeFiles/test_estimators.dir/test_manager.cpp.o CMakeFiles/test_estimators.dir/test_trace_manager.cpp.o  -o ../../../tests/bin/test_estimators -rdynamic ../../../lib/libqmc.a ../../../lib/libqmcwfs.a ../../../lib/libqmcham.a ../../../lib/libqmcbase.a ../../../lib/libqmcutil.a ../../../lib/libqmcdriver.a ../../../lib/libformic_utils.a -lxml2 /soft/apps/packages/climate/hdf5/1.8.16-parallel/gcc-6.2.0/lib/libhdf5.so /soft/apps/packages/climate/szip/2.1/gcc-6.2.0/lib/libsz.so -lz -ldl -lm -lz ../../../lib/libeinspline.a ../../../lib/liboompi.a -ldl -lm -Wl,-rpath,/soft/apps/packages/climate/hdf5/1.8.16-parallel/gcc-6.2.0/lib:/soft/apps/packages/climate/szip/2.1/gcc-6.2.0/lib 
../../../lib/libqmcham.a(NonLocalECPComponent.cpp.o): In function `qmcplusplus::NonLocalECPComponent::evaluateOne(qmcplusplus::ParticleSet&, int, qmcplusplus::TrialWaveFunction&, int, double, qmcplusplus::TinyVector<double, 3u> const&, bool, std::vector<qmcplusplus::NonLocalData, std::allocator<qmcplusplus::NonLocalData> >&) const':
NonLocalECPComponent.cpp:(.text+0x4108): undefined reference to `qmcplusplus::TrialWaveFunction::evaluateRatios(qmcplusplus::VirtualParticleSet&, std::vector<double, std::allocator<double> >&)'
NonLocalECPComponent.cpp:(.text+0x4526): undefined reference to `qmcplusplus::TrialWaveFunction::ratio(qmcplusplus::ParticleSet&, int)'
../../../lib/libqmcham.a(NonLocalECPComponent.cpp.o): In function `qmcplusplus::NonLocalECPComponent::evaluate(qmcplusplus::ParticleSet&, int, qmcplusplus::TrialWaveFunction&, qmcplusplus::TinyVector<double, 3u>&)':
...
```

prckent:
Any significant objections to merging as is (@markdewing , @PDoakORNL )?

markdewing:
I'm okay with merging this.

PDoakORNL:
Looking at this dependency in the source I think target_link_library is the correct way to express this dependency. This way if you have an exe that is only dependent on qmcham it does no also have to link in qmcwfs. CMake handles any cyclic static library dependencies that might arise and you don't have to get the order correct everytime.

ye-luo:
@PDoakORNL I see your point. Since this kind of dependence is not something we are currently taking care of. We need to fix all the places in our CMake together.

PDoakORNL:
I don't think we need to fix everywhere at once, but this causes no problem to doing that so it should just get merged, I'd do it if I could.

PR #719, opened 2018-03-29
Add AFQMC integration tests

jtkrogel:
Add integration tests for diamond, AFM NiO, and NM NiO AFQMC.  

NiO systems did not run on Eos at OLCF but were generated recently at LLNL.  @mmorale3 is looking into the issue.  These tests are likely to fail until the issue is corrected.  Once corrected, the target error bars of the tests will likely need to be adjusted.  Large files needed to run these examples have been uploaded to the ANL Box site.

Several variants are tested for diamond (see the README).  Variants exercising the parallelization option nnodes fail, in part due to incompatibility with the environment on Eos.  As written, to be successful this subset of tests needs to be run on two physical nodes, which the AFQMC code attempts to detect automatically regardless of the number of e.g. MPI processes provided.  This auto-detection appears to fail on Eos, and in any event would not allow this option to be exercised/tested on a single node.  Recommend that a new option e.g. nnodes be created to provide user override of the autodetect feature and enable these tests to run successfully on a single node.


prckent:
Regarding multiple node capabilities for testing, this is a general feature that will need some development with updates to cmakelists and to the job scripts, with graceful handling of the case when only 1 node is available.

jtkrogel:
I thought QMC_DATA was a one-time download of a single directory (with subdirs).  Provided the directory is downloaded and QMC_DATA is set appropriately, the performance, NiO DMC, and NiO AFQMC tests will all work simultaneously.

I will note in the README the specific files that are contained in the QMC_DATA directory.

jtkrogel:
@prckent done.

prckent:
QMC_DATA is sadly a file by file download from Box. We do not have an automated solution. Perhaps this can be developed. Box is a poor solution for our use case, but it is the least worst general access hosting we know about.

PR #722, opened 2018-03-29
Fix pw2qmcpack tests

ye-luo:
pw2qmcpack tests where updated when QMC_RUN_AND_CHECK  API was changed last time.
The improved checking surfaced the issue. The [nightly tests](https://cdash.qmcpack.org/CDash/viewConfigure.php?buildid=20567) failed.
With this fix, qe6.2.1 passes all the tests. No concern.

I also have two minor changes in periodic gaussian.


prckent:
Please keep separate updates separate. Make this two PRs.

PR #726, opened 2018-03-30
Add unit tests for spline2

markdewing:
Add scripts for creating unit test values.  The scripts derive the spline functions symbolically, compute the coefficients, and evaluate the spline and derivatives.


markdewing:
I also updated the bspline notebook on qmc_algorithms to add more text and to solve for coefficients with periodic functions
https://github.com/QMCPACK/qmc_algorithms/blob/master/Wavefunctions/Explain_Bspline.ipynb

ye-luo:
Can we have more template functions? For example when testing SplineBound, it is better to reduce duplicated lines.

PR #737, opened 2018-04-03
Remove unused evaluateHybridSplineComplexToReal_NLPP_kernel

prckent:
Removes the kernel evaluateHybridSplineComplexToReal_NLPP_kernel which is never called. The function that invokes this was already commented out and has also been removed.  The current complex path does not need it. There is no equivalent real kernel, so I presume that this is an unused "optimization".

Removing this code removes the unused MaxQuad variable definition which causes the only warning other than MPI for GPU builds of the code with default compiler flags.


ye-luo:
Since CI currently doesn't check complex build, please comment whether a GPU complex build is fine or not.

prckent:
I checked both real and complex builds. gcc 4.8.5, Intel MPI, CUDA 9.1
```
 cmake -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx -DQMC_CUDA=1 -DENABLE_TIMERS=1 -DQMC_DATA=/data/pk7/QMC_DATA/ ../qmcpack/
 cmake -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx -DQMC_CUDA=1 -DENABLE_TIMERS=1 -DQMC_DATA=/data/pk7/QMC_DATA/ -DQMC_COMPLEX=1 ../qmcpack/
```

PR #746, opened 2018-04-06
Nexus: add support for hybrid representation

jtkrogel:
Add support for hybrid representation and other small changes for running on Mira.


markdewing:
okay to test

Issue #715, opened 2018-03-28
Role of QMC_DATA to be added to manual


prckent:
This has been needed for the performance tests and described in the README.
With #714 it is now required for some of the integration tests.

(QMC_DATA points to the location of large files that we can not reasonably manage with Git).

PR #749, opened 2018-04-06
Update manual to cover QMC_DATA and ENABLE_TIMERS

### LINKED ISSUE #688: ENABLE_TIMERS not described in manual
This is missing from the cmake configure section and readme.md

prckent:
Closes #688 and #715


prckent:
I updated the top level README.md to be consistent with the manual. It mentions the newer options. I also updated the example build.

PR #752, opened 2018-04-08
Adjust one body Jastrow SoA algorithm

ye-luo:
The SoA J1 was written not using the memory pool while J2 and JeeI do.
When a walker with its memory pool loaded, Det and J2 JeeI have the up-to-date value but J1 doesn't.
This is not a problem after all the electrons are proposed to move once but not from the beginning.
This potential risk is secured by checking the ParticleSet.Ready4Measure flag.
However, the flag is delicate and needs a careful maintaining.

This PR changes the algorithm to be aligned as J2 and JeeI. No more need to recompute since it is loaded from the memory pool and the needed size is small, order Nelec.
The state flag ParticleSet.Ready4Measure is also removed.

This PR also fixed the missing names of ParticleSet objects in Sandbox apps.

### DIFF
--- a/src/Particle/ParticleSet.cpp
+++ b/src/Particle/ParticleSet.cpp
@@ -456,7 +456,6 @@ void ParticleSet::update(bool skipSK)
   if (!skipSK && SK)
     SK->UpdateAllPart(*this);
 
-  Ready4Measure=true;
   activePtcl=-1;
 }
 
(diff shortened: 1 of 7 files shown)


prckent:
+1 I am in favor of having one less fragile state machine in the code.  I think the only place where a state machine is required is the drivers, which are explicit state machines.

Are there any more of these in the SoA code?

Issue #753, opened 2018-04-09
All-particle move drivers have bugs


ye-luo:
DMC with and without Tmove are giving bad numbers in the kinetic and Local potential energy terms in the scalar.dat file while the local energy is fine.
```
                            LocalEnergy               Variance           Kinetic                LocalPotential
allp-3.2-aos-no  -10.531050 +/- 0.000184   0.376721 +/- 0.000799   11.483801 +/- 0.001966   -22.014851 +/- 0.002007    
allp-3.2-soa-no  -10.531123 +/- 0.000180   0.377863 +/- 0.002306   11.480341 +/- 0.001736   -22.011464 +/- 0.001765    
allp-3.3-aos-no  -10.531178 +/- 0.000185   0.375630 +/- 0.000321   11.394282 +/- 0.035830   -21.937085 +/- 0.033099    
allp-3.3-soa-no  -10.530922 +/- 0.000183   0.376737 +/- 0.000965   11.509506 +/- 0.055000   -22.014785 +/- 0.046328    
allp-3.4-aos-no  -10.530664 +/- 0.000186   0.376353 +/- 0.000450   11.464139 +/- 0.024494   -21.997251 +/- 0.022930    
allp-3.4-soa-no  -10.531169 +/- 0.000175   0.375465 +/- 0.000334   11.346662 +/- 0.235234   -21.985914 +/- 0.134114    
allp-fix-aos-no  -10.531156 +/- 0.000179   0.375673 +/- 0.000385   11.478436 +/- 0.001857   -22.009592 +/- 0.001894    
allp-fix-soa-no  -10.530796 +/- 0.000180   0.375554 +/- 0.000358   11.484011 +/- 0.001833   -22.014807 +/- 0.001880    

allp-3.2-aos-yes  -10.528419 +/- 0.000174   0.378854 +/- 0.000270   11.525439 +/- 0.001779   -22.053515 +/- 0.001816    
allp-3.2-soa-yes  -10.528126 +/- 0.000174   0.378600 +/- 0.000275   11.523731 +/- 0.001843   -22.051530 +/- 0.001886    
allp-3.3-aos-yes  -10.528237 +/- 0.000176   0.379476 +/- 0.000514   11.505911 +/- 0.027522   -21.991726 +/- 0.041777    
allp-3.3-soa-yes  -10.528385 +/- 0.000178   0.378953 +/- 0.000312   11.451278 +/- 0.073488   -21.943543 +/- 0.108479    
[...]

PR #754, opened 2018-04-09
Fix all-particle move DMC driver

ye-luo:
See more details in #753 for the bug and use that issue to track progress.

This PR adds an integration test for VMC and DMC with all particle moves on diamondC 1x1x1.
Reference values of VMC and DMC are taken from pbyp moves.

This PR largely rewrite the DMC driver routine makes the DMC tests passing but VMC part remains failing. PR #423 also touched the VMC driver but I have no idea if the VMC was broken even before.

### DIFF
--- a/src/QMCDrivers/DMC/DMCUpdateAll.cpp
+++ b/src/QMCDrivers/DMC/DMCUpdateAll.cpp
@@ -49,67 +49,73 @@ void DMCUpdateAllWithRejection::advanceWalker(Walker_t& thisWalker, bool recompu
 {
     W.loadWalker(thisWalker,false);
     //create a 3N-Dimensional Gaussian with variance=1
-    RealType nodecorr=setScaledDriftPbyPandNodeCorr(Tau,MassInvP,W.G,drift);
+    RealType nodecorr = setScaledDriftPbyPandNodeCorr(Tau,MassInvP,W.G,drift);
     //RealType nodecorr = setScaledDriftPbyPandNodeCorr(m_tauovermass,W.G,drift);
     makeGaussRandomWithEngine(deltaR,RandomGen);
-    //if(!W.makeMoveWithDrift(thisWalker,drift,deltaR, m_sqrttau))
-    if (!W.makeMoveWithDrift(thisWalker,drift ,deltaR,SqrtTauOverMass))
-    {
-      H.rejectedMove(W,thisWalker);
-      return;
-    }
     //save old local energy
-    RealType eold    = thisWalker.Properties(LOCALENERGY);
-    RealType signold = thisWalker.Properties(SIGN);
-    RealType enew  = eold;
-    //evaluate wave functior
-    RealType logpsi(Psi.evaluateLog(W));
-    if(UseTMove)
-      nonLocalOps.reset();
-    bool accepted=false;
+    RealType eold = thisWalker.Properties(LOCALENERGY);
+    RealType enew = eold;
+    bool accepted = false;
     RealType rr_accepted = 0.0;
-    nodecorr=0.0;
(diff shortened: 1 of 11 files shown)


Paul-St-Young:
@ye-luo I can confirm your updated DMC driver (dd3a976) outputs the correct values in both `scalar.dat` and `stat.h5`. Although I wonder if the amount of changes can be reduced. Please see the changes I applied to my branch [here](https://github.com/Paul-St-Young/qmcpack/commit/a2421400448738fa1145b3b04483fd555705b896). Basically, I separated calls to `auxH->evaluate` from calls to `auxH->setParticlePropertyList`. The `evaluate` routine accumulates `P.Collectables`, whereas `setParticlePropertyList` overwrites `P.PropertyList`. I believe the behavior difference, i.e. accumulate v.s. overwrite, is the main discrepancy between the `scalar.dat` and `stat.h5` collection paths.

Paul-St-Young:
I also realized that this bug extends to the pbyp driver as well. Namely, the `stat.h5` numbers are wrong when the acceptance rate is low. Basically, every time a move is rejected, `P.Collectables` loses statistics. This bug does not show up normally because there are so many accepted moves (one for each particle*~99%).

One can produce the bug only under special circumstances. Examples I can think of:
1. low acceptance due to large time step or bad wave function.
2. very long run where statistical noise is smaller than the bias.

I have produced the bug using method 1 below

allp-pbyp1_sofk

ye-luo:
@Paul-St-Young Thanks for checking. It is good that the driver is correct now. I largely rewrite the driver more for consistency. Your simple fix can make the S(k) correct but if a quantity depends on the WF, it is still wrong. For this reason, I prefer to have the whole WF recomputed if a move is rejected just like what we changed in the VMC driver. I really don't like `H.rejectedMove`. I get the point of your change of API in the H. However, at this point, I prefer not to add more APIs. I used to find so many useless APIs in TrialWavefunctions and the roles are not clear and logical. For the same reason, I prefer to clean up current APIs.

To be honest, I have several concerns in the implementation of data collection in both PropertyList and P.collectables. A lot of housekeeping work is needed.

ye-luo:
@Paul-St-Young please make an issue describing the different collected data in stat.h5 and scalar.dat. I will keep an eye on it when I clean up the estimators.

jtkrogel:
If I recall correctly, that statistical approach used to make stat.h5 matches scalar.dat for VMC, but for DMC it matches dmc.dat (which does not match scalar.dat!).  The difference in the weighting approaches should vanish with population size.  An observed loss of statistics is another matter, but this should also show up for scalar values in stat.h5 as well as energy normed quantities like the energy density.

prckent:
Concurring with @ye-luo 's comments on the APIs: we need to reduce these and document what we have. They are one of the reasons for code inaccessibility.

jtkrogel:
I strongly agree.

Paul-St-Young:
@ye-luo Thanks for the explanation. I totally agree we should keep the API under control. My [commit](https://github.com/Paul-St-Young/qmcpack/commit/a2421400448738fa1145b3b04483fd555705b896) was more an explanation than a real fix.

Paul-St-Young:
I will make a new PR with all I can find out about the `scalar.dat` v.s. `stat.h5` collection paths. I think the development/observables section would serve as a reasonable home for these explanations.

ye-luo:
@prckent Shall we merge this PR? I think the problem inside this driver is fully resolved.

ye-luo:
@Paul-St-Young Please add an issue for the test you have planned and assign it to yourself. Then make a new PR to close it.

Paul-St-Young:
Got it

prckent:
Approving. Important to complete #756 . For the time being, what advice should we give users regarding use of all particle moves?

PR #760, opened 2018-04-11
Initial testing framework for Nexus

jtkrogel:
Establish an initial testing framework for Nexus.  Currently only the highest level base class is tested thoroughly.

A new Python executable is introduced (ntest) that can be run by the user fully independently of ctest.  Output from this testing tool is similar to that produced by ctest:
```
>ntest

 1/5 generic_logging....................   Passed  0.00 sec
 2/5 generic_intrinsics.................   Passed  0.00 sec
 3/5 generic_extensions.................   Passed  0.02 sec
 4/5 nexus_imports......................   Passed  1.27 sec
 5/5 settings...........................   Passed  0.02 sec

100% tests passed, 0 tests failed out of 5

Total test time = 1.31 sec
```

A command line ctest compatibility mode has also been added, which will produce error codes and messages similar to check_scalars.py.  A new CMake function needs to be added to complete the integration with QMCPACK's CTest system.  Once added, all Nexus tests can be added in the following way:
```
RUN_NEXUS_TEST(generic_logging)
RUN_NEXUS_TEST(generic_intrinsics)
RUN_NEXUS_TEST(generic_extensions)
RUN_NEXUS_TEST(nexus_imports)
RUN_NEXUS_TEST(settings)
```

With the first command simply running: 
```
ntest -R generic_logging --ctest
```
at the command line.


jtkrogel:
@markdewing can you advise how such a CMake function can be added?

markdewing:
There are two options for running ntest
 - running one test at a time (as you suggest)
 - run all the ntest tests at once (as the unit test executables)

The advantage of running the nexus tests individually is the failed test name is more visible when a test fails. Also there can be tests in ntest that are not tested in cmake (in progress, require special environment, etc.)
The disadvantage is it requires duplicate specification - once in ntest, and then again in the CMakeLists.txt file.

I would recommend moving ntest to a new tests/ directory.  This will separate it from the executables used for analyzing qmcpack output.
Also, to keep ntest more scalable for adding more tests, I would recommend putting the tests in separate files from ntest. (Doesn't have to be done in this PR).  The tests/ directory would be a good place to put these files.

The other issue is setting the python path so ntest can find the nexus library.
Adding these lines to locate the library directory relative to the ntest file seems to work
```
scriptpath=os.path.realpath(__file__)
up_one_dir=os.path.dirname(os.path.dirname(scriptpath))
sys.path.append(os.path.join(up_one_dir,"library"))
from generic import generic_settings,object_interface,NexusError
```

Example CMakeLists.txt file (in the nexus/tests directory) (Both options for running are present, should choose and use only one)
```
function(RUN_NEXUS_TEST TEST_NAME)
  #message("Adding nexus test ${TEST_NAME}")
  set(NTEST "${qmcpack_SOURCE_DIR}/nexus/tests/ntest")
  add_test(NAME nexus_${TEST_NAME} COMMAND ${NTEST} -R ${TEST_NAME} --ctest)
  set_tests_properties(nexus_${TEST_NAME} PROPERTIES LABELS "nexus")
endfunction()

RUN_NEXUS_TEST("generic_logging")
RUN_NEXUS_TEST("generic_intrinsics")
RUN_NEXUS_TEST("generic_extensions")
RUN_NEXUS_TEST("nexus_imports")
RUN_NEXUS_TEST("settings")


# Alternate code to run all the ntest tests at once

function(RUN_ALL_NEXUS_TESTS)
  set(NTEST "${qmcpack_SOURCE_DIR}/nexus/tests/ntest")
  add_test(NAME all_nexus_tests COMMAND ${NTEST})
  set_tests_properties(all_nexus_tests PROPERTIES LABELS "all_nxs")
endfunction()

RUN_ALL_NEXUS_TESTS()
```

jtkrogel:
@markdewing thanks for your detailed response.  It looks like the CMakeLists.txt could be placed outside the nexus directory, correct?  I would prefer to keep ctest related tests/scripts outside the Nexus directory for better modularity between Nexus and QMCPACK.

The executables directory is really for any executable associated with Nexus (analysis+).  One reason is to keep "installation" steps to the bare minimum (add nexus/library to PYTHONPATH, add nexus/executables to PATH).

ye-luo:
CMake hierarchy needs to be consistent with nexus to keep good modularity. If I remove the nexus folder, all pieces related to nexus should be either gone or slient. I prefer a `nexus/tests` folder than the top level `tests` folder for things specific to nexus.

jtkrogel:
@ye-luo would your concern be resolved if a directory check is included?  I imagine this is not hard to perform within CMake.

markdewing:
Alternate opinion:
  There is already a CMakeLists.txt file in nexus/examples, and a directory check for nexus/ in the top-level CMakeLists.txt file.
  There necessarily needs to be some piece that is out of place, either in qmcpack or in nexus.   I think it's better to have the CMake code that knows about the details of nexus closer to the nexus code.    If nexus is used independently of qmcpack, then the CMakeLists.txt files in nexus can be ignored.

jtkrogel:
I am OK with adding the necessary CMakeLists.txt to nexus.

@markdewing I like your suggestion to have ntest provide the full list of tests by writing a file.  What is needed in CMake to read/parse a file to extract the list of test names?

markdewing:
Rather than writing to a file, it would also work to use 'execute_process'  to run ntest with an appropriate argument and get the list on stdout (RESULT_VARIABLE argument to execute_process).   The list could either be in a cmake list format (separated by semicolons), or separated by spaces and converted to a list with 'separate_arguments'

prckent:
I believe that this is ready to be merged and effectively reviewed.

Issue #756, opened 2018-04-10
scalar.dat and stat.h5 collection paths are inconsistent


Paul-St-Young:
Track progress of estimator collection fixes per discussion in #754.

Goals of this issue:
- [x] Document collection paths for `scalar.dat` and `stat.h5` in development section of manual.
- [x] Add breaking examples that show `scalar.dat`-`stat.h5` discrepancy in current code.
- [ ] Fix breaking examples

Symptoms of the current bug has manifested in #419, #504, and #753. Time to fix it once and for all.

jtkrogel:
@ye-luo fixed now?

ye-luo:
I tried batched driver and noticed that collectables are not written to the stat.h5 in batched drivers.
https://github.com/QMCPACK/qmcpack/issues/3047
It is not clear to me how much estimators are ported to batched drivers.
Regarding classic drivers, I don't think anything changed.

prckent:
In batched drivers, does anything other than spindensity work? I think the plumbing is simply not there &/or needs to be totally redone. (Maybe @PDoakORNL ?)

jtkrogel:
I know that the spin density works with the batched drivers, but only when in the qmc sections (not hamiltonian) and with special xml.

Is this the current situation: each estimator needs to be refactored specifically for use with the batched drivers?

jtkrogel:
Handled

PR #371, opened 2017-09-12
Rename EstimatorManager EstimatorManagerBase.  

rcclay:
This is pull request 1/3 to resolve issue #214.  This renames EstimatorManager to EstimatorManagerBase, and updates all references to EstimatorManager in the code to EstimatorManagerBase.  While just a rename for now, the next pull request will split current functionality between a derived class EstimatorManager and EstimatorManagerBase.

PR #764, opened 2018-04-12
Add sofk estimator test and manual entry about how estimator works

Paul-St-Young:
**This PR follows up on #756 but only introduce a test and manual updates**

Documenting estimator manager has turned out to be a lot more involved than I expected.
I decided to upload the test examples first to get the discussion started.

The new tests can be ran using `ctest -R estimator-sofk`. I added two `SkEstimator` estimators per run. One outputs to `scalar.dat` while the other outputs to `stat.h5`. DMC is done using a very large time step and 1 step per block. This is a severe test of the ensemble average procedure because both walker weight and walker number undergo large fluctuation.

In the Python script `check_collectables_h5dat`, I get the S(k) value at the largest k vector and compare between `scalar.dat` and `stat.h5`. S(k->large) should fluctuation around 1.0 in both cases. Currently, `scalar.dat` fluctuates around 1.0 but `stat.h5` does not.

sofk_dat-h5


Paul-St-Young:
@markdewing is there an easy way to pass extra arguments to my Python script using the CMake macro `SIMPLE_RUN_AND_CHECK`?

I currently copied and pasted the same script for `pbyp` and `allp`, changing the "prefix" variable only.

Paul-St-Young:
Regarding the actual bug. I think `Collectables` is zeroed out at every step, but it may not be populated if all moves are rejected in `advanceWalkers`.

markdewing:
okay to test

markdewing:
@Paul-St-Young It should be straightforward to add arguments with some changes to the SIMPLE_RUN_AND_CHECK function.  Use the ${ARGN} variable to get extra arguments to that function, and pass it the add_test function after ${check_cmd}

It think it should be enough to change the add_test call to
```
add_test(NAME "${test_name}"
    COMMAND "${check_cmd}" ${ARGN}
    WORKING_DIRECTORY "${work_dir}")
```

WIth this, any extra arguments given to SIMPLE_RUN_AND_CHECK will be passed as extra arguments to the check script

Paul-St-Young:
Thanks @markdewing! You 1-line solution worked beautifully.

Paul-St-Young:
I added documentation on the current estimator collection procedure. I think there are improvements to be made, but my understanding is that @ye-luo is already working on it. 

I will work to fix the current bug within the existing design.

manual/estimator_manager.tex
```diff
@@ -0,0 +1,298 @@
...
+\begin{align}
+E_{DMC}[O] = \frac{1}{N_{step}} \sum_{s} \left\{ \left(\sum_e w_{s,e} O(\bs{R}_{s,e})  \right) / \left( \sum \limits_{e} w_{s,e} \right)  \right\}.
+\end{align}
+
+I will refer to the average in the $\{\}$ as \textit{ensemble average} and the remaining averages \textit{block average}. The process of calculating $O(\bs{R})$ is \textit{evaluate}.
+
+\subsection{Class Relations}
+A large number of classes are involved in the estimator collection process. They often have misleading class name or method name. Document gotchas in the following list:
+\begin{enumerate}
+\item \verb|EstimatorManager| is an unused copy of \verb|EstimatorManagerBase|. \verb|EstimatorManagerBase| is the class used in the QMC drivers. (PR \#371 explains this)
```
markdewing: Can EstimatorManager.cpp and .h be deleted? (See #371).  That would address this point.

ye-luo: Please do not do the deletion of the file but just don't mention it in the tex. I will soon PR some changes including the deletion.

manual/estimator_manager.tex
```diff
@@ -0,0 +1,298 @@
...
+\item \verb|EstimatorManagerBase::Estimators| is completely different from \verb|QMCDriver::Estimators|, which is subtly different from \verb|QMCHamiltonianBase::Estimators|. The first is a list of pointers to \verb|ScalarEstimatorBase|. The second is the master estimator (one per MPI group). The third is the slave estimator that exists one per OpenMP thread.
+\item \verb|QMCHamiltonian| is NOT a parent class of \verb|QMCHamiltonianBase|. Instead, \verb|QMCHamiltonian| owns two lists of \verb|QMCHamiltonianBase| named \verb|H| and \verb|auxH|.
```
markdewing: Based on the names, I would (incorrectly) expect QMCHamiltonian to be derived from QMCHamiltonianBase.  It would be good to say what these are (or if they are defined in a previous section).  QMCHamiltonianBase is the base class for individual Hamiltonian pieces, and QMCHamiltonian is the container for all of them.  H is for Hamiltonian elements that contribute to the energy, and auxH is for elements that do not. (Someone else can comment if this description is wrong?)

manual/estimator_manager.tex
```diff
@@ -0,0 +1,298 @@
...
+\item \verb|QMCDriver::H| is NOT the same as \verb|QMCHamiltonian::H|. The first is a pointer to a \verb|QMCHamiltonian|. \verb|QMCHamiltonian::H| is a list.
```
markdewing: This could subsumed into the previous item by pointing out the QMCDriver::H is of type QMCHamiltonian (and QMCHamiltonian::H was defined in that point)

manual/estimator_manager.tex
```diff
@@ -0,0 +1,298 @@
...
+E_{DMC}[O] = \frac{1}{N_{step}} \sum_{s} \left\{ \left(\sum_e w_{s,e} O(\bs{R}_{s,e})  \right) / \left( \sum \limits_{e} w_{s,e} \right)  \right\}.
+\end{align}
+
+I will refer to the average in the $\{\}$ as \textit{ensemble average} and the remaining averages \textit{block average}. The process of calculating $O(\bs{R})$ is \textit{evaluate}.
+
+\subsection{Class Relations}
+A large number of classes are involved in the estimator collection process. They often have misleading class name or method name. Document gotchas in the following list:
+\begin{enumerate}
+\item \verb|EstimatorManager| is an unused copy of \verb|EstimatorManagerBase|. \verb|EstimatorManagerBase| is the class used in the QMC drivers. (PR \#371 explains this)
+\item \verb|EstimatorManagerBase::Estimators| is completely different from \verb|QMCDriver::Estimators|, which is subtly different from \verb|QMCHamiltonianBase::Estimators|. The first is a list of pointers to \verb|ScalarEstimatorBase|. The second is the master estimator (one per MPI group). The third is the slave estimator that exists one per OpenMP thread.
```
markdewing: It sounds like one big difference between these is related to parallelization - MPI and OpenMP? It seems like it would be helpful to describe how the parallelization structure drives some of structure of the estimator accumulation.   Also, how does this structure affect the weighted summation in E_VMC and E_DMC?

manual/estimator_manager.tex
Paul-St-Young: OK, I will take out this item from the doc

manual/estimator_manager.tex
```diff
@@ -0,0 +1,298 @@
...
+
+Data are outputted to scalar.dat in 4 stages: evaluate, load, unload, and collect. In the evaluate stage, \verb|QMCHamiltonian::Observables| is populated by a list of \verb|QMCHamiltonianBase|. In the load stage, \verb|QMCHamiltonian::Observables| is transfered to \verb|Properties| by \verb|QMCDriver|. In the unload stage, \verb|Properties| is copied to \verb|LocalEnergyEstimator::scalars|. In the collect stage, \verb|LocalEnergyEstimator::scalars| is block-averaged to \verb|EstimatorManagerBase|\\ \verb|::AverageCache| and dumped to file. For \verb|Collectables|, the evaluate and load stages are combined in a call to \verb|QMCHamiltonian::auxHevaluate|. In the unload stage, \verb|Collectables| is copied to \verb|CollectablesEstimator::scalars|. In the collect stage, \verb|CollectablesEstimator|\\ \verb|::scalars| is block-averaged to \verb|EstimatorManagerBase::AverageCache| and dumped to file.
```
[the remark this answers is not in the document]
Paul-St-Young: Agreed. I wrote this summary last, but it would serve as a good overview.

manual/estimator_manager.tex
jtkrogel: This is correct. A "HamiltonianBase" element is really a component of the Hamiltonian for ones that contribute to the energy and just an observable/operator otherwise.  The common denominator for HamiltonianBase is that each derived instance represents a quantum mechanical observable/operator.

prckent:
This PR has not been updated in some while. There are a few outstanding questions that need answering before merging and some conflicts to fix. Should we leave this PR open?

ye-luo:
I fixed the merge conflict. This PR seems not solving our issue but mostly documentation, it can be merged.

PR #755, opened 2018-04-09
Move Tmove operations to NonLocalECPotential class

ye-luo:
Since the introduce of size consistent Tmoves, the DMC drivers gets lengthy.
Indeed, it is more appropriate to put these operations in a routine as part of NonLocalECPotential.
The new routines can be shared by p-by-p and all-p drivers.

The previous method `evaluate(ParticleSet& P, std::vector<NonLocalData>& Txy)`
is replaced by `evaluateWithToperator(ParticleSet& P)` to prepare Tmove and
`makeNonLocalMoves(ParticleSet& P)` to apply Tmove.

The largely duplicated `NonLocalECPotential::evaluate(ParticleSet& P)` and `NonLocalECPotential::evaluate(ParticleSet& P, bool Tmove, std::vector<NonLocalData>& Txy)`
are merged to reduce necessary maintenance.

With #754 in place, the validation reflects that no numerical change is made in this PR.

```
                            LocalEnergy               Variance           Kinetic                LocalPotential
allp-3.2-aos-no  -10.531050 +/- 0.000184   0.376721 +/- 0.000799   11.483801 +/- 0.001966   -22.014851 +/- 0.002007
allp-3.2-soa-no  -10.531123 +/- 0.000180   0.377863 +/- 0.002306   11.480341 +/- 0.001736   -22.011464 +/- 0.001765
pbyp-3.2-aos-no  -10.531236 +/- 0.000183   0.376761 +/- 0.000422   11.486404 +/- 0.001857   -22.017640 +/- 0.001897 
pbyp-3.2-soa-no  -10.530945 +/- 0.000186   0.376305 +/- 0.000367   11.484879 +/- 0.001847   -22.015824 +/- 0.001884 
allp-new-aos-no  -10.530964 +/- 0.000183   0.376607 +/- 0.000976   11.482375 +/- 0.001907   -22.013339 +/- 0.001943 
allp-new-soa-no  -10.530932 +/- 0.000177   0.376067 +/- 0.000424   11.484422 +/- 0.001863   -22.015354 +/- 0.001902 
pbyp-new-aos-no  -10.531145 +/- 0.000177   0.376902 +/- 0.000437   11.486100 +/- 0.001918   -22.017245 +/- 0.001959 
pbyp-new-soa-no  -10.530905 +/- 0.000189   0.376260 +/- 0.000348   11.486298 +/- 0.001900   -22.017203 +/- 0.001949 

                            LocalEnergy               Variance           Kinetic                LocalPotential
[...]

PR #765, opened 2018-04-13
Fixes for all particle move VMC/DMC drivers

ye-luo:
Some left over fix for #754 and #755.
allp DMC Tmove is fixed in the driver with LogValue fix in Jastrow factors.
LogValue needs to be updated by acceptMove such that Psi.getLogPsi() returns the correct value at any time.  Unit tests are updated.
allp VMC failing in SoA is fixed by adding the missing neighbour list construction needed by NLPP. This is not exposed by bccH tests.
The allp VMC tests on diamond had too large time step. This is also fixed and they pass stably now.

### DIFF
--- a/src/Particle/ParticleSet.cpp
+++ b/src/Particle/ParticleSet.cpp
@@ -577,8 +577,11 @@ bool ParticleSet::makeMove(const Walker_t& awalker
 #if defined(ENABLE_SOA)
   RSoA.copyIn(R); 
 #endif
-  for (int i=0; i< DistTables.size(); i++)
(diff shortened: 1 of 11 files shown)


prckent:
Do these pass the long allp tests with AoS and SoA?

ye-luo:
I didn't add long allp tests last time. Since the bugs are fixed, I will add the long tests and reply the question of @prckent.

prckent:
Please add an issue describing the problem

ye-luo:
Issue added.

PR #767, opened 2018-04-13
Clean up the use of donePbyP in drivers

ye-luo:
In the SoA code, electron-ion table not only carries the distances but also a neighbour list for NLPP evaluation. In many cases, when all the distance tables are evaluated from scratch, the neighbour lists need to be updated but only possible through calling donePbyP which is supposed to be used by PbyP drivers only.

This PR changes the default behaviour that the neighbour list is updated automatically when the full table is recomputed. So in most places, the calls to donePbyP can be removed and drivers not tested for SoA may work out of the box.

The only performance penalty I can think of is in optimization when not using NLPP in the correlated sampling. The penalty should be very very small.

### DIFF
--- a/src/Particle/AsymmetricDistanceTableData.h
+++ b/src/Particle/AsymmetricDistanceTableData.h
@@ -172,7 +172,7 @@ struct AsymmetricDTD
   }
 
   ///not so useful inline but who knows
-  inline void evaluate(ParticleSet& P)
+  inline void evaluate(ParticleSet& P, bool update_neighbor_list)
   {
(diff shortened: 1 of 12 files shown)


ye-luo:
There is no state machine inside individual distance tables. It is in ParticleSet.
I updated the comments in ParticleSet.h.

prckent:
I think DistanceTable has state (otherwise it would not need a donepbyp function). The machine that manipulates this is in ParticleSet. Correct?

ye-luo:
@prckent Correct.

Issue #761, opened 2018-04-12
QMCPACK should abort when given bad planewave/bspline orbitals


jtkrogel:
QMCPACK should abort when orbitals coming from pw2qmcpack are malformed.  I observed a case recently (using QE 5.3.0) where selected up orbitals for 3 out of 19 twists had unreasonably large per orbital kinetic energies.  This in and of itself reflects a bug in either pw2qmcpack or QE itself, with the converter being the most likely suspect.

In events similar to this, QMCPACK should abort with an error, rather than running and producing bad data (VMC at least, DMC generally crashes due to the extreme variance).

These cases could be guarded against by computing the per orbital kinetic energies (just summing planewave coeffs x g-vectors^2) and checking against a maximum value.  Since both the total kinetic energy and the number of orbitals scales like N, the per orbital kinetic energy is intensive.  The maximum value I have encountered for this quantity is around 7 Ha, while these failure cases are in the range of 60-120 Ha per orbital.  A maximum value of around 15 Ha might be reasonable.

Separately, the potential bug with the converter needs looking into, but the problem runs take days to converge and I do not know if the problem is easy to reproduce.  Robustness is hard.

ye-luo:
Did your python tool also get a bad KE from the generated h5?
How about having pw2qmcpack computes the KE for each orbital and record them in hdf5? Then in qmcpack computes KE from gvec and check.

jtkrogel:
Yes, post-failure I found the bad KE's directly from the h5 file with my tool.  

I think pw2qmcpack should compute the total KE per supercell k-point and report it in the output for user reference.

Bug checking can take multiple forms based on the per orbital KE's:
1) Orbital KE's are very similar across supercell twist (typically differ by ~0.03 Ha when comparing orbitals with same index, ordered by KS eigenvalue).  Any difference in orbital KE across supercells twists/kpoints exceeding e.g. 1 Ha is a bug.
2) In pw2qmcpack and QMCPACK, an orbital KE > 15 Ha is a bug.
3) In QMCPACK total and per orbital KE could be checked against the values calculated within pw2qmcpack, but as you mention this would require an update to the h5 file format.

I would recommend all of the above with a hard look being given to the KE's computed within pw2qmcpack.  I strongly suspect there is a bug at that level since the DFT total energies looked fine.  

If desired, I can make available the directory where h5 generation with pw2qmcpack was performed (on Eos) for conversion step reruns + bug hunting.

prckent:
[...]

Issue #675, opened 2018-02-16
QMCPACK should abort in the presence of NaN's


jtkrogel:
A recent run gave NaN's at each and every VMC step for some twists.  The code did not abort until the bad info caused DMC to fail.  It would be much better if the code would abort hard at the first instance of NaN's as this would save human and computer time.

Still investigating the root cause of the NaN's; could be faulty input orbitals.
```
                            LocalEnergy               Variance           ratio 
qmc.g000  series 0  -1658.867562 +/- 0.003490   47.508980 +/- 0.167104   0.0286 
qmc.g001  series 0  -1658.869385 +/- 0.003996   47.105770 +/- 0.145762   0.0284 
qmc.g002  series 0  -1658.875468 +/- 0.003755   46.690723 +/- 0.066967   0.0281 
qmc.g003  series 0  -1658.876299 +/- 0.003627   46.500695 +/- 0.079558   0.0280 
qmc.g004  series 0  -1658.870243 +/- 0.003762   46.156389 +/- 0.070319   0.0278 
qmc.g005  series 0  -1658.874163 +/- 0.003922   46.088363 +/- 0.057244   0.0278 
qmc.g006  series 0  -1658.869806 +/- 0.003518   46.120435 +/- 0.073731   0.0278 
qmc.g007  series 0  -1658.877486 +/- 0.004448   46.100426 +/- 0.063137   0.0278 
qmc.g008  series 0  -1658.870285 +/- 0.003658   46.202121 +/- 0.057652   0.0279 
qmc.g009  series 0  -1658.867425 +/- 0.003310   46.391702 +/- 0.059393   0.0280 
qmc.g010  series 0  -1658.876643 +/- 0.003643   46.733503 +/- 0.070403   0.0282 
qmc.g011  series 0  -1658.882400 +/- 0.004074   47.200840 +/- 0.102295   0.0285 
qmc.g012  series 0  -nan +/-     -nan   -nan +/-     -nan     -nan 
qmc.g013  series 0  -1658.862983 +/- 0.004980   47.038696 +/- 0.068098   0.0284 
qmc.g014  series 0  -1658.876883 +/- 0.003361   46.777647 +/- 0.079699   0.0282 
qmc.g015  series 0  -nan +/-     -nan   -nan +/-     -nan     -nan 
qmc.g016  series 0  -1658.873288 +/- 0.003873   46.283234 +/- 0.069075   0.0279 
[...]

PR #770, opened 2018-04-16
Add orbital norm check in SplineAdoptor Readers

ye-luo:
Add the orbital norm check requested by #761.
It many be able to catch #675. 
I ran all the short tests and they are fine.


ye-luo:
Both real and complex code paths use complex plane wave coefficients.

PR #774, opened 2018-04-18
add Blue Waters configuration script

Paul-St-Young:
based on build_olcf_titan.sh

I made sure that cmake variables were set properly in CMakeCache.txt for each build and that they all compiled.


prckent:
OK to test

PR #776, opened 2018-04-23
Adjust the memory query in batched algorithm

ye-luo:
1. The memory size of psiM should be handled outside the SPO routines since psiM is passed in as an argument.
2. Add a fallback evaluateValues and a specialized one for LCAOrbitalSet.


ye-luo:
SPOSetBase.* have been cleaned up.

jtkrogel:
Do any of the evaluate functions you removed match the signature the evaluate call made on line 359 of OrbitalImages.cpp?

prckent:
(Won't merge until @jtkrogel 's query answered).

prckent:
I believe that the CI would not pass because the main app would not build

ye-luo:
@jtkrogel Those evaluate functions for a single particle are untouched. The removed ones are for a range of [First, last) particles. That change was rigorously checked by compilers.

jtkrogel:
OK, wasn't sure if it was compile checked or if that part of the code was walled off from the build (didn't investigate this).  No problem with proceeding.

PR #777, opened 2018-04-23
Change the GPU Sherman-Morrison to the Fahy variant

ye-luo:
All changes are in the CUDA part.
1, Remove the slow "compensated dot" algorithm.
2, Rewrite the step 1 GEMV of the update algorithm.
3, In all_ratios_grad_lapl_kernel, replace the duplicated lines with a loop.

Checked with NiO 128 atom simulation on full Titan and obtained unchanged results.
5~10% overall speed-up.
The lines of source code are significantly reduced.


prckent:
Thank you.

* Have you checked smaller or larger system sizes? (only 1 GPU needed for a few steps, not a full machine run).
* Did you consider fusing any kernels to reduce launch overhead on smaller problem sizes? (perhaps there are sufficient registers or memory on today's GPUs)
* Did you check Volta? (I guess no due to Summit non-availability)

ye-luo:
1. Accuracy on large system. 128 atom cell is the largest one I tried on GPU. On CPU, I checked up to 512 atom with single precision. When you have single precision, it is very hard to check accuracy on a single walker because error quickly grows up after a few single particle moves but the statistically averaged quantity like energy is not sensitive to this error.
2. I only tried Kepler and Pascal. I don't believe it can go wrong on Volta.

prckent:
I am not worried about accuracy, only speed. Is your code always faster or is there a cross-over?
Also, does this normalize the algorithm used in the CPU and GPU implementations?

ye-luo:
It is always faster because the compensated dot is awful and the rewritten GEMV removes performance killing branch divergence.
Now the CPU and GPU implementation has the same computation. The only difference left is GPU saves the new row of A but CPU doesn't.

prckent:
Actual speedup on Kepler GPUs:
image

ye-luo:
Diluted by the heavy recompute. Usually I use 10 steps per block but performance tests do only 5.

prckent:
Agreed. A result nevertheless.

PR #779, opened 2018-04-23
Hybrid orbital representation manual + final tweak

### LINKED ISSUE #745: Manual documentation needed for hybrid representation
The hybrid representation--which gives a lower variance and lower memory orbital representation for planewave/B-spline periodic systems--appears in the tests and is mentioned in the manual.  However input documentation and documentation regarding its general use is missing from the manual.  

Adding this documentation should be prioritized because the hybrid representation should become a standardly used (or at least frequently used) feature in the near future.

ye-luo:
1, The manual section is added.
2, Following the suggestion of @jtkrogel, parameters for splining the radial function are auto determined.
3, The batched algorithm for hybrid now checks, in each evaluation, if it is safe to operate based on the distance. If not, the non-batched route is taken.
4, the batched algorithm tests based on diamondC are removed because they are not hitting the intended code path. Tests based on NiO are added. The batched algorithm reuses the scratch memory from the determinants. Too small system doesn't have sufficient size and takes the usual code-path.

closes #745


ye-luo:
1. CI is fixed. Some intermediate commit should fail SoA but the CI shows pass. Need @grahamlopez to change MP to SoA-MP.
2. get_first_neighbor documentation is updated in the base class.
3. is_batched_safe documentation is added in the source.
4. Add comments in NiO_a4_e48_pp about tested code-path. The batched tag is also mentioned in the manual.

grahamlopez:
@ye-luo I have set `-DENABLE_SOA=1` for the CPU CI runs on Rhea

ye-luo:
@grahamlopez thanks.

PR #780, opened 2018-04-23
Route CUDA_ARCH for ctest

prckent:


prckent:
Agreed. Long and not maintainable.

PR #786, opened 2018-04-25
Add a dummy return to reduce Clang warning.

ye-luo:


ye-luo:
Test this please

prckent:
Just for the record, why aren't we using pure virtual functions here?

ye-luo:
This routine is implemented only in SoA distance tables. No plan to add for the old AoS ones unless this is really needed. If the AoS ones are accessed, the code should stop to catch developer attention.

prckent:
Separate PRs for separate updates please! I don't want us to get in the habit of mixing updates. Next time this happens from core developers I will close the PR. Branches and PRs are cheap.

PR #791, opened 2018-04-26
nexus: fix qmca prefix list when performing twist averaging

jtkrogel:
Fix a small bug introduced in #600 relating to the -a option.


markdewing:
okay to test

PR #793, opened 2018-04-26
nexus: fix default trace plot range in qmca

jtkrogel:
Trace plots were often too zoomed out by default in qmca.  A visual range of +/- 2 sigma is restored as the default.


prckent:
OK to test

PR #796, opened 2018-04-26
nexus: add ability to join subsequent series as a single data source in qmca

jtkrogel:
Add new command line option to qmca (-j).  Used to join/concatenate multiple series as if they appeared as a single file originally.  Equilibrated data are removed prior to join.

As a preparatory step, also make twist averaged calculation more precise.  Specifically, twist average the raw file data first, then perform analysis on the twist averaged data.


markdewing:
okay to test

Issue #789, opened 2018-04-25
Strange behavior of the DMC trial energy


jtkrogel:
During a production Mira run with the most recent SoA version of the code I noticed large spikes/re-equilibration when identical DMC sections were run in series (no change in timestep).  This behavior appears to relate to how the trial energy is reset and it undesirable as it leads to unnecessary equilibration and waste of data.  I don't recall observing similar behavior in earlier versions of the AoS code (e.g. v3.1.1).

The last four sections of DMC data shown in the figures below use the same timestep (0.005/Ha) and each includes 10 warmup steps.  

Behavior of the local energy.  Spikes are visible at the onset of each DMC section.  This is not visible for individual twists, but it is for the twist average.
![local_energy](https://user-images.githubusercontent.com/24547905/39269716-96252212-48a1-11e8-9b56-d25e5497bfc5.png)

Behavior of the trial energy.  The trial energy is apparently reset in a way that lacks continuity between the identical DMC sections.
![trial_energy](https://user-images.githubusercontent.com/24547905/39269721-9c93237e-48a1-11e8-8b8e-995c24b12741.png)

jtkrogel:
Additionally, the trial energy above appears to be undershooting the local energy in the first visible DMC section (timestep=0.02/Ha) resulting in beyond target growth of the walker population:
![num_walkers](https://user-images.githubusercontent.com/24547905/39270231-6dcb21ac-48a3-11e8-820f-594128821dcb.png)

prckent:
Agreed that this is a bug. Is this the development version or the latest release?

ye-luo:
Try a run with samples/walkers per thread <= 1

jtkrogel:
What will this show?  The runs are not totally cheap, ~1.5 hrs on 8192 Mira nodes.

ye-luo:
It probably will show no bumps and confirms that you are affected by the >1 bug.

jtkrogel:
Very well.  If there are no objections I will submit a new job tomorrow.

jtkrogel:
I do not set reconfiguration either way.

ye-luo:
Here in the first section, it takes about 100-150 steps to converge and the later sections needs 50 steps. You need to set your warmupSteps close to that. The trial energy is determined as the current localenergy population average during warmup and the history average of the population averaged value after the warmup. So the trial energy of your first section is very much biased due to a small warmup.

[...]

PR #797, opened 2018-04-27
Remove walker randomization and resolve low walker count restart issue

ye-luo:
To solve the issue reported in #789. All the change relevant to the issue are in QMCUpdateBase.cpp.

When using pbyp the code will randomize the walker when initializing or reusing a driver.
This destroys the already equilibrated population if it is called. For this reason, the randomization is removed and also we have one less hidden VMC driver. The driver unit test was updated.

Note that the initialization of electrons randomly is unchanged but all the walkers will have identical initial electron position. The removed randomization was making all the walkers different by performing a small VMC. This responsibility is moved to the real VMC driver now.
Basically users need to add one step more in the VMC warmup, but who cares.

I removed the updateWalkers method in QMCUpdateBase because it is not needed and the old implementation is no more safe to use.

In addition, I checked the restart and hit a bug related to using memory pool which occurs when you restart DMC with few walkers than the number of threads. this is fixed also in QMCUpdateBase.cpp.

Reference run on 1024 nodes using only one section:
one-section

The other validations have 8 DMC sections. The first 4 have warmup 50 and the other 4 have warmup 2. The 8 sections are ran with 6 in the first run and 2 in a second restart run.

The buggy code always has an energy jump in the beginning of each block.
32 walkers and 32 threads per node, it failed to restart and only had 6 sections
ref-t32-w32
32 wakers and 16 threads per node
ref-t16-w32

Fixed code,
32 walkers and 32 threads per node
fix-t32-w32
32 wakers and 16 threads per node
fix-t16-w32


prckent:
@markdewing Are you OK with the test update?
@jtkrogel Would you argue that we test this independently or agree to the merge?

src/QMCDrivers/QMCUpdateBase.cpp
```diff
@@ -229,95 +238,25 @@ void QMCUpdateBase::initWalkersForPbyP(WalkerIter_t it, WalkerIter_t it_end)
     Psi.copyFromBuffer(W,awalker.DataSet);
     Psi.evaluateLog(W);
     RealType logpsi=Psi.updateBuffer(W,awalker.DataSet,false);
-    awalker.G=W.G;
-    awalker.L=W.L;
-    randomize(awalker);
+    W.saveWalker(awalker);
+    RealType eloc=H.evaluate(W);
+#if (__cplusplus >= 201103L)
```
markdewing: We shouldn't need the C++ version test anymore.

markdewing:
@prckent I am okay with the test change.

jtkrogel:
I think we can merge based on the initial evidence.  Alternatively, I have some runs using the updated code that may finish over the weekend and we will know for sure once they complete.

jtkrogel:
We do not currently have tests that check chained DMC runs do we?  I suppose a current test could/should be extended to have an additional DMC with no warmup.

ye-luo:
@jtkrogel Plan to add a restart test with

ye-luo:
Sorry my mistake clicking the wrong thing.

prckent:
It would help to formalize our thoughts on the different restart scenarios we support and what should happen in each case. Then add tests for each case checking e.g. that the trial energy is set "correctly" on restart.

ye-luo:
I opened #798 to remind me of adding DMC restart tests.

Issue #792, opened 2018-04-26
Tests needed for qmca


jtkrogel:
Since qmca is a user-facing tool, it is desirable to have tests in place to verify its correct operation under known use cases and avoid bugs such as #791.

markdewing:
Two kinds of tests:
1. Internal (unit tests).  This could be provided by the test framework for other parts of nexus in #760
2. External (system tests). Use qmca as from the command line and check the output
     These could go in nexus/tests/executables ?

     For these tests, it might also be useful to have a csv output option to  make parsing qmca output easier

jtkrogel:
Addressed by #2091.

PR #800, opened 2018-04-27
[WIP] Initial version of qmca tests

markdewing:
This is a start to addressing #792 
It might need modification based on the structure of #760 

This only tests the form of the results, looking for regular expressions.    Still needed is some way to test the values that qmca produces.


jtkrogel:
I'm not sure cmake is the right language to use to parse the outputs and perform the checks.  It might be easier to e.g. tokenize the full output from qmca, detect floats, trim the last digit from the floats, and then do a character for character test for equality vs reference outputs.  

Additionally, these are Nexus tests and should go in ntest directly.

markdewing:
Yes, cmake is not the right language to perform more detailed checks on the values of the output.   A python script would be better - following something like the converter tests.

ntest seems like a unit-test framework, but I suppose it can call executables and check the output, too.

jtkrogel:
Right now it just runs what look like unit tests.  Down the road it will drive nexus scripts and check file contents, etc.

PR #806, opened 2018-04-30
Remove AtomicHF

jtkrogel:
Apparently an earlier (non-functional?) version of the functional SQD code.  As far as I know it is unused and the last significant development was in 2004.


grahamlopez:
okay to test

prckent:
Good to remove legacy "duplicates".

Perhaps the SQD code could be named AtomicHF, since that is the most likely use?  This assumes SQD is properly active ?

Issue #807, opened 2018-04-30
SQD build broken by #527


jtkrogel:
Build of SQD was broken with the removal of OhmmsInfo.* in #527.  SQD needs to be updated to use the new reporting system.

jtkrogel:
#808 Fixes build issues other than the reporting system.

jtkrogel:
"Resolved" by #833

PR #808, opened 2018-04-30
Cleanup SQD

jtkrogel:
Remove unsupported files from SQD.  Update to use of std::pow.  Fix CMakeLists.txt header.

SQD was broken by #527.  See also #807.

More cleanup remains for SQD, e.g. there appears to be an ancient QT user interface present.


grahamlopez:
okay to test

prckent:
If we care enough to keep SQD around, we should attempt to add SQD to the default build.

PR #810, opened 2018-04-30
Remove WalkOnSpheres

jtkrogel:
Remove Walk On Spheres (WOS).  First production level publication with QMCPACK (OHMMS-QMC?) by D. Das, R. M. Martin, and M. H. Kalos.  See also

https://doi.org/10.1103/PhysRevE.73.046702
http://mcc.illinois.edu/publications/theses/Das_Dyutiman.pdf
http://mcc.illinois.edu/research/posters/2005/New_MC_Devices_Martin.pdf

Interface to QMCPACK, if there ever was one, has been lost to time.  The last significant development for WOS was in 2004.


grahamlopez:
okay to test

prckent:
Far enough from current uses that I agree we can remove it.

PR #811, opened 2018-04-30
Nexus: add extra protection around file removal

jtkrogel:
ntest generated a failure on oxygen due to missing directory during removal.  Update the file removal code to only care that the directory is gone following the attempt to remove it.


grahamlopez:
okay to test

PR #812, opened 2018-04-30
Remove fully disconnected or never used Hamiltonian elements

jtkrogel:
Most of these are not built or used in any way.  VHXC was built, but as far as I know it is incomplete and has not seen any use since 2009.

More waves of Hamiltonian cleaning to follow.


prckent:
OK to test

PR #814, opened 2018-04-30
Remove model insulator

jtkrogel:
Remove old model insulator files.  Last known development dates to 2006, no use cases are known, and no input/output files exist.


prckent:
OK to test

PR #815, opened 2018-04-30
Remove unused Hamiltonian elements

jtkrogel:
Remove misc unused Hamiltonian elements.

NumericalRadialPotential appears to be a speculative addition from 2007.  

NearestNeighborsEstimator was added by me and was of short-lived interest for streaming indices of electrons nearest to atoms through TraceManager.  

NumberFluctuations may have been used during the Renyi entropy work, but no input exists and the driver is long gone.


prckent:
OK to test

PR #817, opened 2018-05-01
Remove unused reptation estimators

jtkrogel:
These files are not included anywhere.  They appear to originate with some of the helium related work in 2008 but were never expanded beyond the initial commit.


markdewing:
okay to test

PR #818, opened 2018-05-01
Remove legacy files and headers that are never included

jtkrogel:
These files are either legacy einspline or header files that are not included by any other file in the distribution.


grahamlopez:
okay to test

markdewing:
I'm a little concerned about deleting some unused files in utility directories (OhmmsPETE, Utilities, Numerics).   Particularly ones that form a set of functionality.  
OhmmsPETE TinyMatrix - this is apparently unused because Tensor contains all the same functionality and is used instead (?)
 STLEval and STLVectorOperators - looks like it integrates STL containers into expression templates

Utilities FortranString - might be useful if we interface to QE or other codes at a library level (?)

Despite these concerns, I still agree with deleting all these files.   Partly I'm mentioning the files here in case someone comes looking for them in the future.  Though it's probably easier to rewrite them from scratch.

prckent:
Thanks all for looking through these. I'll put a note for future archaeologists in the 3.5.0 release notes. If ever needed we should plan to check back, but the unknown status and lack of tests might make a rewrite of any utility features the best route.

PR #820, opened 2018-05-01
Remove header/cpp file pairs that are totally unused

jtkrogel:
Some header files are included only in a cpp file sharing the same name and nowhere else.  Remove these header/cpp pairs.

Also remove other dangling header files that are the result of an iterative application of the search applied in #818 (i.e. chains of headers that were never included somewhere in the end).


markdewing:
okay to test

markdewing:
Does anyone know why the warp methods are not used?

More general rant: algorithms and methods get published, but sometimes they fall out of favor and the reasons why are usually only passed around in person and never published or documented.  It would be nice if there were more records about 'we tried X and it didn't work, for reasons Y and Z'.

prckent:
Is this the space warp transform used e.g. by Sorella and co to compute forces or something else? e.g. better sampling. Are there any QMCPACK papers using this feature?

jtkrogel:
I don't know why they were abandoned.  That is really a question for @jngkim 

As far as I know, these fell out of use around 2008.  I don't know if they were ever used in production.

prckent:
I am happy for this trimming to go ahead regardless - these are very old files with no current champions or imminent use identified.

PR #821, opened 2018-05-01
[WIP] Change test comparison with reference value.

markdewing:
Use the combination of the error bar on the computed value and the reference value for the pass/fail comparison.  This change allows exact reference values (zero error), like analytic results or those coming from Gamess or Quantum Espresso.  This change should also lead to fewer failures due to random fluctuations in the nightly tests

This would change the process in
There would be no need to estimate an appropriate error bar for the reference value for short and long tests.

See #17 for discussion.

One issue is that a very noisy value can pass this test by having very large error bars.  The fix is to compare the variance in addition to the energy.

Add the variance to a couple of tests.

For the case of the SHO VMC results, the exact value is known, and the local energy is also a constant.   The averaging process introduces some numerical roundoff error.  To account for this, increase the zero error reference by a little bit.


jtkrogel:
The reason I used a longer test to estimate the error bar is because the error bar is itself a statistical quantity and fluctuates about its own average value for many runs.

Basing pass fail on the current (statistical) estimate of the error bar generated within each test should increase the incidence of both false positives and false negatives (where the mean actually varies according to the estimate from the longer run, but the current error bar is underestimated, leading to false failure).

I think the original process is the one that should be used, but it is up for discussion.

markdewing:
I should pull out the SHO changes and submit a separate PR.  That is independent from changing how the error bars are computed.

https://cdash.qmcpack.org/CDash/testDetails.php?test=2023983&build=21692
This failed test would pass with the new criteria. 
I think there would be fewer failures close to the edge of significance.

markdewing:
Comparing the variance should be equivalent to comparing the error bar, correct?
If the change in comparison criteria is made, then adding tests for the variance of every quantity should cover testing the error bars.  (That is one cost of making this change - need to add the variance to all the tested quantities)
Comparing the variance (vs. the error bar) appeals to me because it is independent of the length of the run.

jtkrogel:
It is almost equivalent, with the exception of the autocorrelation time, which, when correctly accounted for, always increases the estimated fluctuations of the mean (error bar).  Autocorrelation estimates become poorer the shorter the run and generally the computed error bars are over-optimistically too small.  

Checking the variance is worthwhile on its own because it is more sensitive to some changes (e.g. a reduction in the quality of orbital representation) than the mean, but it misses the increase in fluctuations due to autocorrelation.

Issue #653, opened 2018-01-26
What are higher wavefunction derivatives used for?


markdewing:
There are higher spatial derivatives of the wavefunction, such as the Hessian in 
evaluateHessian (defined in OrbitalBase.h) and evaluateWithHessian (defined in BasisSetBase.h)

The Hessian appears to be used for the stress tensor in QMCHamiltonians/StressKinetic.cpp.  Is it used anywhere else? And, related, is the stress tensor an important observable?

Then BasisSetBase.h also defines evaluateWithThirdDeriv and evaluateWithThirdDerivOnly, which appear to be used in various overloads of evaluateForWalkerMove in SphericalBasisSet.h.   It appears that these overloads are not used (searching for the GGGVector_t type or grad_grad_grad_Phi doesn't seem to turn up anywhere outside QMCWaveFunctions).
Are these third derivatives used?  What is their intended purpose?

lshulen:
Not sure about the answers to many of these (for example the third deriv stuff), but I consider the stress tensor to potentially be an important observable.  I think it is not used much now for many reasons, but I routinely use it for many classes of calculation in DFT.

prckent:
Stress tensor is important.

If no-one comments on third derivatives lets get rid of them. Anything we can do to simplify OrbitalBase is helpful. We need to evolve OrbitalBase so that only actually used functionality is referenced (!), and that we identify the minimal set of functionality needed to implement (say) a new Jastrow.

Paul-St-Young:
@rcclay do you know what the higher derivatives are used for?

rcclay:
Backflow optimization?

jngkim:
Yes.

Paul-St-Young:
Oh yes. `evaluateWithThirdDeriv` is used in `LCOrbitalSetWithCorrection.h::evaluate_notranspose` for `grad_grad_grad_logdet`. This is in the molecular code path for backflow optimization.

prckent:
Thanks. Some comments about exactly which derivative is being taken and the main use case in orbitalbase would help.

mmorale3:
Optimizing parameters in the backflow transformation requires the mixed second order derivatives, not just the diagonal terms. The formulas are in the BF classes and in David’s papers.

jtkrogel:
Question answered.

Issue #420, opened 2017-10-17
Implement VGH and higher order functions for new splines


rcclay:
Issue highlighted in PR #373 , backflow optimization and stress estimators require VGH and higher order derivatives, which are currently not implemented in the new spline framework.

Issue #200, opened 2017-05-03
Status of backflow code is uncertain


prckent:
There are no tests or reference inputs for the backflow code. This was working a few years ago, but might not be working the production version, particularly with optimization.

First step is to find some inputs and assess functionality: does VMC work? optimization? are old runs reproducible?

Paul-St-Young:
VMC with backflow works in the current version of the code. However, Backflow optimization is broken. Optimization errors out with:
Fatal Error. Aborting at Need specialization of SPOSetBase::evaluate_notranspose() for grad_grad_grad_logdet.

Backtrace shows that the `evaluate_notranspose()` call originated from the `EsinplineAdoptor` class. The `EinsplineSetExtended` class has the needed specialization of `evaluate_notranspose()`. @ye-luo pointed out to me that the switch from `EinspleinSetExtended` class happened at svn revision 6465 or git commit ef857a4a96ac25202bf941aeb87a1a4763cca98e on May 25 2015.

I was able to verify that backflow optimization works in svn revision 6296. Using the optimized backflow, I ran VMC with both rev6296 and the current github develop branch and got the same total energies and variances. I used bcc hydrogen with 54 atoms as a test system, which may be a bit big for a regular test.

The current way the `BackflowTransform` class handles the optimizable parameters is rather sketchy. Instead of using `opt_variables_type myVars;` to map from local to global index, it uses a locally constructed map `optIndexMap`.

In order to restore a reliable backflow implementation, there are at least two things to do:

1. Implement `evaluate_notranspose()` for grad_grad_grad_logdet in `EinspleinAdoptor` class.
2. Revamped the optimization parameter mapping in `BackflowTransform` class.

prckent:
Thanks. Good to hear that VMC reproduces old runs. Almost certainly then, it is only the area of optimization that needs revision.

The assumptions of the old optimization parameter mapping certainly need fixing or verifying. I believe (mentioned by @mmorale3 ) it assumes the backflow parameters  are first.

ye-luo:
Settled by #1204 #1225

PR #823, opened 2018-05-02
Complete the VGH routines of the Spline SPO SoA implementation

ye-luo:
This PR completes the VGH routines in
QMCWaveFunctions/BsplineFactory/Spline{C2C,C2R,R2R}Adoptor.h
The goal is to make it work correctly such that I can remove the AoS implementation.
This SoA spline implementation is always used regardless of AoS or SoA build.
Related to  #200 #420 #653

ToDo:
- [x] add VGH routine check in the unit test. @markdewing I didn't find any check on spline SPO val/grad/lap/hess. Do we have any?
- [x] Remove QMCWaveFunctions/Spline{C2X,R2R}Adoptor.h (AoS) which all fully reimplemented as SoA.
- [x] Since SoA implementation of spline has been used for more than half a year. Infrastructure to AoS can be cleaned up.
- [x] Check backflow calculation with SoA spline in AoS build. Update error handling. use_old_spline is still needed for the optimization. Routines for third derivatives are still not implemented.
- [x] check bccH_2x2x2 and bccH_3x3x3 vmc runs.


markdewing:
There are V/G/H spline tests in spline2/tests.   There are no tests at the Spline*Adoptor level

ye-luo:
I think it is necessary to check Spline*Adoptor level and make sure that phases are applied correctly.

ye-luo:
This is ready for review and merging.

prckent:
Q. How do the bands end up in a different order in the CUDA code?

jtkrogel:
I assume this will have no impact on e.g. particle-hole creation bookkeeping in terms of einspline*.dat and xml input.

ye-luo:
The particle-hole creation is sorted out before this.

prckent:
Not important for this PR, but how hard would it be to get rid of this CPU vs GPU divergence?

ye-luo:
The phase calculation CUDA kernel needs to adopt a consistent ordering.

ye-luo:
Can we merge this?

PR #825, opened 2018-05-02
Remove unused files outside the build system

jtkrogel:
Files outside the build system were identified by parsing CMakeLists.txt.  These are unused and can be removed.


markdewing:
okay to test

prckent:
I don't believe that there are any here, but when removing files please be careful not to remove any potentially useful test codes on core functionality. There are a few that could be the basis of unit tests. Assess on a case by case basis - they still need to be salvageable.

jtkrogel:
Good point.  When performing automated searches for dead code I have manually excluded test code when I could identify it.

PR #830, opened 2018-05-02
SHO test fixes

markdewing:
Fix the failing SHO tests.  
- Add the estimated error on the potential/kinetic values. 
- Add the 'latdev' estimator to check_scalars.py and the CMake function that calls it.
- Add small epsilon to error bars that are exactly zero to account for numerical noise in the averaging.


ye-luo:
If the reference is constant, the error seems large to me. Could you explain a bit more where is the averaging noice from?

markdewing:
Add a set of N floating point values, divide by N, the value is not exact.   The bound is certainly tighter than N*epsilon, but I'm not sure how to set it.  Maybe sqrt(N) would be better?
```python
>>> def f(n):
...   return sum([0.45]*n)/n - 0.45
>>> print f(256), f(256)/sys.float_info.epsilon
1.88737914186e-15 8.5
```

jtkrogel:
For this case, log-log fit gives a linear relationship.  For the test code you gave above, I find that f(N)/sys.float_info.epsilon scales like 0.035*N.

ye-luo:
After run the tests manually, I understand that the tests have two issue. First the kinetic and local potential energy are not constants but error bar was set zero. Second is the error on the total energy which is supposed to be a constant. I saw the error added for the first part and though they are for the round error.

I think linear scaling N*epsilon is fine assuming you value is a constant but off by epsilon. Just be sure mixed precision is fine.

Please cleanup the qmc-ref, there are more files than needed.

prckent:
@jtkrogel OK to merge?

jtkrogel:
It is fine with me.

Paul-St-Young:
Sorry I am late to the party. Question: why is this test failing now when it was passing before?

Paul-St-Young:
The references should be exact with error=0. I thought the test script considers the combined error bar of reference and run. Is this no longer the case?

ye-luo:
1) Your total energy = Kinetic + potential is zero but you are also checking both of them separately with zero error.
2) The check script was not able to handle absolute zero well.

Paul-St-Young:
but this test was passing before. Did something change in the check script?

ye-luo:
I think the total energy check may pass on certain machines. You were luck. but neither Kinetic nor potential.

Paul-St-Young:
Yes, I understand the VMC error bars on kinetic and potential are not zero. This is exactly why I expect these tests to pass. 

The wavefunction I provided is the exact ground-state wavefunction. The references I provided are analytic results so error=0. 

DMC will have mixed estimator error on the kinetic and potential, but VMC should get everything correct within the statistical error bars.

Paul-St-Young:
All the `short-sho` test were passing as of the end of January (e.g. 1d388b6bcad3944db1dc25fcbd4bf7764babd48d)

markdewing:
I think the problem was there was a mistake in the models/CMakeListst.txt file (wrong name of the scalars), and that allowed the tests to silently pass (This silent failure mode has been changed).    The name was fixed at the end of March, and the test comparisons started working, and then showed the failures.

PR #821 attempts to change the scalar check error bar handling to work as you expect.  However, we need a  better handle on the background spurious error rate, which means fixing the tests to work the way the current system expects.

Paul-St-Young:
@markdewing thanks for the explanation. I totally missed the scalar name fix for the ctests...
This explains why the test was "passing" before.

Given the current test-reference error bar handling. I would still expect the kinetic and potential tests to pass with the appropriate Gaussian statistics given the VMC error. I think the reference errors should be kept at 0. Is this correct?

markdewing:
Currently, the reference value is the expected outcome of a QMC run for a given number of blocks/steps/threads/processors.    That includes the mean and the expected error bar.
In practice, this means the reference values is composed of the analytic value (for the mean), and the expected error is taken from a  long QMC run and projected back to the actual size of the short run (using a longer run to smooth out fluctuations in the error bar)
More step-by-step description on the wiki: https://github.com/QMCPACK/qmcpack/wiki/How-to-add-tests

Paul-St-Young:
I see. Given this goal: `the reference value is the expected outcome of a QMC run for a given number of blocks/steps/threads/processors`, then the reference error bar should be set to `sqrt( [variance]/[effective number of samples] ) +/- [machine precision]`.

Thanks for the discussion!

Issue #561, opened 2017-12-12
Purpose and future of SQD and AtomicHF?


markdewing:
What is purpose of the code in the SQD and AtomicHF directories?
Are we going to keep these codes?  If so, we should build and test them regularly.

jngkim:
You can get rid of them.

jtkrogel:
These are numerical Hartree-Fock codes for spherical potentials (e.g. atoms or spherical quantum dots) that can produce orbitals for QMCPACK.  I was going to add a test for this code path soon, but perhaps more discussion is warranted.

jtkrogel:
Votes for keep or remove?  

I am in between myself.  On the plus side this is actually a feature from early in QMCPACK's history that still works.  On the minus side, I don't see it as being terribly useful except in a tutorial setting unless one wants to check the veracity of HF calculations of atoms or unless spherical quantum dots become of interest.

Related to #807.  Suggest we make a decision one way or another and proceed.

jtkrogel:
Having thought about it more, I'm switching to a down vote.

jtkrogel:
"Resolved" by PRs #833 and #806.

PR #833, opened 2018-05-03
Remove SQD

jtkrogel:
SQD provides limited benefit with ensured maintenance overhead.  It was last used peripherally in 2013 (by me).  

See also #561 and #807.  Both issues can be closed with this PR.


markdewing:
okay to test

ye-luo:
Does this clean up the corresponding part inside QMCPACK?

jtkrogel:
No.  I do not know all of the runtime code paths that are touched by a run using a HF atom.  

In the future we can get these by removing the one test case (C_NO) and search for dead code paths relative to all tests.  I don't know of a better way to comprehensively remove a feature like this.

prckent:
Related functionality question:

If someone wanted to run atoms using radial numerical grids, e.g. as emitted by Froese Fischer's suite of MCHF codes or as emitted by many pseudopotential atomic codes, what would they do?

jtkrogel:
They could generalize the existing NO code (which I leave untouched here) to handle multideterminants.  Also, unless I am mistaken, the current Gaussian code using quintic radial splines, so these could be repurposed to handle more than just mappings from radial gaussians.

ye-luo:
Current GTO code allows transforming radial functions to quintic splines on a log grid. It is flexible to handle input radial functions on any grids.

prckent:
OK. Potentially someone might want this to test dirac HF, spin orbit etc., or study atoms in fields etc. Since it is not tied in with the SQD code I am fine for it to go.

anbenali:
I am planning to resurrect NO on grid soon.

prckent:
Also useful for work with DFT functional developers

PR #834, opened 2018-05-03
Clean up QMCTools

jtkrogel:
Most of these are very old (2005-2007) and I believe all are unused.  A more recent one, trace_density was used once by me to postprocess trace data to create resulting density files.  I can assure anyone this route is not worth repeating.


markdewing:
okay to test

markdewing:
Does this superseded #833 ?  This PR also contains SQD removal.

jtkrogel:
It does not need to supersede #833, but it can.  I discovered that many files containing main() draw from similar headers (e.g. from Numerics).  

Removing QMCTools sequentially (in time) after SQD removal guarantees that more orphan files are discovered (e.g. ones that where included in both places).

PR #835, opened 2018-05-03
Remove unused model potentials

jtkrogel:
Remove model systems that have not been used for ~a decade and have no known input/output files required for resurrection/maintenance.


ye-luo:
okay to test

PR #838, opened 2018-05-03
Cleanup released node and AFM determinants from Fermion

jtkrogel:
Remove wavefunction remnants for released node (which has been thoroughly broken for some time) and also the AFM determinant (specialized one time project --AFM hydrogen-- with no existing inputs/outputs).  Also a couple small cleanup items in the Fermion directory.


prckent:
Sad to lose released node, but it was never really there. Selected CI brings another way of checking convergence.

prckent:
OK to test

PR #839, opened 2018-05-03
Remove unused Jastrows

jtkrogel:
Remove old/unused model Jastrows and three-body Geminal and B-spline Jastrows.  No currently operable input exists for these.  Additionally the three-body B-spline Jastrow is incomplete and may never have been used.

Please check the removal of JAABuilder and the changes to JABBuilder carefully.


markdewing:
okay to test

PR #840, opened 2018-05-04
Remove MSDgenerator

jtkrogel:
Remove unused executable MSDgenerator.  Verified with both @lshulen and @mmorale3 that it is no longer used.

Also remove MolecularOrbitalBasis which was missed in earlier cleanup of QMCTools.


prckent:
OK to test

PR #843, opened 2018-05-04
Fix samples check in check_scalars

markdewing:
All code paths need to have the number of samples added to the stats data


markdewing:
I didn't run enough of the short tests before submitting the PR.

PR #845, opened 2018-05-10
Clean up SPOSet Builder creation logic and naming

ye-luo:
Largely renaming files, classes and functions to make them matching the actual role.
In a lot of places, BasisSet actually refers to SPOSet or SPOSetBuilder.
Actual BasisSet creation is now separated from SPOSetBuilder creation and fixes a bug that the code complains about missing basisBuilder(SPOSetBuilder).

### DIFF
--- a/src/QMCHamiltonians/DensityMatrices1B.cpp
+++ b/src/QMCHamiltonians/DensityMatrices1B.cpp
@@ -18,7 +18,7 @@
 #include <Numerics/MatrixOperators.h>
 #include <Utilities/IteratorUtility.h>
 #include <Utilities/string_utils.h>
-#include <QMCWaveFunctions/BasisSetFactory.h>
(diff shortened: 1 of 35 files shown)


prckent:
OK to test

jtkrogel:
Good to separate some of these.

Do short tests pass with these changes?  If so, we can proceed with the merge.

ye-luo:
I ran all the short tests and didn't see any failure increase.

jtkrogel:
Testing system needs thorough checking and cleanup, but that is separate.  Merge can proceed.

ye-luo:
hold a second. Let me change SPOSetBuilderBase.h to SPOSetBuilder.h. Just to have the filename consistent to the class name.

ye-luo:
@prckent Yes. It is good to go as CI passes.
My logic is XXXBuilder builds XXX with its data initialized.
XXXFactory creates XXX only the specialized class for polymorphism.

Issue #125, opened 2017-02-22
What is the preferred way to handle errors and terminate the application?


ye-luo:
To collect the ideas here about the preferred way to handle errors and termination of the app.\
We have several ways to print warning or error(with a stop) and the role and behaviour is not obvious.
Some of them are related to MPI and some are not.

abort(): from C/C++
APP_ABORT("XXXX"): macro to MPI abort, but difficult for unit tests.
exit(1): from C/C++
std::runtime_error("XXXX"): C++
app_error(): stream error output but only rank 0. No termination.

MISC ideas
1, app_error() handling. What should be sent through.
2, error messages from external libraries. Do we make it cleaner or keep them original.
3, all nodes or master only depends on error types.
4, how to integrate errors with external tools, workflow, job scheduler.

markdewing:
Unit testing also needs to be considered - calling abort or APP_ABORT right at the point of fatal errors makes it difficult to test error paths.

prckent:
My main user focused concern is that with typical errors such as missing files or  inconsistent inputs we return a useful message ("Couldn't write new the optimized wave function opt.xml") ahead of any library errors. Otherwise the useful message tends not to be seen and we take the blame. Conceivably this involves both HDF5 and XML handling.

prckent:
What does app_abort() do currently? Does it behave consistently across different systems?

ye-luo:
Paul's question raises the urgency of a preliminary dev guide. APP_ABORT looks like a macro wrapping something up. abort() is also used in the code to stop the code.

jngkim:
Yes, it is just to call mpi finalize.

Never meant to be an error handling method.

ye-luo:
#define APP_ABORT(msg) \
{std::cerr << "Fatal Error. Aborting at " << msg << std::endl; OOMPI_COMM_WORLD.Abort();}

prckent:
@ye-luo Agree with deciding on the preferred way(s) and documenting them. We should simply build up a developer chapter.

jtkrogel:
I think an abort should attempt a clean shutdown of the entire application, if possible (e.g. no broken h5 files).  

I don't have a strong opinion on how this should be handled (function call, macro, exceptions, etc).

prckent:
[...]

PR #847, opened 2018-05-10
Fix the traditional multi determinant implementation

ye-luo:
The slow implementation is broken by the changes over the years.
This PR introduces tests and fixed the code.
Also myG/L and myG/L_temp are removed from DiracDeterminantBase because DiracDeterminantBase only operates on one spin while G/L are over all the electrons. They are in the wrong place.
Change error handling std::cerr + abort() to throw std::runtime_error. More friendly to debuggers.


jtkrogel:
This may merit wider discussion in a GitHub issue.  I am aware that C++ exceptions were intentionally avoided in the past, but I am fuzzy on the logic/details.  Could be fine either way in general, but a discussion is worthwhile.

Probably does not need to affect the limited use in this PR for now.

markdewing:
With C++ exceptions - how do they work within MPI?   Does an unhandled exception terminate the program, or cause it to hang?

markdewing:
On my laptop, OpenMPI and MPICH both terminate the program with an unhandled C++ exception.

ye-luo:
Please comment on #125.

ye-luo:
@prckent in those error messages, the function helps very little since the error is from heavily templated areas. Printing function names doesn't add value. Grepping the full error message is anyway needed.

PR #849, opened 2018-05-11
Fix Nexus LiH example

jtkrogel:
The Nexus LiH example currently generates and up-up term in the two body Jastrow input when only an up-down term should be present.  The current PR fixes this behavior in addition to correcting outdated references to "qmcapp".


markdewing:
okay to test

PR #860, opened 2018-05-11
Fix the h5 files of bccH tests for backflow due to wrong orbital normal values

### LINKED ISSUE #804: Orbitals in backflow tests not normalized
https://cdash.qmcpack.org/CDash/testDetails.php?test=2061411&build=21910
@Paul-St-Young is your modification to the cusp make the orbital denormalized?

Paul-St-Young:
Fixes #804 

1. allow the flag "check_orb_norm" in <sposet_builder> to disable orbital norm check.
2. re-normalized cusp corrected orbitals in short-bccH tests.

@ye-luo please check that the "check_orb_norm" flag is an acceptable way to turn off norm check. I tried to localize the changes to `BsplineReaderBase` and derived classes.


Paul-St-Young:
I also removed the electronic density from the pwscf.h5 files, so the binaries are smaller now.

markdewing:
okay to test

src/QMCWaveFunctions/SplineAdoptorReaderP.h
```diff
@@ -357,7 +357,7 @@ struct SplineAdoptorReader: public BsplineReaderBase
       foundit &= h5f.read(cG,s);
       get_psi_g(ti,spin,cur_bands[iorb].BandIndex,cG);//bcast cG
       double total_norm = compute_norm(cG);
-      if(std::abs(total_norm-1.0)>PW_COEFF_NORM_TOLERANCE)
+      if((std::abs(total_norm-1.0)>PW_COEFF_NORM_TOLERANCE)&(checkNorm))
```
ye-luo: In an if-statement && is preferred. Please also move the checkNorm before the check. Please adjust the others also

prckent:
What is the plan for updating the tests and adding a manual entry about this feature?

(I noticed that bsplinereaderbase knows about many inputs that are not described...)

ye-luo:
GridFactor and Rcut should be removed. They belongs to MultiGridBsplineSetReader.h which was removed by the diet.

prckent:
We should create issues for removal and manual updates if they are not going to be handled in this PR before merge.

ye-luo:
I will do the removal and manual update here.

prckent:
@ye-luo Thanks for the cleanup.

PR #862, opened 2018-05-14
Move molecular orbital coefficients to LCAOSet

ye-luo:
This PR is to keep SPOSetBase light.
Historically, the SPOSetBase carries MO coefficients and its related quantities like Identity, BasisSetSize and the heavy IO routines for the coefficients. This is not good design of a base class.
Now all these parts are flagged by ENABLE_SOA. In the SOA design, these are moved to LCAOrbitalSet class and the IO is moved to builders. A lot of piping are removed.

Orbital rotation part of codes are flagged and now protected by appropriate error messages.


src/QMCWaveFunctions/SPOSetBase.cpp
```diff
@@ -98,6 +86,7 @@ SPOSetBase* SPOSetBase::makeClone() const
   return 0;
 }
 
+#if 0
```
markdewing: Is this code intended to be removed?

src/QMCWaveFunctions/SPOSetBase.h
```diff
@@ -332,10 +310,10 @@ class SPOSetBase: public QMCTraits
 
 
 protected:
-  bool putOccupation(xmlNodePtr occ_ptr);
-  bool putFromXML(xmlNodePtr coeff_ptr);
-  bool putFromH5(const char* fname, xmlNodePtr coeff_ptr);
-  bool putPBCFromH5(const char* fname, xmlNodePtr coeff_ptr);
+  //bool putOccupation(xmlNodePtr occ_ptr);
```
markdewing: should this be removed entirely?

src/QMCWaveFunctions/SPOSetBase.cpp
markdewing: D'oh!

PR #864, opened 2018-05-16
Further fixes for VMC SHO tests

markdewing:
Skip the tests in CUDA builds
Increase the tolerance for mixed precision builds (for BG/Q)
Skip the derivative test in mixed precision (the check script is not in the expected directory in mixed precision builds)


prckent:
Why is the check script in a different location for the mixed precision builds?

ye-luo:
If the grad_lap check script is generic, it is better to be placed with other check scripts.

PR #867, opened 2018-05-21
Nexus: fix QmcpackAnalyzer when energy cost is excluded

jtkrogel:
This is a small fix for a bug noticed by Ye Luo.  Essentially if the energy cost function parameter is not present in an input file for an optimization run an attempted multiplication by None can result.


jtkrogel:
Should work now with no cost provided at all.  In this case the QMCPACK defaults of 0.9 energy and 0.1 variance are used.

markdewing:
okay to test

nexus/library/qmcpack_result_analyzers.py
```diff
@@ -67,13 +67,20 @@ def __init__(self,input,opts,energy_weight=None,variance_weight=None,nindent=0):
...
+                ce = 0
+            #end if
             if ew==None:
                 ew = ce
             #end if
             if vw==None:
                 vw = cv
             #end if
         #end if
+        if abs(ew)<tol and abs(vw)<tol:
```
ye-luo: This fixes my issue but I don't like the logic blocks and tol.
if curv,crv,ce are not specified at all, put the qmcpack default.
if any of them is specified, put zero in non specified ones and do necessary checks (sum=1)

jtkrogel:
Better?

PR #868, opened 2018-05-21
Merge SplineAdoptorReaderP and SplineHybridAdoptorReaderP

ye-luo:
When I first developed the reader for constructing orbitals in the hybrid orbital representation, I copied the whole reader for regular splines and add extra stuff. Overtime, I figured out what are the necessary pieces for the extra work for hybrid. So the largely duplicated lines can be removed.

This PR defines the extra work as virtual functions implemented in the reader for hybrid.
The old reader also has diverged code paths for serial and parallel. The serial is added to avoid memory issue but only works for using 1 MPI tasks. Now I do in-place construction using a single implementation and there is no scratch memory issue.

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/HybridAdoptorBase.h
+++ b/src/QMCWaveFunctions/BsplineFactory/HybridAdoptorBase.h
@@ -88,11 +88,6 @@ struct AtomicOrbitalSoA
     chunked_bcast(comm, MultiSpline);
   }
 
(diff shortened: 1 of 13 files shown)


PDoakORNL:
Ok I'm looking at this now

ye-luo:
The unit test_wavefunction checks the regular spline case.
All the short tests are fine and I also checked with production runs.

PR #871, opened 2018-05-23
Nexus: enable POSCAR and CHGCAR file I/O

jtkrogel:
This PR draws POSCAR file I/O from deeper in the Nexus tree up to the more independent fileio.py module level.  CHGCAR I/O added to enable, e.g. conversion from XCrysDen xsf files to CHGCAR.  

Currently the POSCAR I/O functionality is duplicative w.r.t. portions within structure.py and vasp_input.py.  When tests for these code sections are in place later on, the duplicated code portions in these files will be replaced with usage of the PoscarFile class in fileio.py

### DIFF
--- a/nexus/library/fileio.py
+++ b/nexus/library/fileio.py
@@ -22,11 +22,11 @@
 
 import os
 import mmap
-from numpy import array,zeros,ndarray,around,arange,dot,savetxt
+from numpy import array,zeros,ndarray,around,arange,dot,savetxt,empty
 from numpy.linalg import det,norm
 from generic import obj
-from developer import DevBase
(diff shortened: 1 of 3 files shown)


markdewing:
okay to test

jtkrogel:
@prckent I completely agree.  The support for CHGCAR will likely make its way into a future QMCPACK density postprocessing tool (I have a basic one, but I don't think it is ready for broader use).  

This will also take some small modification to the default grids that QMCPACK uses (histogram cell edges aligning with simulation cell edges, supported in xsf but not CHGCAR) to match CHGCAR (histogram cell centers aligning on simulation cell edges).

PR #874, opened 2018-05-25
Update ppconvert

ye-luo:
1. allows changing density mixing parameter via command line.
2. update ppconvert manual section


Paul-St-Young:
@ye-luo thanks for adding more details on `ppconvert` to the manual!

Could you add a few sentences on how a "reference state" should be chosen?
A few examples would be greatly appreciated.

ye-luo:
@Paul-St-Young I added my limited knowledge of how to set the reference states.

Paul-St-Young:
The following script seems to produce a reasonable upf for silicon
`wget http://www.tcm.phy.cam.ac.uk/~mdt26/pseudo_lib/si/1/summary.txt
wget http://www.tcm.phy.cam.ac.uk/~mdt26/pseudo_lib/si/1/pp.data 
wget http://www.tcm.phy.cam.ac.uk/~mdt26/pseudo_lib/si/1/awfn.data_s2p2_3P

BIN=ppconvert

pot=Si.TN
ref="1s(2)2p(2)"
loc=2

cmd="$BIN --local_channel $loc --casino_pot pp.data --s_ref $ref --p_ref $ref --d_ref $ref"

$cmd --upf $pot.upf >& convert_upf.out
$cmd --xml $pot.xml >& convert_xml.out`

Paul-St-Young:
The local part of potential agree very well with the tabulated data.
However, there seem to be some small differences in the pseudized orbitals.

si_pp

Paul-St-Young:
I should mention that my first instinct was to use `3s(2)3p(2)`, because Wikipedia says Si [Ne] 3s(2)3p(2). However, this naive first attempted failed miserably because `ppconvert` puts nodes into the pseudized orbitals.  i.e. `RadialWF::Solve` uses `int TotalNodes = n-l-1;` Perhaps this is something that should be explicitly warned against in the manual.

ref33_si_pp

ye-luo:
The difference in orbitals may be caused by using a different XC functional.

Paul-St-Young:
The converted pseudopotential can reproduce the Si band structure compared to the standard quantum espresso example (PP/examples/example01). I also tried manually replacing the <PP_PSWFC> and <PP_RHOATOM> entries in the UPF with data from awfn.data (<PP_R> and <PP_LOCAL> already exactly matches pp.data).  I saw no change to the band structure. I think the small difference we saw is insignificant.

tn-si-bands

prckent:
@Paul-St-Young Do you have enough information to convert the potentials that you need? Clearly (i) a set of converted potentials would avoid this problem and (ii) a better write up is needed for ppconvert to be useful or an alternative found.

ye-luo:
@prckent I looked at the archived page when I wrote my guidance for reference states. The following requirement is no more needed.
> Note that the "s_ref" needs to include a 1s state (potentially with 0 occupancy), the "p_ref" needs to include a 2p state, and the "d_ref" needs to include a 3d state.

ye-luo:
If all the pseudopotential files are created by ppconvert, this is not a big issue since QE will rescale them all. If mixed pseudopotential files (from OPIUM and BFD) are used, the starting density may be pretty bad.

PR #876, opened 2018-05-31
Improved mkl fftw

PDoakORNL:
This leads to mkl fftw working for both intel, gcc, and clang.


CMake/FindMKL.cmake
```diff
@@ -87,7 +87,8 @@ try_compile(HAVE_MKL_FFTW3 ${CMAKE_BINARY_DIR}
 
 IF ( HAVE_MKL )
   SET( MKL_FOUND 1 )
-  SET( MKL_FLAGS ${MKL_FLAGS} )
+  SET( MKL_FLAGS ${MKL_COMPILE_DEFINITIONS} )
+  include_directories( ${MKL_INCLUDE_DIRECTORIES} )
   IF ( HAVE_MKL_FFTW3 )
     FILE(WRITE "${CMAKE_CURRENT_BINARY_DIR}/include/fftw3.h" "#include <fftw/fftw3.h>\n" )
```
ye-luo: Can we get a chance to avoid the file creation?

markdewing:
okay to test

prckent:
After this PR, it would be a good time to address #750 (@ye-luo ?)

ye-luo:
Wait till my question is addressed

CMake/FindMKL.cmake
PDoakORNL: Sorry I've been busy.  Yes I think we can avoid it, I was just going for minimal change.

ye-luo: Good. Please remove those lines and check if Intel +MKl and non-Intel +MKL is still fine. Then we will settle the manual. I hope we don't need to touch this part for a long time.

PDoakORNL:
MKL_FLAGS is the historical way we dealt with intel's magical -mkl flag in the rest of the project.  But '-mkl' should, I think, be treated like a compile definition in "Modern" cmake's parlance so for a more flexible and maintable try_compile that's what makes sense to use. 

For the main project we are not yet converting over to setting target_compile_definitions and other target properties so for now I think I just pass it out as MKL_FLAGS. However we have started to use the global include_directories so without the magic flag I use them.

PDoakORNL:
I checked intel and clang 4.  Neither require the copied fftw3.h anymore. Additionally the try_compile checks the fftw3.h is actually in the include_paths that will be used.

ye-luo:
Configuration failed last night on EoS.
```
-- MKL found: HAVE_MKL=TRUE, HAVE_MKL_VML=TRUE, HAVE_MKL_FFTW3=FALSE
```
https://cdash.qmcpack.org/CDash/viewConfigure.php?buildid=23701

PR #877, opened 2018-05-31
nexus: fix array initialization in GAMESS input class

jtkrogel:
This is a very simple bugfix for a problem reported by Cody Melton.  Originally the GIarray constructor was called improperly.


markdewing:
okay to test

PR #878, opened 2018-06-04
Nexus: improve user example tests

jtkrogel:
Improve user example tests in two ways: 

1) Split the user example tests into many, i.e. one for each user example test directory rather than a single monolithic test encompassing all of them. 

2) Improve input file fidelity checks in the case of periodic boundary conditions.  Due to rounding errors, an atom at position 0 may be assigned a position at either 0 or L, which are equivalent in PBC but not w.r.t. plain float comparison.  Check whether a comparison of reference and generated input files is failing on the basis of different atomic positions alone and whether the positions differ w.r.t. PBC or not.


ye-luo:
Okay to test

PR #879, opened 2018-06-05
MultiSlater-Jastrow Orbital Optimization code

spinedaflores:
To whom it may concern,

I am creating a pull request to try to get my multislater-jastrow orbital optimization (MSJ-OO) code into qmcpack. I've left comments on about how the code works and about some deficiencies that I hope to get some input into how to resolve. This work was done by extending the current MultiSlaterDeterminantFast class capabilities.


ye-luo:
@spinedaflores Thanks for contributing to QMCPACK. I read about your added line of codes and this feature is important to us. Meanwhile, I think we should take the chance to improve the design of orbital rotation codes added last year and your new contribution since we are anyway moving to the SoA implementations.

The orbital derivatives can be implemented at the SD or MSD level by taking coefficients from SPO or passing the necessary determinant data to the SPO class and evaluate the derivatives inside the SPO class.

The current implementation is the former one. however, certain assumptions of SPOs are not always fulfilled. For example, the C matrix doesn't exist in numerical SPOs like splines. If we need to do rotation on the spline SPO, we should use spline SPO as a basisset for LCAO and then always apply the rotation at LCAO SPO level. Some part of the rotation code shared by SD and MSD doesn't need to be replicated in SD and MSD since they can be inside LCAO classes.

I would prefer the second design that the derivative evaluation is within SPO class since they are derivatives to the SPO rotation parameters. The needed determinant data is passed through well defined SPO APIs. These APIs should be generic as virtual functions and can be used also for orbital optimization using a different optimization scheme other than doing rotation. It is not simple to figure out the final form of the desired API but we can start by asking what is the necessary info to evaluate the derivative. For example, is the data added in the MSD class really belongs to this class or better to be placed in the SPO class? If it is moved to the SPO class, can both SD and MSD cases share it. If so, it is better to move it in to SPO class. Gradually, we will figure out a good API.

The final goal is to find the optimal way to have a maintainable code with maximal feature flexibility.

ye-luo:
@spinedaflores Please prepare the tests for your added feature. There are [Instructions](https://github.com/QMCPACK/qmcpack/wiki/How-to-add-tests). The run needs to be small but sufficient.

ye-luo:
@spinedaflores Any updates?

spinedaflores:
I apologize for the late response time.

I agree with you first comment about moving the derivatives to the SPO class. But I have not begun to make changes in this direction.

I'm currently trying to implement my work for spline SPO in a similar unfavorable manner I currently have. Once I have this working I will be fully ready to move the optimization from the MultiSlater class to the SPO classes. But I am having difficulty understanding how the spline spo class works/ which class is actually being used. 

I am going to send you an email detailing how I plan to implement orbital optimization for the spline SPO, and would like to get your input on how to move forward.

thank you

grahamlopez:
test this please

spinedaflores:
@ye-luo 
I've re-arranged functions in the way we discussed (moved evaluatederivative function for orbitals into the LCAOrbitalSet class). Could you take a look and see if you're ok with what I have? If the general idea is ok I'll keep polishing up what I have.

ye-luo:
@spinedaflores Thank you. I will take a look next Monday. Could you do a self review on github and see if any change you made was not intended? Then fix them.

build/.gitignore
```diff
@@ -1,4 +0,0 @@
-# Ignore everything in this directory
```
ye-luo: Please recover this file.

ye-luo:
@spinedaflores could you add a test? https://github.com/QMCPACK/qmcpack/wiki/How-to-add-tests
Meanwhile I'm still looking at your code. With a test, I can modify the code with more confidence.

build/.gitignore
ye-luo: Correct. I guess you used to remove build folder completely when you start a new build. However, the `build` folder of QMCPACK is more for display and we have a hidden file.

spinedaflores:
> @spinedaflores could you add a test? https://github.com/QMCPACK/qmcpack/wiki/How-to-add-tests
> Meanwhile I'm still looking at your code. With a test, I can modify the code with more confidence.

I'm waiting to get my account at ALCF re-activated to make these tests. Sorry for the delay

spinedaflores:
@ye-luo 

I've included a test for this as tests/molecules/H4_ae/h4_msj_orb_opt.xml and have placed the reference data in tests/molecules/H4_ae/opt-ref-h4_msj_orb_opt

please note that this test name is similar to h4_orb_opt.xml (which is the sinlge slater orbital optimization)

ye-luo:
@spinedaflores Thank you. Could you do me a favor move all the new test to a new folder H4_msd_ae I also feel H4_ae is a bit crowded.

tests/molecules/H4_ae/H4.wfs_4.xml
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0"?>
```
ye-luo: Once this one is moved. Put a better name.

tests/molecules/H4_ae/opt-ref-h4_msj_orb_opt/H4.wfs_4.wfs
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0"?>
```
ye-luo: This is xml file need .xml

tests/molecules/H4_ae/opt-ref-h4_msj_orb_opt/J1.H.dat
```diff
@@ -0,0 +1,100 @@
+                     0            -0.0675875                     1              -4.41832
```
ye-luo: This file doesn't need to included in the reference.

tests/molecules/H4_ae/opt-ref-h4_msj_orb_opt/J2.ud.dat
```diff
@@ -0,0 +1,100 @@
+                     0               1.10763                  -0.5              0.278323
```
ye-luo: This file doesn't need to included in the reference.

tests/molecules/H4_ae/opt-ref-h4_msj_orb_opt/J2.uu.dat
```diff
@@ -0,0 +1,100 @@
+                     0              0.392521                 -0.25             -0.107449
```
ye-luo: This file doesn't need to included in the reference.

tests/molecules/H4_ae/opt-ref-h4_msj_orb_opt/h4.info.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
```
ye-luo: This file doesn't need to included in the reference.

tests/molecules/H4_ae/opt-ref-h4_msj_orb_opt/h4.s000.qmc.xml
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ye-luo: This file doesn't need to included in the reference.

tests/molecules/H4_ae/opt-ref-h4_msj_orb_opt/h4.s002.qmc.xml
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ye-luo: This file doesn't need to included in the reference.

tests/molecules/H4_ae/opt-ref-h4_msj_orb_opt/h4.s004.cont.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0"?>
```
ye-luo: This file doesn't need to included in the reference.

ye-luo:
Please remove all the *.qmc.xml *.stat.h5 output.cobalt output.error *.info.xml

ye-luo:
Could you please review the PR
1. remove commented lines which are no more needed?
2. I saw boost::format was used, is it absolutely needed? We are sensitive to have extra dependency on boost. Minimal to none use is preferred.
3. formic wrappers can not be used outside the optimizer, please use Numerics/OhmmsBlas.h

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -306,23 +309,48 @@ bool SlaterDetBuilder::put(xmlNodePtr cur)
...
         up_det = new MultiDiracDeterminant((SPOSetPtr) spomap.find(spo_alpha)->second,0);
         app_log() <<"Creating base determinant (down) for MSD expansion. \n";
         dn_det = new MultiDiracDeterminant((SPOSetPtr) spomap.find(spo_beta)->second,1);
         multislaterdetfast_0 = new MultiSlaterDeterminantFast(targetPtcl,up_det,dn_det);
-        //          up_det->usingBF = UseBackflow;
-        //          dn_det->usingBF = UseBackflow;
-        //          multislaterdetfast_0->usingBF = UseBackflow;
+        //remove this later...
+//DELETE LATER        (OrbOpt) ? (multislaterdetfast_0->Orbopt=true): (multislaterdetfast_0->Orbopt=false);
+        (OrbOpt) ? (up_det->Optimizable=true,dn_det->Optimizable=true): (up_det->Optimizable=false,dn_det->Optimizable=false);
```
ye-luo: Directly put orbopt=="yes" without the need of OrbOpt seems cleaner.

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -152,6 +152,11 @@ class SPOSet: public QMCTraits
   virtual void checkObject() const {}
 #endif
 
+  ///test function
+  virtual void test() {}
```
ye-luo: Do we need this one?

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
```diff
@@ -48,7 +55,10 @@ namespace qmcplusplus
     vgl_type Temp; 
     ///Tempv(OrbitalSetSize) Tempv=C*Temp
     vgl_type Tempv; 
-
+    //lookup table mapping the unique determinants to their element position in C2_node vector
+    std::vector< std::vector<int> > lookup_tbl;
```
ye-luo: This data seems MSD specific not generic for SPO.

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
```diff
@@ -60,11 +70,136 @@ namespace qmcplusplus
...
+    {
+      app_log()<<"test function test function SDP \n";
+    }
+    /// create optimizable orbital rotation parameters
+    void buildOptVariables(std::vector<RealType>& input_params, bool params_supplied, std::vector<int> * data, const size_t& nel, std::vector<size_t>& C2node, const int& spin);
+
+    ///helper function to buildOptVariables
+    int build_occ_vec(std::vector<int> * data, const size_t& nel, const size_t& nmo, std::vector<int>* occ_vec);
+  
+    void evaluateDerivatives (ParticleSet& P,
```
ye-luo: This function seems very heavy and seems only used for MSD optimization.
Is it possible to split it the math two levels. In the SPO level, only the evaluation of derivatives or some intermediate data shared by single SD and MSD and then in the single SD and MSD compute the final derivatives?

ye-luo: I'm thinking of the case the spo_up and spo_dn is the same pointer for a restricted HF. This is very common for solid because double the spo adds more memory which we can't afford. LCAO code is less restrictive on memory. So the lazy solution is always building SPO twice even they are identical. In the case optimization, not taking into account spin up/down symmetry may cause severe issue because of spin contamination. This may needs a bit non-trivial modification of the current implementation. I will not block you PR because of this but it seems some thing we must solve.

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
```diff
@@ -60,11 +70,136 @@ namespace qmcplusplus
...
+          const int p = m_act_rot_inds[i].first;
+          const int q = m_act_rot_inds[i].second;
+          // m_first_var_pos is the index to the first parameter of the spin up electrons...
+          const RealType x = myVars[i] = active[loc];
+          //app_log() <<"active ["<< loc << "] = "<< x<< "\n";
+          rot_mat[p+q*nmo] =  x;
+          rot_mat[q+p*nmo] = -x;
+        }
+
+        this->exponentiate_antisym_matrix(nmo, &rot_mat[0]);
```
ye-luo: Remove this-> if not necessary.

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
```diff
@@ -126,6 +261,8 @@ namespace qmcplusplus
         ValueVector_t& psi, GradVector_t& dpsi, ValueVector_t& d2psi) const;
     void evaluate_vgl_impl(const vgl_type& temp, int i,
         ValueMatrix_t& logdet, GradMatrix_t& dlogdet, ValueMatrix_t& d2logdet) const;
+    //helper function to resetParameters
+    void exponentiate_antisym_matrix(const int n, RealType* const mat);
```
ye-luo: Mark this helper function under private

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
spinedaflores: This is true. I was thinking that I would have overloaded functions for evaluatederivative and buildOptVariables specific for the single slater case (if that was your concern). Does that seem reasonable?

spinedaflores:
@ye-luo I think I've address all your comments with 2 exceptions.
1) I was using a lapack function (zheev ) which is why I had included <formic/utils/lapack_interface.h>. But I have now replaced it with <Numerics/Blasf.h> and but had to add the zheev function to the list of lapack fuctions in that file. 
2) You mention how my evaluatederivative function and buildoptvariables function are both specific to MSD wfns. I was hoping to create overloaded functions for the single slater wavefunction to use. Is that acceptable?

ye-luo:
Please add also cheev in Blas.f and make a interface heev similar to hevr in OhmmsBlas.h.
In this way, we won't have issue compiling the mixed precision code.

ye-luo:
Adding virtual functions needs to be careful. For evaluatederivative if we can have one function supporting single and mult- dets. It will be worth doing it.
buildoptvariables depends where it has been called. If it is in the builder, we probably can avoid adding any virtual functions.
We don't need to solve this problem now. Let us get the PR merged first.

src/QMCWaveFunctions/lcao/LCAOrbitalSet.cpp
```diff
@@ -393,6 +393,39 @@ namespace qmcplusplus
       if(Optimizable)
       {
         Table.resize(nel,nmo);
+
+			  Bbar.resize(nel,nmo);
```
ye-luo: Please replace all the tabs.

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
```diff
@@ -263,6 +217,54 @@ namespace qmcplusplus
     //helper function to resetParameters
     void exponentiate_antisym_matrix(const int n, RealType* const mat);
 
+		//helper function to evaluatederivative; evaluate orbital rotation parameter derivative using table method
+		void table_method_eval(std::vector<RealType>& dlogpsi,
+													 std::vector<RealType>& dhpsioverpsi,
+													 const ParticleSet::ParticleLaplacian_t& myL_J,
+													 const ParticleSet::ParticleGradient_t& myG_J,
+													 const size_t& nel,
```
ye-luo: No need of using reference for single plain old datatype.

spinedaflores:
I think I've made all the necessary changes. The previous merge conflict has been resolved.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.h
```diff
@@ -55,6 +55,7 @@ class MultiSlaterDeterminantFast: public WaveFunctionComponent
 
 public:
 
+  bool Optimize;
```
ye-luo: What is the difference between base class Optimizable and your Optimize?

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
ye-luo: You can have two dvaluatederivate interfaces for single and multi dets if that is needed.
buildOptVariable should be the same for single and mult dets.
If any data structure is needed by multi dets only, it needs to be moved to the MSD class and initialized in the MSD::buildOptVariables.
Here lookup_tbl doesn't seems needed for single determinant.

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
```diff
@@ -40,6 +43,10 @@ namespace qmcplusplus
      * makeClone makes a shallow copy
      */
     ValueMatrix_t* C;
+    ValueMatrix_t* m_init_B;
```
ye-luo: documentation needed.

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
```diff
@@ -40,6 +43,10 @@ namespace qmcplusplus
...
+
+    ParticleSet::ParticleGradient_t myG_temp, myG_J;
+    ParticleSet::ParticleLaplacian_t myL_temp, myL_J;
```
ye-luo: These four elements do not really belong to SPO conceptually. If you really need to use them in evaluateDerivates, create it inside the routine.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.h
spinedaflores: 'Optimizable' for the Multi-slater class is set to true if "orbitals" or "CI coeffecients" are going to change. 'Optimize' is for the CI coefficients specifically. So for example if checkinvariables is called and optimize is set to false, then only the Multidiracdeterminant objects will call checkin

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
```diff
@@ -127,6 +213,58 @@ namespace qmcplusplus
...
+                          const size_t nmo,
+                          const int offset1,
+                          double* T
+                          );
+
+
+    ValueMatrix_t  Table;
+    ValueMatrix_t Bbar;
+    ValueMatrix_t Y1,Y2,Y3,Y4,Y5,Y6,Y7,Y11,Y23,Y24,Y25,Y26;
+    ValueMatrix_t pK1,K1T,TK1T, pK2,K2AiB,TK2AiB,K2XA,TK2XA,K2T,TK2T,MK2T, pK3,K3T,TK3T, pK4,K4T,TK4T, pK5,K5T,TK5T;
```
ye-luo: Comments needed for all the member objects.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.h
ye-luo: Is SPOSet::Optimizable a better choice? Make it controlled in the input section of SPO.

spinedaflores: So I didn't do that originally because SPOSet cannot be added to the TrialWavefunction as a TrialWaveFunctionComponent. It now lives inside of MultiDiracDeterminant. Is there a way to change this so that SPOSet can be added to TrialWavefunction and I can make the change you suggested?

spinedaflores: Sorry If my last comment is unclear. I guess my issue is that although SPOset has it's own Optimizable variable, when evaluateDerivative is called to the trialwavefunction it only touches the trialwavefunction components. So how can I get the SPOSet to be called as well?

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
```diff
@@ -259,8 +260,76 @@ namespace qmcplusplus
...
+    $
+    \newline
+    Now with all these matrices and constants the expressions of dhpsioverpsi and dlogpsi can be created.
+    In addition I will be using a special generalization of the kinetic operator which I will denote as O. Our Slater matrix with the special O operator applied to each element will be called B_bar
+
+    $
+    ``Bbar"_{i,j} =\nabla^2 \phi_{j}(r_i) + \frac{\nabla_{i}J}{J} \cdot \nabla \phi_{j}(r_{i})  + \frac{\nabla^2_i J}{J} \phi_{j}(r_{i})
+    $
+    */
+    ValueMatrix_t Table;
```
ye-luo: Short descriptions are preferred in the header files. You can add a pointer to the cpp file for the long description to keep clarity and avoid duplication.
In principle, every data object used only for the table method should not be held by LCAOrbitalSet which need to be agnostic to single or multi SD. Please mention that the following objects are only for the table method in the short description.

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
ye-luo: Please do this

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
ye-luo: @spinedaflores any comment on this?

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
ye-luo: Any comment on this?

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
ye-luo: This was done.

src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
spinedaflores: I’ve placed the member variables in lcao that were only relevant to the table method inside the table_method_eval function. Note that this will make the function evaluation slower.

spinedaflores:
I’ve placed the member variables in lcao that were only relevant to the table method inside the table_method_eval function. Note that this will make the function evaluation slower.

I’ve changed the optimization flag for orbitals to be placed inside the “sposet” start tag instead of using “OrbOpt” inside the “multislaterdeterminant” tag

I’ve removed a helper function “create_tables” in LCAOrbitalSet that was supposed to used for MSD case and Single Slater case. I’ll re-include this when I include the single slater optimization code.

Using just Optimizable for the Multislater class and not having a separate member variable for “optimize CI coefficients” causes a problem with the QMCCostFunction::Return_t QMCCostFunction::correlatedSampling(bool needGrad) function during the three shift adaptive run. The correlatedSampling function utilizes TrialWaveFunction::RealType TrialWaveFunction::evaluateDeltaLog which loops through the trial wavefunction components and checks if the waveFunctionComponent is “Optimizable”. Now since the SPOSet isn’t checked it never calculates the updated value of the Multi-Slater portion of the wavefunction. Do you have a suggestion for resolving this?

ye-luo:
The check-in, evaluate, check-out of parameter should follow the hierarchy.
Namely, the check-in, evaluate, check-out routines of MSD class should invoke the three routines of SPOSet it owns.
SPOSet do have 
```
  virtual void checkInVariables(opt_variables_type& active) {}
  virtual void checkOutVariables(const opt_variables_type& active) {}
```
and LCAOrbitalSet needs to specialize them.

ye-luo:
Does it make sense to have
`Optimizable = Optimizable || Dets[0]->Optimizable || Dets[1]->Optimizable`
When the MSD class is built?

spinedaflores:
I don't think that would be OK for cases in which the user is optimizing the SPOSet while holding the CI coefficients fixed.  If we optimize solely SPOSets, when the MSD->checkin/checkout/evaluatederivatives is called the behavior won't be defined because optimizable is set to true while the number of myVars is zero. Unless I create a new member variable in MSD, something like "CI_Optimizable" I think the suggested change could cause problems without more modifications.

ye-luo:
I would go for CI_Optimizable which will be a flag only for the local parameters excluding the lower levels.
The input flag for MSD should be probably connected to CI_Optimizable instead of Optimizable.

ye-luo:
What is the added test currently doing?

spinedaflores:
The added test performs exclusively orbital optimization (no other parameters are optimized) on a non symmetry H4 geometry.

If you're referring to the fact that the file was modified recently, I just removed "OrbOpt=yes" in favor of having "optimize=yes" in the "start sposet tag"

I've replaced the use of type double with real type so that the mixed_precision flag will compile

ye-luo:
1) Can we have the test case starting with Jastrow optimized?
2) Are opt_vars_up and opt_vars_dn rotation parameters?
I guess they live in LCAOrbitalSet. Should we have their input through sposet tag?
```
      <sposet basisset="LCAOBSet" name="spo-dn" size="4" optimize="yes">
        <occupation mode="ground"/>
        <coefficient size="4" id="downdetC">
  5.04668000000000e-01  4.50114000000000e-01  3.58423000000000e-01  1.26449000000000e-01
 -2.40452000000000e-01 -3.20413000000000e-01  4.66777000000000e-01  7.03549000000000e-01
  8.80080000000000e-02 -5.04842000000000e-01  8.07522000000000e-01 -7.19301000000000e-01
  1.03323500000000e+00 -8.77213000000000e-01 -3.90492000000000e-01  2.12318000000000e-01
</coefficient>
          <opt_vars size="6">
0.0 0.0 0.0 0.0 0.0 0.0
          </opt_vars>
      </sposet>
```

spinedaflores:
1. I think we can have the test start any way you want. I previously had the CI + jastrow parameters fixed and optimized so that just the orbital optimization was the only feature being tested for. 
2. Yes, these are rotation parameters. I'll relocate where these are read in

ye-luo:
1) I mean CI + Jastrow pre-optimized and fixed in this test. Pre-optimization is to reduce the error bar such that we can test the numbers more precisely.

spinedaflores:
OK, I think the test was already built that way, but I'm not 100% sure. So i'll double check that

ye-luo:
Okay to test

ye-luo:
No worry about the gpu build. There is an issue in the CI.

spinedaflores:
I don't think the CI and Jastrow parameters were fully optimized for the test I had up. I'll update it

spinedaflores:
I take back what I said about the test i have added. I think it's fine

prckent:
Can this be merged? My reading of the discussion is that everything requested has been done (thanks)

ye-luo:
Waiting for an email reply from @spinedaflores .

spinedaflores:
@ye-luo 
I believe I've made all the changes we discussed in the email. Let me know if anything else comes up

prckent:
Thanks @spinedaflores . I removed the WIP and await @ye-luo 's comments here. In a *future* PR it would be helpful to have a brief description in the manual. This could be a discussion of the example/test. i.e. just enough to ensure that your new capability can be more easily found. This would also be a good place to cite a published or upcoming paper.

prckent:
Thanks @spinedaflores  @ye-luo

PR #880, opened 2018-06-06
Remove duplicate Nexus example tests

jtkrogel:
Remove ctests that duplicate Nexus user example tests in ntest.  Clean and partly modernize the user example tests.  Fix failing settings test when called from ctest (attempted double read of command line flags).


ye-luo:
Okay to test

PR #882, opened 2018-06-07
Nexus stampede

camelto2:
This enables nexus to run on the Stampede2 machine. This includes the information on the various available queues on Stampede2, including the KNL and the new SKX queues. 

One additional change that was needed was to add the squeue status RD for held jobs, which was missing. Additionally, there was a special case for held jobs on stampede that needed to be addressed (lines 1592-1595).


jtkrogel:
This looks fine to me.  Have you run the "ntest" executable to check for any failures?  The current test set takes <30s to run.

camelto2:
All tests pass for ntest

ye-luo:
A general comment: keep adding stuff in machine.py is not a good.
Just like our build script, every machine should be a small piece of script which can be loaded individually. If a machine retires, we can just remove the file not always editing a single file.

jtkrogel:
I don't see why this is a problem.  I'm fine with a single file.

OK to merge from me.

ye-luo:
Okay to test

PR #884, opened 2018-06-08
Nexus: fix convert4qmc workflow in the case -nojastrow is not provided

jtkrogel:
Nexus originally expected convert4qmc to produce a wfnoj.xml file.  This file is only produced if the -nojastrow flag is provided and so workflows not involving this flag will not complete successfully, since convert4qmc is erroneously assumed to have failed (no wfnoj.xml file present).  

This PR fixes the issue by judging convert4qmc completion as successful if either wfj.xml or wfnoj.xml are produced.  Either file is now accepted as wavefunction input to QMCPACK.

Issue reported by Chandler Bennett.


ye-luo:
Okay to test

Issue #873, opened 2018-05-25
Problems with unicode characters in manual and some LaTeX installations


prckent:
Some LaTeX installs will not build the manual successfully due to unicode characters in the source .tex files.  The characters appears to be mostly pipes and dashes in the tutorials and could be removed or replaced.  (Discovered with a fresh TeX live 2018 install on os x.)

```

! Package inputenc Error: Unicode character │ (U+2502)
(inputenc)                not set up for use with LaTeX.

See the inputenc package documentation for explanation.
Type  H <return>  for immediate help.
 ...

l.46 │

```

ye-luo:
It also hits me 2 minutes ago with Tex live 2018 on linux.

prckent:
I think the preferred route  is to remove the unicode, as opposed to tweaking the package use or adding definitions - we need to move in a direction easier to convert to html or to use a markup language. 

@ye-luo Do you agree?

ye-luo:
I also prefer removing than fixing.

jtkrogel:
Agree.  Prefer replacement with nearest available ASCII (which may constitute a removal in some cases).

PDoakORNL:
Notable TeX Live changes
Two changes may be immediately noticeable. The first is that the default encoding for input sources to LaTeX has changed from ISO Latin 9 to UTF-8 Unicode. Standard ASCII files satisfy both encoding rules and thus are accepted as usual. Latin 9 is an encoding standard that adds to ASCII various accents, umlauts, and special characters like upside down question marks used in Western Europe. From 2018 on, such input files require a LaTeX header line

 
 	\usepackage[latin9]{inputenc}

UTF-8 is an encoding standard for full unicode, which can simultaneously encode most scripts used across the world: Roman, Arabic, Chinese, Hebrew, Hieroglyphics, Japanese, etc. Until now, Latex sources using this encoding required the header line

	\usepackage[utf8]{inputenc}

but this line is now optional because UTF-8 is the default.

PR #886, opened 2018-06-09
Texlive2018 fixes

PDoakORNL:
With a free Friday night I engaged in some latex hacking. Unnecessary but I couldn't help it.  Also not feeling sharp enough for C++. 

No we can have our cross platform texlive2018 build and cool directory tree diagrams.  Prelude to full html output in the next PR.


prckent:
Is someone able to confirm that this still works with older/other tex installs? The manual must build in a wide variety of environments.

ye-luo:
I thought in last conversation on issue #873, we had decided not to open the door to unicode.

ye-luo:
png image is also less supported than pdf image.

prckent:
@ye-luo Agree. Unless there are strong benefits, we should not go in this direction.

prckent:
The PDF image is a PNG though!

[ I wish all the graphics in the manual were vector art since it helps repurposing them and they can always be at a good resolution. ]

ye-luo:
For example, arXiv refuses PNG images but accepts PDF images encapsulating PNG.

prckent:
I am interested in wide opinions about these kinds of updates. For example, one could argue that a complicated build procedure is not a problem provided we can generate PDF and HTML, and keep the current and previous versions online. Similar to our compiler support, building with the last two major versions of texlive might be a reasonable test.

PDoakORNL:
It's 2018 unicode has been around my whole adult life, it is not new, nor should it be exotic. Xetex is packaged in texlive and has also been around forever, it can also handle modern font's properly.

if we provide good build script's or they can download the manual for the release version from the website why do we need to support obsolete tex installs?

When I began work I assigned  #873 to myself in waffle and had worked this out before I knew about Ye's PR. I also did not vote for ASCII which seemed to be because we don't want to take the trouble, I decided to see how hard that was.  Not very, html conversion pretty hard, unicode not hard. I didn't think we had an ideological commitment to english is the best character encoding ;)

I was well into the html conversion by the time I saw #885 and the already merged PR. I think fontspec alone is worth being in xetex regardless of supporting unicode.

ye-luo:
The PR was not from me although I merged it.
Making things more beautiful is not sufficient to use it. I still prefer minimal maintenance. Wider fonts characters support is not necessary, we probably won't write Chinese in the manual.
The most portable solution is pain text or may be a bit markdown. We won't always edit manual with GUI and may just vim or emacs on a terminal. Unicode will definitely complicate the situation. 
I also feel encode is not the major difficulty when converting to good html.
If we are using fancy packages in the current manual, they should probably be cleaned up.

markdewing:
Trying with TeX Live 2015 (Ubuntu 16.04) gives an undefined control sequence error for \IfFontExistsTF

prckent:
@ye-luo One step at a time.

I was unable to build this PR on oxygen with texlive installed with the package manager. Note that a fair number of non-default packages are already needed (bibtopic, placeins) as well as xetex (xetex, collection-xetex).

The \IfFontExistsTF macro was not defined. Commenting this out yielded the following from the build_manual.sh script:

```
(./jastrow_two_body_spline.aux))
Runaway argument?
{\contentsline {subsubsection}{$\mathbf
! File ended while scanning use of \@writefile.
<inserted text>
                \par
l.163 \begin{document}
```

markdewing:
I agree with Ye that we are unlikely to need the expanded character support provided by Unicode in the manual source (at least not right now).

The directory trees might be better rendered using a package (dirtree?) rather the manually-added lines.

Rather than replace pdflatex with xelatex right away, it might be better to add a build_manual_xelatex.sh to introduce a switch more gradually.

PDoakORNL:
As far as unicode in editors:
Emacs has no trouble with this.
vim over iterm2 from oxygen no problem either.  I don't know about the whole chinese unicode character set but monotype box characters are no a problem, hiragana are no problem.

Os's packages are generally out of date and for the OS's use i.e. they don't keep track of the whole spec tree in general just that a set work together as a release.

That's why spack, homebrew and others exist.
```
brew cask install mactex
spack install texlive scheme=full
```
on oxygen this may work...
```
. /data/epd/spack/share/spack/setup-env.sh
spack load texlive
```
I guess I can add some conditionals to still support a latex pdf build but are we really going to get pushback?  I practically never build packages docs if they have a working website.

prckent:
Do not merge until confirmation is posted that this PR builds on ccms-cdash (the VM that builds and hosts the PDF). It will not currently.

ye-luo:
I don't have root on my workstation. I hit the same issue as Paul with texlive 2015.
Repeat my opinion, no need for more character or better looking fonts.

The current problem in manual is too long lines not fully displayed and command lines in the manual cannot copy paste to terminal and work right away. This really affect how the manual can be used.

It will be great if we have a manual which can be opened with web browser.

prckent:
ccms-cdash builds this PR - I falsely asserted that it would not, above. We have new enough TeX there apparently.

I would like to understand the errors with TexLive2015 before proceeding in case there is an easy fix.

PDoakORNL:
So maybe this will satsify everyone, I don't have an obsolete texlive to test with, they seem to purposely make such a thing nontrivial to get.

markdewing:
Both builds worked on Texlive 2013 (Ubuntu 14.04), though the pdflatex build looked bad until the 'cm-super' package was installed ( issue described here: https://tex.stackexchange.com/questions/172079/t1-encoding-causing-bad-pdf-quality )

The font used in the xelatex build looks better in the Chrome pdf viewer than in evince (the default Ubuntu pdf viewer).  Maybe a newer version of evince will work better?

prckent:
Duped by #887

PR #892, opened 2018-06-14
Including optDetCoeffs flag into convert4qmc

mcbennet:
1) Adding "optDetCoeffs" flag to convert4qmc to turn on optimization of CI coefficients (specifically, this sets "optimize" attribute in "multideterminant" element to 'yes') and setting default behavior to false. 

2) Including this functionality into Nexus.

These changes allow the user to avoid having to change the wf xml file by hand after running convert4qmc, in the case that the default behavior is not desired.

If it is decided that the original default behavior of convert4qmc should not be changed (namely optimize='yes') then the flag could be changed to "freezeDetCoeffs" and its impact would be to set optimize='no'


markdewing:
okay to test

jtkrogel:
Having "opt_dets" is fine.  Please also add "opt_det_coeffs" so that command line users can easily transliterate from "optDetCoeffs".

markdewing:
@anbenali Any opinion on changing the default on the determinant coefficient optimization?

ye-luo:
I think this is a good change because Jastrows should be always optimized before det coefs. However, in the noJ case, you may want yes by default. I'm not saying one way or another, you guys decide.

prckent:
Use of multideterminants and potentially optimizing their coefficients is advanced usage. I think their optimization should be off by default everywhere?. Then the behavior between J and noJ example wavefunctions will be the same. (Possibly the optimizers need patching to correctly print "nothing to optimize" and not abort in the case where there are no free parameters - this can be done later.)

PR #894, opened 2018-06-17
Llvm4 cuda build

PDoakORNL:
The addition of the XRAY_GPU_TRACE macro is to allow instrumentation of many small calls on the CPU related to the setup of the GPU code. For builds other than clang+cuda it expands to nothing.  For the clang+cuda combination with -DXRAY_PROFILE=1 llvm-xray instruments the code.

From the perspective of all other builds no change should be seen.

After the CUDA is ported to SOA we can explore removing the attribute macro in favor of a file enumerating functions to be profiled.  I think the attribute macros in the source will make refactoring easier for now.


CMakeLists.txt
```diff
@@ -684,6 +688,24 @@ ELSE(QMC_CUDA)
   MESSAGE(STATUS "Disabling CUDA")
 ENDIF(QMC_CUDA)
 
+SET(XRAY_PROFILE FALSE CACHE BOOL "Use llvm xray profiling")
```
ye-luo: Please add comments.

ye-luo:
Usually I don't like to see compiler/tool related instrumentation in the source code unless is is very local.
We did that for vTune but is within the scope of our timer. Namely, we capture the important pieces with QMCPACK timer and make it reports accurate timing. If profiler needs instrumentation, better to have them within our timer frame.

The current timer doesn't have sufficient resolution in the GPU code path. Probably we need to improve it and then hook the XRay in them.

I don't really think this is entangled with needed SoA change. If you can simplify them by enumerating functions to be profiled, no need to wait. On the other hand, again improve our timer in GPU is always preferred.

GPU_XRAY_TRACE is added sometimes before the function, sometimes after the function, sometimes after function name before arguments...

PDoakORNL:
_The timers do not supply tracing._

C++11 style attributes are tricky they either refer to the element immediately proceeding or after but when in the beginning before a method or function they refer to the whole thing (at least in clang). I'd say the standard is vague and the logic of the clang implementation is not completely transparent to me yet. Also to build cuda code that can be traced along side we need to use the basically obsolete clang 4.0.1. I'd certainly prefer they all go in the same place. 

As I said in the PR I think it will be easier to work with while I'm really working on this. They will be quite easier to remove later. I do not want to constantly have to flip back and forth to another file to see what is instrumented nor do I want to have to be constantly adding and removing things from it.

CMakeLists.txt
```diff
@@ -684,6 +688,24 @@ ELSE(QMC_CUDA)
...
+SET(XRAY_INSTRUCTION_THRESHOLD 200 CACHE INT "Instruction threshold for xray instrumentation")
+SET(XRAY_GPU_MOST FALSE CACHE BOOL "Manually instrument almost all of the CUDA implementations calls on the CPU")
+
+IF(XRAY_PROFILE)
```
markdewing: Can this clang-specific setup be put in CMake/ClangCompilers.cmake ?

CMakeLists.txt
PDoakORNL: I left help strings and documentation in examples/profiling. Could add documentation to the manual or wiki but I don't think this is necessary for this to be merged..

ye-luo: I didn't mean that. Wiki or manual is not necessary for this at the moment. I was asking for just one or two lines of comment what this is. Since it has been moved to Clang, the cached variable is self-explained.

ye-luo:
This is just my general concern. A lot of legacy stuff I saw in the code probably starts like we can put it now, it is convenient and can be removed in the future. Then overtime, people forgot it. Your changes are not a big deal. They should easy to remove by a script.

Going back to the topic. Usually, profiling tools come with a set of APIs to start and pause collection. It is irrelevant to tracing. Tools can record events during the data collected part and provide tracing. Xray doesn't have that kind of APIs?

PDoakORNL:
yes it has that as well.  But control over which functions are have the nop_sled instructions insert so they can later be patched is controlled at compile time.

I'm copying traces that are basically single file webapps that allow graphic exploration to this webserver
http://cdash-minimal.ornl.gov/profiling

PDoakORNL:
Ok made a separate PR for the hdf5 link issue.  And as far as I can see that's the last issue before merging this.

PDoakORNL:
Rebased again.

ye-luo:
Okay to test

PDoakORNL:
Thanks for the much better title.

Issue #893, opened 2018-06-14
NiO benchmark fails on Titan with CUDA and more than 1 MPI rank


markdewing:
Error message:
```
cublas_inverse failed on walker 0, getrf error 125, getri error 177.
Offending hostname = nid02810
Offending device = 0
```
(The error means the one of the factorized matrices has a zero diagonal entry, leading to a singular matrix)

It works okay with one MPI rank.  It fails if using 2 or 4.
Running command is 'aprun -n 2 -N 1 qmcpack --verbosity=high NiO-fcc-S8-dmc.xml'

It worked on the last benchmark run (05/16/2018)
So far the failure is localized between these commits

Works:
```
commit 6c2d9bd4ba1b397badd588789f6c310ada6b9833
Merge: 71f86adb6 02a9e5e11
Author: Mark Dewing <markdewing@gmail.com>
Date:   Mon May 21 13:19:03 2018 -0500

    Merge pull request #823 from ye-luo/spline-VGH

    Complete the VGH routines of the Spline SPO SoA implementation

```

and 

Fails:
```
commit 071c46fef621ba28f5b7f910a43b3ef2224c330f
Merge: 9e4bb0634 24943e67a
Author: Peter Doak <PDoakORNL@users.noreply.github.com>
Date:   Tue May 22 17:38:26 2018 -0400

    Merge pull request #868 from ye-luo/merge-hybrid-reader

    Merge SplineAdoptorReaderP and SplineHybridAdoptorReaderP
```

I suspect something in #868 is causing the problem.  The other PR merged in the range, #867, is a change in Nexus.

ye-luo:
I think #868 is the problem. The communication pattern is changed for C2R case.

prckent:
This indicates we need to update the test matrix to have multiple GPU runs with MPI, since this important breakage was not picked up earlier. Ideally our future CI system would catch this immediately with a deterministic test.

prckent:
@ye-luo Can you look at this or should Mark?

ye-luo:
I will do it.

PR #896, opened 2018-06-18
Fix GPU C2R spline communication

ye-luo:
Address #893


markdewing:
It works on the 32-atom benchmark test case.

PR #897, opened 2018-06-19
Required to avoid linking errors with hdf5

PDoakORNL:
Ok here is the hdf5 linking added which llvm was sensitive too.  Perhaps other tool chains as well if a rpath vs. LD_LIBRARY_PATHS style build environment is used.


ye-luo:
I don't feel this is actually right. HDF5_LIBRARIES is already included in QMC_UTIL_LIBS in the top level and then lower level should only use QMC_UTIL_LIBS consistently. Second, when you archive a library why all the dependent libraries are needed in the archive?

PDoakORNL:
The issue is with symbols the downstream libraries use and that qmcutil does not.

ye-luo:
The current develop builds fine with Clang-3.8+CUDA 8 on my desktop.

markdewing:
Could you be more specific about the linking errors or copy the error message?

PDoakORNL:
Their were H5P symbols that weren't found at link time.  Unfortunately I don't have an environment I can easily reproduce this on right now, and its wasting time. I've moved to more spack in my build which polutes the environment quite a bit.

I think that this is a handbuilt openmpi plus handbuild hdf5 problem. For now I'm just going to rebase this out of the llvm_xray PR and close this one.

PR #898, opened 2018-06-19
Titan requires cmake3 doesn't need 2 cmakes

PDoakORNL:
Installation guide updated to work, build_olcf_titan.sh fixed

cmake no longer needs to be run twice on titan


PDoakORNL:
Do we use another Cray machine with the Cray programming environment build system.
i.e. where 
($ENV{CRAYPE_VERSION}) is defined?
I imagine they might require this same fix.

But I'd rather see it merged than waste time.

ye-luo:
ALCF Theta and NERSC Cori both use Cray PE. The environment variable is set by cray pe module.

ye-luo:
Okay to test

PR #900, opened 2018-06-21
Nexus: add GAMESS examples for RHF, CISD, and CASSCF

jtkrogel:
This PR includes user examples for GAMESS originally contributed by Allison Dzubak.  The examples comprise RHF, CISD, and CASSCF calculations for the water molecule that closely parallel a subset of the calculations in the molecular multideteriminant QMCPACK Lab.  These new examples are also incorporated into the testing system which currently checks the generated input files against included reference files.

In order to support these types of calculations, a limited facility was added to represent folded molecular structures since, if one is using symmetry, GAMESS requires the molecular coordinates in the folded representation.  Care was taken to distinguish this type of folding from translational tiling.  In the future, the required inputs (currently full and folded coordinates) might be simplified further with explicit support for known point group symmetries.


markdewing:
okay to test

PR #903, opened 2018-06-25
Attempt fix for test_MO with Intel SoA build

markdewing:
Add a call to ions.update to ensure that RSoA.copyIn get called.

It appears that ions must be updated before elec (because the call to elec.update uses
data from ions).

With this change, the run has no uninitialized value errors from valgrind (w/o the change, there are some uninit values errors)

The test failure is here.  I could not reproduce it.  Fix is a guess.
https://cdash.qmcpack.org/CDash/testDetails.php?test=2544595&build=24654


markdewing:
test this please

Issue #728, opened 2018-04-02
Incorrect reference values used in long-FeCO6 tests


prckent:
Based on the most recent long runs, https://cdash.qmcpack.org/CDash/viewTest.php?onlyfailed&buildid=20711 it appears that some of the reference values for long-FeCO6 are incorrect. E.g.  long-FeCO6_gms-vmc_b3lyp_noj-1-16-samples . These failures could be hiding other problems.

anbenali:
addressed by #905

PR #905, opened 2018-06-26
Update PySCF converter to Python3; QP parser orbital count with spin fix; Fix FeCO6 and LiH_ae example

anbenali:
- Moving Pyscf Converter to Python 3
- Fixing QP parser to force the same number of orbitals in spin up and spin down calculations for multideterminants
- Fixing FeCO6 example (converged calculation + restart option in Pyscf)


markdewing:
Regarding Python 3 - it would be best to target code that is compatible with Python 2 and 3.

For the print function, add the line "from __future__ import print_function" to the beginning of the Python file, and then the print function will be available in Python 2, as well as being the default in Python 3

There are some packages to hide some of the other compatibility issues (six, future).  The 'future' package contains a script 'futurize' that can help automate some of the translation.

markdewing:
As far as availability of Python 2 and 3 - it seems on Cooley at ALCF the base system python is version 2, but the more advanced python distributions (Ananconda, Intel) are version 3.

anbenali:
The dependence here is only when running pyscf itself. The dependency is not in qmcpack at all since the only file touched is the parser that is called inside pyscf. This file generates the HDF5 file from Pyscf. Since pyscf has moved to python3 and will not guarantee support of python2  I preferred to fix it now since it required minimum changes.

PDoakORNL:
I would advocate longer term we push our users to embrace python environments. If you are running a python code for science it should really have a dedicated stable and known python environment. With how easy [miniconda](https://conda.io/miniconda.html) is to install and use, switching between python 2 and 3 and different versions of python packages is trivial now. It is to treacherous to depend on system python or that a particular python module on a HPC is going to last forever. We should just provide instructions to use conda environments for those that will be using the python elements of the system.

jtkrogel:
If Pyscf will only support python 3 going forward, then Anouar's updates make sense.  The choice of the Pyscf team might spell trouble for being able to run all Python tools consistently in the same environment, though.

I agree that it is best overall to target Python 2+3 compatibility (apparently a foregone conclusion for the Pyscf team).  In the context of Nexus, an upgrade to Python 3 compatibility is somewhere on the horizon (while maintaining reverse compatibility with Python 2 via six or similar), following greater rounding out of the example/test system.

Hopefully Guido van Rossum will think harder before making another round of backwards-incompatible changes in Python 4.

markdewing:
What is the plan for pyscf's support of python?   The current README indicates version 1.5 supports both python 2 and python 3

prckent:
Do the tests require python3?

anbenali:
Not at all. This will only affect running pyscf. Since they already support
python3 nothing matters from our side. Again this is the python script
called from pyscf that generates an HDF5. It is never called by any of our
files including nexus.

prckent:
Since these changes don't affect testing (e.g. a python 3 install would be required alongside python 2), and any users of PySCF will have to have python 3 anyway, I am happy to reset the python 3 tripwire.

If there are any other concerns that need to be request, people should add a blocking review.

@anbenali What other updates do you plan? We won't merge until you remove the [WIP].

anbenali:
Paul,

This PR will concern all molecular tests (FeCO6+LiH) and the manual update
for multideterminants.
Please do not accept just yet. I will keep my hard deadline for Thursday
night.

ye-luo:
@anbenali If possible, separating the test fix and manual updated is preferred. If the test fix lands earlier, we can have the nightly tests running at least 1 night before making a release. manual can be updated even at the last minute.

anbenali:
Ok!
Thanks

On Jun 30, 2018 15:06, "Ye Luo" <notifications@github.com> wrote:

@anbenali <https://github.com/anbenali> If possible, separating the test
fix and manual updated is preferred. If the test fix lands earlier, we can
have the nightly tests running at least 1 night before making a release.
manual can be updated even at the last minute.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/QMCPACK/qmcpack/pull/905#issuecomment-401569094>, or mute
the thread

prckent:
@ye-luo Certainly. We can't make a release unless the integration tests have run. That means they need to be in soon.

anbenali:
This PR will address the following issues:
#729 long-FeCO6_pyscf total energy checks failing
#728 Incorrect reference values used in long-FeCO6 tests 
#448 What should be in the example input files generated by the molecular converter? (already addressed in the WIKI)

tests/molecules/FeCO6_b3lyp_pyscf/CMakeLists.txt
```diff
@@ -1,46 +1,50 @@
...
-  QMC_RUN_AND_CHECK(long-FeCO6_pyscf-vmc_b3lyp_noj
-                    "${CMAKE_SOURCE_DIR}/tests/molecules/FeCO6_b3lyp_pyscf"
-                    vmc_b3lyp_noj
-                    vmc_b3lyp_noj.in.xml
-                    16 1
-                    TRUE
-                    0 FeCO6_PYSCF_SCALARS # VMC
-                    )
+# Ref for long VMC test
+  LIST(APPEND FeCO6_PYSCF_SCALARS "kinetic" "163.8420 0.0022") # kinetic energy
```
ye-luo: Are these scalars used?

anbenali: I preferred to leave the scalars from the reference run.

prckent: Looks like the PYSCF_SCALARS can go...

ye-luo: Keep the reference number is good.
Please change lines from `LIST(APPEND FeCO6_PYSCF_SCALARS "kinetic" "165.48 0.36") # kinetic energy`
To simply `# kinetic energy 165.48 0.36`

ye-luo:
Do we need all `FeCO6_b3lyp_gms/gamess/*.xml` files?

PR #912, opened 2018-07-03
Add option for building with NVTX in the CUDA code

jefflarkin:
This patch does 2 things:

1) Adds an option to cmake that adds a USE_NVTX compiler define and adds the libnvToolsExt library to the link step.
2) Adds nvtx ranges to the DMC::run, VMC::run, and VMC::runWithDrift functions.

This was done to simplify GPU profiling, but will also help @atillack with some debugging we've been doing recently. The changes will have no effect if cmake is run without -DUSE_NVTX=1. This sets a simple path for adding more ranges and markers in the future.


prckent:
OK to test

src/QMCDrivers/DMC/DMC_CUDA.cpp
```diff
@@ -75,6 +78,9 @@ void DMCcuda::checkBounds (std::vector<PosType> &newpos,
 
 bool DMCcuda::run()
 {
+#if USE_NVTX
```
PDoakORNL: I'd rather see these wrapped in macros, They will be easier to strip out if they were unambiguous single line macros.

atillack: Do you mean #if defined(USE_NVTX), since defined(..) is technically a macro, or writing wrappers for all possible ways to inject NVTX commands? There are a handful of ways to strip #ifdef to next #endif blocks. I would do it with awk but there is also unifdef.

ye-luo: Only this place is `#if USE_NVTX`, it needs to be changed to ifdef at least.

CMakeLists.txt
```diff
@@ -678,6 +678,15 @@ ELSE(QMC_CUDA)
   MESSAGE(STATUS "Disabling CUDA")
 ENDIF(QMC_CUDA)
 
+IF(USE_NVTX)
+  IF(HAVE_CUDA)
+    ADD_DEFINITIONS(-DUSE_NVTX)
```
ye-luo: I also prefer to avoid ADD_DEFINITIONS.
If a macro is not used local to a file, we have `src/config.h.cmake.in` to register all the macros.

src/QMCDrivers/DMC/DMC_CUDA.cpp
jefflarkin: Yes, that was mistake, definitely need to fix it. Thanks.

CMakeLists.txt
jefflarkin: Moving to config.h makes sense.

CMakeLists.txt
```diff
@@ -678,6 +678,15 @@ ELSE(QMC_CUDA)
   MESSAGE(STATUS "Disabling CUDA")
 ENDIF(QMC_CUDA)
 
+IF(USE_NVTX)
```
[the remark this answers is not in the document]
jefflarkin: @PDoakORNL I'm a cmake novice...For using `target_link_libraries`, am I correct in thinking that this should go in `src/QMCApp/CMakeLists.txt`?

prckent:
While it would be great if this can be done in the "least polluting" and most extensible way possible, I don't think the current way is problematic. We already have USE_VTUNE_API , and this PR implements something similar at the same level of CMakeLists.txt. The source updates are restricted to CUDA related code, so are not great currently. Improvements can be done iteratively.

Question:
Should it be USE_NVTX_API for consistency?

Request:
Add a short section to manual/external_tools.tex saying what this is and how to activate it.

CMakeLists.txt
```diff
@@ -678,6 +678,16 @@ ELSE(QMC_CUDA)
   MESSAGE(STATUS "Disabling CUDA")
 ENDIF(QMC_CUDA)
 
+SET(USE_NVTX 0 CACHE BOOL "Enable/disable NVTX regions in CUDA code.")
+IF(USE_NVTX)
+  IF(HAVE_CUDA)
+    ADD_DEFINITIONS(-DUSE_NVTX)
```
ye-luo: Can we remove this add_definition? Since it is now handled in config.h

CMakeLists.txt
```diff
@@ -678,6 +678,16 @@ ELSE(QMC_CUDA)
   MESSAGE(STATUS "Disabling CUDA")
 ENDIF(QMC_CUDA)
 
+SET(USE_NVTX 0 CACHE BOOL "Enable/disable NVTX regions in CUDA code.")
```
ye-luo: How about moveing this part next to the VTUNE section?

ye-luo:
Agree with @prckent, we can start this way. Also USE_NVTX_API is a better name.

CMake/GNUCompilers.cmake
```diff
@@ -41,8 +41,8 @@ if(CMAKE_CXX_FLAGS MATCHES "-march=" OR CMAKE_C_FLAGS MATCHES "-march=")
   endif() #(CMAKE_CXX_FLAGS MATCHES "-march=" AND CMAKE_C_FLAGS MATCHES "-march=")
 else() #(CMAKE_CXX_FLAGS MATCHES "-march=" OR CMAKE_C_FLAGS MATCHES "-march=")
   # use -march=native
-  SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -march=native")
-  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native")
+  #SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -march=native")
```
ye-luo: This change slipped.

ye-luo: If you have the problem with native on power, please submit a separate PR transferring the https://github.com/QMCPACK/miniqmc/pull/87

jefflarkin: Yup, just discovered that. Backing it out.

CMakeLists.txt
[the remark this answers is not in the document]
ye-luo: @PDoakORNL Could you give more detail about using target_link_libraries? On the other hand, I don't feel necessary to put them in low levels. First, the drivers may be used by both qmcpack and unit text executables. Second, low level is hard to look for. So putting them at top level is simple and clean.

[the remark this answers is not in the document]
PDoakORNL: You have a point I just want linking to be fast and not just linking everything to everything.  Since this is a dev tool only and the rest of the build system doesn't really provide a guide that can happen in another PR if I or someone else cares enough.

CMakeLists.txt
```diff
@@ -790,6 +791,14 @@ IF (USE_VTUNE_API)
   LINK_LIBRARIES("${VTUNE_ITTNOTIFY_LIBRARY}")
 ENDIF()
 
+SET(USE_NVTX_API 0 CACHE BOOL "Enable/disable NVTX regions in CUDA code.")
+IF(USE_NVTX_API)
+  IF(HAVE_CUDA)
+    LINK_DIRECTORIES("${CUDA_TOOLKIT_ROOT_DIR}/lib64")
```
PDoakORNL: I still have the issue that this link directory is not correct for all platforms.  I will fix.

ye-luo:
@PDoakORNL Changing C++ handling needs to be in a different PR with the raising of cmake version. Your library detection change is an improvement. However, target_link_libraries to qmcutil, qmcham, qmcwfs basically means still going everywhere. I prefer the simple solution that add the library at the top level. This option is only used when profiling with a specific tool, making the linking a bit slow is accepted.

PDoakORNL:
See following commit it was a mistake.

PDoakORNL:
Supposedly libraries coming in via target_link_library enables better link dependency.  But perhaps best left for a different PR.

PDoakORNL:
Now the review scope doesn't change. Project linking is another issue.

jefflarkin:
Yes, this change works for me. Thanks for fixing it up.

Issue #913, opened 2018-07-03
Propagation of type aliases through class hierarchies


PDoakORNL:
Getting your type aliases through a base class like QMCTraits becomes problematic when you have a templated base class. 
```c++
template<class FT>
class OneBodyJastrowOrbitalBsplineSoA :
  public J1OrbitalSoA<FT>
{
  // The following is so we can refer to type aliases(defs) below the
  // templated base class in the object hierarchy
  // Mostly QMCTraits here
  using JBase = J1OrbitalSoA<FT>;
  using RealType = typename JBase::RealType;
  using ValueType = typename JBase::ValueType;
  using GradType = typename JBase::GradType;
  using PosType = typename JBase::PosType;
...
}
```  
In fact getting these type aliases from J1OrbitalSoA has nothing to do with it being a OneBody..'s base class at all. Suggestions? I don't really think QMCTraits belongs in the class hierarchy at all, but sometimes getting a type from your inheritance tree seems more useful.

jtkrogel:
I actually expect a significant amount of our include spaghetti comes from round about type accesses through mechanisms like these.

PDoakORNL:
In terms of consistent types I don't think inheritance is a good way to enforce this and to me it looks ineffective in the current code. I'd prefer matching template parameters or they use a global types object. There are too many layers including MI and simply getting the type through inheritance is too vague. It should be crystal clear which RealType or whatever you are using.

For this I prefer to see something like
`using CTS = CUDAGlobalTypes;`
then CTS::RealType instead of CudaRealType or RealType

or 
using QTS = QMCGlobalTypes;
QTS::RealType instead of RealType where you have gotten it from somewhere in your large MI inheritance tree.

Additionally this allows a class template like this
template<P, V> // P=precisionType, V=valueType
```C++
class Whatever
{
using QTS = QMCTypes<P, V>
...
}
```
This allows generation of however many template classes you need
in the executable.
i.e. in the actual code there is object code
Whatever<double, double>
Whatever<double,std::complex<double>>
Whatever<single>
Whatever<single, std::complex<single>>
And this are distinct types.

One can write another class template like
```C++
template<P, V>
class UsesWhatever
{
using QTS = QMCTypes<P,V>
Whatever<P,V>* myWhatever;
void setWhatever(Whatever<P,V>*);
QTS::ValueType* evalSomething(ParticleSet& P, ...);
}
```

ye-luo:
[...]

PR #914, opened 2018-07-03
[WIP] GPU AoS to SoA just jastrows

PDoakORNL:
So I think this is a successful swap of the GPU Jastro's onto the SoA J1OrbitalSoA and J2OrbitalSoA classes. The tests I expect to pass have passed.  Don't see a significant performance change from AoS GPU on the tests or smaller performance tests (up to a64).

If any of the following seems unbelievable go read about two phase lookup.
Because J1(2)OrbitalSoA is templated the derived classes One(Two)BodyJastrowOrbitalBsplineSoA do not enjoy the derived class syntax you might expect when refering to members and types from the base.

* Base members require access via `this->`
* typedefs/type aliases "inherited" from the templated base classes are not accessible simply. #913 
* The templated derived class must be explicitly instantiated if it has out of line defined methods (these do and right now only <double> is instantiated).  This should be done only once and other compilation units should pick it up through extern. I believe even more care is required if the instantiated template should be exported by a library.

ToDo:
- [ ] Which other types of One(Two)BodyJastrowOrbitalBsplineSoA require instantiation.


ye-luo:
Please
1. Remove J1OrbitalCUDASoA.h and J2OrbitalCUDASoA.h unless they are needed.
2. Merge TwoBodyJastrowOrbitalBsplineSoA to TwoBodyJastrowOrbitalBspline with ifdefs. The same for OneBody.

src/QMCWaveFunctions/Jastrow/OneBodyJastrowOrbitalBsplineSoA.h
```diff
@@ -0,0 +1,193 @@
...
+#include "QMCWaveFunctions/Jastrow/BsplineJastrowCuda.h"
+#include "QMCWaveFunctions/Jastrow/BsplineJastrowCudaPBC.h"
+#include "Configuration.h"
+
+namespace qmcplusplus
+{
+
+template<class FT>
+class OneBodyJastrowOrbitalBsplineSoA :
+  public J1OrbitalSoA<FT>
```
ye-luo: We only have BsplineFunctor working on GPU. So there is no need to make it a template function. The old way is sufficient.

src/QMCWaveFunctions/OrbitalBase.h
```diff
@@ -44,7 +44,7 @@ struct NLjob
     walker(w), elec(e), numQuadPoints(n)
   { }
 };
-#endif
+#endif // defined(QMC_CUDA) && !defined(ENABLE_SOA)
```
atillack: Your #endif says it's for `#if defined(QMC_CUDA) && !defined(ENABLE_SOA)` while the the code reads `if defined(QMC_CUDA)`. Which one is correct?

src/QMCWaveFunctions/OrbitalBase.h
```diff
@@ -606,7 +606,7 @@ struct OrbitalBase: public QMCTraits
     app_error() << "Required CUDA functionality not implemented. Contact developers.\n";
     abort();
   }
-#endif
+#endif //defined(QMC_CUDA) && !defined(ENABLE_SOA)
```
atillack: Same issue (#endif comment does not match #if statement).

grahamlopez:
test this please

src/QMCWaveFunctions/OrbitalBase.h
PDoakORNL: The first, I had a more ambitious port going of which this is a cut down PR. Will fix.

src/QMCWaveFunctions/OrbitalBase.h
PDoakORNL: see above

src/QMCWaveFunctions/CMakeLists.txt
```diff
@@ -115,6 +119,28 @@ IF(OHMMS_DIM MATCHES 3)
     ENDIF(HAVE_EINSPLINE)
   ENDIF(QMC_CUDA)
```
ye-luo: Miss matching like 97

src/QMCWaveFunctions/CMakeLists.txt
```diff
@@ -115,6 +119,28 @@ IF(OHMMS_DIM MATCHES 3)
...
+      Jastrow/TwoBodyJastrowOrbitalBspline.cpp
+      )
+    IF(HAVE_EINSPLINE)
+      SET(FERMION_SRCS ${FERMION_SRCS}
+        EinsplineSetCuda.cpp
+        PhaseFactors.cu
+        AtomicOrbitalCuda.cu
+        )
+    ENDIF(HAVE_EINSPLINE)
+  ENDIF(QMC_CUDA and NOT ENABLE_SOA)
```
ye-luo: Miss matching line 122. case sensitive.

prckent:
OK to test

PDoakORNL:
As far as the history of the copied files, git as it turns out has a switch in log for this --follow.
```shell
[epd@oxygen llvm_nvcc_qmcpack]$ git log --pretty=oneline --follow src/QMCWaveFunctions/Jastrow/OneBodyJastrowOrbitalBsplineSoA.h
e5c0c9e3bf3637dbd10a4c8fa3c4ac2a1b2cb6ee Slighter better type propagation into JastrowBsplineSoA classes
8b8258494b9492db7c6f9b0a88e6f02dd37c6118 CUDA Jastros appear to be working over SoA implementation.
14417e398e918a42953c1ed209afa6b74967be67 Derived CUDA JastroOrbitals rolled back many cuda changes
1ada2a1a927b5ca9679519b073acd8fc1af7a2fc Better Setup for C++11 attributes
3ea9660ae368ecd67f85ac07ac4345b79a19ced0 Found more cpu to cuda call to trace.  Made Macro
23e0200e67565e0dba86b78cefd88695f00cd1ad maybe most host to gpu call instrumented
a2ad08ea1f4c0e7c6454652c71823c3d646e449c Fix the GPU complex WF optimization.
9f56aa0cde00189ce81fe1abde9024ce82603100  Fixing Typos and Creator affiliations
fcaa478fa5c09ee6b62054c7a3114713478e5e67  All header Files have been normalized. History of Authors have also been included. I used latest affiliation and latest known emails except for people I did not know (prior to 2006). Please let me know if any change is needed regarding an affiliation
b8decbc6337ac3813f64443f8bb8c042e7cf7f2c Fixing compile errors with CUDA due to the removal of using namespace std
fae28636dd84a2de6d85200e3631a81718eaeb67 Removing using namespace std from headers
348a402d46c4e2c6def19c701b2fd273024ceed2 Merge CUDA improvements from NCSA branch.
75c8ed1e4fc69438b5d606ea48726dc34c8ac73f code formatting
48e2b11da8c44d73f37a204515e04764617a591e Enable spin-dependent J1.
b87ed9e12f668083a37fb3687a7f112912d7f193 Adding non-periodic evaluation of Bspline Jastrows on GPU.
5a73baac4c1ff951fdfc69cb7a57bf1c0a4e2a1b Renamed files to reflect division between PBC and non-PBC versions.
1d1522d03c4da6e6fcc3d4eb0e15552f0b983a29 Rename CUDA functions for Jastrows to have _PBC appended to make room for non-PBC versions.
06c3ec6ca19a92a49a743a06d55d2b0e5ea433dd Merging GPU code from the branch into the mainline repository.  Many files are affected.  Please consider this an alpha commit, and check results from runs after updating.
```
This performs well.

`git blame -C -C -C` does the same for blame but does not.
So I am still looking at another method that doesn't require git to match up the histories via a search.

ye-luo:
how about copy the old ones to XXAoS and put your changes for SoA on the current file?

PDoakORNL:
I think I found a solution.  If it works I'll close this branch and another PR will appear.  Same changes but better history.

PR #915, opened 2018-07-04
Add pw2qmcpack patch for QE 6.3

ye-luo:
The 5.3 patch is kept since the 6.2.1 patch has not been released in the past.
Compared to 6.2.1 patch, 6.3 patch adds the atomicnumber tag in the ion section of ptcl.xml file generated by pw2qmcpack to avoid NLPP satinity check failure.
Currently pw2qmcpack doesn't provide the input blocks for QMCPACK and hopefully this will be added in the near future.

### DIFF
--- a/external_codes/quantum_espresso/README
+++ b/external_codes/quantum_espresso/README
@@ -1,7 +1,7 @@
 The files in this directory support making a modified version of
 quantum espresso that includes the pw2qmcpack converter.
 
-The download_and_patch_qe6.2.1.sh script will automatically download
+The download_and_patch_qe6.3.sh script will automatically download
 and patch quantum espresso. Similarly named scripts are provided for
 earlier versions. After patching you can move the espresso directory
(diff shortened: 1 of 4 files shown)


ye-luo:
@prckent I added a line of printout `Example configure: ./configure --with-hdf5=path_to_hdf5_include_and_lib`
Currently pw2qmcpack fails at runtime if HDF5 flags are not set, we can actually throw error at compile time and point users to read the full instruction in the manual. Does this solution sounds better?

prckent:
We should throw an error at compile time - it is very frustrating to have pw2qmcpack build but then fail to operate. This could be addressed in a later PR but is a nice feature to include in v3.5.0. What do you think?

ye-luo:
Now an error is given at compilation
`HDF5 flag neither enabled during configure nor added manually in make.inc`

prckent:
We should check this e.g. on oxygen before the release, but since it can't break  current implementations I am happy to merge.

PR #917, opened 2018-07-05
Align pade2 input tag as pade

### LINKED ISSUE #413: Need consistent input tags in pade Jastrows.
Pade uses `var`
Pade2 takes `Var` and `parameter`

Should at least all support `parameter`.

ye-luo:
Now pade2 no more accept parameter but var/Var.
As @jtkrogel pointed out that 'parameter' should only accept a name/value pair and will be parsed by designated parser(validation will be added in the future).
close #413

### DIFF
--- a/src/QMCWaveFunctions/Jastrow/PadeFunctors.h
+++ b/src/QMCWaveFunctions/Jastrow/PadeFunctors.h
@@ -427,7 +427,11 @@ struct Pade2ndOrderFunctor:public OptimizableFunctorBase
     while(tcur != NULL)
     {
       std::string cname((const char*)(tcur->name));
-      if(cname == "parameter" || cname == "Var")
+      if(cname == "parameter")
+      {
+        throw std::runtime_error("Keyword 'parameter' has been replaced by 'var'");
+      }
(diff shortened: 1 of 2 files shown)


ye-luo:
@PDoakORNL Thanks for the correction.

src/QMCWaveFunctions/Jastrow/PadeFunctors.h
```diff
@@ -427,7 +427,11 @@ struct Pade2ndOrderFunctor:public OptimizableFunctorBase
     while(tcur != NULL)
     {
       std::string cname((const char*)(tcur->name));
-      if(cname == "parameter" || cname == "Var")
+      if(cname == "parameter")
+      {
+        throw std::runtime_error("Keyword 'parameter' has been replaced by 'var'");
```
markdewing: This should give some idea of the location in the XML file where change is needed - parameter is not being replace by var everywhere.  Maybe it could read "Keyword 'parameter' has been replaced by 'var' in jastrow/correlation" ?

ye-luo: That is a good point.

PR #918, opened 2018-07-06
Set LCAO pyscf tests based on SOA_SUCCESS

ye-luo:
pyscf uses spherical tensor and only SoA can run.


markdewing:
How is the code supposed to fail in the AoS case - halt with an error message or produce an incorrect answer?

ye-luo:
Yes. An error will appear
https://cdash.qmcpack.org/CDash/testDetails.php?test=2646497&build=25212

PR #923, opened 2018-07-11
Restore the header of bibliography in the manual

### LINKED ISSUE #921: Header of bibliography missing in manual
We have two sections of references, the first one was before Appendix. However, it has no chapter/section title.

PDoakORNL:
So is this what we want?
Fixes #921


markdewing:
okay to test

PR #924, opened 2018-07-12
Remove the Intel specific code-path in einspline_allocator

ye-luo:


markdewing:
Is this being removed because it's unneeded?

ye-luo:
Not necessary and now causing problems with unconventional compilers.

PR #926, opened 2018-07-12
fixing ifdef typo for AoS, been there for a while

PDoakORNL:


ye-luo:
Okay to test

ye-luo:
If this change makes problem, I need to look at the code since I made the mistake.

PDoakORNL:
rhea is having issues

PR #930, opened 2018-07-16
Add unit test for MultiQuinticSpline1D

markdewing:
Also fix a small bug in the 3rd derivative routine.


markdewing:
It was a compile error as soon I as used it in the unit test.  The code was not called from QMCPACK, and being a template, it never got compile

anbenali:
Ouch.. do you have an idea of the effect of this bug. I don't think that it
is too serious but from your test you maybe have a idea?? It should still
be in the nose but I don't see where the "d" comes from.

ye-luo:
@anbenali Third derivative is only needed for the optimization of backflow. Not used at the moment.

markdewing:
The formula is the same as in OneDimGridBase.h, but it was missing the array references.

PR #933, opened 2018-07-18
Enable mkl docs

### LINKED ISSUE #750: ENABLE_MKL not described in manual and README

PDoakORNL:
Adding documentation for ENABLE_MKL.
Closes #750

### DIFF
--- a/manual/installation.tex
+++ b/manual/installation.tex
@@ -184,8 +184,6 @@ compilers, etc.  The list of environmental variables is given below:
 CXX              C++ compiler
 CC               C Compiler
 MKL_HOME         Path for MKL
-MKL_ROOT
-MKLROOT
 LIBXML2_HOME     Path for libxml2
 HDF5_ROOT        Path for HDF5
 BOOST_ROOT       Path for Boost
@@ -220,7 +218,7 @@ ENABLE_TIMERS         Enable fine-grained timers (1:yes, 0:no (default)).
 ENABLE_SOA            (Experimental) Enable CPU optimization and GPU support using
                       Structure-of-Array (SoA) datatypes (1:yes, 0:no (default)).
 \end{verbatim}
-\normalsize
+
 \item General build options
 \verbatimfont{\footnotesize}%
 \begin{verbatim}
@@ -243,7 +241,7 @@ CMAKE_CXX_FLAGS      Set the C++ flags.  Note: to prevent default
                      Also supported: CMAKE_CXX_FLAGS_DEBUG,
                      CMAKE_CXX_FLAGS_RELEASE, and CMAKE_CXX_FLAGS_RELWITHDEBINFO
 \end{verbatim}
-\normalsize
+
 \item Additional QMCPACK build options
 \verbatimfont{\footnotesize}
 \begin{verbatim}
@@ -252,32 +250,38 @@ QMC_EXTRA_LIBS      Add extra link libraries
 QMC_BUILD_STATIC    Add -static flags to build
 QMC_DATA            Specify data directory for QMCPACK performance and integration tests
 QE_BIN              Location of Quantum Espresso binaries including pw2qmcpack.x
-ENABLE_MKL          Enable Intel MKL libraries.
-MKL_ROOT            Path to MKL libraries (only necessary for non intel compilers
-                    or intel without standard environment variables.)
-                    One of the above environment variables can be used.
 \end{verbatim}
-\normalsize
+
 \item libxml2 related
 \verbatimfont{\footnotesize}%
 \begin{verbatim}
 Libxml2_INCLUDE_DIRS   Specify include directories for libxml2
 Libxml2_LIBRARY_DIRS   Specify library directories for libxml2
 \end{verbatim}
-\normalsize
+
 \item HDF5 related
 \verbatimfont{\footnotesize}%
 \begin{verbatim}
 ENABLE_PHDF5    1(default)/0, enables/disable parallel collective IO.
 \end{verbatim}
-\normalsize
+
 \item FFTW related
 \verbatimfont{\footnotesize}%
 \begin{verbatim}
 FFTW_INCLUDE_DIRS   Specify include directories for FFTW
 FFTW_LIBRARY_DIRS   Specify library directories for FFTW
 \end{verbatim}
-\normalsize
+
+\item Intel MKL related
+\verbatimfont{\footnotesize}%
+\begin{verbatim}
+ENABLE_MKL          Enable Intel MKL libraries (1:yes (default for intel compiler),
+                                                0:no (default otherwise)).
+MKL_ROOT            Path to MKL libraries (only necessary for non intel compilers
+                    or intel without standard environment variables.)
+                    One of the above environment variables can be used.
+\end{verbatim}
+
 \item CTest related
 \verbatimfont{\footnotesize}%
 \begin{verbatim}
@@ -285,7 +289,7 @@ MPIEXEC                Specify the mpi wrapper, e.g. srun, aprun, mpirun, etc.
 MPIEXEC_NUMPROC_FLAG   Specify the number of mpi processes flag,
                        e.g. "-n", "-np", etc.
 \end{verbatim}
-\normalsize
+
 \end{itemize}
 
 \subsection{Role of QMC\_DATA}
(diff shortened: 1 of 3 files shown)


manual/installation.tex
```diff
@@ -196,71 +203,89 @@ \subsection{Configuration options}
...
+QMC_COMPLEX           Build the complex (general twist/k-point) version (1:yes, 0:no)
+QMC_MIXED_PRECISION   Build the mixed precision (mixing double/float) version
+                      (1:yes (GPU default), 0:no (CPU default)).
+                      The CPU support is experimental.
+                      Use float and double for base and full precision.
+                      The GPU support is quite mature.
+                      Use always double for host side base and full precision
+                      and use float and double for CUDA base and full precision.
+ENABLE_MKL            Enable Intel MKL libraries.
+MKL_ROOT              Path to MKL libraries (only necessary for non intel compilers
```
ye-luo: ENABLE_MKL and MKL_ROOT are not key build options. Please list them under addition build options

manual/installation.tex
```diff
@@ -392,10 +422,11 @@ \subsection{Using vendor optimized numerical libraries, e.g. Intel MKL}
 \subsubsection{Using Intel MKL with non-Intel compilers}
 
 To use Intel MKL with, e.g. an MPICH wrapped gcc:
+\verbatimfont{\footnotesize}
 \begin{verbatim}
 cmake \
  -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx \
- -DBLA_VENDOR=Intel10_64lp_seq -DCMAKE_PREFIX_PATH=$MKLROOT/lib \
+ -DBLA_VENDOR=Intel10_64lp_seq -DENABLE_MKL -DMKL_ROOT=$MKLROOT/lib \
```
ye-luo: I don't think cmake accepts -DENABLE_MKL without a value.
Is just -D DENABLE_MKL=1 sufficient without DBLA_VENDOR and MKL_ROOT?

prckent:
OK to test

manual/installation.tex
PDoakORNL: If a MKL_HOME, MKLROOT, or MKL_ROOT environment variable is set, then yes it is.

PDoakORNL:
I believe I've fixed the itemize environments. I don't know why I broke them. Spasmodic pasting is what it looks like.

manual/installation.tex
```diff
@@ -331,9 +359,10 @@ \subsection{Example configure and build}
 \end{verbatim}
 
 \item Move to build directory, run cmake and make
+\verbatimfont{\footnotesize}
 \begin{verbatim}
 cd build
-cmake -D CMAKE_BUILD_TYPE=Release ..
+cmake -D CMAKE_BUILD_TYPE=Release -DENABLE_MKL=1 ..
```
ye-luo: In this case, if MKL_HOME is specified. No need to put it on cmake.

manual/installation.tex
```diff
@@ -196,71 +203,89 @@ \subsection{Configuration options}
...
+\normalsize
 \item Additional QMCPACK build options
+\verbatimfont{\footnotesize}
 \begin{verbatim}
 QMC_INCLUDE         Add extra include paths
 QMC_EXTRA_LIBS      Add extra link libraries
 QMC_BUILD_STATIC    Add -static flags to build
 QMC_DATA            Specify data directory for QMCPACK performance and integration tests
 QE_BIN              Location of Quantum Espresso binaries including pw2qmcpack.x
+ENABLE_MKL          Enable Intel MKL libraries.
```
ye-luo: Please do as other libraries. Not mixed with QMC_XXX.
Also mention turn on by default when intel compiler is used.

manual/installation.tex
```diff
@@ -292,13 +318,14 @@ \subsection{Role of QMC\_DATA}
 QMC_DATA/NiO/NiO-fcc-supertwist111-supershift000-S128.h5
 QMC_DATA/NiO/NiO-fcc-supertwist111-supershift000-S256.h5
 QMC_DATA/NiO/NiO_afm_fcidump.h5
-QMC_DATA/NiO/NiO_afm_wfn.dat 
-QMC_DATA/NiO/NiO_nm_choldump.h5   
+QMC_DATA/NiO/NiO_afm_wfn.dat
+QMC_DATA/NiO/NiO_nm_choldump.h5
 \end{verbatim}
-
+\normalsize
```
ye-luo: I sometimes see normalsize, sometimse don't. Could you explain why?

manual/installation.tex
```diff
@@ -392,10 +422,11 @@ \subsection{Using vendor optimized numerical libraries, e.g. Intel MKL}
 \subsubsection{Using Intel MKL with non-Intel compilers}
 
 To use Intel MKL with, e.g. an MPICH wrapped gcc:
+\verbatimfont{\footnotesize}
 \begin{verbatim}
 cmake \
  -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx \
- -DBLA_VENDOR=Intel10_64lp_seq -DCMAKE_PREFIX_PATH=$MKLROOT/lib \
+ -DBLA_VENDOR=Intel10_64lp_seq -DENABLE_MKL=1 -DMKL_ROOT=$MKLROOT/lib \
```
ye-luo: I think if ENABLE_MKL=1 is enabled, BLA_VENDOR is skipped completely.

manual/installation.tex
```diff
@@ -175,10 +179,13 @@ \subsection{Environment variables}
 A number of environmental variables affect the build.  In particular
 they can control the default paths for libraries, the default
 compilers, etc.  The list of environmental variables is given below:
+\verbatimfont{\footnotesize}%
 \begin{verbatim}
 CXX              C++ compiler
 CC               C Compiler
 MKL_HOME         Path for MKL
+MKL_ROOT
```
ye-luo: Sorry. I was giving the wrong info.
Please list only MKL_HOME for environment variable.
When we tell users, just give them 1 but our setting can be smarter internally.

manual/installation.tex
PDoakORNL: You shouldn't see it at all.  Will fix.

manual/installation.tex
PDoakORNL: Correct should be deleted

manual/installation.tex
PDoakORNL: Good that will be less confusing

ye-luo:
A question.
We have MKL_HOME historically. With the improved MKL setting, MKL_ROOT was introduced and used heavily in FindMKL script.
Can we reduce them to one? I prefer MKL_HOME since this is used historically.
When MKL_HOME cmake variable is not defined, we can still look for MKL_ROOT, MKLROOT, ENV{MKLROOT}, ENV{MKL_HOME}

PDoakORNL:
I used MKL_ROOT because it's more often used in cmake FindXXX.cmake includes. Since I've worked on several of them I of course want them to be as consistent as possible inside.
While the Findxxx.cmake's are far not 100% consistent they seem more consistent that different vendors chose environment variables.

ye-luo:
VASP has no configure step and users directly define MKL setting in Makefile.
QE directly inject MKL when Intel compiler is used. If mixing GNU and MKL is needed, modify make.inc.
Gromacs needs -DMKL_INCLUDE_DIR and -DMKL_LIBRARIES explicitly.
LAMMPS has the same setting as Gromacs.
None of them has advanced script like us.

ye-luo:
I'm OK to have either MKL_HOME or MKL_ROOT but not MKL_HOME in the environment with MKL_ROOT in the cmake.
however, Intel defines MKLROOT which is too close to MKL_ROOT.
We can leave the change to a separate PR but I preferred to get this settled before this is exposed to users (release).

PDoakORNL:
It just annoys me. I vote remove MKL_HOME from the example and stay with MKL_ROOT. I think we should be like HDF5 and BOOST who have better find scripts (I think) than than FFTW and LIBXML2.  

Like us findBOOST supports BOOST_HOME, BOOST_ROOT, and BOOSTROOT

I prefer to define everything in the cmake command. But then I'm building many versions and like to be able to capture shell history and not worry about too many enviroment variables. This is not the user situation.

prckent:
My vote is also now for the consistency argument - MKL_ROOT as the primary spec, as per HDF5 and BOOST. We can silently support MKL_HOME for backwards compatibility, but lets advertise MKL_ROOT as the preferred way.

PR #939, opened 2018-07-29
Remove neighbourlist in distance tables

ye-luo:
The neighbour lists carried by distance tables are removed. The only consumer, NLPP evaluation, is updated with an interchange of ion and electron loop.

The neighbour list is supposed to be global. For example, the neighbour list for the electrons near an ion within a cutoff can be used for J1, J3, NLPP, Coulomb. However, due to the need of different cutoffs, the neighbour list cutoff needs to be chosen the maximal value and then filtering is applied again per use. So this actually becomes not efficient. Due to the need of extra memory and consistency maintaining, the neighbour list becomes unfavourable. Thus removed.

### DIFF
--- a/src/Particle/AsymmetricDistanceTableData.h
+++ b/src/Particle/AsymmetricDistanceTableData.h
@@ -172,7 +172,7 @@ struct AsymmetricDTD
   }
 
   ///not so useful inline but who knows
-  inline void evaluate(ParticleSet& P, bool update_neighbor_list)
+  inline void evaluate(ParticleSet& P)
   {
     const int ns=N[SourceIndex];
     const int nt=N[VisitorIndex];
@@ -186,7 +186,7 @@ struct AsymmetricDTD
   {
     APP_ABORT("  No need to call AsymmetricDTD::evaluate(ParticleSet& P, int jat)");
     //based on full evaluation. Only compute it if jat==0
-    if(jat==0) evaluate(P,true);
+    if(jat==0) evaluate(P);
   }
 
   ///evaluate the temporary pair relations
--- a/src/Particle/DistanceTableAA.cpp
+++ b/src/Particle/DistanceTableAA.cpp
@@ -86,7 +86,6 @@ DistanceTableData* createDistanceTable(ParticleSet& s, int dt_type)
           dt = new  SymmetricDTD<RealType,DIM,PPPS>(s,s);
         }
       }
-      o << "    Setting Rmax = " << s.Lattice.SimulationCellRadius << std::endl;
     }
   }
   else if(sc == SUPERCELL_SLAB)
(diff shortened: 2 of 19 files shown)


jtkrogel:
I would have thought that reconstructing a neighbor list for each use separately (J1, J3, NLPP, etc) would have been more expensive, with the added expense growing in the large N limit.  The added check due to cutoff differences is always O(N) rather than O(N^2).  The fact that it is not used apart from NLPP is actually a better argument for the need to generalize its use rather than its removal.

On related note, I am currently implementing a potential that uses the same cutoff as NLPP, so I would very much like to reuse rather than recompute the O(N^2) neighbor list.

prckent:
I generally agree with this simplification, but lets make sure we have considered all the possibilities and are moving in the right direction.

We are optimizing for N from 10-10000

- For small N, any recomputation can be done in cache, and is not very relevant to this discussion
- A global N^2 neighbor list is not scalable with large N
- A "per use" neighbor list benefits from applying unique cutoffs, as you describe
- However, we do lose some reuse possibilities e.g. common one body electron-ion and three-body electron-ion cutoffs, or at least using e.g. a larger electron-ion one-body cutoff to prescreen the three-body term. (I don't believe we do this currently).

Questions: 

- What is the speed of the full range of NiO supercells after this PR? If the logic is correct, the largest should be faster and the smallest little different.
- What is the actual % of time spend doing the neighbor computation versus cell size today?
- We have not discussed (at least on GitHub) adding any kind of Verlet list, or similar hierarchical method to avoid recomputing all the distances each particle move. This has formally better scaling, but one could argue our particle counts are too small. Is there any real citable data  or a  review article that proves this is the case? This is something we might considering introducing later, but if there have been discussions & consideration given we should capture some of that here.

jtkrogel:
@prckent I don't follow your second point: "A global N^2 neighbor list is not scalable with large N".  A neighbor list within a cutoff/cell table is both O(N) in memory and O(N) in use.  Am I missing something?

ye-luo:
Performance wise. I saw 0.5% faster on BGQ with 1000 electrons and NiO 512 atom cell less of 1% change. The NiO case saves 1GB memory.

jtkrogel:
So it is the full neighbor table rather than a neighbor list that uses distance cutoffs.  Fine with me if we remove it.

I guess we will just do full N^2 for short-ranged potential terms until this shows up on a computational intensity profile.

prckent:
Is the following correct? I will update based on comments.

The full e-e neighbor list (N^2 cost) will be recomputed:

- Coulomb e-e potential
- Two body Jastrow e-e 

The electron-ion distances are recomputed (NM cost):

- Local electron-ion
- Non-local electron-ion
- One body jastrow
- Three body jastrow (full e-e is not needed because e-i is screened first)

prckent:
@ye-luo Particle by particle updates are the ideal textbook case for neighbor tables.

ye-luo:
There is no recompute in most cases.
The distance tables are updated on the fly during moves.
e-e table is used by
- Coulomb e-e
- Two body Jastrow e-e
- Three body Jastrow e-e

e-I table is used by
- Local electron-ion
- Non-local electron-ion
- One body Jastrow
- Three body Jastrow (cutoff input based on e-e, cutoff/2 is applied on e-I). A e-I neighbour list is carried internally.

Issue #861, opened 2018-05-11
Remaining tasks for AoS to SoA transition


ye-luo:
This is the final list of tasks that have priority before the remaining AoS code is fully removed from the source.  The AoS removal likely will occur sometime in Oct. 2019.  Even code listed as "priority" below will be removed once the deadline is exceeded if the assignee(s) have not ported by then and this issue will be closed.

All remaining AoS code has been encapsulated by if-defs.  The changes made in the following PR highlight all remaining AoS code:

https://github.com/QMCPACK/qmcpack/pull/1846/files

Priority code to port before the Oct. deadline:
- [x] g(r) (issue: #1699, assignee: @markdewing, PR: #2035)
- [x] Chiesa/Ceperley and bare force estimators (issue: #1795, assignee: , PR: )
- [x] energy density and energy density matrix estimators for EKT (issue: #1796, assignee: @jtkrogel, PR: #1869)
- [x] orbital optimization (issue: , assignee: @spinedaflores , PR: #1893)
- [x]  lattice deviation estimator (issue: , assignee: @Paul-St-Young, PR: #1938)
- [x] "Chiesa" S(k) kinetic energy correction (issue: #2025, assignee: @Hyeondeok-Shin @ye-luo, PR: #2050).
- [x] NLPP derivatives in optimization (issue: #2083, assignee: @ye-luo, PR #2179)

Code that is slated for deletion after the Oct. deadline (volunteers welcome!):
* complex HEG backflow (see #1227)
* spin-dependent electron-ion backflow
* AoS cusp correction code ~~(SoA code has problems, see #1388)~~
* FDLR wavefunction
* numerical orbitals
* stress estimator (see #970) 
* core polarization potential

Note: tests that pass for AoS but fail for SoA:
* long-diamondC_1x1x1_pp-vmc_sdj-1-16-flux (wavefunction error!)
* short-c_no-hf_vmc-1-16 (numerical orbitals)
* long-c_no-hf_vmc-1-16
* short-c_no-sj_dmc-1-16
* long-c_no-sj_dmc-1-16
* short-H2-FDLR-16-1 (FDLR wavefunction)
* short-H2-orb-opt-16-1 (orbital optimization)
* short-H4-orb-opt-dmc-1-16

PDoakORNL:
What about the optimizable wave function support?

I'm working on the CUDA code now.  
I've begun JastrowCUDAClass switch to the OrbitalBase base class.

[...]

PR #946, opened 2018-07-31
Rename base classes

ye-luo:
The following base classes are renamed to reflect the physics and their use and be concise.
OrbitalBase -> WaveFunctionComponent
OrbitalBuilderBase -> WaveFunctionComponentBuilder
SPOSetBase -> SPOSet

As part of #861.


prckent:
I suggest we only merge after v3.5.0 is released, which should happen this week.

ye-luo:
If someone needs to update their branches currently diverged, there is no difference merge before or after 3.5 release.
If someone feels more comfortable to start with a released version, merge now is better than after.
Since we may produce 3.5.1. It has to be delayed even after that. Not good...
@prckent What is the main reason to merge after v3.5.0?

prckent:
We can flag the upcoming changes in the release notes.

ye-luo:
Why not just merge? It is much simpler.

PDoakORNL:
This is developer level change not visible from the user level correct? I don't see why we should delay.  It's certainly going to complicate continued work on my branch.

ye-luo:
8 builds {real,cplx} {AoS,SoA} {full precison, mixed precision} are good with BGClang and Intel compilers.

ye-luo:
CI checks CPU builds with GNU and CUDA AoS builds with GNU.
I manually checked CUDA SoA build real and complex.
Everything is fine.

ye-luo:
If we want to keep the history as linear as possible, this can only be merged after master is merged back to develop.

Issue #819, opened 2018-05-01
Status/future of CSRMC drivers


jtkrogel:
I noticed that the following files are not integrated with the mainline distribution in any way:
```
QMCDrivers/CorrelatedSampling/CSRMC.h
QMCDrivers/CorrelatedSampling/CSRMCUpdateAll.h
QMCDrivers/CorrelatedSampling/CSRMCUpdatePbyP.h
```

Are there any plans for these files?  If there are then we should keep them and otherwise remove them.

rcclay:
We should remove them.  They'll be added if/when they are implemented.

jtkrogel:
OK, will remove.

jtkrogel:
Resolved by #949

PR #949, opened 2018-08-01
Remove CSRMC files

jtkrogel:
Remove the correlated sampling reptation drivers per discussion with @rcclay in #819.


markdewing:
okay to test

PR #953, opened 2018-08-02
Fix BG/Q unit tests

### LINKED ISSUE #941: Unit test failing on BGQ
From last night,
test_unit_spline2 fails with mixed precision build and hangs with full precision build.
https://cdash.qmcpack.org/CDash/testDetails.php?test=2803958&build=26247
test_particle has kc = nan
https://cdash.qmcpack.org/CDash/testDetails.php?test=2803962&build=26247

markdewing:
For the particle test, choose better box vectors.  The previous vectors caused a condition where the real space cutoff became zero, and the inverse cutoff was inf.

For the spline test, the number of splines needs to be a multiple of 4 to ensure the alignment is correct in the BG/Q version of the spline evaluation code.
Also adjust the indexing on gradients and hessians to match how these are written in the evaluation code.

Increase the space for the laplacian - it needs 3 components during evaluation, even though only the first is used in the output.  This was found from running through valgrind.

Fixes #941


prckent:
How do we guarantee memory alignment in production runs on BG/Q and ensure multiples of 4 splines? Are we missing some assertions?

ye-luo:
It seems like spline2 can take any number of splines but actually it has many assumptions.
1. The spline coefficients must be aligned and of course num_spline padded.
2. The result vectors must be aligned and also have last dimension padded.
Since the result vectors are just internal storage, everything is taken care in SplineAdoptors. This requirement is only local and not propagated to up levels.

ye-luo:
@prckent I prefer to do the update separately. @markdewing still needs to check BGQ and update his genration script. I was not touching the generating but directly modifying the generated script.

markdewing:
I checked on BG/Q and it passes

PR #955, opened 2018-08-03
Change BG/Q toolchain so cmake need only run once.

markdewing:
Put CMAKE_*_FLAGS variables in the cache so they are available immediately to the restof the CMake scripts.

See
https://stackoverflow.com/questions/11423313/cmake-cross-compiling-c-flags-from-toolchain-file-ignored
for more information


ye-luo:
Do not update branch.

ye-luo:
Done.

PR #956, opened 2018-08-06
3D Ewald Breakup Scheme

rcclay:
This implements necessary functions for computation of long range sums via the Ewald method in 3D.  Purpose is to serve as a reference for debugging Natoli & Esler optimized breakup schemes for forces and stresses.  This currently works with energies and forces for **all-electron calculations**--the local channel in pseudopotential calculations is not handled correctly. Code exists for computing stresses, but that should be considered experimental until I can test more fully.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -232,6 +232,7 @@ ENDIF()
     LongRange/LPQHIBasis.cpp
     LongRange/LPQHISRCoulombBasis.cpp
     LongRange/EwaldHandler.cpp
+    #LongRange/EwaldHandler3D.cpp
     LongRange/LRCoulombSingleton.cpp
     )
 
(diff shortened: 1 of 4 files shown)


markdewing:
okay to test

markdewing:
It would good to add unit tests or hook this code into the existing unit tests - src/QMCHamiltonians/test/test_coulomb_pbcAA.cpp and test_coulomb_pbcAB.cpp

rcclay:
That's a good idea.  I'll see what I can do tomorrow.

ye-luo:
Just to help us understand the current situation.
Is LongRange/EwaldHandler.cpp the current one in use?
What is the difference comparing it with the one you added?

The added code seems not connected to the main code. It may break without notice when we modify the codes. If we don't want it to be directly connected, connecting it via unit test is a good idea.

rcclay:
@ye-luo EwaldHandler.cpp is what already exists, appears to be for quasi periodic boundary conditions, I don't think it actually does the 3D bulk ewald breakup.  Also, in line with yours and Mark's suggestions, I added unit tests for both energy and force evaluations using EwaldHandler3D.

rcclay:
What's the error?  I can't access the build system logs.

markdewing:
It should send you email with the failed log.   I think it's a unit test failure, which the logs don't give output for.

rcclay:
Ah.  I got it.  That unit test failure is mighty laconic.

rcclay:
OK!  Unit tests fixed, everything is passing, and I've reverted the changes to the Rhea build script.

PR #959, opened 2018-08-08
Reduce SPOSet::evaluateX interfaces

ye-luo:


PDoakORNL:
I see new test failures on OSX with AoS CPU but not with SOA GPU on oxygen. Stay tuned.

ye-luo:
unit_test_hamiltonian failing is unrelated. I added an issue.

PDoakORNL:
Both OSX test failures are regressions from before this PR.

PDoakORNL:
I will now be out till early afternoon. At which point I'll continue, someone is welcome to take over if they have time.

ye-luo:
Although the OSX problem may surface some hidden issue, I don't want it to slow down other activities in general and we can put it aside.

PR #961, opened 2018-08-08
Eliminate ion-ion force computation from ForceBase

rcclay:
Ion-Ion force computation should be done in the force estimators, and not in ForceBase.  In addition to not handling boundary conditions appropriately, this piece of code also causes QMCPACK to segfault since the ion-ion distance table has transitioned to SoA format.


markdewing:
okay to test

PR #963, opened 2018-08-09
Add check_stats script for testing estimators

jtkrogel:
Add a new standalone script (check_stats.py) to be used to implement estimator tests in the ctest framework.  The script is meant to be general purpose for stat.h5 files much in the same way check_scalars.py is for scalar.dat files.

To facilitate testing, new scalar-like streams of data are generated from the underlying estimator data: 1) full sums of estimator data per block, 2) a user specified number of partial sums of estimator data per block.  The statistical mean and error bar is then computed for each scalar-like stream.

Multiple types of tests are performed:   1) Check that full sum means are within nsigma of reference means, 2) Similar checks for partial sum means, but with nsigma correctly adjusted to account for the number of partial sums so that the overall test of partial sums statistically fails according to the user inputted nsigma, 3) For quantities that obey sum rules (e.g. the density), check that the full sum matches the reference value for each block separately, not just for the mean.

Reference data is contained in short data files.  The script is capable of generating reference data from a long reference run, writing out multiple reference files according to the type of tests desired (e.g. short and long), and checking test run against the reference.  

The script is self-documenting (see help); a single example is shown in the comments below.  If further documentation is desired, it will be included in subsequent PR's containing the first tests.  PR's containing tests of the density, spin density, and energy density will follow.

### DIFF
--- a/utils/check_stats.py
+++ b/utils/check_stats.py
@@ -1,0 +1,2304 @@
+#! /usr/bin/env python
+
+from __future__ import print_function
+
+# Statical error checking code for use by testing framework (stat.h5 files)
+# Jaron Krogel/ORNL
+
+
+# check_stats.py packages obj and HDFreader classes from Nexus.
+#   Note that h5py is required (which depends on numpy).
+
+
+
+######################################################################
+# from generic.py
+######################################################################
+
+import sys
+import traceback
+from copy import deepcopy
+import cPickle
+from random import randint
+
+
+class generic_settings:
+    devlog      = sys.stdout
+    raise_error = False
+#end class generic_settings
+
+
+class NexusError(Exception):
+    None
+#end class NexusError
+
+
+exit_call = sys.exit
+
+
+def nocopy(value):
+    return value
+#end def nocopy
+
+
+
+def log(*items,**kwargs):
+    indent=None
+    logfile=generic_settings.devlog
+    if len(kwargs)>0:
+        n=0
+        if 'indent' in kwargs:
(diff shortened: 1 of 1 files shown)


prckent:
OK to test

prckent:
Did you check nsigma=0 and single datapoint cases? (often a cause of division by zero issues)

jtkrogel:
nsigma should always be around 3.0 or greater (success probability >99%).  By this do you mean quantities that have an errorbar of zero?

What do you mean by single datapoint?

prckent:
If the data is exact or not changing, what happens? Single data point = single blocks or whatever the smallest quantum of data is.

jtkrogel:
In cases where the error bar is vanishingly small on the reference data, the error bar for comparison is set to 1e-12 times the expected mean value (see first 6 lines of check_mean function).  This is a tight error bound and results in a pass if the test data are correct.  In the comments above, "   Testing quantity: d full sum" is just such a case and it passes correctly.

jtkrogel:
Estimators writing to stat.h5 instead of scalar.dat generally have many values (e.g. the density) so I haven't checked the 1 value limit as this is not the target of the script (but it is covered already by check_scalars.py). 

If the number of blocks successfully completed by the test run is very small (or zero) it is possible that e.g. the data read/processing will fail in python.  This case is explicitly guarded by try-except.  If the python script fails for any reason, and exit_fail is called indicating test failure.  If the script does run successfully in these cases, it is unlikely that the checks performed on the limited data will pass the 3 sigma tests.

Issue #960, opened 2018-08-08
1RDM needs resurrecting


jtkrogel:
1RDM currently writes all zeroes into the stat.h5 file.  Functionality worked as of SVN revision 6361.  Will track down and repair existing HEG examples from that era.

Once repair is complete, efficiency of 1RDM ratio updates will need to be improved and validity of output checked for large systems.  New issues will be generated for these points once the current one is resolved.

jtkrogel:
Resolved by #967.

See issues #973, #974, #975, #976, and #977 for further needed developments relating to the 1RDM.

PR #967, opened 2018-08-09
Workaround for 1RDM zero problem

jtkrogel:
This PR fixes the manifesting problem with the 1RDM, namely that zeros are populated in the stat.h5 file output.

The overarching issue is a currently unresolved bug in the flow of Traces logic, as it applies to the special case of Traces being used to stream e.g. energetic information to estimators.  Even when the streaming of data is requested, this currently does not happen.

Since this data is only strictly necessary for the energy density matrix (1REDM), a workaround is put in place to allow computation of the 1RDM when the 1REDM is turned off via the input file.

Following this PR, the 1RDM works as originally implemented and is appropriately diagonal for a system w/o a Jastrow (tested explicitly for a small diamond cell).  Further correctness tests and efficiency improvements remain to be made.

A full fix for the data streaming issue will follow the pending implementation of tests for the energy density, 1REDM, and traces.

Input for a working variant of the diamond test case (tests/solids/diamondC_1x1x1_pp) is attached (see [qmc.zip](https://github.com/QMCPACK/qmcpack/files/2275699/qmc.zip)).

### DIFF
--- a/src/QMCHamiltonians/DensityMatrices1B.cpp
+++ b/src/QMCHamiltonians/DensityMatrices1B.cpp
@@ -453,12 +453,14 @@ namespace qmcplusplus
       for(int s=0;s<nspecies;++s)
       {
         out<<pad<<"  matrices/vectors for species "<<s<< std::endl;
-        out<<pad<<"    E_N        : "<< E_N[s]->size()<< std::endl;
+        if(energy_mat)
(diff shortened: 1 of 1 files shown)


jtkrogel:
Addresses #960.

prckent:
OK to test

PR #969, opened 2018-08-09
Fix incorrect sign in LR Breakup

rcclay:
The Esler style long-range breakup has the wrong sign for LRHandlerTemp::evaluateLR(r), but the error is compensated by an incorrect sign when computing the LocalECP in CoulombPBCAB.cpp.  LocalECP energy now correct when using Ewald or Esler breakups.


prckent:
OK to test

PR #987, opened 2018-08-13
Improve JeeI::acceptMove algorithm.

ye-luo:
The cost of updating the neighbour list was N^2, now N.


ye-luo:
I added a bit comment around the lines of code I touched.
I can't document the full algorithm in the source code. It is too much.
I'm completing a paper giving the full algorithm.

PR #991, opened 2018-08-13
Add (real space) derivative of long-range component to LR Handlers.  

rcclay:
When constructing and computing forces associated with local ECP channels, derivatives of the following expression are required:  d/dr(vlocal(r) - r*LR(r)).  This PR adds convenience functions lrDf(r) (like srDf, which is the derivative of the short-range part) to help in computing this.  APP_ABORT is called if the function is not implemented.

### DIFF
--- a/src/LongRange/EwaldHandler.h
+++ b/src/LongRange/EwaldHandler.h
@@ -61,6 +61,7 @@ public:
   EwaldHandler(ParticleSet& ref, mRealType kc_in=-1.0)
     : LRHandlerBase(kc_in)
   {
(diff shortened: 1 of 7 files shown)


ye-luo:
Please add a protected string variable ClassName in the base class and overwrite it by the class name in each derived class constructor. Add the abort implementation to the virtual function of base class with the aborting message containing the ClassName. Then there is no need to put abort in every derived class.

rcclay:
Good suggestion.  Done.

ye-luo:
Okay to test.

Issue #979, opened 2018-08-10
Test for spin density estimator needed


jtkrogel:
Spin density estimator does not yet have a test.

jtkrogel:
Resolved by #992.

PR #992, opened 2018-08-13
Add density and spin density tests

### LINKED ISSUE #978: Test for density estimator needed
Density estimator does not yet have a test.

jtkrogel:
Modify SIMPLE_RUN_AND_CHECK in CMake/macros.cmake to accommodate arbitrary command line inputs to the checking script.

Add tests for the density and spin density estimators based on full and partial sums checked with check_stats.py.  Spin density tests are added for the following cases: VMC+PBC (diamond), DMC+PBC (diamond), VMC+open BC's (LiH dimer).  Density tests are added for the following cases: VMC+PBC (diamond), DMC+PBC (diamond).

All reference data was generated with SVN revision 6259.  These tests catch the current breakage regarding the incorrect weight in the first block of DMC:
```
  Fixed sum per block tests:
    d 0 3.99999792984!=4.0
    u 0 3.99999792984!=4.0
    some per block sums do not match the reference
    status of this test: fail
```

README files have been updated in each test directory.

Resolves #978 and #979.


ye-luo:
Can we move all the added tests to tests/estimators?
Can we move the check scripts from utils to tests/scripts?
Can we have test names as estimator-XXX instead of short-XXX?
short are too heavy now.

jtkrogel:
It seemed more logical to keep the order by system.  For example, we don't have directories for J1, J2, MD, etc, but just diamond.  The tests are based on diamond and LiH, so I am following the original organization of the test files.

What we really need is a listing of all tests documenting what they test and where they are.  These are needs separate from this PR.

jtkrogel:
The short VMC density tests take 5 seconds, while DMC tests take 80 seconds on a 6 year old workstation.  These are not unreasonable runtimes for short tests and they should be kept in the short battery.

jtkrogel:
check_stats.py was placed next to check_scalars.py in a prior approved PR because they have the same general use and function.  Relocating them is fine, but a separate PR or issue should be made for this.

ye-luo:
Currently. when tests are generated in the build folder, the whole folder is copied. Then it becomes very hard to find the actual used input among all the input files. This is the main reason I prefer not to add more in the current testing folders.

Organize tests by feature is fragmentary. We won't do J1/diamond, J2/LiH.

For estimators, the goal is to have a single test running as many estimator as possible instead of one estimator at each test run. At least all the estimators based on particle coordinates is one set of run and all the 1RDM like tests is another. The number of systems used in estimator test will be very limited.

Relocating check_stats.py and check_scalars.py needs a separate PR.

ye-luo:
I'm not saying the test you added is too long. Just I want all the estimator tests in a different category.

jtkrogel:
I'm not sure what directory structure you are proposing for the estimator tests.  Please can you elaborate how this should look when I have N estimators running on M target systems?

I'm also confused about packing many estimators onto a single run.  I don't really mind doing this, but it runs counter to prior requests I have received to make one test for one feature so that causes of failure are readily identifiable.  We can't have it both ways.

A second complication is that we don't yet have a way to add many tests onto a single executable run where each test involves running a script w/ an arbitrary number of command line inputs.  I'm not sure how to do this yet.  Perhaps @markdewing has ideas.

Finally, are these issues we are identifying with the test system generally that need to be discussed, planned, and then implemented across the test system generally?  I'm on a short timeline trying to add these tests before I lose the ability to easily create them based on old versions of the code (my favored test machine is vanishing suddenly Oct. 1 due to age).  I would prefer to be able to add these tests so that we have them, and work on higher level/general issues once we have had time to settle on what we really want.

prckent:
"For estimators, the goal is to have a single test running as many estimator as possible instead of one estimator at each test run." 

I am less sure about this. We need to be able to run on GPUs so this procedure won't work unless e.g. the estimator implementations are updated to skip and not abort.

For now, one or multiple sets of tests for individual estimators seems OK.

ye-luo:
@prckent Important point on how to handle missing implementation.
Then let us put the tests in the per system folder.

jtkrogel:
Why not keep "short" and simply add "estimator"?  It is still useful to mark all tests that complete in the 30s regime.  

I suppose we can standardize markup for Jastrow terms ("J1bs", "J2bs", "J3pl"), determinants ("SD", "MD", "SDBF", "MDBF") and orbitals ("bso", "go", "pwo") and so forth to achieve topical markup in addition to just the estimators.

prckent:
I agree with keeping short and long add adding estimator. The tests can then be filtered/selected and will also be compatible with all the existing automated testing scripts.

ye-luo:
@jtkrogel agree we should standardize naming.

prckent:
@ye-luo So only the test scripts need to move before merging?

jtkrogel:
On second thought, tests are sufficiently modular that also relocating check_scalars.py should be simple.  Doing that now.

ye-luo:
OK to test

jtkrogel:
I believe all concerns are resolved now, correct?

CMake/macros.cmake
```diff
@@ -243,6 +243,8 @@ function(SIMPLE_RUN_AND_CHECK base_name base_dir input_file procs threads check_
     set(check_cmd "${CMAKE_CURRENT_BINARY_DIR}/${full_name}/${check_script}")
   elseif(EXISTS "${CMAKE_SOURCE_DIR}/tests/scripts/${check_script}")
     set(check_cmd "${CMAKE_SOURCE_DIR}/tests/scripts/${check_script}")
+  elseif(EXISTS "${CMAKE_SOURCE_DIR}/utils/${check_script}")
```
ye-luo: Can we now avoid these added lines?
Apart from this, everything is good to go.

PR #994, opened 2018-08-15
Add 1RDM tests

### LINKED ISSUE #973: Test for 1RDM needed
A test is needed for the 1RDM.  Since this has newly been revived (see PR #967) and there is now a route identified to make stat.h5 based tests (see PR #963) this is now possible.

A good candidate for the test is the existing diamond test case.

jtkrogel:
Add short/long cpu tests for 1RDM.  These are not enabled for GPU (should they be?).

Reference data for real and complex builds is necessarily different.  This is because the B-spline orbitals are orthogonal in the complex code, but they are not in the real code.  Following diagonalization of the 1RDM, this results in no difference in the natural orbitals.
```
1rdm real
    1.000  0.     0.     0.    
    0.     1.000 -0.367  0.    
    0.    -0.367  1.000  0.    
    0.     0.     0.     1.000 

1rdm complex
    1.000  0.     0.     0.    
    0.     1.000  0.     0.    
    0.     0.     1.000  0.    
    0.     0.     0.     1.000 
```
Creation of the tests revealed that cloning is broken for DensityMatrices1B in the current code.  The tests pass with a 16 mpi, 1 thread configuration, but fail for 4 mpi and 4 threads.

Resolves #973.  New issue created for cloning bug, see #995.


jtkrogel:
Forgot to update README, commit below fixes this.

prckent:
Please can you see what happens with the CUDA path enabled? How much work might it be to enable? Generally we enable platform specific feature tests when there is a plan to enable them "soon"

jtkrogel:
Will try.

jtkrogel:
On the GPU side, there is no "addEnergy" function implemented (the equivalent of evaluate):
"189: ERROR Need specialization for DensityMatrices::addEnergy(MCWalkerConfiguration &W)."

markdewing:
okay to test

prckent:
Re: GPU. DO you know which class/classes don't have implementation? The abort is from the base class.

Hopefully addEnergy only needs a ratio

jtkrogel:
It is the DensityMatrices class itself that is missing the implementation.

PR #1008, opened 2018-08-16
Merge H5 Parser.

anbenali:
All LCAO codes generate same HDF5 file.  Therefore renamed the PyscfParser to LCAOH5Parser.  

in convert4qmc the tag -pyscf or -orbitals will call the same parser.


markdewing:
The LCAOH5Parser files are missing from the commit?

anbenali:
Of course files were missing... Otherwise how can I be an idiot...

PR #1009, opened 2018-08-16
Backend work for LocalECP Forces

rcclay:
This is PR 1 of 4 for getting Hellman-Feynman forces with pseudopotentials into QMCPACK. This PR focuses on just the LocalECP contribution to the force.  The second PR will focus on implementing the necessary backend stuff for nonlocalPP forces.  The third PR will handle making these capabilities available to the user and adding unit tests.  The fourth will involve implementing the required LR breakup routines into the default Esler style breakup used in QMCPACK.  

All of this is implemented in CoulombPBCAB, and is computed when a "ComputeForces" flag is set to true.  Primary changes are the addition of long range handlers to deal with the forces, and implementing the correct computation of the short-range force component when pseudopotentials are defined.  This works only with the SoA build, but has not been optimized for SoA data layout.  This is indicated in the comments.  

Tests of this were done using the following geometry (Li Z=1 atoms), in periodic boundary conditions, with the EwaldHandler3D breakup (LR_dim_cutoff=75).  

> CELL_PARAMETERS {alat}
6.95218 6.95218 0
0 6.95218 6.95218
6.95218 0 6.95218

> ATOMIC_POSITIONS {alat}
Li   1.0428270   0.9037834   0.5561744
Li   3.4760900   3.4760900   0.0000000
Li   0.0000000   3.4760900   3.4760900
Li   3.4760900   6.9521800   3.4760900
Li   3.4760900   0.0000000   3.4760900
Li   6.9521800   3.4760900   3.4760900
Li   3.4760900   3.4760900   6.9521800
Li   6.9521800   6.9521800   6.9521800

Here's what we get for the forces.  First column, first index is the ionID, second is the cartesian index.  Second column is what QMCPACK is spitting out.  Third is what QE spits out for the local force, fourth is the difference between these two, fifth is the QMC force error, and sixth is the Z-score.  Basically, looks good to me.  

> 

  | QMC | DFT | DIFF | ERR | Z
-- | -- | -- | -- | -- | --
Flocal_0_0 | 0.03911 | 0.03933 | -0.00022 | 0.00014 | -1.58925
Flocal_0_1 | 0.03228 | 0.03224 | 0.00004 | 0.00014 | 0.31743
Flocal_0_2 | 0.01673 | 0.01677 | -0.00004 | 0.00019 | -0.23450
Flocal_1_0 | -0.03067 | -0.03077 | 0.00010 | 0.00015 | 0.65983
Flocal_1_1 | -0.03208 | -0.03223 | 0.00015 | 0.00015 | 1.00453
Flocal_1_2 | 0.01331 | 0.01330 | 0.00001 | 0.00015 | 0.06103
Flocal_2_0 | 0.01809 | 0.01808 | 0.00001 | 0.00016 | 0.03409
Flocal_2_1 | -0.02411 | -0.02428 | 0.00017 | 0.00015 | 1.12233
Flocal_2_2 | -0.02727 | -0.02740 | 0.00013 | 0.00014 | 0.94479
Flocal_3_0 | -0.00150 | -0.00167 | 0.00017 | 0.00015 | 1.14960
Flocal_3_1 | 0.00691 | 0.00700 | -0.00009 | 0.00014 | -0.67593
Flocal_3_2 | 0.01556 | 0.01567 | -0.00011 | 0.00014 | -0.77993
Flocal_4_0 | -0.02503 | -0.02528 | 0.00025 | 0.00019 | 1.31950
Flocal_4_1 | 0.01721 | 0.01718 | 0.00003 | 0.00014 | 0.18500
Flocal_4_2 | -0.02997 | -0.02989 | -0.00008 | 0.00015 | -0.51850
Flocal_5_0 | 0.00786 | 0.00798 | -0.00012 | 0.00014 | -0.85721
Flocal_5_1 | -0.00014 | -0.00014 | 0.00000 | 0.00017 | 0.02385
Flocal_5_2 | 0.01247 | 0.01244 | 0.00003 | 0.00013 | 0.20054
Flocal_6_0 | 0.00621 | 0.00611 | 0.00010 | 0.00013 | 0.76454
Flocal_6_1 | 0.01066 | 0.01085 | -0.00019 | 0.00013 | -1.44996
Flocal_6_2 | 0.00289 | 0.00309 | -0.00020 | 0.00017 | -1.19641
Flocal_7_0 | -0.01105 | -0.01101 | -0.00004 | 0.00013 | -0.30827
Flocal_7_1 | -0.00831 | -0.00838 | 0.00007 | 0.00016 | 0.41609
Flocal_7_2 | -0.00308 | -0.00326 | 0.00018 | 0.00013 | 1.41277


src/LongRange/LRCoulombSingleton.h
```diff
@@ -57,6 +57,8 @@ struct LRCoulombSingleton
    */
   static RadFunctorType* createSpline4RbyVs(LRHandlerType* aLR, mRealType rcut,
       GridType* agrid=0);
+  static RadFunctorType* createSpline4RbyVsDeriv(LRHandlerType* aLR, mRealType rcut,
```
ye-luo: Could you add a comment for this function?

src/QMCHamiltonians/CoulombPBCAB.cpp
```diff
@@ -633,39 +767,57 @@ CoulombPBCAB::evalLRwithForces(ParticleSet& P)
...
-      esum += Qat[jat]*d_ab.rinv(nn)*rV;
-      forces[iat] += Zat[iat]*Qat[jat] * //g(d_ab.r(nn)) *
-                     (d_rV_dr - V)*d_ab.rinv(nn) *drhat;
+      const RealType* restrict dist=d_ab.Distances[b];
+      RowContainerType dr = d_ab.Displacements[b];
+      mRealType esum=czero;
+      for(size_t a=0; a<NptclA; ++a)
+      {
+        //Low hanging SIMD fruit here.  See J1/J2 grad computation.
+        //Too lazy for now.
```
ye-luo: This is good to leave hints/hooks in the code. In this place, doing SIMD is non trivial. No problem leaving such message and we can always revisit later.

src/QMCHamiltonians/CoulombPBCAB.h
```diff
@@ -82,6 +90,13 @@ struct CoulombPBCAB: public QMCHamiltonianBase, public ForceBase
   std::vector<RadFunctorType*> Vat;
   ///Short-range potential for each species
   std::vector<RadFunctorType*> Vspec;
+  ///Short-range potential (r*V) and potential derivative d/dr(rV) derivative for each ion
```
ye-luo: General comment: we should avoid pointers if they can be handled cleanly and then makeClone can be simplified or removed. Let us discuss in the next re-factorization WG.

src/QMCHamiltonians/CoulombPBCAB.cpp
```diff
@@ -569,6 +625,26 @@ void CoulombPBCAB::initBreakup(ParticleSet& P)
...
+  {
+    dAB = LRCoulombSingleton::getDerivHandler(P);
+    if(fV0==0)
+      fV0=LRCoulombSingleton::createSpline4RbyVs(dAB,myRcut,myGrid);
+    if(dfV0==0)
+      dfV0=LRCoulombSingleton::createSpline4RbyVsDeriv(dAB,myRcut,myGrid);
+    if(fVat.size())
+    {
+      app_log() << "  fVat is not empty.  Something is wrong" << std::endl;
+      OHMMS::Controller->abort();
```
ye-luo: Use APP_ABORT. If a non root rank hit here. Root doesn't have a chance to print.

src/LongRange/LRCoulombSingleton.h
rcclay: Got it!

src/QMCHamiltonians/CoulombPBCAB.h
rcclay: Agreed.

src/QMCHamiltonians/CoulombPBCAB.cpp
rcclay: Done.

rcclay:
Added some comments and got rid of the OHMMS aborts.  Hopefully this is better.  

@prckent I have a two atom test (bcc H for now, will probably become Li or C later) with reference values from QE.  I don't want to add it just yet, because there's still a bit of plumbing that needs to be done to make the unit test easy.  As to the orbital question, we can totally use HF to generate the orbitals so there's absolutely no confusion as to what's being calculated.  However, since the e-I force is to first order derived from int V_ext(r) n(r), I think that as long as you run plane-wave DFT and VMC with no Jastrow, and as long as everything is converged between DFT and QMC, you should get the same e-I force contributions between the two because the electron densities are the same.  That's my hunch.

ye-luo:
Okay to test

src/LongRange/LRCoulombSingleton.cpp
```diff
@@ -271,11 +275,54 @@ createSpline4RbyVs_temp(LRHandlerBase* aLR, T rcut, LinearGrid<T>* agrid)
   return V0;
 }
 
+template<typename T>
+OneDimCubicSpline<T>*  
+createSpline4RbyVsDeriv_temp(LRHandlerBase* aLR, T rcut, LinearGrid<T>* agrid)
+{
+  typedef OneDimCubicSpline<T> func_type;
+  if(agrid==0)
```
markdewing: We should use nullptr for comparisons and assignments to pointers, rather than 0 or NULL. It makes the code clearer that a pointer is involved, and it can do better type checking.

rcclay:
Looks like this broke mixed precision.  Investigating.

src/QMCHamiltonians/CoulombPBCAB.cpp
```diff
@@ -62,12 +64,66 @@ QMCHamiltonianBase* CoulombPBCAB::makeClone(ParticleSet& qp, TrialWaveFunction&
...
+  {
+    //V0 is the same object stored in Vat.  Deleting this pointer should take care
+    //of everything all electron.
+    delete V0;
+    V0=0;
+  }
+
+  for(int ig=0; ig<Vspec.size(); ig++)
+  {
+    if(Vspec[ig]) //verbose, but don't want to delete a null pointer.
```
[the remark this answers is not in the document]
rcclay: Awesome.  I will change this then.

markdewing:
Is there a reference for doing the forces for ECP's?

rcclay:
@markdewing Kind of.  I'm writing up everything I'm doing, but the Hellman-Feynman part is based loosely on this:  
https://journals.aps.org/pre/abstract/10.1103/PhysRevE.76.036707 
https://aip.scitation.org/doi/abs/10.1063/1.4948778

I can send you my notes if you want.  They are rough though.

rcclay:
@prckent Are these changes sufficient?

markdewing:
@rcclay  Thanks for the pointer.   
  I don't need to see the notes now.  As long as they end up in the manual (or someplace accessible) when you get them done.

prckent:
@rcclay Please comment/explain reason for
createSpline4RbyVsDeriv_temp
createSpline4RbyVs_temp

Also clean up the editorial
"Probably need to clean up"
"Too lazy for now."

rcclay:
@prckent Added a bit of explanation for why the createSpline functions exist in the header file, and what they calculate.  Also removed the editorial comments.

PR #1012, opened 2018-08-18
[WIP] General Numerical orbitals 

anbenali:
Converter: 
in the HDF5 format, converter requires No change at all. Input files can be genereated using the tag: 
convert4qmc -orbitals MyFile.h5

ParserFromCodeX:
Parser is still being refined. However, It should produce an identical H5 file as every other code (pyscf, QP or gamess). 
Main difference is located in : 
/basiset/atomicBasisSet{id}/basisGroup{id}/grid_name = "numerical"
which will then trigger reading 
/basiset/atomicBasisSet{id}/basisGroup{id}/basis_grid_xy = Matrix(npts_grid,2)

Compared to Gaussian Code, this corresponds to: 
/basiset/atomicBasisSet{id}/basisGroup{id}/grid_name = "Gaussian"
/basiset/atomicBasisSet{id}/basisGroup{id}/RadFunctions = (rad, exponent, contraction)

For accuracy testing and code correctness, initial implementation done with FHI-Aims for the trial wavefunction. Since the code is all electrons, I started with the AoS code. PR will have AoS and SoA implementation. 

AoS route: 

- Fixing the Spherical representation (sign issue) in AtomicBasisBuilder.h
- Function addRadialOrbitalH5(hin,nlms) in NGOBuilder.cpp  called from AtomicBasisBuilder.h will read the grid name. if numerical will call a new function: addNumericalH5 instead of addGaussianH5. 

-Creating addNumericalH5.() in NGOBuilder.cpp.  Function will read the grid, the number of points in the grid and will assign the grid (Matrix(npts,2)) directly to  radorb.  In Gaussian route, this was done through the class : 
Transform2GridFunctor<GaussianCombo<RealType>,RadialOrbitalType> transform(gset, *radorb);
Where:
for (int i=0;i<npts;i++)
    (*radorb)(i)=(guassian_fetchgrid(i)*pow(grid,inversepower);  // some function to put radial function into a loose grid ( see line 85 in Numerics/Transform2GridFunctor.h) 

Note that derivatives at the end of the grid where computed to force the grid to 0 at both ends. This is not necessary for numerical orbitals as the orbitals (and their derivatives cancel per construction at the edges. 
deriv = (deriv*pow(r,np)+static_cast<result_t>(np)*out_(0)*pow(r,np-1));

In the Numerical route, this is simply: 
for (int i=0;i<npts;i++)
    (*radorb)(i)=numGrid[i][1];

We then use 10D cubic splines for interpolation directly by calling 
radorb->spline(0,0,npts-1,0.0);

note that in this scheme, we never call  Transform2GridFunctor.h. 

===============
To do:
- Clean Aims Parser
- Generate AOS tests for correctness
- Port the code to SOA (will still not be usable due to the lack of Cusp Correction)

### DIFF
--- a/src/QMCWaveFunctions/MolecularOrbitals/AtomicBasisBuilder.h
+++ b/src/QMCWaveFunctions/MolecularOrbitals/AtomicBasisBuilder.h
@@ -111,10 +111,10 @@ bool AtomicBasisBuilder<RFB>::put(xmlNodePtr cur)
   else if(Morder == "pyscf")
   {
     expandlm = MOD_NATURAL_EXPAND;
-    addsignforM=tmp_addsignforM;
-    if(sph != "spherical") {
-      APP_ABORT(" Error: expandYlm='pyscf' only compatible with angular='spherical'. Aborting.\n");
-    }
+    addsignforM=1;
+    //if(sph != "spherical") 
+    //  APP_ABORT(" Error: expandYlm='pyscf' only compatible with angular='spherical'. Aborting.\n");
+    
   }
   if(sph == "cartesian" || Morder == "Gamess")
   {
@@ -156,11 +156,11 @@ bool AtomicBasisBuilder<RFB>::putH5(hdf_archive &hin)
   {
     expandlm = NATURAL_EXPAND;
   }
-  else if(Morder == "no" || Morder == "Numerical")
+  else if(Morder == "no")
   {
     expandlm = DONOT_EXPAND;
   }
-  else if(Morder == "pyscf")
+  else if(Morder == "pyscf" || Morder == "Numerical")
   {
     expandlm = MOD_NATURAL_EXPAND;
     addsignforM=tmp_addsignforM;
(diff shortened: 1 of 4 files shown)


jtkrogel:
I don't know how it should be done, but we might want to be careful with naming here to avoid confusion, e.g. we already have an atomicBasisSet with type "Numerical" that reads from an h5 file:
```
<wavefunction name="psi0" target="e">
  <determinantset type="MO" transform="yes" source="atom">
    <basisset>
      <atomicBasisSet type="Numerical" elementType="C" expandYlm="no" href="C.s000.h5">
```

markdewing:
okay to test

prckent:
As this WIP is developed, please add notes for an addition to the developer section of the manual detailing the HDF5 file format. There are at least 3 codes of this class.

prckent:
@jtkrogel @anbenali How would you prefer to update the labels? 

@anbenali if one label is Gaussian, for consistency the other should be Numerical.

ye-luo:
I'm working on cleaning up the input logic of SQD numerical basis set(need both xml and h5) and get ready for the other types of numerical orbitals (h5 only) on SoA code-path
1. the former will be treated as a special case and lives on the xml path. So the 'Numerical' key word will be used in the xml file.
2. the latter will be general and all the information lives inside h5 file.

src/QMCWaveFunctions/MolecularOrbitals/AtomicBasisBuilder.h
```diff
@@ -160,7 +160,7 @@ bool AtomicBasisBuilder<RFB>::putH5(hdf_archive &hin)
   {
     expandlm = DONOT_EXPAND;
   }
-  else if(Morder == "pyscf")
+  else if(Morder == "pyscf" || Morder == "Numerical")
```
ye-luo: Morder is code specific. It needs to be "fhiaims"

src/QMCWaveFunctions/MolecularOrbitals/NGOBuilder.cpp
```diff
@@ -288,17 +284,29 @@ NGOBuilder::addRadialOrbitalH5(hdf_archive & hin, const QuantumNumberType& nlms)
...
     return false;
   }
-  std::string radtype(m_infunctype);
+  std::string radtype;
+  if(hin.myComm->rank()==0){
+      hin.read(radtype,"grid_name");
+      if(radtype!="numerical")
+         radtype=m_infunctype;
+  }
+  hin.myComm->bcast(radtype);
```
ye-luo: This part is incompattble with the input spec used in SoA.

anbenali:
I don't agree. This should not be code specific anymore. As the name says
it should be just the order of the orbitals. I am planning to remove it and
have only spherical or cartesian or none (if that is a thing). Supporting a
new code should be independent from adding a specific tag inside qmcpack
but will have to be at the HDF5 level.

ye-luo:
There is another tag 'type' in atomicBasisSet not implemented in hdf5 but is used in xml. This is more suitable as Numerical. Morder(expandYlm) is a code specific tag with values gaussian, natural, pyscf, Gamess.

ye-luo:
The "input node" print out is not correct. It pretends to report loaded value but the tag names are variable names instead of xml tag names.

prckent:
This PR is nearly 11 month old. Should we close it?

PR #1014, opened 2018-08-19
LCAO SoA builder clean up

ye-luo:
I'm trying to re-enable the numerical orbitals (SQD generated) in LCAO SoA. Somehow I can't go further.
What is done so far:
Builders all have MPI enabled. No more need to access communicator via h5 handle.

What I'd like to do next:
Work with @anbenali to set up the h5 specs to maximize the reuse of routines.


ye-luo:
In this PR, the input spec is untouched. Merging will help interacting with other WIP branches.

anbenali:
I reached similar conclusion. I am at a point where I cannot load numerical
orbitals without serious gymnastics in the SOA code. This is not supposed
to be this way. This PR would prevent me to do weird things and have stupid
if statements and random reads from everywhere in the code...

prckent:
Anything that will reduce the amount of builder weirdness - all indicators that the design needs to be (re)thought.

PR #1015, opened 2018-08-20
Add const to all VirtualParticleSet use and separate SPOMem.

ye-luo:
When I first introduced the batched algorithm, SPOMem was shipped with VritualParticleSet. I had to make a compromise and const could not be applied to VritualParticleSet. This PR allows putting const.


ye-luo:
LCOrbitalSetOpt is from the initial orbital rotation attempt and needs to be cleaned up. Since this is on AoS, the least effort will be made.

prckent:
Please use the standard abort call. For example, for searchability.

PR #1016, opened 2018-08-20
Remove unused function implementations in WaveFunctionComponents

ye-luo:
Removed all unused APIs.


ye-luo:
The copyFrom is probably used by makeClone and maybe also copy constructor to save lines of code.
In the base class, it was commented but not removed. I didn't clean up copyFrom in backflow since it is not a good time to do cleaning in backflow but it should be cleaned up when we seriously touch backflow classes.

prckent:
This doesn't need cleaning up now, but clearly we need to note that copyFrom needs to be used delicately. copyFromBuffer is virtual void as expected.

PR #1021, opened 2018-08-20
Store cusp correction parameters in structure.

markdewing:
Using a structure rather than an array should make the code a little easier to understand.

Also add some headers to describe the columns in the newOrbs* output files.


markdewing:
I can add some comments to CuspCorrectionParameters

PR #1022, opened 2018-08-22
[WIP] Crystal converter

camelto2:
This is the beginning of a converter from Crystal to qmcpack. The converter is only currently interfaced to Crystal09, but will be updated to a more recent version (Crystal17) once I have access to it. The converter is set up to handle real and complex kpoints, but it currently will abort if there are any complex kpoints/orbitals. This will be removed once complex periodic gaussians are implemented. The converter parses the output from crystal, and generates the HDF5 file storing everything. Tests for diamond 1x1x1 and 2x1x1 cells have been created.

The hdf5 output and qmcpack input files are generated by running convert4qmc -crystal outputfilename, where outputfilename is the Crystal output file.

In order for crystal to output the required information, you must add to your crystal input file
PRINTOUT
EIGENALL
EIGENVEC
-999
END


ye-luo:
1. I saw diamond.out.xxxx 'out' is not needed.
2. again big file issue C.qmcapp.xml. Is it necessary to use a different PP than the BFD one? For testing purpose, I'd like have just one. So we can compare across even different codes.

prckent:
Thanks @camelto2 . Should this be a [WIP] work in progress or is there a reason that merging would be helpful? From your comments, I think a molecule would already be supported. How about adding an all electron or pp LiH molecule consistent with our GAMESS test?

camelto2:
@ye-luo the C.qmcpp.xml is actually a BFD potential, so I will link to the same one to avoid redundancy. I will also fix the *.out.* in the tests.

@prckent Changing it to a [WIP] is fine. I will add a molecular test as well

ye-luo:
There is C BFD living in tests/pseudopotentials_for_tests. You can just create a soft link.

ye-luo:
TODO: add a converter test.

grahamlopez:
test this please

grahamlopez:
test this please

ye-luo:
@camelto2 Could you resolve the conflict. Do you think if this PR is ready to be merged?

camelto2:
@ye-luo resolved the conflict. Only issue was the VSVB was removed since I made the WIP. As far as being ready to merge, I wouldn't say that it is. I think there has been some changes since I last worked on this with the periodic gaussian part of the code, so I need to make sure everything still works with the converter and the hdf5. Additionally, I believe there is a normalization issue for my orbitals coming from crystal, but I'll start looking at that to get it resolved.

camelto2:
I got this back up to date with develop. I need to run more thorough tests, but this seems to be correct for the gamma point only. The energies are off compared to QWalk when using non-gamma super_twists. Since gamma is ok, I think the orbitals and their ordering are all correct...so likely it is something to do with the definition of the non-zero supertwists and how that propagates into the evaluation of the wave function.

ye-luo:
> I got this back up to date with develop. I need to run more thorough tests, but this seems to be correct for the gamma point only. The energies are off compared to QWalk when using non-gamma super_twists. Since gamma is ok, I think the orbitals and their ordering are all correct...so likely it is something to do with the definition of the non-zero supertwists and how that propagates into the evaluation of the wave function.

Maybe add a hard abort in the converter saying non-gamma is not validated when it is detected.

prckent:
@camelto2 What are your thoughts on the future of this converter? If (say) only gamma and crystal09 are supported, perhaps merge that and error trap all other combinations? At least then we will have the functionality in mainline.

camelto2:
> @camelto2 What are your thoughts on the future of this converter? If (say) only gamma and crystal09 are supported, perhaps merge that and error trap all other combinations? At least then we will have the functionality in mainline.

I know that @rcclay has expressed interest in crystal for some force related calculations, so having the converter merged in some fashion is probably desirable.  I haven't looked at this in while, so I think there is probably a number of changes needed to be up to date with the current hdf5 format, but that should be quick to fix. My main problem is not having access to crystal anymore, so its hard to generate a number of tests without having someone else generate the output from crystal for me. 

molecules and gamma point solids should work fine. I can add a catch for that and the version so we can merge it when I add the tests.

prckent:
Thanks good point about forces. Clearly we keep this open and eventually merge when updated. We can look into Crystal access.

PR #1023, opened 2018-08-22
Update gold files for converter tests

markdewing:
PR #1008 introduced the /application group to the HDF files.  This updates the gold files to match.

This problem was harder to track down because h5diff prints no output when there are groups in one file
but not another.  The difference is visible with verbose (-v) output from h5diff.

Update the converter test script to output a message in this case.

(If this sort of failure comes up again, the script could be modified to always use -v with h5diff, or rerun h5diff with -v if the comparison shows differences)

Running h5dump and diff on the files showed this was the only difference
```
c1
< HDF5 "gold.orbs.h5" {
---
> HDF5 "test.orbs.h5" {
38a39,52
>          }
>       }
>    }
>    GROUP "application" {
>       DATASET "code" {
>          DATATYPE  H5T_STRING {
>             STRSIZE 7;
>             STRPAD H5T_STR_NULLTERM;
>             CSET H5T_CSET_ASCII;
>             CTYPE H5T_C_S1;
>          }
>          DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
>          DATA {
>          (0): "generic"
```


tests/converter/converter_test.py
```diff
@@ -93,6 +93,7 @@ def run_test(test_name, c4q_exe, h5diff_exe, conv_inp, gold_file, expect_fail, e
                       print("  pass")
                       return True
                    else:
+                      print("h5diff reported a difference")
```
PDoakORNL: But this does report now there is a difference and it can be then looked at with -v?

markdewing: This points out the failure is due to h5diff (vs. the current situation where the output is simply "FAIL")

PR #1024, opened 2018-08-22
Incorrect handling of phase info in ratioGrad

rcclay:
TrialWavefunction::ratioGrad should update PhaseDiff after computation of Psi(r')/Psi(r), and not PhaseValue.  This is also corroborated by use of PhaseDiff in TrialWavefunction::ratio().


rcclay:
During the course of debugging some force stuff, I swapped the ratio() in the nonlocalPP energy for ratioGrad().  I threw away the additional gradient, so other than being slower, it should have exactly reproduced what the current code does.  What I saw instead was a crazy jump in variance and energy.  I traced it back to the fact that Psi.getPhaseDiff() was identically 0, whereas when using ratio(), it would be multiples of Pi.

Issue #997, opened 2018-08-16
Density estimator test need protection for NoMPI


ye-luo:
https://cdash.qmcpack.org/CDash/viewConfigure.php?buildid=26979

jtkrogel:
I don't see why this is causing problems for USE_MPI=0 while others are not.  Shouldn't we just override the number of mpi tasks and omp threads in this case?  I.e. override mpi=4, omp=4 with omp=16 and allow the test to be added for the no mpi case?

prckent:
We should understand the problem and fix it. What is different about this case from other cases with no-mpi and the 4-4 combination?

ye-luo:
the cmake script generating tests find that you request 4 MPI but it can't provide in NoMPI case and a stop is triggered. This is expected behaviour and I mentioned earlier that only have 4x4 will not work with NoMPI.

prckent:
For the solids we are often checking 16x1, 1x16 and 4x4. RUN_QMC_APP defined in macros.cmake takes care to skip tests that don't make sense with no mpi / 1 mpi task.

jtkrogel:
OK, gpu needs 1x16, cpu needs 4x4, and no mpi needs 1x16.  We really only want one mpi/omp configuration per test (it would be overkill to run 1x16, 4x4, and 16x1 for every test), so I suggest we bury these specifications into QMC_RUN_AND_CHECK and SIMPLE_RUN_AND_CHECK, i.e.:
```
if(NOT USE_MPI OR QMC_CUDA)
  set(NMPI 1)
  set(NOMP 16)
else()
  set(NMPI 4)
  set(NOMP 4)
endif()
```
This way we have just one test run per hardware/mpi configuration in a correct way.

jtkrogel:
What gaps?  Why do we need to do 1x16 and 4x4 with MPI and no GPU?

ye-luo:
Since we are not building no MPI case on every platform, only 4x4 doesn't give me if it is broken in the threading or MPI. Testing both 1x16 and 4x4 will help us to tell which one is broken just looking at the nightly results. Running more tests with robot is preferred to manual examination.

jtkrogel:
I think what we are finding is that the CI is getting too heavy time wise.  Integration tests by nature only roughly tell you what is wrong, but there virtue is to cover a broad base of functionality.

[...]

PR #1025, opened 2018-08-22
SIMPLE_RUN_AND_CHECK skip tests needing MPI.

ye-luo:
address #997


jtkrogel:
Thanks @ye-luo.  This looks like a good way of doing it.

PR #1026, opened 2018-08-23
Remove legacy alternateXXX in TrialWaveFunction and WaveFunctionComponent.

ye-luo:
I didn't see any actual implementation in a derived WaveFunctionComponent.


prckent:
More good cleanup. Will wait to hear from @rcclay in case there is some use known/planned.  Could you determine the context in which these functions were added?

ye-luo:
We think they were added for the release node.

prckent:
A shame release node was "lost". Fortunately we have other ways of checking convergence now.

rcclay:
@prckent I can't remember these being used in recent memory, nor am I aware of plans to use them.

PR #1029, opened 2018-08-23
Start adding cusp correction pieces to SoA code.

markdewing:
Copy code for reading the cusp parameter files to the lcao/ directory.

Create routines for splitting out S-orbitals by modifying the MO coefficient matrix.
There are two of these routines.  One (splitEtaPhi) creates two wavefunctions - one with the
S-orbitals on a particular center, and the other wavefunction is everything else.  This is
used when generating the cusp correction.
(Routine was LCOrbitalSetWithCorrection::createLCOSets in AoS code.)

The other routine (removeSTypeOrbitals) removes S-orbitals on all centers and is the part
with no corrections.  This is used when evaluating the corrected wavefunction. The cusp
correction is evaluated separately and is added on.
(Routine was LCOrbitalSetWithCorrection::extractHighYLM in AoS code.)

In order to generalize this code between the two angular types (Cartesian and Spherical), some
methods have been added to find the S-type orbitals.  In particular, queryOrbitalForSType was
added to BasisSetBase.  Adding methods to base classes that are specific to a single
algorithm is less than ideal, but this seemed like the cleanest option.
Hopefully documenting which algorithm uses the method will help keep the confusion under control.

If a future algorithm should need to pick out specific orbitals, queryOrbitalsForSType could be converted to a more general query (eg, accept a functor or something to describe the conditions)

The new files (CuspCorrection.h, CuspCorrection.cpp, and test_soa_cusp_corr.cpp) have been run through clang-format.


src/QMCWaveFunctions/BasisSetBase.h
```diff
@@ -157,6 +157,10 @@ struct RealBasisSetBase
   virtual void setBasisSetSize(int nbs)=0;
   virtual void evaluateVGL(const ParticleSet& P, int iat, vgl_type& vgl)=0;
   virtual void evaluateV(const ParticleSet& P, int iat, value_type* restrict vals)=0;
+  virtual bool is_S_orbital(int mo_idx, int ao_idx) { return false;}
+
+  /// Determine which orbitals are S-type.  Used for cusp correction.
+  virtual void queryOrbitalsForSType(const std::vector<bool> &corrCenter, std::vector<bool> &is_s_orbital) {}
```
ye-luo: if possible, put it as a const member function.

ye-luo:
Is CuspCorrection a builder?
`LCOrbitalSetWithCorrection::createLCOSets` was really bad because it plays both OrbitalSet and builder. In SoA, we need to split the builder from SPOSet derived class. I/O is only allowed in the builder.

markdewing:
What do you mean about createLCOSets being a builder?  That method only modified existing LCOrbitalSet's.  Or do you mean the whole LCOrbitalSetWithCorrection class?

The CuspCorrection class is as big as it's going to be - it only computes the correction, and nothing else.  
 The plan is to create smaller classes than exist in the AoS version.

ye-luo:
You got my point. createLCOSets should live in the builder. LCOrbitalSetWithCorrection has a unfavoured design.

Issue #1028, opened 2018-08-23
Nexus example tests failing on oxygen and other systems


markdewing:
One example
 https://cdash.qmcpack.org/CDash/testDetails.php?test=2533958&build=27324

```
Test name     : example_gamess_H2O
Test exception: 
  example script failed to run
  attempted command: ./h2o_pp_hf.py --generate_only --sleep=0.01
  location: /scratch/pk7/QMCPACK_CI_BUILDS_DO_NOT_REMOVE/build_gcc_mkl_soa/nexus/tests/.nexus_test/user_examples/gamess/H2O
Test backtrace:
  File "/scratch/pk7/QMCPACK_CI_BUILDS_DO_NOT_REMOVE/qmcpack/nexus/executables/ntest", line 469, in run
    self.operation(**self.op_inputs)
  File "/scratch/pk7/QMCPACK_CI_BUILDS_DO_NOT_REMOVE/qmcpack/nexus/executables/ntest", line 1825, in user_examples
    nfail('example script failed to run\nattempted command: {0}\nlocation: {1}'.format(command,path))

Test status: fail
```

jtkrogel:
These tests pass when run manually.  The problem might be a failure to install Nexus on the test machine.

jtkrogel:
Perhaps this can be done by assigning PYTHONPATH explicitly in nexus/tests/CMakeLists.txt.

jtkrogel:
I will try to reproduce and fix.

markdewing:
In addition to fixing the tests, ntest could be updated to make the problem more obvious.  When the execute call (around line 1823) fails, the stdout and stderr from the failed call should be printed out.   If the problem is that Nexus not installed, that would show up as a failed import in the script output.

jtkrogel:
The problem appears to be fixed with #1083.  Nexus tests are passing on CDash on oxygen and on the other testing platforms I looked at.

Agree?

ye-luo:
Please wait for another night because it is hard to read the results from oxygen due to errors generated by compiler changes. Sunday oxygen results are very limited.

prckent:
Ignore all oxygen results for the time being. I am debugging changes in the scripts and compiler versions.

jtkrogel:
This is fully resolved now.  Closing.

PR #1031, opened 2018-08-24
Fix Nexus example tests

jtkrogel:
The example tests were failing because Nexus had not been "installed" on the hosting test environment.  This PR makes that install automatic when running ctest, but it does not alter the user environment (it is active only in ntest and its child proceses).  This is done by enforcing association between the ntest executable and its Nexus source by overriding the host PYTHONPATH variable explicitly via a new command line option to ntest.

Also added is a more informative failure message in the event that an example script fails.  Prior to this PR, the reported error message only mentioned that the example script had failed to run:
```
 Test name     : example_gamess_H2O
 Test exception: 
   example script failed to run
   attempted command: ./h2o_pp_hf.py --generate_only --sleep=0.01
   location: /home/j1k/repositories/qmcpack_fix/build_cpu_real/nexus/tests/.nexus_test/user_examples/gamess/H2O
```
This information is now expanded to include the stdout and stderr output from the example script:
```
 Test name     : example_gamess_H2O
 Test exception: 
   example script failed to run
   attempted command: ./h2o_pp_hf.py --generate_only --sleep=0.01
   location: /home/j1k/repositories/qmcpack_fix/build_cpu_real/nexus/tests/.nexus_test/user_examples/gamess/H2O
   script out:
     
   script err:
       File "./h2o_pp_hf.py", line 3, in <module>
         from nexus import settings,job,run_project
     ImportError: No module named nexus
```
In this case, it clearly shows that the cause of the problem is a failure to install Nexus.

Following this PR all Nexus tests pass on oxygen, even if the host environment variable PYTHONPATH is explicitly left blank.

Addresses #1028.


markdewing:
okay to test

Issue #1027, opened 2018-08-23
Tests using check_stats.py fail due to superstring import


markdewing:
For example,
   https://cdash.qmcpack.org/CDash/testDetails.php?test=2911172&build=27334

Error is 
```
Traceback (most recent call last):
  File "/scratch/pk7/QMCPACK_CI_BUILDS_DO_NOT_REMOVE/qmcpack/tests/scripts/check_stats.py", line 1056, in <module>
    from superstring import valid_variable_name
ImportError: No module named superstring
```

For this particular import, valid_variable_name is already defined, so deleting the import will fix it.
Running pyflakes shows a number other nexus imports that did not get inlined:
```
 from generic import obj
 from developer import DevBase, unavailable
 from debug import *
```
The definition of 'unavailable' is unavailable (and by transititivity, the Void class is also missing)

The 'unavailable' function is only used for to wrap the h5py import.

It looks like all the code is created by inlining from Nexus - is this done manually, or with a script?

The bigger question is whether it is important to mirror the Nexus code, or can the scripts diverge from Nexus?
If they can diverge, the easy fix for h5py is to stop using the 'unavailable' wrapper.  Dropping the try around the h5py import would probably be sufficient to cause the script to stop if h5py is not found.
The top of the script says h5py is required, so using the delayed availability check is not useful (and probably not even desired).

jtkrogel:
Thanks for catching this.  These manual imports were done too sloppily.  The scripts are intended to be self sufficient (no direct imports from Nexus) to be easier to use and not fragile to changes made in Nexus.  

The qmca script is this way already.  I will fix check_stats.py.

jtkrogel:
I believe this is resolved now.  Correct, @markdewing?

markdewing:
Yes, it appears to be fixed.

PR #1034, opened 2018-08-24
Fix check_stats script

jtkrogel:
Fix the check_stats.py script by removing dangling Nexus imports.  Previously the script would fail to run due to these imports being missing in the host environment.  With these imports removed, tests of stat.h5 quantities are correctly performed even if the PYTHONPATH is blank, consistent with check_stats.py being independent of the Nexus installation.

Addresses #1027.


markdewing:
okay to test

PR #1037, opened 2018-08-24
Clean up spline2 folder

ye-luo:
There were a few experimental version of MultiSpline evaluation routines.
It has been reduced to keep only the one in use.
Add a bit description and safeguard in MultiBspline.hpp


src/spline2/MultiBsplineStd.hpp
```diff
@@ -35,7 +35,7 @@ namespace qmcplusplus
   template<typename T>
     inline void 
     MultiBspline<T>::evaluate_vgl_impl(T x, T y, T z, 
-        T* restrict vals, T* restrict grads, T* restrict lapl, int first, int last, size_t out_offset) const
+        T* restrict vals, T* restrict grads, T* restrict lapl, size_t out_offset) const
```
markdewing: out_offset was changed to out_stride elsewhere

markdewing: Nevermind.  I see it really is an offset here.

ye-luo: Let me check. I probably changed one place but forgot another. What is a better out_offset or out_stride?

ye-luo: Changed back to out_offset.

markdewing: out_stride implies the fastest dimension, at least to me.  out_offset seems better, as it is the distance between the SoA elements.

PR #1038, opened 2018-08-25
Avoid increasing on-node population if there are no walkers on-node

ye-luo:
When the low global walker population protection is triggered, the old code assumes every node can boost the number of walkers but this is true only if there were alive walkers(ncopy_w.size()>0) before the load balance. I had a case with no alive walker on a particular node and triggered the safeguard.

This bug only affects very corner case. The usual reason of low walker population is due to a bad WF.


prckent:
Long term I think our load balancing code will have to be rethought. It has unnecessary complexities and subtleties.

PR #1039, opened 2018-08-27
Nexus: make import exceptions less restrictive

jtkrogel:
One reason the Nexus example tests are currently failing on oxygen is due to an uncaught import exception for matplotlib in structure.py.  The matplotlib import is protected by try-except, but only for RuntimeError and ImportError.  

Remove these restrictions so that exceptions are caught regardless of kind and the protections afforded by the unavailable function are in place (i.e. missing imports only cause execution to cease with a raised error if an unavailable imported name is actually accessed somewhere at runtime).  

The intended design is that Nexus should always run in a basic way, provided Python and NumPy are available.

### DIFF
--- a/nexus/library/graph.py
+++ b/nexus/library/graph.py
@@ -18,7 +18,7 @@ from generic import obj
 from developer import DevBase,unavailable
 try:
     from pydot import Dot,Node,Edge,Cluster as pdCluster
-except ImportError:
+except:
     Dot,Node,Edge,pdCluster = unavailable('Dot','Node','Edge','pdCluster')
 #end try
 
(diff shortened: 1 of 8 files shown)


ye-luo:
Has this fix been verified on Oxygen?

jtkrogel:
It needs to run through the CI system, i.e. on an account other than mine.  I would consider this an attempted fix as there might be other issues, but it is in direct response to the manifesting problem shown in last night's test logs.

ye-luo:
This is part of the nightlies not CI. Could you run through ctest -R nexus on oxygen?

jtkrogel:
Running via ctest on the command line passed both on Fri and today on oxygen.  The issue is a change in the local environment: on my user account, matplotlib is setup correctly while from the nightly environment, matplotlib is not available.  We need to wait until it runs tonight to know whether the issue has been resolved.

ye-luo:
I see. Thanks.

PR #1041, opened 2018-08-28
Connect cusp correction to wavefunction evaluation

markdewing:
Set up the correction in the builder.
Only write out the corrected orbital data files in debug mode.   These can be used to compare against the AoS version.

This code does not correctly handle molecules with multiple atoms of the same type (pay no attention to the branch name - I though about fixing this first, made the branch, changed my mind, forgot to change the branch name)
This will be fixed in a future commit.


src/QMCWaveFunctions/lcao/LCAOrbitalSet.h
```diff
@@ -48,6 +50,8 @@ namespace qmcplusplus
     ///Tempv(OrbitalSetSize) Tempv=C*Temp
     vgl_type Tempv; 
 
+    CuspCorrectionBase *cusp;
```
ye-luo: I don't feel right with this addition and all the ifs in the cpp. It is better to have LCAOrbitalSetWithCuspCorrection as a derived class of LCAOrbitalSet. In the derived class, call LCAOrbitalSet::evaluate() + cusp evaluation.

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
```diff
@@ -429,7 +433,102 @@ namespace qmcplusplus
     LCAOrbitalSet *lcos=new LCAOrbitalSet(myBasisSet,ReportLevel);
     loadMO(*lcos, cur);
 
-    //@TODO: add cusp condition
+    if (doCuspCorrection) {
```
ye-luo: This if is very long. Better to be a function.

ye-luo:
CuspCorrectionBase was there before you touch the code. Do you feel this abstraction necessary?

ye-luo:
Mixed precision question. Is the construction always in double?

markdewing:
CuspCorrectionBase doesn't seem like a useful abstraction.  I don't think there are going to be multiple implementations.  I'll remove it.

The cusp correction correction is done in RealType, so no, the construction is not always done in double.   Do we need to support mixed precision with gaussians other than to make it compile?

src/QMCWaveFunctions/lcao/LCAOrbitalSetWithCorrection.h
```diff
@@ -0,0 +1,78 @@
...
+
+
+namespace qmcplusplus
+{
+/** class to add cusp correction to LCAOrbitalSet.
+   *
+   */
+struct LCAOrbitalSetWithCorrection : public LCAOrbitalSet
+{
+  SoaCuspCorrection* cusp;
```
ye-luo: I still don't like using pointer. Is it possible to have it non-pointer? Does it have pointers internally? Can we rely on copy constructor?

markdewing: I'll have to look through the makeClone methods to see if this is possible.

ye-luo: I mean only SoaCuspCorrection not the upper levels.

PDoakORNL: std::unique_ptr or other smart pointer? It could make the SoaCuspCorrection lifetime more explicit.
 
Since this is a new class can we explicitly specify the access?
```C++
class LCAOribtalSetWithCorrections : public LCAOOrbitalSet
{
public:
...
```

markdewing: I think it is possible to refer to SoaCuspCorrection by value, and depend on the copy constructor to carry out the makeClone operations.

ye-luo:
Thank you @markdewing. Your exercise shows we should keep reducing makeClone and pointers.

PR #1042, opened 2018-08-29
Fix cusp correction for multiple atoms of one type

markdewing:
A test case with ethanol has been added.  It has multiple repeated atoms and is non-symmetric so the corrections will be different for each atom.

In SoaLocalizedBasisSet, LOBasisSet only contains unique atoms.
The fix in queryOrbitalsForSType is to use IonID to map from the center index to the atom-type index to access LOBasisSet.

The cusp correction depends on the wavefunction of all the other atoms, so it can be different on each atom. The fix in SoaCuspCorrectionBasisSet is to enlarge its LOBasisSet to include all atoms and remove the IonID lookup.


prckent:
Is it possible to reduce ethanol.s004.opt.xml (>200KiB) in size? Would test with a poor basis set make for a better test?

markdewing:
I will try a smaller basis set

PR #1044, opened 2018-08-30
Nexus:  make user simulation collection optional

jtkrogel:
Simplify Nexus user scripts by no longer requiring the manual collection of generated Simulation objects.

Prior user scripts followed the pattern:
```
sims = []

scf = generate_pwscf(...)
sims.append(scf)

p2q = generate_pw2qmcpack(...)
sims.append(p2q)

qmc = generate_qmcpack(...)
sims.append(qmc)

run_project(sims)
```

Following the PR, user scripts can be simplified to:
```
scf = generate_pwscf(...)

p2q = generate_pw2qmcpack(...)

qmc = generate_qmcpack(...)

run_project()
```

The essential change is that all generated simulation objects are collected by the Simulation class upon instantiation.  This works under the assumption that simulations generated are intended for execution (which generally holds) and avoids potential user-side errors of not passing generated simulations for workflow execution and monitoring.

The user examples are modified to reflect these changes.  All example tests pass locally in either the old or new styles shown above.


markdewing:
okay to test

markdewing:
It might be nice to have documentation or an example of where to find the list of simulations, in case the user needs it.    One of the examples could print the list of simulations to run before calling 'run_project'?

jtkrogel:
I think this case is handled by setting "status = True" via the settings function.  This will print the list of simulations prior to the submission of any jobs. 

In terms of accessing the entire simulation list, I can't think of a good use case for this so I will defer that until a request comes up.  The user is of course free to collect them as before without detriment.

markdewing:
test this please

grahamlopez:
check this please

grahamlopez:
test this please

PR #1046, opened 2018-08-31
Protect boundary conditions which are not implemented.

ye-luo:
Non periodic directions must be placed after the periodic ones.
For example, npp is not supported.


grahamlopez:
test this please

Issue #972, opened 2018-08-10
USE_REAL_STRUCT_FACTOR feature complete?


ye-luo:
Is this macro fully implemented in all the places touching and using structure factor?

I also noticed QMC_SK_USE_RECURSIVE. Is this deprecated?

prckent:
Does the code run with this option or are there gaps in the plumbing?

I also worry about the consistent (?) use of StorePerParticle in StructFact.h/.cpp. More generally, if USE_REAL_STRUCTURE_FACTOR is actually complete (?), shouldn't it be an automatic runtime setting and not a compile time setting? 

SK_USE_RECURSIVE has implementation in src/LongRange/StructFact.cpp. I don't know of any users and we certainly don't have any tests or docs.

jtkrogel:
As far as I know, USE_REAL_STRUCT_FACTOR is an optimization (mostly aimed at Ewald sums?), but I don't know if it is feature complete.  A good proxy for feature completeness would be to run the long tests with this enabled and look for additional breakages.

ye-luo:
USE_REAL_STRUCTURE_FACTOR seems set default in CMakeList.txt ADD_DEFINITIONS.
SK_USE_RECURSIVE is not default and I believe this is obsolete.

prckent:
@ye-luo OK to remove SK_USE_RECURSIVE unless someone speaks up in a few days.

PR #1047, opened 2018-09-01
Cleanup structor factor; Remove QMC_SK_USE_RECURSIVE

ye-luo:
As discussed in #972, remove QMC_SK_USE_RECURSIVE.


ye-luo:
Actually, I feel "omp simd" is a better way to go in the future. eval_e2iphi should be retired.
1. simdized sincos needed by omp simd loop is available via libsmvl(intel) and libmvec(glibc). There is no more need to explicitly call vector specific math functions. With omp simd, a larger loop body can be vectorized instead of only one math function. To call the explicitly vectorized math library, scratch arrays are often needed but they all go away with "omp simd".
2. The vendor vector math library may also contain threading. Our vector math calls are not big enough to high omp parallel overhead. I noticed huge slowdown on such calls when nested parallelism is enabled. In the case here, there is no need to use nested parallelism at all. In other cases, the threaded region needs to be at the outer levels.  So it is better to have explict control than relying on the library which does unnecessary things out of our hands.
3. Profiler shows Intel compiler turns sincos into libsvml calls. With gnu, the glibc 2.23 libmvec sincos bug was triggered, namely the libmvec is used. I saw the new "omp simd"+libsvml implementation do runs faster than calling intel MKL vector math library although the change is negligible comparing to the overall walltime.

ye-luo:
This portion is always very small (32 atom 3%, 128 atom <1%). For a while after we updated the cmake, we didn't use the AMD math library for Titan CPU. After that was fixed, I saw the gain was very marginal. I can't answer for all the platforms but our performance data will tell us more.

The libmvec bug only exists in 2.23 which is used by 16.04LTS but any newer Ubuntu is fine. On my desktop, I just built a newer glibc but removed everything but libmvec and then add the path to the LD_LIBRARAY_PATH.

ye-luo:
The performance degradation is real on non-Intel compilers due to inability of proper vectorization. Since there is only 1 hot path, I just put an ifdef.

prckent:
OK. We can track for other compilers + platforms and then update over time, hoping for simplification.

Do you have any suggestions for how to track these? List of known issues by platform? As several issues, multiple issues? We aren't short of GitHub issues.

grahamlopez:
test this please

Issue #1051, opened 2018-09-06
Remove and/or rename QMC_BUILD_LEVEL


markdewing:
The QMC_BUILD_LEVEL flag could be removed in most cases, and renamed to its actual function in other cases.

The default value is 3.
The only current use I'm aware of is to build the sandbox apps (level 5).  This could use a flag named something like QMC_BUILD_SANDBOX_ONLY.
Does anyone use any of the other levels?

Most of the uses in the code are checking whether it is greater than 1 or 2, so true by default.
A few places check that it is less than 5, and these could be replaced by checking the sandbox define.

jtkrogel:
I would vote for removal if possible since it is rather opaque and not very useful.

ye-luo:
As long as I can build Sandbox, any cleanup is appreciated.

PR #1060, opened 2018-09-11
Remove unused references to cubic splines.

markdewing:
This commit affects the molecular code - both SoA and AoS versions.
The SoA code always uses quintic splines.  Remove the use of cubic splines in the AoS code.

Partially addresses #1051


prckent:
Since the molecular SoA integration tests have been passing, this should be safe to merge.

PR #1075, opened 2018-09-14
Update value in test_einset.cpp

markdewing:
The value winds up out of tolerance in newer clang compilers.  The new value comes from clang 6 in debug mode.
Debug and release values from gcc 4.8.4 are also closer to this value.


jtkrogel:
The value changes by quite a bit.  Is it clear if there was anything wrong with the original value?

markdewing:
I don't know -  I didn't dig into the computation.

Issue #321, opened 2017-08-16
Nexus documentation


jtkrogel:
There are a wide variety of features in Nexus that are not documented in the QMCPACK or Nexus manuals.  This issue page serves as a space to communicate which Nexus features, workflows, or associated tools should be prioritized for documentation. 

User documentation:
- [x] How to perform excited state calculations with QMCPACK
- [ ] Tutorial/walk through for research level calculation (by @prckent and @eneuscamman )
- [ ] Details of automated optimal Jastrow selection (by @kayahans )
- [ ] In code documentation of major user facing functions (generate_qmcpack, etc), accessible via help() (@mcbennet )

Developer documentation: 
- [ ] How to add a new machine
- [ ] How to contribute to the testing system

Examples:
- [ ] Examples demonstrating use of QMCPACK observables (by @kylanpaa )
- [ ] Example illustrating automatic QE restarts (by Ron Cohen)

Requested user meeting topics:
- [ ] Detailed design/use of the Structure class (by @lshulen and @Paul-St-Young )
- [ ] Use of PWSCF and QMCPACK analyzer classes (by @Paul-St-Young and @mcbennet )
- [ ] How to use the testing system and add new tests (by @jtkrogel and @markdewing )
- [ ] How to add a new machine (multiple interest)

See also #322 (Nexus features) and #158 (Nexus tests)

prckent:
I would like to read a detailed walk through of a research level calculation. "How this paper was done". "Figure 1 come from ... ". etc. Hopefully this can be recycled from a recent paper and can make use of data on materialsdatafacility.org

eneuscamman:
I agree.  Even for the simplest case, like DFT -> Simple VMC for Jastrow -> DMC, would be great.  Especially if there were both a solid state and a molecular example.

Paul-St-Young:
I would like to read tutorials on how to build on existing calculations. Namely, tutorials that answer questions such as:
1. I have a folder of 100 DFT calculations. How do I run DMC on top?
2. I have an example DMC calculation. How do I add observables? How do I check the dependence of my results on parameters A,B,C ...?
3. I have a wavefunction for system in condition A. How do I check the quality of this wavefunction for the same system in condition B?
...

kayahans:
[...]

PR #1077, opened 2018-09-19
Nexus: add ALCF machine Theta

jtkrogel:
Add support for Theta at ALCF.  Jobs now accept a hyperthreads option.


markdewing:
okay to test

ye-luo:
Do we have some documentation about how to add a machine description in nexus?

jtkrogel:
No, but this would be a good early candidate for developer docs.  Please add this and any other suggestions to #321.  I've edited the issue header to allow for a checklist of recommendations that can be used to determine priorities.

PR #1080, opened 2018-09-19
Nexus: fix skip submit monitoring hang

jtkrogel:
Problem reported by Allison Dzubak.

The skip_submit flag allows the user to write files (including the flow of dependency information) but avert automatic queue submission.  This allows the user to inspect the files before submission and also to opt to submit by hand, if desired.

The manifesting problem was that if skip_submit is used, a Nexus script is closed, and then skip_submit is turned off prior to rerunning the script, occasionally Nexus will not submit the job contrary to expectations.

This PR attempts to fix the bug by never saving state if skip_submit is used.  Additionally, the changes made here allow Nexus to shutdown cleanly (by marking the skipped simulation and its dependents as blocked) when skip_submit is used since this used to lead to monitoring a job that never was going to submit.


ye-luo:
Okay to test

PR #1082, opened 2018-09-21
Spline threading

ye-luo:
This PR enables nested threading mostly in the Spline SPO.
1. Remove unused USE_VECTOR_ML and PRECOMPUTE_L code path.
2. Enable threading in C2C and C2R adoptors.
3. checkAffinity binary is added for checking affinity.

No performance impact if nested threading is not turned on.


src/QMCApp/QMCMain.cpp
```diff
@@ -84,8 +84,19 @@ QMCMain::QMCMain(Communicate* c)
       << "\n  MPI Nodes            = " << OHMMS::Controller->size()
       << "\n  MPI Nodes per group  = " << myComm->size()
       << "\n  MPI Group ID         = " << myComm->getGroupID()
-      << "\n  OMP_NUM_THREADS      = " << omp_get_max_threads()
       << std::endl;
+  #pragma omp parallel
```
markdewing: This output might be better enclosed in a test for nested parallelism (omp_get_nested), and explicitly say something like 'Nested parallelism enabled' (or disabled).  And then only print the number of threads in each level if nesting is enabled.

src/QMCWaveFunctions/BsplineFactory/SplineC2CAdoptor.h
```diff
@@ -310,9 +299,11 @@ struct SplineC2CSoA: public SplineAdoptorBase<ST,3>
   /** assign_vgl
    */
   template<typename VV, typename GV>
-  inline void assign_vgl(const PointType& r, VV& psi, GV& dpsi, VV& d2psi)
+  inline void assign_vgl(const PointType& r, VV& psi, GV& dpsi, VV& d2psi, int first = 0, int last = -1) const
   {
-    typedef std::complex<TT> ComplexT;
+    // protect last
+    last = last<0 ? kPoints.size() : (last>kPoints.size() ? kPoints.size() : last);
```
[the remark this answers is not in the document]
prckent: I would prefer this was written explicitly with no ternary operators.

src/QMCApp/QMCMain.cpp
markdewing: From the standpoint of settings, I would still consider nested parallelism active even if there were only 1 thread in the inner region.  My concern is being able to discern the two cases from the output - if the user is wondering why there is only 1 thread in the inner region - is it because OMP_NUM_THREADS was set that way, or because OMP_NESTED was set that way.
I didn't know about the nested functions and variables being deprecated - that does complicate the decision.  Do you know what is supposed to replace it ?

markdewing: I found the openmp forum thread where you asked about this topic
http://forum.openmp.org/forum/viewtopic.php?f=23&t=2028

One option would be to test the number of threads at the inner level and if it is 1, print that nesting is not enabled. 
Alternately, using omp_get_nested could be used for now, and changed once it's clear where the next version of openmp is heading.

ye-luo: In 5.0, you will still need OMP_NESTED to control the nested threading. It is deprecated but not removed.
Nothing to replace it because OMP_NUM_THREADS=list OMP_MAX_ACTIVE_LEVELS should be sufficient.
It is not clear to me how to control exactly in the future. From the code, using two parallel region to feel the environment is solid.

ye-luo: Now print "OMP nested threading disabled or only 1 thread on the 2nd level" if the second level size is one.

src/QMCWaveFunctions/BsplineFactory/SplineC2CAdoptor.h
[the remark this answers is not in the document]
ye-luo: This is temporal. I have not settled R2R and hybrid. Once these are done. first and last becomes required and there is no need to provide the default value. C++ only allows default argument value as constexpr. I tried kPoints.size() and got compiler error.

prckent:
If this transitions from experimental/development work to something more concrete it will need docs, tests. For the time being, the standard output should not change visibly to users unless activated.

ye-luo:
I agree with docs and tests but I insist changing
```
  OMP_NUM_THREADS      = 16
```
to
```
  OMP 1st level threads = 16
  Nested OpenMP threading disabled
```
The old output is actually misleading because it looks like printing environment variable but actually it is not. It is printing the number of available threads from the OpenMP runtime.

src/QMCApp/QMCMain.cpp
```diff
@@ -81,11 +81,28 @@ QMCMain::QMCMain(Communicate* c)
...
   app_summary()
-      << "\n  MPI Nodes            = " << OHMMS::Controller->size()
-      << "\n  MPI Nodes per group  = " << myComm->size()
-      << "\n  MPI Group ID         = " << myComm->getGroupID()
-      << "\n  OMP_NUM_THREADS      = " << omp_get_max_threads()
+      << "\n  MPI Nodes             = " << OHMMS::Controller->size()
+      << "\n  MPI Nodes per group   = " << myComm->size()
+      << "\n  MPI Group ID          = " << myComm->getGroupID()
       << std::endl;
+  #pragma omp parallel
```
PDoakORNL: I think this should be in a standalone function/method.

prckent:
General comment motivated by a discussion with @jtkrogel - I think we need to watch out more carefully than we have been for performance regressions on small problems.

src/QMCApp/QMCMain.cpp
ye-luo: I agree with you. The whole MPI/OMP printing, or even more, should go to a function so it can be used in other places like QMC tools. I think it is better to change it in a separate PR.

src/Utilities/UtilityFunctions.h
```diff
@@ -70,6 +70,22 @@ inline void FairDivide(int ntot, int npart, IV& adist)
   adist[npart]=ntot;
 }
 
+/** Partition ntot over npart and the size of each partition is a multiple of base size
```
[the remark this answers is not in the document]
ye-luo: I agree with you. An issue (#1090) has been created.
"change its guard constants so it doesn't look like it should be in the OHMMS directories" could you please explain it a bit more. I'm changing it as a template function to avoid potential multi definition issue and support size_t

ye-luo:
In general, small problem should NOT use nested threading.
There should be no performance regression if nested threading is not turned on.
My scheme only computes a few more integers in the inter level.

src/Utilities/UtilityFunctions.h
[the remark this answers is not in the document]
PDoakORNL: #ifndef OHMMS_UTILITYFUNCTIONS_H
#define OHMMS_UTILITYFUNCTIONS_H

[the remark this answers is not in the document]
PDoakORNL: #ifndef OHMMS_UTILITYFUNCTIONS_H
#define OHMMS_UTILITYFUNCTIONS_H

PR #1083, opened 2018-09-21
Nexus: attempt to make ntest thread safe

jtkrogel:
Addresses #1028 (hopefully).

The ntest script was not originally designed to handle simultaneous execution of multiple instances of itself.  To avoid conflict with prior executions, ntest used to remove the prior execution directory and start from scratch.  This is fine for sequential executions, but not concurrent ones.

This PR attempts to fix the problem by no longer removing the prior execution directory in its entirety, but only piecemeal as a particular test enters a sub-directory to run.  I've verified that this approach works both when running all tests sequentially multiple times and when running all tests concurrently.

### DIFF
--- a/nexus/executables/ntest
+++ b/nexus/executables/ntest
@@ -291,14 +291,30 @@ def nlog_restore():
 
 
 # enter the current testing directory
-def nenter(path=''):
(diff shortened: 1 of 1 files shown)


markdewing:
okay to test

PR #1085, opened 2018-09-24
Script to automate regen/check test reference data

jtkrogel:
This effort follows discussions with @prckent.

The idea is to make a Nexus script that can automatically regenerate and/or check against high statistics reference data for all tests in the current testing system.  The script should be easy to run on all nightly/weekly testing platforms.

The ability to do this routinely will allow us to:
* Perform a thorough check of current reference data used in the tests.  Some tests fail intermittently only because the reference data was not created with sufficient care.
* Determine with high confidence whether or not there are any current breakages that are being signaled by the intermittently failing tests.
* Determine if any testing platforms show specific breakages (while others may not).
* Perform high precision statistical tests prior to major releases as a stronger check than the nightlies, etc.

To facilitate this, the testing script needs to be able to:
- [x] Automatically parse QMCPACK's ctest input and extract the short test information
- [x] Modify short test inputs into high precision inputs (100x blocks, 10x steps)
- [x] Perform reference runs on a single platform first (OIC5)

Leaving the following tasks for future work:
* Perform high precision tests on all testing platforms, including parallel submission to batch queues
* Perform checks against reference data for newly run high precision tests and report failures


markdewing:
Would it be better to modify the macros (SIMPLE_RUN_AND_TEST, QMC_RUN_CHECK) to output the information in a standard format, rather than parsing the CMake files?   This would be more robust against changes to the CMake files.

jtkrogel:
I'm not sure.  One issue I'm facing is that the reference generating script must be able to perform a superset of runs compared to any single invocation of ctest since I am aiming to regenerate all the reference data (which requires driving multiple builds at the same time).  

It might be better to have the feature compatibility, reference data, and code version used to create each subset of reference data reside outside this script and ctest and let it inform/check both.  It could be as simple as some tabulated file that both read from.

markdewing:
That sounds good.

grahamlopez:
test this please

jtkrogel:
Moving to have this merged now with follow on work noted.

ye-luo:
@jtkrogel There is a conflict. Could you get it resolved?

ye-luo:
gcc 6.2 and cuda 10

PR #1086, opened 2018-09-25
Modified Nexus xsf read and write

kylanpaa:
The xsf read and write was changed to Fortran order for data grids, and thus, the transpose options used for manipulation of electron densities are not needed anymore.


prckent:
OK to test

PR #1087, opened 2018-09-25
Fix test_einset failures by increasing tolerance.

markdewing:
Update ComplexApprox to accept a different tolerance as an input.

It would have been nice to read the default epsilon value from the Approx class, but it is a private member with no way to access it.


prckent:
Would catch2 handle this better?

PR #1089, opened 2018-09-27
Nexus: error trap reads of misformatted estimators

jtkrogel:
This is a quick fix for reads of misformatted estimators.  The code now aborts with an error message instead of crashing if a user has written (and tries to read) an input file with unnamed/untyped estimators.


markdewing:
okay to test

PR #1091, opened 2018-10-02
Fix to set epsilon values in complex_approx

markdewing:
It works better if the method that sets the epsilon values actually sets the epsilon value.

### DIFF
--- a/external_codes/catch/complex_approx.hpp
+++ b/external_codes/catch/complex_approx.hpp
@@ -75,6 +75,7 @@ public:
 
     ComplexApprox &epsilon(double new_epsilon)
     {
+      m_epsilon = new_epsilon;
       return *this;
     }
 


markdewing:
This uses epsilon the same way that Catch does - using the flow style where the setter returns a reference to the object, so setters can be chained (e.g. Approx(1.0).epsilon(1e-5).scale(1.0)).
I would prefer not to have different styles of setting epsilon for Approx and ComplexApprox.

The problem is C++ doesn't have named keywords (like Python), so it's not clear at the point of usage what the extra parameter is referring to.   ComplexApprox(1.5, 1e-5) (vs python syntax ComplexApprox(1.5, eps=1e-5).  If there is more than one default parameter, then all the previous ones need to be given, even if you don't want to change the default value. This isn't an issue with ComplexApprox now, it would be an issue if additional parameters get added to match Approx.

The flow style of setting parameters is a workaround for these C++ issues - parameters are named and can  be set in any order, eg. Approx(0.5).epsilon(1e-5)  or Approx(0.5).scale(1.0).epsilon(1e-5)

PR #1095, opened 2018-10-04
[WIP] Batched Evaluations refactored into primary path code

PDoakORNL:
This is intended to allow integration of the existing CUDA into the main SoA code path as well a provide a step towards batched evaluation implementations that easily degrade to single evalultions when necessary. I'm intending minimal changes to existing single evaluation code.

This is handled in a few ways:

### Pattern One
When batched evaluation fall back makes derivation from the single evaluation useful
```C++
template<Batching batching = Batching::SINGLE>
class TrialWaveFunction;

template<>
class TrialWaveFunction<Batching::SINGLE>: public MPIObjectBase
{
...
}

template<>
class TrialWaveFunction<Batching::BATCHED>: public TrialWaveFunction<Batching::SINGLE>
{
...
}
```

### Pattern Two
When it is not valuable to have the templated classes in an inheritance chain.
```C++
template<Batching batching = Batching::SINGLE>
class QMCDriverFactory: public QMCDriverFactoryInterface, public MPIObjectBase
{
...
}

class QMCDriverFactoryInterface
{
...
client calls to factory must come through interface.

}
```

### Pattern Three
When the class simply needs to know whether its in a batched or unbatched environment.
```C++
WaveFunctionPool(Communicate* c, const char* aname = "wavefunction", Batching batching = Batching::SINGLE);
```

### Single Evaluation `QMC_CUDA=0 ENABLE_SOA=1 BuildType=Debug` 
- [x] builds
- [x] passes unit tests
- [x] short-diamondC_1x1x1_pp-vmc_sdj-1-16

### Batch Evaluation CUDA `QMC_CUDA=1 ENABLE_SOA=1 BuildType=Debug`
- [x] builds
- [ ] passes unit tests (fails unit_test_einspline)
- [ ] short-diamondC_1x1x1_pp-vmc_sdj-1-16

### Batch Evaluation CUDA `QMC_CUDA=1 ENABLE_SOA=1 BuildType=Release`
- [x] builds
- [x] passes unit tests
- [ ] short-diamondC_1x1x1_pp-vmc_sdj-1-16


ye-luo:
In Pattern one, I don't see any difference with the following. Templates requiring specialization all the time are not useful templates. EinsplineSet has already been a bad example. The spline adopters are designed not template specialization but more like a datatype.
```C++
class TrialWaveFunction: public MPIObjectBase
{
...
}

class TrialWaveFunctionBatched: public TrialWaveFunction
{
adding all the batched interface
}
```
This way is cleaner than the existing code as the following but it requires extra builder to handle TrialWaveFunctionBatched.
```C++
class TrialWaveFunction: public MPIObjectBase
{
...
#ifdef QMC_CUDA
adding all the batched interface
#endif
}
```

Since I don't feel right with the above two, I agree with your Pattern two with some slight tweaking.
```C++
class TrialWaveFunction: public MPIObjectBase
#ifdef QMC_CUDA
, public TrialWaveFunctionBatchedInterface
#endif
{
...
}

class TrialWaveFunctionBatchedInterface
{
adding all the batched interface
}
```

The Pattern three comes from the extra need in the builder to handle batched and non-batched TrialWaveFunction. I still prefer minimal change in the builder if possible.

PDoakORNL:
Initially I did have many of your first block, SPOSet, SPOSetBatched, TrialWaveFunction, TrialWaveFunctionBatched, etc., this proved very awkward compared to my pattern one.

Pattern One is primarily to enable other class templates with code like the following:

```C++
template<Batching batching>
class QMCDriver
{
...
QMCDriver<batching>* create(MCWalkerConfiguration& w,
                    TrialWaveFunction<batching>& psi,
                    QMCHamiltonian& h, HamiltonianPool<batching>& hpool,WaveFunctionPool& ppool);
...
};

```

Some template classes have complete explicit class template specializations right now because this is just one refactor of the currently separate classes, this paves the way for them to share more commonality. It should reduce how many #ifdefs are necessary in dependent code.

I have left many #ifdefs in place in order to not have this become diverging, the emphasis in this PR is to not change the single evaluation behavior while refactoring to enable a more commonality in the existing batched code.

I want to get away from classes that undergo large mutations through ifdefs. See DiracDeterminantBase.h, TrialWaveFunction.h, etc. Turn on QMC_CUDA and you get a class with double the number of methods that is used differently also based on QMC_CUDA definition.  To me this is a bad smell in the code right now.

ye-luo:
There are two separated questions.
1. Using template or not?
I didn't see any commonality. I only see template specialization and dynamic_cast flying.
2. How to put inheritance?
Probably this make you think template is needed.
```C++
class TrialWaveFunction: public MPIObjectBase
{
...
}

class TrialWaveFunctionBatched: public TrialWaveFunction
{
adding all the batched interface
}
```
This is currently what you are trying and it forces you to touch all the builder, factories.
if you do the following, the single and batching APIs are treated on the same foot and existing builders and factories are ready to be reused. I removed the ifdef because API design is not CUDA specific. See what I did in miniQMC.
```C++
class TrialWaveFunction: public MPIObjectBase, public TrialWaveFunctionBatchedInterface
{
...
}

class TrialWaveFunctionBatchedInterface
{
adding all the batched interface
}
```
In miniQMC, `WaveFunction.h, SPOSet.h, WaveFunctionComponent.h` + drivers have no ifdef.
The only place needs ifdef for CUDA is when the builder instantiates the CUDA derived class. These should be very small ifdefs and they are needed no matter what method you do.

ye-luo:
I have already said SPOSet, SPOSetBatched is bad.
If we put batched API in the base class, there is no need for the template specialization at all.
I need to see incremental change or proof of concept on miniQMC.
In my opinion, this is just creating more difficulty instead of reducing the complexity.

ye-luo:
When the batched APIs are added in the base class, we still only need partial specialization which is what the current code do.

PDoakORNL:
Don't see the advantage of having an ever expanding interface on SPOSet which supports multiple evaluation schemes but only some of which are used by a particular scheme. Sounds just like the current mess.

ye-luo:
The current key issue is not how the classes are inherited but the data structure and algorithm are not compatible and then we can't fallback.
The massive ifdef around CUDA is ugly but we can improve it by putting an interface class and make the base class derived from it.

We do suffer the growth of Interfaces in SPOSet but we don't suffer the growth of classes which needs specialization all the time. Adding intefaces cost less than adding classes where all the new interfaces still get added.

PDoakORNL:
Having an interface that appears and disappears on a class that still has the same naming is still just an ifdef'd expanding base class.

If you are just dealing with the single evaluation interfaces as it stands now the only change to your code is SPOSet*  becomes SPOSet<>* as do DiracDeterminant<>* SlaterDet<>* there are no API changes.

The single evaluation is always the default parameter for these class templates. And the default Batching variable value for  the classes that don't need to be templated is SINGLE. So If you can ignore the <> its all just the same. If you do want to work in one of the specialized classes yes you do have to
```C++
template<>
TemplatedClass<>::Yourmethod()
```
instead of 
```C++
YourClass::YourMethod()
```

If I change my mind about this being a good way to refactor the batched eval the change in the single eval code will be the simple removal of the <> and deletion of the Batching::BATCHED specialization files.

Am I missing some big impact here?  If batching is a waste of time it disappears with almost no impact on the rest of the code.
I think I can in short order stand the current CUDA up in this framework, the single eval is already unchanged. Then it paves the way for considerable improvement and elimination of entire alternate flows through the code, but there still will need to be some specialization for batched algorithms.

ye-luo:
`SPOSet* becomes SPOSet<>* as do DiracDeterminant<>* SlaterDet<>* ` they are API changes and there are 356 files are touched.
If we design the batching API not CUDA specific, they can always exist in the base class without ifdef and just have aborts in the base class implementation. Then there is no need to expand the base class for CUDA.

Even a default is placed for <>, it is unnecessary change.
If batching is no more needed, the code needs to remove the template again and touch 300+ files again.

PDoakORNL:
There are in fact only ~82 files where SPOSet* appears.
It's a stretch to call this an API, it is not used by hundreds of people. Those files are all its clients.

It's making my work easier which to me seems useful. 
Its a move toward a more flexible evaluation that enforces some structure
and safety with regards to algorithm/evaluation/data compatibility.

ye-luo:
I still want no change to the drivers and builders and keep changes local if possible.
```C++
# include <iostream>
# include <vector>

using namespace std;
class SPOSet
{
public:
    virtual void evaluate() const = 0;
    virtual void evaluate_batched() const { cout << "SPOSet::evaluate_batched()" << endl; }
    virtual ~SPOSet() {}
};

class SPOSetCUDA_API
{
public:
    virtual void evaluate_batched() const = 0;
};

class SplineSingle : public SPOSet
{
    void evaluate() const { cout << "SplineSingle::evaluate()" << endl; }
};

class SplineCUDA: public SplineSingle, public SPOSetCUDA_API
{
public:
    void evaluate_batched() const { cout << "SplineCUDA::evaluate_batched()" << endl; }
};


int main()
{
    std::vector<SPOSet*> SPOs;
    SPOs.push_back( new SplineSingle() );
    SPOs.push_back( new SplineCUDA() );

    for( std::vector<SPOSet*>::const_iterator it = SPOs.begin();
       it != SPOs.end(); ++it)
    {
        (*it)->evaluate();
        (*it)->evaluate_batched();
        delete *it;
    }

   return 0;
}
```
Does this enhance safety check? If you miss any CUDA API, the compiler stops you.
It can still do fallback to some extend.

PR #1097, opened 2018-10-05
Make spline evaluation functions independent

ye-luo:
During the miniQMC exercise, I found having the evaluation of splines independent is very useful.
One reason is to reduce the deep levels, separate the data with computing if necessary.
The other reason is for easily switching implementations.

Currently, the evaluate functions are contained by the MultiSpline<T> object. To do the computing,
it is needed to do the following
```C++
MultiSpline<T> bs;
bs->evaluate(...);
```
Currently, the nested OpenMP has been added.
```C++
#pragma omp paralllel
{
  bs->evaluate(..., first, last);
}
```
For accelerators operating on the same granularity, accessing the function via an object pointer adds some difficulty in the compiler optimization. I would like to avoid that as the following
```C++
#pragma omp paralllel
{
  spline2::evaluate3d(bs->spline_m,..., first, last);
}
```
I change all the evaluate functions into free templated inline functions with the master control file MultSplineEval.h. In this file, the spline2 namespace is used instead of the usual qmcplusplus.
The implementation now can be selected in this file. For example, the BGQ implementation has been separated from the conventional one. This change facilitate us bringing in specializations for architectures if necessary.


markdewing:
I agree that it is easier to work with alternative implementations if the spline evaluation is a function.

ye-luo:
1. Comments are added.
2. The BGQ implementation is mostly intrinsics and good for performance on BGQ. bgclang supports it. There was only one file including both BGQ and non-BGQ implementations. I didn't like it because the ifdef region is as large as two giant functions. I often made mistake touching the wrong one. Now they are in two files and the code style is very different. The ifdef is only around the include line in MultiBsplineEval.hpp and much cleaner to read the code.

prckent:
OK on the two files, particularly since BGQ will go away soon.

ye-luo:
The BGQ files should go away when Mira retires.
I'm thinking of adding the spline v2 algorithm from miniQMC as a candidate implementation.
This will be standalone files and selected at MultiBsplineEval.hpp similar to the BGQ implementation. I don't like ifdef longer than a screen. The current amount of files (<20) is reasonable and their roles are clear. Adding a new implementation is only 2 files. We won't get 100 files like the einspline folder used to be.

Issue #1055, opened 2018-09-10
Tests need categorization


jtkrogel:
Currently many tests are failing, but for various reasons (broken feature, SoA incompatibility, poorly constructed test).  There is a need to separate "should pass" from "known to fail" tests so that new breakages can be caught and also so that users can test their builds against what is known to work.

Recommend adding two prefixes to the names of existing and future tests:
- **deteriministic**: current unit tests and future very short deterministic tests.  These must pass under any circumstance on any architecture.
- **unstable**: all tests that currently consistently or intermittently fail.  The unstable prefix will be applied separately per build type (AoS real, SoA complex, CUDA real, etc) since this reflects the current state of affairs.

Need to separate tests into separate dashboards on CDash:
- **deterministic**: "ctest -R deterministic", run nightly (and for CI?)
- **short**: "ctest -R short -E unstable", run nightly
- **long**: "ctest -R long -E unstable", run weekly
- **unstable**: "ctest -R unstable", run nightly?

In addition to the unstable dashboard, unstable tests should be tracked in aggregate by type (via GitHub issues) to gauge progress toward development goals (e.g. the SoA transition, real/complex compatibility, desired CUDA features, etc) vs. known breakages or intermittently failing tests.

Tasks:
- [x] Add deterministic prefix to existing unit tests
- [ ] Add unstable prefix to currently identified failing tests
- [ ] Iterate over nightly/weekly tests until all failing tests are identified and labeled

- [x] Add Deterministic build group
- [x] Add Short build group
- [x] Add Long build group
- [x] Add Unstable build group
- [ ] Modify oxygen test scripts to use new build groups and test categorization
- [ ] Modify bora test scripts to use new build groups and test categorization
- [ ] Modify hyperion test scripts to use new build groups and test categorization
- [ ] Modify ceta test scripts to use new build groups and test categorization

See #1062 for an overarching view of the testing problem and links to its other aspects.  Also see #704 for discussion relating to deterministic tests.

jtkrogel:
I can do the categorization/labeling part.  Would be great if someone else could pick up the dashboard (CDash display) part from there.

ye-luo:
Let us also regularize the naming of tests.
short/long-stable/unstable/determinstic-system-method-WF-estimator?

jtkrogel:
[unstable-]short/long/deterministic-system-method-WF[-estimator]-#mpi-#omp

Perhaps this is a separate task, though.

prckent:
[...]

PR #1098, opened 2018-10-05
Test categorization via ctest labels

jtkrogel:
This PR begins to address #1055.

A standard Python script is introduced to add labels to tests (test_labels.py).  This is done because Python is far superior to CMake for set operations and this allows a single point of modification to the existing ctest framework.

The most basic ctest label is "unstable": tests that may fail and should be excluded.  This is build-specific, so the script accepts the build parameters QMC_CUDA, ENABLE_SOA, QMC_COMPLEX, and QMC_MIXED_PRECISION from the ctest system.

Current labeling outlined in the script is as follows (these cover most unique failure cases I can think of that might be useful to track, address, or selectively run as a developer aiming to fix them):
```
# label descriptions
#
#   main classification
#     stable   - test is expected to pass
#     unstable - test may fail (definitely or with some probability)
#     (stable/unstable are exclusive and comprehensive for all tests)
#
#   reason for instability
#     bug           - test failure definitely a bug and QMCPACK needs to be fixed
#     unsupported   - test fails because feature is not supported and was not before
#     poor_test     - test fails because test is insufficient (false positive)
#     cause_unknown - cause/nature of failure is not known and needs investigation
#     (bug/unsupported/poor/unknown are comprehensive for failures)
#
#   failure type classification of unstable tests
#     hard_fail              - ungraceful crash (segfault, etc) 
#     abort                  - controlled abort 
#     check_fail             - fails rigorous non-statistical check (e.g. checksum)
#     reference_stat_fail    - fails separate reference level check by > 5 sigma
#     definite_stat_fail     - statistical failure by > 5 sigma
#     intermittent_stat_fail - statistical failure by < 5 sigma
#     deterministic_fail     - fails non-rigorous deterministic check
#     (failure types are comprehensive for failures)
#  
#   test quality categories
#     good_test       - test well designed: test fail means there is a bug  
#                       any failure means: action needed to fix QMCPACK
#     poor_test       - test poorly designed and failure likely reflects this
#                       regardless of pass/fail: action needed to improve test 
#     quality_unknown - test quality has not yet been assessed
#                       code failure means: action needed to fix QMCPACK
#                       statistical failure means: prioritize quality assessment
#     (good/poor/unknown are exclusive and comprehensive for all tests)
#
#   put failing tests into the following categories (these determine the others)
#     unsupported
#     hard_fail
#     abort
#     check_fail
#     reference_stat_fail
#     definite_stat_fail
#     intermittent_stat_fail
#     deterministic_fail
#     poor_test
#
# main actions for developers and corresponding categories
#   fix known bugs in QMCPACK  - identified by label "bug"
#   fix poor tests             - identified by label "poor_test"
#   assess test quality        - identified by label "quality_unknown"
#   investigate failure causes - identified by label "cause_unknown"
#                                run reference-level statistical test in this case
```

Tasks for this initial PR:
- [x] Create test labeling script
- [x] Add known universally broken features
- [x] Integrate labeling script into ctest system to label test run
- [x] Propagate labels from test run to subsequent quantity tests
- [x] Add hard failures and aborts by build type
- [x] Add definite and intermittent statistical failures
- [x] Redesign the call tree to centralize execution of test_labels.py 
- [x] Verify that tests run properly

Tasks deferred to later PR's:
* Complete specification of unsupported features by build type.
* Abolishing CMake logic based on build types in the CMakeLists.txt for the tests.  Correct application of the labels will allow full filtering with less boilerplate and allow broken features that are currently walled off to be appropriately addressed.


tests/scripts/test_labels.py
```diff
@@ -0,0 +1,226 @@
...
+    cpu   = not qmc_cuda
+    gpu   = qmc_cuda
+    aos   = not enable_soa
+    soa   = enable_soa
+    real  = not qmc_complex
+    comp  = qmc_complex
+    full  = not qmc_mixed
+    mixed = qmc_mixed
+except:
+    error('extract')
```
markdewing: It might aid debugging if the exception handler printed the exception.

CMake/macros.cmake
```diff
@@ -120,6 +120,11 @@ FUNCTION( RUN_QMC_APP_NO_COPY TESTNAME WORKDIR PROCS THREADS TEST_ADDED ${ARGN}
             MESSAGE("Disabling test ${TESTNAME} (building without MPI)")
         ENDIF()
     ENDIF()
+    IF ( TEST_ADDED )
+       SET(TEST_LABELS "")
+       EXECUTE_PROCESS(COMMAND ${PROJECT_SOURCE_DIR}/tests/scripts/test_labels.py ${TESTNAME} ${QMC_CUDA} ${ENABLE_SOA} ${QMC_COMPLEX} ${QMC_MIXED_PRECISION} OUTPUT_VARIABLE TEST_LABELS)
```
markdewing: The RESULT_VARIABLE could be checked to ensure the script returns success.  When the stdout returned from the script is used to set the labels, there should be a guard involving the exit code.

tests/scripts/test_labels.py
jtkrogel: It's true, but the output from this script is consumed directly by the ctest system (and not printed to user visible stdout), so I thought providing a label was a better way to go.

CMake/macros.cmake
jtkrogel: That's a good suggestion.  How do you access the error code?

tests/scripts/test_labels.py
markdewing: It would be combined with checking the return value in the CMake script - if the script fails, print stdout.   Alternately, the exception handler could print to stderr and the CMake script could capture that and print that if the return code is nonzero.

jtkrogel: That's true and it would be an improvement.  I will add it.

CMake/macros.cmake
markdewing: using the RESULT_VARIABLE parameter

jtkrogel: Sorry, was conflating with OUTPUT_VARIABLE.  Will be straightforward.

jtkrogel: Better now?

jtkrogel:
One thing to note: aborts are apparently being counted as "test passed" when the code is run.  For example, the following passed the main test (then postprocessing checks failed after that):
```
Initializing Hamiltonian from file: NiO_nm_choldump.h5
ERROR  Error opening integral file in SparseGeneralHamiltonian. 
ERROR Error initializing Hamiltonian in AFQMCDriver::setup
ERROR Error in AFQMCDriver::setup(...).
Stack timer profile
Timer  Inclusive_time  Exclusive_time  Calls       Time_per_call
Total     0.0028     0.0028              1       0.002763987
```
This happens for aborts from both the real-space and afqmc codes.
Is this just a matter of correctly propagating and/or checking the errorcode?

CMake/unit_test.cmake
```diff
@@ -1,10 +1,21 @@
...
         #SET_TESTS_PROPERTIES( ${TESTNAME} PROPERTIES ENVIRONMENT OMP_NUM_THREADS=1 )
     ELSE()
         ADD_TEST(NAME ${TESTNAME} COMMAND ${TEST_BINARY})
         #SET_TESTS_PROPERTIES( ${TESTNAME} PROPERTIES ENVIRONMENT OMP_NUM_THREADS=1 )
     ENDIF()
+    SET_PROPERTY(TEST ${TESTNAME} APPEND PROPERTY LABELS unit)
+
+    SET(TEST_LABELS "")
+    EXECUTE_PROCESS(COMMAND ${PROJECT_SOURCE_DIR}/tests/scripts/test_labels.py ${TESTNAME} ${QMC_CUDA} ${ENABLE_SOA} ${QMC_COMPLEX} ${QMC_MIXED_PRECISION} OUTPUT_VARIABLE TEST_LABELS RESULT_VARIABLE SUCCESS)
+    IF( ${SUCCESS} STREQUAL "1" )
```
markdewing: It would be better to test against "0" for success, and everything else is failure.  There can be other failures, such as script not found, which puts "No such file or directory" in that string.

jtkrogel: Good to know.  I will switch to check against "0" for success instead.

prckent:
From tonight, oxygen is using the following for each build type to report to cdash
```
        ctest ${CTCFG} -R deterministic ${LONGCFG} --track Deterministic
        sed -i "s/Deterministic/Short/g" Testing/TAG  # Hack to change cached track
        ctest ${CTCFG} -R 'performance|short' -LE unstable ${LONGCFG} --track Short
        sed -i "s/Short/Unstable/g" Testing/TAG 
        ctest ${CTCFG} -R short -L unstable ${LONGCFG} --track Unstable
```
When this PR is in, all 3 categories should appear. Currently nothing is labeled as unstable. I will PR the script once the script, labels and filters are working as expected. Results for other machines will appear under "Nightly" until their scripts are updated.

CMake/unit_test.cmake
jtkrogel: Done.

jtkrogel:
Ready for final review.

CMake/macros.cmake
```diff
@@ -120,15 +120,33 @@ FUNCTION( RUN_QMC_APP_NO_COPY TESTNAME WORKDIR PROCS THREADS TEST_ADDED ${ARGN}
             MESSAGE("Disabling test ${TESTNAME} (building without MPI)")
         ENDIF()
     ENDIF()
+    IF ( TEST_ADDED )
+       SET(SUCCESS FALSE)
+       # workaround the fact that ENABLE_SOA=="" rather than "0" 
+       #   when -DENABLE_SOA is not explicitly provided
```
ye-luo: Could you merge develop and fix this?

jtkrogel: Done.

jtkrogel:
Anything else needed?

markdewing:
okay to test

ye-luo:
I'm getting with `ctest -R diamond -E "long|dmc"`
```
        Start 592: coverage-diamondC_1x1x1_pp-vmc_sdj-1-1
 40/100 Test #592: coverage-diamondC_1x1x1_pp-vmc_sdj-1-1 ..................................***Failed    0.06 sec
        Start 593: cpu_limit-diamondC_1x1x1_pp-vmc-1-1
 41/100 Test #593: cpu_limit-diamondC_1x1x1_pp-vmc-1-1 .....................................***Failed  Required regular expression not found.Regex=[Time limit reached for]  0.05 sec
        Start 594: cpu_limit-diamondC_1x1x1_pp-vmc-4-4
 42/100 Test #594: cpu_limit-diamondC_1x1x1_pp-vmc-4-4 .....................................***Failed  Required regular expression not found.Regex=[Time limit reached for]  0.06 sec
```
Please also verify all converter test and qe tests.

CMake/macros.cmake
```diff
@@ -120,15 +120,31 @@ FUNCTION( RUN_QMC_APP_NO_COPY TESTNAME WORKDIR PROCS THREADS TEST_ADDED ${ARGN}
             MESSAGE("Disabling test ${TESTNAME} (building without MPI)")
         ENDIF()
     ENDIF()
+    IF ( TEST_ADDED )
```
markdewing: I think this should be ${TEST_ADDED} - since it is an output variable, references in the code need to use the name of the variable in TEST_ADDED, not TEST_ADDED itself.

jtkrogel:
Hmm... perhaps some of the interfaces need changing.  There are three potential weak points here: 

1) I had to add an execute process to test_labels.py in multiple places as the test function tree was not sufficiently modular.
2) The test_labels function output argument needs to be exposed to any tests that add other dependent tests, but should not be exposed otherwise (requires changes to lots of calls for no reason and CMake really doesn't warn you that you have supplied only N-1 arguments to a function that requires N arguments...).
3) Given the complaints regarding odd environments that do not have or make it difficult to install Python, a layer of protection needs to be added to prevent a call to test_labels.py when Python is not available.

To solve these issues (which e.g. Ye's find is tangled up in), I will need to spend some time reengineering the function call tree and push down the call to test_labels.py to a new function at the lowest level (one that directly calls ADD_TEST) so that existing leaf-level test function calls remain unaffected and the exception handling for labels is handled in a uniform way everywhere.

I don't have time to do this now, so this will delay the PR for a bit.

jtkrogel:
Tests run with  `ctest -R diamond -E "long|dmc"` now all pass except for those marked unstable.

jtkrogel:
Ready for review.

CMake/run_qe.cmake
```diff
@@ -7,6 +10,10 @@ FUNCTION( ADD_QE_TEST TESTNAME PROCS TEST_BINARY NPOOL WORKDIR TEST_INPUT)
     ENDIF()
     SET_TESTS_PROPERTIES( ${TESTNAME} PROPERTIES ENVIRONMENT OMP_NUM_THREADS=1 PROCESSORS ${PROCS} WORKING_DIRECTORY ${WORKDIR} )
     SET_PROPERTY( TEST ${TESTNAME} APPEND PROPERTY LABELS "Quantum ESPRESSO" )
+    
+    SET(TEST_LABELS "")
+    ADD_TEST_LABELS( ${TESTNAME} TEST_LABELS )
```
ye-luo: See three lines above. For cases will labels already applied, what is the behavior?

jtkrogel: It uses append so new properties should be added.  In general, set_property + append is safe and set_tests_properties is not.  I had to change the unit tests to set_property + append so labels would not be overwritten.

prckent:
(Updating branch ahead of approved merge to check CI is back)

CMake/test_labels.cmake
```diff
@@ -0,0 +1,15 @@
+
+FUNCTION( ADD_TEST_LABELS TEST_NAME TEST_LABELS )
```
ye-luo: The name of this function is very misleading. TEST_LABELS looks like an input parameter but actually it is for both input and output and so it must be initialized.
All the RUN_QMC_XXX functions also need a bit documentation listing input and output argument.

CMake/run_qe.cmake
markdewing: More generally, is the output variable TEST_LABELS used anywhere?  If not, it doesn't seem necessary?

CMake/test_labels.cmake
prckent: Since @jtkrogel is offline, unless you see something overly bad, please create issues for subsequent revisions (descriptions, side effects, renaming,  etc.). I would like to get the dashboard usable again, and this PR is a dependency.

ye-luo: OK. Let's fix it later.

Issue #1054, opened 2018-09-08
GPU result has number of walkers dependency and doesn't match CPU result


yaoyi92:
Details in the post here. 
https://groups.google.com/forum/#!topic/qmcpack/eOE1eIXAgaE

Not 100% sure whether it is a problem for my build. ctest results here in the post
https://groups.google.com/forum/#!topic/qmcpack/1QzNQoceNHs

![image](https://user-images.githubusercontent.com/6993167/45259435-f0239780-b39a-11e8-95ce-9d80eab075ee.png)

yaoyi92:
This seems to be a Volta V100 specific issue. I don't have this problem compiling and running the same simulation on GTX1080.

Need help from someone with experience of Volta V100.

jtkrogel:
Something does seem to be up with Volta: 
https://cdash.qmcpack.org/CDash/index.php?project=QMCPACK&date=2018-09-07.  

Compare "Volta-GCC-CUDA-Release" with "GCC-CUDA-Release".

prckent:
(GCC-CUDA-Release runs on a Kepler. This is printed on one on the ~third line of output.)

prckent:
To keep this updated: the current belief is that this has hit some kind of edge case (bug) or is a newly surfaced problem on Volta or with recent CUDAs (bug). For whatever reason, our current tests don't show the same problem.  @PDoakORNL will try to chase down.

ye-luo:
Some recent fixes in legacy CUDA probably addressed this issue. Open a new issue if needed.

PR #1101, opened 2018-10-09
Split splines across multiple GPUs on a given node

atillack:
Here is my work-in-progress version of the split splines code for GPU with promising performance numbers:

Run | GPUs | Walkers/GPU | Original Code | New Code w/ Split Splines | Split Spline Cost
-- | -- | -- | -- | -- | --
NiO S32 on SummitDev | 2 | 128 | 129.0 | 148.0 | 1.15
-- | 4 | 128 | 129.0 | 236.5 | 1.83
NiO S32 on Summit | 3 | 128 | 91.3 | 148.1 | 1.62
-- | 3 | 250 | 145.1 | 223.2 | 1.54
-- | 6 | 128 | 92.5 | 223.2 | 2.41
-- | 6 | 250 | 147.3 | 403.7 | 2.74
NiO S64 on Summit | 3 | 45 | 230.5 | 334.7 | 1.45
-- | 6 | 45 | 230.5 | 463.8 | 2.01

Currently, only the spline type used for our NiO example is fully implemented. I will of course implement the missing bits once we finalized the code here. 

In order to use the code and split the spline data memory across multiple GPUs the following needs to be done:

- GPU MPS needs to be used,
- the GPUs need to be visible in each MPI rank,
- the options gpu="yes" and gpusharing="yes" need to be set in the \<determinantset type="einspline"\> section in the \<wavefunction\> definition block

I did test the original code based on an older development tree on both SummitDev and Summit with 2,4 and 3,6 GPUs, respectively (this is the data above). After updating to the current development tree I could only run on two GPUs on SummitDev to confirm performance levels are still the same - with four GPUs I get an MPI error (coll:ibm:module: datatype_prepare_recvv overflowed integer range triggered by comm->gatherv_in_place from gatherv in the new spline adaptor code). I get the exact same error with vanilla 3.5.0 on SummitDev. Since I can't run on Summit right now I can't verify if this is a regression in 3.5.0 (3.4.0 did not have these issues on Summit) or just the particular MPI version on SummitDev.

To finish:
- [x] Output error and exit when split splines and UVM are used together for now
- [x] PhaseFactors.cu is already finished but needs testing of the QMC_Complex code path
- [x] Implement double complex evaluation functions
- [x] Implement float and double wave function evaluation functions
- [x] Fill in the missing split splines memory allocations/break-ups in einspline/multi_bspline_create_cuda.cu
- [x] Implement the MPI host distribution to follow the split splines
- [ ] Add documentation in manual and test case


prckent:
OK to test

atillack:
Rhea (gpu) should compile and run fine now.

ye-luo:
In the v3.5, I changed the way of the spline tables collection using in-place gather in order to require no extra scratch memory. MPI is notorious for using 32 bit integer range and you might have an inferior MPI implementation. We need to find a workaround hopefully this is not a bug.
Could you go to src/spline/einspline_util.hpp in the function gatherv could you printout
 buffer->coefs_size and ncol on the crashed run?
There is an internal control switching between two algorithms. Try reduce (1<<20) and see if that works around the issue.

ye-luo:
I'm trying to understand the algorithm. But directly reading the code may not be the best way.
So throw some questions:
1. Since you are dispatching spline evaluation to remote devices. The result vector is allocated as unified memory and relying on CUDA to migrate the data back to the device for the determinant calculation. Am I right?
2. When you list 6 GPUs, how many MPI ranks do you have on the node? 6 or 1?
3. Assuming the previous answer is 6. The spline table on the host is not distributed among the MPI ranks within the group. Only the table on GPU is distributed when the data is copied to GPU. Am I right?
4. How do you share the GPU memory address across different MPIs? or CUDA handles it for you?
5. How much faster is this comparing to the UVM I implemented? One MPI with one GPU, the part of the spline which doesn't fit stays on the host.

atillack:
@ye-luo:

To your earlier question:

On SummitDev with 4 MPI ranks: buffer->coefs_size=446,772,240 ; ncol = 816 (Working on reducing 1<<20 right now).

Algorithm questions:

1) Yes.
2) 6 MPI ranks. Each rank sees all six GPUs.
3) Yes, you are correct.
4) Cuda IPC memory handles (cudaIpcGetMemHandle and cudaIpcOpenMemHandle)
5) Good question. Let me see if I can do a fair comparison by triggering your UVM code to put half the spline table on GPU and the other on the host.

atillack:
@ye-luo Also, I left your UVM code in place and it should still work even with the split splines...

ye-luo:
@atillack I also need some effort on the spline builder to make the distribution on the host and it should connect to the GPU part seamlessly. I think eventually it will be we first split the table over MPI and then split between device and host using UVM if the table doesn't fit the device.

atillack:
@ye-luo Agreed.

atillack:
@ye-luo You are spot-on with your spline table MPI problem. I went to 1<<16 (to have a lower bound) and that runs through fine.

atillack:
@ye-luo 1<<19 of course works too... (446,772,240 / 816 > (1<<19))

prckent:
It will be interesting to see the Summit numbers due to the better NVLINK support + performance compared to SummitDev.

atillack:
@prckent Absolutely.

@ye-luo I also ran the 128 atom NiO with 4 MPI ranks given the same amount of memory per GPU for the spline table (936 MB): split spline takes 231 s; UVM takes 1023 s on SummitDev.

atillack:
@ye-luo & @prckent I also (of course) found a bug using split splines and UVM at the same time. The fix is to also pass the proper host pointers using ICP. For this WIP PR, do you think I should fix it later in a separate PR and right now just output an error and abort in case that's needed?

ye-luo:
If you need to touch lot of no CUDA related code on the host side. I prefer the change separated.
Could you make a list of needed changes to complete this WIP branch?

atillack:
@ye-luo To finish the current PR:

- Implement float and double real as well as double complex evaluation functions
  - once we are happy with the current complex float implementation there isn't much to do but essentially copying the approach to einspline/multi_bspline_cuda_d/s/z_impl.h
- PhaseFactors.cu is already finished but needs testing of the QMC_Complex code path
- Fill in the missing split splines memory allocations/break-ups in einspline/multi_bspline_create_cuda.cu
  - not hard to do and can follow the implementation for 3d_c
  - I am not sure if we want to have this implemented for 1d as well (I'd argue since it isn't in EinsplineSetCuda.cpp to not bother, but there may be a reason to do it I am overlooking)
- Implement the MPI host distribution to follow the split splines
  - depends on the previous part to be finished and will change that code but will not affect anything else (as far as I can tell)
  - not sure how involved this will be, since I am splitting at the orbital level it shouldn't be too hard

atillack:
@ye-luo These are the changes that the split splines code needs for good performance I want us to be happy about:
- the change in CUDA/gpu_vector.h/cpp to add a managed memory option (by default it is false and does should not affect the original code)
- the change in data structure for split splines (sequential memory for each split) and the additional parameter in the QMCWaveFunctions/PhaseFactors.cu kernels (num_split_splines). I believe the changes to be performance neutral for the original, non-split splines code.
- I made the choice of not having separate kernels for split spline functionality but to integrate it in the existing ones (see einspline/multi_bspline_cuda_c_impl.h). I think I did it in a performance neutral way and not adding more kernels should help readability/maintainability of the code.

ye-luo:
1. I thought as long as the results of spline evaluation are transferred back, the phase can be applied locally. Why PhaseFactors.cu needs to be touched?
2. For convenience, not having separate kernels is good. However, it may trigger register spill and lower the performance. Probably modern architectures are more friendly.

atillack:
@ye-luo 

1) Yes. However, with the current data layout of the evaluation, which is orbital-linear for each walker:
[ orb(1)^(1),...,orb(N)^(1) | ... | orb(1)^(nw),...,orb(N)^(nw) ],
with splines split at the orbital level I got many GPU memory bank conflicts leading to about 2-3x worse performance of the split spline evaluation.

The solution was to go to a layout with the split orbitals each getting their own data block
[ orb(1)^(1),...,(orb(Nsplit)^(1)|...|orb(1)^(nw),...,(orb(Nsplit)^(nw) ||
  orb(1)^(Nsplit+1),...,(orb(2*Nsplit)^(1) ||
  ... ||
  orb(1)^(N-Nsplit),...,(orb(N)^(1) ].
For the non-split spline case this doesn't change the memory access pattern but it did require some modifications to PhaseFactors.cu to make it work.

atillack:
@ye-luo 

2) I agree and kept the additional work (and variables) in these kernels to a minimum.

ye-luo:
@atillack The slicing is introduced to make sure the working set fits the fast memory. The orbital data set is (x,y,z,n). Each band contribute to (:,:,:,n) which goes through the whole memory. So when x is sliced. the working set is only (x,:,:,n) which is much smaller and fits the fast memory. In your case, it is a different problem that MPI cannot handle giant message. You may using the same slicing scheme.
The biggest drawback of the current slicing algorithm is (x,y,z,n) is too big but (y,z,n) is too small. We may need a more flexible implementation. Please submit a separate issue.

src/CUDA/gpu_misc.cpp
```diff
@@ -34,6 +34,10 @@ cublasHandle_t cublasHandle;
 
 size_t MaxGPUSpineSizeMB;
 int rank;
+int relative_rank;
```
ye-luo: In-source documentation is needed.

src/CUDA/gpu_misc.cpp
```diff
@@ -34,6 +34,10 @@ cublasHandle_t cublasHandle;
...
+int device_group_size;
+int* device_group_numbers;
```
ye-luo: It is better to avoid C pointers. Can we used std::vector?

src/CUDA/gpu_vector.cpp
```diff
@@ -24,13 +24,19 @@ namespace gpu
 cuda_memory_manager_type cuda_memory_manager;
 
 void*
-cuda_memory_manager_type::allocate(size_t bytes, std::string name)
+cuda_memory_manager_type::allocate(size_t bytes, std::string name, bool managed)
```
ye-luo: CUDA managed memory has many kinds.
Please avoid putting the flag 'managed' and create an interface instead

Please split the allocation tracking from the allocate function.
Create an allocateManaged interface for calling cudaMallocManaged and taking an argument for the cudaMemAttachGlobal or other kinds.

src/Platforms/devices.h
```diff
@@ -155,37 +172,43 @@ inline void set_appropriate_device_num(int num)
   int deviceCount;
   cudaGetDeviceCount(&deviceCount);
   int num_appropriate=0, device=0;
-  bool set_cuda_device=false;
+  bool set_cuda_device=true;
```
ye-luo: Why the logic is changed?

src/CUDA/gpu_misc.cpp
atillack: Will do.

src/CUDA/gpu_misc.cpp
atillack: Yes, I'll upload it later today.

src/CUDA/gpu_vector.cpp
atillack: I can do that.

src/Platforms/devices.h
atillack: The function's scope changed a little by also populating the gpu::device_group_numbers array. This makes it necessary for the loop to run through fully (as opposed to before where it could stop once the device was set) which necessitates  a test if the device still has to be set. Since I like positive tests (if(true) rather than if(!true)) in loops, the variable set_cuda_device changed meaning to "a device still has to be set", rather than "has been set" as before.
Long story short, I was thinking of renaming the variable but the name is logical enough. I can add a comment what the variable does at its declaration.

src/CUDA/gpu_vector.cpp
atillack: Done, part of the last round of commits.

prckent:
Given all these updates, what is still needed/planned to get this merged in before the release milestone?

ye-luo:
What is the plan regarding the unchecked boxes?
Do we have sufficient guards for not working yet cases?

src/Platforms/devices.h
```diff
@@ -143,6 +143,23 @@ inline int get_device_num()
...
     std::cerr.flush();
   }
+  gpu::relative_rank=relative_ranknum;
+  gpu::device_group_size=num_cuda_devices[rank];
+  if (ranks_per_node[rank]<num_cuda_devices[rank]) // sanity check (can't use more GPUs than ranks per node)
+    gpu::device_group_size=ranks_per_node[rank];
+  gpu::device_group_numbers.resize(gpu::device_group_size);
+  gpu::device_rank_numbers.resize(gpu::device_group_size);
+  int device_rank = 0;
+  for (int i=rank-relative_ranknum; i<rank-relative_ranknum+gpu::device_group_size; i++) // although the indexing looks strange, i is between the actual rank numbers on this rank's node
```
ye-luo: Can we simplify the whole loop as the following?
```C++
  for (int i=0; i<gpu::device_group_size; i++)
    gpu::device_rank_numbers[i] = rank-relative_ranknum+i;
```

src/Platforms/devices.h
```diff
@@ -155,30 +172,34 @@ inline void set_appropriate_device_num(int num)
...
+  bool set_cuda_device=true;
   for (device = 0; device < deviceCount; ++device)
   {
     cudaDeviceProp deviceProp;
     cudaGetDeviceProperties(&deviceProp, device);
     if (((deviceProp.major >= 1) && (deviceProp.minor >= 3)) ||
         deviceProp.major >= 2)
     {
+      if (num_appropriate<gpu::device_group_size)
+        gpu::device_group_numbers[num_appropriate] = device;
```
ye-luo: Is it possible to get ride of device_group_numbers?
get_num_appropriate_devices() already check device properties and if no appropriate device exists, the code aborts in get_device_num().
I think it seems always that device_group_numbers[i]=i

src/Platforms/devices.h
ye-luo: I feel it is more clear to extend the name from set_cuda_device to need_to_set_cuda_device

ye-luo:
It seems that the current implementation only allows sharing among all the MPI within a node.
Using 6 MPI ranks but only sharing among 3 GPUs within a NUMA is not ready. Am I right?

src/Platforms/devices.h
atillack: It'll be in the next upload.

atillack:
@ye-luo GPUs can be oversubscribed with this split spline implementation (that's why the device_group_numbers array is needed). I succesfully tested running 8 MPI ranks on the 4 GPUs of a SummitDev node.

ye-luo:
@atillack I was asking if the code can share spline only among each 2 GPUs and while having 4 MPI ranks (4GPUs). My understanding of the code is that this is not available. Yes. the over-subscription works and we will use it for testing.

atillack:
@ye-luo This is something you can achieve outside of our code with CUDA_VISIBLE_DEVICES. On Summit/SummitDev you could create two resource sets per node each with 3 (2) GPUs. That is also tested to be working.

ye-luo:
@atillack That is great, although this requires the resource manager to do the work for you.
I didn't get your point of over subscription for justifying device_group_numbers. It seems not relevant, the content of device_group_numbers is strictly device_group_numbers[i]=i.

src/Platforms/devices.h
atillack: If we want to support GPU oversubscription or more interesting MPI layouts the answer is no. Also, while it costs almost no memory (a vector of ints) it simplifies the code downstream quite a bit. Furthermore, It could be used to support the "undersubscription" you asked for (6 MPI ranks sharing 3 GPUs) as a config option in our code. I did think about reading the gpu_sharing flag in as a number to use for that but then realized I can achieve the same thing externally.

atillack:
@ye-luo Oversubscription is when MPI ranks > # of GPUs. In this case, device_group_numbers[i] = i % # of GPUs. If you don't want to use the resource manager to solve the question you asked, device_group_numbers could be used for that as in that case device_group_numbers[i] =/= i for the most part.

ye-luo:
@atillack The size of the array never goes beyond device_group_size=min(Ngpu, Nmpi)

prckent:
I think we only need to worry about Summit for now where the resource manager will handle any splitting of resources. A test + docs is more important. How we should test this: will the split spline code work with the small grids used for diamond 1x1x1? What happens if the split spline code is run with 16 MPI on one GPU? (oxygen presents either 2 Keplers or 1 Volta currently).

atillack:
@ye-luo Strangely enough, I agree with you that *currently* device_group_numbers[i] = i when moved up into the loop you're suggesting. It feels a bit wrong though as it makes set_appropriate_dev_num seem somewhat superfluous (well, maybe it is actually) and it seems particularly wrong to not store the values in device_group_numbers where the cudaSetDevice call is. That said, I can move the assignment up and simplify the loop as you suggested but would argue keeping device_group_numbers around to make future extensions straight-forward.

ye-luo:
@atillack It doesn't need to perfect. I think keep device_group_numbers for future use is OK. My major goal is to keep "gpu::" stuff local and I was suggesting to move it up.

ye-luo:
@atillack Actually cudaSetDevice is to set the default device for each MPI rank with round-robin.
device_group_numbers is to define the device ids within a group. They are not closely related.

atillack:
@ye-luo Actually, device_group_numbers is used to store the default device number used for each MPI rank on node and device_group_numbers[gpu::relative_rank] needs to correspond to the current ranks default device or things will break. This is why I was hesitant to move it away from the cudaSetDevice call despite agreeing with you that logically, in the use cases we're interested in, it is just device_group_numbers[i]=i. You made a good point though that it isn't needed in set_appropriate_dev_num.

atillack:
@prckent Small grids should not be a problem but the number of orbitals needs to be large enough (>= #GPUs). Splitting on a single GPU should be covered by the last item on the list. Currently, it'll just without splitting.

atillack:
@prckent @ye-luo I updated the list to reflect more clearly what still needs to done. Implementing the real and double evaluations should be done soon and will also close the next item on the list.

ye-luo:
@atillack Thank your for the update. Let talk over the call more in detail tomorrow.

atillack:
@ye-luo Thanks, Looking forward to it.

To the previous points:

1)+2) Agreed, while it's still in though the next commit moves the device_group_numbers bit to get_num_appropriate_devices which allows to use the original set_appropriate_dev_num. Taking out the test for appropriate devices will make the code lighter by two functions (get_num_appropriate_devices & set_appropriate_dev_num). Since they at this point they are basically sanity checks and don't hurt performance we may as well leave them in (although we could merge them with get_dev_num into a get_and_set_device function)

3) I originally started with this scheme but then realized I can save one call to cudaSetDevice (which is expensive) by having the default device be the last in the loop so I don't need to reset it.

ye-luo:
@atillack 3. Having the local piece last also maximize the time overlap between computing and data movement.

atillack:
Almost finished.

atillack:
@prckent I am not sure how to properly set up a test case. In theory, only gpusharing="yes" needs to be set in the <determinantset> section of a test case with enough orbitals (like NiO) but Cuda MPS also needs to be present which is not within the test framework's control (I think).

ye-luo:
Is there a way in the code to test if MPS is enabled or not? If user turns on the flag but the MPS is not ready, the code can turn off this feature or abort the code.
You can grab an existing diamondC_2x1x1 test and enable sharing as a test.

prckent:
Keep it simple. Lets worry about MPS settings when we have proven that they are needed. For oxygen, where we run the nightlies, I don't think we need be concerned. Later PRs can improve the situation as needed.

atillack:
@prckent @ye-luo I think Ye's idea is a good one and it was simple enough to implement (see next commit). If Cuda MPS is off, a warning is generated and the split splines functionality disabled. This way we can add gpusharing="yes" to a test case and nothing breaks if there is no CUDA MPS.

prckent:
Note: Unable to get this to work with MPS on oxygen. Config instructions at https://docs.nvidia.com/deploy/mps/index.html generate a segv. Probably I don't have the correct/magical combination of cuda aware openmpi and environment settings. Without MPS running the code is appropriately benign.

Can someone confirm that this is working on any machine? I currently don't have access to any other machines with multiple GPUs.

atillack:
@prckent I can run on SummitDev and Summit. Documentation is updated. Can you send me the output which segfaults?

prckent:
The LaTeX is slightly broken. I will push a fix and improvements, then merge.

MPS investigations on oxygen will not occur for a while. Having Summit work is the key thing due to upcoming INCITE usage. (Hence #1054 is an important problem.)

Issue #859, opened 2018-05-11
Need a precise use of MPIObjectBase


ye-luo:
I just noticed MPIObjectBase show up as a base class to several classes owing pointer to Communicate object. The constructor allows optional input pointer. When the input is not give, the global one is taken. initCommunicator() can be used to override the already defined pointer. I can feel the convenience but it is quite unsafe.

I'm thinking of always requiring the input pointer instead of having it optional. This will require all the constructors of classes based on MPIObjectBase to add an extra communicator argument. But the hierachy and consistent can be secured. The pointer to Communicate object should also be also defined as const.

Maybe the MPIObjectBase can be simplified to just keep myComm in the derived class but my requirement is the same, use const as much as possible and create it at construction.

Issue #1103, opened 2018-10-10
Need tests for the feature of ensemble runs


ye-luo:

prckent:
Due to the MPI dependencies it is worth checking both an ensemble spline basis run and an ensemble Gaussian basis run.

ye-luo:
I just did a run with 4 MPI.
The ensemble job contains one Gaussian and one Spline calculation inside and now it works with #1104

prckent:
Please add to the test system. If the test is crash or not, this is fast.

prckent:
Partially addressed by #4091

prckent:
Still needed:
1. Statistical checks of results
1. LCAO case

PR #1104, opened 2018-10-10
Require MPI communicator to be passed to MPIObjectBase

ye-luo:
@anbenali paid the price of issue #859 and was not able to run ensemble jobs.

In this PR, MPIObjectBase is no more allowed to grab the WORLD communicator by default and the communicator pointer must be passed through the class hierarchy. The unsafe initCommunicator is also no more allowed (removed) and ReportLevel is also removed. We have too many similar printing controls like ReportLevel.

One set of changes is in real-space QMC drivers. The second set of changes is in all the builder classes derived from SPOSetBuilder. The communicator must be passed through constructors.

Although I don't like the use of pointer in MPIObjectBase, I cannot do large change at the moment since MPIObjectBase is used in both real-space QMC and AFQMC.


jtkrogel:
Shouldn't this PR follow resolution of #1103?

prckent:
We should get in the habit of committing the tests first or including them in the PR.

ye-luo:
@prckent Constructing such test is non-trivial. The checking mechanism in our testing system doesn't handle ensemble run. However, this fix is immediately needed for runs.

jtkrogel:
Actually, I think this is easily done in the testing system.  Just make a diamond test via ensemble with 4 identical input files listed in the ensemble input.

It should pass the same reference check as the original 16x1 or 4x4 test.  All that is needed is to make check_scalars.py support ensemble averaging like qmca does.

ye-luo:
@jtkrogel This is not a strong test. The old code can also pass this. I need to mix Gaussian and spline and check results individually.

jtkrogel:
I'm actually more concerned about the original case, i.e. that innovations to mpi could crash our current way of doing it, so I would favor including a test first (separate PR) that the current code passes, and then use it to confirm nothing has changed with this PR.

ye-luo:
Let me explain the whole story.
MPIObjectBase constructor has a mechanism adding a default communicator OHMMS::Controller which is WORLD. The old code SPOSetBuilder class has MPIObjectBase as a base class without giving the communicator. When LCAOrbitalBuilder(derived from SPOSetBuilder) tries to build up the basisset and orbitals for different molecules in an ensemble run. The MPI calls via WORLD from the calculation of different molecules can not match due to different basisset and orbitals. Not only the size of message but also the number of calls may differ. Some metadata can be in wrong size and trigger segmentation fault.

The real situation is even worse. When I run an ensemble job mixing Gaussian and Spline. Both Spline and Gaussian hangs. This indicates even in the spline WF part, WORLD is used in some places. Our spline ensemble job runs in the past because WF are similar for each calculation and the MPI calls match by luck.

After the fix is applied, the communicator must be explicitly managed. There is no change to the non-ensemble job. For an ensemble run, each sub run must follow the communicator passed in QMCMain.

jtkrogel:
Definitely sounds shady and worth fixing.

ye-luo:
ReportLevel was an internal state (rank 0 has 1 and 0 otherwise).
SetReportLevel is to set ReportLevel into a different value. I didn't see a case using a different value but just passing the value in.
ReportLevel is used to tell some functions to print or not.
I don't see a value of this rarely used state machine.

Putting rank()==0 is simple and more understandable.

prckent:
Good to remove internal state.

Issue #1020, opened 2018-08-20
Add CMake installation for QMCPACK and NEXUS


prckent:
Based on the request in #1017 , we need to arrange "make install" to work for both QMCPACK and NEXUS. CMake has many options to support this.

I suggest we just focus on the default real cpu build. In later versions we can handle Quantum Espresso, ideally after our converter is in QE, and then complex.

What needs to be installed?

build/bin/*
nexus/executables/*
nexus/library

Although NEXUS is not a python package, there must be some standards to follow.

jtkrogel:
Happy to do the Nexus part, though not sure what is desired in the "make install" context.

naromero77:
For reference, this is what I do in the Spack package for install.
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/qmcpack/package.py#L250

I install all binaries in bin, I copy over the entire Nexus directory, and I "install" the manual.

naromero77:
For Python, there are usually two options:
1. install in "site-packages" subdirectory of the Python distribution
OR
2. Add the appropriate directories to PYTHONPATH environment variable.

jtkrogel:
Thanks Nick.  I've generally taken the second option to "install" Nexus.

prckent:
Regarding NEXUS perhaps copying to $INSTALL_DIR/nexus/...  will suffice. It would be good if qmca would work out of the box.

prckent:
(We'll have to discuss who can handle the QMCPACK side. @PDoakORNL ?)

jtkrogel:
qmca is fully standalone (no install required).  All that is needed is python, numpy, and (optionally) matplotlib.

markdewing:
I can work on the QMCPACK side, unless @PDoakORNL is already working on it.

markdewing:
The bin directory contains `qmcpack` and `convert4qmc`, which are definitely used and should be installed.
But what about these others?  Are they used? Should they be installed?
* getSupercell
* extract-eshdf-kvectors
* checkAffinity

ye-luo:
We may add or remove files over time. So copy bin/* over.

markdewing:
Should we adopt some sort of standard for the names of the executables? (add qmc_ prefix, or at least have qmc in the name, as in convert4qmc)  In the case where someone installs into a common directory (like /usr/local/bin), so the qmcpack-derived executables can be easily identified.

prckent:
[...]

PR #1107, opened 2018-10-11
Fix OSX compilation and rename checkAffinity to qmc-check-affinity

### LINKED ISSUE #1105: checkAffinity does not build on OS X due to missing/undeclared sched_getcpu

ye-luo:
close #1105
The name change is suggested by #1020
I don't have access to any Mac. So the strategy is printing nothing.
Someone needs to check the compilation


prckent:
Confirming that this now compiles on OSX. Checked with clang 6.0.1 (macports), no mpi. -1 is returned from qmc-check-affinity, as expected.

PR #1108, opened 2018-10-12
Reroute Linear Method Middle Shift Output

LeonOtis:
@ye-luo A few lines are added to ensure separate output files are produced for the statistics from the middle shift of the adaptive three shift version of the linear method. Unnecessary uses of 'this->' have also been removed.


prckent:
OK to test

ye-luo:
@LeonOtis Does "ctest -R adaptive" pass?

LeonOtis:
@ye-luo Yes, all four tests from 'ctest -R adaptive' pass.

prckent:
Thanks Leon

PR #1109, opened 2018-10-12
Nexus: small fix to gaussian process code

jtkrogel:
Quick fix to read array dimension.  Before re-wrapping this code as a Nexus module, the variable "d" was obtained unintentionally from the global namespace.


markdewing:
okay to test

markdewing:
Putting a promo for static analysis: pyflakes would catch this.

PR #1110, opened 2018-10-15
Performance test input files for CPU J3 and GPU are generated

ye-luo:
We had CPU, CPU-J3, GPU input files for NiO. The only difference between these sets is just a bit input tag change. I created adjust_qmcpack_input.py script for adjusting input files. Now CPU J3 and GPU input files are derived from CPU input files. The generated files are slightly different from the original ones by sorting the attributes. 96 and 192 atom cells are added to have finer granularity.

This addresses #1111

Thank @markdewing for providing me the initial python script.


tests/performance/adjust_qmcpack_input.py
```diff
@@ -0,0 +1,103 @@
...
+    j3_tree = ET.parse(args.J123)
+    change_jastrow(tree, j3_tree)
+
+  if args.walker:
+    change_number_of_walkers(tree,args.walker)
+
+  if args.output:
+    tree.write(args.output)
+  else:
+    tree.write(fname_in)
```
markdewing: rewriting the input file by default makes me nervous - too easy to mess up something.
clang-format uses the '-i' option to specify in-place editing.

prckent:
Please document and discuss the perceived problems with the current setup in an issue. If the results of the discussion and that this is the best way forward, reopen the PR. We have many of examples of people (non-experts) needing to run the inputs directly and without any additional complications.

PR #1112, opened 2018-10-15
Nexus: enable rsync for Quantum Package

jtkrogel:
Enable rsync between subsequent QM Package runs performed in different directories.

Also include updates to the input spec to allow dft input and revert automatic writes of integrals back to user control.

PR #1113, opened 2018-10-16
Nexus: enable QM Package restart for selected CI

jtkrogel:
Add input variable "converge_dets" to enable auto restart of selected CI runs that have not reached the target number of determinants.

This PR follows (depends on) #1112.


jtkrogel:
NB: I will soon have another small PR that depends on this one.  I am planning to submit it just like this one, but I am open to suggestions if chained PR's are somehow undesirable.

ye-luo:
Okay to test

PR #1114, opened 2018-10-16
Nexus: add failure check to QM Package SCF runs

jtkrogel:
This PR depends on #1113.

SCF runs are now checked for successful completion.  In the event that failure is detected, execution of all dependent runs is blocked.

Also add a small fix for writes of floating point values.


ye-luo:
Okay to test

PR #1116, opened 2018-10-17
Cuda type aliases

PDoakORNL:
I have removed the CUDA typdefs from the class hierarchy.  They are now provided via using a templated on Precision CUDATypeAliases class. This is a simple refactor meaning the code has been rearranged but there is no change in its behavior.  This supports work for refactoring the CUDA and supporting batched algorithms. Due to the lack of encapsulation of the CUDA functionality many files were touched, but in a repetitive mechanical fashion.


PDoakORNL:
test this please

PDoakORNL:
Shorten to CUDATypes?
CUDA_PRECISION_FULL and CUDA_PRECISION still occur here and there in the code I will see if they can be excised. In certain cases CUDA_PRECISION_FULL seems to be used to get full precision regardless of what CUDA_PRECISION is set as.
CUDATypeAliases should ultimately not be visible to anything but the CUDA implementation files so it should not be in the included everywhere "Configuration.h"

PDoakORNL:
```C++
template class<ST, VT, DIM>
class CUDADevice{
using CTA = CUDATemplateAliases<ST, VT, DIM>
//CTA::GradType can be used generically.
}
```

ye-luo:
The whole code is rewritten largely without specialization for complex.
The CPU drivers managed to have no ifdef but the CUDA drivers are ugly with ifdef. This needs to be improved.

ye-luo:
Okay to test

ye-luo:
The CI failure is caused by the GPU AoS build.

ye-luo:
After digging into the code, there is till a long way to have a consistent precision definition even just the CUDA part. Currently, everyone is grabbing the global definition. In principle, the datatype consistency in the APIs needs to be enforced via inheritance while local datatype can be kept flexible.

PDoakORNL:
I don't think inheritance is a good way to enforce this. I'd prefer matching template parameters or they use a global types object. There are too many layers including MI and simply getting the type through inheritance is too vague. It should be crystal clear which RealType or whatever you are using.

For this I prefer to see something like
`using CTS = CUDAGlobalTypes;`
then CTS::RealType instead of CudaRealType or RealType

or 
using QTS = QMCGlobalTypes;
QTS::RealType instead of RealType where you have gotten it from somewhere in your large MI inheritance tree.

Additionally this allows a class template like this
template<P, V> // P=precisionType, V=valueType
```C++
class Whatever
{
using QTS = QMCTypes<P, V>
...
}
```
This allows generation of however many template classes you need
in the executable.
i.e. in the actual code there is object code
Whatever<double, double>
Whatever<double,std::complex<double>>
Whatever<single>
Whatever<single, std::complex<single>>
And this are distinct types.

One can write another class template like
```C++
template<P, V>
class UsesWhatever
{
using QTS = QMCTypes<P,V>
Whatever<P,V>* myWhatever;
void setWhatever(Whatever<P,V>*);
QTS::ValueType* evalSomething(ParticleSet& P, ...);
}
```

ye-luo:
ValueType is a type defined in real when building real or complex when building complex.
This is a restriction to developers and we don't give full freedom.
This convention was used throughout the code and works very clean.
When something is created not following this convention, using a different name to avoid confusion.
My observation in the past shows violating this convention creates a lot of duplicated codes and ifdefs.

PDoakORNL:
As I've pointed out many times existing code is welcome to use the 
CUDAGlobalTypes
Which creates things like
```
class CUDAWhatever 
{
using CTS = CUDAGlobalTypes;
//lets you do what you did before i.e. in pseudo code
QMC_COMPLEX ? CTS::ValueType = std::complex<RealType> : CTS::ValueType = RealType; 
...
};
```
i.e. a CUDAWhatever that could have complex Values or not depending on a particular compile.
In my code in the future I would far prefer to see an object type of
`CUDAWhatever<SINGLE, std::complex<SINGLE>,BATCHED>` in my debugger
And if I was dealing with several at once:
```C++
CUDAWhatever<SINGLE,std::complex<SINGLE>,BATCHED>
CUDAWhatever<SINGLE,SINGLE,MANAGED>
//vs
CUDAWhateverManaged<SINGLE>
CUDAWhateverBatchedComplex<SINGLE>
```
I will fix GPU AoS but lets remember the point is to get rid of it not to degrade design to support it easily.

ye-luo:
User class has freedom to create its own dedicated types while the QMCTypes impose certain restrictions. If your class design in practice shows having template argument is very necessary and helpful in QMCTypes, we will just adopt it.
If possible(seem highly), complex specialization should be avoided.
If you can do the following in a single compilation,
CUDAWhatever<float,std::complex<float>,BATCHED>
CUDAWhatever<float,float,MANAGED>
your second template type must be very local.
After your development, you can just put in the production code.
CUDAWhatever<CTA::RealType,CAT::ValueType,BATCHED>
Anyway, this is for discussion in the future when you show the code.
Please fix the AoS and the we will merge.

ye-luo:
Test this please

PDoakORNL:
AoS Jastrows have some inconsistent naming, the good news is that every rename refactor makes the code more consistent so changing the name again is easier :wink:

Issue #962, opened 2018-08-09
utils directory needs cleaning


jtkrogel:
The utils directory has a large number of scripts and it is unclear which are currently in use.

Here is the current set of scripts:
```
bgrunjobhelper.sh     DMCEnergy.py         OptProgress.pl    PlotTwists.pl
check_input.py        energy.pl            PlotBsplConv.pl   setup-qmc-conf.pl
check_qmcsystem.py    Energy.py            PlotDMC.pl        setup-qmc.pl
check_scalars.py      Forces.py            PlotDMC.py        single_stat.perl
clean_and_link_h5.sh  FWforces.py          plot_fsatom.py    test_import.py
ConfigureScripts.pl   generate.py          PlotTstepConv.pl  TwistAvg.pl
cuspcorrection.py     getBestSupercell.pl  PlotTwistConv.pl
```

Please comment here if you use (or know of the use of) one of these scripts.  Unchampioned scripts will be flagged for removal.

Check the box next to any script below if it is in active use (bonus: add a parenthetical comment to describe its use):
- [X] bgrunjobhelper.sh (used by config/BGQ_Clang++11_ToolChain.cmake to set MPI NUMPROC)
- [ ] check_input.py
- [ ] check_qmcsystem.py
- [X] clean_and_link_h5.sh (used in CMake/run_qe.cmake as part of QE testing)
- [x] ConfigureScripts.pl (update paths to the external tools inside perl scripts)
- [ ] cuspcorrection.py
- [ ] DMCEnergy.py
- [x] energy.pl
- [ ] Energy.py
- [ ] Forces.py
- [ ] FWforces.py
- [ ] generate.py
- [ ] getBestSupercell.pl
- [x] OptProgress.pl
- [ ] PlotBsplConv.pl
- [x] PlotDMC.pl
- [ ] PlotDMC.py
- [ ] plot_fsatom.py
- [ ] PlotTstepConv.pl
- [ ] PlotTwistConv.pl
- [ ] PlotTwists.pl
- [x] setup-qmc-conf.pl
- [x] setup-qmc.pl
- [ ] single_stat.perl
- [x] test_import.py (used by test system, needs to be moved to tests/scripts)
- [x] TwistAvg.pl

jtkrogel:
test_import.py and check_scalars.py are in active use by the testing system.

jtkrogel:
@lshulen are you still using any of the perl/python data analysis scripts here?  

[...]

PR #1120, opened 2018-10-23
Clean up utils directory

jtkrogel:
Remove old utility scripts that are no longer in use.

Addresses #962.


ye-luo:
Just realized that utils/ConfigureScripts.pl is still needed.

jtkrogel:
Restored.

ye-luo:
Okay to test

ye-luo:
Since the CI will not be back until Thur, I will not approve at the moment and let people figure out if there is anything in two days.

PR #1123, opened 2018-10-24
Nexus: enable gamess orbital dependency for quantum package

jtkrogel:
The PR allows Quantum Package runs to depend on GAMESS runs for orbitals.  This means GAMESS can be used for its superior Hartree-Fock convergence properties to kick-start Quantum Package calculations.

Using the dependency looks like this:
```
gms = generate_gamess(
    ...
    )

qp = generate_quantum_package(
    ...
    dependencies = (gms,'orbitals'),
    )
```


ye-luo:
Okay to test

Issue #832, opened 2018-05-03
Update minimum CMake version from 2.8.10


markdewing:
There is a tension with the minimum CMake version required - supporting older versions makes it easier for users, but moving to newer versions makes life easier for developers. 
 This issue is to gather some information on what versions would help developers, and which versions are easily accessible on various platforms.

The current minimum required is 2.8.10, with some additional unit tests enabled for versions > 2.8.11

CMake 3.1 introduces the CMAKE_CXX_STANDARD and CMAKE_CXX_STANDARD_REQUIRED, which should set the flags on the compilers properly, and the existing c++11 flag checking could be removed.

A more modern approach to CMake organization uses target_* commands.  The target_source command was added in CMake 3.1.  I don't know if there are subsequent additions in later CMake versions to better support this style of organization.

prckent:
Someone using creaky machines will have to admit their minimum (maximum?) CMake version. CMake is generally trivial to install, so I am happy to move to (say) 3.1. As with (say) our QE support it would be good to define a policy if possible.

prckent:
No replies yet. Seems like it would be safe to require cmake 3. Package managers have it and the HPC centers have it....

prckent:
We have been using 3.6.3 on oxygen since 9/2016. Unless anyone chirps up by the 9th, when this issue will have been up for a week, lets make 3.6.0 the minimum version.

ye-luo:
ALCF Mira has 3.8.1 and Theta has 3.9.1.

PR #1124, opened 2018-10-24
Update minimum CMake version to 3.6

markdewing:
Implements #832 

Switches to the CMake method of specifying the C++11 flags.
Version checks for older CMake's are removed.

The manual is also updated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,15 @@ PROJECT(qmcpack)
 CMAKE_MINIMUM_REQUIRED(VERSION 3.6.0)
 IF(COMMAND cmake_policy)
   cmake_policy(SET CMP0003 NEW)
+  IF (COMPILING_ON_BGQ)
+    # On BG/Q, the HDF libraries variable includes some supporting shared libraries
+    # such as libm.so.  With the old behavior of CMP0060, these were converted
+    # to -lm for the linker, which then correctly selects the static version.
+    # With the new behavior of CMP0060, the form of the libraries are left alone
+    # on the link line (eg. /usr/lib64/libm.so), and this causes a linker error
+    # because shared libraries are not allowed with static linking.
+    cmake_policy(SET CMP0060 OLD)
+  ENDIF()
   IF(NOT $ENV{CRAYPE_VERSION} MATCHES ".")
     cmake_policy(SET CMP0056 NEW)
     # This policy insures that the CMAKE_EXE_LINKER_FLAGS of the calling project
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,15 +2,15 @@ PROJECT(qmcpack)
 
 
 #####################################################
-CMAKE_MINIMUM_REQUIRED(VERSION 2.8.10)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.6.0)
 IF(COMMAND cmake_policy)
   cmake_policy(SET CMP0003 NEW)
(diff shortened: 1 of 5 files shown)


manual/installation.tex
```diff
@@ -668,6 +676,7 @@ \subsection{Installing on ALCF Theta, Cray XC40}
 \verbatimfont{\footnotesize}%
 \begin{verbatim}
 export CRAYPE_LINK_TYPE=dynamic
+module load cmake/3.9.1
```
ye-luo: Please use cmake/3.11.4
3.9.1 doesn't pick up the parallel hdf5 correctly.

config/BGQ_Clang++11_ToolChain.cmake
```diff
@@ -1,6 +1,8 @@
 set(CMAKE_C_COMPILER mpiclang) 
 set(CMAKE_CXX_COMPILER mpiclang++11)
 
+set(COMPILING_ON_BGQ TRUE)
```
ye-luo: You probably has checked. This doesn't need to be cached, right? I mean only need 1 cmake.

markdewing: The toolchain file gets called many times, including before the regular CMakeLists.txt is processed.  I don't think this needs to be cached.

CMakeLists.txt
```diff
@@ -2,15 +2,24 @@ PROJECT(qmcpack)
 
 
 #####################################################
-CMAKE_MINIMUM_REQUIRED(VERSION 2.8.10)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.6.0)
 IF(COMMAND cmake_policy)
   cmake_policy(SET CMP0003 NEW)
-  IF(CMAKE_VERSION VERSION_GREATER 3.1.3 AND NOT $ENV{CRAYPE_VERSION} MATCHES ".")
+  IF (COMPILING_ON_BGQ)
```
ye-luo: Is it because of HDF5_USE_STATIC_LIBRARIES?

markdewing: Yes.

ye-luo: Is it possible to put an if around HDF5_USE_STATIC_LIBRARIES=off instead of changing the policy.

markdewing: The HDF5_USE_STATIC_LIBRARIES flag seems to have no effect on BG/Q - the list of libraries is the same.
HDF libraries =
/soft/libraries/hdf5/current/cnk-gcc/current/lib/libhdf5.a;/soft/libraries/alcf/current/gcc/ZLIB/lib/libz.a;/usr/lib64/libdl.so;/usr/lib64/libm.so

ye-luo: I see. I think CMake is doing right because HDF5_USE_STATIC_LIBRARIES only means HDF5  is static but the overall linking is still dynamic by default.

markdewing:
test this please

manual/installation.tex
```diff
@@ -876,30 +886,34 @@ \subsection{Installing on NERSC Cori, Xeon Phi KNL partition, Cray XC40}
...
 \end{verbatim}
 
-When the above was tested on 29 August 2017, the following module and
+When the above was tested on 24 October 2018, the following module and
 software versions were present:
 
 \verbatimfont{\footnotesize}
 \begin{verbatim}
-build_cori_knl> module list
+build_core_knl> module list
```
ye-luo: Typo here.

PR #1129, opened 2018-10-26
Reorder splines for simpler parallel construction

ye-luo:
In the C2R case, the splines are reordered based on whether one or two orbitals are generated.
When the splines are constructed in parallel over MPI, the construction uses the old older reading from h5 and then dump the splines into memory with the new order. This makes the the communication at the end of the parallel construction very complicated.
In this PR enables parallel construction directly in the new order and then all the communication part becomes simple and the lines of code reduce.

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/SplineAdoptorReaderP.h
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineAdoptorReaderP.h
@@ -297,20 +297,21 @@ struct SplineAdoptorReader: public BsplineReaderBase
     {
       if(band_group_comm.isGroupLeader())
       {
-        int ti=cur_bands[iorb].TwistIndex;
-        std::string s=psi_g_path(ti,spin,cur_bands[iorb].BandIndex);
+        int iorb_h5=bspline->BandIndexMap[iorb];
+        int ti=cur_bands[iorb_h5].TwistIndex;
+        std::string s=psi_g_path(ti,spin,cur_bands[iorb_h5].BandIndex);
         if(!h5f.read(cG,s)) APP_ABORT("SplineAdoptorReader Failed to read band(s) from h5!\n");
         double total_norm = compute_norm(cG);
         if((checkNorm)&&(std::abs(total_norm-1.0)>PW_COEFF_NORM_TOLERANCE))
         {
-          std::cerr << "The orbital " << iorb << " has a wrong norm " << total_norm
+          std::cerr << "The orbital " << iorb_h5 << " has a wrong norm " << total_norm
                     << ", computed from plane wave coefficients!" << std::endl
                     << "This may indicate a problem with the HDF5 library versions used "
                     << "during wavefunction conversion or read." << std::endl;
           APP_ABORT("SplineAdoptorReader Wrong orbital norm!");
         }
         fft_spline(cG,ti);
-        bspline->set_spline(spline_r,spline_i,cur_bands[iorb].TwistIndex,iorb,0);
+        bspline->set_spline(spline_r,spline_i,cur_bands[iorb_h5].TwistIndex,iorb,0);
       }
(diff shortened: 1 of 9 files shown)


ye-luo:
To hit the code path doing the reordering, the system must be a supercell of at least 2 primitive cells.
I checked all the h5 files used for testing, they are all not large enough. diamondC 1x1x1 and NiO 4atom are both single primitive cells. diamondC 2x1x1 has the same order after reordering, thus it is not a good test case. What will be the system size the deterministic tests? Hopefully that is large enough.

prckent:
The situation for *integration* testing is clearly to run a large enough supercell, but we aren't there yet with deterministic tests due to problems GPU side. What would it take to have a *unit* test for this spline builder functionality (@markdewing )?

ye-luo:
Currently the builder is tested implicitly. Namely, a xml string is passed in and then an SPO is generated. Then we check the SPO evaluation. To test the ordering, I'd like to check the BandIndexMap vector against the recorded value in the unit test.

markdewing:
In order to even guess at a unit test, it would be nice to know
 - what defines the old order
 - what defines the new order
 - why is the ordering/reordering needed
 - where is the point where control flow or code logic changes based on whether the supercell is big enough or not

ye-luo:
old order: the order of orbitals in hdf5, sorted by eigenvalues in DFT.
new order: A pair of splines contributing to two real orbitals is placed before a pair contributing only 1 real orbital in the C2R case. In C2C and R2R, the ordering is untouched.
why: the if-statement can be moved out and have more efficient computation.
C2R,R2R control: UseRealOrbitals(set in EinsplineSetBuilderCommon.cpp) is the state flag to determine if a single spline is needed for an orbital or a pair of splines are needed.

src/QMCWaveFunctions/BsplineFactory/SplineAdoptorReaderP.h
```diff
@@ -297,20 +297,21 @@ struct SplineAdoptorReader: public BsplineReaderBase
...
-          std::cerr << "The orbital " << iorb << " has a wrong norm " << total_norm
+          std::cerr << "The orbital " << iorb_h5 << " has a wrong norm " << total_norm
                     << ", computed from plane wave coefficients!" << std::endl
                     << "This may indicate a problem with the HDF5 library versions used "
                     << "during wavefunction conversion or read." << std::endl;
           APP_ABORT("SplineAdoptorReader Wrong orbital norm!");
         }
         fft_spline(cG,ti);
-        bspline->set_spline(spline_r,spline_i,cur_bands[iorb].TwistIndex,iorb,0);
+        bspline->set_spline(spline_r,spline_i,cur_bands[iorb_h5].TwistIndex,iorb,0);
```
PDoakORNL: `bspline` needs to be plural or `bspline_set`  or `bs_set`

src/QMCWaveFunctions/BsplineFactory/SplineC2RAdoptor.h
```diff
@@ -184,7 +154,10 @@ struct SplineC2RSoA: public SplineAdoptorBase<ST,3>
   /** remap kPoints to pack the double copy */
   inline void resize_kpoints()
   {
+#ifndef QMC_CUDA
+    // GPU code needs the old ordering.
     nComplexBands=this->remap_kpoints();
```
PDoakORNL: This comment is confusing define the ordering in the comment. Then define the new ordering.

Will the GPU code still work? It seems like if it does the i=i mapping is all the GPU code needs or is this just a case that it isn't caught by the unit tests?

ye-luo: The comment is updated.
Yes. The GPU code works. The GPU kernel applying phase need the i-i mapping. Currently our uni tests are not covering the ordering and the integration tests are too small to have a difference in the order.

PR #1132, opened 2018-10-30
Print a warning for the buggy glibc 2.23 in cmake.

### LINKED ISSUE #526: Possible bug using glibc 2.23 and libmvec
As discussed in #375 cmake guards against proven buggy combinations of libraries should be added. The bugs should also be cleanly reproduced and a tester made if practical.

ye-luo:
Closes #526
A warning will be issued when MKL is not used. This still allows the compilation to go.
```
CMake Warning at CMake/FindVectorMath.cmake:45 (MESSAGE):
  Using glibc 2.23 which contains a buggy libmvec crashing QMCPACK.
  Workaround needed for this issue.
```
My workaround is putting a newer libmvec on the LD_LIBRARY_PATH.
At the compile time, the old lib will still be used. I don't even know how to change that.

### DIFF
--- a/CMake/FindVectorMath.cmake
+++ b/CMake/FindVectorMath.cmake
@@ -10,9 +10,7 @@ IF ( HAVE_MKL_VML )
 # We arrive here if MKL was detected earlier by FindMKL
   SET ( HAVE_VECTOR_MATH 1 )
   MESSAGE(STATUS "Using MKL Vector Math functions")
-ENDIF ()
-
-IF ( NOT HAVE_VECTOR_MATH )
+ELSE()
   #MESSAGE(STATUS "Trying MKL VML")
   # Check for mkl_vml_functions.h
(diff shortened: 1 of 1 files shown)


PDoakORNL:
How common is 2.23 for our user base?
Swapping the dynamic library to a newer version is probably as simple as its going to get since glibc is supposed to be backward compatible.
Forcing a newer glibc at compile time is certainly possible but not something cmake will just do for us and not worth automating or documenting for a rare system configuration issue.

ye-luo:
2.23 is the default on ubuntu 16.04 LTS. I can say a lot. But if Intel compiler is in use, libmvec will not be taken.

prckent:
Are we guaranteed to crash or give wrong results? If so, aborting and not warning would be best.  

Ideally gcc would have no nomvec option. Exactly when mvec use is enabled wasn't clear to me by reading the gcc docs.

PR #1134, opened 2018-10-31
Nexus ppconvert

camelto2:
Small addition to nexus for running ppconvert. This allows you to pass additional arguments in ppconvert in case you need to change the density_mix parameter or choose a difference local channel for UPF potentials.


markdewing:
okay to test

PR #1135, opened 2018-10-31
Update to C++14 as the minimum required standard

markdewing:
The only part strictly required is the CMake line `set(CMAKE_CXX_STANDARD 14)`

However, because of library compatibility issues, a configuration check was added.
I don't know if the tested library feature (access tuple by type) is used in the code, but it is a simple and convenient feature to use for the library check.

The error messages from the check attempt to give some advice.

The output from try_compile is written to a file to help in diagnosing issues.  It includes the compile command line and compiler output, making it pretty useful.   (Finding the output in the CMake files otherwise is difficult. I've never been able to find it - it doesn't seem to be in CMakeError.log or CMakeOutput.log)

The manual is updated with directions on different machines (mostly adding 'module load gcc' to the setup) and a section on the issues around the C++ 14 library.
Side note:  'module swap gcc gcc' looks like it should be idempotent, but it's not, at least if the initially loaded gcc and the default gcc versions are not the same.   I wrote it as 'module swap gcc gcc/6.3.0' to avoid the confusing appearance, but it has the same effect.


prckent:
Can you confirm which machines / build recipes  you checked?

prckent:
Lets use standard library in the messaging.

ye-luo:
As far as I know there are only two C++ standard libraries. libstc++(GNU) and libc++(LLVM).
I think we can simply check __GLIBCXX__
https://stackoverflow.com/questions/37118114/value-of-glibcxx-for-each-libstdc-release
I have less knowledge about libc++, it seems that _LIBCPP_VERSION can work.

markdewing:
The CMake messages and manual have been updated to use 'C++ 14 standard library'

prckent:
A direct check of the C++14 library capability is very cmake style and simpler than version checking.

I am happy with these updates.

ye-luo:
Direct check is not safe if the library is partially supporting the feature which is tested.

markdewing:
We can change it if there is a problem.    The test feature (getting tuple element by type) is listed as implemented in GCC 5.  All the other features are listed as implemented as of GCC 5, so this should be a fairly safe check.

Mapping GLIBCXX version numbers to supported standard version looks complicated, especially as 4.9.3 has a date after 5.0.

prckent:
Keep it simple. We can revise what Mark has done in case of issues.

prckent:
Merging. Important to check the nightlies tomorrow.

PR #1138, opened 2018-11-01
Workaround Intel 19u0 internal compiler error.

ye-luo:


prckent:
Q. Do you understand why this works? 
Q. Does the compiler still crash if openmp is disabled?

ye-luo:
Because of that line, the compiler changes code analysis and hit a different code path for code generation. I never tried turn off OpenMP since we rarely compile QMCPACK that way, I guess the compiler may still crash.

PR #1139, opened 2018-11-01
Add space in front of -x check for Intel compilers

markdewing:
Add the space to ensure this only matches options where -x is at the beginning.

Issue showed up passing the option "-cxxlib=/path/linux-rhel7-x86_64".
The '-x' in the middle of the expression would satisfy the match condition.


ye-luo:
I see a problem with CMAKE_CXX_FLAGS="-xCOMMON-AVX512"

markdewing:
D'oh!  Back to the regex drawing board.

ye-luo:
Test this please

PR #1136, opened 2018-10-31
Remove Base from (Multi)DiracDeterminantBase

PDoakORNL:
Small rename refactor.

PR #1140, opened 2018-11-01
updated the CI failure message for clarity

grahamlopez:


tests/test_automation/jenkins_rhea_cpu.sh
```diff
@@ -109,5 +109,14 @@ cp $BUILD_DIR/$BUILD_TAG.o* ../
 
 cp $BUILD_DIR/$BUILD_TAG.o* ../
 
-# explicitly check for correct test output from all builds
+# check for correct test output from all builds
+
+if [ $(grep '100% tests passed, 0 tests failed out of [0-9]*' ../$BUILD_TAG.o* | wc -l) -eq 4 ]
```
ye-luo: Sorry, it needs to be -ne 4.

grahamlopez: sigh, you are right. submitted fix. how embarrassing

markdewing:
It would be nice to check the return values from cmake and make and give some guidance as to which stage started the failure (not for this PR, future idea).   Now it requires some searching to figure out where the failure happened.

grahamlopez:
@markdewing agreed, there is quite a bit more we could be doing.

prckent:
Please could whoever merges this retrigger https://github.com/QMCPACK/qmcpack/pull/1136 ? Hopefully it will work now.

Issue #350, opened 2017-08-29
Revise variables passing from ctest to cmake


ye-luo:
In order to avoid too long ctest command line, the way to pass options from ctest to cmake needs to be revised.

See the exploration in PR #347

HDF5_ROOT is mananged by FindHDF5 module and used as environment variable only. It can be passed in as environment variable.
QMC_COMPLEX is managed by our CMakeLists, it has to be passed by -D explicitly.

PDoakORNL:
So are we actually hitting ARG_MAX to ctest or is this an aesthetics thing? Because you're losing information that will end up being easily visible in cdash and its database when you use environment variables over cmake -D's. I'd prefer to have all the dependencies HDF5, MKL, BOOST, etc. passed in explicitly and recorded in the cdash database. (scroll down to the whole cmake command)[https://cdash.qmcpack.org/CDash/buildSummary.php?buildid=28410] but you don't get the environment (at least by default).

For the end user the opposite is probably optimal, everything comes from the environment and the module load qmcpack makes sure that a series of other modules are loaded so the environmental paths bring in the correct things. So our build should support this when possible.

prckent:
CMake/ctest_script.cmake has many sections like the excerpt below. This is not scalable and is currently the only way to route options to the cmake line when ctest is running the script and driving the build.

```
IF ( QMC_COMPLEX )
  SET( CTEST_OPTIONS "${CTEST_OPTIONS};-DQMC_COMPLEX='${QMC_COMPLEX}'" )
ENDIF()
```

prckent:
I agree with the point about cdash browsability. This is very helpful and not something to lose.

PDoakORNL:
But we can definitely come up with something more scalable than the above.

prckent:
Our current strategy seems OK imo. Reopen if needed.

PR #1141, opened 2018-11-02
Only set ctest options that need to be set

prckent:
Solves the labeling failures in the nightlies


ye-luo:
This is not a scalable solution and why not trying the solution I gave at the end of #350.

prckent:
Getting the dashboard working properly has higher priority.

ye-luo:
Could you give an estimate when we can fix #350? It is clearly getting worse and you were spending a lot of time patching a bad solution.

PR #1143, opened 2018-11-02
Enable processor affinity in ctest and speed up testing

ye-luo:
This significantly reduce testing time on machines with >16 cores
Hyperion time drops from 2+h to <1h.
Bora time drops from 30 mins to 20 mins.
This is a cmake 3.12 feature but it seems ignored correctly by older versions of cmake.

Do NOT merge until Monday.


prckent:
Please check a machines with <16 real cores running openmpi.

ye-luo:
For <16 real cores, there is no change showed last night on ryzen-box. Cores are any way oversubscribed. 
The definition of PROCESSOR_AFFINITY from the documentation
```The set of processors chosen will be disjoint from the processors assigned to other concurrently running tests that also have the PROCESSOR_AFFINITY property enabled.```

prckent:
This will be a big improvement. Can merge when you give the OK.

ye-luo:
This is ready to merge now.

ye-luo:
@prckent I noticed a side effect of this change. When I ran ctest manually, I need to use -j to tell the maximal number processors. Otherwise, it goes to 1 core which is like the behaviour of make. The upside is that people are encouraged to run tests in parallel instead of sequential.

For example,
`ctest -R short-diamondC_2x1x1_pp-vmc_sdj-1-16` runs on 1 core.
`ctest -j 4 -R short-diamondC_2x1x1_pp-vmc_sdj-1-16` runs on 4 core.
`ctest -j 32 -R short-diamondC_2x1x1_pp-vmc_sdj-1-16` runs on 16 core.
`ctest -R unit` runs on 1 core.
`ctest -j16 -R unit` all the unit tests run concurrently each using 1 core.

prckent:
Good catch. In principle this is an improvement, but we'll need to advertise to avoid confusion.

Suggestion:

We add a message at the end of the cmake configure step indicating what to do next, along with some instructions. i.e.

QMCPACK is ready to build. To build QMCPACK: 
make -j 16 # Adjust for core count on your system (we could even check since we have logic in CMake/ctest_script.cmake)

To test QMCPACK after building:
ctest -j 16 -R deterministic -LE unstable # All tests must pass.
ctest -j 16 -E long -LE unstable # Run all the tests except the long ones or those known to be unstable. Tests can fail statistically. Rerun to check they eventually pass.

It would be very slick if we could add the ctest message at the end of the build (make) step since it would encourage people to run the tests.

PR #1144, opened 2018-11-03
Remove CONSTEXPR macro

ye-luo:
CONSTEXPR mcaro was introduced pre-C++11. No more needed.
On hyperion, constexpr doesn't work well with std::complex and use const instead.
The support is currently under C++2a.
All other places of CONSTEXPR are replaced with constexpr.


prckent:
Translation for hyperion: there is an issue with Intel 17.0.8.20180907 ?

ye-luo:
The issue may exist in all Intel 2017. I can't say this is the compiler's fault since this is not standardized yet. We were using the old standard library and it was fine. With a more complicated std::complex in a newer version of GCC standard library, the compiler feels it no more qualified using constexpr and reports an error. Newer versions of Intel compilers probably loosen certain constraints and make it working again. Overall, this needs to be fixed in the specification and then every compiler should make it working. For us, I don't expect a noticeable performance hit and we do want to maximize compatibility.

PR #1148, opened 2018-11-06
QMCPACK build instructions for ARM

ye-luo:
I only need to build hdf5 and newer cmake beyond the script shows.
In the script, I treat armclang as clang to workaround cmake.
Message was sent to arm to improve the cmake compatibility.
C++14 is fine.


ye-luo:
Good point. A manual entry is better. I will move it to the manual and keep it there only.
3. I don't like some hacks on the ARM page. I will ask them to align with us.

ye-luo:
I moved the recipe to manual. But I'm hesitating to cite the link. First, they should follow us not the opposite. We are the official one. Second, links may easily get changed.

prckent:
(I could find two sources of build instructions https://developer.arm.com/products/software-development-tools/hpc/resources/porting-and-tuning/building-qmcpack-with-arm-compiler and https://gitlab.com/arm-hpc/packages/wikis/packages/qmcpack . The second is user contributed and might be the original source. We could PR updated instructions to this wiki e.g. when v3.6.0 is out.)

ye-luo:
@prckent. Correct, I got email from ARM that we can update the gitlab and then they will update the webpage accordingly.

PR #1160, opened 2018-11-16
Test tracks

### LINKED ISSUE #1122: Improved support for ctest test tracks
Current support for multiple test tracks (noted in #1098) requires a hack due to my limited ctest understanding &/or ctest documentation/implementation issues. https://github.com/TriBITSPub/TriBITS/issues/189 and https://cmake.org/cmake/help/latest/command/ctest_start.html suggests that modifying the CTEST_START call might work.

ye-luo:
closes #1122
All the tracks are handled inside ctest_script.cmake and submission to cdash is more frequently.
Look at ryzen-box GCC7-Complex-Mixed-SoA-Release on cdash last night as an example.
Now the configuration and compilation only appear under Nightly.
I used Deterministic, Unstable, Performance and Converter. The Converter track goes to Experimental due to the lack of Track on cdash.

I feel we may not need the Short track since it can be leave under the Nightly together with the Configure and Build and we can save some horizontal space. This can be improved later.


prckent:
Thanks. LGTM.

I am happy with the new test track changes. Breaking out the converter tests seems worthwhile. We can revise if needed as we gain experience and get feedback. Note that I may not be able to revise the oxygen scripts in time for tonight.

ye-luo:
@prckent I guess your scripts should still work although you can clean it up when you are less busy. I do need your help to open a Converter track on Cdash. All the uncategorised submission from machines other than oxygen will automatically get on the new tracks.

prckent:
With the choice of the default submission being to nightly it is imperative that we update all the test machine scripts promptly. Currently the nightly track conveniently separates out the machines that are not following the protocol.

prckent:
Note: The deterministic test track will eventually include more than the unit tests and the labels used on line 321 will need expanding.

PR #1161, opened 2018-11-16
Adjustments and fixes for user meeting

jtkrogel:
The PR contains two small commits needed to support the demos given in the Nexus user and developer meeting today.

Fix a recently introduced bug in job bundling (interaction with skip_submit) and color blocked simulations in black when graphing simulation workflows.


prckent:
OK to test

PR #1163, opened 2018-11-16
Refactor builders for jastrow factors

lshulen:
This addresses Issue #430.  Using templates and specialization, all one and two body radial jastrow factors are built using a new class: RadialJastrowBuilder .  bspline, pade and rpa jastrows are currently supported for both one and two body.  Note that RPA jastrows are currently somewhat broken (see the testing situation) this refactoring does not attempt to fix them, only to build them in the same way as the code currently does.

Areas where testing has been lighter include optimization and the gpu build of the code.


ye-luo:
Okay to test

lshulen:
Hmmm, I will look into this.  Everything built fine on my system with the intel compilers.  Is rhea using gcc?

prckent:
The scripts on rhea used for CI currently invoke gcc 5.3.0

lshulen:
OK, I found the issue.  Not actually sure why intel let it compile, but there you have it.  Anyway, I've pushed a fix to my branch.

src/Numerics/OptimizableFunctorBase.h
```diff
@@ -23,6 +24,7 @@
 #include "Optimize/VariableSet.h"
 #include "OhmmsData/OhmmsElementBase.h"
 #include "OhmmsPETE/TinyVector.h"
+#include <cstdio>
```
ye-luo: iostream?

src/Numerics/OptimizableFunctorBase.h
```diff
@@ -113,5 +123,22 @@ struct OptimizableFunctorBase
 
 };
 
+inline void print(OptimizableFunctorBase& func, std::ostream& os)
```
ye-luo: A free function in the header may cause multi-definition problem. Make it a template(functor) function.

src/QMCWaveFunctions/Jastrow/BsplineFunctor.cpp
```diff
@@ -7,7 +7,7 @@
 // File developed by: Ken Esler, kpesler@gmail.com, University of Illinois at Urbana-Champaign
 //                    Jeongnim Kim, jeongnim.kim@gmail.com, University of Illinois at Urbana-Champaign
 //                    Jeremy McMinnis, jmcminis@gmail.com, University of Illinois at Urbana-Champaign
-//
+//
```
ye-luo: Fashion.

src/QMCWaveFunctions/tests/test_einset.cpp
```diff
@@ -169,7 +168,7 @@ const char *particles =
   REQUIRE(ddpsiV[1](0,1) == ComplexApprox(1.8089479397).compare_real_only());
   REQUIRE(ddpsiV[1](0,2) == ComplexApprox(0.5608575749).compare_real_only());
   REQUIRE(ddpsiV[1](1,0) == ComplexApprox(1.8089479397).compare_real_only());
-  REQUIRE(ddpsiV[1](1,1) == ComplexApprox(-0.07996207476).epsilon(eps).compare_real_only());
+  //REQUIRE(ddpsiV[1](1,1) == ComplexApprox(-0.07996207476).epsilon(eps).compare_real_only());
```
ye-luo: Did this unit test problematic?

lshulen: Actually my system was giving a problem with catch.  ComplexApprox and epsilon were not working together and I could not figure out how to fix it.  If it's working for others it shouldn't stay commented out.

src/QMCWaveFunctions/Jastrow/BsplineFunctor.cpp
lshulen: Nobody has ever accused me of being fashionable before.  How do I accept your suggestion?

ye-luo: Sadly, I feel PR owners didn't self review their PR first and made an effort to reduce the amount diffs. I really wish I could put less comment.

src/QMCWaveFunctions/tests/test_einset.cpp
lshulen: When I build with the intel compiler, I get this:
/remote/lshulen/qmcpack/src/QMCWaveFunctions/tests/test_einset.cpp(171): error: class "Catch::ComplexApprox" has no member "epsilon"
    REQUIRE(ddpsiV[1](1,1) == ComplexApprox(-0.07996207476).epsilon(eps).compare_real_only());
    ^

/remote/lshulen/qmcpack/src/QMCWaveFunctions/tests/test_einset.cpp(171): error: class "Catch::ComplexApprox" has no member "epsilon"
    REQUIRE(ddpsiV[1](1,1) == ComplexApprox(-0.07996207476).epsilon(eps).compare_real_only());

and the build fails.  However gcc does not complain at all.  This was before I started working, on anything though.  I'll remove the comment in my next comment.

src/QMCWaveFunctions/Jastrow/BsplineFunctor.cpp
lshulen: I did comb over this in what I thought was a thorough fashion.  What I have learned is that using the intel compiler in a development situation is a bad idea.  For instance, the reason the tests have been failing is that I had forgot to return the success variable at the end of the put method for the RadialJastrowBuilder class.  However, on intel the unit test still passed.

Anyway, I'm new to the formal development process here.  Let me know if it would be preferable for me to make a clean new pull request, or just fix these few things.

src/Numerics/OptimizableFunctorBase.h
lshulen: I'm not sure what you mean by a template(functor) function.  Do you want it to become a method of OptimizableFunctorBase?  Regarding the multi-definition problem, my understanding was that making it inline solved this issue.  I don't think the function is generic beyond this class hierarchy so I'm not sure that doing something like:
template<class foo>
void print(foo& f, std::ostream& os) 
makes sense to me.  Is that what you had in mind?

src/QMCWaveFunctions/Jastrow/BsplineFunctor.cpp
ye-luo: I wouldn't say doing development with Intel is that bad. At least you can cover two compilers.
Just fix the current branch. No need for a new PR.

src/Numerics/OptimizableFunctorBase.h
ye-luo: Yes.
```C++
template<class FT>
void print(const FT& f, std::ostream& os) 
```

ye-luo: If the compilation becomes very slow, we may break one cpp into several and then we may encounter the multi-definition problem. inline is a weak suggestion to the compiler so not necessarily followed by the compiler.

lshulen: I'm a bit uneasy with the template suggestion.  To me that implies that the function is more general than it is (it's only meant to apply to OptimizableFunctorBase derived objects).  In terms of capturing my intent, I would actually be happier leaving the declaration in the header and putting the definition in a cpp file.  That said, I'll defer to your judgement.  Should I change it to a template?

ye-luo: I don't insist on template. Add a cpp file is also fine to me although there is one more file and you need to edit the cmake file to include it.

prckent: This is a simple "service" function with no performance requirements. The inline is definitely not needed here (as with many uses of inline in our source code). Why not put this function in the .cpp file? Ideally it would be a member function, but provided it is legal just keep it simple.

Luke won't be able to clean up all this code so it is important to strike a balance.

lshulen: OK, making the switch now.  One question though.  Your snippet made the argument const.  That was my first inclination, but the f and df methods of OptimizableFunctorBase are not declared const even though it seems to me they could be.  Should I go ahead and change them and track down the derived classes?

prckent: If it is not too painful due to the cascading dependencies, improving const correctness is worthwhile.

ye-luo: @lshulen if f and df are not const, you can not put the const. Just leave it.

lshulen: I'm trying to make f and df const because I think this is the right thing to do.  Mostly it's not causing a problem.  However, it seems that CubicBSpline is a state machine, so while the intention is that nothing is being changed by calls like splint, there are internal scratch variables that are being changed.  

I'm going to take a bit harder look at this and see if I can make it work out.  Otherwise, I'll just make the argument to print non-const and go from there.

ye-luo: I believe making CubicBspline non-const will be quite a lot of work. We can have the change on builders merged in and then solve the non-const problem. I don't quite understand why we have this one and OneDimCubicSpline which seems doing similar things.

lshulen: I just pushed a change which makes the current code base work at the cost of making a few scratch variables mutable.  To me it made sense to do this and then make an issue to get rid of CubicBspline in favor of OneDimCubicSpline or a direct use of a one dimensional bspline.  I'll be curious to see if you are of the same opinion.

ye-luo: IMO, marking scratch space mutable is dangerous and misleading. When I read the code and see const function. I would expect the code is thread safe and I can call f in a threaded region. With a mutable scratch this is not true. If the purpose is only to get the code to run, I'd like to leave the code without const so it doesn't confuse people. Mutable can be used for some internal parameter which may need some change after calling the constructor but having frequently changing data mutable lost of validity of const.

lshulen:
Figured out problem with previous commit.  However one of the machines I'm testing this on fails on one of the test_hamiltonian tests by ~1e-6 when I use MPI and mixed precision.  I took this to another machine and could not reproduce the problem.  I'm a bit nervous that something is wrong but I wanted to see whether rhea had the same issue.

src/Numerics/OptimizableFunctorBase.h
lshulen: OK, I'll unwind the const business.  In the end, I find myself not happy with the architecture of these sections.  As @ye-luo rightly points out, they are not thread-safe and I can't see how to make them so without serious redesign.  Also, the places I can find them used in the code don't appear to derive much benefit from the scratch spaces other than perhaps avoiding allocation of a few bytes of memory on call.  I was sort of expecting staged evaluation where the bits left in the scratch space were re-used but I don't see where this is happening.

I think I will make an issue and come back to the general design of the functors later.

src/QMCWaveFunctions/Jastrow/JastrowBuilder.cpp
```diff
@@ -110,6 +107,12 @@ bool JastrowBuilder::addOneBody(xmlNodePtr cur)
   bool success=false;
   ParticleSet* sourcePtcl= (*pa_it).second;
   //use lowercase, to be handled by parser later
+  RadialJastrowBuilder rb(targetPtcl,targetPsi,*sourcePtcl);
+  success = rb.put(cur);
+  return success;
+}
+
+  /*
   tolower(funcOpt);
```
ye-luo: Are these commented part all moved to the new builder? If not, please add a comment.

src/QMCWaveFunctions/Jastrow/JastrowBuilder.cpp
```diff
@@ -163,6 +167,12 @@ bool JastrowBuilder::addTwoBody(xmlNodePtr cur)
 {
   ReportEngine PRE(ClassName,"addTwoBody(xmlNodePtr)");
   bool success=false;
+  RadialJastrowBuilder rb(targetPtcl,targetPsi);
+  return rb.put(cur);
+}
+
+
+  /*
```
ye-luo: Are these commented part all moved to the new builder? If not, please add a comment.

src/QMCWaveFunctions/Jastrow/PadeFunctors.h
```diff
@@ -60,14 +61,19 @@ struct PadeFunctor:public OptimizableFunctorBase
   std::string ID_B;
 
   ///default constructor
-  PadeFunctor(): Scale(1.0),ID_A("0"),ID_B("0") { }
+  PadeFunctor(): Scale(1.0),ID_A("0"),ID_B("0"),A(1.0),B0(1.0),Opt_A(false),Opt_B(true) 
+  { 
+    reset();
+  }
 
+  /*
```
ye-luo: There are a few commented builder. Do we still need them?

lshulen: They are no longer needed, I have removed them

src/QMCWaveFunctions/Jastrow/JastrowBuilder.cpp
lshulen: All of this was moved to the new builder.  I have removed the comments.

PR #1164, opened 2018-11-19
Improve test labeling and ctest script

ye-luo:
It seems that there is no need to return TEST_LABELS from ADD_TEST_LABELS function as well as all the functions calling ADD_TEST_LABELS. So I changed all the invocation to pass TEST_LABELS by value and the calls are much simpler.

Also fixed the default track in ctest_script.cmake. If this is not specified, the last portion of test results are submitted to the last track which is "Unstable" and it explains why most of the tests are under the Unstable track. With this change, tests without labels "unit|performance|converter|unstable" will go back to the Nightly track.
We will rearrange these tracks later if needed.

I also noticed a minor issue with the python script. quality_unknown label is always added.


ye-luo:
@prckent a side question. On CDash admin page can your control the order of the tracks? It seems not ordered by alphabet.

prckent:
Track order is easily configurable.

jtkrogel:
@ye-luo there are a few known poor tests and for these "quality_unknown" should not be present, but "poor_test" should.

Most tests have not been checked thoroughly enough yet to classify as good or poor yet.  I'm working on this next as I check the reference values and error bars.

jtkrogel:
There are places where TEST_LABELS needs to be returned.  Any test that strings on others as dependencies needs this.

One example is in QMC_RUN_AND_CHECK where test labels are queried/assigned to the main run and then passed down to the checking processes (see SET_PROPERTY line following call to check_scalars.py in QMC_RUN_AND_CHECK) to be applied as labels to them as well.  

SIMPLE_RUN_AND_CHECK has the same problem.

CMake/macros.cmake
```diff
@@ -169,10 +165,9 @@ FUNCTION(QMC_RUN_AND_CHECK BASE_NAME BASE_DIR PREFIX INPUT_FILE PROCS THREADS SH
 
 
     SET( TEST_ADDED FALSE )
-    SET( TEST_LABELS "")
     SET( FULL_NAME "${BASE_NAME}-${PROCS}-${THREADS}" )
     MESSAGE("Adding test ${FULL_NAME}")
-    RUN_QMC_APP(${FULL_NAME} ${BASE_DIR} ${PROCS} ${THREADS} TEST_ADDED TEST_LABELS ${INPUT_FILE})
+    RUN_QMC_APP(${FULL_NAME} ${BASE_DIR} ${PROCS} ${THREADS} TEST_ADDED "integration" ${INPUT_FILE})
```
jtkrogel: TEST_LABELS is used on line 220 below and needs to be returned here.

CMake/macros.cmake
```diff
@@ -249,8 +244,7 @@ function(SIMPLE_RUN_AND_CHECK base_name base_dir input_file procs threads check_
 
   # add run (task 1)
   set (test_added false)
-  set (test_labels "")
-  RUN_QMC_APP(${full_name} ${base_dir} ${procs} ${threads} test_added test_labels ${input_file})
+  RUN_QMC_APP(${full_name} ${base_dir} ${procs} ${threads} test_added "integration" ${input_file})
```
jtkrogel: test_labels is used on line 275 and needs to be returned here.

jtkrogel:
I don't recall offhand if there are other places like these, but there may be.

CMake/macros.cmake
ye-luo: I just invoke ADD_TEST_LABELS

CMake/macros.cmake
ye-luo: I just invoke ADD_TEST_LABELS

prckent:
General comment: before this PR we were only using labels to indicate the reliability of a test and not the test type. We are selecting for test types based on the name (prefix) and not labels since it makes a very easy story for users & developers. e.g. ctest -R test_type or ctest -R test_type -LE unstable. We could e.g.  add labels based on the name if it had benefits internally.

ye-luo:
Thanks @prckent. I did get confused before you comment. I noticed a few existing labels, for example coverage,unit. We will need to adjust them according to the convention in your comment.

CMake/macros.cmake
jtkrogel: The dependent tests do not have the same name and are not processed the same by the labeling script.  They should just share the labels from the first call.

jtkrogel: The flow of test label information was done intentionally.

jtkrogel: I think so.  What labels does the script report for these when you run it?

ye-luo:
After talking to @jtkrogel return TEST_LABELS is necessary in the current way of handling labels. I need to redo a PR.

PR #1165, opened 2018-11-19
Improve test labeling and ctest script

ye-luo:
Label all the QE workflow tests as "converter".

This also fixed the default track in ctest_script.cmake. If this is not specified, the last portion of test results are submitted to the last track which is "Unstable" and it explains why most of the tests are under the Unstable track. With this change, tests without labels "unit|performance|converter|unstable" will go back to the Nightly track.
We will rearrange these tracks later if needed.


CMake/run_qe.cmake
```diff
@@ -10,10 +10,6 @@ FUNCTION( ADD_QE_TEST TESTNAME PROCS TEST_BINARY NPOOL WORKDIR TEST_INPUT)
     ENDIF()
     SET_TESTS_PROPERTIES( ${TESTNAME} PROPERTIES ENVIRONMENT OMP_NUM_THREADS=1 PROCESSORS ${PROCS} PROCESSOR_AFFINITY TRUE WORKING_DIRECTORY ${WORKDIR} )
     SET_PROPERTY( TEST ${TESTNAME} APPEND PROPERTY LABELS "converter" )
-    
-    SET(TEST_LABELS "")
-    ADD_TEST_LABELS( ${TESTNAME} TEST_LABELS )
-    
 ENDFUNCTION()
```
jtkrogel: Why did you remove the call to ADD_TEST_LABELS here?

ye-luo: The QE tests needs to be under the converter track only. QE tests take three steps pwscf, softlink, qmc. The qmc part picks up other labels because it shares the same generation function as other integration tests. I can't modify it. Before the PR change, labels are applied on pwscf and qmc parts but not the softlink one. It is not consistent and tests should fail because of the missing softlink. So I just want to keep it simple and remove the labels on the pwscf part.

PR #1168, opened 2018-11-23
Remove unused variables

prckent:
Partial source code cleanup based on warnings issued by PGI compiler. 

Did not remove "set but not used" or other more complex or serious warnings.

Full test suite is running. Partially through all the short tests. Will update here when complete.

### DIFF
--- a/src/ADIOS/ADIOS_config.cpp
+++ b/src/ADIOS/ADIOS_config.cpp
@@ -19,7 +19,7 @@
 static bool UseHDF5 = false;
 static bool UseADIOS = false;
 static std::string adios_xml_filename;
-const static std::string empty("");
+static const std::string empty("");
 static bool rdADIOS = false;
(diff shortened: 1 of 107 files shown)


ye-luo:
Does this PR fix the unit test failing with PGI?
We currently suppress unused-value warning also with clang

prckent:
This PR does not address the unit test failure. It only addresses the simplest class of warnings produced by PGI with the flags used when @markdewing added it.  In particular it does not address the warnings that indicate we have poor C++ apart from a couple of definitions that I fixed. This PR should not change any functionality. 

(General comment: Eventually QMCPACK should build warning clean with multiple compilers and with a reasonable amount of pedantry enabled. i.e. With many more warnings enabled that we have today. Buried in some of the trivial warnings are several possible bugs. We are not yet taking advantage of the help we could be getting from the compilers or other tooling.)

PR #1169, opened 2018-11-24
Fix precision issue in 3D natural spline unit test

prckent:
Fixes failing PGI unit test https://cdash.qmcpack.org/CDash/testDetails.php?test=3571223&build=32626

The unit test of the natural (non-periodic) 3D spline gives a NaN with PGI for a grid point exactly on the outermost spline grid limits. By moving the test point inside the limits by 5\*epsilon we avoid a presumed bounds error inside einspline and the NaN return value. 5\*epsilon is required to pass; 4*epsilon still results in NaN.

### DIFF
--- a/src/einspline/tests/test_3d.cpp
+++ b/src/einspline/tests/test_3d.cpp
@@ -21,6 +21,7 @@
 
 #include <stdio.h>
 #include <string>
+#include <limits>
 
 using std::string;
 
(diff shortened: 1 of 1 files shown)


ye-luo:
The einspline evaluation functions require [start, end) in the input.
The multiple-spline evaluation functions has out-of-bound protection but the single-spline evaluation function doesn't implement it. This is a bug. Fortunately, the only place we use single-spline evaluation is MPC which only support periodic and the std::floor is applied on coordinates before calling the evaluation. Then it is safe.

ye-luo:
We are currently at the transition from einspline(C) to spline2(C++). We probably should also move the single spline evaluation to spline2.

prckent:
I could not find a documented spec for the spline bounds and had to read the implementation. This is a bug. We should be clear on behavior in spline2.

ye-luo:
Neither do I. I didn't find any detail in the original einspline library documentation. I figured it out eventually after battling the open/periodic BC multiple times.

PR #1170, opened 2018-11-25
Delayed update on CPU

ye-luo:
This feature can be enabled with all the QMC methods implemented in QMCPACK
via `<slaterdeterminant delay_rank="32">`

Manual, unit test and integration tests are added.

Performance tests are set with `delay_rank="32"`. Currently the input files are generated on the fly.
If they needed to be stored in the repo, I will add them after the review in order to reduce the amount of files.


ye-luo:
1. Will do. I will be more strict on public/private.
2. Could you comment in-source? I'm confused about which macro you refer to.
3. The diamond test has 8 electrons in each spin and the delay was made 5.
4. Will do.
5. Yes. It can be used to access the delay and will be ported to miniQMC.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -560,7 +560,15 @@ bool SlaterDetBuilder::putDeterminant(xmlNodePtr cur, int spin_group, bool slate
     }
 #endif
   }
-  adet->set(firstIndex,lastIndex-firstIndex);
+  if( delay_rank<=0 || delay_rank>lastIndex-firstIndex )
+  {
+    APP_ABORT("SlaterDetBuilder::putDeterminant delay_rank must be positive and no larger than the electron count within a determinant!\n");
```
markdewing: The error message should print the values of delay_rank and lastIndex-firstIndex.

ye-luo:
The previous rank-1 is within this implementation. it is IDENTICAL to the old code if rank_delay=1. There is no need to have two implementation if one can do both work.

prckent:
Comments focus on making DelayedUpdate more accessible to our future selves:

- Doxygen comments needed for all functions in DelayedUpdate.h
- Describe all variables in the state machine in DelayedUpdate.h
- Describe how the state machine works in DelayedUpdate.h (in the description of one of the functions would be fine)
- Is there a tested reason for the large functions with inline hints in DelayedUpdate.h? The moment these are non-trivial in size or have function calls such as BLAS, they should probably be in a .cpp. We do this a lot across the application.
- Add comments about the logic in DelayedUpdate.h. e.g. What does delay_count=0 mean/do?
- Please can you comment the unit test. How are delays checked? (The obvious thing to do here is update one set of matrices rank 1 and another with a delay, and check the ratios, gradients etc.)
- set should be called something more descriptive. e.g. resetFirstParticleAndDelay
- This looks to be direct translation of Tyler's algorithm+implementation (great). Any sneaky extra tricks for efficiency not in that paper should be highlighted.


More general Q:

- Do we ever resize a determinant in the entire application? (Not relevant for this PR, but it is extra complexity that might not be used.)

prckent:
Thanks Ye. Yes set() is part of a larger mess.

I think short-diamondC_2x1x1_pp-delayed_update-vmc_sdj-1-16* needs test labels added. Happily this already runs a smidge faster than the non-delayed version, at least on oxygen.

prckent:
What are the rules around calling completeUpdates ?

ye-luo:
1. We should not reference performance on small tests unless it is 2x or more. Once the PR is merged, the performance tests will reflect the performance gain.
2. tags are not controlled by me and I think we will visit later when categorizing more tests.
3. I added comment in WaveFunctionComponent.h about completeUpdates. It must be called after each substep or step during pbyp move. For each step, we will evaluate hamiltonian and Ainv must be updated. The algorithm allows moving electron in any order but doesn't allow delay the same electron twice. For this reason, it must be called also after each substep.

ye-luo:
Can we merge this? The change in this PR is quite local and clean.

prckent:
Will discuss with @PDoakORNL this afternoon.

PR #1171, opened 2018-11-26
Fixing error in Basiset. Indexing of unique Atomic Basisets was messedup

anbenali:
Bug in the creation of pyscf HDF5.
When removing duplicated basisset (in the case of multiple similar atoms) names of the atoms were entangled.  

Fixed now.


prckent:
OK to test

PR #1172, opened 2018-11-26
Cusp Correction for SOA - Serial

### LINKED ISSUE #1150: SoA cusp correction implementation
Desired but not essential for 3.6.0 (See also #861).

anbenali:
- Cleaned print statments from Mark 
- Store xml file with Coefficients
- Generate Coefficients if not found
- Serial implementation for now 

closes #1150


src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
```diff
@@ -103,8 +103,8 @@ namespace qmcplusplus
   }
 
 
-  LCAOrbitalBuilder::LCAOrbitalBuilder(ParticleSet& els, ParticleSet& ions, Communicate *comm, xmlNodePtr cur)
-    : SPOSetBuilder(comm), targetPtcl(els), sourcePtcl(ions), myBasisSet(nullptr), h5_path(""), doCuspCorrection(false)
+  LCAOrbitalBuilder::LCAOrbitalBuilder(ParticleSet& els, ParticleSet& ions, xmlNodePtr cur)
```
ye-luo: This is a regression.

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
```diff
@@ -310,6 +310,7 @@ namespace qmcplusplus
         if(it == ao_built_centers.end())
         {
           AOBasisBuilder<ao_type> any(elementType, myComm);
+          any.setReportLevel(ReportLevel);
```
ye-luo: This is a regression.

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
```diff
@@ -389,6 +390,7 @@ namespace qmcplusplus
       if(it == ao_built_centers.end())
       {
         AOBasisBuilder<ao_type> any(elementType,myComm);
+        any.setReportLevel(ReportLevel);
```
ye-luo: This is a regression.

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
```diff
@@ -516,32 +670,28 @@ namespace qmcplusplus
     LCAOrbitalSet *lcos = nullptr;
     LCAOrbitalSetWithCorrection *lcwc = nullptr;
     if (doCuspCorrection) {
-      lcwc =new LCAOrbitalSetWithCorrection(sourcePtcl, targetPtcl, myBasisSet, rank()==0);
+      lcwc =new LCAOrbitalSetWithCorrection(sourcePtcl, targetPtcl, myBasisSet, ReportLevel);
       lcos = lcwc;
     } else {
-      lcos=new LCAOrbitalSet(myBasisSet,rank()==0);
+      lcos=new LCAOrbitalSet(myBasisSet,ReportLevel);
```
ye-luo: This is a regression.

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
```diff
@@ -516,32 +670,28 @@ namespace qmcplusplus
     LCAOrbitalSet *lcos = nullptr;
     LCAOrbitalSetWithCorrection *lcwc = nullptr;
     if (doCuspCorrection) {
-      lcwc =new LCAOrbitalSetWithCorrection(sourcePtcl, targetPtcl, myBasisSet, rank()==0);
+      lcwc =new LCAOrbitalSetWithCorrection(sourcePtcl, targetPtcl, myBasisSet, ReportLevel);
```
ye-luo: This is a regression.

ye-luo:
In the mixed precision case, we probably need to do the cusp correction construction in full precision while do the evaluation in single. Do we have a plan how to do this?

prckent:
Re: mixed-precision. I am also interested in the plan, but for an initial implementation I wouldn't object if we only supported double. A working executable path is needed in SoA to help debug #1145 .

prckent:
OK to test

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
anbenali: Fixed

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
anbenali: Fixed

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
anbenali: Fixed

anbenali:
Keep in Mind that the point is to have this available for now for anyone "stuck with a bug" and also to help find errors if any. 

I am making the PR because I needed this code, but all the credit goes to  @markdewing  who made 99.99% of the work. (which means I cannot make the doc for the code)

markdewing:
I will write some documentation for these functions.

prckent:
- Thanks @markdewing 
- @anbenali : remove the WIP when you are finished updating

anbenali:
@prckent I am temptd to stage this in 2 separate PRs. one for Serial and one for parallel implementation just because of the urgency of the matter. I am testing the code on a very large set of molecules and in some cases there is  significant drop in the variance and multidet is slightly stabilized from large fluctuations. 
If that is fine with you I rather do it that way. Otherwise it will be one more week I guess before I can touch the parallel part..

prckent:
@anbenali I assumed this was your plan from the PR name. When the serial implementation is in we can already begin to learn about #1145 and other surprises. Running faster comes later.

prckent:
@markdewing Do you plan to update the code in this PR with some comments, or should we merge now and update later?
@anbenali is this ready to merge apart from the docs? Remove the WIP if so.

markdewing:
I'm working on some comments, documentation, and removing some unused code.  I can probably get that into this branch today.

ye-luo:
Could you please move the test/CuspCorrection/CHN to tests/molecules/CHN_ae?

prckent:
Since users tend to go fishing for inputs, please can you remove the strangeness:

    <parameter name="walkers">1</parameter>
    <parameter name="samplesperthread">1</parameter>
    <parameter name="stepsbetweensamples">10000</parameter>

walkers=1 might have a purpose that I don't appreciate here, but normally we don't specify this in VMC tests. Inputs that don't rely on samples shouldn't specify anything to do with samples.

prckent:
How did you generate the reference error bar on the new tests?

Please generate the reference error bar on the new test by running e.g. 100x longer (or more) than the test, and rescaling appropriately. I assume that this is very fast to run and would also be a good check of the code. While it does not matter much for error bars compared to reference values (reliable deterministic PySCF data used here), @jtkrogel 's observation is that we haven't been careful enough in many of the more recently added tests. Even if not used today it is good to generate these values when the test is generated.

anbenali:
@prckent 
Ref data (inputs and outputs)  is always provided in the qmc-ref directory. Calculations were run for significantly more than 100X.  Ref data from qmca is in the CMAKELIST.txt  at the top of the file (commented). short run value is also provided in the qmc-ref directory. 
Difference between ref run and short run is:
abenali@abenali:~/Work/Development/QMCPACK-Benali/qmcpack/tests/CuspCorrection/CHN/qmc-ref$ qmca -q ev *.scalar.dat

vmc-ref            series 0  -93.339342 +/- 0.000215   2.379711 +/- 0.004316   0.0255 
vmc_short-ref  series 0  -93.319389 +/- 0.010942   2.320352 +/- 0.092366   0.0249 



Are you refering to something else? 

(Removed the Walker=1)

ye-luo:
@anbenali Please keep reference input untouched and only change the input file for tests.

jtkrogel:
@anbenali 
Reference value and error bar should both come from the longest reference run.

Reference error bar for short test can be calculated from the long reference run by rescaling: e.g. if reference run is 100x longer than short, then reference error bar for short test is: 0.000215*sqrt(100+1) = 0.00216.

Reason is basically that short runs produce noisy error bars and thus tests fail with actual frequency deviating from 3 sigma.

Apologies if this is what you did already, your reference of 0.011 looks like it was taken directly from a single short run.

jtkrogel:
@anbenali I also notice that your long reference run uses 200 blocks.  This will also create a variable errorbar.  For generation of reference error bar many more blocks is desired; use 10000 instead (and fewer steps of course).  This will reduce the error bar of the error bar by a lot and also improve estimates of the autocorrelation time.

anbenali:
@jtkrogel regarding the reference containing 200 blocks, this was the past requirement to avoid long files in the repo. got my previous PR blocked by @ye-luo  for doing what you are suggesting. can we have a consensus here?

jtkrogel:
The committed example "ref data" can be short, but the number of blocks actually used to create the reference data and the ref xml file must have more blocks.  I usually include scalar.dat files from a run w/ lower blocks (typically 1000) in the repo and add a README file documenting all the reference data (optional?).

jtkrogel:
@anbenali phone sometime?  Better discussed interactively than through comments here.

prckent:
What was the outcome of the discussions? Merge of this PR is officially pending some simple updates currently. We could choose to require them in a later PR (required for the release).

anbenali:
Hopefully the changes will please all parties..

tests/CMakeLists.txt
```diff
@@ -43,4 +43,5 @@ else()
     SUBDIRS("heg/heg_54_J2rpa")
     SUBDIRS("molecules")
     SUBDIRS("solids")
+    SUBDIRS("CuspCorrection")
```
ye-luo: No more need this

Issue #1068, opened 2018-09-12
Correlations in random number stream based on clock seeding


jtkrogel:
I've noticed issues like this before, but the problem became clearer when performing the 1000x runs for #1067.  Of the 1000x runs which should have been independent, fully 30% were carbon copy reproductions of others.

This came about for two reasons:
1) (1/3 of cases) The random offset, and hence the seed, appears to be set by the second (at least on the machine I ran on, so probably true for others).  Since four runs could be executing at any given time in the queue I was in, occasionally a pair or more would start in the same second.  This should have significant implications (statistical bias) for doing QMC in a "high-throughput" fashion on a larger machine.

2) (2/3 of cases) The random offset appeared to have a maximum range of 1024 (it rolled back to zero after this) so the available seeds would repeat every 20 minutes.  Since the runs started over the course of about 45 minutes, some runs submitted much later also had the same seed profile as earlier ones and were identical to them.

In addition to the carbon copy problem, runs having a different, but similar random offset (say 1020 vs 1021) resulted in highly correlated runs.  This effect is shown in the figure below, where the x axis is run number (0 to 999, zoomed in) which corresponds to different offsets, and each line is the value of the local energy from a particular MC block.  As is evident from the figure, several similar offsets lead to similar, but not identical, results.
![rerun_correlation](https://user-images.githubusercontent.com/24547905/45430508-77ab2800-b673-11e8-83b2-f33eb0d6b125.png)

The effect of reduced statistics due to hidden correlations is also directly evident from using the 1000x runs to produce a mean that should be equivalent to a single run 1000x longer.  The deviation between the two results is 2.6 sigma, with a probability of occurrence of 1/100.

Does anyone know where the offset range of 1024 comes from?

What could be done to avoid statistical bias in high-throughput runs?

Do we know that something like this is not already happening across MPI ranks?

ye-luo:
1024 is from
https://github.com/QMCPACK/qmcpack/blob/d80d1b608105ece8362eeea1dc8c069440dca477/src/OhmmsApp/RandomNumberControl.cpp#L80
The following steps are take:
1. The per rank RNG generates 1 number (offset base) and bcast to all MPI ranks (1024 is found here)
2. Generate prime numbers up to nprocs*omp_get_max_threads()+2)+offset. And use the first nprocs ones to reinitialize per rank RNG.
3. Initialize per thread RNG with other prime numbers.

[...]

PR #1174, opened 2018-11-27
[WIP] Change RNG seed generation scheme.

ye-luo:
step 1, get initial seed by time std::chrono
step 2, generate seeds for RNGs used by MC. Each rank needs omp_get_max_threads()+1 seeds. jump ahead rank()*(omp_get_max_threads()+1) with discard(). Very fast 0.14 sec for 10^8.
step 3, seed the per-rank RNG.
step 4, seed the per-thread RNG.

After dive in the code a bit, I feel a lot of cleaning is needed.
Mostly RandomGenerator_t needs to be simplified. All the information about MPI and threads should be removed.

### DIFF
--- a/src/OhmmsApp/RandomNumberControl.cpp
+++ b/src/OhmmsApp/RandomNumberControl.cpp
@@ -77,22 +77,30 @@ void RandomNumberControl::reset()
 }
 
 /// reset the generator
-void RandomNumberControl::make_seeds()
+void RandomNumberControl::make_seeds(bool init_from_time, uint_type time_seed)
 {
+  // step 1, get initial seed by time
+  if(init_from_time)
+  {
+    time_seed = std::chrono::system_clock::now().time_since_epoch().count();
+    app_summary() << "  Offset for the random number seeds based on time: " << time_seed << std::endl;
+  }
+  mpi::bcast(*OHMMS::Controller,time_seed);
+  // step 2, generate seeds for RNGs used by MC. Each rank needs omp_get_max_threads()+1 seeds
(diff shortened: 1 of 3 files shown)


prckent:
1. Agree that lots of cleaning is needed.
2. I would prefer that we don't touch this shortly before a release and only touch it once. There are many infamous incidents of initialization being done incorrectly.
3. We should move to a scheme that guarantees independent random numbers i.e. different seeds per thread. There is a formally correct way to do this that I outlined earlier today.

prckent:
See #1176

jtkrogel:
What is the typical resolution of "high_resolution_clock" in std::chrono?  

Would really like the probability of repeated clock based seeds to be near zero (e.g. no repeats as in #1068, but guaranteed for an even larger set of runs).

prckent:
I have some better ideas on how to handle this now. A good topic for v4 but we need the batched code first.

prckent:
(Commenting before the stale issue action gets here: it still worth reminding ourselves that the original seeding algorithm in QMCPACK should be replaced, and likely the random generator replaced with a more modern one. This is a delicate topic so whenever this gets tackled there should be discussion. Modern generators have fast skip-forward capabilities that can be exploited to ensure every sequence is distinct. It is also worth investigating an algorithm with less state; the current random.h5 are surprisingly large. Ideally ensemble runs should be rerunnable independently, i.e. enough info printed out to reseed them identically. Keeping the current generator around as an option might preserve some poor structure, so we might replace it + update all the deterministic tests.)

PR #1179, opened 2018-11-27
Update Catch to Catch2

markdewing:
The main purpose for the update is to potentially prepare for an MPI reporter to handle parallel unit tests.  The Reporter interface should be more stable in Catch2 (even if not completely fixed) than the interface from Catch 1.x.

Catch2 internals are simplified by dropping all pre-C++11 code.


markdewing:
Yes.   The gcc 5 host compiler supports c++ 98 by default, but Catch2 needs c++11.
The fix will be either to add a c++11/14 flag to nvcc, or to change the tests to keep Catch out of the .cu files (I haven't look at the latter yet).

PR #1162, opened 2018-11-16
Build and nexus recipe for Cades cluster at ORNL

kayahans:
Build script and additions in Nexus for ORNL local cluster Cades.

PR #1191, opened 2018-11-29
Cleaned build and nexus recipe for Cades cluster at ORNL 

kayahans:


nexus/library/machines.py
```diff
@@ -195,7 +195,9 @@ def __init__(self,
                  machine      = None,
                  account      = None,
                  queue        = None,
-                 bundled_jobs = None,
+		 group_list   = None,
```
markdewing: Indentation here still looks funny - mixed spaces and tabs?

markdewing: Now  I see it's in the argument list where indentation is not significant.    It would be nice if the indentation matches the surrounding code, though.

nexus/library/machines.py
```diff
@@ -238,6 +240,8 @@ def __init__(self,
         self.name        = name
         self.type        = type
         self.queue       = queue
+	self.qos         = qos
```
markdewing: Should use spaces rather than tabs to match the surrounding code.

markdewing:
Okay to test

markdewing:
The machines.py in this PR is identical to the machines.py in #1162 , with all the same merge conflicts.  It looks like this branch was taken from the same starting point as the branch in the other PR, rather than a freshly updated 'develop'.

markdewing:
Looking at the history (https://github.com/kayahans/qmcpack/commits/cades/nexus/library/machines.py ), the last common point is 75c1a84cfcb657b6ee74a14e9e82a27b1db89da8
A minimal diff of the changes can be obtained with:
cp machines.py new_machines.py
git checkout 75c1a84cfcb657b6ee74a14e9e82a27b1db89da8
cp new_machines.py machines.py
git diff machines.py > diff.txt


The tab indentations can be fixed in diff.txt, and this can be applied to an up-to-date version of machines.py with 'git apply diff.txt'

jtkrogel:
CADES does succeed OIC5 right now for practical purposes, however, OIC5 will still be in use in the future so it should not be removed yet.

ye-luo:
No problem. Just keep it until it goes offline.

prckent:
Made a clean PR. Closing.

Issue #1192, opened 2018-11-30
RadialJastrowBuilder adds extra KContainer initialization


ye-luo:
It seems that RadialJastrowBuilder does some extra work.
See below after printing the parameter. The old builder doesn't print "KContainer initialised..."
```
  RadialJastrowBuilder adds a functor with cusp = -0.2500000000
 size = 10 parameters
 cusp = -0.2500000000
 rcut = 5.5727791786
Parameter     Name      Value
uu_0 0.2862235606 1 1  ON 0
uu_1 0.1947736889 1 1  ON 1
uu_2 0.1319544911 1 1  ON 2
uu_3 0.0889339447 1 1  ON 3
uu_4 0.0569557585 1 1  ON 4
uu_5 0.0356595851 1 1  ON 5
uu_6 0.0220695026 1 1  ON 6
uu_7 0.0129608642 1 1  ON 7
uu_8 0.0066010072 1 1  ON 8
uu_9 0.0027871444 1 1  ON 9
  KContainer initialised with cutoff 9.4916334152
   # of K-shell  = 3523
   # of K points = 113160

  NUMBER OF OPT_BREAK KVECS = 31026066914
 finding kc:  1.5418392420 , -1.0000000000
  LRBreakp parameter Kc =1.5418392420
    Continuum approximation in k = [9.4916335819,616.7357058014)
```

lshulen:
OK, I'll take a look.

lshulen:
OK, It looks like I was initializing with RPA when this was not requested.  The previous builder checked to see if this was asked for in the xml.  I'll add the fix now.  Is it OK to have a single pull request that fixes this and the J1 cusp issue?

prckent:
It is preferred to keep things separate but since the fixes and likely small and easily reviewable I think whatever route is expeditious.

Issue #1175, opened 2018-11-27
short-H4-opt-OneShiftOnly-1-16 failing since Nov 20


ye-luo:
Nov 19 to 20 night
https://cdash.qmcpack.org/CDash/queryTests.php?project=QMCPACK&date=2018-11-20&filtercount=1&showfilters=1&field1=testname&compare1=63&value1=short-H4-opt-OneShiftOnly-1-16
Nov 20 to 21 night
https://cdash.qmcpack.org/CDash/queryTests.php?project=QMCPACK&date=2018-11-21&filtercount=1&showfilters=1&field1=testname&compare1=63&value1=short-H4-opt-OneShiftOnly-1-16

The problem seems from #1163, The diamond solid optimization was fine. short-H4-opt-OneShiftOnly-1-16 has J1 with cusp=1. createJ1 needs to handle cusp.

We need to introduce a regression test in the unit test of J1.

lshulen:
OK, will take a look now.  I thought I had handled this, but perhaps just for J2.

ye-luo:
Thank you @lshulen. Hopefully the fix will be simple. You can verify by running
`ctest -j16 -R short-H4-opt-OneShiftOnly-1-16`  a couple of times, if you don't get any failure. It should be good.

prckent:
Interested in what the issue turns out to be here. Perhaps we need to formalize checking the interface between the optimizer and the Monte Carlo.

ye-luo:
Now we actually have the capability of testing multiple sections of QMC methods in one run and thus having more granularity within a single test. Probably adding an energy check on the first VMC block is simple but very helpful.

prckent:
Ideally we have tests that run instantly that developers can use while developing. i.e. deterministic unit and integration tests with good enough coverage to catch these problems.

lshulen:
Another one that I have thought would be nice would be to have some unit type test for setting whether various parameters are on or off for optimization from the input file.  It shouldn't be something that really needs calculations to be performed, but I'm not sure how to do that in practice.

ye-luo:
As a regression test for this case, expand one of the unit test by adding cusp and then check the first element of the coefficient may be enough. https://github.com/QMCPACK/qmcpack/blob/4e650e01d7d738eb6a543371cbc4b955466d82b5/src/QMCWaveFunctions/tests/test_bspline_jastrow.cpp#L313

jtkrogel:
[...]

PR #1193, opened 2018-11-30
Fix jastrow

lshulen:
Fix issue #1192 and #1175 .  Also got help from @markdewing to add to the wavefunction unit test to make sure setting cusp for one body jastrow factors works properly.  This code should now pass short-H4-opt-OneShiftOnly-1-16

### DIFF
--- a/src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
+++ b/src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
@@ -237,6 +237,7 @@ bool RadialJastrowBuilder::createJ2(xmlNodePtr cur)
       rAttrib.add(spB,"speciesB");
       rAttrib.add(pairType,"pairType");
       rAttrib.add(cusp,"cusp");
+      
       rAttrib.put(cur);
       if(pairType[0]=='0')
       {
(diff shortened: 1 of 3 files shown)


src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
```diff
@@ -359,6 +366,10 @@ bool RadialJastrowBuilder::createJ1(xmlNodePtr cur)
       {
 	functor->cutoff_radius = 10.0;
       }
+      if (cusp > 0)
```
ye-luo: Is it better if we set 0 as default.
Here check if <0, abort with error. Then else always call setCusp.

ye-luo:
Okay to test.

src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
ye-luo: Change my mind a bit. Put 0 as default and allow negative cusp for protons.

lshulen:
Thanks Ye

Sent from my iPhone

On Dec 1, 2018, at 11:12 AM, Ye Luo <notifications@github.com<mailto:notifications@github.com>> wrote:


@ye-luo approved this pull request.

I backported the unit test to the old code and it also passes. So the fix is solid.

PR #1194, opened 2018-12-01
AFQMC PR

mmorale3:
Merges llnl::develop.


CMakeLists.txt
```diff
@@ -498,6 +498,40 @@ IF (ENABLE_GCOV)
   ENDIF()
 ENDIF(ENABLE_GCOV)
 
+# Require C++11 while leaving possibility to fall back by modifying this file
```
markdewing: This code should be unnecessary with the updated minimum cmake version and the move to cmake mechanisms to specific the C++ standard.

prckent: @mmorale3 : when I quickly scanned the updated cmakelists.txt it looked like it might actually be from an earlier version - are these updates actually need?

ye-luo:
I saw Communicate was directly replaced with the new wrapper class in the src/io but if I remember correctly, @markdewing was trying to put the new wrapper class contained inside Communicate. I think when @markdewing brings in his change. Things will be more clear.

I was hoping to have changes in src/io as a separate PR but this didn't happen. Some change is needed there but I will have to wait until basic builds+tests is provided. For this reason, please expect slow review.

ye-luo:
I saw 60MB h5 files in the test and it was deep in the commits. We need a cheap test.

CMakeLists.txt
ye-luo: Just remove this part. Everything is already on line 388.

markdewing:
Is it okay if I extract parts of this branch/PR and submit them as separate PR's?    The changes in Numerics look self-contained, for example.

mmorale3:
No problem.

mmorale3:
Everything seems fine. We can always fix any problems.

PR #1195, opened 2018-12-02
Define delta for assert in Quadrature.h

### LINKED ISSUE #1180: Undefined variable in Quadrature.h in Debug mode
#1168 commented out the definition of 'delta' in Quadrature.h.   This value is used later in an assert, which causes a compile failure in debug mode.  
The implementation of 'assert' is a macro that is empty in Release mode, so the values are never compiled or used.

Possible solutions
1. inline the definition of 'delta' into the assert.   Ugly.
2. Keep delta and live with the unused variable error in release mode.  Ugh.
3. Mark the value of delta as used to satisfy the compiler.   There are likely standard techniques or attributes to handle this.  This issue is a marker to research this and put it into the code.

prckent:
Closes #1180 

Ugly solution but this file already has other magic numbers defined similarly. I am surprised the second tolerance is not too small for large integration grids.

### DIFF
--- a/src/Numerics/Quadrature.h
+++ b/src/Numerics/Quadrature.h
@@ -236,16 +236,16 @@ struct Quadrature3D
     assert (xyz_m.size() == nk);
     assert (weight_m.size() == nk);
     double wSum = 0.0;
-    //    const RealType delta=2*std::numeric_limits<float>::epsilon();
+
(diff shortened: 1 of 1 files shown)


ye-luo:
I think it is better because the two places now allow different tolerance.

PR #1198, opened 2018-12-02
Simplify radial jastrow builder

ye-luo:
Mostly to reduce the nested templates.

Also change
```C++
      if (targetPtcl.Lattice.WignerSeitzRadius > 0)
      {
        functor->cutoff_radius = targetPtcl.Lattice.WignerSeitzRadius;
      }
      else if (functor->cutoff_radius < 10.0)
      {
        functor->cutoff_radius = 10.0;
      }
```
to
```C++
      functor->cutoff_radius = targetPtcl.Lattice.WignerSeitzRadius;
```
Because targetPtcl.Lattice.WignerSeitzRadius > 0 for any BC (PBC, OBC, mixed)
cutoff_radius is always checked inside the `functor->put()`.

Note: since we changed to RadialJastrowBuilder, spin polarized J1 is no more accessible. The old class was AoS. The solution will be expanding J1SoA to support the spin case and eventually becomes the only one.


lshulen:
I like almost all of the changes.  There is one thing that I'm not as happy with in the last commit.  In writing RadialJastrowBuilder, I spent a lot of time trying to understand the various includes and headers in BsplineJastrowBuilder and switched to the
#if defined(QMC_CUDA) and defined(ENABLE_SOA) business 
not because it was more elegant, but because it was a bit clearer to me.  What would you think of changing back to this?

Also, I agree about the spin polarized J1.  In fact, I think several of the headers involving SpinJastrowOrbitals could probably be removed.  I kept them around mainly to remind us to properly fix this in the future.

ye-luo:
I will change that part from the nested ifdef back  to
`#if defined(QMC_CUDA) and defined(ENABLE_SOA)`

Hopefully we can remove AoS and absorb DiffXX class inside XX soon.

ye-luo:
@lshulen Are you OK with my change now? If so, please use the reviewer option to approve this PR.

PR #1200, opened 2018-12-03
Excited state calculations with Nexus

kayahans:
With the functionalities here, one can:
- Plot DFT band structure using conventional primitive cell, parse eigenvalues, VBM/CBM and so on
- Find optimal tiling matrix for excitations involving any number of wavevectors
- Perform quasiparticle and optical gap calculations with QMCPACK and Nexus. 

For the first item above, a wrapper is written on SeeK-Path. (https://github.com/giovannipizzi/seekpath) 
Optimal tiling matrix for excitations is implemented using: https://doi.org/10.1103/PhysRevB.92.184301
The contribution also includes instructions on how to perform excited state calculations with QMCPACK and Nexus in their manuals. Examples in the manuals can be produced using this PR. 

Further improvement and testing:
- I think further checks can be useful for excitations where Gamma point is not involved. My initial idea is that using kshift can be sufficient. 
- Parsing the nscf output to find twist and band index of the excited orbitals is still mostly done by inspection, but I think it is better to resolve the item above first. 
- The conventional primitive cell is found for non-magnetic structures. This can be extended to magnetic systems. 
- Further cleanup and any feedback from Jaron or others to make easier workflows here.


prckent:
OK to test

PR #1204, opened 2018-12-03
Backflow SoA Migration.  1/2

rcclay:
This PR implements the 3rd derivatives of single particle orbitals required for backflow optimization using spline2 objects.  This should eliminate the need for the "use_old_splines" flag in the backflow determinant XML block, and thus the need for the legacy EinsplineSetExtended object.  Tested for real and complex AoS builds for now.  Hybrid not implemented.  

Known Issues:
1.)  Vectorization of assign_vghgh is clearly not correct, so the #pragma omp simd directive is commented out.

Future Commits:
1.)  2/3 Fix mixed AoS-SoA distance table issue.  This seems to be the main obstacle to running entirely within SoA.


ye-luo:
Will 2/3 and 3/3 land before the release?

rcclay:
Unclear.  If 2/3 is as straightforward as I think it is and that's the only issue, then sure...  but that's an unknown.  3/3 is pretty basic.

ye-luo:
Could you add the unit tests first in this PR?
The spline 2 test is at src/spline2/tests/test_multi_spline.cpp
SPO test is at src/QMCWaveFunctions/tests/test_einset.cpp

rcclay:
I can do that.  I'll get started on it tomorrow.

ye-luo:
Okay to test

ye-luo:
I removed `use_old_spline="yes"` in bcc 2x2x2 and 3x3x3 tests.
Please pull the code and do `cmake ..` before you try the ctest.
short-bccH_3x3x3_ae-deriv-1-1 is failing with AoS real build.

rcclay:
That commit should fix the problem with the 3x3x3 Real AOS build.  Working on unit tests now.

ye-luo:
Let us know when you finished adding unit tests.

ye-luo:
Please also check if simd is still a problem after your bugfix.

rcclay:
Unit tests are up.

rcclay:
I haven't paid attention to the previous discussions, but in line 172 of "src/QMCWaveFunctions/tests/test_einset.cpp", why is the epsilon there?  Is it necessary?  This gives me some issues with the Intel 17.0.1 with not being able to find the epsilon() function.

markdewing:
The epsilon is there to increase the tolerance of accepted values.   The first PR failed to expose epsilon properly - there's a second PR shortly later that should fix it.

rcclay:
@ye-luo SIMD issue still seems present in SplineC2C.  SplineC2R might work though... as in errors in the the wavefunction parameter derivatives of the local energy seem to be between 1e-8 and 1e-9 across the board.

markdewing:
Also, Intel 17 is going to be unsupported soon.   It fails to compile some specific C++14 features in the MPI wrapper, and it was getting too hard to make workarounds.

rcclay:
@markdewing Good to know.  Thanks!

rcclay:
Cool.  Rhea freaked out at the last added unit test.  I need to get my ORNL account updated to access it.  If we're in a hurry, how hard would it be for someone to execute "ctest --verbose -R deterministic-unit_test_wavefunction" over there?

ye-luo:
Probably you have already read the output of Build # 830. It seems the numerics is a bit off.
I did check last night on AMD ryzen and gcc compiler 7.3. Your unit tests pass. So I'm wondering if this is an issue specifically on rhea. I would put a macro and disable the unit test and we merge the code. You add an issue and investigate it when you get the access to ORNL accounts.

rcclay:
@ye-luo Just got the Build#830 report, and I was afraid of something like that.  I've disabled the test_einset unit tests until I can track down.  I can't reproduce it with a Xeon + Intel 17.0.1.  I'll submit an issue for this.

markdewing:
Some of the previous epsilon changes to test_einset were triggered by clang 7.

PR #1208, opened 2018-12-04
Add qmc to tool names

markdewing:
As discussed in #1020, include a 'qmc' in the executable names.

Also change getSupercell to lowercase-with-dashes style to be more consistent.


prckent:
@jtkrogel Do any parts of NEXUS need to be updated?

jtkrogel:
I would prefer that we just make symlinks upon install with "qmc-" added to the normal name of the tool.  We are doing this just to avoid namespace collisions, right?

markdewing:
Yes, this is to avoid collisions when installing into a shared bin directory.   Symlinks will not help with this  - the normal tool still needs to be put into the bin directory.
I did a quick grep in the nexus library directory and got no hits - I don't think nexus uses these tools.

prckent:
@jtkrogel Lets not have two names. This will be confusing. Since these are mainly legacy tools, I think it is reasonable to rename them as Mark has done. We have to rename them at some point, might as well do it today.

jtkrogel:
Nexus has executables, a couple mainly associated with QMCPACK (qmca, qfit), and a couple for its own use (sim, redo).

Rename via copy into shared bin is OK, but I really don't want to prefix these (for ease of typing) in the parent directory.

jtkrogel:
Is this just for legacy executables?  I see you passed over convert4qmc (no qmc-convert4qmc).

markdewing:
convert4qmc already contains 'qmc'.   It's not the same pattern, but convert4qmc is more heavily used under that name and renaming it would be more involved.

prckent:
This update reduces ambiguity and possible namespace collisions, so is moving in a good direction. Approving.

jtkrogel:
I don't mind with the legacy qmcpack tools.  I also could accept updating qfit/qstat/qtest to qmc-fit/qmc-stat/qmc-test, if desired.  Less sure about qmca (qmc-a?), but I guess if this is more about presence of "qmc" than prefixing, I guess it is covered.

Also less sure what is needed for the purely Nexus tools.  "ntest" was patterned after "ctest".  If we are going for additional prefixing, something like "nx-sim", "nx-redo", and "nx-test" might be consistent.

jtkrogel:
OK, other updates can go in other PR's.  Move discussion back to #1020?

PR #1214, opened 2018-12-04
ORNL CADES support

prckent:
@kayahans Please check that I did not miss anything when I rebuilt this PR with a clean history.

(Replaces #1191 )


ye-luo:
Please also add -DCMAKE_C_FLAGS="-xCOMMON-AVX512" -DCMAKE_CXX_FLAGS="-xCOMMON-AVX512" for the skylake processor because the complier generates AVX2 by default even on Skylake.

prckent:
@PDoakORNL What are your thoughts on these modules? Using this script from my account picks up horrifically old software.

ye-luo:
Since MKL is used, why fftw module is needed? If you have libsci loaded by default, unload it is necessary.

kayahans:
@ye-luo is right, we can unload fftw. openmpi/3.0.0 is also apparently not used since PE-intel loads the following: 1) intel/16.0.1     2) openmpi/1.10.3   3) xalt/0.7.6       4) PE-intel/1.0

prckent:
@kayahans Did you try to use Intel 2018?

kayahans:
@prckent No, it was recommended in CADES website to use PE modules. Intel2018 was added late October after the update. I can try if you would like.

kayahans:
As I remember now, it was because of this:
module load boost
Please load a programming environments (PE):
PE-gnu, PE-intel, or PE-pgi

prckent:
Thanks. I am checking the installed options. gcc is also needed for the C++14 library.

prckent:
(No more updates planned)

PR #1215, opened 2018-12-05
Add install for qmcpack binaries

markdewing:
Also install the qmcpack.settings file.

Addresses the qmcpack binary part of #1020


ye-luo:
Could you add some documentation in README and manual?

prckent:
Thanks. Happy to merge after a small update to the manual.

PR #1217, opened 2018-12-05
Update ORNL OLCF Eos scripts

prckent:
Clones the repo and then submits independent jobs for each build type to work around job and process limits.


ye-luo:
Could take the chance to update ctest_script.cmake by removing eos settings and put them in the job script.

prckent:
I have no bandwidth to do so at this time.

PR #1220, opened 2018-12-06
Nexus PySCF interface

jtkrogel:
This PR contains a Nexus interface to PySCF.  The idea is to be able to drive arbitrary PySCF scripts (based on template files), with the possibility for follow-on QMCPACK calculations.

Generation of Mole/Cell input and the additional conversion step via savetoqmcpack are automated.

Anticipate completion in a day or so.

Tasks:
- [x] Create PyscfInput class to generate/write PySCF scripts based on template files
- [x] Create null PyscfAnalyzer and main Pyscf simulation classes
- [x] Create orbital workflow dependency for PySCF -> QMCPACK 
- [x] Verify that Mole/Cell input generation works
- [x] Verify that managed PySCF runs work
- [x] Verify that managed PySCF+QMCPACK runs work
- [x] Final adjustments for production runs


jtkrogel:
Ready for review.

markdewing:
okay to test

markdewing:
Are examples and/or tests coming?

One of the problems with Nexus is discoverability.  It has a huge number of features and it's hard to find out what they are.   Nexus doesn't document the features of the code it drives (nor should it - that's far too involved), so one has to read the documentation for the target code, and then figure out how to get Nexus to generate that feature.
One problem with code generators or template systems is trying to visualize and understand how changes in the Nexus input change the generated code.   Or as mentioned above, the inverse problem of knowing the desired result in the generated code and figuring out how to change the Nexus input to get that result.

markdewing:
Do you want to include PySCF examples in nexus? Or something else - what do you want to do that involves license policy?

prckent:
Why the license concern? We shouldn't be copying any files from PySCF to have a clean situation. Any examples similar to those in PySCF should be cleanly rewritten.

jtkrogel:
PySCF has a very large number of examples.  Demonstrating how to do some of these with Nexus would be ideal--not copying but basing new files (e.g. the templates) off of them, with any attribution as required.  A situation like this, AFAIK, is a good example of why codes are released as open source.

Apache considers UIUC to be similar to its own license: [http://www.apache.org/legal/resolved.html#category-a](http://www.apache.org/legal/resolved.html#category-a).  There does not appear to be a large risk so long as the requirements are observed.  Is there one I am missing?

Apart from risk, I can understand not wanting to do the added work of meeting the requirements.  I'm quite open to everyone's advice on this.

prckent:
Is there anything in this PR from PySCF covered by their license? Yes or No?

jtkrogel:
Anything else needed here?

prckent:
Thanks Jaron. Just wanted to be sure.

PR #1222, opened 2018-12-06
Add a safe guard to updateInvMat

ye-luo:
I didn't see it causing problem but I feel it more safe to add this guard. It costs nothing.

### DIFF
--- a/src/QMCWaveFunctions/Fermion/DelayedUpdate.h
+++ b/src/QMCWaveFunctions/Fermion/DelayedUpdate.h
@@ -41,14 +41,18 @@ namespace qmcplusplus {
 
       /// pointer to the row of up-to-date Ainv
       const T* Ainv_row_ptr;
-      /// electron id of the up-to-date Ainv_row, checked by ratioGrad
-      int Ainv_row_ind;
+      /** row id correspond to the up-to-date Ainv_row_ptr. [0 norb), Ainv_row_ptr is valid; -1, Ainv_row_ptr is not valid.
+       *  This id is set by getInvRow indicating Ainv_row_ptr has been prepared for the Ainv_row_id row
+       *  ratioGrad checks if Ainv_row_id is consistent. If not, Ainv_row_ptr needs to be recomputed.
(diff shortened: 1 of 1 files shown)


markdewing:
Could you say more about what makes this safe? (Or what was potentially unsafe about not having it set?)   Right now, with the comment in the code, I assume there is something very tricky going on and need to study all the called code carefully so see where Ainv_row_ind is updated between here and where it is set again later.

ye-luo:
It is used to check if getInvRow was called for the same electron when ratioGrad is called.
If they matches, the code will not recomputed AinvRow.
Once a move is accepted or all the pbyp move is done. Ainv_row_ind is marked -1 for safety.
I'm thinking of a case that getInvRow is called but there is no accept. Ainv_row_ind is left not -1.
If the code start to work on a second walker and calls ratioGrad and by chance the Ainv_row_ind matched even getInvRow was not recomputed. The AinvRow from the last walker will be used and things go wrong. However, in reality evalGrad is always called before ratioGrad. Ainv_row_ind is always refreshed.

I feel more safe putting -1 when a move is accept (we ready have it) and all the moves are completed (added in this PR).

prckent:
Obligatory comment: feelings about safety of code points to insufficient testing and poor design in overall the application. We are getting better but there is along way to go.

prckent:
I agree that the comment simply needs improving here.

(For the future, and hopefully becoming less relevant as we remove state information: In general for state machines with exposed functions that can only be called in certain sequences, the state machine should have an explicit internal state where each state is labeled. Functions then assert() that the state machine is in an acceptable state before doing any work. This exposes the logic of the state machine in the source code and makes visible something which is otherwise hidden. If the state machine is complex or important enough, an explicit transition state table can be built to avoid any ambiguity. )

ye-luo:
@markdewing I added point to point reply directly in your comment. Thank you so much for commenting at this level of detail.
@prckent I agree with you, we should document in source state machines as clear as possible. However, in this case, the state goes from -1 to Norb which can't be put in a table.

Issue #1207, opened 2018-12-04
numpy matmul used by NEXUS is not available in older numpy


prckent:
The updated library/structure.py merged last night uses matmul which is not always available. https://cdash.qmcpack.org/CDash/viewTest.php?onlyfailed&buildid=33899

Running develop by hand on oxygen I can confirm that the tests are also broken there.

The library/gaussian_process.py code also uses it, but we don't run that in the tests yet.

Suggestions: add minimum version checks; use our own implementation (either always or when not available);

jtkrogel:
Sounds like it should be replaced by the standard "dot" function if possible.  

@kayahans is the code making use of any matmul functionality that differs from dot?

jtkrogel:
OK, I will change it to dot then.

jtkrogel:
Resolved by #1210.  Closing.

PR #1223, opened 2018-12-06
Nexus: fix issues with CADES

jtkrogel:
Issue #1207 was (and is) resolved by PR #1210 (see nightlies for Dec. 5).

This PR fixes issues recently introduced by PR #1214.  Two issues are fixed: 1) correct indentation of if blocks in the Cades class (this was causing import failures of the machines.py module) and 2) add tests for job submissions on CADES.

All Nexus tests again pass on oxygen.


markdewing:
okay to test

markdewing:
I would recommend converting all the tabs to spaces in nexus code and in the future require only spaces in check-ins (or vice-versa if tabs are preferred) (python -t will issue warnings about mixed tabs and spaces)

PR #1225, opened 2018-12-07
Backflow SoA Migration 2/2

rcclay:
This PR fixes the AoS/SoA distance table discrepancy in the backflow transformation.  This enables me to run VMC/DMC with/without drift within SoA or AoS, using backflow  wavefunctions. 

Some notes:
1.)  This is a "reference" implementation.  Only the interfaces to the distance tables were modified.  No attempt has been made to change the algorithm.  
2.)  Currently, the code is split between AoS and SoA code paths with #ifdef ENABLE_SOA.  While some cringe at macros, I figure the goal is to pull out AoS eventually, so this forked code path shouldn't be long lived.


prckent:
OK to test

ye-luo:
Does the BF optimization work with SoA or not?
Deos BF works with complex? heg sjb is not added with the complex build.
Could you make a summary what is left using BF with the spline SPO?

Please add an issue that we don't have a test for the eI case.
I check the heg tests. They all pass with SoA now.

rcclay:
@ye-luo 
1.) BF optimization seems to work.
2.)  BF works with complex.  All development was done on the complex build with spot checks on the real build.
3.)  For spline SPO, I think all the major functionality has been handled on the spline. For the BF transformations, I didn't implement all the various evaluate() calls present in the AoS version, but its not apparent to me how many of them are actually used.  If an exotic estimator gets broken, I'll fix it.  

Do you mean to add a test case for electron-ion backflow?  Some of the bccH tests certainly use an e-I backflow term...

ye-luo:
#1226 #1227 have been added to track needed tests.

prckent:
Do we have a backflow optimization test?

rcclay:
@prckent I would say yes, and I've verified that things do in fact optimize on the side.  Someone should point me to an optimization test template, or we should chat about what's desired out of such a test for backflow.  Checking to see if it doesn't crash is easy, but BF has always been fickle for optimization.  So saying we should gain X mHa for energy and reduce the variance by Y% for a test pass seems concerning.

prckent:
The art of making an optimizer test is finding a small enough but still meaningful system that can be run fast and reliably enough (3sigma). Perhaps few electron HEG?

PR #1230, opened 2018-12-07
Add unit test for Communicate class

markdewing:
In particular, test the splitting constructor used for ensemble runs.
Test that it splits the processes as expected and creates the group leader communicator.
Purpose is to ensure this functionality doesn't regress when moving to the new MPI wrapper.

Add MPI initialization to the Catch MPI header.   The OHMMS::Controller initialization can
be removed from some of the unit tests that have it.

The test is designed to run under multiple MPI processes, but test system doesn't
handle this yet.  This test can be run manually to check that it works for different
numbers of MPI processes.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -349,6 +349,7 @@ ENDIF()
     SUBDIRS(Estimators/tests)
     SUBDIRS(QMCDrivers/tests)
     SUBDIRS(QMCApp/tests)
+    SUBDIRS(Message/tests)
   endif() #}
 
 endif() #}}}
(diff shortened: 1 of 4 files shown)


prckent:
Is the problem with the test system only that our catch executable ctest macros don't run in parallel, or are more updates needed? RUN_QMC... is run in parallel.

markdewing:
The unit test driver in CMake/unit_test.cmake does launch with MPI, but is set to 1 process.
Changing this code should be all that is necessary.

Adding an MPI-specific reporter to Catch would make the output nicer, but is not strictly required.

PR #1231, opened 2018-12-08
Store r values in vector for 1D quintic splines

kgasperich:
store values to avoid repeated calls to exp

### DIFF
--- a/src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
+++ b/src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
@@ -33,6 +33,7 @@ namespace qmcplusplus
       T upper_bound;
       T OneOverLogDelta;
       double LogDelta;
+      aligned_vector<T> r_values;
 
       inline void set(T ri, T rf, int n)
       {
@@ -44,6 +45,11 @@ namespace qmcplusplus
         double dlog_ratio = log_ratio/static_cast<double>(n-1);
         LogDelta = dlog_ratio;
         OneOverLogDelta = 1.0/dlog_ratio;
+        r_values.resize(n);
+        for (int i=0; i<n; i++)
+        {
+          r_values[i] = ri*std::exp(i*dlog_ratio);
+        }
       }
 
       inline int locate(T r) const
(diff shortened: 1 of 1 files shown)


kgasperich:
This gave ~30% speedup in this part of the code for the few test cases I tried

markdewing:
okay to test

prckent:
@kgasperich Thanks. Seems like an excellent idea to do this. Have you thought through the memory consequences of adding this buffer? How many of these objects do we have and how large are the grids for different sizes or complexities of run?

Unrelated, I notice that the evaluate function strangely includes bounds checks and code for r values under the lower bound but the code for handling r values above the upper bound is commented out. Do you have any sense when/how the lower bound code is needed? The poor specification of bounds requirements in the 3D spline code has historically caused some touble.

kgasperich:
@prckent I believe that there is one of these objects for each basis set shell for each atomic species Someone who is more familiar with this part of the code might know whether this is true or if instead there is a copy for each individual atom (if there is a copy for each atom, this particular vector may or may not be copied depending on how the copy constructor (or makeClone) is implemented).
The default number of points is 1001 when using convert4qmc, so each vector will be ~8 KB (or 4 KB if this is one of the objects becomes single-precision when the code is built with mixed-precision turned on).
I'm not sure what should be taken as a typical use case, but an all-electron second row atom with a quadruple-zeta basis would have 15 shells, and this additional radial grid vector would take ~120 KB of memory if the default of 1001 points is used.

Regarding lower bounds, the default is 1E-6, and I don't have any sense of how often an electron ends up within this region or whether any checks are performed elsewhere (on either the upper or lower bounds).
@anbenali may have more insight on this issue.

prckent:
Thanks @kgasperich . One per basis set shell for each unique atomic species seems reasonable.

Regarding the lower bounds, hopefully @anbenali  will comment here. It would be helpful to get the specification written down.

ye-luo:
If an electron lands below the lower bound (this is still very likely), we must be extremely careful. If this is not handled properly, the catastrophic result can be a DMC blow up.

prckent:
Agreed. But this is not a fault of the spline. It is a fault of the calling code - in this source we have what appears to be a "simple" utility spline code that could be reused elsewhere, but in fact it has undocumented and unclear edge case handling . Presumably the calling code must already implicitly handle the r>r_max case. Not ideal.

Anyway, this PR is a good improvement so will be merged just as soon as the CI has passed.

PR #1234, opened 2018-12-09
Nested threading and improved delayed update documentation

ye-luo:
1. Mention delayed update in performance tests
2. Add description to nested threading.

### DIFF
--- a/manual/build_manual.sh
+++ b/manual/build_manual.sh
@@ -31,6 +31,7 @@ bibtex qmcpack_manual1
 bibtex qmcpack_manual2
 xelatex qmcpack_manual.tex
 xelatex qmcpack_manual.tex
+xelatex qmcpack_manual.tex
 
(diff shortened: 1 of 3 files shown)


ye-luo:
I will add a test for save_wfs probably under the restart tests.
This is not tested but I used it often and pretty sure it is correct. It is shared by CPU and GPU code path.
I'm adding this in the performance test because people profiling our code was seeing quite a lot of time in irrelevant portion. @PDoakORNL has this experience in the past. I have to keep explaining and sending this solution. So putting it in the manual is better.
The spline table construction is in parallel and cost nothing for real run using many node. Thes problem is single node running a very big problem. Performance test is usually this case. Users should not use this unless this is what they really want.
This feature is very useful but it should be moved to input instead of command line.

prckent:
OK. Add a variant of diamond 1x1x1 vmc test or similar where the GPU tests also run.

PR #1235, opened 2018-12-10
Use MPI3 wrapper

markdewing:
Includes some updates from the mpi3 wrapper repo that are needed to compile, especially with gcc 5.3 on Rhea.   Also need load the boost module on Rhea to get a newer boost (1.58)

This PR puts the new mpi3 wrapper in charge of initializing MPI, and the old OOMPI wrappers are initialized from that by passing an MPI communicator.

A check is on the Intel compiler version is added and the manual updated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -392,6 +392,13 @@ ELSE(CMAKE_TOOLCHAIN_FILE)
   # Check that a C++ 14 standard library is configured
   include(CMake/TestCxx14Library.cmake)
 
+  #------------------------------
+  #  check Intel compiler version
+  #------------------------------
+  if ((${COMPILER} MATCHES "Intel") AND (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 18.0))
+    message(FATAL_ERROR "Intel compiler must be version 18 or higher (currently using ${CMAKE_CXX_COMPILER_VERSION})")
+  endif()
+
   #-------------------------------------------------------------------
   #  check MPI installation
   #-------------------------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -392,13 +392,6 @@ ELSE(CMAKE_TOOLCHAIN_FILE)
   # Check that a C++ 14 standard library is configured
   include(CMake/TestCxx14Library.cmake)
 
-  #------------------------------
-  #  check Intel compiler version
-  #------------------------------
-  if ((${COMPILER} MATCHES "Intel") AND (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 18.0))
-    message(FATAL_ERROR "Intel compiler must be version 18 or higher (currently using ${CMAKE_CXX_COMPILER_VERSION})")
-  endif()
-
   #-------------------------------------------------------------------
   #  check MPI installation
   #-------------------------------------------------------------------
--- a/CMake/ClangCompilers.cmake
+++ b/CMake/ClangCompilers.cmake
@@ -1,7 +1,7 @@
 # Check compiler version
-IF ( CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.3 )
+IF ( CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.4 )
   MESSAGE(STATUS "Compiler Version ${CMAKE_CXX_COMPILER_VERSION}")
-  MESSAGE(FATAL_ERROR "Requires clang 3.3 or higher ")
+  MESSAGE(FATAL_ERROR "Requires clang 3.4 or higher ")
 ENDIF()
 
 # Set the std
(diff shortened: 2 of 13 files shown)


CMakeLists.txt
```diff
@@ -392,6 +392,13 @@ ELSE(CMAKE_TOOLCHAIN_FILE)
   # Check that a C++ 14 standard library is configured
   include(CMake/TestCxx14Library.cmake)
 
+  #------------------------------
+  #  check Intel compiler version
+  #------------------------------
+  if ((${COMPILER} MATCHES "Intel") AND (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 18.0))
```
ye-luo: Can we put this into Intel configuration file CMake/IntelCompilers.cmake?

ye-luo:
When building BGQ.
I got tons of the following warning
```
[ 77%] Building CXX object src/QMCWaveFunctions/CMakeFiles/qmcwfs.dir/BsplineFactory/createRealDouble.cpp.o
In file included from /home/yeluo/opt/cleanup/mdewing/src/QMCWaveFunctions/BandInfo.cpp:16:
In file included from /home/yeluo/opt/cleanup/mdewing/src/QMCWaveFunctions/BandInfo.h:20:
In file included from /home/yeluo/opt/cleanup/mdewing/src/Configuration.h:36:
In file included from /home/yeluo/opt/cleanup/mdewing/src/Message/Communicate.h:27:
In file included from /home/yeluo/opt/cleanup/mdewing/external_codes/mpi_wrapper/mpi3/environment.hpp:8:
In file included from /home/yeluo/opt/cleanup/mdewing/external_codes/mpi_wrapper/mpi3/../mpi3/communicator.hpp:39:
In file included from /soft/libraries/boost/1.62.0/cnk-bgclang++11/current/include/boost/archive/detail/common_iarchive.hpp:23:
In file included from /soft/libraries/boost/1.62.0/cnk-bgclang++11/current/include/boost/archive/detail/interface_iarchive.hpp:22:
/soft/libraries/boost/1.62.0/cnk-bgclang++11/current/include/boost/archive/detail/iserializer.hpp:69:7: warning: macro expansion producing 'defined' has undefined behavior [-Wexpansion-to-defined]
#if ! DONT_USE_HAS_NEW_OPERATOR
      ^
/soft/libraries/boost/1.62.0/cnk-bgclang++11/current/include/boost/archive/detail/iserializer.hpp:63:12: note: expanded from macro 'DONT_USE_HAS_NEW_OPERATOR'
        || defined(__SUNPRO_CC) && (__SUNPRO_CC < 0x590)   \
```
@markdewing I saw you metioned BGQ. Did you run any tests there?
Since this just hooks OOMPI with the mpi3-wrapper communicator, should I consider the change at the moment is low risk?

CMakeLists.txt
ye-luo: Is the chain rule of Intel using libstdc++ from gcc causing the so high version?
Do we also need to raise gcc and clang version minimum?

markdewing:
I haven't paid much attention to the warnings since they come from the boost library.

I have had trouble on Cetus with some of the unit tests that read HDF files.   The problem has been happening since before this PR, though.   The nightly unit tests have been passing, so there must be some difference between those tests and my local setup.

CMakeLists.txt
markdewing: No, there are specific bugs in Intel 17.

markdewing: You're right, we should update the minimum versions on clang and gcc for C++ 14 support.

ye-luo:
@markdewing I tried the 1.64 the same warning is still there. I'm not fully sure it is boost problem because if some user defined macros happen before the boost inclusion. Boost can struggle.
BGQ runs are fine. I just checked them again.

CMakeLists.txt
grahamlopez: Shouldn't CMAKE_CXX_STANDARD take care of that for us? The whole point is to not have to fiddle with individual compilers. Sometimes it can't be avoided (e.g. like avoiding version-specific bugs in Intel), but it seems like it shouldn't be necessary just to guarantee standard support.

prckent: Agreed.

The only reason to name specific compiler versions should be if we know they are problematically buggy.

markdewing:
I tested compilation on
 - Cetus
 - Theta
 - Cooley
 - Titan
 - Summit
 - Rhea (CPU and GPU)
 - Eos

And ran unit tests on Titan, Cooley, and Cetus. The tests passed, except the HDF file problems on Cetus that I mentioned above.

prckent:
Did you successfully test the ensemble runs?

markdewing:
I did not try any ensemble runs. 

The unit test issue on Cetus is resolved - I had the wrong directory name in the script I was using to run the tests.   Now all the unit tests pass there.

prckent:
Quick check of ensemble runs was consistent between current develop and this PR.

prckent:
No MPI build also confirmed OK.

PR #1237, opened 2018-12-10
Update dependencies in README.md

prckent:
Minimum versions enforced in #1235

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # Prerequisites
 
- * C/C++ compilers
+ * C/C++ 14 compilers. e.g. gcc 5.0, Intel 2018, clang 3.5 or newer. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,7 +4,7 @@
 
 # Prerequisites
 
- * C/C++ compilers
+ * C/C++ 14 compilers. e.g. gcc 5.0, Intel 2018, clang 3.5 or newer.
```
ye-luo: Please make it more accurate.
"C/C++ 14 compilers" is too vague.
For C compiler we need C99. For C++ we need C++14.
Please also mention that Intel depends on the C++ standard library from gcc and thus gcc version requirement applies to intel as well.

prckent: Updated. Will not go into many more details. This is the readme.

PR #1241, opened 2018-12-11
Reduce hdf_archive constructor.

ye-luo:
Use nullptr as the default communicator pointer. Thus the constructor without arguments is only a special case.

### DIFF
--- a/src/io/hdf_archive.cpp
+++ b/src/io/hdf_archive.cpp
@@ -40,14 +40,6 @@ void hdf_archive::close()
   file_id=is_closed;
 }
 
-void hdf_archive::set_access_plist()
-{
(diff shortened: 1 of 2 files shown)


markdewing:
What is a bit subtle is that the Comm class can be Commuicate* or mpi3::communicator (not a pointer).  The default of setting Comm c to a nullptr only occurs if the default value for Comm is Communicate *.
I think it works okay.

Refactoring hdf_archive now is somewhat moot because it will need to be changed again with Communicate is removed.

ye-luo:
@markdewing Let us leave it the current way and I will close this PR. I agree with you that we should do it when the situation is more clear.

PR #1242, opened 2018-12-11
Nexus: revise directory structure and enable install

jtkrogel:
Addresses the Nexus part of #1020.

An install script is added to Nexus that sets user paths and also copies binaries to a destination, if desired.  When integrated with QMCPACK, CMAKE_PREFIX_PATH is used for this purpose.  The directory structure of Nexus is also updated to more standard usage.

Nexus versioning is set here to 1.6.0.  Future releases will formalize the use of the release number throughout the Nexus package.


markdewing:
okay to test

jtkrogel:
The current setup results in a Nexus install during the initial cmake configuration step.  I should probably try to defer the Nexus install until "make install" is invoked.  I think this can be done by using

install(CODE "execute_process(COMMAND my_script.sh)")

Will try this in a bit.  Hold on merge.

jtkrogel:
Any thoughts on what the flag should be?  "--set_paths"?

prckent:
Re: shell files. I would prefer these are not touched at all, but particularly for this first release. "make install" does not typically modify .rc files. Printing a message on install may be useful however.

jtkrogel:
Will default to printing a message w/o modifying rc files.  A flag can be set to suppress the message and automatically add flags to the rc file.

markdewing:
Installing python library files is a bit more involved than the bin/ files.
Usually the python files go in someplace like
`/usr/local/lib/python2.7/site-packages/` or `~/.local/lib/python2.7/site-packages/`  and are installed using python-specific installation mechanisms.

Installing nexus to a non-standard location and setting the PYTHONPATH is the easiest way to use it, and the installer will work find for that.   My concern is that if people start installing to a standard location (/usr/local) there will be problems.

jtkrogel:
This is a possibility.  Most Python installs assume a setup.py and usage of pip to integrate with the system Python install.  This is not really an option in HPC environs, and it also does not play nice with multiple versions of the software installed simultaneously.  This will clearly require more thought down the road.

Right now, I'm satisfied with moving the binaries appropriately and giving instructions to complete the install (since we aren't going to modify rc files by default with cmake).

jtkrogel:
I've changed the Nexus install to occur only when "make install" is invoked.

I've also added a flag to suppress modifying rc files and this is used in cmake, so "cmake install" will never attempt to modify the rc files.  Instead a message like the following is printed during install:
```
Install the project...
-- Install configuration: "RELEASE"
-- Installing: /home/j1k/tmp/bin/qmcpack.settings
Nexus installation not complete.  Please add the following paths to /home/j1k/.bashrc:
export PATH=/home/j1k/tmp/bin:$PATH
export PYTHONPATH=/home/j1k/repositories/nexus_release/nexus/lib:$PYTHONPATH

-- Installing: /home/j1k/tmp/bin/qmcpack
-- Set runtime path of "/home/j1k/tmp/bin/qmcpack" to ""
-- Installing: /home/j1k/tmp/bin/qmc-get-supercell
-- Installing: /home/j1k/tmp/bin/qmc-check-affinity
-- Installing: /home/j1k/tmp/bin/convert4qmc
-- Set runtime path of "/home/j1k/tmp/bin/convert4qmc" to ""
-- Installing: /home/j1k/tmp/bin/qmc-extract-eshdf-kvectors
-- Set runtime path of "/home/j1k/tmp/bin/qmc-extract-eshdf-kvectors" to ""
```

jtkrogel:
Done unless other changes are desired.

markdewing:
Do we want a variable to turn nexus installation on/off?  (INSTALL_NEXUS)?   That would support installing qmcpack, but not nexus.   The default could be on, but it would be nice to have a way to turn it off, especially if it's being installed where the python libraries might cause a problem.

jtkrogel:
Anything else needed here?

Issue #113, opened 2017-02-09
Code may proceed or crash when Jastrow species do not match particleset groups.  Abort instead.


jtkrogel:
Using e.g. particleset groups Ni1 and Ni2 but Jastrow labels eNi (J1 with elementType="Ni") and/or uuNi (J3 with ispecies="Ni") are incompatible.

Currently the code simply ignores the discrepancy, with the result that the J1 terms are not read in while the J3 ones still are.  The runtime behavior thereafter depends on the build environment: on Eos at OLCF the code proceeds to run to completion, while on Titan builds somewhere after SVN revision 7044 crash with a "double free or corruption" error in eeI_JastrowBuilder::putkids().

Behaviour should be changed to perform a controlled abort if Jastrow labels do not correspond to any particleset species group.

ye-luo:
Please attach your output file.

ye-luo:
Probably we need to add some guards at the builder level.

prckent:
Agree, proper error handling with an informative message needed here.  Use of out of date converters can generate this problem.

#36 is not so different.

jtkrogel:
Resolved by #1244.  Closing.

PR #1244, opened 2018-12-12
Abort if Jastrow species not in ParticleSet

jtkrogel:
Addresses #113.

Guards have been added to Jastrow builders to prevent the code from running if requested Jastrow species are not present in the corresponding ParticleSet.  Previously, the code might silently ignore the input and simply drop the offending Jastrow terms.

The new code was tested explicitly for all species terms in 1 (bspline), 2 (bspline), and 3 (polynomial) Jastrows in the NiO test case.  All abort/error messages have the same form.


markdewing:
okay to test

jtkrogel:
The default values for especies1 and especies2 are both "u", which is correct in this case.

To move away from reliance on default behavior (and non-reading behavior), 'especies="u"' should be replaced with 'especies1="u" especies2="u"'.

prckent:
OK. Please add a bug/issue. This is a place where there I think there should be no default and we should abort if the jastrow is not fully specified.

ye-luo:
I tweaked the code and made the error printing together if both electron species are wrong in J2 and JeeI. Users should see more error with fewer runs.
Regarding JeeI, I agree with @jtkrogel we should enforce specify both eSpecies1 and eSpecies2 even they are the same.
In general we need to enhance our xml parsing classes.

ye-luo:
gcc 6.2 and cuda 10

PR #1245, opened 2018-12-12
Merge AFQMC

markdewing:
This copies most of the code from #1194 

The test directories were not copied, nor the minor update to the manual.  Content in OLD and PREVIOUS_VERSION directories were also not copied.

The code was modified from #1194 to work with the  updated MPI3 wrapper.

There are two changes to the MPI3 wrapper necessary to build it, included in this PR.

The only changes not in the AFQMC directory should be in QMCMain.cpp and the previously mentioned change in the wrapper.


fdmalone:
Ok, but what is the benefit over just merging #1194 and adding Mark's additions on top?

markdewing:
Copying files has a few advantages:
- It clearly removes files from the repository.  There's a 60MB file in the test/ directory that would go into the repository even if removed from the branch (the file size is not huge compared to the repo, but not necessarily small, either)
- It made it easier to separate the PR into smaller pieces that affected portions of the real-space code.    This could also be done with git's cherry-pick and rebase commands, but with a large number of commits it gets unwieldy.

Another problem I'm still thinking through - this is going to create some merge conflicts when Miguel pulls the upstream develop onto his develop.

fdmalone:
Yeah that's the bit I'm worried about. I think I will have to rebase heavily.

markdewing:
I'm looking at ways to remove the large file from the branch while preserving more history.   I think any technique is going to change the commit hashes and cause merging problems.

fdmalone:
Yes, it might just be easier at this point for me to just cherry-pick all of the upstream work onto this once it is merged and start fresh. I think the alternative is too much work at this point and would require going through the pain of rebasing llnl_afqmc onto qmcpack again.

markdewing:
A fresh clone of the qmcpack repo is 458MB.  I would like to avoid adding a file that is > 10% of the repo size.

fdmalone:
Absolutely. In retrospect I could have rebased onto develop and removed this file, those old directories and isolated the real space work so they could be split into separate commits. Then we could have preserved the commits at least.

fdmalone:
But I think the best thing to do now is to just go ahead with this merge. I will deal with the pain on our end.

ye-luo:
@fdmalone Probably this is the best way. Once this gets merged. Everyone working on the AFQMC needs to cherry-pick unmerged changes to the new develop and drop the old one on your side. As @prckent said, once AFQMC development stays close to the main QMCPACK process. No problem should occur again.

prckent:
This uses boost/hana.hpp which is only avail in boost 1.61.0+. We should check the boost version on configure and update our minimum version.

ye-luo:
gcc 6.2 and cuda 10
All the complex builds have afqmc enabled.

Issue #794, opened 2018-04-26
The (automatic) code format


grahamlopez:
It would help to move to automated source code formatting. Some candidate examples are available at https://github.com/qmcpack/src-formatting; more explanation of these examples are in the readme for that repo.

Note that the formatter can be turned off inside the source by enclosing regions with appropriate comments (`// clang-format off/on`). I didn't search for everywhere this will be needed; I think we can pick them up as people run into them. We probably want try to minimize how often we use that feature (e.g. only for arrays/matrices of constants), as it adds to maintenance overhead and erodes the advantage of automation.

Let's use this issue thread to discuss which of these formats we would like to use, and any ways to tweak them.

prckent:
My vote is for custom. I find the placement of the opening and closing {} helpful to scan the logical flow of the code.

jtkrogel:
My main issue with all the current options is that they make heavy math illegible.

Compare for the following files:
./src/einspline/multi_bspline_eval_d_std3.cpp
./src/einspline/nubspline_eval_d_std.cpp
./src/QMCWaveFunctions/Fermion/BFsplineFunctor.h
./src/QMCWaveFunctions/Jastrow/eeI_Jastrow

Spaces before and after */,< is excessive and the line wrap cutoff could be extended.  Note that even the original code is not ideal on math presentation as it was passed through another auto-formatter on 25 Apr 2013.

Perhaps some of this can be addressed with additional customization?

jtkrogel:
Good to know that this can be selectively applied.  Perhaps adjusting the rules also will help (favoring more compact math).  There will be many cases that I don't have time to find that will/could be handled poorly.  I suggest instead that we shape the rules to render the above files (and others we find) well so that we can apply the formatter to the whole of the code without concern.

Overall, I would strongly prefer to not introduce spaces before/after */,<.  I might also add +- to this list as sometimes spaces help, but often they hinder readability by expanding e.g. for loop declarations a lot.  These cases could be handled manually (less space in for loop decls, more space in full math expressions) for readability if an auto-tool cannot be customized enough.

jtkrogel:
Last time the code formatting was performed by using astyle: [http://astyle.sourceforge.net/](http://astyle.sourceforge.net/)

jtkrogel:
Presumably this would render the files listed above sufficiently well.

jtkrogel:
> I suggest that we just fix these as we encounter them after the fact

[...]

PR #1248, opened 2018-12-13
Counting Jastrow Factors

bwvdg:
This is an implementation of Counting Jastrow factors, which are a type of four-body jastrow factor using a set of gaussian quotients as Jastrow basis functions. These basis functions depend on explicit electronic coordinates and are handled separately from the existing DistanceTable-based Jastrow framework. More information can be found at https://arxiv.org/pdf/1811.03191.pdf .

The Counting Jastrow class is separated into three layers: the outer layer (CountingJastrow.h) explicitly handles the log-linear Jastrow parameters, the middle layer is a wrapper for the gaussian quotient basis functions (CountingRegion.h), and the final layer contains the primitive basis function components (3d-gaussians) themselves (CountingFunctor.h). A builder class (CountingJastrowBuilder), a working test (short-H4-cjs-16-1), and a basic cjf generation utility (utils/ncjf_gen.py) are also included.


prckent:
Many thanks for this!

You labeled the PR as a work in progress [WIP]. Do you plan any other updates?

prckent:
One question:

In the test you have:
LIST(APPEND H4_CJS_SCALARS "totenergy" "-2.0772 0.0095") # total energy

Where did this reference energy and error bar come from? In our experience this needs to come from a run with (say) 100x more blocks in order to have reliable statistics. (We currently believe that most of the statistical failures on https://cdash.qmcpack.org/CDash/viewProjects.php are due to insufficiently accurate reference data due to overly short reference runs. They will be updated soon.)

prckent:
Please could you remove the large sections of commented code? You can always add them back uncommented in a future update should they need to be activated.

prckent:
We should aim to get this merged in before the new release, which we are working on now. The release will be followed by #794 which will require rebasing these changes.

bwvdg:
I labelled this WIP because I was concerned about some code style (ex. using std::function as a getter/setter for data storage vectors). I don't plan on any substantial changes (such as those that might affect a test) any time soon.

I removed DiffCountingJastrowOrbital.h, which is unimplemented and should be the only source of the large code comments.

I took the energy and sigma directly from the test itself and should currently match the statistics calculated by check_scalar.py. I have numerical results for the same test under much better statistical conditions, but the test doesn't pass under the short conditions. The test currently consists of three linear method steps followed by a VMC run. I would like the test to include optimizer/evaluateDerivatives code, but I can't increase the sample size very much without making it take longer than about a minute. How should I modify the test?

prckent:
A way to generate reliable reference data:
- Run a long vmc with x100 more blocks than the short test that can run in about a minute
- Compute average properties and error bars
- Put the computed average properties from above with appropriately rescaled error bars (x10 here) in the test definition

Averaged properties from shorter runs tend to be just bad enough that we see one or two sigma failure rates, and not the desired 3 sigma.

src/OhmmsData/libxmldefs.h
```diff
@@ -70,7 +70,8 @@ bool putContent(T& a, xmlNodePtr cur)
   std::istringstream
   stream((const char*)(xmlNodeListGetString(cur->doc, cur->xmlChildrenNode, 1)));
   stream >> a;
-  return stream.good();
+  return !stream.fail();
```
ye-luo: Could you revert these changes or there is a reason? It seems irrelevant to this PR.

prckent:
Revision of above statement: if you are testing optimization (vs a simple VMC run), we are open to ideas. We would like separate tests for optimization and for VMC. The latter we know how to make reliable enough.

src/QMCWaveFunctions/Jastrow/CountingFunctor.h
```diff
@@ -0,0 +1,433 @@
+
+#ifndef QMCPLUSPLUS_COUNTING_FUNCTOR_H
```
ye-luo: Missing licence header. Please add and update the header at the beginning of every file you introduced.

src/QMCWaveFunctions/Jastrow/CountingFunctor.h
```diff
@@ -0,0 +1,433 @@
...
+    {
+      os << std::endl;
+      for(int J = 0, IJ = I*3; J < 3; ++J, ++IJ)
+        os << boost::format("  %10.5f") % A[IJ];
+    }
+    os << std::endl << "  opt_A: ";
+    std::copy(opt_A.begin(), opt_A.end(), std::ostream_iterator<bool>(os,", "));
+    os << std::endl << "  B: "; 
+    for(auto it = B.begin(); it != B.end(); ++it)
+      os << boost::format("  %10.5f") % (*it);
```
ye-luo: Can we avoid boost here?

src/QMCWaveFunctions/Jastrow/CountingJastrowBuilder.cpp
```diff
@@ -0,0 +1,190 @@
...
+  {
+  ClassName="CountingJastrowBuilder";
+  NameOpt="0";
+  TypeOpt="unknown";
+  RegionOpt="normalized_gaussian";
+  SourceOpt=targetPtcl.getName();
+}
+
+template<class precision, template<class> class CountingRegionType>
+class CountingJastrowTypeHelper
```
ye-luo: Can we remove this class and directly put its content at its use? It seems it is only used once.

src/QMCWaveFunctions/Jastrow/CountingJastrowOrbital.h
```diff
@@ -0,0 +1,741 @@
+#ifndef QMC_PLUS_PLUS_COUNTING_JASTROW_ORBITAL_H
+#define QMC_PLUS_PLUS_COUNTING_JASTROW_ORBITAL_H
+
+#include "Configuration.h"
+#include "Particle/ParticleSet.h"
+#include "QMCWaveFunctions/WaveFunctionComponent.h"
+
+#include "boost/format.hpp"
```
ye-luo: No boost please.

src/OhmmsData/libxmldefs.h
bwvdg: The put function logic didn't consistently work without this, since stream.good() returns false if the stream reached eof.

ye-luo: Changing this may affect all other places using this function.
If xml input doesn't provide sufficient values to feed the request, it is a failure.
Please give your logic with an example.

bwvdg:
I'll add the VMC test using the higher-resolution results.

I know in previous versions of QMCPACK, I've been able to get matching output down to machine precision even for multiple optimizer steps. For optimization, would it be unreasonable to test for consistency using a particular random seed (or even a small set of them) rather than for statistical accuracy? I do know this complicates future changes to walker loading/sampling or random number generation, but it seems more robust to optimizer divergences or small errors in a short statistical test, with the added benefit of not requiring a long sampling procedure.

grahamlopez:
okay to test

src/OhmmsData/libxmldefs.h
ye-luo: !stream.fail() is correct. Please remove all the commented stream.good().

src/QMCWaveFunctions/Jastrow/CountingJastrowOrbital.h
```diff
@@ -0,0 +1,742 @@
...
+#include "Particle/ParticleSet.h"
+#include "QMCWaveFunctions/WaveFunctionComponent.h"
+#include "QMCWaveFunctions/Jastrow/CountingRegion.h"
+
+namespace qmcplusplus
+{
+
+template <class RegionType>
+class CountingJastrowOrbital: public WaveFunctionComponent 
+{
```
ye-luo: Could you change the class and file names to CountingJastrow. Calling it Orbital only adds confusion.

src/OhmmsData/libxmldefs.h
ye-luo: I did it.

src/QMCWaveFunctions/Jastrow/CountingJastrowOrbital.h
ye-luo: I changed it.

ye-luo:
@bwvdg I fixed the complex compilation issue and CI builds all pass now. It is not intended to use this Jastrow with the complex code but to get the compilation through. If someone wants to run the complex code, careful check is still needed.

Please complete the following:
1) upload your run which generated the reference numbers. Add a folder qmc_cjs_ref under H4_ae and put **only** the xml input, standard output and *.scalar.dat 
2) about std::function. It seems you are building a matrix with vector storage. In QMCPACK, you can use the matrix class defined in OhmmsPETE/OhmmsMatrix.h. The matrix class accepts RealType/ValueType/PosType/GradType as its template argument.

bwvdg:
Sorry for the delay. I moved the test to its own folder and replaced the indexing functions with Matrix classes.

ye-luo:
@bwvdg I was not asking to move the tests but to add the input and output of the run you generates the reference data. The test and reference run should be different. See https://github.com/QMCPACK/qmcpack/wiki/How-to-add-tests.
You can take tests/molecules/LiH_dimer_pp as example.

bwvdg:
Alright, I moved the test back. I have the input, scalarfile, and the output in the reference folder and the test in the original place.

ye-luo:
@bwvdg The reference run should be much longer and wider than the test. You can't use the same one.

bwvdg:
Wasn't thinking. I added the correct reference.

ye-luo:
Please follow all the steps in
Write a README, update the test reference number based on the long reference run.
Once you think everything is ready to merge, remove the [WIP]

bwvdg:
I added the readme and updated the reference number.

ye-luo:
I revised the error bar based on the reference run uploaded by @bwvdg. The short test needs only 8 seconds on my workstation. There is no long test at the moment and we can add it after we observing the short one for a while.

prckent:
Agree.

prckent:
Many thanks for completing this @bwvdg

PR #1250, opened 2018-12-13
Nexus: add kpoints for solid-state PySCF runs

jtkrogel:
Write out k-points for use by PySCF.  This is a step toward enabling complex gaussian PBC simulations in QMCPACK.


ye-luo:
gcc 6.2 and cuda 10

PR #1251, opened 2018-12-13
Nexus: fix GAMESS environment setup on ALCF machines

jtkrogel:
Environment variables are set by Nexus in job submission scripts.  On ALCF machines like Vesta, these need to be provided directly to runjob.  As a corner case, environment variables for GAMESS were not being set in this required way.

This PR fixes the issue (basically due to initialization ordering in the Gamess class) and adds tests that include rendering of job submission commands that include user requested environment variables.


ye-luo:
gcc 6.2 and cuda 10

PR #1255, opened 2018-12-14
Nexus: add job_ref_table option to nxs-test

jtkrogel:
Left the door open on one of the last commits.  Instead of commenting/uncommenting bits of code, add a flag to nxs-test when job/machine reference data needs to be updated.


markdewing:
Built and tested on Rhea with 1 configuration (CPU, MPI).

PR #1263, opened 2018-12-15
Add batched dgemm routines.

fdmalone:
Some additional blas routines needed in upstream AFQMC. @mmorale3


mmorale3:
This is temporary and will be removed in a future commit (within the next
few weeks).
The new linear algebra on the AFQMC side will not rely on Blasf.h or on
OhmmsBlas anymore.
If this is a problem for the release, then I can remove it right now.

Miguel

ye-luo:
Not a problem for release. Open an issue for future reference.

PR #1264, opened 2018-12-15
HDF archive for multi array.

fdmalone:
@mmorale3

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -120,6 +120,10 @@ if (HAVE_MPI)
   INCLUDE_DIRECTORIES("${qmcpack_SOURCE_DIR}/external_codes/mpi_wrapper")
 endif()
 
+if (HAVE_LIBBOOST)
+    INCLUDE_DIRECTORIES("${qmcpack_SOURCE_DIR}/external_codes/boost_multi")
+endif()
+
 ####################################
 # create liboompi
 ####################################
--- a/src/io/hdf_archive.h
+++ b/src/io/hdf_archive.h
@@ -22,6 +22,7 @@
 #include <io/hdf_stl.h>
 #include <io/hdf_hyperslab.h>
 #include <io/hdf_double_hyperslab.h>
+#include <io/hdf_multi.h>
 #if defined(HAVE_LIBBOOST)
 #include <io/hdf_ma.h>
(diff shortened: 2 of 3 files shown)


src/io/hdf_archive.h
```diff
@@ -22,6 +22,7 @@
 #include <io/hdf_stl.h>
 #include <io/hdf_hyperslab.h>
 #include <io/hdf_double_hyperslab.h>
+#include <io/hdf_multi.h>
```
ye-luo: Please move this inside HAVE_LIBBOOST

ye-luo: Please also clean up hdf_ma.h. I guess it is legacy now.

src/CMakeLists.txt
```diff
@@ -120,6 +120,10 @@ if (HAVE_MPI)
   INCLUDE_DIRECTORIES("${qmcpack_SOURCE_DIR}/external_codes/mpi_wrapper")
 endif()
 
+if (HAVE_LIBBOOST)
```
ye-luo: Change my mind a bit. MA is not part of boost yet. It must have its own macro if needed.
At the moment, MA is shipped within the code and used by default. We don't need to use a macro.
@markdewing what do you think?

src/io/hdf_archive.h
ye-luo: Changed my mind on put it inside HAVE_LIBBOOST macro. See above.

mmorale3:
MA is actually boost's multi_array. multi is Alfredo' library, so
technically hdf_ma could be guarded.
In any case, boost is a requirement, no?

Miguel

ye-luo:
I was so confused by boost multi_array and Alfredo's boost-multi.
Is Alredo's multi now included as external_codes/boost_multi?
Since it is not part of boost, guard it with HAVE_LIBBOOST in src/CMakeLists.txt is inappropriate.

Are both multi_array libraries simultaneously used in the AFQMC code?

Boost is a requirement at the moment.

mmorale3:
Alfredo’s multi is now included with QMCPACK, but I haven’t finished replacing Boost’s multi_array with it yet. This is coming very soon, probably within the next 2-3 weeks, but it is not done yet. So for a bit longer, both libraries are used simultaneously.

PR #1271, opened 2018-12-17
Nexus: update qmca quantities and tighten logic

jtkrogel:
Update qmca quantity abbreviations to include RMC, CSVMC, and AFQMC quantities.

Also tighten the logic surrounding total samples, total time and efficiency.  The logic changes consider the minimum amount of information needed to compute each quantity separately.  This way the maximum set of quantities is made available to the user.


jtkrogel:
Sure.  Imagine you don't have what is needed to compute "Efficiency" (e.g. LocalEnergy is missing because you are running CSVMC).  In the original code, "TotalSamples" would not be computed, but in the modified code, it will.

ye-luo:
@jtkrogel Thank you.

Issue #1253, opened 2018-12-14
Regen/check test reference data


ye-luo:
Work left by PR #1085

    Perform high precision tests on all testing platforms, including parallel submission to batch queues
    Perform checks against reference data for newly run high precision tests and report failures

PR #1274, opened 2018-12-18
Nexus: add basic analysis capability to qtest

jtkrogel:
The qtest test script can now postprocess reference data regenerated with Nexus, including performing joins of separate reference runs.

This is an incremental update.  More work is needed for completion of qtest (see #1253)


markdewing:
okay to test

ye-luo:
Okay to test

tests/scripts/qtest
```diff
@@ -1058,6 +1258,307 @@ def run_adjusted_tests(short_tests):
 
 
 
+#=================================#
+# analyze_tests and sub-functions #
+#=================================#
+
+# quantity mapping for check_scalars.py to ctest/cmake (see macros.cmake)
```
markdewing: These mappings are maintained in four places - CMake/macros.cmake, check_scalars.py, qmca, and qtest.
Would it be worth creating a single location for these?  (Future change, not for this PR)

One complication is check_scalars.py and qmca are standalone files and import only standard python modules.   Adding an import of a local module doesn't seem like a good idea.
Perhaps a code generation scheme?  One file contains these mappings and generates the form used in macros.cmake, check_scalars.py, and qmca.   qtest imports from nexus, so it could import the necessary module.

Code generation has a comprehensibility cost as well.    The question is if maintenance benefit for a single definition of these quantity mappings is worth the cost?

jtkrogel: The mappings themselves differ in each location, so I'm not sure that centralizing them makes sense.  Possibly the name mapping within the testing system itself could be made consistent and be centrally located in check_scalars.py itself (macros.cmake (and qtest) could then get its mapping list by calling check_scalars.py by adding the appropriate command line flag).  This would keep things centralized, readable, and avoid code generation or local modules.

Should we make an issue for this, discuss more, and then act?  Probably does not need to be resolved for this PR to proceed.

jtkrogel:
Any decision/request for this PR?

Issue #1266, opened 2018-12-15
Remove cblas calls from Blasf.h


mmorale3:
Routines were temporarily included in a recent commit for batched blas calls in AFQMC.
Will be removed in future release with the introduction of new linear algebra backend in AFQMC code.

prckent:
The batching interface could be useful in real space as well.

ye-luo:
When the real space code needs them, we can reintroduce them in a separated header.

PR #1278, opened 2018-12-20
Realign LLNL AFQMC with QMCPACK develop

fdmalone:
The following brings the main AFQMC development branch back in line with qmcpack develop following #1245. I have also updated the AFQMC unit tests to use the benchmark results in the input files which they read. I'm not sure whether AFQMC is actually set up for CI but at least now the unit tests will be testing something more concrete.

I'll shortly start adding short and long AFQMC tests, as well as consolidate the integral generation scripts which we will distribute with QMCPACK.

There are still batched dgemm routines added which will eventually be removed following #1266.

Additionally there are some gentle temporary hacks to mpi3 and boost_multi which will dealt with more cleanly in the future @mmorale3. These modifications shouldn't affect anything outside of AFQMC.


markdewing:
okay to test

ye-luo:
I noticed some changes are in boost_multi and mpi_wrapper. What is the best strategy and our policy?
Should we only allow changes directly in the upstream library or we maintain a diverged local copy?

mmorale3:
As Fionn mentioned, this is a temporary modification which we will remove
soon.
It does not affect anything outside AFQMC.

In the future, we will do PRs that modify code outside AFQMC independently,
but this was so small we decided to do it all together.

ye-luo:
@mmorale3 Thank you for the promptly reply. I missed the last paragraph. My fault. I ran CI on my workstation. The complex builds with AFQMC all pass.

PR #1284, opened 2018-12-22
Enable AFQMC by default with the MPI complex full precision build.

### LINKED ISSUE #1252: AFQMC not enabled in CI
Currently CI scripts on Rhea do not enable BUILD_AFQMC? Does anyone remember the history?
Consider AFQMC development will be more close to the main repo, we need to turn that on.

- [x] boost 1.61 needed in CI scripts. Rhea default is 1.58
- [ ] ~~update CI script to enable BUILD_AFQMC on complex builds.~~

ye-luo:
The default value can always be overridden via command line options.
closes #1252


ye-luo:
Some painful chain reaction. AFQMC->boost::hana->gcc 6->cuda 9. Only CUDA 8 is available on rhea.

ye-luo:
@mmorale3 @fdmalone is there AFQMC GPU code in the develop? Is QMC_CUDA reused to access AFQMC GPU code?

mmorale3:
This is coming in the next 2 week or so. I’m moving code from miniAFQMC to the main code. In the process I’ll fix mixed precision and real builds. I’ll also stop using boost::hana, it was just for convenience.

prckent:
@ye-luo Please file a ticket with OLCF requesting up to date software versions on rhea.

fdmalone:
Why is the CPU build failing? I don't have access to the details.

PR #1285, opened 2018-12-23
Remove legacy tau profiler interface

### LINKED ISSUE #1281: Is qmcpack/src/tau legacy and removable?
What is the status of this code? Unless someone champions it we should remove it.

prckent:
Closes #1281


ye-luo:
Could you please remove all the 'grep -R TAU_PROFILE' lines?
Everything inside ppconvert can be ignored.

PR #1289, opened 2018-12-29
Multideterminants from HDF5 File. 

anbenali:
HDF5 Reader for  CI.
- SlaterDetBuilder gets a new function to read from HDF5. Multideterminants are stored in HDF5 using 64 bits representation of the usual excitation string. 
For example a determinant: "11100000" will be stored as the long integer: 7. If the number of orbitals (CI states) is larger than 64, the the number of integers to store the representation is increased. 
Using this representation,   a Trial wavefunction of 2M determinants goes from 457M to 77M. 
Significant speedup in the initialization of the wavefunction is also  observed. Benchmark will be provided. 

- Cleaning inconsistencies in the size of basis set. We always assumed that N_MO=N_AO. Not always the case. loops are now corrected (will verify if fixes known bugs). 

- Converter is now able to run with flag "-hdf5" with multideterminants. However, only to be used for additional tests. QP dumps complete HDF5 file compatible with QMCPACK. 

 - Removed converter for VSVB. VSVB code has significantly changed since version beta and will not provide same output. Better remove. 

- Converter section of the manual updated

- C2_pp test updated to reflect HDF5, XML and multidets in general.


ye-luo:
Okay to test

prckent:
Q. Does the new det coeff cutoff successfully drop zero coefficient determinants? If so, please clarify the language in the manual to state this.

tests/molecules/C2_pp/CMakeLists.txt
```diff
@@ -1,54 +1,70 @@
-
+IF (ENABLE_SOA)
```
ye-luo: I guess the AoS should also work. MSD operates at the level above SPOSet.

tests/molecules/C2_pp/CMakeLists.txt
```diff
@@ -1,54 +1,70 @@
...
                     1 16
-                    ${MP_SUCCESS}
-                    1 C2MSDJS_VMC_SCALARS # VMC
+                    TRUE
+                    1 C2_PP_MSDJ_SHORT # VMC
                     )
-
-  QMC_RUN_AND_CHECK(short-C2_pp-msdj-traditional_vmc
+ 
+  QMC_RUN_AND_CHECK(vmc_short_pp_msdj-XML
```
ye-luo: Please add back a test specific for the traditional algorithm code path. The default option is the table method.
You can find the input difference by comparing qmc_short_pp_msdj.xml and qmc_short_pp_msdj-traditional.xml in the develop branch.

ye-luo:
C2.ezfio was added as the original data from QP.
Similar to QE, we only keep the input and output file but we don't keep outdir.
The same logic here, we would like to keep the original input for QP instead of intermediate files with QMCPACK tests. The internal format of C2.ezfio may change from version to version and the file can be out-of-date very quickly. QP should be reliable enough to reproduce its previous result (within a reasonable accuracy) from very simple input. If keep C2.ezfio is the only option (I hope not), it must be a tar instead of a folder with many small files.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -1218,4 +1244,194 @@ bool SlaterDetBuilder::readDetList(xmlNodePtr cur, std::vector<ci_configuration>
...
+  spoAttrib.add (Dettype, "type");
+  spoAttrib.add (nstates, "nstates");
+  spoAttrib.add (cutoff,"cutoff");
+  spoAttrib.add (multidetH5path,"href");
+  spoAttrib.put(DetListNode);
+  if(ndets==0)
+  {
+    APP_ABORT("size==0 in detlist is not allowed. Use slaterdeterminant in this case.\n");
+  }
+  if(Dettype != "DETS" && Dettype != "Determinants")
```
ye-luo: Just spot this by change. More for discussion, not asking for a change in this PR.
Is it better to support only Determinants in the future? I don't see any benefit keeping both except supporting legacy.
@anbenali Does the converter only write DETS or Determinants?

tests/molecules/C2_pp/CMakeLists.txt
anbenali: It still works for the XML version. The point is to have the HDF5 version by default.  I will fix.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
anbenali: this is for the case of CSF.  Since I do not use CSFs and the only way (for now) to generate a CSF is through Gamess and I will not touch gamess unless Paul Kent forces me to do so I only focused on what is needed to do science.
This being said, CSFs are important and should still be supported. Also, it is unlikely to endup with millions of determinants in CSF so the XML path is fine.

ye-luo: Yes. I mean to keep minimal amount necessary input tags. We probably should only keep CSF and Determinants and deprecate all others by stopping the code.

PR #1291, opened 2019-01-03
Switch to multi array in AFQMC

fdmalone:
Completes switch to Alfredo's multi array library in AFQMC. Removes boost::hanna which should fix build errors.


mmorale3:
There's a small bug fix to boost-multi/multi/array_ref.hpp,
will be updated with official library after Alfredo makes the fix on his end.

prckent:
OK to test

ye-luo:
Do we still use boost and Afredo's multi array? Or no more boost one?

mmorale3:
Alfredo's multi should replace boost's multi_array entirely. There should
be no more usage of boost's multi_array.
If anything is left it is because I missed it.
For example, writing this I realized that I didn't removed the template
specializations for boost's multi_array in the hdf and type_traits folders.

I still need boost, because I use boost::variant and boost::iterator.
When we move to c++17, I can stop relying on boost entirely.

ye-luo:
@mmorale3 Thank you. Please remove the left non used code when you have time. Your explanation made the status of multi_array very clear. It is great that hana is gone. We won't remove the dependency of boost until all your needs are covered by C++.

PR #1300, opened 2019-01-04
Changes to walker set

mmorale3:
Small changes to the shared walker set. I stopped using unique_ptr, which was never a good idea to begin with. I also fixed a bug in boost::mpi3::intranode::array_ptr in the implicit conversion to const pointers.


prckent:
OK to test

mmorale3:
test this please

grahamlopez:
test this please

PR #1302, opened 2019-01-07
Add complex build afqmc tests

fdmalone:
Currently only complex builds of AFQMC are supported, but the tests in tests/solids are set up for real builds.

I've adapted the unit tests to long and short tests for complex builds of AFQMC for use in CI.
The tests currently cover Cholesky factorized integrals with RHF, UHF and particle-hole multi determinant trial wavefunctions, and ISDF/THC factorized integrals with an RHF trial wavefunction.


markdewing:
okay to test

PR #1303, opened 2019-01-08
Cuda backend and kernels for AFQMC

fdmalone:
Implements a number of cuda kernals for upcoming AFQMC gpu implementation.


prckent:
OK to test

mmorale3:
How do I send the command to retest?

prckent:
The (re)testing is done on a polling basis, so if the CI has been approved to run once it should rerun each time the PR is updated.

PR #1305, opened 2019-01-09
Add missing NOMSD wavefunction test data.

fdmalone:
I forgot to add the NOMSD trial wavefunction data necessary for some of the AFQMC unit tests to run.


markdewing:
okay to test

prckent:
Fionn  - will the AFQMC code need many more files like these as part of routine testing? We try to be careful not to include "large" files in the git or distribution. A single 1M file is fine in my opinion, but we don't want to have too many.

fdmalone:
No, this should be the last of it. I could actually remove the larger file in this case and store the test information elsewhere. I'll do a force push.

PR #1310, opened 2019-01-10
Add regenerated scalar reference data

jtkrogel:
Add a file with newly regenerated reference means and error bars for scalar type test runs.  The reference data in the file was created with the qtest script.  See additional notes in the file itself for details.  

The data in the file will be used to update erroneous statistical tests.  Additionally, the file contains reference data for more quantities than we are currently testing against, which will allow us to add tests of more quantities down the road if we see fit.  It is envisioned that more data will be added to the file as more statistical tests undergo similar quality checks and reference data updates.

A significant subset of the reference runs were performed independently by Hyeondeok using the same reference versions of the code, but a different machine and build environment.  All reference values computed by Hyeondeok and the qtest script agree within 2 sigma using the very small error bars from the reference runs (most within 1 sigma) and computed target error bars for the short runs agreed within 5%.  This independent confirmation gives much more confidence in the reference data produced using this method across the broader set.

### DIFF
--- a/tests/reference_data/scalar_reference_full.dat
+++ b/tests/reference_data/scalar_reference_full.dat
@@ -1,0 +1,847 @@
+
+# Description of data
+#   ref_code      :  QMCPACK version used to generate the reference data.
+#   ref_code_hash :  Git hash of QMCPACK version used to generate the reference data.
+#   test_name     :  Name of short test.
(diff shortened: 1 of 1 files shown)


ye-luo:
Okay to test

PR #1311, opened 2019-01-11
Fix: NoMPI builds broken by counting Jastrow factor bug

### LINKED ISSUE #1307: NoMPI builds broken by counting Jastrow factor
See e.g. https://cdash.qmcpack.org/CDash/viewBuildError.php?buildid=38283

bwvdg:
Fix #1307.


ye-luo:
Okay to test

PR #1313, opened 2019-01-14
Add build type info to reference data

jtkrogel:
As requested by @ye-luo, add build variant information to the test reference data file.


ye-luo:
Okay to test

PR #1321, opened 2019-01-17
support for llvm ASAN use -DLLVM_SANITIZE_ADDRESS=1

PDoakORNL:
MSAN is also included in theory but I have not gotten a complete set of instrumented libraries built yet and the blacklist/whitelist feature seems buggy. So it is not useful yet in my opinion.


CMake/ClangCompilers.cmake
```diff
@@ -32,6 +32,8 @@ SET( CMAKE_C_FLAGS_RELEASE     "${CMAKE_C_FLAGS_RELEASE} -ffast-math" )
 SET( CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -ffast-math" )
 SET( CMAKE_C_FLAGS_RELWITHDEBINFO     "${CMAKE_C_FLAGS_RELWITHDEBINFO} -ffast-math" )
 SET( CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -ffast-math" )
+SET( CMAKE_C_FLAGS_DEBUG     "${CMAKE_C_FLAGS_DEBUG} -fno-omit-frame-pointer -fstandalone-debug" )
```
ye-luo: Please a comment and separates from setting optimization flags.
No BOOST_HANA

ye-luo:
Please mention in the manual even just a line how to enable this feature.

prckent:
Maybe test drive on a few systems before putting in the manual? Key for me would be a recipe that I could use on a regular Linux workstation e.g. oxygen.

PDoakORNL:
These flags will work fine if you are building with the default llvm7 from spack plus everything else we use.

manual/installation.tex
```diff
@@ -324,6 +324,15 @@ \subsection{Configuration options}
                        e.g. "-n", "-np", etc.
 \end{verbatim}
 
+\item LLVM/Clang Developer Options\\
```
ye-luo: It is better to put them in the tool section near profiling tools.

prckent: Ye - this is a cmake config option and needs to go here.  This section should be comprehensive. I suggest merging now and adding an additional set of notes in the tools section after we have some experience.

ye-luo: This is a compiler or tool specific option.
Listing an entry here is fine but the following bulk description must be moved to the tools section.

We were not listing CMake options for profiling tools. If the policy is having all the CMake options listed here. We need to list all of them.

ye-luo: I'd like to have an entry as brief as possible. Even not listing the two options here. Just put a ref to the tool section.

ye-luo:
Okay to test

PR #1324, opened 2019-01-18
[WIP] Update test reference data in CMakeLists

jtkrogel:
Update the reference data in CMakeLists.txt files for a large subset of short and long scalar statistical integration tests.  The updates aim to correct the mean/error bar of the reference data.  In many cases the changes are small.  Efforts have been made to correctly preserve exact means and error bars where these are expected (exact wavefunctions, Hartree Fock, etc).

This is PR is mostly complete, but the error bars will benefit from another pass to be sure that all the right scaling factors were used for each test separately.  Once the remaining tasks are done, i will remove the WIP status.

Once this PR is in, updated tests that fail should be regarded as signifying actual change (and likely bugs).  

Given the large number of tests (and the large statistical fluctuations that introduces in aggregate), I recommend that we update to a 4 or 5 sigma error bar check instead of 3.  This is a one line change.

Remaining work:
- [ ] Update RMC reference values (single test)
- [ ] Final pass over error bars that disagree
- [ ] Update target error bars in scalar_reference_full.dat to match
- [ ] Add "good_test" label to all updated tests


prckent:
Would it be possible to get this updated a minimal amount and then merged, with to-be-finished items created as new issues?

jtkrogel:
Probably.  Likely best not to do this just prior to release.  My current priority is spin dependent 1RDM.  Afterward I can certainly come back to this before doing anything else.

PR #1325, opened 2019-01-21
Update XL compiler configuration

ye-luo:
This PR updates the XL compiler configuration.
There was a compiler bug failing a few unit tests. To trace the bug, I added more unit tests.
Now the bug has been identified and it is related to "omp simd" in a performance non-critical spot in OhmmsSoA/PosTransformer.h.
So I added a workaround and reported the bug to XL team. Once I got some feedback, it will be reported to OLCF as well.

Currently, the XL compiler is not ready for C++14 and I need to use a few hacks. This will be resolved in the near future. The build recipe on Summit is pasted here for reference and it should be simplified when XL is ready.
```
cmake -DCMAKE_C_COMPILER=mpixlc -DCMAKE_CXX_COMPILER=mpixlC -DENABLE_SOA=1 -DQMC_MIXED_PRECISION=1 -DCMAKE_CXX_FLAGS="-D__cplusplus=201402L -isystem /sw/summit/gcc/6.4.0/include/c++/6.4.0/powerpc64le-none-linux-gnu -qgcc_cpp_stdinc=/sw/summit/gcc/6.4.0/include/c++/6.4.0" -DCMAKE_CXX_STANDARD_LIBRARIES=/sw/summit/gcc/6.4.0/lib64/libstdc++.a ..
```


prckent:
What is the signature of the bug?      #pragma omp simd aligned(x,y,z) ? We have many other #pragma omp simd aligned in the source.

ye-luo:
"omp simd" was not correct in this matrix transpose case with XL compiler. We don't use transpose in other places.

trudeaun:
@ye-luo Hi Ye, do you know if a PMR was opened with IBM about the problem you mentioned?

ye-luo:
@trudeaun I don't know about the PMR. I forwarded my previous email to you. If you don't get it. Ping me via my Argonne email.

PR #1326, opened 2019-01-21
Default is to read all determinants through cutoff set to 1e-20

### LINKED ISSUE #1317: Updating gold files of converter tests needed
Here is one example.
https://cdash.qmcpack.org/CDash/testDetails.php?test=3997758&build=38934

anbenali:
quick fix for cutoff in converter test
close #1317


ye-luo:
Okay to test

PR #1334, opened 2019-01-24
Fix for distance table soa heap buffer read overflow

PDoakORNL:
Several issues here,
SoaDistanceTableAA::update is either mis documented or doing someting different. What are the valid values for iat? I'm reasoning this out and will write a unit test, but if you feel responsible for this feel free to chime in.

The underlying simd::copy_n `/src/simd/algorithm.h` function when called as in SoaDistanceTableAA after getAlignedSize and not just as an alias of std::copy_n may not be causing an error its still incorrect code if it is intended to 'copy a cache line' of data allocated with aligned_alloc or posix_memalign. While in release you can copy from an invalid pointer to T1* you shouldn't. This does trip a heap buffer read overflow in LLVM ASAN. However you can legally copy from the allocated memory via void* of one aligned_alloc vector data region to another. This should result in the "optimal" copy instruction as well since std::memcpy should be heavily optimized on all platforms and compilers. 

Note src/simd/allocator.hpp is also involved via getAlignedSize and use of this function with a type of size larger than ALIGN is a div 0 error. I need to think through what would happen is sizeof(T) was not a divisor of ALIGN.

If used as in SoaDistanceTableAA For example (not saying SoaDistanceTableAA ever does this but if you used its code as an example)
I have a aligned vector A of two doubles and vector B of 4 doubles and ALIGN=32 
first = A.data()
result = B.data()
copy_n(first,getAlignedSize<double>(2),result)
Now I have undefined garbage in B[2] and B[3]

### In conclusion
when std::copy_n, the developer is responsible to guarantee sufficient size of A and B.
The Temp_r was not providing sufficient size before the fix. The bug doesn't affect any result because the incorrectly copied data is not used in the later calculation.

### DIFF
--- a/src/Particle/SoaDistanceTableAA.h
+++ b/src/Particle/SoaDistanceTableAA.h
@@ -131,9 +131,9 @@ struct SoaDistanceTableAA: public DTD_BConds<T,D,SC>, public DistanceTableData
     if(iat==0) return;
     //update by a cache line
     const int nupdate=getAlignedSize<T>(iat);
-    simd::copy_n(Temp_r.data(),nupdate,Distances[iat]); // [iat - 1] also passes tests
+    std::copy_n(Temp_r.data(),nupdate,Distances[iat]); // [iat - 1] also passes tests
     for(int idim=0;idim<D; ++idim)
-      simd::copy_n(Temp_dr.data(idim),nupdate,Displacements[iat].data(idim)); // [iat - 1] also passes tests
+      std::copy_n(Temp_dr.data(idim),nupdate,Displacements[iat].data(idim)); // [iat - 1] also passes tests
   }
 
 };
--- a/src/Particle/SoaDistanceTableAA.h
+++ b/src/Particle/SoaDistanceTableAA.h
@@ -118,15 +118,22 @@ struct SoaDistanceTableAA: public DTD_BConds<T,D,SC>, public DistanceTableData
     return index;
   }
 
-  ///update the iat-th row for iat=[0,iat-1)
+  /** update the iat-th row for iat=[0,iat-1)
+   * into element iat of Distances and Displacements skiping element iat - 0
+   * doesn't really seem correct
+   * unless Distances[0] is original and Distances[1] is the proposed.
+   * in which case the brief above is just misinformation
+   * copying to Distances, Displacements iat or iat-1 both pass test_particles
+   * test_wavefunctions
+   */
   inline void update(IndexType iat)
   {
     if(iat==0) return;
     //update by a cache line
     const int nupdate=getAlignedSize<T>(iat);
-    simd::copy_n(Temp_r.data(),nupdate,Distances[iat]);
+    simd::copy_n(Temp_r.data(),nupdate,Distances[iat]); // [iat - 1] also passes tests
     for(int idim=0;idim<D; ++idim)
-      simd::copy_n(Temp_dr.data(idim),nupdate,Displacements[iat].data(idim));
+      simd::copy_n(Temp_dr.data(idim),nupdate,Displacements[iat].data(idim)); // [iat - 1] also passes tests
   }
 
 };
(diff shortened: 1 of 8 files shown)


src/Particle/SoaDistanceTableAA.h
```diff
@@ -118,15 +118,22 @@ struct SoaDistanceTableAA: public DTD_BConds<T,D,SC>, public DistanceTableData
...
   }
 
-  ///update the iat-th row for iat=[0,iat-1)
+  /** update the iat-th row for iat=[0,iat-1)
+   * into element iat of Distances and Displacements skiping element iat - 0
+   * doesn't really seem correct
+   * unless Distances[0] is original and Distances[1] is the proposed.
+   * in which case the brief above is just misinformation
+   * copying to Distances, Displacements iat or iat-1 both pass test_particles
+   * test_wavefunctions
```
ye-luo: Non of the unit test covers this function.
The math needs iat instead of iat-1.
Only the lower triangle of the N by N(Npad) symmetric distance matrix is updated excluding the diagonal.
The correctness of the upper triangle is ignored.
Currently the distances is still N by Npad. Because some code path still uses the upper triangle which is correct when the full table is updated from scratch.
The displacements only stores the lower triangle with each row padded.
Eventually all the codepath should be fixed using only the lower triangle and distances should store only half.

src/simd/algorithm.hpp
```diff
@@ -25,13 +25,20 @@ namespace qmcplusplus {
...
+    inline void copy_n(const T1* restrict first, size_t count, T2* restrict result) 
       {
-//#pragma omp simd 
-        for(size_t i=0; i<count; ++i)
-          result[i]=static_cast<T2>(first[i]);
+//#pragma omp simd
+
+	//If anything is going to emit SIMD instructions
+	//std::memcpy should
+	std::memcpy(result, first, count * sizeof(T1));
```
ye-luo: There is no alignment assumption.
memcpy copy memory by bytes. When we need to copy float to double, this is wrong.

src/Particle/SoaDistanceTableAA.h
PDoakORNL: I think you must write a unit test showing this in code. I'm still not sure I understand you properly and the comment for update conveyed none of this at all.  That said this PR doesn't really deal with this and if it is currently correct it should behave the same.  The meat is in the other files.

This comment or something more "correct" should go here but even this is a huge improvement over the original.

ye-luo:
getAlignment 0 issue can be guarded with static_assert for zero cost.

ye-luo:
Your last example can happen. The developer is responsible to make sure both A and B have sufficient length to cover getAlignedSize(2).

src/simd/algorithm.hpp
ye-luo: I can't quite understand your comment. I think the old one was correct. We can stay on the phone and talk after the meeting

prckent:
- Requirements should be documented
- Code should protect against errors where practical

ye-luo:
Please refrain from formatting the code. It makes code review harder.
I actually prefer to get ride of simd::copy_n. The performance matters little and we don't need to maintain it ourselves.

ye-luo:
In which system, did you get an overflow? What are the vector and copy sizes causing overflow?

prckent:
@ye-luo Everyone has permission to format code they work on.

ye-luo:
@prckent We need to first format whole code once. Then everyone is free to format.

prckent:
@ye-luo Please add a comment about why/where padding is done/necessary. Some but not all (?) structures are being padded to SIMD width?...

ye-luo:
@prckent Padding is not a global requirement. Only the performance critical portion needs to be padded if it is relevant.

ye-luo:
I do approve Peter's change but since I made non cosmetic change to the code as well. I can't approve this PR.

PR #1336, opened 2019-01-25
Fix Multiquinticspline 1D heap overflow

PDoakORNL:
Problem:
Another HEAP BUFFER OVERFLOW Should only effect LCAO.  LogGridLight return loc outside of the integer indexing implicit in the grid. 
This resulted in MultiquinticSpline reading data outside of the allocated Coeffs table when LogGridLight returned an out of bound loc value. Everyone seemed to assume a r (distance) larger than the largest r_value in the log grid would never be passed but it happened with some frequency in @anbenali example run.

Fix:
LogGridLight now refuses to hand back an index larger than the highest grid point index. Multiquintic spline returns what I hope is the correct constant value outside of r_max. Either would fix this particular problem but I think both are required for the classes to be standalone safe.

TBD:
Unit tests that verify both classes don't return to their bad behavior.

Still to be resolved how and why r's > rmax are getting into the SoaDistanceTableAA. The big r's are close enough to r_max that some sloppy floating point math somewhere could be to blame.


ye-luo:
LCAO is different from other use cases. For each atomic center, all the AOs are evaluated together. That is why it is being called Multi. Just like the Spline SPO.

ye-luo:
Actually, it is expected, at least by me,  that r>=r_max will be passed in.
In this case, SoaDistanceTableBA is relevant.
If the math is bad at r very close to r_max, the quintic spline can be the problem.
When we spline a function with n grid point, the number of knots need to be n+2 if it is cubic spline.
Probably more knots for higher orders.

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
```diff
@@ -123,60 +136,73 @@ struct MultiQuinticSpline1D
...
-  inline void evaluate(T r, T* restrict u, T* restrict du, T* restrict d2u) 
+  /** Approximation of getting the asymtotic constant spline values
+   */ 
+  inline void assignLimitValues(T* u)
+  {
+    int loc = myGrid.size() - 1;
+    const size_t offset=loc*6;
+    const value_type* restrict a=(*Coeffs)[offset+0];
+    for(size_t i=0; i<num_splines_; ++i)
+      u[i]=a[i];
```
ye-luo: This doesn't seem right. All the AO are expected to be zero beyond r_max.

ye-luo:
Correct myself. It seems that 
`if(r_new>Rmax) continue;` is always placed before `MultiRnl->evaluate`
How can r_new>Rmax get passed in?
For sure, it needs to be changed to `r_new>=Rmax`.
I'm wondering if the Rmax outside and the Rmax inside is not consistent.

anbenali:
This would explain a lot of behaviors (but at this point、any work to clean
the lcao code gives me hope)

PDoakORNL:
The CuspCorrectionAtomicBasis now receives an r_max in a sane way and doesn't evaluate when r >= r_max, however I'm not yet zeroing the outputs of its evaluations.

ye-luo:
In the spline code, when r>=rmax happens, the code should stop. Then the developer is responsible to protect the calling code.

PDoakORNL:
MultiQuinticSpline1D now throws a std::domain_error if it is called with r>=rmax. And my test cases don't throw it due to corrections in code above in the call stack.

Also fixed the magic number hard code of 100 for r_max in RadialOrbitalSetBuilder<COT>::finalize() this seems like it was added instead of actually calculating a valid r_max for slaters.  I now handle this by setting the m_rcut_safe to at least 100 if a slater orbital is included in the set.

prckent:
The "const double eps=1e-6;" in find_cutoff is also magic

PDoakORNL:
It seems to be what the gamess orbitals are coming in with for their eps, but I'm not really sure.

I think both this and the max cutoff for Slater orbitals / max max cutoff allowable for gaussians should come in from input with a sensible default defined. That default for max_rmax/slater cutoff should get passed into the constructor for the RadialOrbitalSetBuilder. 

The default for radial grid cutoff eps sounds like it should be a global property and not be defined inside a terminal function but accessible to it from global read only property data.

prckent:
Regardless, if this PR makes the code less illegal, that is good progress.

PDoakORNL:
IMHO this is done. LCAO still needs work but this is one step in the right direction.

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
PDoakORNL: this now zeros

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
```diff
@@ -305,8 +336,11 @@ struct MultiQuinticSpline1D
       }
     }
   }
+  
+  int getNumSplines() const { return num_splines_; }
+  void setNumSplines(int num_splines) { num_splines_ = num_splines; }
```
ye-luo: Add inline keyword before these two functions.

src/QMCWaveFunctions/lcao/SoaAtomicBasisSet.h
```diff
@@ -27,15 +27,15 @@ namespace qmcplusplus
     struct SoaAtomicBasisSet
     {
       typedef ROT RadialOrbital_t;
-      typedef typename ROT::value_type value_type;
+      using ValueType = typename ROT::value_type;
```
ye-luo: ValueType in the source code is reserved for data types turning complex when the code is built complex.
In the AO basis, it is always real only even in the future when complex support of LCAO is added.
Please restore this change as well as all the ValueType

src/QMCWaveFunctions/lcao/RadialOrbitalSetBuilder.h
```diff
@@ -177,9 +172,9 @@ class RadialOrbitalSetBuilder: public MPIObjectBase
 };
 
   template<typename COT>
-    RadialOrbitalSetBuilder<COT>::RadialOrbitalSetBuilder(Communicate* comm)
+  RadialOrbitalSetBuilder<COT>::RadialOrbitalSetBuilder(Communicate* comm, int radial_grid_size)
```
ye-luo: Could you explain a bit how radial_grid_size is handled after the change?
This maybe affects the input of grid.

src/QMCWaveFunctions/lcao/RadialOrbitalSetBuilder.h
```diff
@@ -400,11 +397,11 @@ class RadialOrbitalSetBuilder: public MPIObjectBase
   void RadialOrbitalSetBuilder<COT>::addGaussian(xmlNodePtr cur)
   {
     int L= m_nlms[1];
-    using gto_type=GaussianCombo<double>;
+    using gto_type=GaussianCombo<ValueType>;
```
ye-luo: Using double is on purpose. When constructing splines in the mixed precision build, Gaussian are always initialized in double. After the spline parameters are constructed, the evaluation goes to lower precision.

src/QMCWaveFunctions/lcao/SoaCuspCorrectionBasisSet.h
```diff
@@ -145,10 +166,7 @@ struct SoaCuspCorrection
    */
   inline void evaluateVGL(const ParticleSet& P, int iat, VGLVector_t& vgl)
   {
-
-    constexpr RealType czero(0);
-
-    myVGL=czero;
+    myVGL=0.0;
```
ye-luo: More of a personal taste. I prefer RealType(0) instead of 0.0. Complier should turn RealType(0)  to 0.0 or 0.0f based on RealType.

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
```diff
@@ -52,9 +59,14 @@ namespace qmcplusplus
...
       }
 
+      // Consider a way to get this without branch via math operations
       inline int locate(T r) const
       {
-        return static_cast<int>(std::log(r/lower_bound)*OneOverLogDelta);
+        int loc=static_cast<int>(std::log(r/lower_bound)*OneOverLogDelta);
+	// if(loc >= r_values.size())
+	//   loc = r_values.size() - 1;
+	return std::min(static_cast<unsigned long>(loc), r_values.size() - 1);
```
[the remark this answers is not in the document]
PDoakORNL: But that may not be zero just close. I thought you wanted to throw the exception?

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
```diff
@@ -66,24 +78,28 @@ namespace qmcplusplus
...
-        //return r-lower_bound*std::exp(loc*LogDelta);
+        loc=locate(r);
         return r-r_values[loc];
       }
+
+      inline int size() { return r_values.size(); }
     };
 
   /** multivalue implementation for OneDimQuintic
+   *  While its implied this spline could have complex values
```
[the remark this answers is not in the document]
PDoakORNL: complex gaussian basis do exist

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
```diff
@@ -117,88 +134,94 @@ struct MultiQuinticSpline1D
     return myGrid.upper_bound;
   }
 
+  /** Approximation of getting the asymtotic constant spline values
+   *  Since this is getting used only for AO's should be 0
+   */ 
+  inline void assignLimitValues(T* u)
```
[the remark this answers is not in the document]
PDoakORNL: I thought it might be convenient to only have to change from std::fill to another method in one place.

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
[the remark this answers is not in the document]
ye-luo: I'm thinking of throw the expect before line 65.

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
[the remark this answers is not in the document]
ye-luo: In theory, yes. Do any of our workflow using complex gaussian basis? Are there any major codes support that seriously? Do we plan to support that in the future?

src/QMCWaveFunctions/lcao/SoaAtomicBasisSet.h
PDoakORNL: ValueType having a special global meaning like that is a design issue. You should be able to build complex and real simultaneously and the local meaning of a type should not require special global knowledge of the code. This is the ValueType for a radial function used by this basis set. It must be the type returned by RF(r).  This could be mathematically a real, complex, quaternion or whatever.

src/QMCWaveFunctions/lcao/RadialOrbitalSetBuilder.h
PDoakORNL: Now it is a magic number in one place.  It will be easier to work on making it something more optimized than just 1001 typed into the source in multiple places.

src/QMCWaveFunctions/lcao/SoaCuspCorrectionBasisSet.h
PDoakORNL: The compiler produces exactly the same instructions either way. I took this to be an attempt to write generic code that would support complex types.
Only in that case is T t = T(0) different than writing (T t = 0 | T t = 0.0 | T t = 0.0f); 

The only time you need to write 1.0f instead of 1 for a literal constant is if you have something like
T t = 1 / 2; In modern C++, definitely not true in Fortran.

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
```diff
@@ -456,18 +459,19 @@ namespace qmcplusplus
       // loop over MO index - cot must be an array (of len MO size)
       //   the loop is inside cot - in the multiqunitic
       SoaCuspCorrection::COT *cot = new CuspCorrectionAtomicBasis<RealType>();
-      cot->AOs.initialize(radial_grid, orbital_set_size);
-      cot->ID.resize(orbital_set_size);
-      for (int mo_idx = 0; mo_idx < orbital_set_size; mo_idx++) {
-        cot->ID[mo_idx] = mo_idx;
-      }
+      cot->initializeRadialSet(*radial_grid, orbital_set_size);
+      //How is this useful?
```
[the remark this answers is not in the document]
markdewing: The indirection through the ID array was there in CuspCorrectionAtomicBasis when I started - I don't know the intended purpose.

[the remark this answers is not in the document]
markdewing: It can probably be removed.

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
```diff
@@ -435,9 +435,12 @@ namespace qmcplusplus
 
     std::vector<bool> corrCenter(num_centers, "true");
 
+    //What's this grid's lifespan?  Why on the heap?
```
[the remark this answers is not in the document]
markdewing: I think it can go on the stack instead.   OneDimQunticSpline only uses it to set a type (??), and MultiQuinticSpline grabs some values from the grid and doesn't keep a copy of the pointer (it should take a const reference instead).

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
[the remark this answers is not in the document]
prckent: @PDoakORNL Take care not to get caught up in the sticky mess that is LCAO - cleanup is good, but not if it delays this PR...

src/QMCWaveFunctions/lcao/RadialOrbitalSetBuilder.h
ye-luo: Probably it should be replace with OHMMS_PRECISION_FULL

src/QMCWaveFunctions/lcao/SoaAtomicBasisSet.h
ye-luo: I think the reason of using value_type is to do some local template. ValueType is reserved for global meeting. This is an convention. The real issue is mixing them here.

src/QMCWaveFunctions/lcao/SoaCuspCorrectionBasisSet.h
prckent: IMO either are fine. I tend to prefer 0.0 since it is clear that a floating point number is intended when the source is read. If we truly think it matters, the coding guide should be updated first.

src/QMCWaveFunctions/lcao/RadialOrbitalSetBuilder.h
prckent: Special handling for extra precision in mixed precision builds should be commented. A perfect example of something subtle but important ("Always use full precision here since the fudge function is numerically sensitive in large supercells") that will be not obvious to anyone but the original author.

prckent:
Please remember that @anbenali is waiting on this. The logical overhaul needs to be done later. For now avoiding the illegal memory accesses should be the focus - which is the title of the PR.

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
PDoakORNL: inline is implicit, if you want it to avoid a multiple definition error, which I haven't see yet it could be added if warranted but I'd rather see that, that's a bad code smell that should be justified.

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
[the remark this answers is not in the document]
PDoakORNL: If we throw in LogGridLight we never do this, I deleted it.

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
ye-luo: Class member function should have no multi-definition issue anyway. inline is never enforced it is just a suggestion the the compiler. Since the function is so small, I expect the compiler to do the inlining anyway.

PDoakORNL:
test this please

src/QMCWaveFunctions/lcao/MultiQuinticSpline1D.h
PDoakORNL: You are right because class member function defined in the class declaration are implicitly inline so its always redundant. If it is advantageous to inline the compiler will. What should happen here is all the other member definitions need inline removed.

ye-luo: I don't think all the member functions are inlined. If they are big, they will not be inlined.
The linker knows how to deal with multi definition of class member functions.
inline keyword is always a suggestion to the compiler. The existing use of inline is appropriate I think and there is no need of removal.

PDoakORNL: "A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function. "

PDoakORNL:
I think to issue here is fact that each function has its own r_max but we eventually want a common grid and r_max

ye-luo:
I think when the multi radial function class is being constructed, the rmax can be different from the input. However, once the class has been built. There should be just one common Rmax. When the evaluation being called, the caller function needs to query Rmax.

prckent:
Hopefully we are done with updates to this code. Assuming this PR really does fix the bugs, I don't think anyone will being science limited by this code or its performance.

PR #1342, opened 2019-01-29
 Fix to gamessAsciiParser.cpp for ALDET with cores

mcbennet:
Fix for GAMESS ALDET conversions when there are a non-zero number of core orbitals.
The changes modify the number of orbitals in the configurations to be equal to nca+nea (ncb+neb) -- and all core orbitals are occupied.


markdewing:
okay to test

PR #1344, opened 2019-01-30
Remove QMC_BUILD_LEVEL

### LINKED ISSUE #1051: Remove and/or rename QMC_BUILD_LEVEL

markdewing:
The default had been set to 3, and was only changed to build the sandbox apps.
Remove the ifdefs in most places, and replace with QMC_BUILD_SANDBOX_ONLY where appropriate.
Fixes #1051.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,8 +60,8 @@ MESSAGE(STATUS "QMC_SYMLINK_TEST_FILES = ${QMC_SYMLINK_TEST_FILES}.  ${SYMLINK_M
 ######################################################################
 # Build level
 ######################################################################
-SET(QMC_BUILD_LEVEL 3 CACHE INTEGER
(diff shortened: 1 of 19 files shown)


markdewing:
I think it is the latter (QMC_BUILD_SANDBOX_ONLY is not defined if  the corresponding CMake variable is off.   I will change the uses to check for definition rather than value.

PR #1348, opened 2019-01-31
Relocate command line save_wfs as XML input tag save_ceofs

### LINKED ISSUE #1259: Test of save_wfs option is needed
Per #1234

ye-luo:
Manual updated.
Test added but it doesn't test seriously.
Solve part of #146
Close #1259


ye-luo:
I updated the printout reflecting the input and output of h5.
cmake hijacks all the output which I can't use for grep.
Here in the save_coefs test, check.sh only checks the existence of scalar.dat at the moment. I need more work to use check_scalar.py.
For coefficients, I don't have a good way to check the real-space spline coefficients.

In there restart test, the check.sh do check the numbers. We are still failing that in certain cases.

PR #1349, opened 2019-01-31
Updating converter/test_aldet6 gold files

mcbennet:
Given #1342 , the aldet6 test will now fail. Updating gold files to mend this.


markdewing:
okay to test

PR #1351, opened 2019-02-01
Nexus: fix CHGCAR conversion

jtkrogel:
Small fix for XSF to CHGCAR conversion.  Bug introduced with changes to internal representation of XSF data a while back (C vs. Fortran array order).


markdewing:
okay to test

PR #1353, opened 2019-02-01
Debug build was broken.  Check the debug build

PDoakORNL:
When you are developing first build debug, test, then build release and test, then push.  

All the assert's and what not do nothing otherwise.


ye-luo:
My mistake. I changed the code from pointers to references without checking all the builds.

PR #1359, opened 2019-02-05
Reenable and test evaluateGradSource functions.

rcclay:
This PR revives and adds unit tests for limited APIs used to compute ionic gradients of the trial wave function, electron gradient, and electron laplacian objects.  These changes are not user facing yet, but are a necessary prerequisite for computing Pulay terms of nonlocal PP derivatives and for constructing ZVZB estimators.  These are the major changes:
1.)  add evaluateGradSource(...) functions to new BsplineOrbitalSet object. 
2.)  add and implement evaluateGradSource(...) to J1OrbitalSoA object.  
3.)  add finite difference test for ionic \nabla (ln \Psi_T) to wave function tester.  
4.)  add unit test for one body jastrow orbital.  

Note:  Wave function tests of ion derivatives will crash currently, because working arrays are not allocated automatically.  APP_ABORTS are thrown until initialization and checks of existing code are performed.


rcclay:
Actually, hold on this.  Might be a mixed precision issue.

rcclay:
We cool.  For now...

prckent:
Thanks. Happy to merge once the CI passes. Do you expect to do the implementation to fix the crash in the wave function tests of ion derivatives shortly?

rcclay:
@prckent I was going to push this down the road, since this doesn't impede development, and there are some bigger fish to fry at the moment.  The "fix" is automatically setting "ionDerivs" value from XML, and double checking that the the determinant formulae are right, which only matters for the LCAO code.  For now, hard coding ionDerivs=true and commenting out APP_ABORT creates working code for einsplines, which is what I care about.  I'm happy to push this up the priority list though...  this commutes with everything.

prckent:
@rcclay Please follow the most expeditious implementation route for forces. If needing we can use  the test labeling system to avoid the failing tests contaminating the dashboard.

markdewing:
Where did the unit test values originate from?

rcclay:
Dumped the various gradients corresponding to the configuration in the unit test using the AoS code.  I guess to make this rigorous, I could compute these from finite differencing.

rcclay:
Ok.  Using same parameters as in the unit test, I can reproduce the values in the unit test using the wave function tester.  Comparison between what comes out of evaluateGradSource and finite difference shown.  

[wftest.000.txt](https://github.com/QMCPACK/qmcpack/files/2837527/wftest.000.txt)

markdewing:
This is sufficient for the unit tests - it's more important to document where the values came from than the particular means of generating them.

markdewing:
okay to test

ye-luo:
The state machine ionDerivs seems not very necessary.
1) It has been used to resize data objects in DiracDeterminant. If this is the only purpose, It is better to just create a function doing these resize and call it in the places where these data objects are used.
2) It is also used inside SPOSet::evaluateGradSource to skip the whole calculation. Why these function needs to be called with all the computation skipped.
@rcclay do you think we can get rid of ionDerivs everywhere?

rcclay:
@ye-luo I'm happy to lose the ionDeriv state machine--I only used it because it was preexisting.  I haven't quite figured out how I want to (re)organize everything on the initialization and run side, so a better solution might be a PR or two out.

ye-luo:
Data objects used during VMC/DMC for energy only are always resized.
Those for particular estimator or optimization are resized only at their use.

prckent:
Unnecessary state machines are unnecessary. Excellent candidate for removal.

PR #1360, opened 2019-02-05
Isolate boost::multi hdf template specialisations.

fdmalone:
Removes include of boost::multi template specialisations from being included in hdf_archive by default. Prevents boost::multi related warnings when building the real space code with intel.

PR #1361, opened 2019-02-06
Re-enable AFQMC restarts

fdmalone:
Allows AFQMC calculations to be restarted again. Currently only the serial writing of restart files is permitted.


prckent:
OK to test

fdmalone:
Pushed a fix following merge of PR #1360.

PR #1362, opened 2019-02-06
Pyscf lcao converter fixes

camelto2:
This PR attempts to resolve some inconsistencies in the LCAO converters. These fixes are needed for the crystal converter, since it depends on LCAOH5Parser. 

First there is a typo in LCAOH5Parser which reads SpinRestricted from the hdf5 file variable "SpinResticted" which is resolved. Also, the pyscf converter does not ever write a SpinRestricted variable, and instead is working with SpinUnRestricted. I fixed the PyscfToQmcpack converter to use and write SpinRestricted as a variable. 

Additionally, there seems to be an inconsistency with the use of spin_multiplicity in the PyscfToQmcpack and LCAOH5Parser. LCAOH5Parser is based on QMCGaussianParserBase, which uses SpinMultiplicity correctly as 2S+1. However, the PyscfToQmcpack writer for the hdf5 writes "spin", which uses the pyscf spin variable which is 2S. LCAOH5Parser was reading SpinMultiplicity from this "spin", which is incorrect. I have added a change to make the LCAOH5Parser read in the spin as 2S, and set SpinMultiplicity correctly as 2S+1. Another change would be to save in the hdf5 "spin_multiplicity". Either works, but it will be needed to be correct for the crystal converter. 

@anbenali


markdewing:
okay to test

prckent:
Test this please

anbenali:
These modifications are light and I will adapt to them for the following pushes! thanks for the additions!

Issue #1329, opened 2019-01-22
Formatting Issues in manual


PDoakORNL:
List of bad formatting in manual.  If you wrote this section consider fixing it.
Build the manual to see how your fix looks.
See below for the current guidance on formatting different sections

- Additional Tools
  - [x] badly formatted XML
  - [x] Aquiring  pseudopotentials -- verbatim section overruns.
- Lab3
  - [x] appendix XML
- Lab4
  - [x] Supertwist section verbatim overruns

## Formatting suggestions
XML should be included like this
```latex
\begin{lstlisting}[style=XML]
  <simulationcell>
    <parameter name="lattice">
      3.8       0.0       0.0
      0.0       3.8       0.0
      0.0       0.0       3.8
    </parameter>
    <parameter name="bconds">
       p p p
    </parameter>
    <parameter name="LR_dim_cutoff"> 20 </parameter>
  </simulationcell>
\end{lstlisting}
```

prckent:
Peter - If you can, please assign specific fixes to individual people. Otherwise they won't get done.

PR #1364, opened 2019-02-06
Consistent lstlisting and inlinelstlisting for Manual

PDoakORNL:
Addressing #1329 and additional issues.
I've also updated the contributing to the manual section.
Note that tabularx cannot take "verbatim" elements so we cannot use the inline formats there.
Those remain manually specified \texttt \textit \textbf etc.


PDoakORNL:
Also fixed the overflowing xml spec table issue

ye-luo:
What is this file?
manual/utils/fixing_verbatims.org

ye-luo:
I can't build the manual with pdflatex due to `\ttfamily` but it can be caused by my old textlive2015.
Can you try pdflatex?

prckent:
The manual will only build with recent texlive 2018 and maybe 2017 from memory. Use xelatex and the script.

I'll check if this builds in our VM.

ye-luo:
@prckent Did you succeed with the pdflatex from textlive 2018?
I can build the manual in the develop branch with texlive 2015.
Here `\ttfamily` stops pdflatex with both textlive 2015 and 2017.
Not a fan of xetex.

jtkrogel:
I can build with pdflatex from texlive 2015.  The changes in listing fonts, etc, amount to a big improvement.

At the moment, the pdflatex listings look a lot better than the xelatex ones.  There is a change in verbatim font that significantly degrades readability in xelatex IMO.  See e.g. Listing 13.4 with pdflatex and xelatex.  I'm guessing the font could be changed in xelatex?

Another small beef: pink urls?

PDoakORNL:
Link color changed.  TexLive 2018 pdflatex build works. @jtkrogel its hard to know which font you are picking up for the XeLaTeX build.  That build first tries for DejaVu Sans Mono and if it can't find that Courier New and if it can't find that you end up with the default monospace font.

ye-luo:
Texlive 2015 pdflatex works as well now. @PDoakORNL could you please resolve the conflict?

prckent:
I was able to build in the VM, as expected. For reference this is a full "binary" install of TexLive2018. 
Since 99% of users interact with the manual via the PDF produced in the VM and the *official* build_manual.sh, how this looks is most important. Ideally the PDF produced other ways would not be much different, but we shouldn't spend much effort on it.

- The XML listings are much improved
- With the Link color fix, the URLs are now blue and clearly visible.
- The table of contents entries, which are clickable, are now ~black. Before the link color fix in this PR they were red. Making these visible again would be useful. Before this PR they were red outline boxes around black text - ugly but usable.
- The font in the PDF still the same as https://docs.qmcpack.org/qmcpack_manual.pdf , which is fine by me.

@PDoakORNL Can you tweak the table of contents entries color?

PDoakORNL:
Yes what color, could be blue too, I think that is commonly understood as a link. I thought all the red looked a little crazy.

prckent:
I am happy with the PDF produced in the VM with the latest updates.

ye-luo:
@markdewing is not available today.

PR #1365, opened 2019-02-06
Improve Nexus handling of excited state calculations

jtkrogel:
This PR improves the Nexus interface to SeeK-path for excited state calculations.  The QMCPACK lab on excited states is updated to reflect the new simplified interface.

Interaction with SeeK-path is now pulled up into the Structure class.  Users request SeeK-path functionality via two keywords added to generate_structure.  

If keyword use_prim is set (`use_prim=True`), the Structure class uses SeeK-path to automatically determine the primitive cell and sets the internal axes and atomic positions to match.

If keyword add_kpath is set (`add_kpath=True`), the Structure class uses SeeK-path to automatically add k-points corresponding to high symmetry paths in the first BZ to the structure object.  This can be used subsequently to easily perform band structure calculations and select k-points to target for QMC excited state runs.

Further refinements to the lab will be made by @kayahans following the merge of this PR.


prckent:
OK to test

PR #1366, opened 2019-02-07
Re-enable AFQMC estimator HDF5 estimator output

fdmalone:
Prints back propagated RDMs to file with a controllable blocking parameter to reduce output. I also added a unit test. In the process of writing the test I discovered a bug for UHF trial wavefunctions, which is also now fixed.


prckent:
OK to test

ye-luo:
Test this please

PR #1371, opened 2019-02-08
Refactor the src/spline2 folder and allows using memory allocator in many places.

ye-luo:
1. VectorSoaContainer accepts allocator as a template argument.
2. Redesign BsplineAllocator which now takes a memory allocator as a template argument.
3. MultiSpline uses BsplineAllocator underneath and thus accepts memory allocator as well.
4. All the C code are removed from src/spline2. Most of them are turned into C++ template code in BsplineAllocator. Thus malloc/free and customized einspline_alloc/einspline_free are replaced by new/delete and the allocator.allocate/deallocate.
5. Initializing the multi-spline structure from already initialized single spline is the only way from now on. Previously supported initialization from any array of real-space values directly to the multi-spline structure is dropped to keep the overall code tight. As an alternative, it is simple to just build a single spline, initialize it with the data array and copy it into the big table.

Hopefully this PR reduces the hassle in the spline2 folder.


src/spline2/BsplineAllocator.hpp
```diff
@@ -0,0 +1,263 @@
...
+        for (int iy=0; iy<My; iy++)
+          for (int iz=0; iz<Mz; iz++) {
+            intptr_t doffset = iy*Mz+iz;
+            intptr_t coffset = iy*Nz+iz;
+            einspline::find_coefs_1d(spline->x_grid, xBC, data+doffset, My*Mz,
+                            spline->coefs+coffset, Ny*Nz);
+          }
+
+        // Now, solve in the Y-direction
+        #pragma omp parallel for
```
PDoakORNL: These pragma omp parallel for's are the only thing that has come over from einspline that bothers me.  But I expanded the test program and they do help in the case of larger grids.  It seems for the smaller sizes we generally use that they don't do a great deal ~15%.

ye-luo: Currently, the splines are constructed sequentially. Threads are anyway idle and gaining some thing is always good.

ye-luo:
Some hints for reviewers. The rough idea is like the following.
MultiSpline is like a container class std::vector which takes an allocator and owns a memory object.
BsplineAllocator is more of a hyper allocator doing a bit more than a memory allocator. It generally serves allocating and deallocating the memory object in MultiSpline although BsplineAllocator does a bit more.

src/QMCWaveFunctions/BsplineFactory/contraction_helper.hpp
```diff
@@ -0,0 +1,72 @@
...
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2019 QMCPACK developers.
+//
+// File developed by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//
+// File created by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
```
PDoakORNL: One of these functions is only used in test code and two only in SplineR2RAdoptor but they are just floating around here with no scope. Where are they going?

ye-luo: They are all used in the C2R, C2C and R2R adopters and it is the reason of the move.

ye-luo: I actually didn't pay attention to move the test. Can I do it in a later PR?

PDoakORNL: ok why aren't they scoped into those? Then they should go into the base class.  
call via
```C++
BaseType::v_m_v(...)
```

ye-luo: Did you mean SplineAdoptorBase? I can move them in.

ye-luo: Actually not. The Base class was not specialized to do 3D by definition.

PDoakORNL: Yes that's what I meant. The unit test can stay put if it can adapt.

PDoakORNL: Or write an issue for it. And disable it for the moment.

ye-luo: An issue is created.

grahamlopez:
Test this please

PR #1372, opened 2019-02-09
Rename initGPU to finalizeConstruction for more general use

ye-luo:
This is needed for my OpenMP offload.
Hooks finalizeConstruction in BsplineSet.h
PR separately for easy review.

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/BsplineSet.h
+++ b/src/QMCWaveFunctions/BsplineFactory/BsplineSet.h
@@ -82,16 +82,9 @@ struct BsplineSet: public SPOSet, public SplineAdoptor
     return SplineAdoptor::estimateMemory(nP);
   }
 
-  inline void evaluate(const ParticleSet& P, int iat,
-                       ValueVector_t& psi, GradVector_t& dpsi, ValueVector_t& d2psi)
+  inline void finalizeConstruction()
   {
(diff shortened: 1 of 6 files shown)


prckent:
Please add a description of what this function is responsible for. Then we will see about finding an improved name.

ye-luo:
Here is what I put in SPOSet.h
```
  /** finalize the construction of SPOSet
   *
   * for example, classes serving accelerators may need to transfer data from host to device
   * after the host side objects are built.
   */
```

prckent:
Ideally in a future refactor the need for this call would disappear, with one constructor/setup/initialization call actually doing all the setup & initialization. Any chance to get rid of it now? Given that there are no arguments to the call, why can it not be done earlier?

ye-luo:
The initialization of spline takes 3 steps.
1. create the memory space for splines
2. initialize the values or read from pre dumped files.
3. GPU only. Transfer to the device.
Currently the builder is not aware of the accelerator. It calls finalizeConstruction in step 3 and let the implementation decide. If we need to have the builder aware of the GPU, both the builder and the object class needs to be GPU aware and way more lines of codes will be needed than only keep the lightweight GPU specialized code (step 1 and 3) all live in the object.

prckent:
I was thinking about a possible merge of 2 and 3 perhaps 1-3 in future. No reason these could not be merged in improved overrall design. The builders should definitely not know any more than they do now - they are already too complex already for too little return :-)

Issue #1369, opened 2019-02-07
eval_UBspline for single splines with natural boundary conditions do not protect from out of bounds values


PDoakORNL:
In the case of `xBC.lCode==NATURAL` floating point input value x => x_grid.end results in a heap buffer overflow read.
This is the result of turning that x into an i for coefficient access.
Similar issue for the 2d and 3d.

PDoakORNL:
Also UBsplines of all types with PERIODIC boundary conditions are not periodic.

PDoakORNL:
These are both expected behavior from the old school einspline c library.  We now assert on out of bounds values for UBSplines.  Of course you will not get the performance hit of checking or the safety unless you build in Debug. But ultimately this is something that should not be address in einspline but in a future C++ bspline implementation.

PR #1373, opened 2019-02-11
Fix the residual from getSplineBound.

ye-luo:
Related to issue #1369 but in a different place.
spline2 is only intended for PBC or anti-PBC.

### DIFF
--- a/src/spline2/MultiBsplineData.hpp
+++ b/src/spline2/MultiBsplineData.hpp
@@ -31,9 +31,24 @@ namespace spline2
 template<typename T, typename TRESIDUAL>
 inline void getSplineBound(T x, TRESIDUAL& dx, int& ind, int ng)
 {
-  T ipart;
-  dx=std::modf(x,&ipart);
-  ind = std::min(std::max(int(0),static_cast<int>(ipart)),ng);
+  // lower bound
+  if (x < 0)
+  {
+    ind = 0;
+    dx = T(0);
+  }
+  else
+  {
+    T ipart;
+    dx=std::modf(x,&ipart);
+    ind = static_cast<int>(ipart);
+    // upper bound
+    if (ind > ng)
+    {
+      ind = ng;
+      dx = T(1) - std::numeric_limits<T>::epsilon();
+    }
+  }
 }
 
 /** class for cublic spline parameters
--- a/src/spline2/tests/test_multi_spline.cpp
+++ b/src/spline2/tests/test_multi_spline.cpp
@@ -34,13 +34,13 @@ void test_spline_bounds()
   // check clamping to a maximum index value
   x = 11.5;
   spline2::getSplineBound(x, dx, ind, ng);
-  REQUIRE(dx == Approx(0.5));
+  REQUIRE(dx == Approx(1.0));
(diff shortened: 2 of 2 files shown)


src/spline2/MultiBsplineData.hpp
```diff
@@ -31,9 +31,24 @@ namespace spline2
...
 inline void getSplineBound(T x, TRESIDUAL& dx, int& ind, int ng)
 {
-  T ipart;
-  dx=std::modf(x,&ipart);
-  ind = std::min(std::max(int(0),static_cast<int>(ipart)),ng);
+  // lower bound
+  if (x < 0)
+  {
+    ind = 0;
+    dx = T(0);
```
PDoakORNL: So can this spline be std::complex or some other type that needs to be explicitly constructed with 0 to be zero?  The T(0) implies that.

src/spline2/MultiBsplineData.hpp
```diff
@@ -31,9 +31,24 @@ namespace spline2
...
+  else
+  {
+    T ipart;
+    dx=std::modf(x,&ipart);
+    ind = static_cast<int>(ipart);
+    // upper bound
+    if (ind >= ng)
+    {
+      ind = ng;
+      dx = T(1) - std::numeric_limits<T>::epsilon();
```
PDoakORNL: This is the behavior I think is strange.  Imagine ng=2 I pass x=2.5. the periodic function is sin(x/pi) but I get back 0. Now with the > there I get the value of at 0.5 but at 3.0 I suddenly get the value of 2.0.

ye-luo: ng is not the grid point. It is the upper bound. So > is correct, >= is not correct.

if there is n grid point. If in the 3D periodic spline case, 
getSplineBound(,,,ng=n-1)
in the 1D open BC spline case,
getSplineBound(,,,ng=n-2)

in your example, ng=2, x=2.5. [0,3) is the allowed range.
you get integer part 2 and fractional part 0.5
in your example, ng=2, x=3.0. [0,3) is the allowed range.
you get integer part 2 and fractional part 1-epsilon.

Does this make sense?

src/spline2/MultiBsplineData.hpp
ye-luo: spline is real only but can be float or double.
The compile should optimized it out. T is plain-old type and the compiler should optimize it out.

src/spline2/MultiBsplineData.hpp
PDoakORNL: Yes that makes sense but it must be documented is some clear way upper bound is not clear ng is the highest grid point of the implementation.  This seems like its a private static member function of some type class you're avoiding defining. The code could be more self documenting right now its a collection of functions from einspline.  Are we going somewhere like that?

ye-luo: We did enough with wrapping einspline with C++ in the past and spline2 is the templated C++ version and we are moving away from einspline. The getSplineBound is used by 1D and 3D cases and I made it a free template function which can be always inlined.

grahamlopez:
Test this please

PR #1374, opened 2019-02-11
Improve hdf_archive error handling and interface

prckent:
- Modify the hdf_archive::read and ::write functions to check the return HDF5 error codes internally and return void. Historically, the error codes were largely unchecked by the calling code which had the responsibility. With this PR, the majority of the HDF5 I/O calls are now error checked.
- Introduce readEntry and writeEntry functions that return bool error codes with identical functionality to the previous read and write functions. Suggestions for better names welcome.
- This update was motivated by a recent bug in the converters that an error check would have caught. 
- Thanks to the AFQMC folks for checking most of their HDF5 calls.


prckent:
(Removed WIP. Checked gcc8+MKL with AFQMC and Intel2018 by hand. No GPU checks yet.)

ye-luo:
I cannot build it with SoA. More readEntry change is needed.

prckent:
Thanks. Updated for SoA.

grahamlopez:
Test this please

ye-luo:
Does the unit test have a case checking the failed read?
If not, could you add one? Just readEntry someone existing dataset.

prckent:
Added. Good suggestion.

PR #1378, opened 2019-02-13
Refactor SPOSet::evaluateValues

ye-luo:
SPOSet::evaluateValues arguments are revised. It is used by WavefunctionComponent::evaluateRatios for the batched evaluation of WF ratios in NLPP.

Previously, I thought providing scratch space is necessary from caller class. After settling the batched algorithm and nested parallelism, I found thhe needed space is very small and can be just kept local with greater flexibility. The code is much cleaner and safer without SPOMem.

In addition to making a portable algorithm for both the CPU and GPU, I found it necessary to move the dot production inside the evaluateValues and return only ratios instead of SPO values. After this change, I feel the routine call be called better as evaluateDetRatios. I have not done the name change and expect suggestions for names.

Note: I have made the renaming.


prckent:
Since SPOSet is an important class and you have been staring at the interface, please could you add doxygen comments for *all* the functions?

ye-luo:
@prckent I added comments on all the functions for the CPU code. Macro protected AoS or CUDA interfaces are undocumented.

prckent:
@ye-luo Thanks. Very helpful to note the functions we should remove in future as well.

ye-luo:
It is now specialized to compute the determinant ratios.
DetInvMat[iat] needs to pass in as psiinv and the return value is ratios.

If a future need for evaluateValues occurs, we just add evaluateValues do exactly as the name describes. Currently we don't need evaluateValues and thus there is no reason to keep it.

ye-luo:
Sorry. Some renaming is incorrect. sed mistake.

prckent:
Regarding improving the interface, last week I twice received feedback that the large number of different evaluates (etc.) across the codebase makes it harder to follow. It certainly makes it difficult to search for specific functionality.

My suggestion for *future* updates of (say) SPOSet is to actually refer to SPOSet in many of the function names. e.g.  evaluateSPOSetValues, and variants.  If we can keep the interfaces to different abstract classes distinct this should improve accessibility across the whole application. Currently we can "evaluate" too many distinct things. Since the arguments vary by base class,  this does not help usability much. If they e.g. all took a single coordinate and did the same thing, we would keep them the same.

Better suggestions welcome.

ye-luo:
After going through all the interfaces in SPOSet, I feel all the evaluate() is a headache and we probably misused function overloading.
Names need to be short but self-explaining.
SPOSet::evaluateValue
SPOSet::evaluateValueGradientHessian
I don't think class name SPOSet is needed in the function name.
I also don't like plurals. But I don't know what is the best practice.

prckent:
I don't claim to have a great solution yet, but the problem reported is that we have

Apple::evaluateValue(fruit,tastiness);
Car::evaluateValue(road,direction,speed);

This makes the code much less accessible. e.g. grep evaluateValue is not informative. And we have a huge proliferation of evaluate* functions.

jtkrogel:
What about ditching evaluate all together and simply describing the return type:

SPOSet spo;
spo.orbitalValues(...);
spo.orbitalValGradHess(...);

WaveFunctionComponent psi;
psi.wavefunctionValue(...);

prckent:
@jtkrogel A more elegant and effective suggestion.

ye-luo:
I insist keeping the verb in the function. During our last discussion about naming rules, we define function names as verbObject. It needs to be evaluateABC, setDEF, getGHI.

'Orbital' is ambigioius can be molecular orbital or atomic orbital.
To distinguish the different evaluate Values between difference base classes,
I'm thinking of
SPOSet spo;
spo.evaluateSPOValues(...);
spo.evaluateSPOValGradHess(...);

This is more tricky. Do we need to distinguish the functions between
TrialWaveFunction and WaveFunctionComponent
WaveFunctionComponent psi;
psi.evaluateWaveFunctionValue(...); // fresh evaluation
psi.getWaveFunctionValue(...); // get the saved value if there is any.
...

prckent:
Anything that is less ambiguous and clearer would be an improvement.

(For functions that use fully recalculated vs cached data, this should be in the title. e.g. "FromScratch" vs "". Making the state machine clearer is definitely needed. )

ye-luo:
Test this please

PR #1380, opened 2019-02-14
Add deterministic test for solid 'diamondC_2x1x1_pp' and add scalars for deterministic test for molecule 'LiH_dimer_pp'.

Hyeondeok-Shin:
Add some deterministic tests for the full precision build. 

1. Add deterministic test for scalars 'ionion', 'localECP', 'nonlocalECP', and 'flux' on a molecule 'LiH_dimer_pp' testset for more detailed test. 

2. Add deterministic tests for a solid 'diamondC_2x1x1_pp' testset.

References were obtained from AoS/real/full precision build with clang-4.0 in BG/Q. Tests were passed with gcc 8.2.1 and intel 19.0.2.187 in my workstation, but it needs more feedbacks from other compilers and machines.

Hyeondeok Shin


ye-luo:
det_qmc_short_vmc_dmc_reconf.in.xml Is it necessary to keep the one for reconfiguration? It is not correct at the moment and may cause false alarm.

Hyeondeok-Shin:
Thanks for the comments. 'det_qmc_short_vmc_dmc_reconf.in.xml' is removed and the step has changed. Please let me know there is still a problem I have to update.

prckent:
LGTM

ye-luo:
Okay to test

PR #1384, opened 2019-02-15
Remove state from quintic spline

markdewing:
This is the first step in remove state from the spline evaluations.  There are two parts

1. Remove the location index (Loc) from OneDimGridBase.  The 'locate' function now returns the index rather than storing it.   Changing the rest of the code to use this might result in some duplicate calls to 'locate', but that should not have a big performance impact.

2. Completely remove state from OneDimQuinticSpline.   A new function (getIndexAndDistanceFromGrid) returns the location index and the distance - these are the only values computed by updateForQuintic, so moving that functionality to the quintic splint functions should not have a big performance impact.  This also removes a duplicate call to 'locate'.
Also mitigating any performance impact, quintic evaluations are only done here for the AoS code - the SoA code use the multi quintic evaluation in lcao/MultiQuiniticSpline1D.h.

Removal of the cubic spline state will be in a future PR.


prckent:
Mark - assuming the CI passes I think this could be approved as is. While massaging the class, could you make everything private apart from the appropriate functions?

prckent:
Will merge now since there will be a follow-up PR

markdewing:
Yes, I can also change the public/private access in a future PR.

PR #1385, opened 2019-02-15
Remove cubic spline state

markdewing:
The cubic spline state was extracted to a new class, CubicSplineEvaluator.
Ye said the grid manager optimization was not used, so removing it should not cause a performance problem.

All the evaluation routines were made `const` (including those from the quintic splines).
The internal quintic spline routines were made private.


prckent:
I'll approve since we are close to the weekend test window, but comments defining the remaining public functions would be appreciated in a later PR.

prckent:
Thanks for adding the private declarations.

PR #1387, opened 2019-02-16
NonLocalECPComponent Unit Test

rcclay:
This implements a test for NonLocalECPComponent.  If QMC_COMPLEX and ENABLE_SOA are enabled, Hellman-Feynman forces are checked as well.  Otherwise, only the value of the nonlocal channels on a compressed sodium dimer is checked.


markdewing:
okay to test

PR #1393, opened 2019-02-18
Improve nested threading

ye-luo:
This PR improves the performance and handling of nested threading.
1. Threading is enabled on Coulomb after PR #1385 made cublic spline thread-safe.
2. Introduce BlasThreadingEnv class to manage the threading of threaded BLAS/LAPACK libraries when being called from OpenMP region. This is a much better solution than using environment variables forcing the number of threads globally.
3. Implement nested threading in delayed update in case of serial BLAS.
4. Manual updated to explain the "Serial or multi-threaded library" in the installation and how to use nested threading.


src/Numerics/BlasNestedThreadingService.h
```diff
@@ -0,0 +1,47 @@
...
+
+  public:
+  /// Constructor, obtains the number of threads at the next level
+  BlasNestedThreadingService();
+
+  /// get num_threads
+  int getNumThreads() { return num_threads; }
+
+  /// Detect the threading environment and preset BLAS libraray
+  void presetBLASNumThreads();
```
markdewing: Would this be better if the name started with just 'set' rather than 'preset'? 
(In my mind, 'preset' indicates a setting at the very beginning of the run, not something that is set before a call to MKL)

src/QMCWaveFunctions/Fermion/DelayedUpdate.h
```diff
@@ -204,10 +206,51 @@ namespace qmcplusplus {
...
-          for(int i=0; i<delay_count; i++) tempMat(delay_list[i], i) -= cone;
-          BLAS::gemm('N', 'N', norb, delay_count, delay_count, cone, V.data(), norb, Binv.data(), lda_Binv, czero, U.data(), norb);
-          BLAS::gemm('N', 'N', norb, norb, delay_count, -cone, U.data(), norb, tempMat.data(), lda_Binv, cone, Ainv.data(), norb);
+          BlasNestedThreadingService knob;
+          // always use serial when norb is small or only one second level thread
+          bool use_serial(norb<=256||knob.getNumThreads()==1);
+          if(use_serial || knob.NestedThreadingSupported())
+          {
+            // threading depends on BLAS
+            if(!use_serial) knob.presetBLASNumThreads();
```
markdewing: This looks like a good use for scoping to set/unset the BLAS threading environment.  Maybe another class (BlasThreadingEnv ?) that does RAII and takes a flag to run in serial or not?

ye-luo: Good suggestion. Done.

markdewing: Thanks, it looks better now

src/Numerics/BlasNestedThreadingService.h
PDoakORNL: This should just be set.  It will then form a logical pair with unset.

config/BGQ_Clang_ToolChain.cmake
```diff
@@ -1,6 +1,9 @@
 set(CMAKE_C_COMPILER mpiclang) 
 set(CMAKE_CXX_COMPILER mpiclang++11)
 
+# set the linker if extra wrapper is needed for profiling tools like hpctoolkit.
```
PDoakORNL: If we're going to support hpc toolkit it shouldn't be through comments in certain config's. If its just a one time thing put in the manual or wiki and pass it as a -DCMAKE_CXX_LINKER=...

manual/installation.tex
```diff
@@ -485,6 +485,19 @@ \subsubsection{Using Intel MKL with non-Intel compilers}
...
 directories (etc.) and is often /opt/intel/mkl
 
+\subsubsection{Serial or multi-threaded library}
+\label{sec:threadedlibrary}
+Vendors may provide both serial and multi-threaded versions of their libraries.
+Using the right version is critical to QMCPACK performance.
+QMCPACK makes calls from both inside and outside threaded regions.
+When being called from outside an OpenMP parallel region, the multi-threaded version is preferred for the possibility of using all the avaiable cores.
+When being called from every thread inside an OpenMP parallel region, the serial version is preferred for not oversubscribing the cores.
+Fortunately nowadays the multi-threaded versions of many vendor libraries (MKL, ESSL) are OpenMP aware.
```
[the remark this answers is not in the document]
ye-luo: Since I started to use MKL and ESSL 10 years ago, it is already like this. I don't know which version.

src/Numerics/BlasNestedThreadingService.h
ye-luo: All gone as BlasThreadingEnv is introduced.

config/BGQ_Clang_ToolChain.cmake
ye-luo: Unnecessary. Removed.

manual/installation.tex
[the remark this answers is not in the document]
PDoakORNL: Ok we'll just have to fill this in as we learn more.

PR #1395, opened 2019-02-19
Add Calculation of Pulay Terms to NonLocalECPComponent

rcclay:
This PR adds a function to NonLocalECPComponent to spit out the contributions to the Nonlocal pseudopotential force coming from ionic derivatives of $Psi_T$.  This also adds a unit test of this feature.

### DIFF
--- a/src/QMCHamiltonians/NonLocalECPComponent.cpp
+++ b/src/QMCHamiltonians/NonLocalECPComponent.cpp
@@ -375,6 +375,192 @@ NonLocalECPComponent::evaluateOneWithForces(ParticleSet& W, int iat, TrialWaveFu
   return pairpot;
 }
 
+NonLocalECPComponent::RealType
+NonLocalECPComponent::evaluateOneWithForces(ParticleSet& W, ParticleSet& ions, int iat, TrialWaveFunction& psi, 
+    int iel, RealType r, const PosType& dr, 
+    PosType & force_iat, ParticleSet::ParticlePos_t & pulay_terms, bool Tmove, std::vector<NonLocalData>& Txy) const
+{
+  constexpr RealType czero(0);
+  constexpr RealType cone(1);
+
+  //Array for P_l[cos(theta)].
+  RealType lpol_[lmax+1];
+  //Array for P'_l[cos(theta)]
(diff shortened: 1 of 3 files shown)


markdewing:
okay to test

ye-luo:
1. evaluateOneWithForces(HF+Pulay)
duplicates
2. evaluateOneWithForces(HF)
duplicates
3. evaluateOne
I'm feeling a design issue.
Long term Foces should be seperated and move to an estimator.
For the moment, I'm trying to figure out why 1,2 duplication is needed.
Why the pulay term needs to be seperated from the HF contribution in force?

rcclay:
evaluateOneWithForces(HF) will probably be deprecated/removed, once I understand the behavior and accuracy of the estimators.  At the moment, computing the pulay terms could add quite a bit of computational expense over PP+HF evaluation, so leaving the option to not evaluate them is nice.  Also, the HF terms here don't have an infinite variance issue out of the box, whereas the Pulay terms would.

prckent:
Please add a function definition (doxygen comment). evaluateOne what? A better name is needed here.

markdewing:
Variables and parameters reuse PosType and ParticleSet::ParticlePos_t for their type.    It might be clearer if some additional types were added specifically for forces - even a simple typedef of ForceType for PosType and ForceVector_t (or some better name?) for ParticleSet::ParticlePos_t would be enough.   This is something more pervasive than just this PR - if changed, it should be in a separate PR.

rcclay:
@markdewing That's a reasonable suggestion.  I can definitely do that in a future PR.

rcclay:
@ye-luo Current design is to go into the physical hamiltonian components and implement the analytic derivatives.  A "force estimator" will trigger computation of these various derivatives, and then collect/assemble them.  Treating it so that every observable that contributes to $E_L$ will have a $\partial(E_L)$ contribution.

ye-luo:
@rcclay I agree with you, each physical hamiltonian component to $E_l$ has an interface for ionic derivatives. Then the force estimator can call theses interfaces and then assemble the force. The universal estimator manager handles all the data collection and implements proper back propagation.

PR #1396, opened 2019-02-20
Nexus: update and expand Gaussian Process code

jtkrogel:
This PR updates the gaussian_process module with version 2 of Rick Archibald's Gaussian Process optimization code.  The procedural code is wrapped in a new class-based interface to enable easier use and carefully controlled saving/loading of the GP optimization state each iteration.

This interface will be used to integrate Nexus-based DFT and QMC workflows with Gaussian Process optimization to target structural optimization of systems with less than about 20 free parameters.

A test function interface is also included to test drive the optimizer on known optimization target functions.


markdewing:
okay to test

PR #1397, opened 2019-02-20
Add Local Kinetic Energy "Pulay Term" Contribution

rcclay:
This adds a function for computing the ionic gradient of the local potential energy (BareKineticEnergy).  A unit test is added which tests the value, HF term (there's none), and "Pulay" term.  These reference values are constructed by explicit finite differences of the local kinetic energy in the wave function tester.

### DIFF
--- a/src/QMCHamiltonians/BareKineticEnergy.h
+++ b/src/QMCHamiltonians/BareKineticEnergy.h
@@ -45,6 +45,21 @@ inline T laplacian(const TinyVector<std::complex<T>,D>& g, const std::complex<T>
   return  l.real()+OTCDot<T,T,D>::apply(g,g);
 }
 
+template<typename T, unsigned D>
+inline T dlaplacian(const TinyVector<T,D>& g, const T l, const TinyVector<T,D>& gg, const T gl, const T ideriv)
+{
+  return gl + l*ideriv + 2.0*dot(g,gg) + dot(g,g)*ideriv;
+}
+template<typename T, unsigned D>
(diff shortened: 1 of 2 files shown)


src/QMCHamiltonians/BareKineticEnergy.h
```diff
@@ -20,6 +20,7 @@
 #include "Particle/WalkerSetRef.h"
 #include "QMCHamiltonians/QMCHamiltonianBase.h"
 #include "ParticleBase/ParticleAttribOps.h"
+#include "QMCWaveFunctions/TrialWaveFunction.h"
```
ye-luo: Is this line really needed?

ye-luo:
Okay to test

src/QMCHamiltonians/BareKineticEnergy.h
rcclay: Actually, I don't think it is.

rcclay: Rhea seems to think it is necessary.  Adding it back in.

ye-luo: The major drawback of including TrialWaveFunction.h is that it will be propagated.
QMCHamiltonianBase.h intentionally use "class TrialWaveFunction;" to avoid the header inclusion and the derived class cpp includes TrialWaveFunction.h. In your case, the class is a template and there is no cpp. So you have to include the header file.
I believe BareKineticEnergy.h itself will not be propagated too far. So the inclusion is fine.

PR #1399, opened 2019-02-22
Enable LCAO in complex build

ye-luo:
This PR enables complex LCAO infrastructure.
The design is to support complex MO using complex basis set (complex AO with real radial functions).
The actually functionality will be completed by @anbenali 

Significant changes are
1. Separate functions for cusp correction from LCAOrbitalBuilder to its own header and cpp files.
2. Rename RealBasisSetBase to SoaBasisSetBase which now serves both real and complex.
3. Orbital optimization and cusp currections are not enabled in the complex build.

### DIFF
--- a/src/QMCWaveFunctions/lcao/CuspCorrectionConstruction.cpp
+++ b/src/QMCWaveFunctions/lcao/CuspCorrectionConstruction.cpp
@@ -1,0 +1,256 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2016 Jeongnim Kim and QMCPACK developers.
+//
+// File developed by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
+//                    Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//
+// File created by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+#include "QMCWaveFunctions/lcao/CuspCorrectionConstruction.h"
+
+namespace qmcplusplus
+{
+
+  // Modifies orbital set lcwc
+  void applyCuspCorrection(const Matrix<CuspCorrectionParameters> &info, int num_centers,
+                            int orbital_set_size, ParticleSet& targetPtcl, ParticleSet& sourcePtcl,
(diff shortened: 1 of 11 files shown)


anbenali:
Yessss!!!!

prckent:
Will complexifying the cusp correction implementation require more than a scattering of T and "auto" throughout, to replace use of realtype etc., or are there greater issues?

anbenali:
I am not sure of the meaning of cusp correction with complex wavefuction..
is it implying doing the real part and imaginary part independently? Or is
there something more? It feels more like a research topic than simple
implementation.. at this point requiring cusp correction for complex lcao
means we plan to do finite size corrections for all electrons solids.. not
quiet a priority at this point...

jtkrogel:
The derivative of the cusp cancels the divergence in the Coulomb potential, which is real.  I think the cusp must therefore apply to the modulus of the wavefunction, but not the phase.

jtkrogel:
It should apply to real and imaginary parts equally, in a fashion that is equivalent to multiplying each orbital by it's own real valued, short ranged one body Jastrow function (same function for real and imaginary parts).  This is the only way not to affect the phase.

src/QMCWaveFunctions/lcao/CuspCorrectionConstruction.cpp
```diff
@@ -0,0 +1,256 @@
...
+    eta.setOrbitalSetSize(lcwc.OrbitalSetSize);
+    eta.BasisSetSize = lcwc.BasisSetSize;
+    eta.setIdentity(false);
+
+
+    std::vector<bool> corrCenter(num_centers, "true");
+
+    //What's this grid's lifespan?  Why on the heap?
+    LogGrid<RealType>* radial_grid = new LogGrid<RealType>;
+    radial_grid->set(0.000001, 100.0, 1001);
```
prckent: Where do these magic numbers come from?

prckent: (Wondering if we should start flagging magic numbers consistently in the source. For sure we will be bitten one day.)

ye-luo: I created an issue #1400 for discussion or trouble shooting magic number issues. So we don't need to delay the merge of this PR.

ye-luo:
Regarding complex cusp correction, I have not thought about how to do it since I have not read the method. It is better to leave it unfixed at the moment and fix it when the LCAO complex is working and getting tested.

PR #1401, opened 2019-02-22
QMCA save averaged data to scalar files

camelto2:
This enables qmca to save the averaged data from the -a option to scalar files for
each series. This will allow one to use qmc-fit to perform a time-step
extrapolation on twist averaged data, for example. One only needs to add the option --save_average, and it will store avg.sXXX.scalar.dat files for each series. Then use qmc-fit on the new scalar files to extrapolate your twisted averaged data.

### DIFF
--- a/nexus/bin/qmca
+++ b/nexus/bin/qmca
@@ -2685,7 +2685,6 @@ QMCA examples:
         opt = self.options
         if opt.average and opt.save_average:
             for fmap in self.file_map:
-                print(fmap.data.first())
                 ndata = len(fmap.data.first())
                 data = [arange(ndata)]
                 keys = 'index'
--- a/nexus/bin/qmca
+++ b/nexus/bin/qmca
@@ -2685,7 +2685,8 @@ QMCA examples:
         opt = self.options
         if opt.average and opt.save_average:
             for fmap in self.file_map:
-                ndata = len(fmap.data['LocalEnergy'])
+                print(fmap.data.first())
+                ndata = len(fmap.data.first())
                 data = [arange(ndata)]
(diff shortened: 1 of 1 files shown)


markdewing:
okay to test

nexus/bin/qmca
```diff
@@ -2677,6 +2683,18 @@ QMCA examples:
             #end for
         #end for
         opt = self.options
+        if opt.average and opt.save_average:
+            for fmap in self.file_map:
+                ndata = len(fmap.data['LocalEnergy'])
```
jtkrogel: Use fmap.data.first() instead.  Depending on the method, LocalEnergy may or may not be the key used for the total energy.

nexus/bin/qmca
```diff
@@ -2677,6 +2683,18 @@ QMCA examples:
...
+                data = [arange(ndata)]
+                keys = 'index'
+                fmt = ['%i']
+                for key in fmap.data.keys():
+                    data.append(fmap.data[key])
+                    keys+='     {}'.format(key)
+                    fmt.append('%.10e')
```
jtkrogel: Format control expressions to match surrounding code, i.e. use "#end if" and "#end for".

jtkrogel:
@camelto2 just a couple of small changes, see above.

ye-luo:
Test this please

ye-luo:
@jtkrogel Good to merge?

jtkrogel:
Yes.

ye-luo:
Test this please

ye-luo:
Test this please.

PR #1402, opened 2019-02-25
Clang format stage 1

PDoakORNL:
First stage of clang-format of code base.


src/Lattice/Uniform2DGridLayout.cpp
```diff
@@ -46,54 +42,54 @@ void Uniform2DGridLayout::makeShell(std::vector<SingleParticleIndex_t>& RS,
...
     }
   }
   RS.resize(ic);
   FS.resize(ic);
-  int irc = static_cast<int>(4.0*rr*scaleL);
-  int ir=0;
-  std::map<int,std::vector<SingleParticleIndex_t>*>::const_iterator cit = rs.begin();
-  while(cit != rs.end())
+  int irc                                                                = static_cast<int>(4.0 * rr * scaleL);
+  int ir                                                                 = 0;
```
jtkrogel: Alignment here is a bit extreme.  Perhaps not a big deal if infrequent.

ye-luo: It is due to the next line with a = very far away.
Currently clang-format is not smart enough to handle this.
Hopefully they will introduce some heuristics or machine-learning algorithm to improve this.

PDoakORNL: adding a blank line fixes this.  For now clang-format doesn't provide a penalty weighting for this particular reformat so this is what we can do.

PR #1403, opened 2019-02-25
Exclude -ffast-math when building ppconvert

ye-luo:
When -ffast-math is used with GNU and Clang, some values become NaN and cause infinite loops in ppconvert. This PR only workaround the problem. It **does not** fix the cause of the NaN.


lshulen:
This looks fine to me.  What is the origin of the NaN's in the first place?  Is this an issue with PBE having subtleties when the value of the wavefunction gets small?

ye-luo:
@lshulen I have no idea. Etrial goes NaN. I added a trap at very top level.
@lshulen Since you've already reviewed the PR and think it is good to go. Please use the review tool to approve the PR. We need more people involved in the review process.

PR #1404, opened 2019-02-26
Fix TM1/3 when using both batched algorithm and delayed updates

ye-luo:
1. Fix Tmove v1 and v3 when using both batched algorithm and delayed updates
2. Move the state machine invRow_id to the DiracDeterminat class
3. ratio, ratioGrad, evalGrad inside DelayedUpdate class is not necessary and moved back to DiracDeterminat.

### DIFF
--- a/src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
+++ b/src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
@@ -32,7 +32,7 @@ namespace qmcplusplus
  *@param first index of the first particle
  */
 DiracDeterminant::DiracDeterminant(SPOSetPtr const spos, int first):
-  DiracDeterminantBase(spos,first), ndelay(1)
+  DiracDeterminantBase(spos,first), ndelay(1), invRow_id(-1)
 {
   ClassName = "DiracDeterminant";
 }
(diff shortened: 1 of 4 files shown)


src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
```diff
@@ -131,9 +134,16 @@ DiracDeterminant::ratioGrad(ParticleSet& P, int iat, GradType& grad_iat)
   const int WorkingIndex = iat-FirstIndex;
   UpdateMode=ORB_PBYP_PARTIAL;
   GradType rv;
-  curRatio = updateEng.ratioGrad(psiM, WorkingIndex, psiV, dpsiV, rv);
-  grad_iat += ((RealType)1.0/curRatio) * rv;
-  RatioTimer.stop();
+
+  // check invRow_id against WorkingIndex to ensure getInvRow() called before ratioGrad()
+  // This is not a safety check. Some code paths do call ratioGrad without calling evalGrad first.
```
prckent: Commentary: Unfortunate that this is needed, but how would anyone ever know to call evalGrad before calling ratioGrad?

ye-luo: Good point. I have update the comment which is more clear as an optimization.

PR #1405, opened 2019-02-26
Create preliminary ZVZB Force Estimator

rcclay:
This PR puts in necessary framework for:
1.)  An estimator to evaluate and accumulate all correct ionic derivatives to assemble a ZVZB force estimator.
2.)  Consistent interface in QMCHamiltonianBase and derived objects for evaluating ionic derivatives. 

Quite a lot of stuff has been touched, but before I go any farther, I figure I'll submit this as a digestible PR.  I'm holding off on a unit test until I improve the long-range handler, mainly because the finite difference results for the LocalECP force converge much more quickly than the ewald3D direct evaluation, and this is something I've been meaning to do anyways.  The plus side is that the difference between the analytic derivative of E_L and its finite difference I believe is attributable to this slow convergence, and so this estimator should be codewise and formally correct. 

By the way, expect a pretty large refactoring of the legacy force code once this is ready for production.  For now, things are going to get a little messier before they get cleaner.


markdewing:
okay to test

ye-luo:
Good to discuss here.
I saw function names evaluateWithIonDerivs and evaluateIonDerivs at QMCHamiltonian and QMCHamiltonianBase. The role is not very clear.

Could you explain a bit more the call logic?
In the driver, we call the following now
1. H.evaluateWithToperator for the local energy
2. compute the new walker weight based on the local energy.
3. H.auxHevaluate compute all the estimators.

For the force, are you intended to
1. H.evaluateWithToperator for the local energy + ion derivatives if needed by the estimator
2. compute the new walker weight based on the local energy.
3. H.auxHevaluate compute all the estimators. The force estimator assembles the saved pieces from 1 without computing the derivatives.

This solution has the minimal computational duplication but the QMCHamiltonianBase::evaluate needs to include force terms in the arguments. A switch is also needed to control the computation of forces when the frequency of estimator computation is less frequent.

Another way is
1. H.evaluateWithToperator for the local energy.
2. compute the new walker weight based on the local energy.
3. H.auxHevaluate compute all the estimators. The force estimator calls H to evaluate the ion derivatives only and then assemble the force.

On the positive side, there is less state machine needed and we don't need to grow the number of arguments in the energy evaluation API.
The negative side, there may be a bit computation redundant but it can be small if you compare the cost to the force.

Also think about if you need to add stress, how would you prefer the API to be?

rcclay:
No problem!  
So the way the current code works is if ComputeForces=True, evaluate(P) for observables drops into an appropriate evaluate() call that calculates the forces.  These are stored in ParticlePos_t force arrays derived from ForceBase class.  Each estimator dumps its value and forces directly to the scalar.dat and hdf5 files.  

The way I propose, and where this PR is going, is the last possibility you suggested.  When H.auxHevaluate is called, the combined force estimator calls the H.evaluateIonDerivs, which returns the value, hellman-feynman like d/dR, Zero-Variance d/dR component (which I've been calling Pulay, might change the name of this in a future refactor), and d/dR wavefunction only components, which are use to assemble the estimator.  All H.evaluateIonDerivs does is cycle through the QMCHamiltonianBase objects and call their evaluateWithIonDerivs functions, each appending their contribution to the passed in hf_force and pulay_term force arrays.  

On the plus side, I think this code approach should allow me to eliminate ForceBase and a lot of the state machine stuff in the various estimators.  You can see in this PR that I've spun off the force evaluation into their own functions (evaluateWithIonDerivs).  I will need to add something like QMCHamiltonian::initializeIonDerivs(...) and TrialWaveFunction::initializeIonDerivs(), (likewise for QMCHamiltonianBase and WaveFunctionComponent) to handle initialization.  The main state information I'm thinking about adding is based on SPO ionic derivatives: if there's no explicit dependence on ion coordinates, I really want DiracDeterminant to skip all the matrix algebra.  
 
My preferred API is to continue using this template, where an evaluateXDeriv[s] is added to QMCHamiltonian to iterate over the QMCHamiltonianBase objects, each calling their evaluateXDeriv[s] implementations.  Part of me feels like there should be a more elegant solution to this that is less verbose, but this is at least transparent.

ye-luo:
1. Why do you need initializeIonDerivs? Just resize the memory at H/WF.evaluateIonDerivs call.
2. The assumption of no ionic derivative in SPO is not OK. LCAO will contain ionic derivatives. The API must be made generic.
3. "I really want DiracDeterminant to skip all the matrix algebra." What do you mean by "skip". We can query the ion derivatives of SPOSet and then skip the matrix operation if that is possible.
4. evaluateWithIonDerivs do you need the local energy term or just the ionic derivatives when computing the forces? I guess putting "With" in the name means you will compute the local energy contribution as well, right? In the TrialWavefunction each function has a corresponding one with the same name at WaveFunctionComponent. The same here for QMCHamiltonian and QMCHamiltonianBase. Why do you need to use different names?

ye-luo:
There is another problem in general. The dependence chain
ParticleSet
TrialWavefunction depends on ParticleSet
QMCHamiltonian depends on ParticleSet and TrialWavefunction
Estimator can depends all the above.
Now the code put estimators as hamiltonians and the QMCHamiltonian is playing two roles and got messy. This needs to be resolved by a better design in the near future.

rcclay:
Just going point by point:
1.)  You mean initialize the temporary arrays and what not at the beginning of H/WF.evaluateIonDerivatives?  I think thats OK.  The most memory intensive stuff will be in the DiracDeterminant object, but compared to the amount of work that needs to be done, creating temporary arrays shouldn't be bad. 
2 & 3.) I'm definitely not assuming that SPO's don't depend on ionic coordinates.  What I want though is a flag in SPOSet (indicating if it depends explicitly on ionic coordinates or not) that DiracDeterminant can grab, so that DiracDeterminant knows if it can skip the matrix algebra or not.  
4.) I'll elaborate on this point in a separate post.  As for why I used different names, I kinda wanted different names for QMCHamiltonian and QMCHamiltonianBase, but I can keep the names the same.  Naming is an issue, and I'm open to suggestions.  

Can you clarify by what you mean by "Now the code put estimators as hamiltonians and the QMCHamiltonian is playing two roles and got messy."

ye-luo:
1). The arrays are not not temporal and belong to the class. They are resized only within the corresponding evaluation functions. The price of allocation is only paid once and only when the feature is used.
2&3). Yes. This is what I'd like to have. Not the way currently in DiracDeterimant class, the IonDerivs flag is ugly.

QMCHamiltonian owns H which corresponds to physical hamiltonian terms contributing to the local energy and auxH corresponds to estimators. They are basically two separated code paths interfaces though one class. The assumption of auxH for estimators and H for local energies is not the same.
Force(an element of auxH) depends on H. This is not directly related to this PR but if you have any thoughts about it. Please mentioned them.

rcclay:
In the long run, it might be worth considering forking off auxH into its own class...  like "Operators", "Observables" or something.  Most ZVZB style estimators I've seen (mainly forces, but also pressure/stress, force constant matrix, what not) are mainly derivatives of the local energy, which itself intrinsically depends on physical hamiltonian, wave function, and particle position.  This funky dependency I have sort of tries to recognize this, but I'm open to suggestions.  Here's what I would like out of a future code design:

1.)  Easy access to the local energy and all its derivatives. Having the various derivatives of local energy live in QMCHamiltonian and QMCHamiltonianBase seems logical.  Furthermore, adding new derivatives just involves implementing functions, and doesn't add data structures to the underlying QMCHamiltonianBase object, with the goal being to populate them and wait for an estimator to scrape their values.    Previously, there was a lot of state information associated with initializing ForceBase and various code paths in QMCHamiltonianBase::evaluate() just to allow computation and scraping of these values.  Doing it this way though, I think I can eliminate ForceBase and a lot of state info.

2.)  Ability to easily determine which combination of derivatives of the local energy I want to evaluate and how often.   The existence of an estimator for the derivative sort of takes care of this, and doesn't depend on the details of what physical Hamiltonian the user specified.

rcclay:
As an aside, here's what I'm trying to compute:
-F_{ZVZB} = d(E_L) + 2(E_L - <E>) dPsi/Psi
This is equivalent to the standard Assaraf-Caffarel form:
-F_ZVZB = [(dH)Psi]/Psi + [(H-E_L)dPsi]/Psi + 2(E_L-<E>) dPsi/Psi.

Currently, the first term and second term are spit out as the "Hellman-Feynman" and "pulay terms" respectively in things like evaluateWithIonDerivs, although really the second term is the "zero-variance" term.  The E_L contribution is also returned from evaluateWithIonDerivs.  With E_L and dPsi/Psi computed in the force estimator, this gives me everything I need to compute the second term.  When everything is validated, I intend to merge the Hellman-Feynman and ZV terms and work directly with d(E_L), E_L*dPsi/Psi, and dPsi/Psi.  I'm keeping them separate for now because of infinite variance issues, that I would like to verify are solved once I implement a guiding function.

ye-luo:
Thank you for the explanation in detail. Now I remember E_L is obviously needed when computing any kind of derivatives. In terms of how to handle arguments, this is research related. You have all the freedom to change and it may change over time once you try more things like the stress.

src/QMCHamiltonians/QMCHamiltonian.h
```diff
@@ -266,6 +266,10 @@ class QMCHamiltonian
       std::vector<RealType>& dhpsioverpsi,
       bool compute_deriv);
 
+  Return_t evaluateIonDerivs(ParticleSet& P, ParticleSet& ions, TrialWaveFunction& psi,
```
ye-luo: Please add documentation for this function

rcclay: Done

PR #1406, opened 2019-02-28
Device implementation

fdmalone:
GPU implementation for the k-point version of the AFQMC code.


prckent:
What are the minimum CUDA and GPU versions for this to work? E.g. Can this run on a Kepler, even if slowly?

ye-luo:
Okay to test

mmorale3:
I tested the code in Volta and Pascal. In principle, the real limitation is the need for at least gcc6 and/or clang4. It should work with sm3.5 from the cuda point of view.

CMake/GNUCompilers.cmake
```diff
@@ -43,8 +43,8 @@ if(CMAKE_CXX_FLAGS MATCHES "-march=" OR CMAKE_C_FLAGS MATCHES "-march=")
   endif() #(CMAKE_CXX_FLAGS MATCHES "-march=" AND CMAKE_C_FLAGS MATCHES "-march=")
 else() #(CMAKE_CXX_FLAGS MATCHES "-march=" OR CMAKE_C_FLAGS MATCHES "-march=")
   # use -march=native
-  SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -march=native")
-  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native")
+#  SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -march=native")
+#  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native")
```
ye-luo: Could you explain why you need to comment out this?

mmorale3:
Ups, this is a mistake. I couldn’t compile in my laptop without commenting this and the change propagated into the code in llnl. I just forgot to restore it.

prckent:
Version dependencies: Good. This will make it easy to test nightly/weekly on oxygen.

mmorale3:
Paul,

Don’t enable afqmc yet on nightly tests with gpus, we still need to clean up the unit tests on the gpu code. Right now all of them are being run, even those meant for the cpu only. Fionn is cleaning this up and will fix it in his next pr.

PR #1409, opened 2019-03-01
Add CoulombAA SoA Force path.  CoulombAB bug fix.

rcclay:
As the title says:  this adds a code path so that AA particle sets (read ions) can have forces computed correctly within the SoA build.  Moreover, there was a forgotten term in the expression for the CoulombPBCAB force, which has been added.   Ion-Ion force validated against quantum espresso, and electron-ion force separately validated.

### DIFF
--- a/src/QMCHamiltonians/CoulombPBCAA.cpp
+++ b/src/QMCHamiltonians/CoulombPBCAA.cpp
@@ -329,22 +329,69 @@ CoulombPBCAA::evalSRwithForces(ParticleSet& P)
 {
   const DistanceTableData &d_aa(*P.DistTables[0]);
   mRealType SR=0.0;
-  for(int ipart=0; ipart<NumCenters; ipart++)
+  if(d_aa.DTType == DT_SOA)
   {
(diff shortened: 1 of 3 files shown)


markdewing:
okay to test

src/QMCHamiltonians/CoulombPBCAA.cpp
```diff
@@ -36,11 +36,12 @@ CoulombPBCAA::CoulombPBCAA(ParticleSet& ref, bool active,
   DistanceTableData *d_aa = DistanceTable::add(ref,DT_SOA_PREFERRED);
   PtclRefName=d_aa->Name;
   initBreakup(ref);
-//  if(!is_active)
-//  {
+  ref.turnOnPerParticleSK();
```
ye-luo: PerParticleSK is not cheap.
Can it be protected by the ComputeForces flag?

rcclay: It can!  I'll do that :)

Issue #1412, opened 2019-03-02
Long Range Handlers Need Refactoring


rcclay:
I think the way the long range handlers deal with different breakups, (i.e. for energy, force, and stress), is a mess (I'm to blame for this :P).  Basically, a single handler (lets take LRHandlerTemp) for example, is designed to do the breakups and simultaneously manage breakups for the energy, force, and stress.  Moreover, if I want to do some but not all of these breakups, I have to add more functions to independently initialize and populate the data structures required to do the breakup.  

I propose to:
1.)  Split LRBreakup into 3 derived classes, one for energy, one for force, and one for stress.  Basically, DoBreakup will be different for all of these, but otherwise, the arguments and returns will be the same.  

2.)  Template LRHandlerBase and derived classes on the breakup scheme.  There will thus be only one set of fit coefficients, and one set of long-range fourier components.  

3.)  Eliminate the data structures associated with every other breakup scheme, and eliminate a lot of evaluate/initialize calls to other breakup schemes.   

Temporary solution:  
LRHandlerSRCoulomb will be the force-only breakup handler until the refactor is done.

ye-luo:
The idea of base class is to use virtual functions. virtual function cannot be templates functions.
The derived class is supposed to be template functions. Currently, LRHandlerTemp and LRHandlerSRCoulomb have a template parameter of the breakup basis which implicitly refer to the break up method used.
However, I only found LRHandlerTemp using LPQHIBasis and LRHandlerSRCoulomb using LPQHISRCoulombBasis.
Q1. Do both really need to put as a template argument?
Q2. In your re-factor scheme, do you plan to put LRHandlerBase* energyhander, forcehandler and stress hander. Each one can be individually initialized with one kind of LRBreakup? or what is your intention?

I do agree using one breakup scheme class per breakup basis instead of putting one Breakup class for all breakup basis. On the other hand, I'm not sure you need to set three derived classes but probably just three classes since they share little. data structure is different, call interfaces are different...

rcclay:
The breakup basis is less important here than the breakup equations that need to be solved to get the breakup coefficients.  Basically, the equations you need to solve for energy, force, and stress differ by k-dependent weighting factors.  This is why LRBreakup.h has DoBreakup, DoGradBreakup, etc.  As it stands, LRHandlerTemp and LRHandlerSRCoulomb solve the same equations, they just populate the input arrays slightly differently. It would be nice to choose which of these DoXBreakup calls gets done in a clean manner.

PR #1413, opened 2019-03-02
Long range handler bug fixes.

rcclay:
This PR does two things:

1.)  Performs some preparatory bug fixes and implementation to allow LRHandlerTemp and LRHandlerSRCoulomb to work with forces.  Main ones are implementing the evaluateLR and lrDf functions in LRHandlerSRCoulomb, and implementing/referencing the correct derivative of the LPQHI basis function for LRHandlerTemp.  Both can now be used to compute forces, but having this full capability for both won't happen until a refactor. 

2.)  Makes LRHandlerSRCoulomb the default force handler, and it only does forces.  This is a temporary state of affairs until a refactor can be done, as referenced in issue #1412.  Why?  Because initialization of LRHandlerTemp to do both types of breakups without doubling the startup overhead for everyone else's runs will take some hack-ish coding, and I think this is best handled with a refactor.


markdewing:
okay to test

prckent:
Ye, Mark - any concerns?

PR #1415, opened 2019-03-04
Next stage of clang-format Lattice,Particle,io,type_traits,QMCWaveFunctions

PDoakORNL:
picking up the format again.


src/Particle/Walker.h
```diff
@@ -159,93 +160,90 @@ struct Walker
...
-  {
-    return Rhok_GPU.data();
-  } GPU_XRAY_TRACE
-  inline CUDA_PRECISION_FULL* get_rhok_ptr (int isp)
-  {
-    return Rhok_GPU.data() + k_species_stride * isp;
-  } GPU_XRAY_TRACE
+      Rhok_GPU.resize(num_species * k_species_stride);
+  }
+  GPU_XRAY_TRACE
```
markdewing: It looks like GPU_XRAY_TRACE is placed at the end of the function definition?
If so, it's clearer in the original formatting.   Blank lines separating the functions would also work to make it clear which function GPU_XRAY_TRACE modifies (assuming the formatter would preserve the blank lines)

src/Particle/FastParticleOperators.h
```diff
@@ -199,149 +184,150 @@ struct ConvertPosUnit<ParticleAttrib<TinyVector<T,3> >,Tensor<T,3>, 3, false>
...
-    else if(Y>=PLUSONE) Y-=PLUSONE;                 \
-    if(Z<EPS)           Z+=PLUSONE;                 \
-    else if(Z>=PLUSONE) Z-=PLUSONE
-
-#define TWO_DIM_BOUNDARY_BLOCK(X,Y,EPS,PLUSONE) \
-    if(X<EPS)           X+=PLUSONE;                 \
-    else if(X>=PLUSONE) X-=PLUSONE;                 \
-    if(Y<EPS)           Y+=PLUSONE;                 \
-    else if(Y>=PLUSONE) Y-=PLUSONE
+#define THREE_DIM_BOUNDARY_BLOCK(X, Y, Z, EPS, PLUSONE) \
```
markdewing: I prefer the original formatting on this macro - putting the conditional and the expression on the same line makes it compact, and easier to see regularity of +=/-= operators
Any other opinions?  Is worth it to put the format exception markers around these macros?

PDoakORNL: I think its just time to remove that macro.  However its heaviest presence is in QMCWaveFunctions. I'll look at adding them to this PR.

PDoakORNL:
Also GPU_XRAY_TRACE has been removed.

ye-luo:
Need to enable C++11 for nvcc.

PR #1416, opened 2019-03-04
Allow Pseudopotential Forces in Real Build

rcclay:
This PR does the following to allow force computation in the real build (targeting openBC's):
1.)  Fixes a known bug in the nonlocalPP evaluation with real build.  I've removed the QMC_COMPLEX=1 guard in the test_ecp unit test, and everything should pass.
2.)  Implements Hellman-Feynman force evaluation for LocalECPotential (SoA only).  Basically a copy-paste job from CoulombPBCAB.cpp.


markdewing:
okay to test

PR #1417, opened 2019-03-05
Check compiler and libstdc++ compatibility

### LINKED ISSUE #1341: Detect and protect C++ standard library version for non GNU compilers
It happened a few times that people try to use Intel with the most recent gnu standard library.
For example, Intel 18 with gnu 7 or 8.
Due to the fact that Intel 18 supports C++14 not C++17 while newer standard library uses newer language features, the compilation aborts with error.
We need to find a way to catch this scenario in CMake. Currently we only check the lower bound.

ye-luo:
closes #1341 
The current scheme is compiling a toy code with version checks.
Intel 18.0 with gnu 7.1 passes the check and QMCPACK gets compiled.
Intel 18.3 with gnu 8.2 failed the check as intended.
`    #error too new libstdc++ from GNU for Intel 18, use GNU version <= 7`
Intel 19.2 with gnu 8.2 passes the check and QMCPACK gets compiled.
It seems that I can't reproduce some claims about gnu + intel 18 issue.

This check can be expanded for other compilers like IBM, Clang, PGI.


ye-luo:
> Please update the messages:
> "libstdc++ version check" -> "libstdc++/C++ compiler version compatibility check"

Updated

PR #1418, opened 2019-03-05
squeue correction for nexus

camelto2:
This PR fixes a small issue with squeue usage on the NERSC machines.
Evidently, the NERSC machines used to print squeue information in a
different order, but do so no longer. This fixes the reading of the
squeue output to gather the correct information. In the case that the
squeue REASON is longer than a single string (as is often the case on the NERSC
machines), now we only read the first 8 variables since the reason string is
never used anyway.

Additionally, I added the remaining squeue job states, because the NERSC
machines often have jobs in this various job states causing nexus to
fail.

@jtkrogel


markdewing:
okay to test

nexus/lib/machines.py
```diff
@@ -1291,7 +1291,21 @@ def __init__(self,
                                  R = 'running',
                                  S = 'suspended',
                                  CD = 'complete',
-				 RD = 'held'
+				 RD = 'held',
```
ye-luo: No tab please

camelto2: sorry, fixed

prckent:
Tested on both Edison and Cori?

camelto2:
@prckent It has been checked for both edison and cori, as well as checking  that it does not disrupt the behavior on another squeue machine like stampede.

Issue #512, opened 2017-11-20
Interfacing QMCPACK to other periodic codes


prckent:
Are there any write ups on the format of either g-space or real-space inputs? e.g. To implement QE 6.2 , ABINIT, QBox, or a real space code, where should one look?

The manual https://docs.qmcpack.org/qmcpack_manual.pdf is silent on this topic, but perhaps there are some old docs, wiki pages, writeups with (old, unsupported) wfconvert?

An alternate route would be to implement loading of trial wavefunctions in new formats directly in QMCPACK, reverse engineering what is done for the currently supported formats.

prckent:
For reference, I'll note here that to-date we have not received a request to interface to another periodic code besides QE, but we would certainly give any request strong consideration if it would help a new user or enable some research.

PR #1425, opened 2019-03-07
Qbox Converter

lshulen:
This introduces a barebones converter to make eshdf files for qmcpack from QBox sample files.  It has been tested for several different cases including multiple species, non-orthorhombic boxes, net-magnetic moment not equal to zero and arbitrary k-points.

The biggest item that needs to be addressed (as well as numerous policy / documentation questions) is that for the moment QBox does not put the eigenvalues in the sample file.  I've asked Francois Gygi to think about remedying this, but for the moment I am just putting ascending values for the energy of each SPO starting at -5000.  This means that one likely cannot use the converter for doing a supercell with k-points (although you might get lucky if each k-point has the same occupation).  I would also be happy for suggestions as to how to do this in other ways (although parsing the regular QBox output file to get it would be rather painful).

I also have a number of other questions:
 -- What documentation is needed for this?
 -- What sort of tests should be developed?
 -- What parts of my coding style etc. are unsuitable for inclusion in QMCPACK?

I expect there will be some consternation as I have used my own very simple XML parser rather than libxml2.  I have done this because it is reasonably fast and does not require holding the entire contents of the XML file in memory (rather I am reading a single band at a time).

I have also used a loose wrapper to FFTW (see FftContainer.h).  I have not looked closely at whether there is something already in the qmcpack source that would be a better fit.


prckent:
No problem with the XML parsing choice, imo.

prckent:
The FFT is needed because QBox stores orbitals in real space?

lshulen:
Yes, the qbox sample file stores only real space wave functions.  (as an aside, it took me the better part of a couple of days to figure out that I was using the wrong direction for the FFT)

prckent:
This will also be helpful for interfacing with RMG :-)

lshulen:
Note, that this partially addresses issue #512  .  Probably given the machinery here, it would be fairly straightforward to add an abinit converter based on this skeleton.  (Although I'm not volunteering at the moment)

src/QMCTools/CMakeLists.txt
```diff
@@ -47,6 +50,15 @@ ADD_EXECUTABLE(qmc-check-affinity check-affinity.cpp)
 INSTALL(TARGETS qmc-check-affinity
         RUNTIME DESTINATION bin)
 
+ADD_EXECUTABLE(QBoxConverter QBoxConverter.cpp XmlRep.cpp WriteEshdf.cpp)
```
markdewing: To follow the pattern, the name of the converter should have qmc in it.   Maybe qmc-qbox-converter ?

markdewing:
Would tinyxml2 meet your needs for an XML parser? Are the QBox files large enough that partial reading of the file is necessary?
A second XML parser in the code would be okay, but it would be nice to reuse something existing.   The use of XML in QMCPACK is simple enough that it might be feasible at some point to replace libxml with tinyxml (and remove a library dependency).

prckent:
By the same logic, updating the XML handling here would not be much work...

lshulen:
Regarding tinyxml, I did look at it briefly (also rapidxml), but it appears to require having the whole contents of the xml file in memory.  This may have been premature optimization, but I remember we used to have a big problem when working with large wavefunctions going from quantum espresso to qmcpack (think large molecules in a big box or disordered condensed phases) until Ye parallelized the converter.  I chose to go the current route because I would only need enough memory in the place where the conversion is being done to store a single band (actually maybe two copies of the band the way I have done it) rather than the entire wavefunction at once.  

That said, if all we want the qbox -> qmcpack route for is doing simple tests regarding the kleinman-bylander vs semilocal issue, then we could absolutely use one of the other xml parsers and just not worry about memory.

markdewing:
@lshulen Thanks for the longer description - it helps to have this written down.

I think it's okay to keep using the parser for this prototyping stage.  We should at some point make a clear decision - continue using this one, or work on switching to something else (tinxml2? libxml? a streaming C++ parser like expat.  Though I expect the cost of another external dependency would lean strongly against this option)

One point that would make it easier to justify using a custom XML parser is to reduce/constrain the scope - if the scope is 'can parse QBox XML files', the testing and validation load is much easier than being able to parse arbitrary XML.

lshulen:
A question about tests.  I could certainly provide a couple of qbox sample files and hdf outputs from this tool as a reference.  You could then have the converter run a check to see if the produced files differ.  On my machine, using diff to do this fails (not sure why exactly, perhaps hdf puts some sort of timestamp in the hdf archive).  However, h5diff correctly sees that the contents of the files are the same.  I'd rather not require access to h5diff for the test to work though.  Is this the correct route?  Should I build a little stand alone tool to compare the contents of the hdf files without h5diff?

ye-luo:
@lshulen h5diff is the correct way and we are already using it in other converter tests.

lshulen:
> @lshulen Thanks for the longer description - it helps to have this written down.
> 
> I think it's okay to keep using the parser for this prototyping stage. We should at some point make a clear decision - continue using this one, or work on switching to something else (tinxml2? libxml? a streaming C++ parser like expat. Though I expect the cost of another external dependency would lean strongly against this option)
> 
> One point that would make it easier to justify using a custom XML parser is to reduce/constrain the scope - if the scope is 'can parse QBox XML files', the testing and validation load is much easier than being able to parse arbitrary XML.

@markdewing, this is an excellent point.  In fact, I can think of several things that are not quite standards compliant about my xml parser.  For instance, I'm not currently chasing href= attributes, so my tool would not necessarily allow parsing qmcpack input files depending on how they are generated.  If people decided they liked my parser, this could be fixed without too much issue, but full standards compliance is certainly not something I have been chasing.

lshulen:
> @lshulen h5diff is the correct way and we are already using it in other converter tests.

Cool, one more question then.  qbox is writing real space wavefunctions that I am doing an FFT on before writing the hdf file.  For this reason, I'm not expecting 16 digits of accuracy on different architectures for the wavefunction values when running the converter.  Do you currently use -d or -p in the converter tests?  If so, what threshold do you think is appropriate?

ye-luo:
We can enable the -p or -d after the nightly runs and figure out what is an appropriate value.
We had a norm check of the hdf5 PW orbitals, did you get qmcpack stopping you because of that?

ye-luo:
I vaguely remember that Qbox write wavefunction binary into XML. If this is true, our customized XML parse is probably needed. I don't have a concern in the FFT but the hdf5, can we use the hdf_archive? or it is too hard to use? Let us know your experience. We probably need more documentation and designing there.

lshulen:
@ye-luo
Regarding the norm of the wavefunction, I've not had any problems with that.  Upon reading the wavefunction, I divide all elements by sqrt(number of elements).  I also divide them by the same factor after doing the fft into k-space.  After this, the norm seems to be fine. (I've checked it and it is 1 both in real and k-space for all orbitals, there are actually a couple of lines of commented code in the converter (WriteEshdf.cpp around L137) that spit out the L2 norm in both real and k-space).

For qbox, I've been using the -ascii option in writing the sample file.  If you do not do this, it is required to be in little-endian binary.  I can look into being able to handle this as a feature for the future.  Currently the converter will die with an error message if you have a sample file containing binary rather than ascii data.

For your question about hdf5, is hdf_archive the way qmcpack is internally reading and writing hdf?  If so, I can certainly look into it, but have not yet done so.

prckent:
While this is a WIP, I think it would be a good idea to clearly determine what is needed to get this bare bones version merged and what might be improved in later improvements as experience is gaines.

lshulen:
I appreciate the comment.  Here is my proposed list:

- [ ] Create section in manual covering operation of the tool
- [ ] Create a test comparing output with a "gold standard" file
- [ ] Add a warning message to output with multiple k-points stating that using this for supercell calculations is unlikely to work
- [ ] Explore whether hdf_archive could be used for writing the output rather than the direct method currently in the tool

One could also argue that the last one could wait until after the initial merge.  Does this sound about right?

prckent:
IMO, the first three. The hdf_archive can be tidied later.

prckent:
In view of our push to progress WIP PRs faster, what truly needs to be fixed here or is just about ready to be added to the PR? e.g. A test would be ideal, but replacing hdf_archive could be done later.

lshulen:
Agreed, will try to get the first three done today.

lshulen:
The first three items on the list should now be addressed.  Due to the extremely long life of this as a PR, there is a minor conflict (couple character change in file name in CMakeLists.txt).  Not sure how to resolve this.  If it's too cumbersome, I could just create a new PR starting from the most recent develop...

prckent:
(Will look at this after the 3.8.0 release which will likely occur tomorrow)

PDoakORNL:
The converter should have Qbox in its name. 

I think we should follow our naming conventions in all new code.
Relevant here: AClassOrTypeName

lshulen:
@prckent, the testing situation is concerning to me.  I used the relative tolerance option, which as you have pointed out, is susceptible to small fluctuations for values near zero.  If the threshold is increased to account for that, I personally would be concerned about 1% fluctuations in the wavefunction value induced by the converter affecting the results.  I'm in favor of switching to using an absolute tolerance given this, perhaps given your experiment above, would "-d 1e-9" be agreeable?  I suppose the most important check would be to see that the VMC energy without a jastrow would be the same to within a small tolerance, but this test is very much not designed to do that.

@prkent and @PDoakORNL.  I named the executable convertpw4qmc to be an echo of the convert4qmc tool that handles conversion from gaussian type codes.  I had initially thought to use qbox in the name, but since the initial WIP PR, I have been working to add functionality to make hdf5 wave functions from the quantum espresso output for calculations involving spinors.  Since I was using the same code base for both things, I thought it better to have a single executable that could potentially handle making wave functions from any DFT code based on plane waves, so I changed to a generic convertpw4qmc.  

@PDoakORNL the change in naming conventions can be done.  I've been playing with these concepts in other codes for some time.  If I change all of the variable / class names it will fork this code from those.  This is the reason I haven't done this so far, but if it is the price for entry into qmcpack, as I said, I can make the change.

prckent:
How far from supporting wavefunctions from QE are you ? I thought more entries would be needed in the QE HDF5 files, implying a long timescale. If not too far off the general name of the converter makes sense. (It would be a big improvement to not have to have to maintain our pw2qmcpack for QE.)

prckent:
Would "-d 1e-9" be agreeable? Yes.

Historically the kinds of bugs that have been made in converters or that have been introduced by changes in the upstream codes, have all resulted in large changes in the wavefunctions. Using the -d setting is fine by me.

( My tests were on an OSX box with macports installed gcc and fftw )

PDoakORNL:
@lshulen re: naming, If that's the intention for the converter then that makes sense. 

re: Naming conventions: In my opinion following the collectively agreed coding conventions is the price.

prckent:
Lets follow our naming conventions.

lshulen:
OK, will make a PR shortly to fix the test and to update the naming conventions.  As far as the QE converter aspect, between the xml and hdf files, I think that I have everything that I need.  Last I looked, there was even a file containing the charge density in hdf.  I need to look up the format for that eshdf, but I think it should be straightforward.

lshulen:
@prckent commits should address requested changes.

prckent:
Confirming that the revised test works for me. Thanks for the source edits.

prckent:
Test this please

prckent:
OK to test

PR #1428, opened 2019-03-08
Improve single det class inheritance

ye-luo:
This PR cleans up the inheritance of single determinant classes.
DiracDeterminantBase is the base class
DiracDeterminant is the only single det class.
DiracDeterminantOpt is removed since it was intended for orbital rotation but not touched since 2010 and we don't have input example and test. All the orbital optimization will be directly implemented in DiracDeterminant by the Berkeley effort.
DiracDeterminantWithBackflow is directly based on DiracDeterminantBase. The change I made was very little. It indicates using DiracDeterminant as a base class is not necessary.


ye-luo:
Test this please

ye-luo:
The old comment was funny. WaveFunctionComponent contains the full description of evaluateLog.

PR #1430, opened 2019-03-11
Add and tidy applications aborts

prckent:
1. QMCHamiltonianBase::addEnergy was missing an abort, causing the CUDA code to crash when certain unimplemented functionality was used (user reported).
2. Tidied many APP_ABORT calls for consistency.
3. I noticed that unit_test_hamiltonian calls WaveFunctionComponent::evalGradSource incorrectly. The function in the base class should never be called. The abort there was commented out, and still is. Simulations run correctly with the abort present, so this is an issue in the test setup.

### DIFF
--- a/src/QMCWaveFunctions/SPOSet.h
+++ b/src/QMCWaveFunctions/SPOSet.h
@@ -20,7 +20,6 @@
 #ifndef QMCPLUSPLUS_SINGLEPARTICLEORBITALSETBASE_H
 #define QMCPLUSPLUS_SINGLEPARTICLEORBITALSETBASE_H
 
-#include "Message/Communicate.h"
 #include "OhmmsPETE/OhmmsArray.h"
 #include "Particle/ParticleSet.h"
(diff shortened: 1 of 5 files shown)


src/QMCWaveFunctions/SPOSet.h
```diff
@@ -218,9 +219,7 @@ class SPOSet: public QMCTraits
   virtual void
   evaluate (const ParticleSet& P, PosType &r, ValueVector_t &psi)
   {
-    app_error() << "Need specialization for SPOSet::evaluate "
-                << "(const ParticleSet& P, PosType &r).\n";
-    abort();
+    APP_ABORT("Need specialization for SPOSet::evaluate(const ParticleSet& P, PosType &r)\n");
```
ye-luo: Aiming at reducing header inclusion. Please move the function definition into SPOSet.cpp and remove "#include "Message/Communicate.h"

prckent: Good point. Updated.

Issue #1408, opened 2019-03-01
nexus tile may duplicate atom position


Paul-St-Young:
I have attached an example where nexus.Structure::tile produced a supercell with two atoms on top of each other. @jtkrogel do you see an easy fix for this? 

I found a workaround by adding "use_ase" flag to tile(). May I PR this as a feature?
```
diff --git a/nexus/lib/structure.py b/nexus/lib/structure.py
index a374ca728..9f2d6143d 100755
--- a/nexus/lib/structure.py
+++ b/nexus/lib/structure.py
@@ -2947,6 +2947,7 @@ class Structure(Sobj):
 
 
     def tile(self,*td,**kwargs):
+        use_ase            = kwargs.pop('use_ase',False)
         in_place           = kwargs.pop('in_place',False)
         magnetic_order     = kwargs.pop('magnetic_order',None)
         magnetic_primitive = kwargs.pop('magnetic_primitive',True)
@@ -2979,8 +2980,18 @@ class Structure(Sobj):
 
         self.recenter()
 
+        # get atomic positions "pos"
+        if use_ase:
+          from ase import Atoms
+          from ase.build import make_supercell
+          atext = ''.join(self.elem)
+          pbc = [True if b == 'p' else False for b in self.bconds]
+          a0 = Atoms(atext, cell=self.axes, positions=self.pos, pbc=pbc)
+          a1 = make_supercell(a0, tilematrix)
+          pos = a1.get_positions()
+        else:
+          pos  = self.tile_points(self.pos,self.axes,tilematrix,tilevector)
         elem = array(ncells*list(self.elem))
-        pos  = self.tile_points(self.pos,self.axes,tilematrix,tilevector)
         axes = dot(tilematrix,self.axes)
 
         center   = axes.sum(0)/2
```

[tile.zip](https://github.com/QMCPACK/qmcpack/files/2920268/tile.zip)

prckent:
Better to fix the bug in tile. This should never happen.

jtkrogel:
I'm looking at this now.  Apparently performing the shear operations to obtain the diagonal tiling can depend on the order in which you perform them.  I will explore whether an alternative based on row reduction and pivoting would do the right thing.

jtkrogel:
@Paul-St-Young, please see #1432.  Can you confirm that the fix works for you?

Paul-St-Young:
Yes, I can confirm the problem is fixed. Please go ahead and close this issue once #1432 is merged.
[...]

PR #1432, opened 2019-03-11
Nexus: make tiling more robust

jtkrogel:
This PR fixes the issues raised in #1408.  Full matrix tilings are trickier than first appreciated.  

Nexus attempts to arrive at an effective diagonal tiling by reducing the tiling matrix into a volume preserving diagonal form via successive shear operations.  This usually works, but the diagonal tiling apparently depends on the order of the shearing operations (e.g. you can sometimes get either (1,1,4) or (1,2,2)) as revealed by #1408.

This PR added tests of diagonal and matrix tiling and also a search over permutations in the shearing operation order.  This change worked for the problem case of #1408, however it did not work for all of the randomly generated tiling matrices just added to the test set.

A final brute force fallback operation was added to catch the remaining cases: essentially find a bounding box for the tiled supercell in the basis of the untiled axes, generate new atomic coordinates in this bounding box, and then remove duplicate points.  This approach was avoided originally because it's worst case scaling is order det(T)^3 rather than order det(T) as above.


Paul-St-Young:
This does resolve the problem case I found. Thanks @jtkrogel!

markdewing:
okay to test

PR #1433, opened 2019-03-12
Fix backflow

ye-luo:
#1428 introduced a bug and caused segmentation fault due to missing resizing of vectors.
This PR should fix problems like https://cdash.qmcpack.org/CDash/testDetails.php?test=4493186&build=47567


prckent:
That all the runs did not crash indicates we need to be doing some regular memory checking in the nightlies.

ye-luo:
Only backflow has the issue. It is expected that most runs should be fine.

prckent:
Perhaps oxygen did not pickup the changes when it ran, but the BF tests passed on oxygen https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=short-bccH_2x2x2_ae-deriv-1-1&date=2019-03-12

ye-luo:
The commit log in the output file shows oxygen had started testing before you merged my PR.

PR #1435, opened 2019-03-12
NEXUS user guide builds with xelatex and fixes

prckent:
Build the NEXUS user guide with xelatex consistent with QMCPACK manual. With this change the user guide now builds. Shell escape still required for pygments and minted package use. Fixed missed/duped references.

### DIFF
--- a/nexus/docs/build_nexus_user_guide.sh
+++ b/nexus/docs/build_nexus_user_guide.sh
@@ -1,24 +1,23 @@
 #!/bin/sh
-if [ -z "$(command -v pdflatex)" ]; then
-  echo "Error: pdflatex is required for build_nexus_user_guide.sh"
+echo "----♦----♦----♦----♦----♦----♦----♦----♦----♦----♦----♦----♦----"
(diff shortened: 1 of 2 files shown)


ye-luo:
@jtkrogel Any concern?

jtkrogel:
@prckent Thanks for the updates.

I assume building w/ pdflatex has no issues with your changes.  If so, please also include a script to build it this way (e.g. build_nexus_user_guide_pdflatex.sh).

Nexus versioning is now tied to the QMCPACK release cycle, with the current version being 1.6.0.  See nexus.py line 246:
```
parser = OptionParser(usage=usage,add_help_option=True,version='%prog 1.6.0')
```

Should this be included in the manual now, or does the version need to be easier to query by the manual build script?

prckent:
I was not able to build the NEXUS manual using pdflatex due to issues of unicode handling.

Versions can be fixed in a later PR. e.g. Grep the version from nexus.py and put on the first page, and have an override for the development version.

jtkrogel:
I will add the pdflatex script later provided I can build with it.

PR #1437, opened 2019-03-13
Manage raw MPI communicator in Communicate

ye-luo:
Initially, I need to create a communicator to handle multiple GPUs assignment to MPI ranks.
Then I realized that more fix is needed before doing that.

This PR is to adjust how the raw communicator myMPI is managed in Communicate.
Before this PR, there are three cases to initialize Communicate.
1. Communicate constructed from a raw communicator and its value is copied to myMPI.
2. Communicate split a parent communicator and assign the sub communicator to myMPI.
3. Communicate copies the communicator by value from mpi3 wrapper.

In case 1,3 myMPI is from external. case 2 myMPI is local.
For this reason, the destruction of Communicate is not fully clean.
With the new change in case 1 and 3, the raw communicator passed in is duplicated by MPI_Comm_dup and no more relies on the original raw communicator.
There is no change in case 2.
The destructor fully destroys myMPI by MPI_Comm_free and exits clean.

Now I can create a node level Communicate object which can be used local.
In general, having any sorts of global communicator is discouraged.

More notes: OOMPI communicator object doesn't handle free of the MPI raw communicator. So there is no double free problem. In the future when mpi3 wrapper object comm should be the only one owing a raw communicator and myMPI will be just a reference.

Some side change: I was doing changes back and forth and those changes just got in. Headache to separate them. Fortunately, they are very small.
Testlibstdc++ needs CXX14 flag to have XL through on Summit.
qmc_common.compute_device is used like ifdef and changed to use ifdef explicitly.


prckent:
How does this complement/conflict with the new MPI wrapper? ( @markdewing, @ye-luo  )

markdewing:
Ye and I had a discussion and this code is fine for now.  It will eventually be replaced by the new mpi wrapper code.   There's no good way to replace the code in intializeAsNodeComm yet.

prckent:
Test this please

PR #1441, opened 2019-03-13
OpenMP offload of spline SPO C2R

ye-luo:
This PR enables the OpenMP offload of the spline SPO calculation from the CPU build. Only C2R case is implemented at the moment. All other cases will be added later.
To access this feature, -DENABLE_OFFLOAD=1 is needed. Currently, only XL compiler is fully supported. Clang will be the next.
The input tag is the same as using the CUDA code, putting gpu="yes" on the line specifying SPO is sufficient. Offload can be selected on/off by this tag. It will make some heavy portion of the calculation to GPU.


ye-luo:
Test this please

prckent:
One of my concern with #1444 will be that that we are directly calling CUDA blas. These should be behind a wrapper of our own ("acceleratorBlas") as is done in the AFQMC code to allow for straightward substitution by AMD/INTEL/MAGMA/XYZ BLAS.  This is an architectural problem that should not be allowed to take hold. Generally, if whenever we make a vendor specific call we should ask ourselves what is generic and put it behind an interface/abstraction layer that we own. We should not have CUDA* called directly by our QMC algorithms. Allocators etc. should be similarly protected - that is the whole point of having them.


Is there anything similar here?

ye-luo:
This PR is OpenMP only and has no contamination from vendor specific libraries.

prckent:
Approving post discussions

PR #1442, opened 2019-03-13
clang-format of QMCWaveFunctions/Fermion folder

PDoakORNL:


src/QMCWaveFunctions/Fermion/DelayedUpdate.h
```diff
@@ -19,179 +19,271 @@
...
+    delay_list[delay_count] = rowchanged;
+    // the new Binv is [[X Y] [Z x]]
+    BLAS::gemv('T', norb, delay_count + 1, cminusone, V.data(), norb, psiV.data(), 1, czero, p.data(), 1);
+    // x
+    T y = -p[delay_count];
+    for (int i = 0; i < delay_count; i++)
+      y += Binv[delay_count][i] * p[i];
+    Binv[delay_count][delay_count] = y = T(1) / y;
+    // Y
+    BLAS::gemv('T',
```
ye-luo: Is there a way to teach clang format that don't break lines for a function call no matter how long the call is?
If you look for t3_contract in Spline?2?Adoptor.h They are all ugly.
When a function is defined breaking the long list of argument is good, but not the place we call it.

ye-luo: Or at least not one argument per line.

prckent: I would like to keep the reformatting moving. Can we either live with the ugliness or protect these sections?

ye-luo: @prckent I'm in favour of merge.
@PDoakORNL do you know if there is such a setting in clang-format?

PR #1444, opened 2019-03-13
Add delayed updates using cublas in the CPU code path

ye-luo:
To access the new code use -DENABLE_CUDA=1.
ENABLE_CUDA is mutual exclusive to QMC_CUDA.
I had discussion with @mmorale3. It seems that both real-space and afqmc needs to avoid QMC_CUDA which only supports double precision on the host-side.
ENABLE_CUDA will turn on the cublas implementation of delayed update in the CPU code path.
Putting gpu="yes" as an attribute of slaterdeterminant turns on this feature(by default on).

More details:
1. DiracDeterminant class becomes a template class  with delayed update engine as the template parameter. This avoids a lot of code duplication.
2. DelayedUpdateCUDA is the implementation mostly using cublas and still one cuda kernel is needed.
3. allocators based on CUDA malloc functions are added.
4. Vector and Matrix container classes support memory attributes to handle device only memory space. Initialization is only done when the memory space is host accessible.
5. Platforms/accelerators.cpp is added to initialize GPUs. MPI ranks are equally distributed to the GPUs. When assigning 2 GPUs, 4 MPI ranks get 0,0,1,1. This is opposite to the round-robin scheme in QMC_CUDA code path.
6. qmc_common.default_accelerator_id is added. In the future if sharing data among accelerators are needed, a local communicator can be created from the parent communicator and the shared device list can be reconstructed based on the default_accelerator_id per rank without querying devices.


ye-luo:
First reason, GPU to CPU socket affinity affects performance.
With ENABLE_CUDA, 4 MPI ranks get 0,0,1,1. Rank 0&1 on the first socket talk to GPU 0 and rank 2&3 on the second socket talk to GPU 1
With QMC_CUDA, 4 MPI ranks get 0,1,0,1. Rank 1&2 introduce cross-socket traffic.

mmorale3:
This assumes -d packed in summit, which is what we want.
The real problem is lack of consistency in task placement between different systems. 

Miguel

ye-luo:
All the system I have seen put MPI tasks either packed or spread but never round-robin by default.
For this reason, having GPU assignment round-robin by default doesn't make sense. However, if round-robin assignment of GPUs is needed, it can be added very easily. For the moment, I don't see this need.

ye-luo:
Question from @prckent One of my concern will be that that we are directly calling CUDA blas. These should be behind a wrapper of our own ("acceleratorBlas") as is done in the AFQMC code to allow for straightward substitution by AMD/INTEL/MAGMA/XYZ BLAS. This is an architectural problem that should not be allowed to take hold. Generally, if whenever we make a vendor specific call we should ask ourselves what is generic and put it behind an interface/abstraction layer that we own. We should not have CUDA* called directly by our QMC algorithms. Allocators etc. should be similarly protected - that is the whole point of having them.

src/QMCApp/QMCMain.cpp
```diff
@@ -82,10 +83,15 @@ QMCMain::QMCMain(Communicate* c)
   qmc_common.print_git_info_if_present(app_summary());
   app_summary() << "=====================================================\n";
   qmc_common.print_options(app_log());
-  app_summary() << "\n  MPI Ranks             = " << OHMMS::Controller->size()
-                << "\n  MPI Ranks per group   = " << myComm->size()
-                << "\n  MPI Group ID          = " << myComm->getGroupID()
-                << "\n  MPI Ranks per node    = " << NodeComm.size() << std::endl;
+  app_summary()
+      << "\n  MPI Ranks in total    = " << OHMMS::Controller->size()
```
prckent: "Total number of MPI Ranks" or  "Total MPI Ranks"

src/QMCApp/QMCMain.cpp
```diff
@@ -82,10 +83,15 @@ QMCMain::QMCMain(Communicate* c)
...
+      << "\n  Number of rank groups = " << myComm->getNumGroups()
+      << "\n  MPI Group ID          = " << myComm->getGroupID()
+      << "\n  MPI Ranks in group    = " << myComm->size()
+      << "\n  MPI Ranks per node    = " << NodeComm.size()
+      << std::endl;
+  // assign accelerators within a node
```
prckent: Add a note for QMC_MPI=0 builds?

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
```diff
@@ -0,0 +1,205 @@
...
+#include "QMCWaveFunctions/Fermion/delayed_update_helper.h"
+#include <cuda_runtime_api.h>
+#include "CUDA/cudaError.h"
+
+namespace qmcplusplus {
+
+  class Range
+  {
+    // [first, last) rows of Ainv
+    int first, last;
```
prckent: Thanks for putting the class data private :-)

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
```diff
@@ -0,0 +1,205 @@
...
+#include <OhmmsPETE/OhmmsMatrix.h>
+#include "CUDA/CUDAallocator.hpp"
+#include <Numerics/CUDA/cuBLAS.hpp>
+#include "QMCWaveFunctions/Fermion/delayed_update_helper.h"
+#include <cuda_runtime_api.h>
+#include "CUDA/cudaError.h"
+
+namespace qmcplusplus {
+
+  class Range
```
prckent: What is this class for?

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
```diff
@@ -0,0 +1,205 @@
...
+    inline int getOffset(int index) const
+    {
+      if(!checkRange(index)) throw std::runtime_error("index not in range \n");
+      return index-first;
+    }
+    inline bool checkRange(int index) const { return (index>=first) && (index<last); };
+  };
+
+  template<typename T>
+    struct DelayedUpdateCUDA
```
prckent: struct or class, public/private issues?

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
```diff
@@ -0,0 +1,205 @@
...
+        cublasErrorCheck( cublasSetStream(handle,hstream), "cublasSetStream failed!" );
+      }
+
+      ~DelayedUpdateCUDA()
+      {
+        cublasErrorCheck( cublasDestroy(handle), "cublasDestroy failed!" );
+        cudaErrorCheck( cudaStreamDestroy(hstream), "cudaStreamDestroy failed!" );
+      }
+
+      inline void resize(int norb, int delay)
```
prckent: reset? reconfigure?

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
```diff
@@ -0,0 +1,205 @@
...
+      {
+        //tempMat.resize(norb, delay);
+        V.resize(delay, norb);
+        U.resize(delay, norb);
+        p.resize(delay);
+        Binv.resize(delay, delay);
+        // prefect 8% more rows corresponding to roughly 96% acceptance ratio
```
prckent: predict not prefect? or did you mean something else?

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
```diff
@@ -0,0 +1,205 @@
...
+      inline void initializeInv(const Matrix<T>& Ainv)
+      {
+        cudaErrorCheck( cudaMemcpyAsync(Ainv_gpu.data(), Ainv.data(), Ainv.size()*sizeof(T), cudaMemcpyHostToDevice, hstream), "cudaMemcpyAsync failed!");
+        // safe mechanism
+        delay_count = 0;
+        prefetched_range.clear();
+      }
+
+      template<typename VVT>
+      inline void getInvRow(const Matrix<T>& Ainv, int rowchanged, VVT& invRow)
```
prckent: Define

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
```diff
@@ -0,0 +1,205 @@
...
+        BLAS::ger(delay_count, delay_count, cminusone, Binv[delay_count], 1, Binv.data()+delay_count, lda_Binv, Binv.data(), lda_Binv);
+        // Z
+        for(int i=0; i<delay_count; i++)
+          Binv[delay_count][i] *= -y;
+        delay_count++;
+        // update Ainv when maximal delay is reached
+        if(delay_count==lda_Binv) updateInvMat(Ainv,false);
+      }
+
+      inline void updateInvMat(Matrix<T>& Ainv, bool transfer_to_host = true)
```
prckent: define

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
```diff
@@ -0,0 +1,205 @@
...
+      {
+        // update the inverse matrix
+        if( delay_count>0 )
+        {
+          const T cone(1);
+          const T czero(0);
+          const int norb=Ainv.rows();
+          const int lda_Binv=Binv.cols();
+          const T cminusone(-1);
+          cudaErrorCheck( cudaMemcpyAsync(U_gpu.data(), U.data(), norb*delay_count*sizeof(T), cudaMemcpyHostToDevice, hstream), "cudaMemcpyAsync failed!");
```
prckent: Worth putting the formulae here or referring to file with the formulae

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminant.h
```diff
@@ -54,7 +54,7 @@ class MultiSlaterDeterminant : public WaveFunctionComponent
   NewTimer RatioTimer, RatioGradTimer, RatioAllTimer, UpdateTimer, EvaluateTimer;
   NewTimer Ratio1Timer, Ratio1GradTimer, Ratio1AllTimer, AccRejTimer, evalOrbTimer;
 
-  typedef DiracDeterminantBase* DiracDeterminantPtr;
+  typedef DiracDeterminantBase* DiracDeterminantBasePtr;
```
prckent: Good catch

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -550,11 +556,18 @@ bool SlaterDetBuilder::putDeterminant(xmlNodePtr cur, int spin_group)
...
     }
+#endif
+#if defined(ENABLE_CUDA)
+    else if (useGPU == "yes")
+    {
+      app_log() << "Using DiracDeterminant with DelayedUpdateCUDA engine" << std::endl;
+      adet = new DiracDeterminant<DelayedUpdateCUDA<ValueType>>(psi,firstIndex);
+    }
 #endif
     else
```
prckent: Robustness suggestion - check for useGPU=="no", give error on other inputs. Currently "Yes" would give an unexpected result.

src/QMCWaveFunctions/Fermion/delayed_update_helper.cu
```diff
@@ -0,0 +1,104 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2018 QMCPACK developers.
```
prckent: 2019

src/QMCWaveFunctions/Fermion/delayed_update_helper.cu
```diff
@@ -0,0 +1,104 @@
...
+  return make_cuComplex(cuCrealf(x)-1.0f, cuCimagf(x));
+}
+
+template<>
+__host__ __device__ __inline__ cuDoubleComplex subtractOne<cuDoubleComplex>(cuDoubleComplex x)
+{
+  return make_cuDoubleComplex(cuCreal(x)-1.0f, cuCimag(x));
+}
+
+template<typename T, int BS>
```
prckent: Please define these kernels. What they do etc.

src/QMCWaveFunctions/Fermion/delayed_update_helper.h
```diff
@@ -0,0 +1,38 @@
...
+// File created by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+#ifndef CUDA_DELAYED_UPDATE_HELPER_H
+#define CUDA_DELAYED_UPDATE_HELPER_H
+
+#include <complex>
+
+void applyW_stageV_cuda(const int *delay_list_gpu, const int delay_count,
```
prckent: A definition would be appreciated. Here preferred to the implementation file.

src/simd/MemorySpace.hpp
```diff
@@ -0,0 +1,25 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2018 QMCPACK developers.
```
prckent: 2019

src/CUDA/CUDAallocator.hpp
```diff
@@ -0,0 +1,116 @@
...
+//
+// Copyright (c) 2018 QMCPACK developers.
+//
+// File developed by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//
+// File created by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+// -*- C++ -*-
+/** @file CUDAallocator.hpp
+ */
```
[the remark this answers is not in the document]
prckent: Please add explanation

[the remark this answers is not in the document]
ye-luo: This is for C++ memory allocators one with cudaMalloc and the other is cudaMallocManaged. They are used for our container class Vector and Matrix. This enables very simple code changes with memory on different places.

src/CUDA/CUDAallocator.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+#include <cstdlib>
+#include <stdexcept>
+#include <cuda_runtime_api.h>
+#include "CUDA/cudaError.h"
+
+namespace qmcplusplus
+{
+  template<typename T>
+  struct CUDAManagedAllocator
```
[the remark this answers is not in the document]
ye-luo: This is based on C++ allocator example. It needs to be struct and public. I will add docs.

src/Numerics/CUDA/cuBLAS.hpp
```diff
@@ -0,0 +1,121 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2018 QMCPACK developers.
```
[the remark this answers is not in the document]
ye-luo: This is the year that file was added to the repo not necessarily the same year as the PR merge.

src/OhmmsPETE/OhmmsMatrix.h
```diff
@@ -21,18 +21,18 @@
 
 namespace qmcplusplus
 {
-template<class T, typename Alloc = std::allocator<T>>
+template<class T, typename Alloc = std::allocator<T>, unsigned MemType = MemorySpace::HOST>
 class Matrix
 {
```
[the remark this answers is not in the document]
ye-luo: Both Vector and Matrix classes have accurate public/private attributes. They are already proper OO classes. I may need to change the private to protect but it is not needed at the moment.

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
ye-luo: Sorry. I mean prefetch.

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
ye-luo: reset and be reset values. reconfigure is not clear.
This is resize internal storage. So I still prefer resize.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
ye-luo: I prefer not to add a check here. OhmmsAttributeSet class which passes input needs to do this work.

src/CUDA/CUDAallocator.hpp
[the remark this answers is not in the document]
ye-luo: Done.

src/Platforms/accelerators.cpp
```diff
@@ -0,0 +1,106 @@
...
+  int residual = num_ranks % num_devices;
+  int assigned_device_id;
+  if (rank_id < min_ranks_per_device * (num_devices - residual))
+    assigned_device_id = rank_id / min_ranks_per_device;
+  else
+    assigned_device_id = (rank_id + num_devices - residual) / (min_ranks_per_device + 1);
+  return assigned_device_id;
+}
+
+void assignAccelerators(Communicate& NodeComm)
```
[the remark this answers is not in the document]
ye-luo: defined in the header now.

src/QMCApp/QMCMain.cpp
ye-luo: Now prints "This binary is built without MPI"

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
ye-luo: I need to do this in later PR for an accurate pointer.

src/CUDA/cudaError.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <iostream>
+#include <string>
+#include <sstream>
+#include <stdexcept>
+#include <cuda_runtime_api.h>
+
+#define cudaErrorCheck(ans, cause) { cudaAssert((ans), cause, __FILE__, __LINE__); }
+/// prints CUDA error messages. Always use cudaErrorCheck macro.
+inline void cudaAssert(cudaError_t code, const std::string& cause, const char *filename, int line, bool abort=true)
```
markdewing: The abort parameter appears not to have any effect.

mmorale3:
Regarding the last question, the plan is to make QMC_CUDA refer to legacy
CUDA code and use a new flag for new developments. Once this PR is
accepted,
I will switch to ENABLE_CUDA in AFQMC and will not reference QMC_CUDA
anymore.

Miguel

ye-luo:
@mmorale3 in the AFQMC if CUDA is enabled, is CPU and GPU code selectable from input file?

src/CUDA/cudaError.h
ye-luo: Fixed.

mmorale3:
No, cuda build forces execution on gpu. It can be a runtime decision given the current design, but I didn’t see a point to it.

ye-luo:
It very useful for debugging. The other reason is memory, if GPU memory is not enough, I can run certain portion on CPU.

mmorale3:
In AFQMC, it is the same code path, so you just need to compile without
QMC_CUDA to do what you mention.

prckent:
(History shows that the ability to do a cpu vs gpu comparison is extremely useful for debugging. i.e. Same code paths, equivalent implementations. )

ye-luo:
OhmmsMatrix / OhmmsVector change comes after I tried a few ways back and forth with DelayedUpdateCUDA implementation and finally find this way is the best.

Regarding 1CPU, 2GPU configurations situation. The ENABLE_CUDA configuration will be used for AFQMC production but experimental for real-space. The goal is eventually getting rid of all the code-path under QMC_CUDA. If some one wants a simulation with large memory that doesn't fit GPU and still use GPU whenever possible. They can use this version of code. Why do we want another build option? ENABLE_CUDA is starting from CPU code and ensures feature complete.

ye-luo:
For testing, we will need a third category different from the existing CPU, GPU.
The AFQMC and GPU accelerated portion of real-space will be tested with ENABLE_CUDA .

ye-luo:
Anything left blocking the merge?

ye-luo:
@prckent Anything left you still have concern?

src/OhmmsPETE/OhmmsMatrix.h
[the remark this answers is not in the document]
prckent: I take back my comment - too used to seeing public variables elsewhere in the source. Only definitions here (good).

prckent:
Sorry for the slight breakage!

PR #1447, opened 2019-03-14
Enforce the destruction of myMPI after myComm

ye-luo:
A bug was introduced in #1437.
OOMPI object must be destructed before the MPI raw communicator gets freed.
I was doing that by calling the destructor by hand before freeing the raw communicator.
This is wrong because the destructor will be called again by the auto generated code in the Communicate class and try to access the communicator which has been freed already.
The solution is using a helper class which has a destructor freeing the raw communicator.
As long as it is declared before the OOMPI object, the destruction order is correct.

### DIFF
--- a/src/Message/Communicate.cpp
+++ b/src/Message/Communicate.cpp
@@ -43,14 +43,18 @@ int TagMaker::CurrentTag = 1000;
 Communicate* OHMMS::Controller = new Communicate;
 
 //default constructor: ready for a serial execution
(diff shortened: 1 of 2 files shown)


markdewing:
How did this bug manifest?  I'm curious because I'd like to make the code works correctly when using the new mpi wrapper.

ye-luo:
I caught this on Titan. I got "invalid communicator" error from MPI.

PR #1448, opened 2019-03-14
Switch some constructors to the new mpi wrapper

markdewing:
The constructor accepting a bare MPI communicator was changed to accept an mpi3::communicator instead.

The splitting constructor for ensemble runs and the shared node initialization functionnnow use the mpi wrapper for implementation.

The goal is to make the new wrapper responsible for management of the MPI communicators.
This might mean that #1447 is not necessary.

### DIFF
--- a/src/Message/Communicate.cpp
+++ b/src/Message/Communicate.cpp
@@ -62,10 +62,6 @@ Communicate::Communicate(const mpi3::environment &env):
 
 Communicate::~Communicate()
 {
-#ifdef HAVE_MPI
-  myComm.~OOMPI_Intra_comm();
-  if(myMPI!=MPI_COMM_NULL) MPI_Comm_free(&myMPI);
-#endif
   if(GroupLeaderComm!=nullptr) delete GroupLeaderComm;
 }
 
@@ -88,13 +84,10 @@ Communicate::Communicate(const mpi3::communicator &in_comm):
 Communicate::Communicate(const Communicate& in_comm, int nparts)
 {
   std::vector<int> nplist(nparts+1);
-  int p=FairDivideLow(in_comm.rank(), in_comm.size(), nparts, nplist); //group
-  int q=in_comm.rank()-nplist[p]; //rank within a group
-
-  comm = in_comm.comm.split(p, q);
+  int p = FairDivideLow(in_comm.rank(), in_comm.size(), nparts, nplist); //group
+  comm = in_comm.comm.split(p, in_comm.rank());
   myMPI = &comm;
   myComm = OOMPI_Intra_comm(myMPI);
-
 
   d_mycontext=myComm.Rank();
   d_ncontexts=myComm.Size();
@@ -119,7 +112,7 @@ Communicate::Communicate(const Communicate& in_comm, int nparts)
 void Communicate::initialize(const mpi3::environment &env)
 {
   comm = env.world();
-  MPI_Comm_dup(&comm, &myMPI);
+  myMPI = &comm;
   myComm = OOMPI_Intra_comm(myMPI);
   d_mycontext = myComm.Rank();
   d_ncontexts = myComm.Size();
--- a/src/Message/Communicate.cpp
+++ b/src/Message/Communicate.cpp
@@ -35,6 +35,10 @@
 #include "ADIOS/ADIOS_config.h"
 #endif
 
+#ifdef HAVE_MPI
+#include "mpi3/shared_communicator.hpp"
+#endif
+
 
 //static data of TagMaker::CurrentTag is initialized.
 int TagMaker::CurrentTag = 1000;
@@ -68,15 +72,17 @@ Communicate::~Communicate()
 //exclusive:  OOMPI, MPI or Serial
 #ifdef HAVE_OOMPI
 
-Communicate::Communicate(const mpi_comm_type comm_input):
+#ifdef HAVE_MPI
+Communicate::Communicate(const mpi3::communicator &in_comm):
   d_groupid(0), d_ngroups(1), GroupLeaderComm(nullptr)
 {
-  MPI_Comm_dup(comm_input, &myMPI);
+  comm = mpi3::communicator(in_comm);
+  myMPI = &comm;
   myComm = OOMPI_Intra_comm(myMPI);
-  // TODO: mpi3 needs to define comm
   d_mycontext = myComm.Rank();
   d_ncontexts = myComm.Size();
 }
+#endif
 
 
 Communicate::Communicate(const Communicate& in_comm, int nparts)
@@ -84,26 +90,24 @@ Communicate::Communicate(const Communicate& in_comm, int nparts)
   std::vector<int> nplist(nparts+1);
   int p=FairDivideLow(in_comm.rank(), in_comm.size(), nparts, nplist); //group
   int q=in_comm.rank()-nplist[p]; //rank within a group
-  MPI_Comm_split(in_comm.getMPI(),p,q,&myMPI);
-  myComm=OOMPI_Intra_comm(myMPI);
-  // TODO: mpi3 needs to define comm
+
+  comm = in_comm.comm.split(p, q);
+  myMPI = &comm;
+  myComm = OOMPI_Intra_comm(myMPI);
+
+
   d_mycontext=myComm.Rank();
   d_ncontexts=myComm.Size();
   d_groupid=p;
   d_ngroups=nparts;
+
   // create a communicator among group leaders.
-  MPI_Group parent_group, leader_group;
-  MPI_Comm_group(in_comm.getMPI(), &parent_group);
-  MPI_Group_incl(parent_group, nparts, nplist.data(), &leader_group);
-  MPI_Comm leader_comm;
-  MPI_Comm_create(in_comm.getMPI(), leader_group, &leader_comm);
+  nplist.pop_back();
+  mpi3::communicator leader_comm = in_comm.comm.subcomm(nplist);
   if(isGroupLeader())
     GroupLeaderComm = new Communicate(leader_comm);
   else
     GroupLeaderComm = nullptr;
-  MPI_Comm_free(&leader_comm);
-  MPI_Group_free(&leader_group);
-  MPI_Group_free(&parent_group);
 }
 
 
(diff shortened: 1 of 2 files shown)


src/Message/Communicate.cpp
```diff
@@ -35,6 +35,10 @@
 #include "ADIOS/ADIOS_config.h"
 #endif
 
+#ifdef HAVE_MPI
+#include "mpi3/shared_communicator.hpp"
```
ye-luo: We only use the shared_communicator via the split_shared function.
This header inclusion seems not very logical.
It is probably a design issue in mpi3

src/Message/Communicate.cpp
```diff
@@ -68,43 +68,39 @@ Communicate::~Communicate()
...
-  MPI_Comm leader_comm;
-  MPI_Comm_create(in_comm.getMPI(), leader_group, &leader_comm);
+  nplist.pop_back();
+  mpi3::communicator leader_comm = in_comm.comm.subcomm(nplist);
   if(isGroupLeader())
     GroupLeaderComm = new Communicate(leader_comm);
   else
     GroupLeaderComm = nullptr;
-  MPI_Comm_free(&leader_comm);
-  MPI_Group_free(&leader_group);
```
ye-luo: Here I free the groups after using them.
The mpi3 wrapper subcomm routine doesn't do it.
@correaa Do you think a fix is needed in subcomm?

correaa:
On Thu, Mar 14, 2019 at 11:13 AM Ye Luo <notifications@github.com> wrote:

> *@ye-luo* commented on this pull request.
> ------------------------------
>
> In src/Message/Communicate.cpp
> <https://github.com/QMCPACK/qmcpack/pull/1448#discussion_r265702959>:
>
> >    if(isGroupLeader())     GroupLeaderComm = new Communicate(leader_comm);
>    else
>      GroupLeaderComm = nullptr;
> -  MPI_Comm_free(&leader_comm);
> -  MPI_Group_free(&leader_group);
>
> Here I free the groups after using them.
> The mpi3 wrapper subcomm routine doesn't do it.
> @correaa <https://github.com/correaa> Do you think a fix is needed in
> subcomm?
>

Not freeing the `mpi3::communicator leader_comm = ...;` is not a problem.
You can always artificially scope it if you want be sure that the resource
is released.

The problem is more likely this `GroupLeaderComm = new
Communicate(leader_comm);`
because I don't know if the constructor of `Communicate` is already
stealing the resources of `leader_comm`.

ye-luo:
@correaa leader_comm is fine. It will be release when this function ends.
Communicate(leader_comm) is also fine.
The constructor(in_comm) does
```
  comm = mpi3::communicator(in_comm); // you have a MPI_Comm_dup beneath
```
So it doesn't steal leader_comm.

I mean here
```
        communicator subcomm(Vector const& v) const{
                MPI_Group old_g;
                MPI_Comm_group(impl_, &old_g);
                MPI_Group new_g;
                MPI_Group_incl(old_g, v.size(), v.data(), &new_g);
                communicator ret;
                MPI_Comm_create(impl_, new_g, &ret.impl_);
                return ret;
        }
```
That old_g and new_g are not freed after the creation and use.

correaa:
Hi Ye, you are right, I was not freeing the groups. I corrected it now. I
forgot this function existed, I think it was from a time I didn't have a
mpi3::group wrapper. Probably it is a bad function and I would like to
removed.

In any case, if I understand correctly that whole function is just to
divide the communicator.

If you want a fair share division you can do this:
```
    mpi3::communicator third = world/3;
```
see the example in `test/communicator_split.cpp`

Which translates into:
```
world[0]: I am 'world' rank 0 and 'third' rank 0
world[1]: I am 'world' rank 1 and 'third' rank 1
world[2]: I am 'world' rank 2 and 'third' rank 2
world[3]: I am 'world' rank 3 and 'third' rank 0
world[4]: I am 'world' rank 4 and 'third' rank 1
world[5]: I am 'world' rank 5 and 'third' rank 2
world[6]: I am 'world' rank 6 and 'third' rank 0
world[7]: I am 'world' rank 7 and 'third' rank 1
```

If you don't like using math operators you can do the same with:
```
    mpi3::communicator third = world.split(world.rank()/3);
```

ye-luo:
@correaa 
The underlying dividing algorithm of comm/3 has a few issues.
1. When dividing 8 ranks in 5 gorups, the color is 0, 0, 1, 1, 2, 3, 3, 4 based on your int color = rank()*n/size() formula. I prefer to have 0,1,2,2,3,3,4,4. In QMCPACK, most dividing has fewer items in the lower end. The size of each group increases monotonically.
2. rank()*n can overflow. 100000 ranks into 100000 groups is possible.
Anyway, I will need to save the color for the group id. So using the explicit split(color) is good.

correaa:
Thanks for the feedback Ye,

On Mon, Mar 18, 2019 at 9:37 PM Ye Luo <notifications@github.com> wrote:

> @correaa <https://github.com/correaa>
> The underlying dividing algorithm of comm/3 has a few issues.
>
>    1. When dividing 8 ranks in 5 gorups, the color is 0, 0, 1, 1, 2, 3,
>    3, 4 based on your int color = rank()*n/size() formula. I prefer to have
>    0,1,2,2,3,3,4,4. In QMCPACK, most dividing has fewer items in the lower
>    end. The size of each group increases monotonically.
>
> There are many ways to near-evenly divide a communicator, so I that is one
option.
For a fine grained splitting you can use the explicit `.split(key[,
color])` function. Usually color is not necessary.

Is there a reason to have the smallest communicators to have the lower
ranks in the big communicator?
If you are aware of a general reason for that I could change the default.

Does qmcpack relies on that?


>    1. rank()*n can overflow. 100000 ranks into 100000 groups is possible.
>
> good point. I should be more careful with this convenience functons.
This is now fixed in the last commit.

>
>    1. Anyway, I will need to save the color for the group id. So using
>    the explicit split(color) is good.
>
> And where do you store(save) that color? I have to test it a bit better,
but the idea is that the communicator have an automatic naming scheme.

I changed the default to be consistent with qmcpack, but I wouldn't rely on
that.

Thanks,
Alfredo

correaa:
correction: I mean the opposite `comm.split(color[, key])`, `key` is
usually not necessary. -- A

ye-luo:
@correaa 
1. Having lower ranks with larger communicators is actually better. There is no impact in a week scaling scenario. In a strong scaling scenario, the first group can finish its work quicker with more ranks. The QMCPACK choice is flexible.
2. Sometimes the group lead (rank 0 in the split comm) needs to do a gather. All the group info is currently stored in the wrapper class Communicate. Do you have a way to split work into groups and derive a group leader communicator?

correaa:
On Tue, Mar 19, 2019 at 4:26 PM Ye Luo <notifications@github.com> wrote:

> @correaa <https://github.com/correaa>
>
>    1. Having lower ranks with larger communicators is actually better.
>    There is no impact in a week scaling scenario. In a strong scaling
>    scenario, the first group can finish its work quicker with more ranks. The
>    QMCPACK choice is flexible.
>
> So the point is that, if the same amount of work is given to the new
smaller communicatorst, and parallelization does what is supposed to do
then the lower ranks of the initial communicator will finish first.
Finishing in rank order seem to be a good property to have (for making
other assumptions), is that the reason?

BTW, this is the opposite you asked, right?

for example 8 processes divided in 5 will be 0, 0, 1, 1, 2, 2, 3, 4.
Would it be an acceptable default?


>
>
>    1. Sometimes the group lead (rank 0 in the split comm) needs to do a
>    gather. All the group info is currently stored in the wrapper class
>    Communicate. Do you have a way to split work into groups and derive a group
>    leader communicator?
>
> Sorry, I don't understand the question. Do you want to have a
subcommunicator (not a rank) that is a leader of other communicators?
Forgive my ignorance, but the only way I see doing that is by giving it a
special name for example, or the other option is to add custom MPI
attributes but that will require a certain convention.

This is not implemented, but I can do it:

```
auto third = comm/3;
if(third.leader()){ // leader() is true for the communicator with lower
ranks in comm;
   ... leader group code
}
```

Do you want this perhaps?

ye-luo:
@correaa Please please reply directly on github website instead of email. It is very hard to read.
1. 0, 0, 1, 1, 2, 2, 3, 4 is acceptable. This is opposite to what I asked for in the beginning.
2. 0, x, 1, x, 2, x, 3, 4 I want the ranks which is not marked x to form a group leader communicator.
Currently we do the group division on our own and use split to create the groups and use
comm.subcomm(nplist) to construct the group leader communicator.
nplist= {0, 2, 4, 6, 7}. So both communicators are consistent.
If I do comm/3, I need a way to generate the nplist. I don't see an easy way.

correaa:
Hi Ye, I added the capability of adding arbitrary attributes to a communicator. In particular the "color" attribute is set *after* a split operation. 

See the example here: https://gitlab.com/correaa/boost-mpi3/blob/master/test/communicator_split.cpp

```
	mpi3::communicator s = c.split(color);
        assert( mpi3::any_cast<int>(s.attribute("color")) == color );
```

correaa:
@ye-luo @markdewing 

the communicator now has three members, `.divide_low(n)` (group with small number of ranks first), `.divide_high(n)` (groups will small number of ranks last), `.divide_even(n)` (groups are all same size, some communicator will be empty).

`.operator/(int n)` defaults to `.divide_high(n)`.

for more control use `.split(color[, key])`.

All communicators created with split (directly or indirectly) have a custom `.name()` and also a custom attribute `.attribute("color")` that keeps the value passed on `split`. 
For example `.attrubute("color")==0` can be interpreted as a leader in some cases.

Should I close this issue?

ye-luo:
@correaa what I need is actually not the leader group but having the rank=0 of each group (group leaders) to form a communicator.

ye-luo:
@markdewing could you please pull mpi3 to get the "free group" bugfix in and then reopen the PR. I'd like to get this merged. Changing the splitting can be done later.

correaa:
> @correaa what I need is actually not the leader group but having the rank=0 of each group (group leaders) to form a communicator.

Ah, sorry, I didn't read carefully. Ok, how is this achieved now?

Isn't a cartesian communicator better for that?

Actually, now that I think about it is not that hard.

```
	mpi3::communicator third = world/3; // or other division
	mpi3::communicator leaders = world.keep(third.root()); // same as world.split(third.root()?0:mpi3::undefined);
```

See the example `test/communicator_split.cpp` in GitLab 
https://gitlab.com/correaa/boost-mpi3/blob/master/test/communicator_split.cpp

Also, feel free to open issues there if the conversation becomes abstracted from QMCPACK.
https://gitlab.com/correaa/boost-mpi3/issues

correaa:
BTW,  I think a 2D communicator can be simulated in the same way:

```
assert(world.size() == nrows*ncols);
auto row = world/nrows;
auto col = world.split(row.rank());
```

PR #1452, opened 2019-03-14
Update check_affinity: Ordered by MPI_rank

TApplencourt:
A race condition existed between mpi process.  This pull request fixes the issue.

Modification:
1.  "OpenMP enables..." is printed only once. 
2. For loop over MPI_rank + MPI_barrier.


markdewing:
okay to test

Issue #1449, opened 2019-03-14
Test of MPC interaction needed


prckent:
This popular feature is untested. A variant of the diamond 1x1x1 test could be made.

prckent:
Completed. Thanks @Hyeondeok-Shin . Will need to monitor, assess and possibly improve if the code fails on certain builds or the tests prove "too unreliable".

PR #1453, opened 2019-03-15
MPC Flag & MPC test is added in diamondC_1x1x1 testset, and the precision has changed to the double. 

Hyeondeok-Shin:
1. MPC flag is added in macros.cmake.

2. Deterministic tests for diamondC_1x1x1_pp were very sensitive with the machine and compiler because they were in single precision, so all deterministic tests have changed to double precision.

3. MPC test is added in deterministic-diamondC_1x1x1_pp testsets. All references for the double precision are obtained in AOS_real_fullprecision build using clang in BG/Q.


prckent:
OK to test

prckent:
@Hyeondeok-Shin Please could you add MPC checks to the regular short and long diamondC_1x1x1 in a separate PR as well? (Requires a super long run to get the reference data).

prckent:
Partially addresses #1449

PR #1457, opened 2019-03-15
Add QE6.4 patch

### LINKED ISSUE #1451: Quantum Espresso v6.4 converter
Now that QE 6.4 is available, the converter should be updated to support it.

ye-luo:
close #1451 
- [ ] review summit build recipe. @prckent could you update the manual if you find 6.4 is simpler on summit?


prckent:
Approving, but completely unchecked before #1477 is done. This won't break our 6.3 support.

PR #1460, opened 2019-03-18
Tidy afqmc unit tests

### LINKED ISSUE #1424: Updated tests for AFQMC code
This is a reminder that we need to update and (re)activate at least a few nightly integration and unit tests for the k-point AFQMC code ahead of the new release.

fdmalone:
Tidied up afqmc unit tests and enabled unit test for kpoint Hamiltonian which is currently the only Hamiltonian type tested on GPUs. Should resolve  #1424.

The Wavefunction and Propagator factories are now tested on all wavefunction + hamiltonian combinations which I achieved by adding two arguments to the catch argument parser (--hamil and --wfn). This also enables the unit tests to be run with new wavefunctions / hamiltonians more easily. The command line arguments are passed to the unit tests via the global variables UTEST_HAMIL and UTEST_WFN which are included through catch_mpi_main.hpp. This seems to be the recommended way from catch.


prckent:
OK to test

PR #1461, opened 2019-03-18
Fix AFQMC compilation with boost 1.61

ye-luo:
Just found this problem on my desktop.
I asked OLCF to enable 1.61 on rhea so I can test against the minimum version we support.


ye-luo:
Test this please

Issue #1436, opened 2019-03-12
Naming rules for cmake flags and macros


ye-luo:
I found a list of cmake flags and macros with the following pattern
BUILD_XXX
ENABLE_XXX
HAVE_XXX
QMC_XXX
I'd like a consistent way to define new ones. The existing ones should not be changed.

1. BUILD_XXX, ENABLE_XXX, QMC_XXX can are cmake flags.
2. HAVE_XXX should not be user provided cmake flags but should be flags set by CMAKE internally.
3. BUILD_XXX, ENABLE_XXX are expected to be 1/0.
4. QMC_XXX will be more flexible for strings.

Between ENABLE and BUILD, I prefer ENABLE.
If you participate other project and find their naming rule is very clear. Please comment.

ye-luo:
Years after working with CMake. One thing is clear. All QMCPACK specific options should start with `QMC_`

jtkrogel:
Outdated

PR #1462, opened 2019-03-19
Update cades build script

jtkrogel:
Due to recent changes in the CADES environment, the build script ceased to work.  This PR updates the build script to match the build recipe used in the CADES qmcpack module.


prckent:
Thank for improving this.

config/build_ornl_cades.sh
```diff
@@ -7,28 +7,36 @@
...
 # Set environment variables
 export FFTW_HOME=$FFTW_DIR/..
+export BOOST_ROOT=$BOOST_DIR
 
 CMAKE_FLAGS="-DCMAKE_C_COMPILER=mpicc \
              -DCMAKE_CXX_COMPILER=mpicxx \
              -DCMAKE_C_FLAGS=-xCOMMON-AVX512 \
-             -DCMAKE_CXX_FLAGS=-xCOMMON-AVX512"
+             -DCMAKE_CXX_FLAGS=-xCOMMON-AVX512 \
+             -DHAVE_MKL=1"
```
ye-luo: Please use -DENABLE_MKL=1 instead of -DHAVE_MKL=1.
See the logic here #1436

ye-luo: Did you find our FindMKL not working as expected even with intel compiler?

jtkrogel: Switched to -DENABLE_MKL=1.  Seems to work fine.

ye-luo: I’m still wondering why you need it. It should not be needed when intel compiler is in use.  What happens if you don’t put it.

jtkrogel: Also works fine without ENABLE_MKL.  Removing it now.

ye-luo:
Okay to test

PR #1463, opened 2019-03-20
Nexus: update for Cori

jtkrogel:
This PR contains modifications to handle the dual nature of Cori.  Cori contains both Haswell and KNL nodes which effectively comprises two different machines from the point of view of job submission.  

The updates allow the Cori supercomputer class to update its internal state relating to Haswell/KNL node anatomy when processing the job inputs prior to creating job submission scripts.   I could have instead created two classes, e.g. CoriHaswell and CoriKNL, but this would preclude submission to both partitions from within the same Nexus script as is possible with the current modifications.


ye-luo:
Okay to test

PR #1466, opened 2019-03-20
Workaround rhea issue

ye-luo:
Force the correct libstdc++ via LD_PRELOAD


ye-luo:
For some reason, the spack managed software stack on Rhea has some issue and cmake warns about it. On day 1 when the software stack of rhea is refreshed, I noticed that when building code manually but our CI doesn't trigger that. Now our CI gets a similar behaviour. I believe it needs rhea admin to fix the problem.

PR #1469, opened 2019-03-21
clang-format of QMCDrivers

PDoakORNL:


ye-luo:
@PDoakORNL Could you fix the conflict? It is bad to keep it to next week.

PR #1470, opened 2019-03-22
Add MPC test to diamondC_1x1x1_pp testset.

Hyeondeok-Shin:
MPC test is added on following inputs in diamondC_1x1x1_pp testset.

qmc_short.in.xml
qmc_long.in.xml
qmc_short_hybridrep.in.xml
qmc_short-meshf.in.xml
qmc_short_vmc_dmc.in.xml
qmc_short_vmc_dmc_allp.in.xml
qmc_long-meshf.in.xml
qmc_long_vmc_dmc_allp.in.xml
qmc_long_vmc_dmc.in.xml

References were generated in AoS/real/full-precision build using gcc6.2.0 and tests were passed in the same build in clang4.0 and intel19 compiler in the workstation.


markdewing:
okay to test

prckent:
Q. What is the cpu or gpu time cost of adding MPC to these tests? It should be trivial, but if not we might consider not adding MPC to the full test of modified tests.

PR #1474, opened 2019-03-23
Reenable afqmc rdms

fdmalone:
Allows for 1RDMs (mixed or back propagated) to be computed in AFQMC. They are currently tested through a unit test, however I will add further deterministic tests in the future.


ye-luo:
Okay to test

PR #1476, opened 2019-03-24
Print additional afqmc estimators

fdmalone:
Optionally print mixed estimates for energy components of the total energy (real part only). Helpful for debugging. I also adding padding of hdf dataset block index to avoid string sorting inconsistencies


ye-luo:
Okay to test

ye-luo:
Test this please

Issue #1473, opened 2019-03-22
Real and complex disagree in deterministic test


ye-luo:
Currently, the deterministic test reference values are generated from the real code and the complex code failed. I did an experiment by rotating the phase of the orbital. The real code do change values but the complex code doesn't. It is suspicious to me that there might be a bug in the real code.

prckent:
Which tests? Please be as specific as possible.

ye-luo:
Deterministic integration tests are not labelled correctly as deterministic. It is in the nightly category now.
https://cdash.qmcpack.org/CDash/viewTest.php?onlyfailed&buildid=49206

prckent:
I think what you are seeing is simply that these tests don't work yet. They are developmental and not production unfortunately. Fixing this needs real effort to be spent understanding the difference and then fixing them, not simply increasing the tolerance as some have suggested. Whikei would like to be wrong I suspect there is a subtle bug in at least QMCPACK's periodic spline handling.

prckent:
@jtkrogel What is the procedure for assigning the correct label to these tests?

jtkrogel:
I will label them over the next couple of days.  The labeling script needs a small update for this (originally meant to add descriptive labels to failure cases of statistical tests, now used for general labels).  

I will post instructions for adding tests to the deterministic category as well as for adding new labels in the future.

jtkrogel:
I assume we just want pattern based labeling in this case (tests with names starting with "deterministic" are labeled as such) so we don't have to update any internal lists.

prckent:
Pattern match would be simplest here and save work in future.

jtkrogel:
See #1483.  Any deterministic tests added in the future will be labeled automatically.

If new labels are desired later on, the test labeling script (`qmcpack/tests/scripts/test_labels.py`) can be easily updated in the following way.  

To add a labels for a new fancy test type, you can pattern match based on the test name as follows:
```
    if test.startswith('deterministic'):
        labels.append('deterministic')
    #end if
    if 'fancy_new_test' in test:
        labels.append('fancy')
        labels.append('new')
    #end if
```

PR #1483, opened 2019-03-25
Label deterministic tests

jtkrogel:
Update the test labeling script to label deterministic tests based on name pattern matching.

See discussion in #1473.


jtkrogel:
Note: unit tests were marked only with the label "unit" before this.  They now are marked both with "unit" and "deterministic" labels.  The tracks on CDash may need to be updated to catch the new "deterministic" label.

markdewing:
okay to test

ye-luo:
Test this please

PR #1484, opened 2019-03-26
Fix real build

fdmalone:
This PR 1) fixed several bugs/missing implementations of collinear calculations, 2) fixes the real build with/without mixed precision, 3) small bug fixes, 4) adds gpu port for some estimators.

Th unit tests for the real build are broken, since the integral files have complex integrals. We need real versions of these to get the unit tests working. CI can't enable the real build yet, since the tests are broken.

This will be sorted out in an upcoming PR.


ye-luo:
Okay to test

PR #1490, opened 2019-03-27
References for real/complex build are separated for deterministic-diamondC_2X1X1_pp testset.

Hyeondeok-Shin:
Now references for real/complex build are separated for a deterministic test for diamindC_2x1x1_pp testset in order to check consistency in each build.
 References for the complex build are generated with cplx/AoS/full-precision build using clang, and tests are passed with the same build using intel19 and gcc-6.2.0 compilers.

P.S : I'm now separately checking real vs.complex determinism in the HEG, and I'll be sharing result when they have done.


prckent:
Interesting that a separate set of checks is need for the delayed update. Ideally this would not be the case.

prckent:
OK to test

prckent:
This worked in the nightlies :-) Much cleaner results now - just the mixed precision tolerance and a very few minor afqmc and pgi issues for us to fix.

PR #1491, opened 2019-03-27
Update configs for SOA default

prckent:
This naive update to the example/reference build scripts ensures that executables are still named correctly by specifying ENABLE_SOA=0 for non-soa builds.


ye-luo:
@jtkrogel is OIC gone?

jtkrogel:
It is currently shut down.  We had expected it to be resurrected at some point, but this has not happened yet and may not.

PR #1493, opened 2019-03-27
Nexus: update version to 1.7.0

jtkrogel:
Update Nexus version to 1.7.0 ahead of QMCPACK v3.7.0 release.


markdewing:
okay to test

PR #1498, opened 2019-03-27
Nexus: updates for Gaussian Process optimization

jtkrogel:
This PR contains updates needed to run the most recent version of Rick's GP code on CADES at ORNL.


markdewing:
okay to test

PR #1501, opened 2019-03-28
Use correct test input path for short/long afqmc tests.

fdmalone:
UHF test benchmark will be updated in future due to change in wavefunction.


ye-luo:
Okay to test

Issue #1480, opened 2019-03-25
Broken mixed precision HDF5 converter tests


prckent:
Probably a trivial precision issue but should be checked carefully in case there is serious breakage. The failing tests are converter_test_HDF5_FeCO6 and converter_test_HDF5_LiH_qp. 
Besides being important to check, these  tests are the ones blocking us from having an aesthetically pleasing "all green" set of converter checks.

<img width="1549" alt="image" src="https://user-images.githubusercontent.com/13683814/54923910-9bcca900-4ee1-11e9-823a-b9da57fc3154.png">

anbenali:
So I found the reason why this is failing. 
The Atomic coordinates don't match after 10^-7 since the ParticleSet is defined as RealType and therefore will be half precision. The clean solution would be to keep the Atomic Positions as double but unfortunately we use   ParticlePos_t for both atoms and general particles (electrons etc.. ). Forcing the type to double would be risky as it propagates deep in the code. 

Fixing this can be done in 2 ways: modifying the Python comparison script to compare each group in the HDF5 file and doing a difference between (ref-test < 10^-5) instead of full HDF5 diff. This can be painful specially that it implies reading all the WF coeffs and comparing them, knowing that they are not affected. this is a heavy task as it implies having a full HDF5 reader... 
Or I can just add an HDF5 reference for mixed precision. This solution would take a few minutes to implement. 

Any thought?

ye-luo:
h5diff allows tolerance -d or -p. We probably just need an appropriate tolerance.

anbenali:
these options dont work but if I find the correct syntax it would be the best solution

anbenali:
Yes this works. 
I will push

PR #1504, opened 2019-03-28
Bug 1480 - Fix HDF5 converter tests via tolerance

anbenali:
Fixing #1480  by forcing H5diff to fail only for differences that are greater than the limit delta given as 0.000001 in the atomic positions.  This issue comes from using ParticleSet to define the atoms position and Particle set takes RealType wich can be either Float or double.


ye-luo:
Okay to test

prckent:
Happily, these worked in the nightlies.

Issue #1052, opened 2018-09-06
1 body reduced density matrix estimator produces incorrect results


jptowns:
Collecting the "dm1b" estimator on a VMC calculation of FeO with a single determinant PBE orbital basis with no Jastrow factor should produce a diagonal density matrix. This system contains 22 electrons per spin channel, so a visualization of the 1rdm for, say, the "up" channel should show a strongly diagonal structure for the first 22 states. Yet, for QMCPACK built on 04 Sept. (#1046) it does not:

![rdm_grid](https://user-images.githubusercontent.com/25334555/45186995-8c5b6c00-b1ec-11e8-85ce-ea37cf91b94c.png)

The salient input block is in the specification of the hamiltonian:
```
<?xml version="1.0"?>                                                                                                  
<qmcsystem>                                                                                                            
  <hamiltonian name="h0" type="generic" target="e">                                                                    
    <pairpot name="IonIon"   type="coulomb" source="i" target="i"/>                                                    
    <pairpot name="ElecElec" type="coulomb" source="e" target="e"/>                                                    
    <pairpot type="pseudo" name="PseudoPot" source="i" wavefunction="psi0" format="xml">                               
      <pseudo elementType="Fe" href="../wfns-and-pseudos/Fe.opt.xml"/>                                                 
      <pseudo elementType="O"  href="../wfns-and-pseudos/O.opt.xml"/>                                                  
    </pairpot>                                                                                                         
    <estimator type="dm1b" name="1rdms">                                                                               
      <parameter name="basis"            >  dm_basis     </parameter>                                                  
      <parameter name="points"           >  10           </parameter>                                                  
      <parameter name="samples"          >  10           </parameter>                                                  
      <parameter name="warmup"           >  1            </parameter>                                                  
      <parameter name="timestep"         >  0.25         </parameter>                                                  
    </estimator>                                                                                                       
  </hamiltonian>                                                                                                       
</qmcsystem>                                                                                                           
```

And the VMC calculation was specified by:
```
  <qmc method="vmc" move="pbyp" checkpoint="0">                                                                        
    <parameter name="walkers"       >       8  </parameter>                                                            
    <parameter name="timestep"      >  0.2000  </parameter>                                                            
    <parameter name="useDrift"      >      no  </parameter>                                                            
    <parameter name="steps"         >      64  </parameter>                                                            
    <parameter name="blocks"        >      64  </parameter>                                                            
[...]

Issue #995, opened 2018-08-15
1RDM cloning is broken


jtkrogel:
As revealed by #994, the 1RDM works in the current code only when 1 OMP thread is used per MPI.  This indicates that cloning (per thread deep copy) of the DensityMatrices1B class, or perhaps it's contained SPOSet, is broken.

Fixing this issue is critical for large scale production runs involving the 1RDM.

prckent:
General question: Is there a predictor for when deep copy is broken? e.g. Search for a specific call. This is unlikely to be the only broken estimator.

jtkrogel:
The only way to really tell is to compare the output from an MPI-only run and an MPI+OMP run.

ye-luo:
The AoS code requires many deep copy to handle pointers especially in the LCAO case. We should in generally move away from makeClone and rely on good copy constructors. Even if we have to use makeClone at the moment, try to use copy constructor inside.

ye-luo:
The AoS code requires many deep copy to handle pointers especially in the LCAO case. We should in generally move away from makeClone and rely on good copy constructors. Even if we have to use makeClone at the moment, try to use copy constructor inside.

jtkrogel:
I'm not sure how this is really different.  There will always be added code and interface functions needed to handle pointed to data.

# Master copy
Psi->spo = new spo
H->spo = Psi->spo

# cloning
Psi0 = Psi->makeClone() # Psi0->spo==spo0, a new spo copy
Psi1 = Psi->makeClone() # Psi1->spo==spo1

H0 = H.makeClone() # should not make a new copy of spo
H0->spo = Psi0->spo
H1 = H.makeClone()
H1->spo = Psi1.spo

I don't see how we can handle relinking cases like these (we are trying to deep copy a DAG by copying individual elements) with only pure copy constructors.

ye-luo:
If there is a pointer, an explicit copy constructor is needed otherwise just rely on compiler generated copy constructor instead of maintaining explicit makeClone everywhere. The compiler does the whole DAG.

Copy constructor of Psi:
  spo = new spo_type(*spo_in);
spo_type has compiler generated copy constructor.

@jtkrogel In your example
Psi = new Psi(spo) on each thread is actually not guaranteed to have the identical Psi data as the master if the Psi on the master was modified after the initial construction.
Using more compiler generated copyConstructor helps consistency.

jtkrogel:
This conversation really needs a whiteboard, so I will defer it until later.

[...]

PR #1509, opened 2019-03-28
Fix 1RDM cloning

jtkrogel:
Addresses #995.  The 1RDM now produces correct results for diamond (B-spline orbitals) and HEG (planewave orbitals) examples with AoS/SoA.  This was checked explicitly by observing the trace of the density matrix for 2 mpi 1 thread and 1 mpi 2 threads for diamond/HEG AoS/SoA.

The cloning issue, at least w.r.t. the current QMCPACK codebase, was traced to the use of the "normalized" variable.  Essentially, normalization was performed on the master thread (setting normalized=true) and then normalized was copied via cloning.  Since normalized was true in the clones, independent normalization was not performed on the thread copies.  This was fixed by copying the computed norm values from the master copy to the thread copies during cloning.

The need to compute the norm (via approximate Riemann integration) was removed by noting that current HEG and B-spline orbitals are normalized according to volume (integral over the cell yeilds volume instead of 1.0).  While volume normalization is assumed, the user now has the option to specify whether the orbitals are normalized to volume (volume_normed="yes") or not (volume_normed="no", norm is 1.0).

The statistical 4 mpi 4 omp tests now pass for the complex SoA and AoS codes, though still with a high intermittent failure rate.  The real code does not pass the tests because the real space orbitals are not orthogonal.  This might be the issue observed in #1052.


markdewing:
okay to test

Issue #1508, opened 2019-03-28
New Nexus test failures


ye-luo:
https://cdash.qmcpack.org/CDash/viewTest.php?onlydelta&buildid=50013

prckent:
Something broke last night.

Note that several PRs went in yesterday, before and possibly also after the nightlies started.

prckent:
Run by hand on oxygen
```
[pk7@oxygen build_intel2018_latest]$ ctest -R nexus
Test project /home/pk7/projects/qmc/git_QMCPACK_qmcpack/build_intel2018_latest
      Start 917: ntest_nexus_generic_logging
 1/16 Test #917: ntest_nexus_generic_logging ................   Passed    0.16 sec
      Start 918: ntest_nexus_generic_intrinsics
 2/16 Test #918: ntest_nexus_generic_intrinsics .............   Passed    0.10 sec
      Start 919: ntest_nexus_generic_extensions
 3/16 Test #919: ntest_nexus_generic_extensions .............   Passed    0.10 sec
      Start 920: ntest_nexus_nexus_imports
 4/16 Test #920: ntest_nexus_nexus_imports ..................   Passed    1.79 sec
      Start 921: ntest_nexus_settings
 5/16 Test #921: ntest_nexus_settings .......................   Passed    0.71 sec
      Start 922: ntest_nexus_machines
 6/16 Test #922: ntest_nexus_machines .......................   Passed    3.47 sec
      Start 923: ntest_nexus_structure
 7/16 Test #923: ntest_nexus_structure ......................   Passed    1.79 sec
      Start 924: ntest_nexus_pwscf_input
 8/16 Test #924: ntest_nexus_pwscf_input ....................   Passed    0.53 sec
      Start 925: ntest_nexus_example_gamess_H2O
 9/16 Test #925: ntest_nexus_example_gamess_H2O .............***Failed    1.15 sec
      Start 926: ntest_nexus_example_qmcpack_oxygen_dimer
10/16 Test #926: ntest_nexus_example_qmcpack_oxygen_dimer ...***Failed    0.80 sec
      Start 927: ntest_nexus_example_qmcpack_H2O
11/16 Test #927: ntest_nexus_example_qmcpack_H2O ............***Failed    0.78 sec
      Start 928: ntest_nexus_example_qmcpack_LiH
12/16 Test #928: ntest_nexus_example_qmcpack_LiH ............***Failed    0.84 sec
[...]

PR #1510, opened 2019-03-28
Nexus: fixes for ctest

jtkrogel:
This PR attempts to address #1508.  

When driving the Nexus test script from ctest, there was a missed change from nexus/library to nexus/lib.  Ctest would set PYTHONPATH to the former, which remains an existing path if a repository has been updated in place over time (nexus/library was renamed nexus/lib in a prior PR).  The resulting tests therefore used an older copy of Nexus (contained in nexus/library) to run the current tests, which recently had reference files updated, causing the failures.  Updating the ctest pythonpath to point to nexus/lib resolves the issue.


markdewing:
okay to test

PR #1512, opened 2019-03-28
clang-format src files, PETE, all Particle, Platforms

PDoakORNL:


ye-luo:
Can we exlcude PETE? It is from another library.

PDoakORNL:
I can revert that

PR #1516, opened 2019-03-29
clang-format QMCTools

PDoakORNL:


ye-luo:
Have you checked ppconvert build by hand?

PDoakORNL:
Good idea

PDoakORNL:
ppconvert builds with gcc 6.4.0 and does not crash.

prckent:
Should be OK to merge now that we have done the release.

Issue #1440, opened 2019-03-13
DMC LocalECP incorrect in GPU code on titan using CUDA 9.1


jtkrogel:
Disagreement between CPU and GPU DMC total energies was observed for a water molecule in periodic boundary conditions (8 A cubic cell, CASINO pseudopotentials, Titan at OLCF, QMCPACK 3.6.0).  Issue originally reported by Andrea Zen.  Original inputs and outputs: [TEST_DMC.zip](https://github.com/QMCPACK/qmcpack/files/2961648/TEST_DMC.zip)

From the attached outputs, the VMC energies agree, while the DMC energies differ by about 0.3 Ha:
```
#VMC
>qmca -q e *s001*scalar*
dmc_cpu  series 1  LocalEnergy           =  -17.183577 +/- 0.007486 
dmc_gpu  series 1  LocalEnergy           =  -17.152789 +/- 0.018592 

#DMC
>qmca -q e *s002*scalar*
dmc_cpu  series 2  LocalEnergy           =  -17.220971 +/- 0.000968 
dmc_gpu  series 2  LocalEnergy           =  -16.869061 +/- 0.003256 
```

The difference is entirely attributable to the local part of the ECP:
```
#DMC
>qmca -q l *s002*scalar*
dmc_cpu  series 2  LocalECP              =  -41.436580 +/- 0.021199 
dmc_gpu  series 2  LocalECP              =  -41.026695 +/- 0.028982 
```

Note: the DMC error bars are not statistically meaningful here (10 blocks), but the difference is large enough to support this conclusion.

The oddity here is that the error is only seen in DMC and it is limited to a single potential energy term.  This may indicate a bug in LocalECP that surfaces with increased walker count on the GPU (1 walker/gpu in VMC, 320 walkers/gpu in DMC).  Likely, a series of VMC runs with increasing number of walkers will show this.

prckent:
The local ECP kenel is one that is known to not be reproducibile between runs, i.e. is buggy. Something to do with walker and GPU thread/block count. Previously the differences have been small enough to be ignorable; this problem indicates it must be fixed.

You don't state explicitly, but is the non-local ECP term correct?

jtkrogel:
The non-local ECP term appears to be correct.

prckent:
To save time debugging this, for the next 3 weeks the necessary pwscf file is at
<https://ftp.ornl.gov/filedownload?ftp=e;dir=WATER>
[...]

PR #1518, opened 2019-03-29
Fix parseCasino

### LINKED ISSUE #1511: Segmentation fault when using Casino PP
It should be simple to fix.

ye-luo:
Closes #1511


prckent:
It occurs to me that a bug like this could be responsible for #1440

PR #1534, opened 2019-04-04
Cache distance table description for later output

markdewing:
Currently the distance table output occurs during the middle of wavefunction output.
Store the description and output it later during the ParticleSetPool output.

### DIFF
--- a/src/Particle/DistanceTableData.h
+++ b/src/Particle/DistanceTableData.h
@@ -398,6 +398,15 @@ struct DistanceTableData
       Temp.resize(N[SourceIndex]);
     }
   }
+
+  /// Description generated during creation, to be printed later
+  std::ostringstream description;
+
+  void output_description(std::ostream& os)
+  {
+    os << std::endl;
+    os << description.str();
+  }
 };
 } // namespace qmcplusplus
 #endif
--- a/src/Particle/DistanceTableData.h
+++ b/src/Particle/DistanceTableData.h
@@ -399,14 +399,6 @@ struct DistanceTableData
     }
   }
 
-  /// Description generated during creation, to be printed later
-  std::ostringstream description;
-
-  void output_description(std::ostream& os)
-  {
-    os << std::endl;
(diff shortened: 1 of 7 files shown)


src/Particle/DistanceTableData.h
```diff
@@ -398,6 +398,15 @@ struct DistanceTableData
       Temp.resize(N[SourceIndex]);
     }
   }
+
+  /// Description generated during creation, to be printed later
+  std::ostringstream description;
```
ye-luo: Hopefully in the near future, we will change most objects of DistanceTableData to private.
I'm thinking of keeping the class light.
On the other hand, this printout doesn't need to be kept the whole life time of DistanceTableData.
For this reason, is it better if we expand the createDistanceTable by introducing 1st argument a stream and print to the stream. The stream is stored in ParticleSet.

src/Particle/ParticleSet.h
```diff
@@ -160,6 +160,9 @@ class ParticleSet : public QMCTraits, public OhmmsElementBase, public PtclOnLatt
   ///distance tables that need to be updated by moving this ParticleSet
   std::vector<DistanceTableData*> DistTables;
 
+  /// Descriptions from distance table creation.  Same order as DistTables.
+  std::vector<std::string> distTableDescriptions;
```
ye-luo: What is your consideration of keep one for each table? I'm naively thinking just one description for all tables.

markdewing: That requires changing how the description is generated.  I'm trying to make the minimal amount of changes to this code (for now).

ye-luo: I see. I did a quick try and hit some issues. Basically, we need to improve the use of C++. AoS is the problem again.

prckent: Agree. The AoS/SoA stuff should either totally go (simpler) or we should use some C++ trickery such that there are no switches and everything always works.

Issue #1533, opened 2019-04-04
evaluateHessian needed for J1/J2 SoA


jtkrogel:
This is needed to support both Stress evaluation and L2 potentials.  Both are currently broken in SoA.  This issue was listed in #861, but did not have its own issue until now.

ye-luo:
@jtkrogel close and update 861?

jtkrogel:
Agreed.  Closing this issue as resolved by #1535.

PR #1535, opened 2019-04-04
SoA implementation of J1/J2 evaluateHessian

jtkrogel:
Addresses #1533 as part of #861.

The new SoA implementation was checked against AoS deterministically using a modified version of the NiO L2 statistical integration test.  Checks were performed without Jastrow (worked already), J1 only, and J1+J2, see below.
```
>qmca -e 0 -q e 0*J0/*scalar*
01_aos_deterministic_J0/NiO-fcc-S1-vmc  series 1  LocalEnergy  =  -368.208074 +/- 2.183875 
04_soa_deterministic_J0/NiO-fcc-S1-vmc  series 1  LocalEnergy  =  -368.208074 +/- 2.183875 

>qmca -e 0 -q e 0*J1/*scalar*
02_aos_deterministic_J1/NiO-fcc-S1-vmc  series 1  LocalEnergy  =  -323.706678 +/- 11.355735 
05_soa_deterministic_J1/NiO-fcc-S1-vmc  series 1  LocalEnergy  =  -323.706678 +/- 11.355735 

>qmca -e 0 -q e 0*J2/*scalar*
03_aos_deterministic_J2/NiO-fcc-S1-vmc  series 1  LocalEnergy  =  -371.940356 +/- 0.936515 
06_soa_deterministic_J2/NiO-fcc-S1-vmc  series 1  LocalEnergy  =  -371.940356 +/- 0.936515 
```

### DIFF
--- a/src/QMCWaveFunctions/Jastrow/J1OrbitalSoA.h
+++ b/src/QMCWaveFunctions/Jastrow/J1OrbitalSoA.h
@@ -124,6 +124,35 @@ struct J1OrbitalSoA : public WaveFunctionComponent
     return LogValue;
   }
 
+  void evaluateHessian(ParticleSet& P, HessVector_t& grad_grad_psi)
+  {
+    const DistanceTableData& d_ie(*(P.DistTables[myTableID]));
+    valT dudr, d2udr2;
+
+    Tensor<valT, DIM> ident;
+    grad_grad_psi = 0.0;
+    ident.diagonal(1.0);
+
+    for (int iel = 0; iel < Nelec; ++iel)
+    {
+      const valT* dist          = d_ie.Distances[iel];
+      const RowContainer& displ = d_ie.Displacements[iel];
+      for (int iat = 0; iat < Nions; iat++)
+      {
+        int gid    = Ions.GroupID[iat];
+        auto* func = F[gid];
+        if( func != nullptr)
+        {
+           RealType r    = dist[iat];
+           RealType rinv = 1.0 / r;
+           PosType dr    = displ[iat];
+           func->evaluate(r, dudr, d2udr2);
+           grad_grad_psi[iel] -= rinv * rinv * outerProduct(dr, dr) * (d2udr2 - dudr * rinv) + ident * dudr * rinv;
+        }
+      }
+    }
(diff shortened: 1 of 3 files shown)


ye-luo:
Could you expand the unit tests?
src/QMCWaveFunctions/tests/test_bspline_jastrow.cpp

jtkrogel:
Make an issue for further work?  Adding a deterministic test would be less work than the unit test.  Unit tests use data from finite differences, which likely originate from a script I don't have (@markdewing ?).

markdewing:
The values for the unit tests come from gen_bspline_jastrow.py.   The derivatives are computed in the function 'gen_bspline_jastrow' (around line 120).

ye-luo:
It's not necessary to do finite difference. Just do the minimal.
For the moment, there is no need to check the functor.
Just check J1 and J2.evaluateHessian call.

markdewing:
I was looking at the wrong part of the unit tests - I see some of it comes from finite differences, and I don't know how those were computed.

ye-luo:
Just like https://github.com/QMCPACK/qmcpack/blob/c985be2bdf0ef04b179c4e6d4ac7645891e61203/src/QMCWaveFunctions/tests/test_bspline_jastrow.cpp#L140

prckent:
I agree with Ye that unit testing the current value would be fine imo. Then at least we'll catch a change or compiler optimization bug that breaks the code.

jtkrogel:
Very well.  Will get to this in time.

jtkrogel:
Done.  Unit tests confirmed to pass for AoS/SoA real/complex.

As an item for future investigation, I noticed that evaluateHessian has side effects.  For example, sometimes if evaluateLog is not called after evaluateHessian some other unit test would fail.  This is likely due to an incompatible change of internal state.

markdewing:
okay to test

ye-luo:
@jtkrogel I just commented the evaluateLog after evaluateHessian. The test passes. I tried both gcc7.3 on AMD ryzen and intel 18.3 on Xeon. Could you tried again on the develop branch by commenting the evaluateLog line? What is your testing platform?

PR #1538, opened 2019-04-08
Nexus: updates for Quantum Package 2.0

jtkrogel:
Update Nexus to interact with version 2.0 of Quantum Package.  This was a major change in the QP source and user input.

Also add a user example for Quantum Package: running Hartree-Fock on a water molecule.


grahamlopez:
Okay to test

PR #1544, opened 2019-04-11
Nexus: symmetrize kpoint grids with spglib

jtkrogel:
In this PR, support is added to automatically find irreducible supercell k-points with spglib.  This feature will be useful in the future to support irreducible twist averaging.


markdewing:
okay to test

PR #1547, opened 2019-04-14
Remove workaround in SplineC2ROMP

ye-luo:
The allocator bug has been fixed when fixing the unit test before the v3.7 release.


ye-luo:
GCC crashed in CI.

ye-luo:
Test this please

PR #1548, opened 2019-04-15
Enable MultiSlaterDeterminantFast class to work with complex CI coefficients

chrislzhao:
@ye-luo This is the first PR of a sequence of PRs, which eventually enables LM optimizer to optimize complex wave function variables. In the current PR, the multiSlaterDeterminantFast class is modified so that it could take complex CI coefficients from input and compute its energy.


markdewing:
okay to test

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -717,7 +721,11 @@ bool SlaterDetBuilder::createMSDFast(MultiSlaterDeterminantFast* multiSD, xmlNod
       {
         //std::stringstream sstr;
         //sstr << "CIcoeff" << "_" << i;
-        multiSD->myVars->insert(CItags[i], (*(multiSD->CSFcoeff))[i], true, optimize::LINEAR_P);
+        #ifdef QMC_COMPLEX
+          multiSD->myVars->insert(CItags[i], std::real((*(multiSD->CSFcoeff))[i]), true, optimize::LINEAR_P);
```
ye-luo: I guess this is just a workaround to get the code building.
Could you add a note
```
// TODO: to be fixed by the complex VariableSet.
```
Could you add such TODO notes in all the places needing a fix later.

ye-luo:
What kind of test do you think is good for checking this case?
LiH with 2 dets? We do need separated tests for real and complex because the xml input files are different due to complex.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
chrislzhao: @ye-luo I have added the note.

ye-luo: Is this the only place? Or you have others in mind. Please do all places.

chrislzhao: Should be done by now.

chrislzhao:
@ye-luo I think LiH with 2 Dets is a good test.

ye-luo:
Could you set up the test?
VMC LiH with 2 dets and Jastrow. Jastrow is optimized for single det. The weight of the two dets can be a bit artificial. Based on your previous experience, please set the starting weights not too good because we will add optimization test on top of it and not too bad because we don't want crazy starting point for optimization.

chrislzhao:
@ye-luo For the tests, all we need now is to get the VMC energy right, correct? Since the optimization is presumably not working for complex CI coefficients.

ye-luo:
Correct. We only need tests for VMC at the moment.
If possible, please create a LiH test based on the existing LiH_solid_1x1x1_pp.
Please reuse LiH-gamma.pwscf.h5. So we don't need to add more big h5 files into the repo.

chrislzhao:
@ye-luo I have added the tests.

tests/solids/LiH_solid_1x1x1_pp/md_rc_vmc_LiH-gamma.xml
```diff
@@ -0,0 +1,112 @@
...
+  <wavefunction name="psi0" target="e">
+    <determinantset type="einspline" href="LiH-gamma.pwscf.h5" tilematrix="1 0 0 0 1 0 0 0 1" twistnum="0" gpu="yes" meshfactor="1.0" source="i"  precision="single">
+      <comment text="create the spline-based basis set builder" />
+      <basisset name="spline-basis" type="einspline" />
+
+      <comment text="create the spline-based single particle orbital sets" />
+      <sposet name="spo-spline-fixed" basisset="spline-basis" size="6" optimizable="no" />
+      <sposet name="spo-up"    basisset="spo-spline-fixed" size="6" optimizable="no" />
+      <sposet name="spo-dn"    basisset="spo-spline-fixed" size="6" optimizable="no" />
+
```
ye-luo: @chrislzhao before you start, please pull the most recent change.
Could you please remove all the basisset. I understand that you need that for optimization but we will not go through the AoS routine but only the SoA route via LCAOSet which already has rotation support. For this reason, how to setup the basisset input will be different. In this PR, we need to minimal input that supports running MSD solid.

ye-luo:
In addition, we need to save a "very long" reference run and also have long tests.

chrislzhao:
@ye-luo Do I need to add both CPU and GPU tests?

ye-luo:
@chrislzhao no need to create GPU tests, the CUDA code is not interoperable with host determinants.

PR #1549, opened 2019-04-15
LCAO SoA Implementation of Hessian and Gradient Hessian

rcclay:
This PR implements the machinery for computation of the SPO hessian and gradient hessian for LCAO type SPO's with the SoA code version.  Immediate effects of this are:
1.)  Backflow local energies can now be evaluated with LCAO's, and 
2.)  Backflow optimization should work with LCAO's now.  

This is also a necessary prerequisite for ionic derivatives of LCAO type SPO's, which will be the next PR.


markdewing:
okay to test

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -256,6 +256,18 @@ class SPOSet: public QMCTraits
...
+   * @param P current ParticleSet
+   * @param iat active particle
+   * @param psi values of the SPO
+   * @param dpsi gradients of the SPO
+   * @param grad_grad_psi hessians of the SPO
+   * @param grad_grad_grad_psi grad hessians of the SPO
+   */
+  virtual void
+  evaluate(const ParticleSet& P, int iat,
+           ValueVector_t& psi, GradVector_t& dpsi, HessVector_t& grad_grad_psi, GGGVector_t& grad_grad_grad_psi)=0;
```
ye-luo: I think it is better to define APP_ABORT as the base class implementation instead of implementing every APP_ABORT at every place.

src/QMCWaveFunctions/tests/test_dirac_det.cpp
```diff
@@ -168,6 +175,20 @@ void FakeSPO::evaluate(const ParticleSet& P,
     psi[i] = a(iat, i);
   }
 }
+
+void FakeSPO::evaluate(const ParticleSet& P,
```
ye-luo: Is this function  been called?

src/QMCWaveFunctions/SPOSet.h
rcclay: Sounds reasonable.  I'll do that instead.

src/QMCWaveFunctions/tests/test_dirac_det.cpp
rcclay: It's not called.  This is a side effect of having this function pure virtual in the base class, and therefore having to have an instantiation for each derived class.  This can go away when I implement the suggestion you made previously.

src/QMCWaveFunctions/SPOSet.h
ye-luo: Could you also take care of the above function? It think it has the same situation.

rcclay:
@ye-luo Made the changes you suggested.  Let me know if this is satisfactory.

rcclay:
Failure to fetch.  That's a new one.

markdewing:
The failure to fetch in the CI is transient - retriggering a build works around it.  This is becoming annoyingly frequent.

markdewing:
test this please

rcclay:
Test this please.

rcclay:
Oops.  Sorry!

src/QMCWaveFunctions/tests/test_MO.cpp
```diff
@@ -144,6 +144,57 @@ void test_He(bool transform)
     REQUIRE(dpsi[0][2] == Approx(0));
     REQUIRE(d2psi[0] == Approx(-0.2618497452));
 
+    //==========Hessian and Grad Hessian Test==========
+    /*
+    // Generated from gen_mo.py for position [0.1, 0.0, 0.0]
+    REQUIRE(values[0] == Approx(   0.9671159429));
```
ye-luo: Why are these tests commented?

rcclay: I generated the test data at the beginning of the project, but as I went on, I felt like the HCN example was a lot more thorough/discriminatory than the Ne or He tests, so I didn't enable the tests for those.

ye-luo: Does it hurt anything if we enable tests?
I think it is better to have them enabled or deleted not commented.

rcclay: It doesn't hurt anything except add quite a bit to the source. I might delete these, since I imagine I have other more important tests to add soon.

PR #1550, opened 2019-04-17
Drift modifiers

ye-luo:
template function getScaledDrift is used to apply limits to DMC drift.
In order to include more modification schemes, a new abstraction layer is introduced to allow expansion in the future.
DriftModifierBase, DriftModifierUNR
This PR also allows handing parameters of each schemes.
`<parameter name="drift_UNR_a"         >    1.0   </parameter>`


jtkrogel:
I prefer "drift_UNR_a".  The "modifier" part is a bit redundant with providing a parameter from the user point of view (even though it reflects association with code buried deeper).  Simpler is better.

ye-luo:
Do not merge. I will adjust the PR based on the review comments.

ye-luo:
@rcclay I thought about moving branchWeight stuff from SimpleFixedNodeBranch. it takes a bit effort to clean up the code and I will do it in a later PR.
@jtkrogel @rcclay good name suggestions. I will change them.

PR #1553, opened 2019-04-18
More smooth functions and schemes for hybrid orbital representation

ye-luo:
Make it easier to customize smooth functions.
No change to the default. Our manual has been updated as well.


src/QMCWaveFunctions/BsplineFactory/HybridAdoptorBase.h
```diff
@@ -416,7 +418,16 @@ struct HybridAdoptorBase
...
+  RealType f, df_dr, d2f_dr2;
+  /// smoothing schemes
+  enum
+  {
+    CONSISTENT = 0,
+    SMOOTHALL,
+    SMOOTHPARTIAL
+  } smooth_scheme;
+  /// smoothing function
+  int smooth_func_id;
```
PDoakORNL: I will not approve this kind throw away of enum variable, especially as it is set far from here and used in place of what should be a proper functor or templated function. And its actually part of the "state"/recipe of the spline evaluation but its hidden here.

src/Numerics/SmoothFunctions.hpp
```diff
@@ -0,0 +1,125 @@
...
+#ifndef QMCPLUSPLUS_SMOOTH_FUNCTIONS_H
+#define QMCPLUSPLUS_SMOOTH_FUNCTIONS_H
+
+#include <cmath>
+
+namespace qmcplusplus
+{
+struct SmoothFunctions
+{
+  enum
```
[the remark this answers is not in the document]
ye-luo: Actually it should be LEKS2018 based on the paper authors.
Regarding doing a cpp file, for function at this size, it is probably not necessary.
The other drawback of cpp file is that a compilation time error becomes link time error if someone put complex type in it.

src/QMCWaveFunctions/BsplineFactory/HybridAdoptorBase.h
```diff
@@ -630,18 +641,68 @@ struct HybridAdoptorBase
...
-    const RealType scale  = ctwo / (cutoff - cutoff_buffer);
-    const RealType x      = (r - cutoff_buffer) * scale - cone;
-    const RealType cosh_x = std::cosh(x);
-    const RealType tanh_x = std::tanh(x);
-    df_dr                 = -chalf / (cosh_x * cosh_x) * scale;
-    d2f_dr2               = -ctwo * tanh_x * df_dr * scale;
-    return chalf * (cone - tanh_x);
+    const RealType scale  = cone / (cutoff - cutoff_buffer);
+    const RealType x      = (r - cutoff_buffer) * scale;
+    if(smooth_func_id == SmoothFunctions::YE2018)
```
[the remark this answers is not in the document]
ye-luo: I took your first suggestion. I think we can't put every parameter on templates.

ye-luo:
I replaced all the enum with named enum and put the definition close to its use.

ye-luo:
@PDoakORNL is my last fix enough?

src/QMCWaveFunctions/BsplineFactory/HybridAdoptorBase.h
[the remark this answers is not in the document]
PDoakORNL: This is fine, for this sort of branching its mostly just style.

src/Numerics/SmoothFunctions.hpp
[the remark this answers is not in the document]
PDoakORNL: That link time error makes it clear the author assumed that type would not be used in the template.
There are a great many headers like this in the code they really add up. Why add more.

And we assume complex basis are not used so either we should make the intent clear with:
```c++
#include <type_traits>
...
static_assert(std::is_integral<T>::value, "Integral required.");
```

PDoakORNL:
The code changes are enough, I'm checking just how hard explicitly instantiating and linking the templates would be.

ye-luo:
I'm not sure it is worth putting more options as template parameters.
Neither a lot of template parameters nor a lot of state machine is good anyway.
The smoothing_scheme will not expand in the future.
smoothing_function is more valuable to because it may grow.

PDoakORNL:
I didn't add any template parameters just put the functions into a cpp.  Explicit instantiation brought up sincos has no overide for float it has a different name.  That is fixed as well.

src/Numerics/SmoothFunctions.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "Numerics/SmoothFunctions.hpp"
+
+namespace qmcplusplus
+{
+
+/** wraps untemplated cmath function
+ *  unless wall clock is actually effected considering dropping cmath dependence 
+ */
+template<typename T>
+void sincosT(T x, T* y, T* z);
```
ye-luo: sincos is managed at
config/stdlib/math.h

ye-luo:
@PDoakORNL thank you. I was thinking about other things.

src/Numerics/SmoothFunctions.cpp
PDoakORNL: What happens if I want to build both float and double?

ye-luo: some platform doesn't have sincosf or a slow sincosf.

ye-luo: Actually I don't understand your question. We always build both float and double. The config/stdlib/math.h makes sure sincos for float and double can always be called.

PDoakORNL: Whoa I'd forgotten that file, the headers lead me to believe we were really using cmath sincos.  I don't have a great feelling about this sort of technique.  libc calls should just be the libc calls.

ye-luo: That file used to handle more math functions. Now C++11 ships more and we can do less.
Probably it is time to change the name. I just don't like a separate implementation floating.

PDoakORNL: But it does look like we can't quite, I will point at config/stdlib/math.h to avoid the evil of two different overrides.

PR #1556, opened 2019-04-19
Fix clang warnings in GreenFunctionModifiers

markdewing:
In addition to issuing a warning at compile time, clang causes the code to abort if the function takes a return value but no return statement is present.

Also add a virtual destructor to the base class to address another clang warning.


ye-luo:
I have another bug in it. hold a second.

PR #1559, opened 2019-04-22
Implement Ionic Derivatives of LCAO-type SPO's for SoA

rcclay:
This implements all analytic ion derivatives of the LCAO-type single-particle orbitals required for ZVZB force estimation--in particular, ionic derivatives of the SPO, SPO electron gradients, and SPO electron laplacians.  I've also added unit tests of these new API's, with reference values generated from finite differences.  This code is included in src/QMCWaveFunction/tests/gen_mo.py.


rcclay:
Ok to test.

PR #1560, opened 2019-04-22
[WIP - Cosmetic level. functionality implemented] Evaluation of Complex Wavefunctiuon in LCAO.

anbenali:
Evaluation of Complex wavefunction for solids. (case of Kpoints)

- All input are read in complex.
- All data is complex and bcasted correctly properly accross the code
- The Tiling of the cell is done outside QMCPACK in the Pyscf converter. 
- Pyscf converter now stores one real matrix and one complex matrix stored in: 
 eigenstate_0 
 eigenstate_0_imag

-Modification of the hierarchy in the HDF5 where /KPOINT_0/ becomes /Super_Twist/
 Backwards compatibility is preserved for real wavefunction. 

- Create a function to compute all phases in one shot and store them. Evaluated in builder (LCAOrbitalBuilder::EvalPhaseFactor). 

- Phase Factor applied in SoaAtomicBasisSet.h 

TO DO (these are cosmetic and verification oriented) :

- Clean Converter generator. 
- Create tests
- Verify that the phase factor applies to first derivatives AND second Derivatives. @lshulen / @jtkrogel ? can you confirm? 

- Modify functions: LoadCMatrix and LoadCMatrix_cplx to become only one templated function with some specialization for cplx.  @markdewing ?
- Modify function: LCAOrbitalBuilder::EvalPhaseFactor to have specialization for Complex.  @markdewing  ?

### DIFF
--- a/src/QMCTools/LCAOH5Parser.cpp
+++ b/src/QMCTools/LCAOH5Parser.cpp
@@ -63,7 +63,7 @@ void LCAOParser::parse(const std::string& fname)
   hin.read(PBC, "PBC");
   hin.pop();
 
-  std::cout << "Periodic Boundary Comditions: " << (PBC ? ("yes") : ("no")) << std::endl;
+  std::cout << "Periodic Boundary Conditions: " << (PBC ? ("yes") : ("no")) << std::endl;
   std::cout.flush();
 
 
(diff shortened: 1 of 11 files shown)


rcclay:
@anbenali Why wouldn't the phase factor apply to first and second derivatives?

anbenali:
This PR is obsolete and was replaced by #1578 until we understand how to apply the phase factor properly.

PR #1561, opened 2019-04-24
Revive Derivative of Determinants

rcclay:
Looks like there's a legacy implementation of force derivatives for determinants which seems to work now that the LCAO ionic gradients are in place.  This PR does the following:

1.)  Adds a "prepareIonDerivs()" function to TrialWaveFunction and WaveFunctionComponent classes.  This does nothing by default, but for DiracDeterminant, it resizes all the temporary arrays needed for force computation.  This should be called before any force evaluations are done.  For fast force evaluation in the future, behavior for this will be nontrivial.  

2.)  Set "ionDerivs" SPOSet class member to True if SPOSet has an explicit ion dependence.  Determinant math only done if ionDerivs=True.  Otherwise it's bypassed.

No unit test for this yet, but WaveFunctionTester on the H2O molecule example without a 3-body jastrow agrees with finite difference results.


jtkrogel:
Is the legacy code part of Ken's earlier force efforts?

rcclay:
@jtkrogel It is!  This is going to scale at least like O(N^4) to compute all forces, but from what I can determine, it's at least formally correct.

rcclay:
Don't merge yet.  Chasing down a weird bug.

rcclay:
After some digging, the bug I found is unrelated to this PR and definitely pre-exists it.  Everything works as it should on my sodium test problem.  
 
Bug manifests as a crash on line 623 QMCHamiltonian.cpp, in function QMCHamiltonian::evaluateIonDerivs, which is called by the ACForce estimator (src/QMCHamiltonians/ACForce.cpp:116) from auxH evaluate in initWalkersForPbyP in the VMC driver.  The test system is the H2O molecule found in tests/molecules/H2O_dimer_sep_pp/ with the 3 body jastrow removed.  I'll dig in later.

src/QMCWaveFunctions/Fermion/DiracDeterminant.h
```diff
@@ -97,6 +97,9 @@ class DiracDeterminant : public DiracDeterminantBase
 
   ValueType ratioGrad(ParticleSet& P, int iat, GradType& grad_iat);
   GradType evalGrad(ParticleSet& P, int iat);
+
+  void prepareIonDerivs();
```
ye-luo: This function is expected to be private and only called within DiracDeterminant class.
Calling from outside and passing through all wavefunction hierarchy is bad.
Also it is better to call it resizeScratchObjectsForIonDerivs.

src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
```diff
@@ -459,96 +467,96 @@ typename DiracDeterminant<DU_TYPE>::GradType DiracDeterminant<DU_TYPE>::evalGrad
...
-    {
-      lapl_phi_Minv(i, j) = 0.0;
-      for (int k = 0; k < NumOrbitals; k++)
-        lapl_phi_Minv(i, j) += d2psiM(i, k) * psiM(j, k);
-    }
-  for (int dim = 0; dim < OHMMS_DIM; dim++)
+  GradType gradPsi(0.0);
+  if(Phi->hasIonDerivs())
   {
+    Phi->evaluateGradSource(P,
```
ye-luo: call resizeScratchObjectsForIonDerivs() before this line.

src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
```diff
@@ -306,16 +296,32 @@ void DiracDeterminant<DU_TYPE>::evaluateRatiosAlltoOne(ParticleSet& P, std::vect
...
 {
-  if (!ionDerivs)
-    APP_ABORT("template<typename DU_TYPE>DiracDeterminant<DU_TYPE>::evalGradSource.  Determinant not initialized for "
-              "force computations.");
-  Phi->evaluateGradSource(P, FirstIndex, LastIndex, source, iat, grad_source_psiM);
-  return simd::dot(psiM.data(), grad_source_psiM.data(), psiM.size());
+  GradType g(0.0);
+  if(Phi->hasIonDerivs())
+  {
+    Phi->evaluateGradSource(P, FirstIndex, LastIndex, source, iat, grad_source_psiM);
```
ye-luo: Call resizeScratchObjectsForIonDerivs before this line.

src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
```diff
@@ -326,103 +332,105 @@ typename DiracDeterminant<DU_TYPE>::GradType DiracDeterminant<DU_TYPE>::evalGrad
...
-                          source,
-                          iat,
-                          grad_source_psiM,
-                          grad_grad_source_psiM,
-                          grad_lapl_source_psiM);
-  Phi->evaluate_notranspose(P, FirstIndex, LastIndex, psiM_temp, dpsiM, d2psiM);
+  GradType g(0.0);
+  if(Phi->hasIonDerivs())
+  {
+    Phi->evaluateGradSource(P,
```
ye-luo: Call resizeScratchObjectsForIonDerivs before this line.

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -139,7 +139,10 @@ class SPOSet: public QMCTraits
     return OrbitalSetSize;
   }
 
-
+  inline bool hasIonDerivs() const
```
ye-luo: This is a good move, please mark ionDerivs protected.
```
  ///flag to calculate ionic derivatives
  bool ionDerivs;
```
I think the meaning of this flag should be
```
  ///true, if the derived class has non-zero ionic derivatives.
```
Please add a consistent comment of this function as well.

src/QMCWaveFunctions/WaveFunctionComponent.h
```diff
@@ -250,6 +250,11 @@ struct WaveFunctionComponent: public QMCTraits
     return GradType();
   }
 
+  /** For given particle set configuration, initialize whatever intermediate objects are required
+  * for force evaluation (preferably fast).  Default does nothing.*/ 
+  virtual void prepareIonDerivs()
```
ye-luo: Resizing is only needed locally. Please remove all prepareIonDerivs but DiracDeterminant.

src/QMCWaveFunctions/Fermion/DiracDeterminant.h
rcclay: This function is where I got hamstrung by the current API.  Right now, the evaluateGradSource functions take an ion index and return a 3d gradient associated with that ion's ionic derivatives.  So I could pull specific force gradients out, or I could compute the gradients for all N ions.  If I want to do the latter, I want to allocate and populate temporary arrays once at the beginning of all these computations if the electron particle set has changed.  Having this function here was in anticipation of that, although I could probably just add a function "evaluateGradSource" function that computes ALL the force derivatives, and does the allocation within itself.  I think I might do that...

src/QMCWaveFunctions/SPOSet.h
rcclay: Will do!  Thanks!

src/QMCWaveFunctions/Fermion/DiracDeterminant.h
ye-luo: memory allocation cost is paid only once even if you call resize infinite times. That is why I insist moving it into the evaluation function. Is it a concern for you?

My understanding is that evaluateGradSource should compute ALL force derivatives. It doesn't make sense to me just do one and it stops us writing very efficient codes.
Is evalGradSourcep used? If not, could you delete it?

src/QMCWaveFunctions/SPOSet.h
ye-luo: Please delete the setIonDerivs.

rcclay:
@ye-luo  I think I addressed everything you brought up.  Does it look OK now?

src/QMCWaveFunctions/Fermion/DiracDeterminant.h
rcclay: The resize itself is not a problem.  It's if I have to do stuff other than resize.  But I think adding another function is the right way to handle this case.  Your suggestion to add a resize to the current function is the best solution.

src/QMCWaveFunctions/WaveFunctionComponent.h
```diff
@@ -165,11 +165,6 @@ struct WaveFunctionComponent: public QMCTraits
     Optimizable = optimizeit;
   }
 
-  inline void setIonDerivs(bool calcionderiv)
-  {
-    ionDerivs=calcionderiv;
```
ye-luo: Do we still need ionDerivs in WaveFunctionComponent.h? If not, please remove it.
If there is any unused ionDerivs, please remove them.

rcclay: Done.

rcclay:
test this please

PR #1562, opened 2019-04-24
Nexus: fix jastrow dependency for unreweightedvariance

jtkrogel:
Quick fix for processing of jastrow dependencies following optimization when using only unreweighted variance.


markdewing:
okay to test

PR #1567, opened 2019-04-26
Update nightly

ye-luo:
Update the script for bora
Add a nightly job script for summit


ye-luo:
Q1. Running 1 build with almost all the tests takes about 2h. Different build needs to be on different jobs.
Q2. Correct. CUDA only supports Intel 18.

PR #1568, opened 2019-04-26
Tidy energy print and print total energy of starting determinant.

fdmalone:
Very minor formatting change for energy printout of starting determinant so I don't have to add the numbers together myself.


markdewing:
okay to test

PR #1569, opened 2019-04-29
Editing pass on Sections 1-5 of manual

prckent:
Edits from Charlie Horak (ORNL) for style and consistency plus bibliography fix.

### DIFF
--- a/manual/features.tex
+++ b/manual/features.tex
@@ -17,7 +17,7 @@ available or can be quickly  brought to the full production level.
 \item Excited state calculations via flexible occupancy assignment of Slater determinants
 \item All electron and nonlocal pseudopotential calculations
 \item Casula T-moves for variational evaluation of nonlocal
-  pseudopotentials (nonsize consistent and size-consistent variants)
+  pseudopotentials (non-size-consistent and size-consistent variants)
 \item Wavefunction optimization using the ``linear method'' of Umrigar
   and coworkers, with arbitrary mix of variance and energy in the
   objective function
--- a/manual/features.tex
+++ b/manual/features.tex
@@ -1,29 +1,29 @@
 \chapter{Features of QMCPACK}
 \label{chap:features}
 \section{Production-level features}
-The following lists the main production level features of QMCPACK. If
-you do not see a specific feature that you are interested in listed,
-see the remainder of this manual and ask to see if specific feature is
-available or can be made full production level quickly.
+The following list contains the main production-level features of QMCPACK. If
+you do not see a specific feature that you are interested in,
+see the remainder of this manual and ask whether that specific feature is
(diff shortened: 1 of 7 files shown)


manual/features.tex
```diff
@@ -1,38 +1,38 @@
...
+\item Wavefunction updates using optimized multideterminant algorithm of Clark et al.
 \item Backflow wavefunctions
 \item One, two, and three-body Jastrow factors
 \item Excited state calculations via flexible occupancy assignment of Slater determinants
-\item All electron and non-local pseudopotential calculations
-\item Casula T-moves for variational evaluation of non-local
-  pseudopotentials (non-size consistent and size-consistent variants)
+\item All electron and nonlocal pseudopotential calculations
+\item Casula T-moves for variational evaluation of nonlocal
+  pseudopotentials (nonsize consistent and size-consistent variants)
```
markdewing: I don't think 'nonsize' is a word, and 'size' should be more connected to 'consistent' the preceeding 'non'.   Should this be 'non-size-consistent'?

prckent: Fixed

manual/installation.tex
```diff
@@ -1562,26 +1555,25 @@ \subsubsection{NiO performance tests}
...
-  \caption{System sizes and names for NiO performance tests. GPU performance
-    tests are named similarly but have different walker counts.}
+
   \label{tab:niotests}
 \end{center}
 \end{table}
 
 \subsection{Troubleshooting tests}
-ctest reports briefly pass or fail of tests in printout and also collects all the standard outputs to help investigating how tests fail.
+Ctest reports briefly pass or fail of tests in printout and also collects all the standard outputs to help investigating how tests fail.
```
markdewing: I think CTest should be capitalized in the same style as CMake

manual/running.tex
```diff
@@ -102,31 +102,31 @@ \subsection{Nested OpenMP threads}
...
 
 \subsection{Memory considerations}
-When using threads, some memory objects shared by all the threads. Usually these memory are read-only when the walkers are evolving, for instance the ionic distance table and wavefunction coefficients.
+When using threads, some memory objects are shared by all the threads. Usually these memory objects are read only when the walkers are evolving, for instance the ionic distance table and wavefunction coefficients.
```
markdewing: The second sentence should leave the hyphen in 'read-only'.   While 'read only' parses correctly, it does not convey the right intention - 'read-only' means 'not written', while 'read only' means 'only read during this phase'.

prckent: Fixed

manual/installation.tex
prckent: Fixed throughout

PR #1573, opened 2019-05-01
Fix value of sendcount passed to gatherv_n in dumpHDF5.

fdmalone:
Sendbuff array was not being resized to zero causing gatherv_n to hang
if number of walker blocks to write > 1. sendcounts is determined from
SendBuff.num_elements().


ye-luo:
Okay to test

Issue #1563, opened 2019-04-25
UseBoundBox can be true in OpenBC's


rcclay:
To handle open boundary conditions, it appears as though the UseBoundBox flag used by ParticleSet to auto-reject moves outside of a bounding box is set to "false" by initializing molecular type SPO's, and defaults to "True" otherwise.  It's NOT set out of explicit regard for the boundary conditions.  Doing a jastrow-only wave function in OpenBC's results in an insidious bug.  

I think that the default behaviour should be to set this flag to False, and then set to true if periodic boundary conditions occur or some other piece of code specifically asks for it.

jtkrogel:
Is this in place to prevent attempted out of bounds evaluation of the B-spline table for the orbitals?  If so, it is better logic to set to true only when B-spline orbitals are present.

ye-luo:
Current code
```
if(UseBoundBox)
{
  // this is likely the spline case
  // orbitals are defined in a box. only the directions with P boundary condition have orbitals defined form -inf to +inf.
  outOfBound() // if any direction of reduced coordinates displacement goes larger than 0.5, mark invalid
  isValid() // if any open direction of reduced coordinates of new position is not in the range [0,1), mark invalid
}
else
{
  // PW orbital only supports PPP
  // LCAO orbital is defined in the whole space no matter what BC is forced.
}
```
It is not clear to me if the outOfBound check is really helpful. I did see some random LiH runs failed due to this check and the electron gets stuck.
isValid check seems reasonable.

The other way of fixing this removing all the checks. Assign orbital values to zero if the electron goes out of the box at open directions with spline orbitals. Such moves are rejected naturally due to the ratio 0.

Any preference?

ye-luo:
A cleaner logic.
When a bound box is necessary?
1. OpenBC like a molecule, no latticed provided from input. using default lattice, a huge box.
2. A lattice from input and mixed P and N boundary conditions. outOfBound to guard too large moves. isValid only make coordinates out side the box in the open directions. A lattice with OpenBC or PPP are covered in this case.
So we only need to activate the bound check if a lattice is provided from user input (xml or hdf5). We don't need to know what SPO is in use.

prckent:
What are "too large moves"  and when do they occur?

ye-luo:
For a proposed move, the reduced coordinates of delta R are restricted to [0, 0.5] in each direction. This is the currently choice in the code. IMO, [0,1) is more reasonable.

PR #1575, opened 2019-05-03
Remove CheckBoxConds

ye-luo:
Keep only the used codepath.
For #1563 discussion.


prckent:
Just to confirm for records, this change will ensure we have consistent handling of the box independent of basis?

ye-luo:
This PR only trims of unused lines of code. No change in code execution. I found the old code hard to read, so made this simplification.

PR #1576, opened 2019-05-03
Deterministic tests for mixed precision build are added.

Hyeondeok-Shin:
Deterministic tests for mixed precision build are added on diamondC_1x1x1_pp, diamondC_2x1x1_pp, and LiH_dimer_pp testset. 

References for real_MP and complex_MP are separated, and tolerances are set to 5x larger value than the maximum variation between the full-precision vs. mixed precision results obtained within intel19, gcc-6.2.0, and clang7.0 compilers in the workstation.

This multiplication factor of 5 for MP will be adjusted after obtaining additional MP tests from other machines and compilers in order to relax the tolerance for MP build.

Hyeondeok Shin


prckent:
@Hyeondeok-Shin Thanks. I discussed with @jtkrogel yesterday but did not comment in time. Please can you list here how the reference data was generated? I assume full double or complex with one particular compiler.

Hyeondeok-Shin:
@prckent As the MP reference, the values are fixed as the values obtained from full-precision build with Clang4. 
For the error bar in the MP reference, I obtained MP observables using 3 different compilers (gcc 6.2.0, intel19, Clang7), and picked the largest variation for full vs. mixed precision build between these 3 results. So, errors in the MP reference come from one of these 3 compilers, and it is not a specific one compiler because a compiler showing the largest variation in MP against FP is not uniform.

Please let me know if additional information is needed.

prckent:
@Hyeondeok-Shin good for me.

prckent:
Test this please

PR #1581, opened 2019-05-05
Update qmca AFQMC local energy translation.

fdmalone:
The headers of the AFQMC scalar output files have changed.


markdewing:
okay to test

Issue #1339, opened 2019-01-28
List of actions for code formating


ye-luo:
We are aiming at a full repo (src) reformatting very soon.
If you have significant changes not merged yet, please comment.
We are making a plan and strategy to minimize impact on every developer.

To be completed before the action
1. [x] Update manual with day-to-day formatting procedure (@PDoakORNL )
2. [x] Merge PR #879 (msdj orbital optimization)
3. [x] Merge PR #1279 (GPU delayed updates)
4. [x] Merge CUDA AFQMC code or avoid initially reformatting AFQMC
5. [x] Pick a date for reformatting.
5. [x] Reformat.
6. [x] Discuss with and help authors of any pending PRs at time of reformat

prckent:
For anyone new to this topic, see discussion in #794

prckent:
Please do not unpin this issue.

PDoakORNL:
For the first checkbox I propose adding an href from the manual to the appropriate section of the wiki.  At any particular time the best way we've found to do the formatting and its current place in the workflow should be there.

PDoakORNL:
I propose creating the reformatting PR tomorrow.

ye-luo:
Can we start with a trial PR formatting one or two folders under src now to see if there is any issue?

PDoakORNL:
What we are talking about with function calls clang-format call bin-packing.
It's not very flexible i.e. it either goes to the max width or does one per line.
I prefer one per line to parsing several lines of max character width arguments but not enough to be a no vote.

Since @ye-luo is a strong BinPackArguments:true vote it's currently 1:0 to change this.

Thumbs up to change to true for remainder of reformat and future use. 
Voting will when the next reformat PR is merged and currently it will be changed to true.

```C++
BinPackArguments (bool)
```
If false, a function call’s arguments will either be all on the same line or will have one line each.
```C++
    true:
    void f() {
      f(aaaaaaaaaaaaaaaaaaaa, aaaaaaaaaaaaaaaaaaaa,
        aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa);
    }

    false:
    void f() {
      f(aaaaaaaaaaaaaaaaaaaa,
        aaaaaaaaaaaaaaaaaaaa,
        aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa);
    }
```

ye-luo:
[...]

PR #1583, opened 2019-05-06
Clang format final sweep

PDoakORNL:
Doing a final sweep before closing #1339


ye-luo:
I noticed that the files under src/QMCWavefunctions excluding the subfolders were not formatted. Could you do that as well?

ye-luo:
I'm working on this PR now.

PR #1585, opened 2019-05-08
Fix crashing due to mishandling of DriftModifier pointers

### LINKED ISSUE #1584: QMCPACK segfaults with multiple DMC blocks
Running QMCPACK on a simple system (diamond, fcc tiled to cubic, BFD potential) segfaults when there is more than one DMC section.  The segfault occurs on the first step of the second DMC section.

This is observed in the current develop, but not an earlier version (Dec. 2018).  The segfault happens for SoA/AoS, real/complex, mpi/serial.  I've observed it both on my laptop and within an AWS instance.

Files to reproduce the problem are attached.
[dmc_bug.zip](https://github.com/QMCPACK/qmcpack/files/3154557/dmc_bug.zip)

ye-luo:
Close #1584.

Function QMCDriver::process is supposed to handle consecutive driver input blocks without destroying and creating the same driver. I introduced a bug by recreating DriftModifier without updating its pointer used in QMCUpdateBase. With this PR, DriftModifier can be created only once unless a driver is destroyed. The side effect of this fix is that DriftModifier is not allowed to switch between multiple consecutive DMC blocks. This is only a minor issue.

I thinking in the future, we should get rid of reusing drivers. Every driver input section should be a fresh one. The destroying and creating cost is really negligible. This also get rid of the unintended setting left from a previous driver input section.

### DIFF
--- a/src/QMCDrivers/GreenFunctionModifiers/DriftModifierBuilder.cpp
+++ b/src/QMCDrivers/GreenFunctionModifiers/DriftModifierBuilder.cpp
@@ -27,7 +27,6 @@ DriftModifierBase* createDriftModifier(xmlNodePtr cur, const Communicate* myComm
     DriftModifier = new DriftModifierUNR;
   else
     myComm->barrier_and_abort("createDriftModifier unknown drift_modifier " + ModifierName);
-  DriftModifier->parseXML(cur);
   return DriftModifier;
 }
 
--- a/src/QMCDrivers/QMCDriver.cpp
+++ b/src/QMCDrivers/QMCDriver.cpp
@@ -216,8 +216,9 @@ void QMCDriver::process(xmlNodePtr cur)
     branchEngine->setEstimatorManager(Estimators);
     branchEngine->read(h5FileRoot);
   }
-  if (DriftModifier != 0) delete DriftModifier;
-  DriftModifier = createDriftModifier(cur, myComm);
+  if (DriftModifier == 0)
+    DriftModifier = createDriftModifier(cur, myComm);
+  DriftModifier->parseXML(cur);
 #if !defined(REMOVE_TRACEMANAGER)
   //create and initialize traces
   if (Traces == 0)


ye-luo:
For the moment, there is only one DrfitModifier type. So the current change 1) removes the crash 2) allows changing modifier parameters. So far the fix is enough. There is no need of abort because user anyway can't change the DrfitModifier type because there isn't a second type.
So I think we should accept this PR as it is.

I agree with @PDoakORNL that the current setting is problematic in the future.
I went through the code looking for how reusing a driver is handled. It is more complicated than I thought. There are some adhoc input parsing in the DMC driver but not others. DrfitModifier is used by all drivers and I need to find a clean solution for all the drivers.

ye-luo:
I cannot quickly do more than the current fix and it is important to fix the crash.
1) The crashing is fixed 2) Input parameter change is still allowed at every input block.
There is no second DriftModifier type and there is no need to abort.
So I think the current PR is sufficient.

I agree with @PDoakORNL. The current situation will be a problem when a second DriftModifier type is added.
I went through the code to see how the reusing of drivers is handled. I saw ad-hoc stuff in the DMC driver but not other drivers. DriftModifier is used by all the drivers and I need a better solution than a simple fix. It needs a bit of time.

ye-luo:
I cannot quickly do more than the current fix and it is important to fix the crash.
1) The crashing is fixed 2) Input parameter change is still allowed at every input block.
There is no second DriftModifier type and there is no need to abort.
So I think the current PR is sufficient.

I agree with @PDoakORNL. The current situation will be a problem when a second DriftModifier type is added.
I went through the code to see how the reusing of drivers is handled. I saw ad-hoc stuff in the DMC driver but not other drivers. DriftModifier is used by all the drivers and I need a better solution than a simple fix. It needs a bit of time.

ye-luo:
I cannot quickly do more than the current fix and it is important to fix the crash.
1) The crashing is fixed 2) Input parameter change is still allowed at every input block.
There is no second DriftModifier type and there is no need to abort.
So I think the current PR is sufficient.

I agree with @PDoakORNL. The current situation will be a problem when a second DriftModifier type is added.
I went through the code to see how the reusing of drivers is handled. I saw ad-hoc stuff in the DMC driver but not other drivers. DriftModifier is used by all the drivers and I need a better solution than a simple fix. It needs a bit of time.

ye-luo:
I cannot quickly do more than the current fix and it is important to fix the crash.
1) The crashing is fixed 2) Input parameter change is still allowed at every input block.
There is no second DriftModifier type and there is no need to abort.
So I think the current PR is sufficient.

I agree with @PDoakORNL. The current situation will be a problem when a second DriftModifier type is added.
I went through the code to see how the reusing of drivers is handled. I saw ad-hoc stuff in the DMC driver but not other drivers. DriftModifier is used by all the drivers and I need a better solution than a simple fix. It needs a bit of time.

prckent:
After consideration, I suggest we make a ticket about the remaining issues and merge this update.

PDoakORNL:
We are going to live with this for now but it must be dealt with #1589

PR #1586, opened 2019-05-08
Stop using variable size array in einspline

ye-luo:
HAVE_C_VARARRAYS caused find_coefs_1d_d failing silently. I got NaN on AWS nodes.
Simply remove them. find_coefs_1d_d is only called when initializing splines.


ye-luo:
With this fix, the short tests pass but the test_einspline is still problematic with gcc-7 and gcc-8 on AWS. I'm still battling.

@PDoakORNL feel free to delete more stuff before merging. I will signal you when my side is done.

ye-luo:
I added GNU compiler flag "-Werror=vla" to stop any VLA in the C code. I think we will do that for the C++ code as well but there are much more fix needed in the C++ code.
HAVE_C_VARARRAYS is fully cleaned up.

ye-luo:
The fix to test_einspline is also kill all the VLA in the unit tests.
This again highlight the importance of killing all the VLA in the code.

ye-luo:
@PDoakORNL is there anything you still would like to change before merging?

PR #1590, opened 2019-05-08
Nexus: Fix orbital file path handling for pyscf to qmcpack

jtkrogel:
Fix the handling of the H5 orbital file path when using convert4qmc to "convert" the orbitals from PySCF to QMCPACK.

Unlike other cases, convert4qmc does not actually create an h5 file in this case.   Instead a pre-existing h5 file made at the save_to_qmcpack step from PySCF needs to be used.  This fix passes the correct path info along, which comes from the xml input files created by convert4qmc.

This fix is needed for the workshop, please expedite.


ye-luo:
Okay to test

PR #1591, opened 2019-05-09
Nexus: pyscf expects kpoints in bohr regardless of input units

jtkrogel:
Fix for Pyscf workflows involving k-points.  Regardless of whether input structure is in Angstrom or Bohr units, Pyscf expects k-points in Bohr units.

This update is needed for the workshop, please expedite.


prckent:
OK to test

PR #1595, opened 2019-05-13
Enable NewTimer class being safely used in nested OpenMP regions + minor intel compiler configuration cleaning

ye-luo:
The change in the timer class has been verified in miniQMC.
Some minor Intel compiler configuration cleaning.


prckent:
Please update the PR title to describe the update

PR #1596, opened 2019-05-13
Move glibc version check to GNU compiler configuration

ye-luo:
The glibc issue only affects GNU compiler when it turns math function inside "omp simd" into calls to libmvec shipped by glibc.

This PR reduces necessary warning to users. Now if the issue is catched, a hard stop is in place.


prckent:
Good move. I believe PGI and LLVM use their own implementations and only GNU is using the glibc implementation.

PR #1600, opened 2019-05-17
Require the same data type between container and allocator

ye-luo:
This PR enforces a container and its allocator using the same data type.
This is a restriction stronger than std::vector.
std::vector documentation says "The behaviour is undefined if Allocator::value_type is not the same as T".


ye-luo:
Test this please

PR #1601, opened 2019-05-18
[WIP] Complex LCAO trial Wavefunction 

anbenali:
Phase applied to Value and Gradient. Terms in Laplacian strange and need to be carful with the second derivatives. 

- All coordinates of Super twists passed in a clean manner and code protected for the phase between complex and real  build of the code. 

- Added test to try recovering the phase. It is a 1x1x1 cell with a shift in the twist (0.33 0.33 0.33). This should avoid tiling and force only QMC phase. 

- The Laplacian term is located in  srd/QMCWavefunction/lcao/SoaAtomicBasisSet.h line 255. 
- I will try to write it in the weekend. 

Anyone feel free to give it a try if you have spare time.


prckent:
- To make the test harder, consider a lower symmetry twist with all components not equal. The  LiH arbitrary k-point tests for splines uses 0.1,0.2,0.3.
- Remove the .chk file unless needed

prckent:
OK to test

ye-luo:
Could you please start a new PR with tests only?
I'm checking a few things on the complex LCAO, your temporal change to LCAO code breaks the build.
I don't think it is necessary to fix all build variants since we are trying to get math correct first.
So if the tests can be added separately, we can merge the tests and l can use it instead of waiting for fixing every piece.

rcclay:
What's failing right now?

anbenali:
There is some build variant that I broke.. I tested for real and complex
but maybe something else is not right.

For the purpose of the task, I am still unable to get back the HF energy
from the test provided. The laplacian is however not implemented

prckent:
@anbenali You should have received email from Jenkins. Looks like some types are wrong - missing "cast" or similar. I am surprised that this compiled for you. Did you check in all your changes?

```
/autofs/nccs-svm1_compilers/gcc/rhel7-x86_64/6.2.0/include/c++/6.2.0/complex:386:5: note:   template argument deduction/substitution failed:
In file included from /tmp/jenkins-rhea-cpu-702-src/src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp:24:0:
/tmp/jenkins-rhea-cpu-702-src/src/QMCWaveFunctions/lcao/SoaAtomicBasisSet.h:252:35: note:   mismatched types 'const std::complex<_Tp>' and 'float'
            dpsi_z[ib] +=de_mikr_z * ( ang * vr ) + (ang * gr_z  + vr * ang_z ) * e_mikr;
                         ~~~~~~~~~~^~~~~~~~~~~~~~
/tmp/jenkins-rhea-cpu-702-src/src/QMCWaveFunctions/lcao/SoaAtomicBasisSet.h:252:80: error: no match for 'operator*' (operand types are 'float' and 'std::complex<double>')
            dpsi_z[ib] +=de_mikr_z * ( ang * vr ) + (ang * gr_z  + vr * ang_z ) * e_mikr;
                                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
```

anbenali:
@prckent : 
Fixed all build issues (due to mixed precision and a nicer compiler from mysides who throuhgh a warning on an syntax error instead of dying.. 

As for the example, agreed that a complex case is better but for now I am not able to reproduce a simple case. Once the laplacian is written properly and we can reproduce the HF energy of a 1/3 1/3 1/3 twist we will adapt more complicated test. 

@ye-luo : Can you work with this example or do you prefer another PR?

ye-luo:
I can run the test case you sent me earlier. This is fine.
To facilite the PR merging, we can separate adding tests and fixing the code because the latter takes a bit time. I can't directly merge this WIP branch to my working branch because you also changed some lines of code and there are conflicts. So I can't use your tests easily but only running them manually.
Yes. I mean start a new PR with tests only.

anbenali:
@ye-luo : I thought about separating the PR for the test, but the test fails and is intended to fail until we fix the code. I  added explicitely the test in this case so whoever wants to give it a try can have an example to test on. once the test is fixed then you are right that it might be better to just push more tests in a subsequent PR. For now we need the phase.

ye-luo:
Yes. I want these tests to fail. For the moment, the code changes are still exploration. So we can't get the PR merged. If the tests can be added separately, the new PR for the tests can get in without any problem.

prckent:
The WaveFunctionTester might be helpful at this point.

anbenali:
What I am sure off is that if I take the phase factor to be exp(-ikr) and
apply gradients and laplacian my total energy is -3ha instead of -11. If I
use exp(ikr) (note the sign) energy is -10ha. There might be some sign or
2_PI issue somewhere...

Would the WF tester help in this case?

anbenali:
@ye-luo : I added the phase higher in the call with cleaner evals of gradients and Laplacian. Still failing... 


I agree that exp(-ikr) r refers to th e absolute  electron coordinate so it does not make sens to use the dR from the Ion like initial attempts. but if using only  Coordinate we are missing the translations. Without translation, it is pointless to compute the phase...

PR #1606, opened 2019-05-21
Nexus: expand error detection for PWSCF

jtkrogel:
Detect PWSCF aborts characterized by "'Error in routine" in the log output and mark these as failures.


markdewing:
okay to test

PR #1610, opened 2019-05-23
AFQMC bug fixes for gcc8

### LINKED ISSUE #1475: Test failure of deterministic-unit_test_afqmc_matrix
Opening an issue for the test failure noted in #1467. The test fails with gcc/8.2.0 on oxygen and I can reproduce with gcc/8.1.0 locally. The copy assignment on line 222 of AFQMC/Matrix/tests/test_csr_matrix.cpp seems to be causing the code to crash. Potentially a c++17 issue. Could you take a look @correaa

fdmalone:
Should resolve #1475.


mmorale3:
they were fails kn gcc8 only, but they should have been everywhere.

prckent:
Please try to write specific PR titles - it helps when writing the human readable changelog and release notes.

prckent:
OK to test

PR #1611, opened 2019-05-24
No VLA in C++

ye-luo:
Since we have been bitten by VLA so many times, I was trying to get rid of all of them.
I only fix the part that is used currently but not all the parts compiled. Some spline evaluation routine in einspline still has VLA but they are not used any more, I tend to remove them later instead of fixing them.
-Wvla is turned on by default with GNU and we will catch more VLA.

In the NLPP, the code was not using VLA originally but to make the evaluateOne a const member function many temporal arrays are created. This const seems not necessary and this PR reverts this part of change and use temporal arrays owned by the class. @rcclay Please have a close look.


rcclay:
I haven't forgotten about this.  I can vouch for everything except NonLocalPseudoPotential, which I'm going to have to carefully go through.

prckent:
@ye-luo

1. Does the revised code pass a bounds check, e.g. using -D_GLIBCXX_DEBUG with gnu to check the new container access? (If this fails, presumably the old code was also problematic)
2. What part of evaluateOne causes a problem with constness?

ye-luo:
1. I was not brave enough to turn on bound checking. I think we should do this at least in deterministic tests. Also add address sanitizers. Currently, all the tests are running fine.
2. std::vector<ReaType> psiratios was the WF ratios for all the QP. It was defined in the class and used as a scratch space by evaluateOne. There was an intention to distribute all the electron ion pairs in NLPP calculation over threads and each pair needs to call evaluateOne. For this reason, evaluateOne is protected with const and psiratios_[nknots] was added inside evaluateOne and it was VLA. After adding new T-move schemes, I realized that the electron loop cannot be parallelized because of dependency. Another reason is threading electron-pair needs a lot of scratch memory to store orbital values and also bad for cache at large problems. I eventually abandoned this idea and added the threading inside evaluateOne. Now, evaluateOne only needs to be called serially in each walker. There is no need of constness any more. In fact, the restored class owned objects can be reused by multiple functions and it avoids allocating and deallocating on the fly.

prckent:
@ye-luo Thanks. 

@rcclay OK to merge?

PR #1612, opened 2019-05-27
Update container design

ye-luo:
In #1444, containers were expanded to support memory accessibility control via template arguments.
This PR updates that design with allocator_traits. The accessibility is directly specified by the allocator.
Most of the static_assert has been replaced with using SFINAE.


src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
```diff
@@ -62,23 +62,23 @@ class DelayedUpdateCUDA
   Matrix<T, CUDAHostAllocator<T>> U, Binv;
   Matrix<T> V;
   //Matrix<T> tempMat; // for debugging only
-  Matrix<T, CUDAAllocator<T>, MemorySpace::CUDA> U_gpu, V_gpu, Binv_gpu, temp_gpu, Ainv_gpu;
+  Matrix<T, CUDAAllocator<T>> U_gpu, V_gpu, Binv_gpu, temp_gpu, Ainv_gpu;
```
PDoakORNL: The style guide indicates we agree on one variable declaration per line. Here it is also inconsistent.

src/OhmmsPETE/OhmmsVector.h
```diff
@@ -131,12 +128,12 @@ class Vector
     if (n > nAllocated)
     {
       resize_impl(n);
-      if (MemType == MemorySpace::HOST)
+      if (allocator_traits<Alloc>::is_host_accessible)
```
PDoakORNL: I'm not fond of this return to runtime branching on allocator_traits. But the whole nAllocated, nLocal, n + attacheReference situation is not real clean and probably should be resolved in another PR first.

src/OhmmsPETE/OhmmsMatrix.h
```diff
@@ -107,38 +107,32 @@ class Matrix
...
-    static_assert(MemType == MemorySpace::HOST, "Matrix::operator= MemType must be MemorySpace::HOST");
     resize(rhs.D1, rhs.D2);
-    return assign(*this, rhs);
-  }
-
-  inline const This_t& operator=(const This_t& rhs) const
-  {
-    static_assert(MemType == MemorySpace::HOST, "Matrix::operator= MemType must be MemorySpace::HOST");
-    return assign(*this, rhs);
+    if (allocator_traits<Alloc>::is_host_accessible)
```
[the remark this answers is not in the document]
ye-luo: copy constructor and assign operator are special. Because of the compiler generated versions, I cannot make SFINAE working as intended and it caused problems. So I'm leaving if-statement here, in fact we can have if constexpr in the future to kill any overhead.

src/OhmmsPETE/OhmmsVector.h
ye-luo: The attachReference is for other purpose and I agree it needs improving.
Here I prefer focusing on this PR and get it finished.
I will move the branch outside.

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
ye-luo: Will fix.

ye-luo: Fixed.

PR #1614, opened 2019-05-28
Nexus: allow bypass of some SLURM settings for SuperMUC-NG

jtkrogel:
Give more user control over SLURM settings to SuperMUC-NG.  Inputs generated in the typical way should be fine, but this gives more flexibility if desired.


markdewing:
okay to test

PR #1615, opened 2019-05-29
Update oxygen test scripts

prckent:
PGI2019 and boost version increases


ye-luo:
What version of boost we should use in the nightlies? Minimum supported version or newer version?
I feel better testing the minimum version. If the current minimum version is not sufficient, we raise the minimum version.

prckent:
I updated Boost as part of chasing down MPI wrapper and general AFQMC C++ breakage, warnings etc. etc.. I am not going to change this back on oxygen since it takes a bunch of human time that I don't have to chase dependencies. Please make another issue on the boost version.

prckent:
What I think needs to happen on oxygen and elsewhere is a set of builds using "oldest supported" versions and "latest released and supported" versions.

ye-luo:
@prckent Please comment what you observed on oxygen with 1.61. Error or warning?
There is no need to investigate why but just list what you observed.

More discussion for version at #1616

PR #1617, opened 2019-05-30
update script for summit nightlies

grahamlopez:
This is a PoC for getting the Summit nightlies going. Right now, it only runs Real-SoA once a day at 1:05 AM local Summit time. I have started these and they currently running and reporting to cdash.

Future improvements might include:
- running more variants
- building on the computes to be good citizens
- building/running in parallel on one or more computes
- building/running in memory for possibly better performance


prckent:
Curiously no tests show up in the nightly category, while deterministic, unstable, performance, & converter all have entries on today's cdash.

grahamlopez:
Erm... what determines which cdash categories get populated?

prckent:
ctest_script.cmake. I am puzzled why there were no entries also. We could leave it running for a few more days to verify what is the stable situation.

grahamlopez:
okay, fine with me

ye-luo:
The tests included in the nightly ran last. I think during unstable, ctest was stopped by the stop-time to avoid the job being killed by walltime. Thus no tests in the nightly categora got run.

PR #1620, opened 2019-05-30
Require QE binaries are present when QE_BIN specified

prckent:
+Small improvement to pw2qmcpack patch readme.


CMakeLists.txt
```diff
@@ -34,6 +34,26 @@ SET(QMCPACK_VERSION_MINOR 7)
 SET(QMCPACK_VERSION_PATCH 9)
 SET(QMCPACK_VERSION "${QMCPACK_VERSION_MAJOR}.${QMCPACK_VERSION_MINOR}.${QMCPACK_VERSION_PATCH}")
 
+
```
ye-luo: Could you move this to cmake/run_qe.cmake

prckent: Done. ( The call is early in the cmake so that the user does not have to endure extensive cmake configuration before any abort if qe is not found. )

ye-luo: However now, pw.x and and pw2qmcpack will be checked twice. I still don't like this place. If you think QE check needs to be a head, we can also have python check ahead. Let me think a bit more.
Currently run_qe is only used in pw2qmcpack tests. For this reason, it is natural to put it there.

prckent: Why will pw.x be checked twice? I added a function that is called once as far as I can tell.

prckent: run_qe is now our home for qe related macros, functionality.

ye-luo: You are right. A function is good.

prckent:
Test this please

PR #1622, opened 2019-05-31
Improve AFQMC scripts

fdmalone:
Some updates for AFQMC wavefunction generation.


prckent:
Can you be more specific? e.g. includes k-point handling bugfix.

prckent:
OK to test

fdmalone:
1. Appends Wavefunction by default to hamiltonian hdf5 file if hamiltonian is requested.
2. Optionally disable hamiltonian generation.
3. Determines number of electrons from wavefunction hdf5 rather than hamiltonian for template input generation.
4. Fixes unit test as a result of hdf5 wavefunction output.

Issue #1624, opened 2019-06-03
Need smart pointers. Using LCAOrbitalSetWithCorrection as a starting example


ye-luo:
#1623 exposed a problem that we need to reduce the use of raw pointers and use smart pointers if pointers are needed to share data between class objects.

I extract lines of code to better describe a right way to do in my mind.
```
struct LCAOrbitalSetWithCorrection : public LCAOrbitalSet
{
  SoaCuspCorrection cusp;

  SPOSet* LCAOrbitalSetWithCorrection::makeClone() const
  {
    LCAOrbitalSetWithCorrection* myclone = new LCAOrbitalSetWithCorrection(*this);
    myclone->myBasisSet                  = myBasisSet->makeClone();
    myclone->IsCloned                    = true;
    return myclone;
  }
}

struct SoaCuspCorrection
{
  typedef CuspCorrectionAtomicBasis<RealType> COT;
  aligned_vector<COT*> LOBasisSet; // raw pointer, bad

  SoaCuspCorrection(const SoaCuspCorrection& a) = default;
}

template<typename T>
class CuspCorrectionAtomicBasis
{
  typedef MultiQuinticSpline1D<T> RadialSetType;
  RadialSetType AOs;
  std::vector<T> phi, dphi, d2phi;

  CuspCorrectionAtomicBasis(const CuspCorrectionAtomicBasis& a) = default;

  inline void evaluate(const T r, T* restrict vals)
  {
     phi.resize(nr);
  }
}

template<typename T>
class MultiQuinticSpline1D
{
  coeff_type* Coeffs; // raw pointer, bad
  bool own_spline;

  MultiQuinticSpline1D(const MultiQuinticSpline1D& in) = default;

  MultiQuinticSpline1D<T>* makeClone() const
  {
    MultiQuinticSpline1D<T>* myclone = new MultiQuinticSpline1D<T>(*this);
    myclone->own_spline              = false;
    return myclone;
  }
}
```

Each thread has a copy of `LCAOrbitalSetWithCorrection`. The makeClone function triggers the copy constructor and copies the pointers of `SoaCuspCorrection::LOBasisSet` by value instead of deep copy. Multple threads attempt to resize in `CuspCorrectionAtomicBasis::evaluate` and breaks the code.

[...]

PR #1625, opened 2019-06-03
Fix cusp correction evaluation.

### LINKED ISSUE #1623: Cusp correction crashes
See e.g. https://cdash.qmcpack.org/CDash/testDetails.php?test=5233020&build=60682 
These are Intel2018 release builds. short-chn_ae_cuspCorrection-vmc-1-16 is crashing somewhere in the initial VMC. A PR from last week is likely at fault since the runs appeared stable before.

image

ye-luo:
Closes #1623
This is a workaround. phi, dphi, d2phi are allocated and freed locally and may cause performance penalty.

A more clean fix is needed as #1624 described.


prckent:
Provided the performance is OK, this is enough and need not be revisited. We don't have any molecular performance tests yet however - I'll check the status of them.

grahamlopez:
test this please

PR #1627, opened 2019-06-04
Save Optimized CI coefficients in HDF5 file

anbenali:
When reading CI coeffs from H5 files, Optimization of the coefficients was possible but it was impossible to chose the best coeffs, or restart from coefficients. 

This PR dumps the new optimized coeffs in H5 *.s0XX.opt.h5 file and allows to restart from optimized coeffs.

To save in HDF5:
the following tag needs to be added in the optimization block:
  <qmc method="linear" move="pbyp" checkpoint="-1" hdf5="yes" >

After the optimization block is done, (s001 for example), the following files are created:
*.s001.opt.xml 
*.s001.opt.h5

*.s001.opt.xml is the same as usual except that now it contains a member "opt_coeffs=*.s001.opt.h5" 
  <detlist size="1487" type="DETS" nca="0" ncb="0" nea="2" neb="2" nstates="85" cutoff="1e-2" href="../LiH.orbs.h5" opt_coeffs="LiH.s001.opt.h5"/>

The file *.s001.opt.h5 contains ONLY the new CI coefficients. 

The file *.s001.opt.xml can be used as a new wavefunction file. However, when parsing the file, if the builder finds the tag "opt_coeffs". the CI coefficients will be read from the *.s001.opt.h5 file. 

TESTS:
There are 3 tests:

1- test1: LiH_ae_noj: This tests multideterminants and recovers the Variational energy from sCI (CIPSI). 
2- test2: LiH_ae_j: This tests  simple VMC energy for a multideterminant trial wavefunction with 1,2 and 3 body Jastrows and a truncated multidet expansion (1e-3). 
3- test3: LiH_ae_Opt_CI: This test is particular as it starts from 1,2 and 3body Jastrow optimized with CI coefficients truncated at 1e-3. the new CI coeffs are stored in the s010.opt.h5 file and the WF used is s010.opt.xml where the appropriate Jastrows are stored.  This last test is the only one that tests properly one part of the newly added features of the PR. 

Missing:
- Deterministic test for the CI optimization. 

The logic behind this is to force the user to always have the set Orbitals+CI+Jastrows together. 

It is easy now to also store the Jastrows (and the orbitals) or any optimized value in the H5File but  it is a design change to move the Jastrow outside the XML and requires more discussions.


markdewing:
okay to test

prckent:
reportH5 -> ciCoeffsInH5File or similar

Looking through this WIP reminds me that the mixing of XML handling and h5 handling is not ideal and will be a problem when appropriate disinfection is carried out. This is a problem across the entire app, but try to put h5 handling in separate functions to keep the h5 code clean of our bad XML ways. i.e. writeCICoeffsH5(..) should not touch XML.

anbenali:
@prckent 
I added comment to explain the change in call order when separating HDF5 from XML. 
I added fail safe in case CI optimize=no but hdf5=yes. Instead of creating empty hdf5 and wrong *.opt.xml now nothing happens and old behavior is secured. 

Remember that I made a unilateral decision to save only the coefficients in HDF5 file. Other option would have been to clone completely the previous HDF5 and only update the coefficients, but we would end up with a new H5 file for every step. If any better solution I am more than happy to discuss it. (@jtkrogel @ye-luo )

jtkrogel:
@anbenali but the optimal MD coefficients are written to a new H5 every series, right?

anbenali:
@jtkrogel Yes at each series there will be a new *.sXXX.opt.h5 that contains only the new CI coeffs. 
the corresponding *.sXXX.opt.xml file will have something similar to this:

        <detlist size="1487" type="DETS" nca="0" ncb="0" nea="2" neb="2" nstates="85" cutoff="1e-2" href="../LiH.orbs.h5" opt_coeffs="LiH.s042.opt.h5"/>

Where there is now a opt_coeffs path. 
When restarting from the the optimized coefficients, you will need the old WF H5 file. All coefficients will be read, then the optimized ones will replace the old ones. 
The reason behind is that during optimization some coeffs my go below the cutoff specified in the WF. When we read the coeffs initially we sort them again. So it becomes important to read the old WF to avoid weird situations where the cutoff does not match the requested WF.

anbenali:
Ready to review (and merge)

anbenali:
>     * Please rename the _Opt_ files _opt_ consistent with our other optimization runs
Done

>     * Use standard doxygen comments and language for your function comments, e.g. reportParamtersH5
Done

src/QMCDrivers/QMCFixedSampleLinearOptimize.cpp
```diff
@@ -528,8 +531,12 @@ bool QMCFixedSampleLinearOptimize::put(xmlNodePtr q)
 #endif
       optTarget = new QMCCostFunction(W, Psi, H, myComm);
     optTarget->setStream(&app_log());
+    if (ReportToH5=="yes")
+       optTarget->reportH5=true;
     success = optTarget->put(q);
   }
+  if (ReportToH5=="yes")
```
ye-luo: Could you explain why these two lines needs to be repeated?

src/QMCDrivers/QMCCostFunctionBase.h
```diff
@@ -77,6 +77,10 @@ class QMCCostFunctionBase : public CostFunctionBase<QMCTraits::RealType>, public
   ///process xml node
   bool put(xmlNodePtr cur);
   void resetCostFunction(std::vector<xmlNodePtr>& cset);
+  ///Save opt parameters to HDF5
+  bool reportH5;  
+  bool CI_Opt;  
+  char newh5[128];
```
ye-luo: Need comments on both CI_Opt and newh5. newh5 better to be replaced with std::string.

src/QMCDrivers/QMCFixedSampleLinearOptimize.cpp
anbenali: The writing of the CI from H5 is activated by the tag hdf5=yes 
 
it can only be read in this function bool QMCFixedSampleLinearOptimize::put(xmlNodePtr q)
but it is needed to write from optTarget.  

optTarget needs to be created before we can  pass the value of ReportH5. OptTarget is created only once but if you have multilple optimization blocks you can decide to save only a few blocks to H5. We therefore need to read again the value hdf5 and pass it to optTarget. hence one definition in the if optTarget==0.
Now in the optTarget==0 there is a call to 
    success = optTarget->put(q);
this function needs to know if ReportH5 is yes or no before being called. This is why i cannot use just the one call outside the if statement.

anbenali:
@prckent @ye-luo 
Do the changes meet the requirements? 
If so it would be nice to merge it so I can move to the next PR.

PR #1628, opened 2019-06-05
Editing pass on several manual chapters

prckent:
From Charlie Horak/ORNL


prckent:
Added WIP since new chapters are coming in

prckent:
Removed the WIP. Will open a new PR for further edits as they are done.

PR #1630, opened 2019-06-05
Partial occupancies

fdmalone:
Deals (in some way) with AFQMC trial wavefunction generation for metallic systems with partially occupied bands by automatically generating small multi determinant trial wavefunction from degenerate states. This should be considered very experimental.

Along the way fixed some interface inconsistency issues for molecular wavefunction generation and I've added some flexibility for input file generation.


prckent:
OK to test

PR #1631, opened 2019-06-06
Tighten tolerances for mixed precision deterministic tests

Hyeondeok-Shin:
Tolerances for mixed precision deterministic test are relaxed. (diamondC_1x1x1_pp, diamondC_2x1x1_pp, and LiH_dimer_pp testset.)

 The mixed precision results are collected from the cdash, vesta with clang, and workstation with clang, gcc, and intel compiler, and the largest deviations from the reference among those are set as the reference error bars.

 Obviously, it is empirical, but we'll be able to figure out whether the mixed precision tests are within the bound of the reference relaxed from various compilers and machines.
     
 Hyeondeok Shin


prckent:
Please point to some test runs that have been failing in cdash or list some specific problems that have been encountered. A first step on encountering problems should be to make an issue.

I would rather see some investigation of what is going on in these cases and a tightening of tolerances as a result vs risk letting problems through by continuing to widen the tolerances.

prckent:
I made the above comment before reviewing the numerical values. It seems that most of the changes actually *tighten* the tolerances. A few are increased. Tightening the tolerances is a good thing, so no objections.

I am still interested in what prompted this update and why the tolerances can now be smaller. e.g. Do new clang results fall better to center of the distribution of results giving us a better average to work with?

prckent:
OK to test

grahamlopez:
test this please

grahamlopez:
test this please

Hyeondeok-Shin:
@prckent Previous tolerance was set as 5x larger value than the maximum deviation. It is promising to pass, but it is too large as the deterministic test for the mixed precision build, and hard to measure whether it is working normally and variation from the full-precision build. 
 Now tolerances are adjusted as actual values of the maximum deviation between collected mixed precision runs, so all tolerances are decreased from 5x scale to 1x. (Please remind me if there is something increased in the updated values.)

grahamlopez:
test this please

grahamlopez:
test this please

prckent:
Test this please

Issue #848, opened 2018-05-11
RPA Jastrow is broken and needs to be fixed


jtkrogel:
RPA Jastrow is currently broken and needs a fix.  There is a test in place to verify the correctness of any fix attempts.  A fixed RPA Jastrow should lead to improved finite size corrections currently being worked on by @rcclay.

See QMCPACK/tests/heg/heg_54_J2rpa/README

PR #1635, opened 2019-06-12
Resurrect RPAJastrow

Paul-St-Young:
Responding to #848 

Two main things I did:
  1. use 100 instead of 12 1D grid points to evaluate the Bspline basis functions. The old setting resulted in very poor transfer of the short-range function from `ShortRangePartAdapter` to `BsplineFunctor`.
  2. use `kSpaceJastrow` instead of `LRTwoBodyJastrow`.

I think the RPAJastrow is mostly fixed. However, more testing and cleaning are required.
Most importantly, I had to re-enable `LPQHIBasis.h::df`. `df(0)` is used to set the cusp of the short-range part. I am not convinced that this is right. I expected cusp=-0.5, but `-df(0)=-0.544`.

Is `LRTwoBodyJastrow` still maintained at all? If not, then I am in favour of removing it.


markdewing:
okay to test

Paul-St-Young:
@ye-luo thanks for pointing out the unit test example. I will add the unit test and remove `LRTwoBodyJastrow`.

One question: why was `LPQHIBasis.h::df` slated for deprecation? Is it broken?

jtkrogel:
There is something I don't understand here.

The RPA Jastrow is a specific form (Gaskell form resulting from an Ewald sum of a particular radial function), so I don't understand how substituting in the kSpaceJastrow for the LR part reproduces this Jastrow.

For the particular form, see Ceperley, Phys. Rev. B 18, 3126 (1978) ([https://doi.org/10.1103/PhysRevB.18.3126](https://doi.org/10.1103/PhysRevB.18.3126)) equations 8 (rpa), 9 (yukawa), and the Appendix.

The main point of this Jastrow, as I understand it, is to mirror the Ewald potential in the Jastrow pseudopotential so that the long ranged component is automatically correct.  This has overlap with a k-space form, but it is not the same.

Paul-St-Young:
@jtkrogel indeed, the LR part alone does not reproduce the RPA form (eg. Gaskell). `RPAJastrow` has both a LR and a SR part. When added together, the sum reproduces the RPA form. The current `RPAJastrow` works as follows:
1. Put the analytical RPA Jastrow potential U_k in reciprocal space. Only k vectors compatible with the supercell are used.
2. Perform an optimized breakup to cut the k-space parameters at `Kc` at the cost of introducing a short-range function from `r=0` to `r=Rcut`. Now U_k = FT[U^{sr}]_k + U^{lr}_k
3. Represent U^{sr}(r) using `BsplineFunctor` and represent `U^{lr}_k` using `kSpaceJastrow`

I have checked that the breakup works correctly for the Gaskell form. Namely, FT[U^{sr}]_k + U^{lr}_k is indeed = U_k Gaskell. `RPAJastrow` was broken before because: a. U^{sr}(r) was not accurately represented and b. `LRTwoBodyJastrow` uses `e^{-U/2N}` instead of `e^{-U/2V}`.

I thought about attempting to fix `LRTwoBodyJastrow`, but I noticed that the code is not used anywhere else and is replicating part of `kSpaceJastrow`'s capabilities, so I wanted to remove it instead.

prckent:
Please update the manual when finalized

jtkrogel:
@Paul-St-Young, thanks.  I missed how the k-space coefficients were initialized.

Are you able to reproduce the J2 RPA test from 2008?  See qmcpack/tests/heg/heg_54_J2rpa.

The input file is heg.rpa.rs5.xml and the total energy at that time was:
```
>qmca -q e -e 2 qmc-ref/*s001*scalar*
qmc-ref/HEG.rs5.rpa  series 1  LocalEnergy = -4.209725 +/- 0.012941
```

With your new B-spline approach, you might beat this energy with the same inputs.

src/QMCWaveFunctions/Jastrow/RPAJastrow.cpp
```diff
@@ -173,7 +193,9 @@ void RPAJastrow::makeShortRange()
   // J2OrbitalSoA<BsplineFunctor<RealType> > *j2 = new J2OrbitalSoA<BsplineFunctor<RealType> >(targetPtcl,IsManager);
   TwoBodyJastrowOrbital<BsplineFunctor<RealType>>* j2 =
       new TwoBodyJastrowOrbital<BsplineFunctor<RealType>>(targetPtcl, IsManager);
-  size_t npts    = 12;
+  size_t nparam  = 12; // number of Bspline parameters
```
rcclay: Hard coded number of bspline knots gives me the willies.  Why 12, and will this come back to bite us if the cell geometry is changed significantly?  Perhaps a comment explaining why 12, or increasing it to a larger, safer number.

src/QMCWaveFunctions/Jastrow/RPAJastrow.cpp
```diff
@@ -185,7 +207,7 @@ void RPAJastrow::makeShortRange()
   Y[npts]              = 0.0;
   std::string functype = "rpa";
   std::string useit    = "no";
-  nfunc->initialize(npts, X, Y, SRA->df(0), Rcut, functype, useit);
+  nfunc->initialize(nparam, X, Y, cusp, Rcut+tiny, functype, useit);
```
rcclay: A comment on why "tiny" is being added to Rcut would be nice.  This seems to be the result of a "gotcha"?

src/LongRange/LPQHIBasis.h
```diff
@@ -88,7 +88,6 @@ class LPQHIBasis : public LRBasis
   }
   inline mRealType df(int n, mRealType r) const
```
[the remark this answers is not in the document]
Paul-St-Young: will do!

src/QMCWaveFunctions/Jastrow/RPAJastrow.cpp
Paul-St-Young: 12 was there before I touched the file. I intend to make it an input parameter eventually. Should I do it in this PR?

src/QMCWaveFunctions/Jastrow/RPAJastrow.cpp
Paul-St-Young: "tiny" is there because of numerical issue. If I set Rcut to L/2 exactly, then BsplineFunctor may complain.

The old code had `Rcut-0.1`, which is strictly worse IMO. I think "tiny" should also be an input parameter that defaults to a very small number.

ye-luo: What BsplineFunctor complained?

prckent:
In my opinion reactivation of this code and non-trivial improvements should be in separate PRs. Better to iterate fast. e.g. Add input paramters in a later PR.

src/QMCWaveFunctions/Jastrow/RPAJastrow.cpp
rcclay: Not necessary, but definitely leave a note/comment for posterity.

Paul-St-Young:
@jtkrogel 
> Are you able to reproduce the J2 RPA test from 2008? See qmcpack/tests/heg/heg_54_J2rpa.

Yes with an asterisk.

After using Ray's fixed `dh_dr`, the cusp becomes correct, and I do get an energy that is within error bars or the old result. Although the error bar of the old calculation is too large for a conclusive comparison.
```
HEG.rs5.rpa  series 1  -4.191195 +/- 0.002719    1.3   0.052024 +/- 0.000880    1.0   0.0124 
```
old
```
HEG.rs5.rpa  series 1  LocalEnergy = -4.209725 +/- 0.012941
```

If I reoptimize the short-range part, then my results can be improved.
```
rpa  series 0  -4.205725 +/- 0.000132    1.0   0.030828 +/- 0.000095    1.1   0.0073
```

src/QMCWaveFunctions/Jastrow/RPAJastrow.cpp
Paul-St-Young:
```
ERROR ERROR: Error in BsplineFunctor::initialize: r > cutoff_radius.
Fatal Error. Aborting at BsplineFunctor::initialize
```

jtkrogel:
@Paul-St-Young Excellent.  This qualifies as passing in my book.  Great work Paul.

We should create an issue or similar to prompt the regeneration of the reference data and update the integration test + remove it from "developer" status.

Paul-St-Young:
@ye-luo @rcclay I think I have addressed all your concerns. Please see if this PR is good to merge now.

The unit test is not great right now. It is more of a regression test that makes sure we are at our current state without really checking the absolute correctness of the code. I have a few ideas for better tests, but I would rather do them in a future PR along with other non-trivial improvements. Please let me know if this is OK.

ye-luo:
@jtkrogel Please add your review. If you are satisfied with this PR, we will merge it ASAP.

PR #1641, opened 2019-06-17
Optimize offload

ye-luo:
1. Add a few allocator CUDALockedPageAllocator and PinnedAllocator/PinnedAlignedAllocator.
2. introduce MultiBsplineEvalHelper.hpp, all the helper functions are moved to this class. getSplineBound is moved there. computeLocationAndFractional is introduced to determine the location of an electron on the spline grid and compute all the prefactor. Separating this from the spline evaluation adds more flexibility and facilitate future needs.


prckent:
As previously discussed, we shouldn't do any further work as regards offload, performance portability or refactoring of the main app until a proper design and responsibilities for it have been worked out. To do so risks building things that we are only going to have to tear down and does not allow others to weigh in appropriately (who have been waiting for me to catch up, as requested).

PR #1642, opened 2019-06-17
Nexus: update outdated cubic specifier to alat for QE

jtkrogel:
Remove the outdated "cubic" specifier on the CELL_PARAMETERS card and use "alat" instead.

Improvement suggested by Ron Cohen.


prckent:
OK to test

PR #1643, opened 2019-06-17
Add OMP parallelization for cusp correction

markdewing:
Add timers for cusp correction.
 
Parallelize over both the MO's and centers.

The diff view is more complicated that necessary because of an indentation change to a large section of code from adding the pragma omp parallel scope.    The changes in body should be
 1. Add timer start/stop around minimizeForRc
 2. Declare and initialize local (private) versions of targetPtcl, sourcePtrc, phi, and eta.
 3. Change targetPtcl, sourcePtcl, phi, and eta to local variable equivalents (localTargetPtcl, localSourcePtrc, local_phi, and local_eta)

In theory, targetPtcl and sourcePtcl could be declared firstprivate and the compiler should use the copy constructor to generate the private versions.  Works with clang, but causes an internal compiler error with Intel 19.0.0.120.

The MO and center loops are interchanged again.  In theory splitting the MO's into eta and phi only needs to be done once per center.   By putting the MO loop on the inside, hopefully some data might still be in cache when doing the same split again.   The time spent splitting into eta and phi isn't enough to be worth any more complicated scheme to avoid recomputing it.


markdewing:
test this please

prckent:
@PDoakORNL do you have time to look at this?

PDoakORNL:
I will look this today.

PDoakORNL:
test this please

src/QMCWaveFunctions/lcao/CuspCorrectionConstruction.cpp
```diff
@@ -211,55 +223,84 @@ void generateCuspInfo(int orbital_set_size,
...
-          break;
-        }
-      }
-
-      if (corrO)
+      for (int mo_idx = start_mo; mo_idx < end_mo; mo_idx++)
       {
-        OneMolecularOrbital etaMO(&targetPtcl, &sourcePtcl, &eta);
-        etaMO.changeOrbital(center_idx, mo_idx);
+        app_log() << "   Working on MO: " << mo_idx << " Center: " << center_idx << std::endl;
```
PDoakORNL: This should be eliminated, the debugger can also tell you every time you are here. Or at least make it debug only. You should not writing from every thread and profiling, since there should be a mutex back there somewhere.

src/QMCWaveFunctions/lcao/CuspCorrectionConstruction.cpp
```diff
@@ -211,55 +223,84 @@ void generateCuspInfo(int orbital_set_size,
   int start_mo = offset[Comm.rank()];
   int end_mo   = offset[Comm.rank() + 1];
   app_log() << "  Number of molecular orbitals to compute correction on this rank: " << end_mo - start_mo << std::endl;
-  for (int mo_idx = start_mo; mo_idx < end_mo; mo_idx++)
+
+#pragma omp parallel
```
PDoakORNL: I'd like to see  us move toward abstract or C++ concurrency constructions. The implicit capture of variables by the omp parallel pragma contributes to hazy scope and fears about race conditions.  It makes it very convenient to write monster methods that build up a huge local namespace.
Is the potential confusion about thread scope worth it just to have what I think is initialization code be faster?

src/QMCWaveFunctions/lcao/CuspCorrectionConstruction.cpp
```diff
@@ -211,55 +223,84 @@ void generateCuspInfo(int orbital_set_size,
...
+    LCAOrbitalSet local_eta(eta);
+    local_eta.myBasisSet = eta.myBasisSet->makeClone();
+    local_eta.IsCloned   = true;
+    local_eta.C          = nullptr;
+    local_eta.setIdentity(false);
+
+// Specify dynamic scheduling explicitly for load balancing.   Each iteration should take enough
+// time that scheduling overhead is not an issue.
+#pragma omp for schedule(dynamic) collapse(2)
     for (int center_idx = 0; center_idx < num_centers; center_idx++)
```
PDoakORNL: This should be the beginning of an actual lambda or function that controls what's available at thread scope. The signature and documentation should clearly show what is const/immutable at the thread scope and what is the shared object being updated. All the construction above would also cleary be in inside the thread "body" i.e. the lambda or function.

src/QMCWaveFunctions/lcao/CuspCorrectionConstruction.cpp
```diff
@@ -211,55 +223,84 @@ void generateCuspInfo(int orbital_set_size,
...
+            pos[i] = (i + 1.0) * dx;
+          }
+
+          RealType eta0 = etaMO.phi(0.0);
+          ValueVector_t ELorig(npts);
+          CuspCorrection cusp(info(center_idx, mo_idx));
+          computeTimer->start();
+          minimizeForRc(cusp, phiMO, Z, rc, Rc_max, eta0, pos, ELcurr, ELideal);
+          computeTimer->stop();
+          info(center_idx, mo_idx) = cusp.cparam;
```
PDoakORNL: Is this the only shared object being updated? That should be documented.

src/QMCWaveFunctions/lcao/CuspCorrectionConstruction.cpp
prckent: Background info: The Cusp Correction construction/calculation is distressingly slow, so maintainable improvements are definitely welcome imo.

src/QMCWaveFunctions/lcao/CuspCorrectionConstruction.cpp
markdewing: I think it's important that there be some feedback to the user that progress is being made.   The loop body is time-consuming enough that any lock here should not be an issue.  I'll add an omp critical pragma to keep the output clean, and to make it clear there is a lock.

src/QMCWaveFunctions/lcao/CuspCorrectionConstruction.cpp
markdewing: Integrated the parallel scope with the for loop scope.  This will result in additional allocations/deallocations with the objects, but it shouldn't affect the run time.

ye-luo: Sorry I still have question with respect to the last change.
```
#pragma omp parallel
{
  ParticleSet localTargetPtcl(targetPtcl);
  #pragma omp for
  for(..)
  {
    do something on localTargetPtcl;
  }
}
```
This is a valid OpenMP code and this is a recommended optimization to avoid allocation and deallocation. I had a couple times corrected by OpenMP experts that my expectation of "#omp for" of concurrent execution of the loop iterations was wrong. "#omp for" is a workshare construct and indicating the loop can be distributed among the threads spanned by the parallel region. It didn't say that the loop iterations are independent and can be executed concurrently. Instead in OpenMP 5.0, the loop construct is introduced indicating that the loop iterations are independent and this definition aligns with the concept of the concurrent loop in C++ and Fortran.

Going back to the example, the thread scope is the parallel region and the localTargetPtcl is defined in the right scope.
It should have no difference with a code without OpenMP.
```
{
  ParticleSet localTargetPtcl;
  for(..) // The effect of "omp for" is changing the lower and upper bound of the loop, although it depends on the scheduling
  {
    do something on localTargetPtcl;
  }
}
```

The actual code did suffer from allcation/deallication and the imbalance between iterations make all the overhead amplified. I remember that when I profiled the cusp correction, the constructor and destructor took a lot of time. For this reason, I think the old way from @markdewing  of doing parallel region is preferred.

ye-luo: @PDoakORNL can we be a bit more flexible allowing this optimization to reduce overhead? it does introduce more data in the thread scope but scope and lifetime is quite clear.

PDoakORNL: I'm aware openmp magically creates the "right" number of thread local localTargetPtcl's here.
I'm not claiming its not "good" openmp code it is valid. However openmp is a collection of compiler magic funneling into a runtime that is not a part of C++. A great deal is implicit and therefore easier for developers to neglect thinking through. 

Good options exist in std C++ and C++ libraries to deal with concurrency concerns around races and efficient scheduling. Most of them are far clearer in there semantics. 

So if LCAOOrbitalSet is so expensive to construct local copies of, is it actually even necessary?
Is anything but their C Matrices mutable in thread scope?

ye-luo: I think there is a reason that C++ threading is not favoured in HPC land. Its tasking based model is slow for small loop counts and short loop body and controlling affinity is hard. OpenMP is easy to use and the code is much simpler. Simple and clean code is often winning. I don't think C++ thread is good enough to meet the performance of my nested OpenMP implementation.

Going back to the code. The code is very ugly, it is absolute not necessary to make so many big local object copies. I guess only one row of the C matrix is used each time. It is horrible. At this moment, there is no need to seriously rewrite this part of code. We just need something working a bit better.

markdewing: I did some profiling on my laptop, and there doesn't seem to be significant overhead in creating the LCAOrbitalSet copies.

PDoakORNL: @ye-luo I would rather leave the code as is less dependent on openmp syntax. Is that acceptable?

I also think that we'd be better off refactoring egregious design issues when they are the cause of performance issues. Although according to Mark this isn't really the hotspot.  @markdewing  Can you tell what it is?

ye-luo: As @markdewing confirming that overhead is small. Putting object within the innermost scope is the cleanest way. If this really impact a workload, we can revisit this.

markdewing: Bulk of time (85%) is in DGEMM called from LCAOrbitalSet::evaluate, called from OneMolecularOrbital::phi_vgl, called from getCurrentLocalEnergy.
This is for a larger system, and I killed it part way through the cusp correction (so pretty much all the time in the run was spent doing cusp correction)

ye-luo: This is not the type of overhead I worried. I mean measuring the time of cusp construction with and without the optimization of copy.

PR #1644, opened 2019-06-17
Fix opt.tex so the manual builds.

markdewing:
And a few minor language changes.


prckent:
Test this please

prckent:
Confirming that the manual does actually build now.

PR #1645, opened 2019-06-18
Nexus: fix structure skew function

jtkrogel:
Fix small bug in skew function.


markdewing:
okay to test

markdewing:
Could you provide more context - using transpose vs. no transpose feels rather subtle.    Is there a reference for applying a skew transformation?   How was this bug discovered?      The routine is called by stretch and adjust_axes - is skew ever called by scripts directly?  What are some common uses for it?

jtkrogel:
This function was seldom used and the bug was found by trying to use it.  The job of the function is to apply a matrix operation consistently to the internal data of the structure class.  The real space (A) and reciprocal space (B) axes should always satisfy the relationship B.T = 2*pi*inv(A).  The skew function broke this relationship (the incoming matrix was not appropriately applied to the reciprocal axes) and the use of transpose there restores it.

The inconsistency was identified by the error checking code in the Structure.check_consistent function that is used in situ to guard against inconsistencies between the real and reciprocal space axes.  Since the problem was caught by this check (which now passes for the rotation matrix I was using) and this check is still in place to mitigate against problems, I think the risk of the change (which I think is mathematically correct anyway) is small.

markdewing:
Thanks @jtkrogel .  This description is what I wanted.

PR #1648, opened 2019-06-19
Nexus: add MD analysis capabilities

jtkrogel:
In this PR, the ability to gather and analyze relevant data from QE molecular dynamics runs is added to PwscfAnalyzer.

As a start, fast data scraping is enabled by using memory mapped files.


markdewing:
okay to test

jtkrogel:
This one can move forward.  I'll come back with any needed updates later, but the code functions at a basic level at present.

prckent:
How extensive to you plan to make this / have you received requests to support this?

[ (i) A wider support for MD might require support for multiple runs - e.g. I have never published an MD from less than 20 concatenated runs. (ii) There are a large number of existing analysis codes for MD trajectories and well supported formats. e.g. via mdtraj, ASE, lammps, VASP, nMOLDYN/MMTK...   ]

jtkrogel:
I've received requests for basic support from Ron Cohen.  I don't have extensive/elaborate plans but will respond at some level to feature requests (even if the best action is just to refer to other existing tools).  Ron is trying out restarts, which unless I am missing something are similar to SCF or relax restarts which Nexus already supports.

prckent:
Sounds Good To Me

Issue #1651, opened 2019-06-19
Cleaning up ParticleSet


ye-luo:
ParticleSet has multiple issues at the moment.
1. ParticleSet::addTable and copy constructor has hard-coded table 0 as the e-e distance table
2. distance tables are exposed as public member objects.
I would like to clean up all the related pieces and enhance security.
There will be multiple PRs before the goal is achieved.
3. Update the He example in the manual.

PR #1652, opened 2019-06-19
Remove DistanceTable::add

ye-luo:
Part of #1651 
ParticleSet::addTable becomes the only way adding distance tables to ParticleSet.
DistanceTable::add is removed.

In the use case
```
DistanceTableData* d_ii = DistanceTable::add(*ions, DT_SOA_PREFERRED);
```
Becomes
```
const int d_ii_ID = ions->addTable(*ions, DT_SOA_PREFERRED);
const auto& d_ii = *ions->DistTables[d_ii_ID];
```
Next PR will use ParticleSet::getDistTables(id) to get a constance reference of distance table and tables can be accessed only by index. The use of pointer to distance tables will be minimized as well.


jtkrogel:
You just changed something that touched code in a lot of places, e.g. the estimators.  How do you know this did not break currently working ones, or add an additional layer of breakage to those that currently do not?

prckent:
@jtkrogel Has an excellent question.

ye-luo:
@jtkrogel has a valid question. This PR doesn't have significant impact on the code but I'm working on the next PR and I did see breakage in following changes. Our test coverage is significantly lagging behind.

I removed the files with header file change only. All the following files has literal changes.
```
 a/src/QMCHamiltonians/CoulombPBCAA.cpp b/src/QMCHamiltonians/CoulombPBCAA.cpp
 a/src/QMCHamiltonians/CoulombPBCAA.h b/src/QMCHamiltonians/CoulombPBCAA.h
 a/src/QMCHamiltonians/CoulombPBCAB.cpp b/src/QMCHamiltonians/CoulombPBCAB.cpp
 a/src/QMCHamiltonians/CoulombPBCAB.h b/src/QMCHamiltonians/CoulombPBCAB.h
* a/src/QMCHamiltonians/CoulombPotential.h b/src/QMCHamiltonians/CoulombPotential.h
** a/src/QMCHamiltonians/DensityEstimator.cpp b/src/QMCHamiltonians/DensityEstimator.cpp
 a/src/QMCHamiltonians/ECPComponentBuilder.h b/src/QMCHamiltonians/ECPComponentBuilder.h
* a/src/QMCHamiltonians/EnergyDensityEstimator.cpp b/src/QMCHamiltonians/EnergyDensityEstimator.cpp
 a/src/QMCHamiltonians/ForceBase.cpp b/src/QMCHamiltonians/ForceBase.cpp
 a/src/QMCHamiltonians/ForceCeperley.cpp b/src/QMCHamiltonians/ForceCeperley.cpp
 a/src/QMCHamiltonians/ForceChiesaPBCAA.cpp b/src/QMCHamiltonians/ForceChiesaPBCAA.cpp
** a/src/QMCHamiltonians/LatticeDeviationEstimator.h b/src/QMCHamiltonians/LatticeDeviationEstimator.h
* a/src/QMCHamiltonians/LocalCorePolPotential.cpp b/src/QMCHamiltonians/LocalCorePolPotential.cpp
* a/src/QMCHamiltonians/LocalCorePolPotential.h b/src/QMCHamiltonians/LocalCorePolPotential.h
 a/src/QMCHamiltonians/LocalECPotential.cpp b/src/QMCHamiltonians/LocalECPotential.cpp
* a/src/QMCHamiltonians/LocalECPotential.h b/src/QMCHamiltonians/LocalECPotential.h
* a/src/QMCHamiltonians/LocalMomentEstimator.cpp b/src/QMCHamiltonians/LocalMomentEstimator.cpp
* a/src/QMCHamiltonians/LocalMomentEstimator.h b/src/QMCHamiltonians/LocalMomentEstimator.h
* a/src/QMCHamiltonians/MPC.cpp b/src/QMCHamiltonians/MPC.cpp
* a/src/QMCHamiltonians/MPC_CUDA.cpp b/src/QMCHamiltonians/MPC_CUDA.cpp
* a/src/QMCHamiltonians/MomentumEstimator.cpp b/src/QMCHamiltonians/MomentumEstimator.cpp
* a/src/QMCHamiltonians/NonLocalECPComponent.cpp b/src/QMCHamiltonians/NonLocalECPComponent.cpp
* a/src/QMCHamiltonians/NonLocalECPotential.cpp b/src/QMCHamiltonians/NonLocalECPotential.cpp
* a/src/QMCHamiltonians/PairCorrEstimator.cpp b/src/QMCHamiltonians/PairCorrEstimator.cpp
* a/src/QMCHamiltonians/PulayForce.cpp b/src/QMCHamiltonians/PulayForce.cpp
 a/src/QMCHamiltonians/QMCHamiltonian.cpp b/src/QMCHamiltonians/QMCHamiltonian.cpp
* a/src/QMCHamiltonians/SpaceGrid.h b/src/QMCHamiltonians/SpaceGrid.h
*** a/src/QMCHamiltonians/StressKinetic.cpp b/src/QMCHamiltonians/StressKinetic.cpp
* a/src/QMCHamiltonians/StressPBC.cpp b/src/QMCHamiltonians/StressPBC.cpp
*** a/src/QMCHamiltonians/StressPBCAA.cpp b/src/QMCHamiltonians/StressPBCAA.cpp
*** a/src/QMCHamiltonians/StressPBCAB.cpp b/src/QMCHamiltonians/StressPBCAB.cpp
* a/src/QMCHamiltonians/ZeroVarianceForce.cpp b/src/QMCHamiltonians/ZeroVarianceForce.cpp

* no unit test
** no unit test but has integration test
*** currently not connected to the build. @rcclay prefers to keep these files at the moment.
```
Any idea on a strategy?

PDoakORNL:
I like the removal of a circular dependence between ParticleSet and DistanceTable and I think for this PR the test is basically compilation since this will fail to compile in any code that is not updated. No logic appears to be changed. I will comment on the test coverage for the rest of the series elsewhere.

However, I think the design of keeping a vector index as if it were a const handle has some potential issues.
Later on for efficiency someone decides to delete a table from the middle of distance tables then indexes become invalid or subtlely wong. It is not expressed in the code that this is forbidden. Of course the code is a rather a mess once you look at what ParticleSet::addTable actually does.

Solutions would include:
*  Big scary comments explaining why there is no removeTable (probably not worth it if the myDistTableMap optimization is retained)
*  Use a std::map<std::string, DistTableData*> instead of std::vector<DistTableData*> instead remove myDistTableMap, the code already appear to try to enforce/use as an optimization that the std::string key is unique.
This would be robust against later update of the ParticleSet DistanceTableData container.

PDoakORNL:
test_particle.cpp --> testParticleSet.cpp
Should be updated to access its DistTables through the getDistTables method in the tests.
It needs test cases added that insure that the ID mapping works. i.e. addTable keeping the ID then getTable via the ID get have correct values. Including tables that should be optimized to one table and tables that shouldn't.

I think unit tests need to be written for the Hamiltonians. But I think its mostly the compiler that's going to check this change. Make the DistTable container private which will break direct access. 

The Hamiltonians should really have at least some unit tests. Can we find a limited set of methods to test in addition to valid instantiation that would at least document correct code and output for them with respect to creating/referencing and using the ParticleSet.getDistanceTable(ID).  

Maybe there is a way to template, automate or macro it into a slightly less repetitive process.

ye-luo:
Yes. This PR is checked by compilers. I did the checking with GNU and Intel. 

The indexing scheme is used for ParticleSet and VirtualParticleSet(VP) in the following way.
When you have electron ParticleSet.
electron.DistTables[0] and [1] are e-e and e-I tables.
When an electron is moved to quadrature points, a VP is constructed and the tables of VP must be consistent as well.
VP.DistTables[0] and [1] are vp-e and vp-I tables. Then a lot of places evaluate(VP) can be directly used and achieves code reuse. This is an example.
https://github.com/QMCPACK/qmcpack/blob/ede0ab1afae75d18c42d847cfb6f82a396495aed/src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h#L422
This is what I need at the moment. I didn't use std::map a lot in the past, perhaps we can improve this.

At the moment, we don't allow removing distance tables. Every consumer, WaveFunctionComponent and QMCHamiltonianBase use addTable to query the table it needs and store the index as a const class member integer. This is safe.

The effort is to stop all the manipulation of DistTables outside ParticleSet.

ye-luo:
@PDoakORNL You read my brain.
The next PR protests the whole DistTables. addTable has been clean up.
https://github.com/ye-luo/qmcpack/tree/protect-DT

prckent:
So we merge before the nightlies go through and check in the morning.

Our CI should be good enough to catch the different build combinations, but we will see.

PDoakORNL:
So looks like it was fine

PR #1653, opened 2019-06-20
Nexus: Adding support for ghost atoms

shivupa:
Allows nexus to understand ghost atoms. 

Why this is necessary:
The converters from quantum chemistry codes (e.g. gaussian/gamess) don't currently know about ghost atoms either, but they will parse the correct basis functions/MO coefficients. These just get assigned to atoms of type "". Replacing these with something reasonable like "X" and trying to uses nexus post calculation for analysis results in
```
PhysicalSystem error:
    particle X is unknown
  exiting.
```
This fixes the above error so now QmcpackAnalyzer can be used on systems with ghost atoms.


markdewing:
okay to test

jtkrogel:
This looks like it should work, though permanently adding atom X to the periodic table is unusual.

Does the following work?
```
from physical_system import Matter,Ion
from qmcpack_analyzer import QmcpackAnalyzer

Matter.particle_collection.add_particle(Ion('X'))

qa = QmcpackAnalyzer('./your.in.xml')
qa.analyze()
```

shivupa:
```
Matter.particle_collection.add_particles(Ion('X',mass=0,charge=0,spin=0,protons=0,neutrons=0))
```
This does work. I could use this since I'm guessing ghost atoms aren't very commonly used.

jtkrogel:
I think having a way of supporting ghost atoms is a good idea.

One possibility is to add a function like this to physical_system.py
```
def ghost_atoms(*names):
    for name in names:
        Matter.particle_collection.add_particles(Ion(name,mass=0,charge=0,spin=0,protons=0,neutrons=0))
    #end for
#end def ghost_atoms
```
and use it outside
```
ghost_atoms('X')
qa = QmcpackAnalyzer(...)
```
or move its use into QmcpackAnalyzer instead
```
qa = QmcpackAnalyzer(...,ghost_atoms=['X'])
```

If having an atom X permanently available is more convenient, then I'm fine if you want to add this line
```
Matter.particle_collection.add_particles(Ion('X',mass=0,charge=0,spin=0,protons=0,neutrons=0))
```
after `Matter.set_particle_collection(Particles(plist))` in physical_system.  This way 'X' is always there as a vacuous ion, but it is not in the periodic_table which is queried for real data.

shivupa:
That makes more sense, and this way ghost atoms don't have to be named "X". I'll update this PR.

shivupa:
@jtkrogel How does this look?  I didn't love having to pop from kwargs since it is mutating the object, but it was necessary otherwise the `__init__` for QmcpackAnalysisRequest would break.

nexus/lib/qmcpack_analyzer.py
```diff
@@ -213,6 +214,18 @@ def __init__(self,arg0=None,**kwargs):
...
         #end if
 
+        if 'ghost_atoms' in kwargs:
+            ghosts = kwargs.pop('ghost_atoms')
+            if isinstance(ghosts,str):
+                ghost_atoms(ghosts)
+            #end if
+            elif isinstance(ghosts,list):
+                for center in ghosts:
+                    ghost_atoms(center)
```
jtkrogel: Prefer ghost_atoms(*ghosts) over the loop.

jtkrogel:
Just one small requested change, then good to go.

nexus/lib/qmcpack_analyzer.py
```diff
@@ -213,6 +214,18 @@ def __init__(self,arg0=None,**kwargs):
             del kwargs['analyze']
         #end if
 
+        if 'ghost_atoms' in kwargs:
+            ghosts = kwargs.pop('ghost_atoms')
+            if isinstance(ghosts,str):
+                ghost_atoms(ghosts)
+            #end if
```
jtkrogel: Remove #end if as it is joined with the following else.

shivupa:
Fixed!

jtkrogel:
Looks good.  Thanks.

PR #1656, opened 2019-06-24
Nexus: fix miscategorized starting_charge

jtkrogel:
The PWSCF input parameter "starting_charge" was miscategorized as a simple float, whereas it is really an array of floats.  This prevented appropriate input generation when using this parameter.

Other arrays that are dimensioned by ntyp were explicitly checked for correct categorization.


markdewing:
okay to test

PR #1659, opened 2019-06-24
Nexus: protect pseudopotential search

jtkrogel:
Nexus expects pseudopotential files to follow a generic naming format like [atomic symbol].anything.you.like.[file extension].  Previously, all items present in the pseudopotentials directory provided to the settings function via the "pseudo_dir" argument where expected to have this format.

Under the old behavior, uncaught crashes could occur when files or directories contained within the pseudopotentials directory did not follow this format.  This PR now explicitly neglects all sub-directories present (instead of demanding that they be files) and ignores file names that do not match the required patterns.  Descriptive log entries are written to stdout to inform the user about the neglected files/directories to inform them of the behavior.

Issue reported by Ron Cohen.


prckent:
Good to know this restriction to avoid breaking it in QMCPACK examples. e.g. Au_QMCPACK.xml, Au_QBOX.xml vs Au.QMC.xml & Au.QBOX.xml

jtkrogel:
I misspoke slightly: Au.QMCPACK.xml, Au_QBOX.xml, Au-QMC.xml all work (delimiter for atomic symbol can be ".", "_", or "-").  The atomic symbol is also case insensitive and can be followed by any number of integers preceding the delimiter (Au1.QMC.xml, Au2_QMC.xml).

ye-luo:
Okay to test

PR #1654, opened 2019-06-21
Protect DistTables in ParticleSet

ye-luo:
Part 2 of #1651
Updated ParticleSet::addTable() and secure DistTables manipulation in copy constructor.
Distance can only be access via index and const reference only from outside ParticleSet via ParticleSet::getDistTable().

Since this PR, the DistTables[0] is no more garanteed to be self-self type. Every distance table consumer should query the table ID by addTable() and use the table by getDistTable(id). I have fixed all the places with [0] hard coded.

I'm able to identify all the left AoS only features by greping addTable
```
QMCHamiltonians/EnergyDensityEstimator.cpp     :    dtable_index = Pdynamic->addTable(*Pstatic, DT_AOS);
QMCHamiltonians/LocalMomentEstimator.cpp       :  : ions(srcs), d_table_ID(elns.addTable(ions, DT_AOS))
QMCHamiltonians/StressPBC.cpp  :      ei_table_index(elns.addTable(ions, DT_AOS)), ee_table_index(elns.addTable(elns, DT_AOS))
QMCHamiltonians/LatticeDeviationEstimator.cpp  :      myTableID_(P.addTable(sP, DT_AOS)),
QMCHamiltonians/PulayForce.cpp :    : ForceBase(ions, elns), Ions(ions), Electrons(elns), Psi(psi), d_ei_ID(elns.addTable(ions, DT_AOS))
QMCHamiltonians/LocalCorePolPotential.cpp      :      d_ie_ID(els.addTable(ions, DT_AOS)), d_ii_ID(ions.addTable(ions, DT_AOS))
QMCHamiltonians/StressPBCAA.cpp:      Ps(ref), is_active(active), d_ee_ID_(ref.addTable(ref, DT_AOS))
QMCHamiltonians/StressPBCAB.cpp:      d_ei_ID_(elns.addTable(ions, DT_AOS))
QMCHamiltonians/ForceBase.cpp  :  : ForceBase(ions, elns), d_ei_ID(elns.addTable(ions, DT_AOS))
QMCHamiltonians/ZeroVarianceForce.cpp  :    : ForceBase(ions, elns), Ions(ions), Electrons(elns), Psi(psi), d_ei_ID(elns.addTable(ions, DT_AOS))
QMCWaveFunctions/IonOrbital.cpp:  myTableID      = ptcls.addTable(CenterRef, DT_AOS);
[...]

PR #1661, opened 2019-06-25
[WIP] unit test failure investigation

ye-luo:
I will run a few experiments.

### DIFF
--- a/src/QMCWaveFunctions/ExampleHeComponent.cpp
+++ b/src/QMCWaveFunctions/ExampleHeComponent.cpp
@@ -127,6 +127,11 @@ ExampleHeComponent::ValueType ExampleHeComponent::ratio(ParticleSet& P, int iat)
   double log_v_old = -Z * (r_old)-u_old;
   double log_v_new = -Z * (r_new)-u_new;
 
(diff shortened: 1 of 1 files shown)


ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
Issue captured by a CI failed run.
```
Checking
r12 old and new = 1.5652475842 1.5652475842
r old and new = inf inf
u old and new = -0.3474933237 -0.3474933237
log_v old and new = -inf -inf
```
ei_table has an issue. This is a bug in the AoS build but it doesn't affect real calculations.
ions.RSoA is updated by ions.update() only when the build is AoS, ions has tables and the first table is SoA. In a real simulation, ion-ion table is always added for Coulomb and RSoA is updated correctly.
In the case of unit test, ions particle set doesn't carry ion-ion table and thus RSoA is uninitialized. If the uninitialized values are 0, the unit test pass. If not, NaN is a frequent error.

I have already fixed this bug in #1654 when I was doing the cleaning work. I just added one more unit test to capture this case.

PR #1663, opened 2019-06-26
Enabling a target shift value in the adaptive optimizer

eneuscamman:
This commit allows the user to specify that the adaptive optimizer should choose the shift value closest to a target shift value so long as that shift's update's cost function value is not too much higher than that of the other shifts that could be chosen.  This helps avoid runaway increases in the adaptive shift which can effectively end an optimization prematurely.  By default, this behavior is disabled and the adaptive optimizer functions as before.


markdewing:
okay to test

src/QMCDrivers/QMCFixedSampleLinearOptimize.h
```diff
@@ -129,6 +129,10 @@ class QMCFixedSampleLinearOptimize : public QMCLinearOptimize, private NRCOptimi
   RealType max_relative_cost_change;
   ///max amount a parameter may change relative to current wave function weight
   RealType max_param_change;
+  /// the tolerance to cost function increases when choosing the best shift in the adaptive shift method
+  RealType cost_increase_tol;
+  /// the shift_i value that the adaptive shift method should aim for
+  RealType target_shift;
```
ye-luo: Since it is targeting shift_i, put an explicit name seens better target_shift_i

manual/opt.tex
```diff
@@ -251,6 +253,10 @@ \subsubsection{adaptive}
   \item \ixml{shift_s}.  This is the initial coefficient used to scale the overlap-based stabilizer.
                             More stable but slower optimization is expected with a large value.
                             The adaptive method will automatically adjust this value after each linear method iteration.
+  \item \ixml{target_shift}.  If set greater than zero, the adaptive method will choose the update whose shift_i value is closest to
```
ye-luo: Update here accordingly for the name change.

src/QMCDrivers/QMCFixedSampleLinearOptimize.h
eneuscamman: Change made and committed to my fork.

manual/opt.tex
eneuscamman: Manual updated accordingly.

prckent:
Thanks for also updating the manual. Would be good to have a small example after more experience is gained.

PR #1664, opened 2019-06-26
Reduce and rename makeMove interfaces

ye-luo:
All particle move routines are rename to avoid confusion.
Single particle move routines are tightened and all the computation part is moved to
computeNewPosDistTablesAndSK() which is "protected" in the class.


prckent:
What does makeMoveAndCheck actually check?

ye-luo:
The following lines are added to the ParticleSet.h
   * When a Lattice is defined, passing two checks makes a move valid.
   * outOfBound(displ): invalid move, if displ is larger than half, currently, of the box in any direction
   * isValid(Lattice.toUnit(activePos)): invalid move, if activePos goes out of the Lattice in any direction 
   * Note: activePos and distances tables are always evaluated no matter the move is valid or not.
More discussion on the logic is in #1668.

ye-luo:
Test this please.

Issue #975, opened 2018-08-10
Efficiency of 1RDM needs to be improved


jtkrogel:
As was done for n(k), more efficient ratio computation needs to be implemented for the 1RDM.

jtkrogel:
Satisfied by #1672.

PR #1672, opened 2019-06-27
Optimize ratio evaluation in 1RDM estimator

jtkrogel:
This PR addresses #975.

I did a series of performance profiling runs by rejigging the NiO performance tests to include the 1RDM.  These tests showed that a ~4x performance penalty is found for production system sizes when using 1 non-local sample per atom.  Higher sampling densities would result in proportional cost increases (2x more samples leads to 2x slowdown).  Below is a plot showing the BlockCPU time in DMC vs. number of atoms/samples relative to a run of the same size containing no estimator.

baseline_reltime_vs_atoms_samples

By inserting timers into the 1RDM eval, I was able to see the fraction of time spent on each sub-component.  Components include ratio eval, spo eval (labeled "basis" below), and matrix products.  As a function of system size, the ratio eval clearly dominates (approaches 100% of the overall time):

baseline_1rdm_fractions

With the performance improvements in this PR (solely due to Ye's evaluateRatiosAlltoOne function), the added cost over and estimator-free run drops to about 30% regardless of system size:

optimized_reltime_vs_atoms_samples

The balance between different evaluation components is now much closer with matrix products and spo evals becoming competitive with ratio evals as the leading source of efficiency loss:

optimized_1rdm_fractions

In production runs, the spo eval and matrix products are likely to become even more significant since the performance tests are limited to an artificially small basis that comprises just the occupied states.  In a production setting, spo evals are likely to grow by 2-3x and matrix products possibly by 4-10x.  The performance analysis will be redone in a more production setting once support has been added for independent spin up/down bases (which will further strain the spo eval).  In any event, the ratio contribution seems to now be handled.

### DIFF
--- a/src/QMCHamiltonians/DensityMatrices1B.cpp
+++ b/src/QMCHamiltonians/DensityMatrices1B.cpp
@@ -360,13 +360,15 @@ void DensityMatrices1B::initialize()
     normalize();
   }
 
-  eval_timer               = TimerManager.createTimer("DensityMatrices1B::evaluate", timer_level_medium);
-  gen_samples_timer        = TimerManager.createTimer("DensityMatrices1B::generate_samples", timer_level_fine);
-  gen_sample_basis_timer   = TimerManager.createTimer("DensityMatrices1B::generate_sample_basis", timer_level_fine);
-  gen_sample_ratios_timer  = TimerManager.createTimer("DensityMatrices1B::generate_sample_ratios", timer_level_fine);
-  gen_particle_basis_timer = TimerManager.createTimer("DensityMatrices1B::generate_particle_basis", timer_level_fine);
-  matrix_products_timer    = TimerManager.createTimer("DensityMatrices1B::evaluate_matrix_products", timer_level_fine);
-  accumulate_timer         = TimerManager.createTimer("DensityMatrices1B::evaluate_matrix_accum", timer_level_fine);
+  const TimerNameList_t<DMTimers>
+    DMTimerNames = {{DM_eval              , "DensityMatrices1B::evaluate"},
+                    {DM_gen_samples       , "DensityMatrices1B::generate_samples"},
+                    {DM_gen_sample_basis  , "DensityMatrices1B::generate_sample_basis"},
+                    {DM_gen_sample_ratios , "DensityMatrices1B::generate_sample_ratios"},
+                    {DM_gen_particle_basis, "DensityMatrices1B::generate_particle_basis"},
+                    {DM_matrix_products   , "DensityMatrices1B::evaluate_matrix_products"},
+                    {DM_accumulate        , "DensityMatrices1B::evaluate_matrix_accum"}};
+  setup_timers(timers,DMTimerNames,timer_level_fine);
 
   initialized = true;
 }
@@ -585,7 +587,7 @@ void DensityMatrices1B::warmup_sampling()
 
 DensityMatrices1B::Return_t DensityMatrices1B::evaluate(ParticleSet& P)
 {
-  eval_timer->start();
+  timers[DM_eval]->start();
   if (have_required_traces || !energy_mat)
   {
     if (check_derivatives)
@@ -597,7 +599,7 @@ DensityMatrices1B::Return_t DensityMatrices1B::evaluate(ParticleSet& P)
     else
       APP_ABORT("DensityMatrices1B::evaluate  invalid evaluator");
   }
-  eval_timer->stop();
+  timers[DM_eval]->stop();
   return 0.0;
 }
 
@@ -623,7 +625,7 @@ DensityMatrices1B::Return_t DensityMatrices1B::evaluate_matrix(ParticleSet& P)
   generate_sample_ratios(Psi_NM);     // conj(Psi ratio) : particles x samples
   generate_particle_basis(P, Phi_NB); // conj(basis)     : particles x basis_size
   // perform integration via matrix products
-  matrix_products_timer->start();
+  timers[DM_matrix_products]->start();
   for (int s = 0; s < nspecies; ++s)
   {
     Matrix_t& Psi_nm     = *Psi_NM[s];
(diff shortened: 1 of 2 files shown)


jtkrogel:
Should have mentioned: the new code passes the existing 1RDM VMC tests for diamond.

ye-luo:
Could you include your timers as well?
You can find examples in #1664.

src/QMCHamiltonians/DensityMatrices1B.cpp
```diff
@@ -1140,19 +1141,20 @@ void DensityMatrices1B::generate_sample_basis(Matrix_t& Phi_mb)
...
+    Pq.makeVirtualMoves(rsamples[m]);
+    Psi.evaluateRatiosAlltoOne(Pq, psi_ratios);
+
+    // collect ratios into per-species matrices
+    int p = 0;
+    for (int s = 0; s < nspecies; ++s)
     {
-      PosType& Rp = Pq.R[p];
-      for (int m = 0; m < samples; ++m, ++nm)
+      Matrix_t& P_nm = *Psi_nm[s];
```
ye-luo: Not necessary to fix now but keep in mind of replacing pointers with containers when you are refactoring old codes.

jtkrogel:
Sure, I will add the timers.

ye-luo:
I wrote a wiki page for adding timers https://github.com/QMCPACK/qmcpack/wiki/How-to-add-timers
Please update accordingly.

ye-luo:
use timer_level_fine only. Timers can nest even within the same level. medium is reserved for sections inside drivers.

prckent:
@ye-luo Please can your move the timers info to the manual and remove from the wiki. The how to add tests section can also be moved/merged. Reason: Trying to have a single place for development info, and this is currently the manual.

jtkrogel:
Is the use of enum + initializer function strictly necessary, or just use of timer_level_fine?  I'm missing the usefulness of the former here.

ye-luo:
@prckent 
@markdewing and I are still evolving the wiki page and then will move to the manual.

ye-luo:
@jtkrogel There is no initializer function. It is a definition of static const class member data.
Having an array of timers is better than a set of individual timers. Having enum is better than directly using integer indices.

Please do both the enum and timer_level_fine change.

jtkrogel:
@ye-luo The setup_timers function is what I meant by initializer function.  IMHO, this (enum, name map, timer list) just adds extra code and layers of naming (which reduces readability) over individual timers.  From this point of view, I see no advantage, but if it is the required pattern then I will follow it.

Your documentation does not specify the type definition of myTimers.  What is it?

markdewing:
For the timers, it feels slightly neater to have a single member variable storing all the timers, rather than a number of member variables.  This does come at the cost of a little bit more code and an extra enum.

ye-luo:
@jtkrogel I updated the wiki. Defining TimerNames to the class is only necessary when multiple constructors exist. In your case, it can be defined local to the constructor. It is cleaner to have enum and a vector of timers.

jtkrogel:
@ye-luo Done.

src/QMCHamiltonians/DensityMatrices1B.cpp
```diff
@@ -576,6 +587,7 @@ void DensityMatrices1B::warmup_sampling()
 
 DensityMatrices1B::Return_t DensityMatrices1B::evaluate(ParticleSet& P)
 {
+  timers[DM_eval]->start();
```
ye-luo: Please use RAII scoped timer.

src/QMCHamiltonians/DensityMatrices1B.cpp
```diff
@@ -874,6 +891,7 @@ DensityMatrices1B::Return_t DensityMatrices1B::evaluate_loop(ParticleSet& P)
 
 inline void DensityMatrices1B::generate_samples(RealType weight, int steps)
 {
+  timers[DM_gen_samples]->start();
```
ye-luo: Please use RAII scoped timer.

src/QMCHamiltonians/DensityMatrices1B.cpp
```diff
@@ -1128,39 +1147,45 @@ void DensityMatrices1B::get_energies(std::vector<Vector_t*>& E_n)
 
 void DensityMatrices1B::generate_sample_basis(Matrix_t& Phi_mb)
 {
+  timers[DM_gen_sample_basis]->start();
```
ye-luo: Please use RAII scoped timer.

src/QMCHamiltonians/DensityMatrices1B.cpp
```diff
@@ -1128,39 +1147,45 @@ void DensityMatrices1B::get_energies(std::vector<Vector_t*>& E_n)
...
   }
+  timers[DM_gen_sample_basis]->stop();
 }
 
 
 void DensityMatrices1B::generate_sample_ratios(std::vector<Matrix_t*> Psi_nm)
 {
-  int p = 0;
-  for (int s = 0; s < nspecies; ++s)
+  timers[DM_gen_sample_ratios]->start();
```
ye-luo: Please use RAII scoped timer.

src/QMCHamiltonians/DensityMatrices1B.cpp
```diff
@@ -1128,39 +1147,45 @@ void DensityMatrices1B::get_energies(std::vector<Vector_t*>& E_n)
...
       }
     }
   }
+  timers[DM_gen_sample_ratios]->stop();
 }
 
 
 void DensityMatrices1B::generate_particle_basis(ParticleSet& P, std::vector<Matrix_t*>& Phi_nb)
 {
+  timers[DM_gen_particle_basis]->start();
```
ye-luo: Please use RAII scoped timer.

src/QMCHamiltonians/DensityMatrices1B.h
```diff
@@ -138,6 +153,15 @@ class DensityMatrices1B : public QMCHamiltonianBase
 
   RandomGenerator_t* uniform_random;
 
+  NewTimer* eval_timer;
```
ye-luo: Remove all these no more used timers.

src/QMCHamiltonians/DensityMatrices1B.cpp
jtkrogel: Please update your document ("recommended" -> "required when possible").  It currently lists both methods as options.

ye-luo:
Okay to test

prckent:
Really good to see. Dramatic speedup.

PR #1665, opened 2019-06-26
modifications for supermuc and bohr in qe

recohen:
changes for superMUC computer and use of bohr for lattice as required now in quantum espresso

PR #1678, opened 2019-06-28
Nexus: update tests for recent changes

jtkrogel:
Update test reference data to reflect changes made by PR #1665.  PR #1665 updated the definition for the SuperMUC machine (corrected number of cores) which is now accomodated in the tests.  That PR additionally changed the default option for PWSCF card inputs from "alat" to "bohr".  This has also been updated in the test data.

Additional options were also added for input generation to specify any kind of card option, which was used to regain backwards compatibility with older inputs.


markdewing:
okay to test

PR #1679, opened 2019-06-29
Update multi and cdash bug fixes

### LINKED ISSUE #1674: Broken Intel 2018 complex build due to AFQMC KP3IndexFactorization problems
https://cdash.qmcpack.org/CDash/viewBuildError.php?buildid=64229

### LINKED ISSUE #1675: Broken Clang7 Complex build due to AFQMC THCOpps failed assert
https://cdash.qmcpack.org/CDash/viewBuildError.php?buildid=64255

### LINKED ISSUE #1677: Broken Clang6-CUDA10-Complex build due to AFQMC MPI3 issues
https://cdash.qmcpack.org/CDash/viewBuildError.php?buildid=64336

Another breakage. There is a chance that this is related to the build configuration.

fdmalone:
Update boost multi array library and fix conj issue. Should resolve #1674 resolve #1675 and resolve #1677 (@mmorale3 ?)


prckent:
OK to test

prckent:
(Multi has an FFTW interface ??!!  A topic for another day, but it seems out of place given multi's supposed mission. Plus surely an FFT abstraction is needed...)

prckent:
@fdmalone this LGTM but since you requested a review from @mmorale3 I'll wait for the review+approval before merging.

mmorale3:
the FFTW interface is some experimentation with possible extensions of the library. The point is to let the library handle the backend implementation for things like fftw, hlas, lapack, etc, with full gpu support transparent to the user like afqmc does

correaa:
@prckent 

There are two separate references to fftw in multi, one is just to show how to use Multi with C-interfaces the right way. https://gitlab.com/correaa/boost-multi/blob/master/tests/array_legacy_c.cpp the other is the "adaptor".

I struggled a lot with adding the adaptors, ideally they should be separate libraries or a huge "example case". 

The main purpose of the adaptors is  to define an interface for common system libraries, the second purpose is to have a default fallaback cpu-backend. I agree that Multi is not for that, but realistically, for lack of a better place I added the adaptors and the cpu backends because a hand made use of Multi in this context will likely end up in an underutilization of Multi.

PR #1680, opened 2019-06-29
New Functor for Short Ranged en Jastrow

eneuscamman:
I have implemented a new functor intended to act as an optimizable, short-ranged one-body Jastrow factor for correcting Gaussian orbitals in the cusp region.  Unlike the pre-computed cusp correction, this functor can respond to changes in the wave function during VMC optimization, which we expect to be important as we begin to explore core excitations in first row atoms.  With a core excitation, the electron density near the nucleus changes dramatically, and the resulting orbital relaxations should be accompanied by the ability to change the detailed behavior near the nucleus while maintaining exact cusp conditions.  This functor seeks to provide this capability.


markdewing:
okay to test

manual/opt.tex
```diff
@@ -247,16 +247,16 @@ \subsubsection{adaptive}
 
 Additional information:
 \begin{itemize}
-  \item \ixml{shift_i}.  This is the initial coefficient used to scale the diagonal stabilizer.
+  \item \ixml{shift\_i}.  This is the initial coefficient used to scale the diagonal stabilizer.
```
ye-luo: backslash is not needed in \ixml{}

ye-luo: Please do not try to fix this issue here. We will do that with a separate PR.

ye-luo: I just got this fixed already. git pull before you continue other changes.

src/QMCWaveFunctions/Jastrow/ShortRangeCuspFunctor.h
```diff
@@ -0,0 +1,622 @@
...
+            sstr << ID_B << "_" << i;
+            myVars.insert(sstr.str(), B.at(i), Opt_B, optimize::OTHER_P);
+          }
+        } else if ( optimize == "no" ) {
+          Opt_B = false;
+        } else if ( optimize != "string_not_set" ) {
+          PRE.error("Unrecognized value for \"optimize\". Should be either yes or no", true);
+        }
+
+      }
```
prckent: Add an abort here if the cname is unrecognized i.e. There is a typo.

markdewing:
test_wf.cpp contains some tests for a Pade functor.   Some of the values for comparison were computed by hand.
test_user_jastrow.cpp tests different functions and derivatives for consistency, but has no comparison to an external or hand-computed reference.  Modifying it to work with another functor should only require some adjustments to set up the variational parameters.

eneuscamman:
OK, I'll try to see to these things sometime this week...

src/QMCWaveFunctions/Jastrow/ShortRangeCuspFunctor.h
eneuscamman: So far it is experimental, but in early testing it does a better job than using a short-ranged, finely spaced spline form.  As we learn more, the preferred functional form for these purposes may evolve...

eneuscamman:
I've added the unit test and the extra error trap that Paul wanted.  Let me know if there's anything else to do.

prckent:
@eneuscamman Can you remake opt.tex from the current version on develop or otherwise fix the conflict? opt.tex was edited and has modest changes throughout. You should be able to put in your new section again easily.

eneuscamman:
@prckent The conflict in the manual optimization section is now resolved.

eneuscamman:
@ye-luo The unit test is ready.  I think this is ready to merge.

ye-luo:
Test this please

PR #1681, opened 2019-06-30
Make PyscfToQmcpack.py python3 compatible

aannabe:
make PyscfToQmcpack python3 compatible

### DIFF
--- a/src/QMCTools/PyscfToQmcpack.py
+++ b/src/QMCTools/PyscfToQmcpack.py
@@ -10,7 +10,7 @@
 ## File created by: Anouar Benali, benali@anl.gov, Argonne National Laboratory
 #######################################################################################
 
-
+from __future__ import print_function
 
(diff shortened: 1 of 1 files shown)


prckent:
Any idea what is going on in the "Strange If Statement"? Why do we even print a message?...

anbenali:
It should be straight forward and the if statement was there before we
could handle complex. I am rewriting and cleaning  a bit that converter but
will get back to it once my mother leaves next week... between baby sitting
and mother sitting I am a bit overwhelmed:)

prckent:
@aannabe  Nice to have such a simple upgrade to Python3. Happy to merge.

ye-luo:
Does python 2 complain? or we don't need python 2 in this case.

markdewing:
Good point @ye-luo , the code needs 'from __future__ import print_function' to be most compatible between Python 2 and 3.
 
The change probably works under Python 2 without that, but it prints a tuple containing the string.    Which may not be a problem in this case.  
Adding the import will make it easier to avoid accidentally adding Python 2 print statements.

markdewing:
okay to test

PR #1682, opened 2019-06-30
Fix Clang warning and LocalMomentEstimator crash

ye-luo:
In src/QMCDrivers/QMCCostFunctionBase.cpp, I only add parenthesis in line 274 and 279 but the formatting showed a lot of change.

### DIFF
--- a/src/QMCHamiltonians/LocalMomentEstimator.cpp
+++ b/src/QMCHamiltonians/LocalMomentEstimator.cpp
@@ -20,7 +20,7 @@
 namespace qmcplusplus
 {
 LocalMomentEstimator::LocalMomentEstimator(ParticleSet& elns, ParticleSet& srcs)
-  : ions(srcs), d_table_ID(elns.addTable(ions, DT_AOS))
+  : ions(srcs), d_table_ID(elns.addTable(srcs, DT_AOS))
 {
   int num_species = elns.groups();
   const SpeciesSet& e_species(elns.getSpeciesSet());
--- a/src/QMCDrivers/QMCCostFunctionBase.cpp
+++ b/src/QMCDrivers/QMCCostFunctionBase.cpp
@@ -149,8 +149,8 @@ QMCCostFunctionBase::Return_rt QMCCostFunctionBase::computedCost()
 {
   //   Assumes the Sums have been computed all ready
   //Estimators::accumulate has been called by correlatedSampling
-  curAvg_w        = SumValue[SUM_E_WGT] / SumValue[SUM_WGT];
-  curVar_w        = SumValue[SUM_ESQ_WGT] / SumValue[SUM_WGT] - curAvg_w * curAvg_w;
+  curAvg_w         = SumValue[SUM_E_WGT] / SumValue[SUM_WGT];
+  curVar_w         = SumValue[SUM_ESQ_WGT] / SumValue[SUM_WGT] - curAvg_w * curAvg_w;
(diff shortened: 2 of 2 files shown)


src/QMCHamiltonians/LocalMomentEstimator.cpp
```diff
@@ -20,7 +20,7 @@
 namespace qmcplusplus
 {
 LocalMomentEstimator::LocalMomentEstimator(ParticleSet& elns, ParticleSet& srcs)
-  : ions(srcs), d_table_ID(elns.addTable(ions, DT_AOS))
+  : ions(srcs), d_table_ID(elns.addTable(srcs, DT_AOS))
```
prckent: This looks consequential. Are there any tests on this code?

ye-luo: We don't have any test for this class. I have no idea what this class does.
The initial code was
```
LocalMomentEstimator::LocalMomentEstimator(ParticleSet& elns, ParticleSet& srcs)
  : ions(srcs)
{
  d_table_ID = elns.addTable(ions, DT_AOS);
}
```
I change the code to 
```
LocalMomentEstimator::LocalMomentEstimator(ParticleSet& elns, ParticleSet& srcs)
  : ions(srcs), d_table_ID(elns.addTable(ions, DT_AOS))
{
}
```
This was wrong because the order of initialization list doesn't affect the actual order of initialization which is determined by the order of declaration. Clang reports warn of using uninitialized objects.

In the future, we need to turn on compiler warning or even error all unused and uninitialized variables.

prckent: Paging @jtkrogel

prckent: It is in the manual. Code is clearly AoS only. Maybe there is a test (I can't currently check)

Best not to change the code, assuming the old way compiles.

Once the new nightlies have been stabilized we will add a build with more warnings and checking enabled.

prckent: There is a test in tests/estimator/latdev. pandas has to be installed for it to be enabled & run.

Since Jaron is out I suggest simply not modifying this code.

ye-luo: @prckent I can't get your point for not fixing this. The warning is accurate and surfaces a bug I just introduced a few days ago. I hit similar issues when I was cleaning up addTable.
I made a test by manually adding the xml entry to activate this estimator and got a segfault which disappeared after this fix. I think we need this PR for solving the segfault. I have no idea what this estimator exactly does and hence can't construct a real test to check the result numbers.

What did you mean "in the manual" and "tests/estimator/latdev"? I didn't find any manual section or test files correspond to this estimator particularly.

prckent: I was looking at the latticedeviation and not localmoment estimator.

PDoakORNL: @ye-luo I have to agree here. I'd say fixing the initialization argument from ions to srcs just restores the original behavior from ~2 weeks back. There the distance table didn't see the ions particleset until it was in the constructor body of LocalMomentEstimator so this didn't occur. So this should be merged.

The sanest test here would be to get to the point where we can watch for compiler warnings in the CI. Probably for both gcc and clang.

PR #1684, opened 2019-06-30
Use shared_ptr in adopters and update and document MultiBspline and MultiBspline1D

### LINKED ISSUE #954: Improve robustness and documentation of spline2
Re: #953 

Document restrictions/assumptions of spline2 at the function level. Add protections (assertions etc.) where practical.

ye-luo:
In MultiBspline1D, spline_m has been replaced with a pointer which is the way used in MultiBspline (3D).
Because spline_m itself contains an extra pointer and spline_m is a C structure, the full destruction of spline_m and spline_m->coefs needs to be at MultiBspline1D and MultiBspline. Just consider these two Multi classes as containers. These containers are strictly not copyable and assignable.

All the spline adopter class are cleaned up as well in this PR. The old code uses two pointers (SplineInst and MultiSpline) to distinguish the master thread which is responsible for the destruction of SplineInst. This PR introduce std::shared_ptr to manage the SplineInst pointer.

Documentations are added to MultiBspline and MultiBspline1D. Thus close #954.

In the near future, all the C struct will be replaced with C++ and QMCPACK will be C++ only. We did that in miniQMC already.


src/QMCWaveFunctions/BsplineFactory/SplineC2ROMP.h
```diff
@@ -263,7 +263,7 @@ struct SplineC2ROMP : public SplineAdoptorBase<ST, 3>
   ///GGt data pointer
   const ST* master_GGt_ptr;
 
-  SplineC2ROMP() : BaseType(), nComplexBands(0), SplineInst(nullptr), MultiSpline(nullptr)
+  SplineC2ROMP() : Base(), nComplexBands(0), SplineInst(nullptr), MultiSpline(nullptr)
```
PDoakORNL: Did you overlook changing this to
```C++
SplineC2ROMP() : Base(), nComplexBands(0)
```

PDoakORNL: It also look like to me if the object is constructed through the copy constructor MultiSpline will never be valid.  If it is valid you'll call delete on a std::shared_ptr which you should not.

ye-luo:
Hold a second. I need to change to smart pointer in SplineC2ROMP.h as well.

src/spline2/MultiBspline.hpp
```diff
@@ -27,10 +27,12 @@
...
-   * @tparam T the precision of splines
-   * @tparam ALIGN the alignment of the orbital dimention
-   * @tparam ALLOC memory allocator
-   */
+ * @tparam T the precision of splines
+ * @tparam ALIGN the alignment of the orbital dimention
+ * @tparam ALLOC memory allocator
+ *
+ * This class contains a pointer to a C object, copy and assign of this class is forbidden.
+ */
```
PDoakORNL: This should be a class
private:
SplineType* spline_m;
public:
SplineType* getSplinePtr();

Will prevent issue like in SplineC2R.

src/spline2/MultiBspline1D.hpp
```diff
@@ -30,48 +32,59 @@ struct MultiBspline1D
   ///define the real type
   using real_type = typename bspline_traits<T, 1>::real_type;
   ///actual einspline multi-bspline object
-  SplineType spline_m;
+  SplineType *spline_m;
```
PDoakORNL: as with MultiBspline  why not a class

src/QMCWaveFunctions/BsplineFactory/SplineC2ROMP.h
ye-luo: Not done yet. I was waiting for the other PR touching this class. Cleaning it now.

ye-luo: Done.

src/spline2/MultiBspline.hpp
ye-luo: Good suggestion. Done. "Will prevent issue like in SplineC2R." What is the issue you refer to?

src/spline2/MultiBspline1D.hpp
ye-luo: Done.

src/spline2/MultiBspline.hpp
PDoakORNL: The assignment of the extra Multipline pointer wouldn't have compiled. Takes a bit more obliviousness to take and keep an extra copy of the pointer in new code.

PR #1685, opened 2019-07-01
Fix broken LaTeX in manual optimization section

prckent:


prckent:
Why did you restore the ixml tags? I intentionally removed those since(1) we use texttt elsewhere (2) each additional tag in the latex will make a future conversion to markdown harder.

ye-luo:
We have most places \ixml when listing an xml tag. \texttt is used only in the tables. Base on manual Chapter 24 \ixml is preferred over \textt unless I misunderstood something.

prckent:
OK.

PR #1688, opened 2019-07-01
Use standard cmake find package for libxml2

### LINKED ISSUE #1687: Replace FindLibxml2QMC.cmake by standard cmake find_package
Support for libxml2 is much better than when libxml2 integration was first implemented and I suggest/question whether our custom code is needed any more. 

- Directly calling find_package(LibXml2) worked on both oxygen and summit and correctly picked up e.g. spack installed versions of LibXml2, unlike the existing code.
- Nearly all our build recipes use the default LibXml2 location, which the CMake package should pickup. The Blue Gene toolchains are the only active examples with customization, but the existing cmake_find_root_path should continue to work.

prckent:
Use the standard cmake find_package implementation for libxml2. Tested on oxygen, summit, cori.

Closes #1687


prckent:
As sign of this working will be the appearance of libxml2 2.9.1 and 2.9.9 in the builds for older and newer compiler versions in the nightlies, respectively.

prckent:
Test this please

(DNS failure in rhea-cpu)

ye-luo:
I think LIBXML2_HOME is no more effective. Could you update the README.md?
LIBXML2_INCLUDE_DIR and LIBXML2_LIBRARY are the standard solutions.

prckent:
Done. Good catch.

manual/installation.tex
```diff
@@ -459,8 +457,8 @@ \subsection{Build scripts}
 
 cmake                                                \
   -D CMAKE_BUILD_TYPE=Debug                         \
-  -D Libxml2_INCLUDE_DIRS=/usr/include/libxml2      \
-  -D Libxml2_LIBRARY_DIRS=/usr/lib/x86_64-linux-gnu \
+  -D LIBXML2_INCLUDE_DIR=/usr/include/libxml2      \
+  -D LIBXML2_LIBRARY=/usr/lib/x86_64-linux-gnu \
```
ye-luo: I checked that LIBXML2_LIBRARY is not a path but the exact library.
For example.
-D LIBXML2_LIBRARY=/usr/lib64/libxml2.so

The above
`Library location for libxml2` probably is not clear enough.

prckent: Fixed

prckent:
Confirming that this worked in the nightlies e.g.
https://cdash.qmcpack.org/CDash/buildSummary.php?buildid=64826
```
-- Found LibXml2: /home/pk7/apps/spack/opt/spack/linux-rhel7-x86_64/gcc-8.3.0/libxml2-2.9.9-jdohhm4pev3mjpnsp4i6vejwibjdgadu/lib/libxml2.so (found version "2.9.9") 
```
https://cdash.qmcpack.org/CDash/buildSummary.php?buildid=65010
```
-- Found LibXml2: /home/pk7/apps/spack/opt/spack/linux-rhel7-x86_64/gcc-5.5.0/libxml2-2.9.1-rp5bnl3dku7sp7rfywset6ldnams2eyj/lib/libxml2.so (found version "2.9.1")
```

Issue #1061, opened 2018-09-11
Complete list of tests needed


jtkrogel:
Many tests have been added, but there still remain many features that need to have tests added.  The list below is meant to capture all features that do not yet have a test, or known breakages that do not yet have a corresponding test so that corresponding fixes can be pursued and validated.

Real space estimator tests needed:
- [ ] Chiesa kinetic energy correction (PR: , see also #1003 )
- [ ] MPC potential energy correction (PR: , see also #1002 )
- [ ] momentum distribution (PR: , see also #627 )
- [ ] pair correlation function (PR: , see also #396 )
- [ ] structure factor--SkEstimator (PR: , see also #1000 )
- [ ] structure factor--StaticStructureFactor estimator, spin resolved (PR: , see also #999 )
- [ ] structure factor--SkAllEstimator, finite size corrections (PR: , see also #1001 )
- [ ] energy density matrix (PR: , see also #981, #983 )
- [ ] traces (PR: )
- [ ] forward walking (PR: , see also #1004 )
- [ ] core polarization potential (PR: , see also #1005 )
- [ ] force estimators in open BC's (PR: , see also #332 )
- [ ] force estimators in periodic BC's (PR: , see also #332 )

Real space wavefunction tests needed:
- [ ] excited states in Slater determinant (PR: , see also #740 )
- [ ] matrix tiling of B-spline orbitals (PR: , see also #575 )
- [ ] quasi-2D boundary conditions (?) (PR: , see also #578 )

Real space method tests needed:
- [ ] pade2 Jastrow optimization (PR: , see also #412)
- [ ] orbital optimization with Bspline orbitals (PR: , see also #637, #638)
- [ ] orbital optimization demonstrating incorrect energy with LCAO (PR: , see also #638)
- [ ] CSVMC with openmp enabled (PR: , see also #362)
- [ ] RMC test demonstrating BlockWeight problem (PR: , see also #180)
- [ ] DMC test demonstrating stochastic reconfiguration problem (PR: , see also #18 )
- [ ] DMC restart test (PR: , see also #798 )
- [ ] DMC size consistent T-moves with splines (PR: , see also #630 )

Real space deterministic tests needed:
(plan out first set of tests to add here, see also #704 )

Auxiliary field estimator tests needed:

Auxiliary field method tests needed:

jtkrogel:
@rcclay are there other RMC features that aren't covered yet?

[...]

Issue #740, opened 2018-04-04
Tests for excitations and ground state modifications needed


prckent:
I don't believe our multideterminant tests fully cover the common cases of addition or subtraction of an electron for EA/IP calculations. The multideterminant and configuration interaction tests largely cover the case of creating an exciton, but an explicit promotion test would not be amiss, particularly with splines. e.g. Using carbon diamond 1x1x1.

If we conclude that these cases are fully covered I will close the issue.

PR #1690, opened 2019-07-02
Add deterministic tests for excited state promotion

Hyeondeok-Shin:
This is a part of the complete test list #1061 and may relevant to #740.

  Deterministic tests for the excited state are added to diamondC_1x1x1_pp and diamondC_2x1x1_pp testset. VMC & DMC tests are added and an electron at Gamma point is promoted.
  This test will provide a simple and quick statistical test for the excited state for a solid system.
   
  References are generated in Intel Xeon with gcc 6.2.0 and tests were passed with intel19 and clang 7.0 compiler.
  Tolerances for the mixed precision are preliminary set as the values for the ground state, and they will be adjusted after checking CDash if it is needed.

  Hyeondeok Shin


prckent:
Thanks. Important to get some coverage of this functionality.

* Do we have integration tests for these? A quick grep in qmcpack/tests suggests not. Do you have the reference data to hand? We can create a new issue on this topic.
* (General point) When creating tests we should think about the code coverage. In this case a DMC test is not strictly necessary since the QMC drivers have no idea whether they are running a ground or excited state calculation. i.e. VMC should be enough. For some properties there might be a case to make that the DMC is more sensitive to mistakes in the wavefunction and hence a test is worthwhile. Now that we have them we should keep them.

prckent:
Changed the PR name to be more accurate. This is important for writing the changelog.

prckent:
OK to test

prckent:
Approving. At this point the CPU CI runs have passed, indicating that the CMakeLists.txt is not broken.

ye-luo:
@prckent There is no statistical test in this PR. @Hyeondeok-Shin commented on the #1691 that these tests will be added in 1-2 days.

PR #1693, opened 2019-07-02
Two year rule and software versions in README.md

prckent:
Write down the "two year rule" and fix typos and leftovers in the README.

The list of tested software versions corresponds exactly to what is tested on oxygen. This is now done in an easily updatable manner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ performance and easiest configuration.
 Nightly testing currently includes the following software versions on x86:
 
 * Compilers
-  * GCC 8.3.0, 7.4.0, 5.50
+  * GCC 8.3.0, 7.4.0, 5.5.0
   * Clang/LLVM 7.0.1, 6.0.1, 4.0.1
   * Intel 2018.3
   * PGI 19.4
--- a/README.md
+++ b/README.md
@@ -1,29 +1,51 @@
 # Getting and building QMCPACK
 
- Obtain the latest release or development copy from http://www.qmcpack.org
+ Obtain the latest release from https://github.com/QMCPACK/qmcpack/releases or clone the development source from
+ https://github.com/QMCPACK/qmcpack. 
 
 # Prerequisites
 
- * C++ 14 and C99 capable compilers. e.g. gcc 5.0, Intel 2018, clang 3.5 or newer. 
+ * C++ 14 and C99 capable compilers. 
  * CMake, build utility, http://www.cmake.org
- * BLAS/LAPACK, numerical library, use platform-optimized libraries
- * Libxml2, XML parser, http://xmlsoft.org/
+ * BLAS/LAPACK, numerical library. Use platform-optimized libraries.
+ * LibXml2, XML parser, http://xmlsoft.org/
  * HDF5, portable I/O library, http://www.hdfgroup.org/HDF5/
- * BOOST, peer-reviewed portable C++ source libraries, http://www.boost.org , version 1.61.0 or higher.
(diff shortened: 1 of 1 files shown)


prckent:
Minor edits should be made directly to the PR for efficiency. Larger changes can be discussed.

prckent:
I will update the manual after this is merged.

prckent:
Release dates:

boost_vnew=1.70.0 # Released 2019-04-12
boost_vold=1.61.0 # Released 2016-05-13

gcc_vnew=8.3.0 # Released 2019-02-22 
gcc_vold=5.5.0 # Released 2017-10-27
gcc_vintel=7.4.0 # Released 2018-12-06

hdf5_vnew=1.10.5 # Releeased 2019-02-28
hdf5_vold=1.8.19 # Released 2017-06-16

cmake_vnew=3.14.4 # Released 2019-05-15
cmake_vold=3.8.2 # Released 2017-05-31

ompi_vnew=4.0.1 # Released 2019-03-26
ompi_vold=2.1.1 # Released 2017-05-10

libxml2_vnew=2.9.9 # Released 2019-01-03 
libxml2_vold=2.9.1  #Released 2013-04-19

fftw_vnew=3.3.8 # Released 2018-05-28
fftw_vold=3.3.4 # Released 2014-03-16

llvm_vnew=7.0.1 # Released 2018-12-21
llvm_vold=4.0.1 # Released 2017-07-04
llvm_vcuda=6.0.1 # Released 2018-07-05

README.md
```diff
@@ -1,29 +1,51 @@
...
+We aim to support open source compilers and libraries released within two years of each QMCPACK release. Use of software versions over
+two years old may work but is discouraged and untested. Proprietary compilers (Intel, PGI) are generally supported over the same
+period but may require use of an exact version. We also aim to support the standard software environments on Summit at OLCF, Theta
+at ALCF, and Cori at NERSC. Use of the most recently released compilers and library versions is particularly encouraged for highest
+performance and easiest configuration.
+
+Nightly testing currently includes the following software versions on x86:
+
+* Compilers
+  * GCC 8.3.0, 7.4.0, 5.50
```
ye-luo: The two year rule for compilers are applied to the major release number. 5.5 is checked as 5.1 (first 5) which should be 4 years old. Minor releases are only bug fixes but we need major releases for features.

grahamlopez: FYI, this is not how versioning is typically done. From semver.org:

    MAJOR version when you make incompatible API changes,
    MINOR version when you add functionality in a backwards-compatible manner, and
    PATCH version when you make backwards-compatible bug fixes.

I didn't know we only applied the "two year rule" to major versions; that seems like a loss to me.

ye-luo: https://en.cppreference.com/w/cpp/compiler_support
Since gcc 5 and clang 4 I rarely see minor version giving new C++ feaures.
Supporting new C++ standard features seems considered as "incompatible API changes".
If we use a feature in gcc 6, it probably doesn't exist in gcc 5.5
We can set a finer policy per compiler.

prckent: Clearly we could choose to finesse the definitions in future. For now I don't think it matters. 

I have elected to avoid specifying exact software *combinations* or specifying which standard C++ libraries that we test with since it more complicated than necessary for now.  If someone asks or runs into trouble we can do so in future. I can foresee needing to specify other tested platforms and known not working software combinations.

Intel2018 is running fine (at least where the QMCPACK complex build is not broken by AFQMC...).

Issue #1649, opened 2019-06-19
Counting Jastrow Code


PDoakORNL:
Since this code is getting touched again the following should be resolved.

##QMCWaveFunctions/CountingFunctor.h
Mismatched class declaration and file name
No Unit Test.
typdef vs using
Inherits from QMCTraits
Has template parameter it doesn't use.

##QMCWaveFunctions/CountingRegion.h
Mismatched class declaration and file name
No Unit Test.
typdef vs using
Inherits from QMCTraits
Has template parameter it doesn't use.

ye-luo:
@bwvdg Could you have a look?

bwvdg:
I made the appropriate code convention changes. I'm currently writing the unit tests for these classes and will submit a pull request soon.

ye-luo:
Unit test inplace

PR #1695, opened 2019-07-03
Counting Jastrow Update

bwvdg:
- fixed segfaults from new VariableSet::print
- made formatting/convention changes according to #1649 
- added GaussianFunctor and CountingJastrow unit tests.
- optimized variables are now output to *.opt.xml
- added automatic countingjastrow/*.opt.xml xml generation based on a source ParticleSet
- improved printing and removed unused/untested code.

There are some somewhat questionable additions, including VariableSet::getLoc and QMCCostFunctionBase::InitVariables. These were added to work around VariableSet's sometimes uneven handling of variables depending on their optimization flag.

Particular variables of the CountingJastrow are known to be redundant, but implementing these as hard constants makes the parameter input very inflexible. Handling these unoptimized parameters alongside the optimizable ones reasonably made these additions necessary.


src/QMCDrivers/QMCCostFunctionBase.h
```diff
@@ -250,6 +252,8 @@ class QMCCostFunctionBase : public CostFunctionBase<QMCTraits::RealType>, public
    * is normally the same as OptVariables.
    */
   opt_variables_type OptVariablesForPsi;
+  // unchanged initial checked-in variables
+  opt_variables_type InitVariables;
```
ye-luo: it concerns me that this extra InitVariables adds more memory and bookkeeping. For other parameters checked in from the XML input and output to the opt.xml don't need initVariables. Why it is necessary to have InitVariables particularly for CJ.

src/QMCWaveFunctions/Jastrow/CountingJastrow.h
```diff
@@ -35,18 +35,18 @@ class CountingJastrow : public WaveFunctionComponent
 
   // Jastrow linear coefficients
   Matrix<RealType> F;
-  std::vector<RealType> G;
+  //std::vector<RealType> G;
```
ye-luo: Delete if you don't want it any more.

src/QMCWaveFunctions/Jastrow/GaussianFunctor.h
```diff
@@ -9,22 +9,27 @@
...
 #include "OhmmsData/AttributeSet.h"
+#include "Optimize/VariableSet.h"
 #include <array>
 
 namespace qmcplusplus
 {
-// doesn't inherit from OptimizableFunctorBase since this is a function of the entire position vector
-template<class T>
-class GaussianFunctor : public QMCTraits
+
```
ye-luo: Need some comment for this class documenting all the use cases and limitations. The comment you deleted seems like useful info.

This is GaussianFunctor can not be used directly in J2 to replace the BsplineFunctor, having similar name is confusing. Please find a name more verbose.

src/QMCDrivers/QMCCostFunctionBase.h
bwvdg: My Jastrow factor contains variables that are not constants (i.e. can be input by the user) yet should not be optimizable (due to parameter redundancies). This is because the Jastrow basis (the "counting functions") are built from a set of normalized gaussians {g_I}: 

C_I(r_i) = (g_I(r_i)) / (\sum_{J} g_J(r_i) } ), 

The gaussian exponents of each g_I are used as optimizable parameters. Dividing by/multiplying by all the counting functions by an arbitrary gaussian function produces the same set of counting functions, and optimizing the all of the gaussian parameters together would therefore produce singular behavior in the linear method. One of the gaussian functions is designated as a 'reference' and should be flagged as not optimizable; requiring the user to use a constant reference gaussian would make the input file both more difficult to read and write.

The only way to write the counting jastrow parameters to the opt.xml files is to read these non-optimized variables from the wavefunction VariableSet. However, variables which are flagged as not optimizable are completely unregistered from both OptVariablesForPsi and OptVariables before this function is called. As a result, I copy the initial variables a single time to another VariableSet before printing so I have their values on hand.

Without doing something like keeping the unoptimized variables in one of the two present VariableSets or moving part of this xml updating function to the WavefunctionComponent class itself, there's not another obvious, simple way to get these values of these non-optimizable variables into the final opt.xml file.

src/Optimize/VariableSet.h
```diff
@@ -124,6 +124,18 @@ struct VariableSet
     return -1;
   }
 
+  inline int getLoc(const std::string& vname) const
```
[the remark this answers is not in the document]
bwvdg: I call this in resetParameters for both the CountingJastrow and GaussianFunctor classes. I add non-optimizable parameters to their VariableSet so that they their values may be read by the opt.xml updating function for reasons I describe in the next comment. 

Values in the Index array are updated before the non-optimizable parameters are removed, yet are not updated after. As a result, using getIndex returns the wrong index, and the incorrect parameter is updated.

Alternatives here are to reset the indices after these parameters are removed, either directly in the VariableSet call, or by adding another call to checkOutParameters either in QMCCostFunctionBase or to each of the wavefunction's resetParameters functions. 

I'm sort of puzzled why this indexing scheme is built on these locally-cached indices when it produces this additional complexity.

src/QMCWaveFunctions/Jastrow/GaussianFunctor.h
bwvdg: I've readded that comment and changed GaussianFunctor to CountingGaussian and NormalizedGaussianRegion to CountingGaussianRegion for consistency and clarity.

src/QMCWaveFunctions/Jastrow/CountingJastrow.h
bwvdg: Done.

src/QMCDrivers/QMCCostFunctionBase.h
ye-luo: I agree with your analysis but I prefer a harder but logically clean route.
I think we need to clean up how to output an input xml block or hdf5 tree for WavefunctionComponent and SPOSet.

In "./molecules/H4_ae/qmc_cjs_ref/H4_cjs_ref.wfs.xml" is the function g0 fixed?
By looking at the input file, I saw the first xml seems treated fixed implicitly.
I would expect a control key optimize="yes"/"no"
I remember in Jastrow factors, when optimize="no", the code doesn't optimize certain correlation pairs and the opt.xml can still print them. Could you have a look of how the two body Jastrow xml was handled?

ye-luo:
Okay to test

bwvdg:
The changes should be done.

prckent:
Thanks @bwvdg !

PR #1697, opened 2019-07-03
Fix build issues

### LINKED ISSUE #1696: No MPI build broken by spline updates
It appears that #1684 broke the no-MPI build.

ye-luo:
1. Most of time M_PI is defined in math.h but this is not part of the language standard.
So define M_PI, if M_PI was not previously defined in included files.
2. `<memory>` is needed by smart pointers.
3. Intel workaround needs to cover all 19 updates.

Closes #1696

### DIFF
--- a/src/Numerics/Ylm.h
+++ b/src/Numerics/Ylm.h
@@ -17,6 +17,7 @@
 #include <cmath>
 #include <complex>
 #include "OhmmsPETE/TinyVector.h"
+#include "config/stdlib/Constants.h"
 
 namespace qmcplusplus
 {
--- a/src/ParticleBase/RandomSeqGenerator.h
+++ b/src/ParticleBase/RandomSeqGenerator.h
@@ -21,6 +21,7 @@
 #include "OhmmsPETE/OhmmsMatrix.h"
 #include "ParticleBase/ParticleAttrib.h"
 #include "Utilities/RandomGenerator.h"
+#include "config/stdlib/Constants.h"
 
 /*!\fn template<class T> void assignGaussRand(T* restrict a, unsigned n)
   *\param a the starting pointer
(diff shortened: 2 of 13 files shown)


ye-luo:
The value is set only if the system header doesn't include M_PI.

The value is copied from GNU C library.
math.h:# define M_PI		3.14159265358979323846	/* pi */

PDoakORNL:
Ok mine had more digits but the following code indicates this seems ok
```C++
#define M_PI  3.14159265358979323846264338328
#define  M_PI2 3.14159265358979323846

#include <iostream>

int main()
{
 std::cout << M_PI << " " << M_PI2 << '\n';
 double em_pid = M_PI;
 float  em_pif = M_PI;
 double em_pid2 = M_PI2;
 float  em_pif2 = M_PI2;
 std::cout << std::setprecision(32) << em_pid << " ?= " << em_pid2 << '\n';
 std::cout << (em_pid == em_pid2) << '\n';
 std::cout << em_pif << " ?= " << em_pif2 << '\n';
 std::cout << (em_pif == em_pif2) << '\n';
 return M_PI == M_PI2;
}
```

```shell
(base) -bash-4.2$ ./test_mpi
3.14159 3.14159
3.1415926535897931159979634685442 ?= 3.1415926535897931159979634685442
1
3.1415927410125732421875 ?= 3.1415927410125732421875
1
```

prckent:
Clarification/reminder needed: The Intel workaround is needed because of a bug introduced in Intel2019 that was not present in 2018? What is the symptom of the missing workaround in 2019?

ye-luo:
The workaround is intended to solve Intel 19 compiler crash. It doesn't exist in Intel 18.

Issue #1691, opened 2019-07-02
Short and long excited state statistical integration tests needed


prckent:
Once the tests in #1690 have run successfully, we should add long reference tests.

Hyeondeok-Shin:
@prckent I can generate references for the integration tests in 1~2 days, it can be updated soon.

jtkrogel:
These now exist.

PR #1702, opened 2019-07-05
Add excited state statistical integration tests

Hyeondeok-Shin:
This is for #1691 and related to #740.

 Short and long statistical integration tests for excited state are added on diamondC_1x1x1_pp and diamondC_2x1x1_pp testset. VMC and DMC tests are added.
 Number of blocks and steps are the same as the tests for the ground state, so cpu time is almost consistent with the corresponding ground state test.

 Hyeondeok Shin


PDoakORNL:
test this please

ye-luo:
Test this please

jtkrogel:
Please update the readme detailing how the reference data were obtained.  Which version of the code was used for this?

Hyeondeok-Shin:
@jtkrogel Thank you for the comments, information is added to README and reference outputs are added to qmc-ref on diamondC_1x1x1. I didn't add the information and references on diamondC_2x1x1 because everything except system size is the same as the 1x1x1 tests. Please let me know if you're feeling 2x1x1 information is also needed.
  For the reference, QMCPACK 3.7.9 (current developer version) was used.

jtkrogel:
In general the earliest known working version should be used when making tests.  For the deterministic tests, we decided on a fixed version (3.6.0).

I would like to know that current develop reproduces the behavior of an earlier version.  I propose v3.1.1 (if others think a later/earlier version is more trustworthy, please chime in).  Please regenerate the reference data with this version.  If the reference data from the current develop and v3.1.1 agree, then no changes are needed to files in this PR.

prckent:
@jtkrogel Is there any reason not to use 3.0.1? Is that too hard to build / there have been too many input changes? I don't believe that there were intentional algorithm changes between then and now, but we do know that bugs and uncertainty were introduced along the 3.x timeline.

jtkrogel:
I have no objection to 3.0.1.

Hyeondeok-Shin:
@jtkrogel @prckent Then, I'll check the reference on 3.0.1.

Hyeondeok-Shin:
@jtkrogel @prckent There are little differences on the DMC reference (10x more blocks than long test) between 3.0.1 and 3.7.9.

-diamond 1x1x1
3.7.9
series 1  -10.056785 +/- 0.000298   0.411150 +/- 0.000471   0.0409
3.0.1
series 1  -10.057283 +/- 0.000320   0.411110 +/- 0.000441   0.0409

-diamond 2x1x1
3.7.9                                                                
series 1  -21.356382 +/- 0.001157   0.966079 +/- 0.007642   0.0452
3.0.1
series 1  -21.359004 +/- 0.001223   0.956384 +/- 0.001429   0.0448

They are still at around 2 sigma (1.7 sigma for 1x1x1 and 2.3 sigma for 2x1x1), but it seems the references are need to be changed as 3.0.1 as we promised. I'll update when VMC reference runs have completed and long & short tests in current version are ok with the references in 3.0.1. (sigma seems pretty large than I expected.)

jtkrogel:
These differences seem just fine: 0.25(22) mHa/atom for 1x1x1 and 0.66(42) mHa/atom for 2x1x1.

For consistency, updating the reference to 3.0.1 makes sense, but apparently there is no significant difference (systematic or statistical).

Hyeondeok-Shin:
@jtkrogel @prckent Now the references are replaced to 3.0.1 outputs.

jtkrogel:
Looks good to me.

prckent:
Test this please

PR #1704, opened 2019-07-08
Adding vmc_batch option for qmc_mode input

PDoakORNL:
1. Adding the vmc_batch input option and initial test for it. 
2. Made QMCRunType a scope enum class, to facilitate later templating on this the enum. 
3. Fixed issue with qmcham library not having a unit test variant linked to qmcfakerng.
4. Added pure virtual base class QMCDriverInterface to temporarily support new QMCDriverUnified base class.


src/QMCApp/QMCDriverFactory.cpp
```diff
@@ -89,117 +95,112 @@ bool QMCDriverFactory::setQMCDriver(int curSeries, xmlNodePtr cur)
...
+
+bool QMCDriverFactory::setQMCDriver(int curSeries, xmlNodePtr cur, QMCDriverFactory::DriverAssemblyState& das)
+{
+  unsigned long newQmcMode = das.what_to_do.to_ulong();
   //initialize to 0
-  QMCDriver::BranchEngineType* branchEngine = 0;
+  QMCDriver::BranchEngineType* branchEngine = nullptr;
   if (qmcDriver)
   {
-    if (newRunType != curRunType || newQmcMode != curQmcMode)
```
ye-luo: It seem that reusing a driver from the previous section of the same driver type is removed. Is this a safe change not affecting the code behaviour? For example, if I run multiple DMC sections with identical setting, everything is good at the transition from one section to another?

I feel more tests will be needed if we change the production code path.

src/QMCApp/QMCDriverFactory.cpp
```diff
@@ -76,8 +76,14 @@ void QMCDriverFactory::putCommunicator(xmlNodePtr cur)
   }
 }
 
-bool QMCDriverFactory::setQMCDriver(int curSeries, xmlNodePtr cur)
+/** Read the xml specifify the driver for this QMC section
+ *
+ *  Copy ellision should result in just a move of the
```
ye-luo: elision? single l?

src/QMCApp/QMCDriverFactory.cpp
PDoakORNL: I will write another test.  I'm "sure" that the branch engine brings everything needed.

PDoakORNL: The statistical tests for NiO_a4_e48_pp with back to back DMC sections pass, It is looking like quite an undertaking to write an integration test that checks this. Since this behavior is to be eliminated I think the system level testing is what we want to fall back on here. If these tests are insufficient what do you think we should do?

PDoakORNL:
test this please

ye-luo:
```
  <qmc method="vmc" move="pbyp" gpu="yes">
    <estimator name="LocalEnergy" hdf5="no" />
    <parameter name="walkers">               16 </parameter>
    <parameter name="warmupSteps">            5 </parameter>
    <parameter name="substeps">               4 </parameter>
    <parameter name="steps">                 16 </parameter>
    <parameter name="blocks">                 5 </parameter>
    <parameter name="timestep">             1.0 </parameter>
    <parameter name="usedrift">             yes </parameter>
    <parameter name="samples">              256 </parameter>
  </qmc>
  <qmc method="dmc" move="pbyp" gpu="yes">
    <estimator name="LocalEnergy" hdf5="no" />
    <parameter name="targetwalkers">       2000   </parameter>
    <parameter name="warmupSteps">          100 </parameter>
    <parameter name="steps">                 10 </parameter>
    <parameter name="blocks">               100 </parameter>
    <parameter name="timestep">            0.01 </parameter>
    <parameter name="reconfiguration">       no </parameter>
    <parameter name="nonlocalmoves">         v1 </parameter>
    <parameter name="blocks_between_recompute">         1 </parameter>
  </qmc>
  <qmc method="dmc" move="pbyp" gpu="yes">
    <estimator name="LocalEnergy" hdf5="no" />
    <parameter name="targetwalkers">       2000   </parameter>
    <parameter name="warmupSteps">          100 </parameter>
    <parameter name="steps">                 10 </parameter>
    <parameter name="blocks">               100 </parameter>
    <parameter name="timestep">            0.01 </parameter>
    <parameter name="reconfiguration">       no </parameter>
    <parameter name="nonlocalmoves">         v3 </parameter>
    <parameter name="blocks_between_recompute">         1 </parameter>
  </qmc>
```
I ran the following input only got 2 scalar.dat one for VMC and a second one for the first DMC section. I didn't get the third scalar.dat. I saw some append="yes" in the printout but I didn't see the second scalar.dat getting longer. I don't remember if the "append" feature still works, it is intended when users need to restart a run and get more statistics piped to the previous output files. It needs to be requested from an input key not automatically.

I believe something went wrong. Diff the printout can help debugging.

PDoakORNL:
Thanks @ye-luo this isn't something the existing statistical tests catch or the deterministic tests address. Nor something I thought to look for. I'll look at this behavior in detail, there was a good deal of spaghetti with respect to append and reuse of drivers spread across several files. I will write one or more unit tests that enforce the correct behavior.

But as you mentioned elsewhere we should have a deterministic test with a VMC and multiple DMC sections. And there we should as a backstop test check that all the expected files are generated by the new test for multiple DMC sections.

ye-luo:
No problem. Tests are never enough and we just need to write more. In terms of appending output feature, we can probably deal it later.

prckent:
Unfortunately append is almost undocumented and definitely untaught and not explicitly tested. The default is listed as "yes" in the driver options table in Chapter 10 of the manual. We might consider what to do about append in another issue. This PR should not change the situation from what we currently have.

PDoakORNL:
The default behavior in "develop" for one DMC section following another is `<qmc ... append="no">`.
It now is in this PR as well. 

Append will never be true unless explicitly set in input i.e.
```xml
<qmc method="vmc_batch" move="pbyp" append="yes">
    <estimator name="LocalEnergy" hdf5="no" />
    ...
</qmc>
```

Each DMC section in this PR now produces its own s00X.dat and s00X.xml files as in develop.

However if append(n+1) is explicitly set true in successive qmc sections with method(n+1) == method(n) it is now more broken, in "develop" the current step in a reused driver's CurrentStep was not reset if append == true. However this effected only dmc.dat for a section not scalar.dat. i.e. dmc.dat was append to where as scalar.dat was overwritten. Additionally the ...cont.xml appears to be missing the current step field the driver of a successive run would need to append.

ye-luo:
Please delete all the left resetComponents

src/QMCApp/QMCDriverFactory.cpp
```diff
@@ -89,117 +95,105 @@ bool QMCDriverFactory::setQMCDriver(int curSeries, xmlNodePtr cur)
...
     }
     else if (qmc_mode.find("dmc") < nchars)
     {
-      newRunType = DMC_RUN;
+      das.new_run_type =QMCRunType::DMC;
     }
     else if (qmc_mode == wf_test_name)
     {
-      newRunType = WF_TEST_RUN;
+      das.new_run_type =QMCRunType::WF_TEST;
```
ye-luo: Formatting? `das.new_run_type = QMCRunType::WF_TEST;`

ye-luo:
@PDoakORNL Once the left changes are fulfilled. Let us merge ASAP.

PDoakORNL:
I believe I've addressed the the remaining changes.

Issue #1574, opened 2019-05-02
Cmake versions higher that 3.14.3 cannot find packages 


anbenali:
with cmake versions equal and newer explicitely than 3.14.3 there is a change in policy that makes it hard to find libraries like HDF5.
```
[benali@hopcroft build_real]$ ../../Libraries/cmake-3.14.3/bin/cmake --help-policy CMP0074
CMP0074
-------

``find_package()`` uses ``<PackageName>_ROOT`` variables.

In CMake 3.12 and above the ``find_package(<PackageName>)`` command now
searches prefixes specified by the ``<PackageName>_ROOT`` CMake
variable and the ``<PackageName>_ROOT`` environment variable.
Package roots are maintained as a stack so nested calls to all ``find_*``
commands inside find modules also search the roots as prefixes.  This policy
provides compatibility with projects that have not been updated to avoid using
``<PackageName>_ROOT`` variables for other purposes.

The ``OLD`` behavior for this policy is to ignore ``<PackageName>_ROOT``
variables.  The ``NEW`` behavior for this policy is to use
``<PackageName>_ROOT`` variables.

This policy was introduced in CMake version 3.12.  CMake version
3.14.3 warns when the policy is not set and uses ``OLD`` behavior.
Use the ``cmake_policy()`` command to set it to ``OLD`` or ``NEW``
explicitly.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.
```

The Cmake output gives this warning:
```
-- Found LibXml2: /lib64/libxml2.so (found version "2.9.1") 
CMake Warning (dev) at CMakeLists.txt:604 (find_package):
  Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.
  Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  Environment variable HDF5_ROOT is set to:

    /home/benali/src/Libraries/hdf5-1.10.5/Libraries

  For compatibility, CMake is ignoring the variable.
This warning is for project developers.  Use -Wno-dev to suppress it.

-- HDF5: Using hdf5 compiler wrapper to determine C configuration
-- Found HDF5: hdf5;/usr/lib64/libz.so;/usr/lib64/libdl.so;/usr/lib64/libm.so  found components:  C 
-- Using HDF5 non-scalable serial I/O due to the lack of library support for parallel
```
==============
[...]

PR #1705, opened 2019-07-08
fixing CMP0074 policy setting to prevent cmake warnings

PDoakORNL:
Fixing #1574 we now set this policy.  Most except for FindCUDA which use CUDA_TOOLKIT_ROOT_DIR most of the Find<PACKAGE_NAME>.cmake modules follow this policay as well. I don't think we should map CUDA_ROOT --> CUDA_TOOLKIT_ROOT_DIR since we will should switch to CUDA as a language support and this eccentricity of the FindCUDA.cmake is well known, i.e. we're all used to CUDA_TOOLKIT_ROOT_DIR.


prckent:
Presumably no updates to our build recipes are needed. Correct?

PDoakORNL:
Nothing should stop working, its possible that a problem like @anbenali will be fixed (unless in a Find<Package Name>.cmake.  Some warnings will go away.

PR #1715, opened 2019-07-10
[WIP][No CI] Edited SlaterDetOpt to be able to use LCAOrbitalSet;

spinedaflores:
Still need to clean up syntax/code and resolve a memory error that is causing a memory leak / seg fault.
I am starting the pull request to get feed back on what is currently done and to get help on the memory error

### DIFF
--- a/src/QMCWaveFunctions/SPOSet.h
+++ b/src/QMCWaveFunctions/SPOSet.h
@@ -157,6 +157,21 @@ public:
   virtual void checkInVariables(opt_variables_type& active) {}
   virtual void checkOutVariables(const opt_variables_type& active) {}
 
+  virtual void evaluateDerivatives(ParticleSet& P,
+                                          const opt_variables_type& optvars,
+                                          std::vector<ValueType>& dlogpsi,
+                                          std::vector<ValueType>& dhpsioverpsi,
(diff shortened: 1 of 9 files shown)


src/QMCWaveFunctions/SPOSet.h
```diff
@@ -157,6 +157,21 @@ class SPOSet : public QMCTraits
...
+                                          const opt_variables_type& optvars,
+                                          std::vector<ValueType>& dlogpsi,
+                                          std::vector<ValueType>& dhpsioverpsi,
+                                          const ValueType& psiCurrent,
+                                          const ValueMatrix_t& M,
+                                          const ValueMatrix_t& Minv,
+                                          const ValueMatrix_t& Bbar,
+                                          const int m_first_var_pos,
+                                          const std::vector<std::pair<int, int>> m_act_rot_inds,
+                                          const int m_nel,
```
ye-luo: All the rotation related parameters need to be encapsulated inside the SPO class. The determinant class doesn't need to know any details how SPO is optimized. Thus, m_act_rot_inds probably should not show up in the arguments.

What is Bbar? Please document the added function as well all the arguments. Please do the same thing for evaluateDerivatives used by MSD as well.

PR #1718, opened 2019-07-12
Rename addOrbital and define is_fermionic by dervived classes instead of a builder

ye-luo:
1. Rename addOrbital to addComponent in TrialWaveFunction
2. define is_fermionic of WaveFunctionCompoenent by derived classes instead of addOrbital


prckent:
(Wait for any comments from @PDoakORNL before merging)

PR #1719, opened 2019-07-12
Add private/public in NonLocalECPComponent

ye-luo:
More concise access control.
`friend class NonLocalECPotential_CUDA;` is introduced only as a temporal workaround.

### DIFF
--- a/src/QMCHamiltonians/NonLocalECPComponent.cpp
+++ b/src/QMCHamiltonians/NonLocalECPComponent.cpp
@@ -162,7 +162,7 @@ NonLocalECPComponent::RealType NonLocalECPComponent::evaluateOne(ParticleSet& W,
   {
     RealType zz = dot(dr, rrotsgrid_m[j]) * rinv;
     // Forming the Legendre polynomials
-    lpol[0]          = cone;
+    lpol[0]           = cone;
     RealType lpolprev = czero;
     for (int l = 0; l < lmax; l++)
     {
@@ -304,7 +304,7 @@ NonLocalECPComponent::RealType NonLocalECPComponent::evaluateOneWithForces(Parti
     cosgrad[j] = rinv * uminusrvec;
 
     RealType udotgradpsi = dot(gradpsiratio[j], rrotsgrid_m[j]);
-    wfngrad[j]          = gradpsiratio[j] - dr * (udotgradpsi * rinv);
+    wfngrad[j]           = gradpsiratio[j] - dr * (udotgradpsi * rinv);
     wfngrad[j] *= sgridweight_m[j];
 
     // Forming the Legendre polynomials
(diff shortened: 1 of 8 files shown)


src/QMCHamiltonians/NonLocalECPComponent.cpp
```diff
@@ -550,17 +550,17 @@ NonLocalECPComponent::RealType NonLocalECPComponent::evaluateOneWithForces(Parti
     {
       //Note.  Because we are computing "forces", there's a -1 difference between this and
       //direct finite difference calculations.
-      lsum           += vrad[l] * lpol[angpp_m[l]] * psiratio[j];
-      gradpotterm_   += vgrad[l] * lpol[angpp_m[l]] * psiratio[j];
+      lsum += vrad[l] * lpol[angpp_m[l]] * psiratio[j];
+      gradpotterm_ += vgrad[l] * lpol[angpp_m[l]] * psiratio[j];
       gradlpolyterm_ += vrad[l] * dlpol[angpp_m[l]] * cosgrad[j] * psiratio[j];
-      gradwfnterm_   += vrad[l] * lpol[angpp_m[l]] * wfngrad[j];
+      gradwfnterm_ += vrad[l] * lpol[angpp_m[l]] * wfngrad[j];
```
jtkrogel: These changes will just be reverted again by auto-format, but I guess it is no big deal.

PDoakORNL: At the moment I don’t think so I didn’t have the bandwidth to fight for my consistent assignment operator PR to clang-format. I think we just have to let these go for now.

ye-luo: Actually this is changed by clang-format not me.

src/QMCHamiltonians/tests/test_ecp.cpp
```diff
@@ -248,7 +248,7 @@ TEST_CASE("Evaluate_ecp", "[hamiltonian]")
   //This line is required because the randomized quadrature Lattice is set by
   //random number generator in NonLocalECPotential.  We take the unrotated
   //quadrature Lattice instead...
-  nlpp->rrotsgrid_m = nlpp->sgridxyz_m;
+  nlpp->copyGridUnrotated();
```
PDoakORNL: Several Things Here:
1. Ideally a class API shouldn't contain (public) testing only functions.  Doubly so for one that modifies the sense of a member variable.
2. If this is really essential I would separate it into a friend test class or function just built into the test executable not the hamiltonian class library.
3. I think the correct way to test this would be with a mock RandomGenerator somehow delivered to nlpp so that we are testing a behavior of the class(es),
without making modifications to their internals or their normal sequence of states.
4. The fact this will take a little work is indicative of the murky creation and dependency of an nlpp. And offers a chance for improving the design.

@ye-luo Do you think any progress in this direction is within the scope of this PR?
I'm willing to except this member function as a refactoring stage toward better design and testing for ECPCompoenents in general.

ye-luo: I didn't think harder how to enable tests safely in this case. I actually didn't like the API I added currently in the PR. Adding a public API (security hole) for testing purpose is not a good thing. I would go for a friend function in this case because I can't access the class name in catch. Currently I'm not working on the creation of a better unit test or re-factoring this class. I need to add a feature in this class for science but don't want to mix the feature with the basic cleaning work.

prckent: IMO: Peter's #3 is the correct answer, and #4 is correct analysis. That said I am happy to take this PR as being on the route to a better design and enabling a new science feature.

ye-luo: @PDoakORNL I did 2 at the moment. I agree we should do 3. @markdewing introduced USE_FAKE_RNG which overall indicates some issue in the design of passing RNGs.

PDoakORNL: This works for now. 3 Is probably best done in tandem with some other architecture changes. We should think in general about how we handle the distribution and ownership of RNGs.

PR #1720, opened 2019-07-12
Nexus: update version number to 1.8.0

jtkrogel:
Update to version 1.8.0 ahead of the release.


jtkrogel:
Yes, and we probably should from here out.

ye-luo:
Okay to test

PR #1723, opened 2019-07-12
Nexus: initial Sphinx docs

jtkrogel:
Add means to generate html/pdf developer docs via Sphinx.  For example, html can be generated by running the Makefile as: `make html`.  The added files were autogenerated by Sphinx and hand modified to adjust to the Nexus source.  

More adjustments need to be made to rendered style, etc. before this PR goes in.

The main goal here is to enable rapid generation of developer docs as this will speed the addition of properly formatted (NumPy-style) docstrings, similar to those already present in gaussian_process.py.


prckent:
Why did you need to modify the Sphinx generated files?

jtkrogel:
AFAIK this is common practice.  The Sphinx files are autogenerated by a wizard program just to get you started and then you customize them over time to suit your needs.

Paul-St-Young:
> (Google-style) docstrings

@jtkrogel It looks like `gaussian_process.py` has numpy-style docstrings rather than Google-style (extension="napoleon"). Is this the intention?
I typically use Google-style because it is easier to write, but I also like numpy style because it is easier to read. Just want to make sure which one we are picking because switching further down the line would be more painful.

jtkrogel:
@Paul-St-Young I've looked into this a little more closely and I'm coming down on the side of NumPy style.

jtkrogel:
This is fine to go in now; the basic style is now settled.  Expansions will be made in the future.

ye-luo:
Okay to test

PR #1724, opened 2019-07-13
Implement flag for test verbosity during cmake configuration

### LINKED ISSUE #1224: Reduce output about testing during cmake configuration
Currently we largely print one entry for each test. This is not actionable and tends to bury useful message about tests that were not added for some reason. It is also far out of proportion to the other configuration output. I suggest we only print when broad categories of tests are not added, and give the reason.

prckent:
Add new cmake flag QMC_VERBOSE_TEST_CONFIGURATION and update messaging throughout cmake configuration to reduce output when QMC_VERBOSE_TEST_CONFIGURATION=0 (default).

I will update the documentation after initial review.

Closes #1224


prckent:
Updated. Changed the flag to QMC_VERBOSE_CONFIGURATION since there will likely be other uses. Updated the docs.

markdewing:
An alternate implementation would be to create a function `TEST_CONFIG_MESSAGE` to replace the existing calls to `MESSAGE` and  put the if test for `QMC_VERBOSE_CONFIGURATION` inside that function.
  The alternate implementation has the advantage to putting the test in single place.  It has the disadvantage of adding an extra function (and layer of indirection).
  The solution in this PR has the advantage of being more direct and obvious, and the disadvantage is it requires adding the if-test everywhere it is needed.

prckent:
I thought about implementing a function at the time but went the simple route. Over time the trend should be to less of these as we move towards one build that does everything. I noticed that the cmake code around the tests is duplicated in several places so there are other efficiencies to gain.

We might need to add some more messages so that cmake looks like it is making enough progress.

ye-luo:
We will add more tests and I feel the current scheme is inferior to a function MESSAGE_VERBOSE.
If we need to rename the option or move something from MESSAGE to MESSAGE_VERBOSE, a function is much simpler. Let us get it right now.

prckent:
Done.

I moved the include of macros.cmake to the topmost CMakeLists.txt to have visibility of MESSAGE_VERBOSE and reduced our overzealous reinclusion of that file throughout.

PR #1725, opened 2019-07-13
Remove coefficientHistory

ye-luo:
Data is collected in coefficientHistory but never taken out.
Basically, it is unused code stick in a bad place.


prckent:
If we wanted to update one of the existing optimizers to make a better choice about final coefficients, e.g. to take the best value of the objective function over the last 5 cycles, would any of this code be useful? (My guess is not and the cost of recreating it is low...)

ye-luo:
In this case, recreating new code is simpler than archaeology on unused legacy code.

Issue #1640, opened 2019-06-15
Improve RPAJastrow Usability


Paul-St-Young:
From #1635, we identified a few issues with the current implementation of RPAJastrow.
This issue will serve as a tracker for a future PR with the desired fixes

- [ ] `nparam`, `npts`, and `tiny` should be input parameters
- [ ] enhance wavefunction output
- [ ] more explanations needed in the manual
- [ ] validate unit test and check Bspline as well as kSpaceJastrow coefficients
- [ ] enable HEG integration test

jtkrogel:
@Paul-St-Young still working on this or close?

Paul-St-Young:
@jtkrogel this fell off my radar.
Please close unless there is further interest.

PR #1727, opened 2019-07-13
output long-range Jastrow coefficients

Paul-St-Young:
Tangentially related to #1640, this PR outputs coefficients in `kSpaceJastrow` to file.
Functions designed to mimic the way `RadialJastrowBuilder` outputs "J1.*.dat" and "J2.*.dat".


ye-luo:
Could you make a unit test for `kSpaceJastrow`? Although the RPA one uses it inside. A separate unit test is better.

More of just a question: Does the one and two body share data or they are quite separate? I have a feeling that it is better to have them in separate classes.

Paul-St-Young:
OK, I will think of a unit test.

Regarding your question, my impression is that the one- and two-body parts do not share data. I am not sure why they are not separated classes, although I did not make this class. @rcclay probably knows this better.

prckent:
Test this please

ye-luo:
Okay to test

PR #1728, opened 2019-07-13
[nexus] add generate_kspace_jastrow

Paul-St-Young:
Added `generate_kspace_jastrow` in the same style as `generate_jastrow1,2,3`.

I want to add a default long-range Jastrow to `generate_jastrow`, but I need help with setting reasonable defaults. I think a safe default is: no one-body part and RPA two-body part. This means `kc1=0`, `nk1=0`, `kc2=1~2*kf`, `nk2=[# of kshells inside kc2]`.

To enable this default, I need access to Fermi kvector magnitude `kf` and count the number of kshells inside `kf`. Some routines in `structure.py` (designed for tiling), can be used for this purpose. I am tempted to add `calc_kf` and `count_nsh` to `structure.py`, but I am not sure if this is the best route. @jtkrogel what do you think?


jtkrogel:
Adding functions to structure.py is fine.  Names like `kf_rpa` and `count_kshells` would be more descriptive IMO.

nexus/lib/qmcpack_input.py
```diff
@@ -5254,6 +5255,57 @@ def generate_jastrow3(function='polynomial',esize=3,isize=3,rcut=4.,coeff=None,i
...
+    coeff1 (list, optional): one-body Jastrow coefficients, default None
+    coeff2 (list, optional): two-body Jastrow coefficients, default None
+  Returns:
+    QIxml: kspace_jastrow qmcpack_input element
+  """
+
+  if coeff1 is None: coeff1 = [0]*nk1
+  if coeff2 is None: coeff2 = [0]*nk2
+  if len(coeff1) != nk1: raise RuntimeError('coeff1 mismatch')
+  if len(coeff2) != nk2: raise RuntimeError('coeff2 mismatch')
```
jtkrogel: Please use the standard error handling mechanisms in Nexus.  In this case, `QmcpackInput.class_error('coeff2 mismatch','generate_kspace_jastrow')` would do fine.  These errors can all be converted to raise type errors under the hood (as nxs-test does) while defaulting to "abort with traceback" for typical use.

nexus/lib/qmcpack_input.py
```diff
@@ -5254,6 +5255,57 @@ def generate_jastrow3(function='polynomial',esize=3,isize=3,rcut=4.,coeff=None,i
...
+
+  corr1 = correlation({
+    'type': 'One-Body',
+    'symmetry': symm1,
+    'kc': kc1,
+    'coefficients': section({
+      'id': 'cG1', 'type': 'Array',
+      'coeff': coeff1
+     })
+  })
```
jtkrogel: This looks a little heavy syntactically.  Does this not work:

  corr1 = correlation(
    type = 'One-Body',
    symmetry = symm1,
    kc = kc1,
    coefficients = section(
      id = 'cG1', type = 'Array',
      coeff = coeff1
     )
  )

I generally prefer this form for Nexus since it removes two layers of markup: one for the dict "{}" and one for each key "''".

Paul-St-Young: I use dictionary in my own scripts because then I can pass the dictionary around to other simple scripts that know nothing about nexus. Now, since this function is being added to nexus, I agree that the internal syntax is cleaner.

Paul-St-Young:
@jtkrogel I address your concerns and added the default k-space jastrow to `generate_jastrows`. Please see if this PR is good to merge now.

I added a few unit tests to the folder `nexus/tests/unit` in the style of pytest/nosetest. Is this OK with you?

prckent:
@jtkrogel  Please can you mark this as changes requested if that is still the situation.

jtkrogel:
Sorry, I'm being a little slow getting to this.

nexus/lib/physical_system.py
```diff
@@ -635,6 +635,15 @@ def ae_pp_species(self):
         #end if
         return ae_species,pp_species
     #end def ae_pp_species
+
+    def kf_rpa(self):
+      import numpy as np
```
jtkrogel: Move this to a top-level import.  Nexus will move to this standard pattern over time, but no need for local import.

nexus/lib/physical_system.py
```diff
@@ -635,6 +635,15 @@ def ae_pp_species(self):
...
+      nelecs = self.particles.electron_counts()
+      volume = self.structure.volume()
+      kvol1 = (2*np.pi)**3/volume  # k-space volume per particle
+      kfs = [(3*nelec*kvol1/(4*np.pi))**(1./3) for nelec in nelecs]
```
jtkrogel: Prefer ndarray return type.

nexus/lib/qmcpack_input.py
```diff
@@ -4786,6 +4787,11 @@ def generate_jastrows(jastrows,system=None,return_list=False,check_ions=False):
         if '3' in jorders and have_ions:
             jterm = generate_jastrow('J3','polynomial',3,3,4.0,system=system)
         #end if
+        if 'k' in jorders:
+            kcut = max(system.rpa_kf())
+            nksh = system.structure.count_kshells(kcut)
+            jterm = generate_kspace_jastrow(0, kcut, 0, nksh)
```
jtkrogel: Use keyword inputs where possible as these break less easily under future changes than positional arguments do.

nexus/lib/qmcpack_input.py
```diff
@@ -5254,6 +5260,59 @@ def generate_jastrow3(function='polynomial',esize=3,isize=3,rcut=4.,coeff=None,i
 #end def generate_jastrow3
 
 
+def generate_kspace_jastrow(kc1, kc2, nk1, nk2,
+  symm1='isotropic', symm2='isotropic', coeff1=None, coeff2=None):
+  """ generate <jastrow type="kSpace">
+
+  Args:
```
jtkrogel: Move to NumPy style docstrings.

nexus/lib/qmcpack_input.py
```diff
@@ -5254,6 +5260,59 @@ def generate_jastrow3(function='polynomial',esize=3,isize=3,rcut=4.,coeff=None,i
...
+    coeff1 (list, optional): one-body Jastrow coefficients, default None
+    coeff2 (list, optional): two-body Jastrow coefficients, default None
+  Returns:
+    QIxml: kspace_jastrow qmcpack_input element
+  """
+
+  if coeff1 is None: coeff1 = [0]*nk1
+  if coeff2 is None: coeff2 = [0]*nk2
+  if len(coeff1) != nk1:
+    QmcpackInput.class_error('coeff1 mismatch', 'generate_kspace_jastrow')
```
jtkrogel: Add `#end if` closures.

nexus/lib/structure.py
```diff
@@ -1223,6 +1223,36 @@ def distances(self,pos1=None,pos2=None):
...
+      # check tilevec input
+      for nt in tilevec:
+        if nt % 2 != 0:
+          msg = 'tilevec must contain even integers'
+          msg += ' so that kgrid can be zero centered.'
+          Structure.class_error(msg, 'count_kshells')
+        #end if
+      #end for
+
+      import numpy as np
```
jtkrogel: Move to top level import.

nexus/tests/unit/test_qmcpack_input.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
+def test_generate_kspace_jastrow():
```
jtkrogel: Similarly repackage into nxs-test with the contents of "main" and all other local functions going into a new function called "qmcpack_input".

jtkrogel:
@Paul-St-Young Thanks for your patience.  You caught me in the middle of settling on docstring styles.

Most of the remaining requested changes are small.  Hopefully repackaging the tests will not take too much effort.

Your inclusion of documentation and tests are greatly appreciated.

Paul-St-Young:
@jtkrogel sorry I did not see your comments until now. I addressed most of your comments (everything except for the unit tests).

I am a little intimidated by the size of `nxs-test`. Do you think it makes sense to include only the difficult tests (like job monitoring and integration tests) in `nxs-test` and split out simple unit tests of stand-alone objects like `strcuture` and `qmcpack_input` in standard pytest/nosetest style?

I prefer the latter because:
1. Many users can contribute tests on different objects simultaneously without merge conflict.
2. New users need not learn the internal syntax of `nxs-test`, only the standard pytest/nosetest syntax.
3. This helps make `nxs-test` less intimidating to edit.

Please let me know if this is OK with you (i.e. keep the unit tests as is).
If you still prefer keeping all the tests in one file, then I will repackage the tests tomorrow (hopefully...)

nexus/tests/unit/test_physical_system.py
```diff
@@ -0,0 +1,43 @@
...
+    net_charge = 1,
+    net_spin = 1,
+    H = 1
+  )
+  kfs = ps.kf_rpa() 
+  assert np.isclose(kfs[0], 1.465, atol=1e-3)
+  assert np.isclose(kfs[1], 1.465/2**(1./3), atol=1e-3)
+#end def test_kf_rpa
+
+def test_count_kshells():
```
jtkrogel: Move `test_count_kshells` into a new file named `test_structure.py` as this tests a Structure object rather than a PhysicalSystem object.

jtkrogel:
@Paul-St-Young nxs-test needs to be able to run all the tests.  I agree that supporting pytest style development is also desirable.  

I have a solution in the works that will allow easy extension of nxs-test to include tests developed in standard pytest style that I will put in a PR after this one goes in.  This PR can go in once the tests are split up as requested above.

jtkrogel:
NB: PR #1765 is the PR referred to above.

Paul-St-Young:
@jtkrogel I split out `test_count_kshells` as you requested.

Thanks for being accommodating to these different styles!

jtkrogel:
Thanks, and you're welcome.  Enabling standard style unit tests should increase the number of contributed tests, which is highly useful.

prckent:
Test this please

prckent:
OK to test

Issue #1729, opened 2019-07-15
Deterministic integration tests for CUDA code: add reference data or disable?


prckent:
Currently the deterministic tests are run for the legacy CUDA code. Necessarily they will always fail due to the different numerics. i.e. We rely on the statistical integration tests for correctness and to detect meaningful change.

It would be helpful to (1) have the deterministic tests always work ("ctest -L deterministic") and (2) catch changes on Volta, which I think is a safe reference for now. Do we add reference data or disable the numerical checks for now?

jtkrogel:
I vote for adding data for at least a subset of the deterministic tests.

ye-luo:
Let us do disabling first. It is basically adding if-else which will be anyway needed for adding a separate set for CUDA.

Hyeondeok-Shin:
I'm going to work with separated deterministic test for CUDA code as @jtkrogel mentioned. As a first step, I'll disable the deterministic test for CUDA with the current reference now as @ye-luo comment in order to avoid confusion.

ye-luo:
Legacy CUDA has been removed from the code base https://github.com/QMCPACK/qmcpack/pull/4431

PR #1730, opened 2019-07-15
Disable deterministic integration tests for CUDA build

Hyeondeok-Shin:
As discussed in #1729, we're going to make separated references of deterministic integration tests for CUDA code. 
 As a first step, deterministic integration tests are disabled for CUDA build with current references in order to avoid confusion. They will resume with separated references for CUDA build.

 Hyeondeok Shin


prckent:
Test this please

PR #1731, opened 2019-07-16
Orbital Optimization for DiracDeterminant Class

spinedaflores:
I'm trying to prepare DiracDeterminant class to perform orbital optimization.

The first step is to prepare the class to be able to read in more than the
 minimal amount of orbitals. This would correspond to changing the "size" parameter within the "determinant" tag in the input to be greater than the default minimum of the number of electrons present.

On a clean branch, changing the value of size to read in all MO will cause a segfault due to the SPOSet reading in more orbitals than DiracDeterminant believes there are.

In  **void DiracDeterminant<DU_TYPE>::set** function I  simply change one line 
from  **resize(nel, nel);**  to  **resize(nel, Phi->getOrbitalSetSize());**
But I found that this does not produce the behavior I'd like. I can compare runs with the same wfn but using the multislater class and I get different values returned for ratio func during the accept/reject phase.  I think the DelayedUpdate.h may not be updating the psiM, dpsiM, d2psiM objects correctly and I concluded this by printing these objects out using the DiracDeterminant class and the MultiSlaterFast class. If I return the value of size in the input to be equal to the original value of  number of electrons the print out from both classes match. But when I change size to read in all MOs then I have the bad behavior I have described. 

Could you help me determine what part of DelayedUpdate.h is causing the bad upate of  psiM, dpsiM, d2psiM. I have a suspicion that is has to do with the sherman morrison inverse updating but it's hard for me to follow what the code currently says. 

I've attached the simple inputs that I use to confirm when things are running correctly. The SD folder is the input for the Diracdeterminant class to be used. The MSD is the folder for the mutlislaterdetfast class to be used. To see the bad behavior for the SD input when the value of size is set to 4 in the determinant tag. If it is set to 2 then there are no issues but then not all the MOs have been read in.
[small_test.zip](https://github.com/QMCPACK/qmcpack/files/3394719/small_test.zip)

### DIFF
--- a/src/QMCWaveFunctions/CMakeLists.txt
+++ b/src/QMCWaveFunctions/CMakeLists.txt
@@ -185,6 +185,7 @@ IF(OHMMS_DIM MATCHES 3)
 ENDIF(OHMMS_DIM MATCHES 3)
 
 SET(FERMION_SRCS ${FERMION_SRCS}
+  Fermion/RotationHelper.cpp
   Fermion/DiracDeterminant.cpp
   Fermion/SlaterDet.cpp
   Fermion/SlaterDetBuilder.cpp
--- a/src/QMCWaveFunctions/Fermion/DelayedUpdate.h
+++ b/src/QMCWaveFunctions/Fermion/DelayedUpdate.h
@@ -106,7 +106,7 @@ public:
     }
     const T cone(1);
     const T czero(0);
-    const int norb     = Ainv.rows();
+    const int norb     = Ainv.cols();
     const int lda_Binv = Binv.cols();
     // save Ainv[rowchanged] to invRow
     std::copy_n(Ainv[rowchanged], norb, invRow.data());
@@ -128,7 +128,7 @@ public:
   {
     const T cminusone(-1);
     const T czero(0);
-    const int norb     = Ainv.rows();
+    const int norb     = Ainv.cols();
     const int lda_Binv = Binv.cols();
     std::copy_n(Ainv[rowchanged], norb, V[delay_count]);
     std::copy_n(psiV.data(), norb, U[delay_count]);
@@ -165,7 +165,7 @@ public:
     // update the inverse matrix
     const T cone(1);
     const T czero(0);
-    const int norb = Ainv.rows();
+    const int norb = Ainv.cols();
     if (delay_count == 1)
     {
       // this is a special case invoking the Fahy's variant of Sherman-Morrison update.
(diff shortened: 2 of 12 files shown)


ye-luo:
Building a determinant needs to 2 steps.
1. build the SPO
2. put the SPO into a determinant.
Having more than minimum orbitals is a feature in the SPO. determinants neither single nor multi don't need to know this detail. All the rotation happens inside SPO not determinants.

The set(M,N) function seems allowing `N!=M`. This is legacy. Currently only N=M is supported and only N=M should be supported.
When you need more SPO for rotation. you may need scratch space inside the roration SPO class. When the result is returned to the determinant, the size must be the one after rotation not before. So the sizes of psiM dpsiM d2psiM should not be changed.

spinedaflores:
@ye-luo  The most recent commit has code that works with the simple test I uploaded in the first comment. One thing to note is that I found the original issue I had mentioned ( the one in which reading in more than the minimal orbitals will cause the DiracDeterminant class to not evaluate the values of psiM dpsiM or d2psiM correctly on a clean branch without any of my additions of code for OrbOpt.)  This issue is in LCAOrbitalSet.cpp  in evaluate_vgl_impl function. It uses OrbitalSetSize to copy values from "temp" to the passed in ValueMatrices. The valuematrices will be of dimension nel,nel no matter how many orbitals are read in so the std::copy writes at addresses out of bounds. I only fixed the evaluate_vgl_impl at line 358 for now because I don't know how you want to address this, but other functions will need this fix too. 
Let me know how to proceed once you've had a chance to look at the additional code. I'm sure there are changes for efficiency we can make.

ye-luo:
@spinedaflores we need to arrange a meeting to discuss this.

prckent:
What is the current status of this PR?

spinedaflores:
I just need to get the threading working. I'm working on it currently.

ye-luo:
@spinedaflores Which test works? I tried `ctest -j 16 -R orb` and all the tests failed. I believe the msd case failure is not related to this PR. It was caused by some changes from @chrislzhao and hopefully you two had a plan to fix this. I'd like to fix a few things in this PR but I need a test first.

spinedaflores:
@ye-luo  Sorry about the delay. I think I resolved the threading issue I had. The only test I was running was an orbopt test from the last orbopt pull request, making sure I  get the exact same results. I'm going to clean up changes and push the code over the weekend. But for now does the rotationhelper class do what you were envisioning?

ye-luo:
@spinedaflores I will try the new PR today.
succeeded by #1893

PR #1732, opened 2019-07-16
Add QE 6.4.1 patch

### LINKED ISSUE #1722: Quantum Espresso v6.4.1 support
http://www.quantum-espresso.org/news-events/news/quantum-espresso-v-6-4-1

Ye: As we discussed, move this to the 3.8 Milestone if it looks simple enough. Otherwise we can do a v3.8.1.

ye-luo:
Closes #1722

### DIFF
--- a/external_codes/quantum_espresso/README
+++ b/external_codes/quantum_espresso/README
@@ -1,7 +1,7 @@
 The files in this directory support making a modified version of
 quantum espresso that includes the pw2qmcpack converter.
 
-The download_and_patch_qe6.4.sh script will automatically download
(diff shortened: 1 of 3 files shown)


ye-luo:
It runs well last night. https://cdash.qmcpack.org/CDash/viewTest.php?onlypassed&buildid=67577

prckent:
Since there is no way this can break rhea-cuda-experimental, I'll ignore that CI result as an infrastructure problem.

prckent:
Updated the nightlies to use 6.4.1. Only needed to change the version number in the scripts.

PR #1734, opened 2019-07-16
[WIP] Convert timer pointers to shared pointers

markdewing:
The timer objects are one of the leaky resources identified by valgrind.    The TimerManager class contains a list of timer objects and they all need to still be alive when the timers are analyzed and output.    This makes timers a good candidate for shared pointers.

Steps:

1. Create a type alias for NewTimer *, which will later be set to to a shared pointer.    For the purpose of design feedback, only one pointer was changed to NewTimerPtr.  If this design is okay, the rest of the "NewTimer *" locations will be changed to NewTimerPtr to complete this PR
2. In another PR, change NewTimerPtr to std::shared_ptr<NewTimer>.

### DIFF
--- a/src/Utilities/NewTimer.h
+++ b/src/Utilities/NewTimer.h
@@ -128,10 +128,12 @@ public:
 // N = 2 gives 16 nesting levels
 typedef StackKeyParam<2> StackKey;
 
+using NewTimerPtr = NewTimer *;
+
 class TimerManagerClass
 {
 protected:
-  std::vector<NewTimer*> TimerList;
+  std::vector<NewTimerPtr> TimerList;
   std::vector<NewTimer*> CurrentTimerStack;
   timer_levels timer_threshold;
   timer_id_t max_timer_id;


ye-luo:
The owner ship of the timer pointer must be made clear. Here is my thinking.
1. timer pointers must be created and owned by timer manager. Created first and then registered in manager should not be allowed.
2. Each thread doesn't own the shared_ptr and should use weak_ptr.

Some extra things to do for timers. I saw a lot of enum and functions in the global scope. They need to be moved to the class local scope.

markdewing:
A weak_ptr must be converted to a shared_ptr before use.   This seems like it would make it awkward to use to call start/stop.

A weak_ptr also doesn't quite match up to the model we want.   A weak_ptr says 'I don't own this, and it might go away at anytime, and I don't want to prevent it from going away'.   In our case, being owned by the TimerManager (especially if your condition 1 is met) means the timers are in no danger of going away.   We want a pointer that says 'I don't own this, but I know it won't go away'.  Handing out a raw pointer or a shared pointer seems equally valid in this case.

ye-luo:
If the ownership is clear and restricted to the manager. Do we still need the shared_ptr? Why not having the manager owing unique_ptr and then the consumer class just own a copy of the raw pointer?

ye-luo:
Also the name of the class, 'new' is not a future proof name. My understanding for this class is MasterThreadTimer.

markdewing:
My original rationale for using shared_ptr was then it would all just work and we could ignore lifetime and ownership issues.   But maybe that's too lazy an analysis approach.

PDoakORNL:
Here's how I'd like the timers to work based on changes I've tried to them in the past. This way you don't have raw pointers in the "users" of the timers and timer life span is clear and straightforward.

```C++
class NewTimer {
  std::string name_;
  int resolution_;
public:
 NewTimer(const std::string name,const int resolution) : name_(name), resolution_(resolution) {}
  std::string& getName() { return name_; }
...
};

class TimerManager
{
private:
  std::vector<std::unique_ptr<NewTimer>> timers;
public:
  static TimerManager& get()
  {
    static TimerManager instance;
    return instance;
  }
  
  NewTimer& addTimer(const std::string name, const int res)
  {
    timers.push_back(std::make_unique<NewTimer>(name, res));
    return *(timers.back());
  }
...
};

class Determinant
{
 NewTimer& Update, Ratio, Inverse;
 public:
 Determinant() : Update(TimerManager::get().addTimer("Update", 1)),
   Ratio(TimerManager::get().addTimer("Ratio", 1)),
    Inverse(TimerManager::get().addTimer("Inverse", 1))
      {}
  ...
};
```

ye-luo:
`get().addTimer` can be simplified just marking addTimer static.

PDoakORNL:
I guess how we deal with the current "global" nature of the timer manager is a matter of taste. Thread safety should also be updated. Going with the static variable/methods route.

```C++
class TimerManager
{
private:
  static std::vector<std::unique_ptr<NewTimer>> timers_;
  static std::mutex mutex_timers_;
public:
  static  NewTimer& addTimer(const std::string name, const int res);
...
};

NewTimer& TimerManager::addTimer(const std::string name, const int res)
  {
    std::lock_guard<std::mutex> guard(mutex_timers_);
    timers_.push_back(std::make_unique<NewTimer>(name, res));
    return *(timers_.back());
  }
```

ye-luo:
We currently use omp critical because addTimers will be called from OpenMP region. I guess it also not guaranteed that mutex always work with OpenMP.

If it returns reference, we cannot have a vector of timers.

PDoakORNL:
1. It seems you are correct in openmp reserving the right to break std::mutex in an openmp region. That is another reason to abstract away and ultimately limit use of openmp in the code. omp_critical is basically a global mutex. Having a properly scoped C++ mutex is much better in my opinion. But outside of the scope of this PR.

So raw pointers then. I think the code being clear about object ownership and lifetime are worth it. References can still be used for consumers when appropriate. 

A named omp_critical section should be used as addTimer is not the only place undefined behavior could occur and the rest of TimerManager should be reasoned through for thread safety.
 
```C++
NewTimer* TimerManager::addTimer(const std::string name, const int res)
  {
    generic_lock(timer_manager_vector_mutex)
    timers_.push_back(std::make_unique<NewTimer>(name, res));
    return timers_.back().get();
  }
...
 Determinant() : Update(*(TimerManager::addTimer("Update", 1))),
    Ratio(*(TimerManager::addTimer("Ratio", 1))),
    Inverse(*(TimerManager::addTimer("Inverse", 1)))
      {}
```

markdewing:
It looks like the new first step is to convert all initializations of timers to use createTimer and remove uses of the (old) addTimer.

I prefer @PDoakORNL 's suggestion of static get vs. making addTimer static because it still allows multiple TimerManager objects for unit testing.

ye-luo:
@markdewing Could you explain a bit more "it still allows multiple TimerManager objects for unit testing."

markdewing:
In test_timers.cpp, each test creates a separate TimerManager for the duration of that test.   If the TimerManager is static, then there is only one global copy, and each test much use that copy.  Each test then needs to reset or clear the TimerManger to avoid interfering with other tests.

PDoakORNL:
Sadly the singleton is just a better looking lazily intialized global and shares this issue.  You can write singleton of the older pattern with a private instance pointer and then add a delete method, but really TimerManager should just be an object in the scope of qmcapp.cpp main.

PR #1736, opened 2019-07-16
[nexus] optimal_tilematrix_hnf

Paul-St-Young:
I added an alternative optimal supercell search method to nexus' native `optimal_tilematrix` based on PRB 92, 184301 (2015). I added unit tests for the new functions, but I made two esoteric choices that probably require administrative approval:
1. GNU v3 license
  The core code is fortran and is taken and modified from the supplementary materials of PRB 92, 184301 (2015). My understanding is that GNU v3 is compatible with our NCSA license, so all should be well, but I just want to double-check to be sure.
2. f2py in nexus
  I put the fortran code in `nexus/lib/forlib`. It needs to be compiled using `f2py` into a Python library. I have included a makefile that should work on most Linux distributions (with numpy pre-installed). I think this dependence on `f2py` is OK, because `f2py` comes with numpy, which is very prevalent. @jtkrogel what do you think?


markdewing:
Code under the GPL is not compatible with the NCSA license (in the sense that including GPL-licensed code infects the rest of the code to fall under the GPL)

This code looks small enough that it could be rewritten in python fairly easily, and that would probably sidestep both issues.

nexus/lib/structure.py
```diff
@@ -441,7 +441,58 @@ def __call__(self,T):
     #end def __call__
 #end class MaskFilter
 mask_filter = MaskFilter()
-            
+
+def rwsc(axes, dn=1):
+  """ radius of the inscribed sphere inside the real-space
+  Wigner-Seitz cell of the given cell
+
+  Args:
```
jtkrogel: Need NumPy style docstrings.

nexus/lib/structure.py
```diff
@@ -441,7 +441,58 @@ def __call__(self,T):
...
+
+def optimal_tilematrix_hnf(axes, nprim, nc=5):
+  import numpy as np
+  try:
+    from forlib.gencell import nsupercell, generate_all_tilematrices
+    from forlib.gencell import generate_all_supercells
+  except ImportError:
+    msg = 'Failed to import from gencell.\n'
+    msg += 'Please make sure nexus/lib/forsub folder is compiled.'
+    raise RuntimeError(msg)
```
jtkrogel: Use bare "error" function from developer.py instead of raise.

nexus/lib/structure.py
```diff
@@ -441,7 +441,58 @@ def __call__(self,T):
...
+  # fourth, back solve for the tilematrix
+  scell = scells[:, :, iopt]
+  topt = np.round(np.dot(scell, np.linalg.inv(axes))).astype(int)
+  ropt = radii[iopt]
+  # last but not least, check optimal tiling matrix
+  nvol = np.round(abs(np.linalg.det(topt))).astype(int)
+  assert nvol == nprim
+  rw = rwsc(np.dot(topt, axes))
+  assert np.isclose(ropt, rw)
+  return topt, ropt
```
jtkrogel: Needs function closure comment.

nexus/tests/unit/test_gencell.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python
+import numpy as np
+from forlib.gencell import nsupercell, generate_all_tilematrices
+from forlib.gencell import generate_all_supercells
+
```
jtkrogel: Move contents to existing "structure" function in nxs-test.  Replace assert with nassert().  Check that "nxs-test -R structure" passes.

jtkrogel:
GPL is an issue.  What are the prospects for a rewrite?  Separately, is speed a concern?

prckent:
For convenience, the cited PRB is https://doi.org/10.1103/PhysRevB.92.184301

Paul-St-Young:
@markdewing @jtkrogel @prckent Thanks for the feedback! Bummer this cannot be included in the main distribution.

My preferred resolution here is to create a separate repository with the GPL license, then put a link under `external_codes`. After that, I can write a small nexus component to drive this external code. I am against a Python rewrite, because:
1. The fortran implementation is clean, fast, and presumable well tested by the authors.
2. It still feels like a violation of the GPL license in spirit.

Is this OK with everyone?

nexus/lib/structure.py
```diff
@@ -441,7 +441,58 @@ def __call__(self,T):
     #end def __call__
 #end class MaskFilter
 mask_filter = MaskFilter()
-            
+
+def rwsc(axes, dn=1):
```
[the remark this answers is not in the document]
Paul-St-Young: Oops, I just did not see `structure.rwigner`. I have not tested speed, but there is probably no need for a rewrite.

jtkrogel:
If a rewrite is too big an ask, then this really leaves GPL + any integration issues.  

Perhaps instead extend the `install` script distributed with Nexus to include a `-e`/`--external` option to download and install the contents of the separate repo you propose?  This would similarly avoid GPL contamination and have better scope.  

Overall I would prefer direct integration via f2py following download rather than an OS driver as this is more seamless user-side.

prckent:
Even without the license issues, the complexity <> benefit tradeoff does not seem worth it to me via the current route. Too easy for the feature to go untested or fail to install somewhere.

prckent:
Lets talk through this offline.

jtkrogel:
@Paul-St-Young I discussed this a bit with Paul and here are the options.

1. Are you willing to reimplement in Python?  If so, this is the preferred route.
2. If not, then the following would be OK as an intermediate term solution:
* Place GPL fortran code in a separate repository.
* Update the `install` script to include a `-e` option, which, when invoked, creates an `external_codes` directory under `nexus`, downloads the external repository, and compiles the source as needed.
* Update the Nexus source such that Nexus will abort with an error message if a call to `optimal_tilematrix_hnf` is attempted when `install -e` has not been used.

Either way, please update the unit test to be part of `test_structure.py` (with a similar guard against the call in the event of no install).

Paul-St-Young:
@jtkrogel I can reimplement in Python if this is really the preferred path.
Let me contact the original authors to make sure they are OK with this.

jtkrogel:
@Paul-St-Young Thanks for your flexibility.

prckent:
@Paul-St-Young It is a published algorithm in a scientific journal. A better question is whether they have or know of a BSD licensed python implementation.

Paul-St-Young:
@jtkrogel you are welcome! @prckent this is the first time I've had to deal with licensing issue, so I wanted to be extra careful to not step on any toes. It seems we should be able to make all parties happy :)

Paul-St-Young:
After offline discussion, I agree to make a Python implementation of the algorithms in question with the original authors' blessings. I am closing this PR in favor of a new one when the Python implementation is done.

prckent:
@Paul-St-Young Thanks

PR #1742, opened 2019-07-17
Fix phmsd

fdmalone:
Reimplements the rediagonalisation of the CI Hamiltonian matrix for PHMSD wavefunctions and prints the resulting variational energy. Potentially needs refactoring. Originally read_ph_wavefunction_ would build the ph_struct structures necessary to construct the PHMSD wavefunction. I've split this into two steps where now first we read the list of ci coefficients and occupied strings, compute the variational energy / rediagonalise the hamiltonian and then build the ph_structures. This avoids needing to rebuild / reset the values of the ci coefficients in the ph_structs which looked messy.

I don't really think the the wavefunction factory is the best place for the new routines. I think the PHMSD class should probably take the ci coefficients and occupied orbitals as arguments to the constructor and build the ph_struct internally. Or at least, these additional routines to compute the variational energy should live elsewhere.

I still think there are some issues with how the mean field subtraction term is being calculated, but will leave this to a future pull request.

- [x] Add test input
- [x] Ensure real build works


fdmalone:
This is good to merge.

ye-luo:
Okay to test

PR #1743, opened 2019-07-18
Fix flattening of lists of irregularly sized arrays.

fdmalone:
The list of mo occupancies can potentially contain numpy arrays of different lengths if basis functions are removed due to linear dependencies. ravelling these arrays will not yield the expected result because of this so I had to explicitly unpack the arrays into a big one.


prckent:
OK to test

PR #1745, opened 2019-07-18
Add determinant localization approximation (DLA)

ye-luo:
From https://arxiv.org/abs/1907.04432. @zenandrea please have a look
still need
- [x] manual entry
- [x] unit test

@PDoakORNL Please have a look if I use enum in a good way or not.


prckent:
@PDoakORNL Are you OK with this code?

src/QMCWaveFunctions/TrialWaveFunction.cpp
```diff
@@ -335,6 +333,28 @@ TrialWaveFunction::ValueType TrialWaveFunction::full_ratio(ParticleSet& P, int i
...
 
+TrialWaveFunction::RealType TrialWaveFunction::calcRatio(ParticleSet& P, int iat, ComputeType ct)
+{
+  ValueType r(1.0);
+  std::vector<WaveFunctionComponent*>::iterator it(Z.begin());
+  std::vector<WaveFunctionComponent*>::iterator it_end(Z.end());
+  for (int ii = V_TIMER; it != it_end; ++it, ii += TIMER_SKIP)
+  {
+    myTimers[ii]->start();
+    if (ct == ComputeType::ALL || ((*it)->is_fermionic && ct == ComputeType::FERMIONIC) || (!(*it)->is_fermionic && ct == ComputeType::NONFERMIONIC))
```
PDoakORNL: I think this is the absolute limit for how complex a conditional should need to be.

src/QMCHamiltonians/ECPotentialBuilder.cpp
```diff
@@ -63,15 +63,23 @@ bool ECPotentialBuilder::put(xmlNodePtr cur)
 #ifdef ENABLE_OFFLOAD
   NLPP_algo = "batched"; // set "batched" as the default
 #endif
+  std::string useDLA("no");
```
PDoakORNL: for new code we should use the new naming conventions.  use_DLA

prckent: +1

PDoakORNL:
Also I think the enum is fine.

src/QMCWaveFunctions/TrialWaveFunction.cpp
ye-luo: Do you want me to change something or this is just a comment?

src/QMCHamiltonians/ECPotentialBuilder.cpp
ye-luo: Will fix it soon.

src/QMCWaveFunctions/TrialWaveFunction.cpp
PDoakORNL: Just a comment. Although I think if we end up needing more logic here we should consider a different approach to dispatching this call.

PDoakORNL: And the ComputeType tag sets us up to do that

ye-luo:
@PDoakORNL variable name has been updated.

ye-luo:
Test this please

PR #1750, opened 2019-07-19
Optimization Methods

LeonOtis:
I have been working with @eneuscamman on multiple wave function optimization algorithms. We have implemented several gradient descent based methods as well as a hybrid approach that alternates between descent and the blocked linear method, where the sections of descent optimization identify important directions in parameter space for use in the BLM. These methods are discussed in more detail in sections 2.4 and 2.5 of a recent paper: https://pubs.rsc.org/en/content/articlelanding/2019/CP/C9CP02269D#!divAbstract

Most of my changes have been related to the QMCFixedSampleLinearOptimize class, along with some additions to QMCCostFunction and parts of the linear method engine. User input can specify whether one of the descent methods or the hybrid method should be used as well as step sizes for different parameter types, though defaults are provided. Currently, the descent and hybrid methods are only for ground state optimization. 

For the hybrid method, the number of samples used by the descent section is typically different from that used by the BLM and is provided with a separate input. My current way of ensuring the number of samples changes between descent and BLM iterations has been through modifying the QMCDriver and VMC classes to have knowledge of whether the hybrid method is being used and if the current iteration is descent or BLM, but I am very open to other ways of accomplishing this.

I have also included a basic test of some partial descent optimization followed by VMC (short-H4-descent-16-1), but could provide other tests if needed. Thanks in advance for any and all feedback.


prckent:
Thanks for this! It will take some time to go through.

prckent:
OK to test

prckent:
Test this please

prckent:
I was able to build this locally and have
	short-H4-descent-16-1
	short-H4-descent-16-1-totenergy
pass.

prckent:
@ye-luo , @LeonOtis It is good to see all the updates to this PR, but please could you add a comment here about the current status? What remains to be done?

ye-luo:
To integrate new optimization methods, we feel the need of adding an abstraction layer to capture all the optimizer operations. Considering that optimizers are very different from each other, we start with a small goal which is to integrate the descent and hybrid methods with minimal hack into the existing code. The initial PR adds code in many places, especially in QMCFixedSampleLinearOptimize. Now we are gradually moving them into a new abstraction layer.
- [x] create DescentEngine and HybridEngine
- [] reduce the amount of switches
- [] fix threading

Long term, adaptive and oneshiftonly will also have their own engines and being refactored out of QMCFixedSampleLinearOptimize. Due to the complexity of the code for WF optimization, we are still learning and the current design is clearly not final but just a small step.

ye-luo:
Okay to test

ye-luo:
@LeonOtis We need the following.
- [ ] unit tests for descent and hybrid engines.
- [ ] integration test for hybrid case.

prckent:
Manual entry?

ye-luo:
@prckent I think we will make it in v3.9.

prckent:
Wait.. make a TeX mistake

PR #1752, opened 2019-07-19
Add deterministic integration test for CUDA build on diamondC testsets

Hyeondeok-Shin:
This PR is for #1729.

 References for deterministic tests on CUDA build are added on diamondC_1x1x1_pp and diamondC_2x1x1_pp testsets.

 References were obtained using QMCPACK 3.6.0 on Bora (Quadro P1000). 
 Tests were fully passed on Rhea (Tesla K80) and Summit (Tesla V100), but it needs to keep watching because I saw some tests sometimes fail.

 Hyeondeok Shin


Hyeondeok-Shin:
@prckent Thanks for the comment, because diamondC_1x1x1 has the same references between real & complex, now they are merged and there is now no more IF(QMC_COMPLEX) block on diamondC_1x1x1_pp testset.

prckent:
Test this please

ye-luo:
Okay to test

PR #1753, opened 2019-07-19
Fix reading of initial wavefunction from hdf5.

fdmalone:
Fixes an issue with reading initial wavefunction from hdf5. Only affects systems where nalpha != nbeta


prckent:
OK to test

PR #1754, opened 2019-07-19
Afqmc wavefunction converter

fdmalone:
Adds wavefunction converter script from old afqmc format to hdf5 and also for the real space multi-determinant wavefunction format. In a subsequent PR I will remove plain text wavefunctions.


markdewing:
okay to test

Issue #1746, opened 2019-07-18
Reading in a custom external potential defined on a grid.


kryczko:
Hi all,

I presently have a script written in python to generate a custom external potential. I would like to then read in this potential to qmcpack (via whatever method is easiest i.e. hdf5 seems to be built in) and then perform VMC calculations of electrons interacting with this external potential. Does anyone have ideas what the best/easiest way to do this is?

Kevin

markdewing:
The simplest example of an external potential is probably a harmonic potential in `src/QMCHamiltonians/HarmonicExternalPotential.{h,cpp}`.   An example input using the potential is in `tests/models/sho/vmc_sho.xml`.

HDF files are read using hdf_archive in `src\io`.

This isn't much guidance, but it's a place to start looking.

kryczko:
I saw those files, but the SHO potentials are centred about the atoms and the potential still depends on positions of the particles. Since I don't have an analytic form for external potentials, I would need to have my potential defined numerically on a grid.

markdewing:
See PR #1756 for an example on creating a grid-based potential.    It reuses the existing einspline code to do spline interpolation of values in 3D.

kryczko:
Thanks for providing that code, it looks like it will be very useful. I'm thinking in my case, I would do exactly the same, except my data Array would now consist of the potential I would read in from an hdf5 file. Does that sound like it would do the trick? Is there visualization tools I could use to make sure everything is as it should be?

markdewing:
Reading 'data' from hdf5 should work.  I'm working on updating the example to read from a simple hdf file.

kryczko:
Great, I look forward to seeing the code.

markdewing:
PR is updated.   The name of the file and all the dimensions are fixed in the source, which should be good enough to get started.
I don't know about visualization  - you could make a loop and print out the results of eval_UBspline_3d_d to a file.

I've been using the harmonic potential on the grid to ensure the code gives the same answer as the existing harmonic potential.  In the input file I set the random seed ( `<random seed="1"/>`) and set the mass and energy are 1.0.

kryczko:
Great, sounds good. What is the block of xml that is going into the input file? Should it be something like:
```
<hamiltonian ... >
  <extpot type="GridExt" />
  ...
</hamiltonian>
```

markdewing:
The type of extpot is the value tested in HamiltonianFactory.cpp - `grid` in this case.  So the xml input would look like
```
[...]

PR #1756, opened 2019-07-22
[WIP] Example of grid-based external potential.

markdewing:
Start of an example for #1746.  Creating a PR to make it easier to see the code and code changes.

Uses einspline for grid interpolation.   The boundary conditions on the spline grid are open - this could be changed to periodic depending on the system.

Creates a grid that reproduces the harmonic potential (with mass and energy set to 1.0)

In the input file, the hamiltonian section expot element would look like
```
    <hamiltonian name="h0" type="generic" target="e">
      <extpot type="grid"/>
    </hamiltonian>

```

I think evaluate_sp and the tracemanage code can be ignored for now (?)


markdewing:
I'm going to close this.  It is here to make the example code changes easier to see, and it can still serve that purpose closed.

PR #1767, opened 2019-07-27
C++ codes use strictly cmath instead of math.h

ye-luo:
rename ../src/config/stdlib/math.h to ../src/config/stdlib/math.hpp to avoid confusion
implement qmcplusplus::iszero using C++11 functions only. The implementation is near identical to math.h from glibc.
The non standand iszero from glibc is no more used.


src/QMCHamiltonians/SpaceGrid.cpp
```diff
@@ -15,7 +15,7 @@
 #include <OhmmsData/AttributeSet.h>
 #include <Utilities/string_utils.h>
 #include <cmath>
```
markdewing: Duplicate include of cmath

prckent:
Many crashes in AFQMC tests.

ye-luo:
My second commit only removed a duplicated include<cmath>. I can't understand how the first CI passed, the second CI failed.

ye-luo:
Test this please

ye-luo:
AFQMC unit tests run fine on my workstation.

ye-luo:
quirks didn't appear this time. For this PR, I have nothing really special for @PDoakORNL to review.

PR #1773, opened 2019-07-31
VMC batch develop

PDoakORNL:
MCPopulation introduced.
This will take over rank level population state from MCWalkerConfiguration

Due to interface changes QMCDriverNew and VMCFactoryNew are needed.
I'm starting them very close to the original QMCDriver and VMCFactory so that refactoring will be easier to follow.

A VMCBatched driver can now be created while preserving existing functionality.


src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -0,0 +1,390 @@
...
+  m_param.add(UseDrift, "usedrift", "string");
+  m_param.add(UseDrift, "use_drift", "string");
+
+  prevSteps               = nSteps;
+  prevStepsBetweenSamples = nStepsBetweenSamples;
+}
+
+bool VMCBatched::run()
+{
+  //   resetRun();
```
markdewing: Is the commented-out code intended to be used in the future?

markdewing:
"McPopulation - over 10 billion walkers branched"

markdewing:
Could you say more about MCPopulation.    It appears not to contain the walker storage?  Why does it have the ion ParticleSet and Properties?

src/QMCDrivers/VMC/VMCBatched.cpp
PDoakORNL: I'm trying to fix it so as much as possible the changes from VMC --> VMCBatched will possible to follow. So some if it is intended to be there in the future some of it will never compile since changes have already been made in the class hierarchy and API's.

src/QMCDrivers/VMC/VMCFactoryNew.cpp
```diff
@@ -0,0 +1,48 @@
...
+                                       QMCHamiltonian& h,
+                                       ParticleSetPool& ptclpool,
+                                       HamiltonianPool& hpool,
+                                       WaveFunctionPool& ppool,
+                                       Communicate* comm)
+{
+  int np = omp_get_max_threads();
+  //(SPACEWARP_MODE,MULTIPE_MODE,UPDATE_MODE)
+  QMCDriverInterface* qmc = nullptr;
+  if (VMCMode == 0 || VMCMode == 1) //(0,0,0) (0,0,1)
```
ye-luo: Let us kill this state machine. We will only have 1 VMC driver. PbyP and allP moves are internal details.

ye-luo:
> My hope is to make the changes from VMC --> VMCBatched possible to follow. So some if it is intended to be there in the future and some of it will never compile and be replaced since changes have already been made in the class hierarchy and API's.

@PDoakORNL VMC itself was made super heavy and very hard to follow. I'd like to start VMCBatched with all the functions as no ops and implement them if they are justified to be kept forward. What do you think?

prckent:
Whatever route is taken, we need a working batched VMC in only a few hundreds of hours from today. This doesn't preclude a no-op route but suggests that progressive refactoring would be best. The drivers need a lot of work besides batching but we should keep this work disentangled.

prckent:
As an example of something that can be done elsewhere the VMC equilibration/warmup is clearly done in a nonobvious place, but from a batching point of view it does not matter.

src/QMCDrivers/VMC/VMCFactoryNew.cpp
prckent: One thing at a time

ye-luo:
If the goal is to have a working batched VMC for demonstration and leave the driver cleaning work later, I'm OK with what is here now.

@PDoakORNL are QMCDriverNew and VMCBatched mostly copied and VMCBatched with all the copied content commented?

PDoakORNL:
Now that current production VMCDriver is protected my plan is to refactor with batching as the main goal. There are many other things that should be fixed but until we get batching and then until we can replace the existing driver I'd like to push as much of that as is sane. I want to get to refactoring in place instead of on a separate source path sooner than later. This will hopefully reduce the divergence and merge issues down the road.

ye-luo:
@PDoakORNL Since now the code diverges from the production code path, we have a bit less constraints. Functionality first, this is required before switching over.

prckent:
Please add doxygen comments as this expands.

prckent:
Test this please

ye-luo:
@PDoakORNL The GPU build is broken. This is not a CI issue.

PDoakORNL:
Apologies on that.  I need to protect the legacy CUDA build from some of the new tests.

ye-luo:
Test this please

Issue #1772, opened 2019-07-31
Broken restart code after recent updates


prckent:
https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=restart-8-2-check&date=2019-07-31

The MPI+threads restarts ("8-2") are failing on all platforms after last weeks updates. The pure MPI restart checks are passing.

PDoakORNL:
The problem appears to actually be in VMC section following VMC section. Somehow here reusing the driver allowed some state to be passed forward that was not formally accounted for.

PDoakORNL:
The issue is with the second vmc section output of the initial run. But I'm not sure that the reference output is correct at all.

The reference output is not what I would expect. Can someone explain [initial.out](https://github.com/QMCPACK/qmcpack/blob/develop/tests/io/restart/ref/initial.out) data for the restart 8-2 test (reported to be from 3.0.0)
```shell
=========================================================
  Start VMCSingleOMP
  File Root qmc_short.s001 append = no 
=========================================================
Using existing walkers 
  Resetting Properties of the walkers 1 x 15

<vmc function="put">
  qmc_counter=1  my_counter=1
  time step      = 0.0000e+00
  blocks         = 1
  steps          = 8
  substeps       = 0
  current        = 0
  target samples = 0.0000e+00
  walkers/mpi    = 16

  stepsbetweensamples = 88
```
I don't see any reason that 8 steps should be put or that 88 steps between samples should be arrived at from the following input xml [qmc_short.in.xml](https://github.com/QMCPACK/qmcpack/blob/develop/tests/io/restart/qmc_short.in.xml) looks like this
```XML
   <qmc method="vmc" move="pbyp" checkpoint="-1">
      <parameter name="walkers"             >    16              </parameter>
      <parameter name="blocks"              >    1               </parameter>
      <parameter name="steps"               >    1               </parameter>
      <parameter name="subSteps"            >    0               </parameter>
      <parameter name="timestep"            >    0.0             </parameter>
      <parameter name="warmupSteps"         >    0               </parameter>
   </qmc>
```
I'd also add that the ref output reports to be for a 4 mpi x 4 omp job and not 8 x 2 or 16 x 1.

The current code reports stepsbetweensamples = 2 for the second block of vmc in the intial run.
Both the reference and the current code report stepsbetweensamples = 2 for the restart run.

PDoakORNL:
[...]

PR #1777, opened 2019-08-01
Fix restart test

ye-luo:
subSteps = 0 caused problem #1772 when Nwalker > Nthread.
The test restart-8-2 has Nwalker=16 and Nthread=2.
The distance tables in ParticleSet are not always recomputed when a walker is loaded.
Because they are costly and anyway will be overwritten during pbyp moves.
Without substeps, they are uninitialized and caused wrong results.
In the restart-1-16, the uninitialized value match the recompute in the beginning of a driver and thus pass.

This PR add checks to make sure blocks, steps, substeps are always positive.
`timestep = 0.0` becomes the only way forcing electrons not moving.

removed restart tests reference output because they are not really relevant to the test. The old data is self generated.


ye-luo:
Test this please

PDoakORNL:
test this please

Issue #1776, opened 2019-08-01
Broken CUDA deterministic tests


prckent:
e.g. https://cdash.qmcpack.org/CDash/viewTest.php?onlyfailed&buildid=70893

These need investigating. Could be a tolerance issue or a sign of unreliability.

Hyeondeok-Shin:
@prckent It seems larger tolerance with mixed precision in oxygen than the other environment. (I didn't check on oxygen when making tolerance.) I also found tolerance on MPC for DMC cycle 2 on DMC multisection on CUDA build is missing, so I'll fix them.

ye-luo:
I'd like to have a multi VMC/DMC tests with 2 VMC sections and 2 walkers. Is it better to have a separate test or expand the existing test?

prckent:
@Hyeondeok-Shin we have a Volta now, but I recall you tested on Volta...

@ye-luo It is cheap to add tests and knowing that a 1 walker test passes but a 2 walker test does not would be helpful => separate tests. I assume you are concerned about logic errors in the legacy CUDA?

ye-luo:
@prckent my concern is from the restart test fix I just submitted. 1 walker per thread and 2 walkers per thread is different. Current deterministic tests run with only 1 thread.

prckent:
Understood. More tests then. 

(I'll take back part of my comment about the tests being cheap: our run time for the deterministic tests that run QMC is surprisingly long given the tiny system size and few steps, often a few seconds. We might profile them sometime to see if there is an easy win.)

Hyeondeok-Shin:
@prckent @ye-luo I'll make separated PR for fixing the problem and 2 walkers for thread.

Hyeondeok-Shin:
@prckent I'd like to keep watching more test results on oxygen(Tesla V100-PCIE-32GB). There are three reasons.

 1. I double-checked test but they're obviously passing on summit. (Tesla V100-SXM2-16GB)
 2. Tolerances on oxygen are too large compared to the other environment.  
 3. In oxygen, the same testsets are failing also on the full-precision build.

 It seems not a simple tolerance issue in the mixed precision. Only oxygen with Tesla V100-PCIE-32GB is giving significantly different references to the others. It needs more discussions.

 I'll make PR for missing tolerance for MPC soon.

ye-luo:
All looks fine now https://cdash.qmcpack.org/CDash/viewTest.php?onlypassed&buildid=196096

PR #1778, opened 2019-08-02
Add missing tolerance for MPC on deterministic multi DMC sections.

Hyeondeok-Shin:
This PR is a part of fixing issue #1776 .

Because a tolerance in MPC for DMC multisection on CUDA build was missing. This MPC test was always broken.
So, missing tolerances for MPC is added to pass the test.

Hyeondeok Shin


ye-luo:
Okay to test

PR #1782, opened 2019-08-05
ci diagonalization improvement

mmorale3:
Placed Hamiltonian matrix in shared memory to reduce memory redundancy.
Also, H matrix is only constructed now if new ci coefficients are requested.


markdewing:
okay to test

src/AFQMC/Wavefunctions/WavefunctionFactory.cpp
```diff
@@ -976,41 +976,48 @@ void WavefunctionFactory::computeVariationalEnergyPHMSD(TaskGroup_& TG, Hamilton
...
     using RVector = boost::multi::array<RealType,1>;
     using CMatrix = boost::multi::array<ComplexType,2>;
     // Want a "unique" solution for all cores/nodes.
     if(TG.Global().rank() == 0) {
-      std::pair<RVector,CMatrix> Sol = ma::symEig<RVector,CMatrix>(H);
+      std::pair<RVector,CMatrix> Sol = ma::symEigSelect<RVector,CMatrix>(H,1);
       app_log() << " - Updating CI coefficients. \n";
       app_log() << " - Recomputed coefficient of first determinant: " << Sol.second[0][0] << "\n";
       for(int idet=0; idet < ndets; idet++) {
-        ComplexType ci = Sol.second[idet][0];
```
fdmalone: What's the reason for row/column difference?

mmorale3: the new routine returns a selected number of eigenvalues, which already come in this order from the fortran call to lapack. I decided that it is more natural this way and also avoid additional memory allocation and transposition in the eigenvalue routine.

PR #1783, opened 2019-08-06
Add quantum espresso support to convertpw4qmc (also enable spinors)

lshulen:
This code utilizes the xml output of quantum espresso and the HDF5 files it generates to make an eshdf file suitable for qmcpack.  In order to run this one does:
convertpw4qmc data-file-schema.xml
where data-file-schema.xml is a file generated in the output structure of a typical quantum espresso calculation.  It is also designed to work with noncolinear wave functions that will be needed for enabling spin orbit coupling in qmcpack.  

Other than for spinors, currently the eshdf file generated should be nearly identical to that generated by pw2qmcpack with a couple of exceptions.  First some of the unused symmetry information is omitted.  Second, the total number of gvectors differs slightly, but in all cases I have found this is just zero padding that is of a different length.  Finally, espresso does not currently export the atomic number or valence charge of ionic species in the xml file, so the eshdf file doesn't have these either.  Discussions are ongoing with the espresso developers about this, but this information should NOT be required by qmcpack as it is specified elsewhere.

Finally, for noncolinear wave functions, an attribute "electrons/has_spinors" is added that is an integer flag.  If it is 1, the wavefunctions should be treated as spinors.  To this end, number_of_electrons will contain {n, -1} specifying that the second species is invalid.  Then the coefficients are placed in spin_0 and spin_1 respectively.

The big items remaining on my TODO list are to add details covering this to the manual, to develop the proper tests and to make sure that none of the above changes cause QMCPACK's parsing to generate errors incorrectly.  Likely the existing integration tests that run espresso could be extended as this would avoid adding a number of additional hdf files to the repository.


lshulen:
Pending acceptance of PR #1789, a patch is ready to eliminate the local HDF wrappers and use hdf_archive.  Per discussions with @ye-luo, I will propose a format change to ESHDF whereby the integer indices will be written for every k-point rather than only to k-point 0 and mapping everything to this list.  This will require some testing and updating of builders currently in the code.

prckent:
@rcclay In the future, should we rename the QMCPACK vectors simply as cell_vectors? Calling supercell vector primitive vectors is nonsensical...

ye-luo:
Since we will raise the version of our eshdf format. Introducing breaking changes fixing issues is desired.

lshulen:
With Cody's help, this now handles multiple k-points correctly and has rudimentary testing.  Some documentation has been added to the manual.  Finally, for the multiple k-point functionality, the scheme adopted by the previous pw2qmcpack converter has been adopted whereby all g-vectors are made into a map and are now present and in the same order in every k-point.  Future changes to have g-vectors per k-point are deferred to a future pull request.

lshulen:
Thanks to @camelto2 for running a few tests on QE 6.5.  Checks of a few cases suggest that the converter is working well for that as well.  This can be considered for issue #2230.

ye-luo:
It does require changing XML input files to add bypass flags, right?

lshulen:
For "normal" calculations this is not necessary.  @camelto2 and @rcclay would be better to comment on non collinear wave functions.

lshulen:
Just checked with @camelto2.  It looks like things work without bypass for noncolinear calculations.  However, the code currently aborts complaining about the missing valence etc the tare not provided in the xml part of espresso.  

as an example:
ERROR Primitive cell ion 0 vs supercell ion 0 atomic number not matching: 0 vs 6 for a diamond

This has been discussed somewhat before.  I feel this should be a warning rather than a hard abort, but I believe we can run with a bypass.

camelto2:
The failing test on the GPU side seems to be test_Coulomb_CUDA...which as far as I'm aware nothing in this PR should be  touching any of that. So not sure what is going on there

PDoakORNL:
Generally the CI runs 2 jobs and can run multiple ctest tests at once.  This fails in some cases when contention occurs for the GPU and this causes spurious test failure. I'm working on a good fix without completely giving up concurrent tests, at the moment I'm just running it again.

prckent:
@camelto2 It probably is test_coulomb_cuda that is bad, and as you point out it has nothing to do with this PR, so we'll eventually bypass or fix it. It is a bit concerning that this test has become unreliable recently...

prckent:
Test this please

prckent:
How important are the wfc1.h5 and charge-density.h5 files? (~1MiB total).

lshulen:
Perhaps @camelto2 should comment to be sure.  I believe he put them there for the integration test.  The converter gets the wave function and charge density from these and then puts them in the eshdf file that is generated. If I'm correct then the test would not work without them.

camelto2:
Both are in fact needed for the integration test to work. They are the output from a small QE calculation located in solids/monoO_noncollinear_1x1x1_pp. The test does a symlink of the eshdf file there, and then runs an h5diff on  the converted eshdf file from the wfc1 and charge-density file.

prckent:
By hand CI:
```
--- START build_gccnew Thu Apr 9 14:46:11 EDT 2020
100% tests passed, 0 tests failed out of 38
--- START build_gcccuda Thu Apr 9 14:48:48 EDT 2020
100% tests passed, 0 tests failed out of 36
--- START build_gcccuda_full Thu Apr 9 14:50:58 EDT 2020
100% tests passed, 0 tests failed out of 36
--- START build_gcccuda_complex Thu Apr 9 14:53:13 EDT 2020
100% tests passed, 0 tests failed out of 49
--- START build_gccnew_complex Thu Apr 9 14:55:40 EDT 2020
100% tests passed, 0 tests failed out of 51
--- START build_gccnew_nompi Thu Apr 9 14:58:33 EDT 2020
100% tests passed, 0 tests failed out of 26
--- START build_gccnew_nompi_complex Thu Apr 9 15:00:17 EDT 2020
100% tests passed, 0 tests failed out of 26
--- START build_clangnew Thu Apr 9 15:02:00 EDT 2020
42% tests passed, 22 tests failed out of 38
--- START build_clangnew_complex Thu Apr 9 15:03:19 EDT 2020
31% tests passed, 35 tests failed out of 51
--- START build_clangnew_mixed Thu Apr 9 15:04:46 EDT 2020
42% tests passed, 22 tests failed out of 38
--- START build_clangnew_complex_mixed Thu Apr 9 15:06:07 EDT 2020
31% tests passed, 35 tests failed out of 51
--- START build_clangnew_aos Thu Apr 9 15:07:35 EDT 2020
44% tests passed, 20 tests failed out of 36
--- START build_clangnew_complex_aos Thu Apr 9 15:08:57 EDT 2020
33% tests passed, 33 tests failed out of 49
```

camelto2:
@prckent I just pushed a fix to @lshulen's fork. Once he merges into his branch, it should fix the problem you pointed out. I check it with clang9 and the same array initialization was present. It should now compile and passed all unit tests on my end

ye-luo:
Okay to test

ye-luo:
@camelto2 I updated this branch with latest develop. Could you resolve the conflict on your PR on Luke's fork?

camelto2:
@ye-luo, I just checked my PR on Luke's branch and it is up to date with  the latest develop as well. It should be able to be merged in. I'll ping Luke

camelto2:
forgot to run clang-format on the new files. One last PR to his fork has the correctly formatted code. then hopefully this is good to go

lshulen:
OK, working on commenting...

lshulen:
I believe this should address the changes requested by @rcclay.  Please advise if the format of the comments is not what is requested and include an example (or point to the appropriate section of the manual) if they are not.  Note that I could not find any Doxygen style comments in the QMCTools directory to use as examples.

lshulen:
I'm not understanding the CI failure.  When I look at the logs, I see things like:
hudson.plugins.git.GitException: Command "git checkout -f ca2f3c3609c3e8714e975c5f7876073941a27180" returned status code 128:
stdout: 
stderr: fatal: reference is not a tree: ca2f3c3609c3e8714e975c5f7876073941a27180

Have I mucked something up with my git or is CI having an issue?

prckent:
ALCF CI is not able to complete the checkout for some reason.

I am running checks by hand and will post here in ~30mins.

camelto2:
> --- START build_clangnew_mixed Fri Apr 10 15:00:28 EDT 2020
> 97% tests passed, 1 tests failed out of 38
> --- START build_clangnew_complex_mixed Fri Apr 10 15:02:59 EDT 2020
> 98% tests passed, 1 tests failed out of 51
> --- START build_clangnew_aos Fri Apr 10 15:05:42 EDT 2020


looks like there are two tests failing in clang mixed precision. Is this expected

prckent:
@camelto2 Unfortunately the failures in mixed precision unit tests are expected.

nitrogen, our AMD Rome box highlights some bugs and/or test tolerances that are bad, probably some of both. e.g. https://cdash.qmcpack.org/CDash/viewTest.php?onlyfailed&buildid=111890 . This shows that we have more work to do. In the above it is deterministic-unit_test_hamiltonian that fails (@rcclay ) 
```
...git_QMCPACK_qmcpack/qmcpack/src/QMCHamiltonians/tests/test_force.cpp:315: FAILED:
  REQUIRE( force.forces[0][0] == Approx(9.24061106) )
with expansion:
  9.24078f == Approx( 9.24061106 )
```

PR #1784, opened 2019-08-06
Thread abstraction task block

PDoakORNL:
To prevent openmp contamination in new unified driver.  The barrier logic was necessay in miniqmc to prevent destructors deleting cuda allocated memory being called before all the crowds finished.  These memory deallocations cause device scope synchronization and a huge performance hit.


ye-luo:
I think this is beyond what we need for one driver step 1.
1. The wrapper you add is not sufficient for the minimal need in a VMC driver. At least, omp_get_num_threads() and omp_get_thread_num() for population chunking are missing. Let us first do step 1 with minimal OpenMP and keep your thread block design in mind. Then we deal with wrapping after all the use patterns are shown by step 1.
2. We don't have much experience std::threads on HPC systems. For example, how to control affinity on a multi socket system with hyper-threads. It needs to be investigated as well.
3. Your CUDA async example in miniqmc is a nice optimization but it will not be used in step 1 as well. If we need that implementation, it will probably not be CUDA.

Could you keep this for your own at the moment?

PDoakORNL:
I really would like to get this in now. Enough new code is getting written that dependence on opemmp should be abstracted, I'd also add that is this was already demonstrated in miniqmc and I think enforces clear thinking about the concurrency requirements.

To answer your questions:
1. No, but its sufficient for my next driver PR. It interoperates fine with raw openmp directives (obviously only for the OPENMP specialization) but prevents new code dependent on them. It can be extended as necessary for current needs.
2. For std::thread one generally uses taskset if you only care about a particular socket. That would be sufficient if std::thread is even needed. Since the bottlenecking in small systems is due to insufficient threads feeding the GPU not thread scheduling or core level data locality.
3. For concurrency research going forward (and this is largely a waste of time in miniqmc). I would like to reserve the ability to use std::thread, boost threads, or hpx without the need to create a long term divergent branch. Sooner than later I would like no new code to be written that cannot interoperate with different threading/tasking backends. 

3. b) I don't think its been decided that there will be no CUDA batching in step 1. I'm certainly putting the splines in if I have time.

By writing unit tests that attempt to hold there scope to their actual unit it won't matter that most of the code is openmp only at this time.

prckent:
@ye-luo , @PDoakORNL Did you discuss this last week in person? Can we go ahead with the merge?

ye-luo:
@PDoakORNL is simplifying the PR to minimize the complication.

src/Concurrency/TasksOneToOneOPENMP.hpp
```diff
@@ -0,0 +1,62 @@
...
+{
+  const std::string nesting_error{"TasksOneToOne should not be used for nested openmp threading\n"};
+  if (omp_get_level() > 0)
+      throw std::runtime_error(nesting_error);
+  int throw_count = 0;
+#pragma omp parallel num_threads(num_threads_) reduction(+ : throw_count)
+  {
+    try
+    {
+      f(omp_get_thread_num(), std::forward<Args>(args)...);
```
ye-luo: Do we need a header inclusion for std::forward?
Ping me when this is done. Seems to me the last needed update.

PDoakORNL: we should need <utility> so I added it.

PDoakORNL:
test this please

PDoakORNL:
gcc6.2.0 is having trouble with std::atomic. I'm trying to reproduce this.  No luck with gcc 7.3.0

PDoakORNL:
I've reproduced this now. Power9 doesn't require libatomic. Dizinf.

PDoakORNL:
So gcc 6.5.0 couldn't figure out the auto& and std::ref interaction.  While gcc 6.2.0 seemed only to need libatomic linked.

PDoakORNL:
In light of this x86 gcc6 issues I've decided to protect the std::thread implementation behind a build option QMC_EXP_THREADING until I can test on more platforms and compilers.

PDoakORNL:
test this please

PR #1785, opened 2019-08-07
Fix UHF logic for afqmc molecular wavefunction.

fdmalone:
Remove walker_type string in favour of uhf bool.


ye-luo:
Okay to test

PR #1787, opened 2019-08-08
Fix a bug in Adaptive_Three_Shift

chrislzhao:
@ye-luo This PR fixes a bug that affects the correctness of correlated sampling in the adaptive_three_shift method.

### DIFF
--- a/src/QMCDrivers/QMCFixedSampleLinearOptimize.cpp
+++ b/src/QMCDrivers/QMCFixedSampleLinearOptimize.cpp
@@ -952,15 +952,11 @@ bool QMCFixedSampleLinearOptimize::adaptive_three_shift_run()
 #ifdef HAVE_LMY_ENGINE
   // call the engine to perform update
   EngineObj->wfn_update_compute();
-//std::cout << "optimization here 0.5" << std::endl;
 #else
   solveShiftsWithoutLMYEngine(shifts_i, shifts_s, parameterDirections);
 #endif
 
   // size update direction vector correctly
-  //for (int i = 0; i < EngineObj->good_solve().size(); i++)
-  //  app_log() << EngineObj->good_solve().at(i) << "  ";
-  //app_log() << endl;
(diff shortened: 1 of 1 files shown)


markdewing:
okay to test

src/QMCDrivers/QMCFixedSampleLinearOptimize.cpp
```diff
@@ -1051,7 +1051,7 @@ bool QMCFixedSampleLinearOptimize::adaptive_three_shift_run()
     for (int j = 0; j < parameterDirections.size(); j++)
     {
       if (j != central_index)
-        parameterDirections.at(j).at(i + 1) -= parameterDirections.at(1).at(i + 1);
```
ye-luo: Please remove the following commented line.

chrislzhao: Done

ye-luo:
Please explain what is the bug and how the correct is affected. Everything else looks good.

chrislzhao:
@ye-luo In the adaptive three shift method, three different shift values are used to stabilize the LM eigensolver, and leads to three different sets of wave function parameter updates. It will then do a correlated sampling with wave function parameters updated based on the medium-sized shift as the guiding function, in order to decide which shift is the best (gives the lowest cost function). This bug affects the correlated sampling part, in which the code will need to loop over shifts and update the wave function to make sure the wave functions are updated correctly based on that shift. Within the loop, it needs to check whether the shift is the medium sized one, and if it is then there is no needs to update the wave function. However, this check was wrongly coded. Instead of check "k == central_index", it does "k == num_shifts". Therefore the code used an incorrect wave function as the medium shift updated wave function. 

Another change I made would not change the current code. At line 1054 I simply change parameterDirections.at(1) to parameterDirections.at(central_index). Since 3 shifts are used, central_index is 1. But this may be a problem in the future when we decide to use more shifts.

PR #1789, opened 2019-08-13
Add functionality to hdf_archive

lshulen:
This adds the ability to use hdf_archive to write from std::vector to datasets of arbitrary dimensionality specified by the user.  It also adds the ability to read from slices of a dataset to a std::vector.

The new interfaces are explained in comments in hdf_archive.h and minimal unit tests are provided in test_hdf_manual_reorder.cpp.


markdewing:
okay to test

lshulen:
After discussions with @ye-luo, I've decided to move this to WIP.  There may be a way to cleanly do most of what I want with things like hyperslab_proxy.  I will try to rewrite the unit test using the previous API.  If I can do that, I will delete my changes to hdf_archive but keep the new unit test to document the functionality.

src/io/tests/test_hdf_manual_reorder.cpp
```diff
@@ -0,0 +1,99 @@
...
+using namespace qmcplusplus;
+
+
+TEST_CASE("hdf_write_reorder", "[hdf]")
+{
+  hdf_archive hd;
+  hd.create("test_write_reorder.hdf");
+
+  vector<double> v{1.1, -2.2, 3.3, -4.4, 5.5, -6.6, 7.7, -8.8, 9.9};
+  vector<hsize_t> dims{3,3};
```
markdewing: Would the test be more powerful if the dimensions were not square?   I would be concerned about getting the dimensions reversed at some point in the code path.

markdewing: I looked more closely and see the following test handles the non-square case while reading.

lshulen:
Anyone have any thoughts?  There are three options specified in the unit test (test_hdf_manual_reorder) on the functionality that I need.  I prefer the third.  It would be nice to introduce some functionality that does not require the ability to resize the data structure that you are reading into as this would allow for a more generic use of the wrappers.  (note that Miguel did something along these lines for functionality involving multi).  However, I really just want to get convertpw4qmc working so that I Ray and I can progress with the spin orbit work.

I will leave this as WIP so people can comment for another day, then I'll just change it to match my preference tomorrow if nobody objects.

src/io/tests/test_hdf_manual_reorder.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+  hd.write(allData, "matrix");
+  hd.close();
+
+  hdf_archive hd2;
+  okay = hd2.open("test_read_partial.hdf");
+  REQUIRE(okay);
+
+  // method 1 (relying on vector in hdf_stl)
+  vector<int> readSpec{1,-1};
```
PDoakORNL: naming conventions. And here (std::)vector isued while ohhms Vector was used above.

src/io/tests/test_hdf_manual_reorder.cpp
```diff
@@ -0,0 +1,210 @@
...
+#include "io/hdf_archive.h"
+#include <iostream>
+#include <vector>
+#include <OhmmsPETE/OhmmsVector.h>
+#include <OhmmsPETE/OhmmsMatrix.h>
+#include <OhmmsPETE/OhmmsArray.h>
+
+using std::cout;
+using std::endl;
+using std::vector;
```
PDoakORNL: I think this is best not to use since the code contains its own Vector type.  Also it is not used uniformly below.

src/io/tests/test_hdf_manual_reorder.cpp
```diff
@@ -0,0 +1,210 @@
...
+  REQUIRE(m(1,1) == Approx(v[4]));
+  REQUIRE(m(1,2) == Approx(v[5]));
+}
+
+
+// do this in three ways.  
+// 1. relying on my overloading in std::vector (do not perfer this, too specific)
+// 2. rely on my understanding of hyperslab_proxy (seems to require you to know
+//          too much about the dimensions in the dataset and is prone to user error)
+// 3. use hyperslab_proxy to handle the back end api, but specify the slice to choose
```
PDoakORNL: Do we want to have people reading into different buffer object types? Is that why 3 would be better?

PDoakORNL:
Also why two different proxy objects?  If we go with 3 can we get rid of one?

src/io/tests/test_hdf_manual_reorder.cpp
lshulen: The initial code I wrote was using std::vector for the descriptor.  In method 3, I switched to a templated integer container type (IC) as is used elsewhere in the hdf_archive stuff.  That way one could use either.

src/io/tests/test_hdf_manual_reorder.cpp
lshulen: Fair enough, can clean this up and just explicitly use std:: in the unit test if needed.

src/io/tests/test_hdf_manual_reorder.cpp
lshulen: This is a philosophical question about how hdf_archive works.  Currently if you want to a 2D dataset out of an HDF5 file, it wants you to read into a 2d type of object like Ohmms::Matrix.  It seems a somewhat inconsistent choice to say when you are reading partial content out of a field that you should use a linear buffer like array regardless of the dimensionality of the slice you are reading.  This is not to say that I always agree with that design choice, but we should probably be consistent.

ye-luo:
I prefer method 2 which is more general but we need to trim down input and build smarter default.

src/io/tests/test_hdf_manual_reorder.cpp
PDoakORNL: That's probably better then.

src/io/tests/test_hdf_manual_reorder.cpp
PDoakORNL: I think its worth while to make the tests as easy to understand as we can.

PDoakORNL:
Added a container_proxy for Matrix which I think is necessary here.

src/io/tests/test_hdf_manual_reorder.cpp
PDoakORNL: I think you are correct and I found a way to make that less painful.  See commits.  I'm out as a reviewer now.

src/io/hdf_dataproxy.h
```diff
@@ -47,6 +48,45 @@ inline bool get_space(hid_t grp, const std::string& aname, int rank, hsize_t* di
...
+    }
+  }
+
+  offset.resize(readSpec.size());
+  count.resize(readSpec.size());
+  numElements = 1;
+  for (int i = 0; i < readSpec.size(); i++)
+  {
+    offset[i] = readSpec[i];
+    count[i] = 1;
```
ye-luo: readSpec is a mixed presentation of offsets and counts. The default count = 1 is very cryptical.

src/io/hdf_archive.h
```diff
@@ -203,6 +235,37 @@ class hdf_archive
...
+  {
+    if (Mode[NOIO])
+      return true;
+    hid_t p = group_id.empty() ? file_id : group_id.top();
+    h5data_proxy<T> e(data);
+    return e.read(p, aname, readSpec);
+  }
+
+  template<typename T, typename IC>
+  bool readEntry2(T& data, const IC& readSpec, const std::string& aname)
```
ye-luo: This is the API used by method 3.
I'm concerned about adding this very specific APIs.
This PR only needs read with a particular pattern. Then if we need to write, add a symmetric API.
Then we need to read/write hyperslab with both offsets and counts, we add another pair of APIs?
Method 2 hyperslab is a more generic method. One for all these cases.

src/io/tests/test_hdf_manual_reorder.cpp
```diff
@@ -0,0 +1,212 @@
...
+  dims_unused[0] = 1;
+  dims_unused[1] = 2;
+  TinyVector<int, 2> dims_local;
+  dims_local[0] = 1;
+  dims_local[1] = 4;
+  TinyVector<int, 2> offsets;
+  offsets[0] = 1;
+  offsets[1] = 0;
+
+  hyperslab_proxy<Matrix<double>, 7> pxy1(outbuffer1, dims_unused, dims_local, offsets);
```
ye-luo: We can do smarter defaults for dims_unused, offsets, dims_locals instead of forcing users to specify all of them.

When using hyperslab, I consider
(offset)
(offset, dim_local)
(offset, dim_local, dim_all)
all useful for reading
but only the last one is valid for writing.

src/io/hdf_dataproxy.h
```diff
@@ -47,6 +48,45 @@ inline bool get_space(hid_t grp, const std::string& aname, int rank, hsize_t* di
   return thesame;
 }
 
+/** free function to go from spec to dimensionality of memory space and data space
```
ye-luo: I do see the need of query dataset on the file via hdf_archive. it is currently only allowed inside the class.

lshulen:
To me the goal of wrapping the entire functionality of hdf5 is way beyond the scope of a couple of classes inside the IO directory of QMCPACK.  Making things simpler and less error prone is fine, but this is getting overwrought.  What about chunking, complex data types compression etc?  I disagree with much of the design of the way this wrapper is written anyway.  I think it overly restrictive to have to define datatypes to match the dimensionality of the data in the hdf dataset (as apparently did Peter where he introduced a workaround that is satisfactory for the test case but this would have to be repeated with every new type under consideration).  Then using hyper slab we revert this to a linear datatype because the complexity gets too much.  

We clearly have different concepts about where to go with this, which is fine.   I have made a good faith effort to work with what is here, but I feel in the end I have wasted a bunch of effort.  I am withdrawing this PR and will stick with the wrappers around bare hdf calls that I had in convertpw4qmc.  I am willing to harden that with unit tests as needed, but I don't feel like having this debate anymore is a worthwhile use of time.

prckent:
One step at a time folks. I had assumed that this would go in. The choice is bare HDF5 in convert4pwqmc or the contents of this PR (@ye-luo ).

mmorale3:
I haven’t read the thread completely yet, but I think this can already be done with the current specializations of read/write for multi:: array_ref. I use this all the time in afqmc.

ye-luo:
There are good things coming out of this PR. I'd like to keep them.
We are not supporting full hdf5 features and then it means there is restriction we put in the wrapper.

There are two cases.
1. writing using hdf_hyperslab. Users do need to specify a bit more info and this is required by the hdf5 library. hdf_hyperslab supports stripping (hyperslab selectron) on the filespace but requires memory side a single contiguous space. If striping on the memory side is needed, hdf_double_hyperslab is available.

2. writing full data. We currently make it as simple as h5.write(obj, name) and use template traits to figure out the metadata of obj. This brings convenience that you don't need to say any info about obj. However if some datatype was not supported, we need to implement new template specialization. Thanks @PDoakORNL. I see this is better than passing multi dimension info via an argument std::vector.

We do need to improve documentation in this part of code.
I do want to avoid HDF5 becoming another XML problem which we are having now. The use pattern is simple and the wrapper is capable. Then we stick to the wrapper.

ye-luo:
I reopened the ticket because I'm taking action to continue fixing the existing problems and take the burden from @lshulen. I didn't mean @lshulen need to address the requested changes. 

I worked a bit last night. https://github.com/ye-luo/qmcpack/tree/simplify-hdf_hyperslab
Although not yet finished.
For the two points I mentioned above.
Full data read/write. I find a bit redundancy in the template specialization.
hypeslab. I feel the template specialization currently used is overcomplicated. We only need to specialize each container std::vector<T>, Vector<T>, Matrix<T>, Array<T> instead of per underlying datatype Matrix<std::complex<T>>.

There is some reshaping and resize mechanism involved and I'm looking at logical way of doing it.
In addition, I'm look at a convenient way of querying the underlying data dimension.

ye-luo:
I'm able to enable complex inplace without adding any template specialization.
There is a high chance to enable the following for reading
```
hyperslab_proxy<Matrix<complex<double>>, 3> pxy3(outbuffer3, readspec);
   hd2.read(pxy3, "matrix");
```
althought safeguard is needed for preventing writing with hyperslab constructed this way.

lshulen:
My personal preference here would actually be to remove method 1 altogether.  I think you get the same thing out of method 3 and it pollutes the code less.  Also, on calling the method for the third option read2. I had done this as a placeholder so that everything could coexist.  My thought would be to either name it read and replace the call from method 1, or to name it something like readHyperslab (or whatever the style guide says the right capitalization / _ convention is for a member function).  What would be preferred?

Luke

ye-luo:
I mean "cleaning" deleting method 1 since you commented in the source this is not the preferred way.
readHyperslab is a better name. It avoids the problem of many "read" functions only differ by arguments. We have enough trouble with SPOSet::evaluate.

lshulen:
Cleanup is finished

ye-luo:
Is the added writeEntry correspond to method 1 or 3?

lshulen:
This is a good catch.  The writeEntry method was added so that I could reshape linear memory into an arbitrary hdf dataset.  As this has refactored towards using the in place functionality, I removed the necessary hooks in dataproxy to make that work.  This method should similarly be removed.  I'll take care of it in the repo now.

PR #1793, opened 2019-08-14
Quick fixes to allow more efficient working with force estimators

tiihonej:
Two one-liners to allow
1) adding 'forces=yes' to Hamiltonian components in nexus (still has to be done manually)
2) fixing PBC simulations involving such terms from crashing

Since I'm relatively new to the code, I'm marking this WIP to discuss (especially with @rcclay) whether the fixes should be done more thoroughly. If not, WIP status can be lifted.

### DIFF
--- a/src/LongRange/LRHandlerBase.h
+++ b/src/LongRange/LRHandlerBase.h
@@ -193,7 +193,7 @@ struct LRHandlerBase
       PosType k = kpts[ki];
       for (int iat = 0; iat < Zat.size(); iat++)
       {
-        grad1[iat] -= Zat[iat] * k * Fkg[ki] * (e2ikrA_r(iat, ki) * rhokB_i[ki] - e2ikrA_i(iat, ki) * rhokB_r[ki]);
+        grad1[iat] -= Zat[iat] * k * Fk[ki] * (e2ikrA_r(iat, ki) * rhokB_i[ki] - e2ikrA_i(iat, ki) * rhokB_r[ki]);
       }
     }
 #endif
--- a/src/LongRange/LRHandlerBase.h
+++ b/src/LongRange/LRHandlerBase.h
@@ -193,7 +193,7 @@ struct LRHandlerBase
       PosType k = kpts[ki];
       for (int iat = 0; iat < Zat.size(); iat++)
       {
-        grad1[iat] -= Zat[iat] * k * Fk[ki] * (e2ikrA_r(iat, ki) * rhokB_i[ki] - e2ikrA_i(iat, ki) * rhokB_r[ki]);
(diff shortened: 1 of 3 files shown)


src/LongRange/LRHandlerBase.h
```diff
@@ -193,7 +193,7 @@ struct LRHandlerBase
       PosType k = kpts[ki];
       for (int iat = 0; iat < Zat.size(); iat++)
       {
-        grad1[iat] -= Zat[iat] * k * Fkg[ki] * (e2ikrA_r(iat, ki) * rhokB_i[ki] - e2ikrA_i(iat, ki) * rhokB_r[ki]);
```
ye-luo: The change indicates a bug. Why you say it just makes the code more efficient?
If it is a bug, we need a test (hopefully unit test) to capture the bug and get pass after the fix.

tiihonej: Sorry, a bad choice of words. By efficient I did not mean execution of the code but the workflow: fixing a bug such as this allows for more straightforward post-processing (and has to be done anyway). I'm currently working with @rcclay on an alternative fix using EwaldHandler3D, which apparently is better tested to start with. Will push it here shortly.

rcclay: @ye-luo LRHandlerBase has not been validated for forces--only EwaldHandler3D.  Ideally, this whole part of the code should be error trapped for users wanting to use forces. Since this is an experimental feature, this hasn't been done yet.

jtkrogel: @rcclay Could you comment a little more about the functionality of/distinction between LRHandlerBase and EwaldHandler3D?  What are the responsibilities of each?

jtkrogel:
@rcclay What would the timeline/path be for moving the changes into develop?  If the changes support science, then I would lean toward a short path.

What unit tests do you expect to break?

rcclay:
@jtkrogel The path forward is fixing and validating the Esler style breakup for use with forces.  This way of doing things is significantly more complicated in ewald...  like ideally, you would have a breakup for energy and a breakup for forces (and a separate one for stresses), and you would like to have some code logic to make sure that unnecessary breakups aren't done.  It's a fair amount of code complexity and debugging when for Ewald it just works out of the box.  

I expect this would break all the unit tests in PBC that calculate energies, and depending on the tolerances on the statistical tests, probably those those as well.  Esler breakup converges a lot faster w.r.t. kspace cutoff than Ewald, and that error will be very detectable.

tiihonej:
Okay, in that case I will close the pull request for now and continue using it only for testing of forces.

PR #1794, opened 2019-08-14
Enable Hamiltonians to deal with complex parameters

chrislzhao:
Hi @ye-luo this is the third part of a sequence of PRs that eventually make LM work with complex wave function parameters. In this PR, the Hamiltonian is made compatible with complex wave function parameters in the following way:
1) Kinetic energy operator is now compatible with complex gradients and laplacians. 
2) Nonlocal ECP with AoS (only) can compute local energy and its derivatives w.r.t. complex wave function parameters.


ye-luo:
Why evaluateDerivativesForNonLocalPP and evaluateDerivatives needs to be seperated? I understand you need complex but I think the large duplication can be reduced.

chrislzhao:
@ye-luo I separated these two functions because 

1) evaluateDerivativesForNonLocalPP avoids evaluating energy derivatives that are not used in nonlocal ECP derivatives 

2) for multiSlaterDeterminantFast class, its evaluateDerivatives function would not work for the purpose of nonlocal ECP derivatives. Because to evaluate the derivative of nonlocal ECPs, one needs to evaluate wave function derivatives for each point on the spherical grid. For each point, the particle set class object would first move to that point, evaluate wave function ratios and derivatives, and move back. However, for multiSlaterDeterminantFast class, its determinant values are never updated based on the new point, instead, it computes the new determinant values and stores them as "new_detValues". But the evaluateDerivatives uses "detValues", which are never updated. Therefore I have to make a separate function to use "new_detValues".

ye-luo:
1. evaluateDerivatives is supposed to evaluate Wavefunction derivatives. The Kinetic energy derivative needs to be taken out. As a temporal solution, could you move WF derivative part from evaluateDerivatives to evaluateDerivativesWF and make it complex, call evaluateDerivativesWF from evaluateDerivatives for backward compatibility and call evaluateDerivativesWF in NLPP derivatives. I'd like to reduce code duplication.
2. In NLPP derivative, the code moves one electrons to a quadrature point and accept the move.
At the this point, if the whole WF is recomputed from scratch, there should be no change. Then it calls evaluateDerivative to evaluate all the derivatives with respect to all the parameters. Then it proposes a reverse move and accept it. By design, this should work with any WaveFunctionComponents. If evaluateDerivatives of multiSlaterDeterminantFast doesn't work this way, it needs a fix. I don't see anything as "new" because all the moves are accept and the WF is current.

chrislzhao:
@ye-luo I can move the wave function derivatives outside of the evaluateDerivatives function. 

In terms of NLPP, yes the move is accepted, but only on the ParticleSet part. The wave function needs to call evaluateLog to update all of its values, which is not done in the current NLPP code. For Jastrow factor it is fine since it only needs the distance table, which is updated with the ParticleSet.

ye-luo:
This indicates the old code is broken in logic. Probably in the past only Jastrows are optimized and there is no need to accept the WF but now you do need to accept the WF as well.
It should be
```
for each QP
{
  W.makeMove
  psi.ratio
  psi.acceptMove
  W.acceptMove

  psi.evaluateDerivatives
  // reverse direction
  W.makeMove
  psi.ratio
  psi.acceptMove
  W.acceptMove
```
This is a working solution. but a nice solution will be using VirtualPaticleSet with all the QPs batched and no real moves involved. I'm working on this but this is outside the scope of this PR and takes time to land. In this PR, we just need a working solution here.

chrislzhao:
@ye-luo I have modified the code in a different way to avoid duplication. Now the evaluateDerivative function has a one more parameter: wf_deriv_only. This parameter is set to be false by default, but once it is true, all the codes used to compute kinetic energy derivatives will be skipped. This is set to be true in the nlpp derivative function. 

I have also modified the nlpp evaluateValueAndDerivative function so that it calls psi.evaluateDeltaLog once the particle is moved on the grid.

ye-luo:
@chrislzhao I think your last commit needs to be reverted.
1. wf_deriv_only flag flying here and there inside the routine is not acceptable. break the evaluateDerivative and take out the last step which calculates the kinetic energy derivatives. Some intermediate computation which you flagged can be always computed, although it s not optimal.
When you have
```
boo(flag)
{
  if(flag)
  {
    do A
  }
  do B
}

main()
{
   boo(flag)
}
```
it is better to call explicitly.
```
booAB()
{
  do A
  booB
}

booB()
{
  do B
}

main()
{
  if(flag)
    booAB()
  else
    booB()
}
```
2. Please do not randomly shopping functions. How did you find evaluateDeltaLog suitable to use? This function cost N^3. If you accept a move, the cost is only N^2.

chrislzhao:
@ye-luo How would this avoid code duplication? If you factorize the code used to compute the wave function derivatives and make it a new function. Then one needs to go through the almost the same code to compute kinetic derivatives. Are you suggesting that I store the intermediates used for kinetic energy derivative in the function for wave function derivatives? So that it can be used later on?

ye-luo:
@chrislzhao emailed you for scheduling a call directly.

chrislzhao:
@ye-luo I have made the changes based on our discussion.

chrislzhao:
@ye-luo I have separated the evaluation of wave function derivatives and kinetic energy derivatives.

src/QMCWaveFunctions/WaveFunctionComponent.h
```diff
@@ -351,6 +351,10 @@ struct WaveFunctionComponent : public QMCTraits
                                    std::vector<ValueType>& dlogpsi,
                                    std::vector<ValueType>& dhpsioverpsi);
 
+  virtual void evaluateDerivativesWF(ParticleSet& P,
```
ye-luo: Documentation. Please mention the difference from evaluateDerivatives.

src/QMCWaveFunctions/TrialWaveFunction.cpp
```diff
@@ -667,6 +668,38 @@ void TrialWaveFunction::evaluateDerivatives(ParticleSet& P,
   }
 }
 
+void TrialWaveFunction::evaluateDerivativesWF(ParticleSet& P,
+    const opt_variables_type& optvars,
+    std::vector<ValueType>& dlogpsi,
+    bool project)
+{
+  //     // First, zero out derivatives
```
ye-luo: Do these commented line still make sense? Update or delete.

src/QMCDrivers/QMCCostFunction.cpp
```diff
@@ -294,18 +294,13 @@ void QMCCostFunction::checkConfigurations()
         std::vector<Return_t> Dsaved(NumOptimizables, 0.0);
         std::vector<Return_t> HDsaved(NumOptimizables, 0.0);
 
-//FIXME the ifdef should be removed after the optimizer is made compatible with complex coefficients
-#ifndef QMC_COMPLEX
+        //FIXME the ifdef should be removed after the optimizer is made compatible with complex coefficients
```
ye-luo: Remove?

src/QMCDrivers/QMCCostFunction.cpp
```diff
@@ -411,18 +406,17 @@ void QMCCostFunction::engine_checkConfigurations(cqmc::engine::LMYEngine* Engine
         std::vector<Return_rt> rDsaved(NumOptimizables, 0.0);
         std::vector<Return_rt> rHDsaved(NumOptimizables, 0.0);
 
-//FIXME The ifdef should be removed after the optimizer is compatible with complex wave function parameters
-#ifndef QMC_COMPLEX
-        psiClones[ip]->evaluateDerivatives(wRef, OptVariablesForPsi, Dsaved, HDsaved);
-#else
+        //FIXME The ifdef should be removed after the optimizer is compatible with complex wave function parameters
```
ye-luo: Remove?

src/QMCDrivers/QMCCostFunction.cpp
```diff
@@ -436,7 +430,7 @@ void QMCCostFunction::engine_checkConfigurations(cqmc::engine::LMYEngine* Engine
           der_rat_samp.at(i + 1) = rDsaved.at(i);
 
         // evaluate local energy
-        etmp = hClones[ip]->evaluate(wRef);
+        //etmp = hClones[ip]->evaluate(wRef);
```
ye-luo: Remove 2 lines?

src/QMCDrivers/WaveFunctionTester.cpp
```diff
@@ -1839,8 +1839,8 @@ void WaveFunctionTester::runDerivNLPPTest()
   int Nvars = wfVars.size();
   std::vector<ValueType> Dsaved(Nvars);
   std::vector<ValueType> HDsaved(Nvars);
-  std::vector<RealType> rDsaved(Nvars);
-  std::vector<RealType> rHDsaved(Nvars);
+  //std::vector<RealType> rDsaved(Nvars);
```
ye-luo: Remove?

src/QMCHamiltonians/NonLocalECPComponent.cpp
```diff
@@ -140,11 +140,7 @@ NonLocalECPComponent::RealType NonLocalECPComponent::evaluateOne(ParticleSet& W,
     {
       deltaV[j] = r * rrotsgrid_m[j] - dr;
       W.makeMove(iel, deltaV[j]);
-#if defined(QMC_COMPLEX)
-      psiratio[j] = psi.ratio(W, iel) * sgridweight_m[j] * std::cos(psi.getPhaseDiff());
-#else
-      psiratio[j] = psi.ratio(W, iel) * sgridweight_m[j];
-#endif
+      psiratio[j] = psi.full_ratio(W, iel) * sgridweight_m[j];
```
ye-luo: Rename the full_ratio function to evaluateFullRatio in TrialWaveFunction

src/QMCHamiltonians/NonLocalECPotential.deriv.cpp
```diff
@@ -103,23 +103,23 @@ NonLocalECPComponent::RealType NonLocalECPComponent::evaluateValueAndDerivatives
...
-  Matrix<RealType> dratio(optvars.num_active_vars, nknot);
-  std::vector<RealType> dlogpsi_t(dlogpsi.size(), 0.0);
-  std::vector<RealType> dhlogpsi_t(dlogpsi.size(), 0.0);
+  Matrix<ValueType> dratio(optvars.num_active_vars, nknot);
+  std::vector<ValueType> dlogpsi_t(dlogpsi.size(), 0.0);
+  std::vector<ValueType> dhlogpsi_t(dlogpsi.size(), 0.0);
 
-  std::vector<ValueType> dlogpsi_ct(dlogpsi.size(), 0.0);
-  std::vector<ValueType> dhlogpsi_ct(dlogpsi.size(), 0.0);
+  //std::vector<ValueType> dlogpsi_ct(dlogpsi.size(), 0.0);
```
ye-luo: Remove!

src/QMCHamiltonians/NonLocalECPotential.deriv.cpp
```diff
@@ -137,26 +137,39 @@ NonLocalECPComponent::RealType NonLocalECPComponent::evaluateValueAndDerivatives
     for (int j = 0; j < nknot; j++)
     {
       PosType pos_now = W.R[iel];
+      //W.makeMoveAndCheck(iel, deltarV[j]);
+
+//#if defined(QMC_COMPLEX)
```
ye-luo: Remove!!

src/QMCHamiltonians/NonLocalECPotential.deriv.cpp
```diff
@@ -137,26 +137,39 @@ NonLocalECPComponent::RealType NonLocalECPComponent::evaluateValueAndDerivatives
...
+//      psiratio[j] = psi.ratio(W, iel) * std::cos(psi.getPhaseDiff());
+//#else
+//      psiratio[j] = psi.ratio(W, iel);
+//#endif
+      psiratio[j] = psi.full_ratio(W, iel);
       psi.resetPhaseDiff();
 
       //use existing methods
       W.acceptMove(iel);
+      psi.acceptMove(W, iel);
```
ye-luo: The call sequence matters.
W.makeMove
psi.ratio
psi.accept
W.accept

src/QMCHamiltonians/NonLocalECPotential.deriv.cpp
```diff
@@ -137,26 +137,39 @@ NonLocalECPComponent::RealType NonLocalECPComponent::evaluateValueAndDerivatives
...
+      psi.evaluateDerivativesWF(W, optvars, dlogpsi_t);
       for (int v = 0; v < dlogpsi_t.size(); ++v)
-        dratio(v, j) = std::real(dlogpsi_t[v]);
+        dratio(v, j) = dlogpsi_t[v];
 
       PosType md = -1.0 * deltarV[j];
       W.makeMove(iel, md);
       W.acceptMove(iel);
+      ValueType tmp_ratio = psi.full_ratio(W, iel);
+      psi.acceptMove(W, iel);
```
ye-luo: Following the above call sequence? Do we have sufficient test? A test with Jastrow should catch this.

src/QMCHamiltonians/QMCHamiltonian.h
```diff
@@ -43,6 +43,7 @@ class QMCHamiltonian
   typedef QMCHamiltonianBase::RealType RealType;
   typedef QMCHamiltonianBase::ValueType ValueType;
   typedef QMCHamiltonianBase::Return_t Return_t;
+  typedef QMCHamiltonianBase::Return_ct Return_ct;
```
ye-luo: Is Return_ct type still used? If not, remove all

src/QMCWaveFunctions/DiffWaveFunctionComponent.h
```diff
@@ -84,6 +84,14 @@ struct DiffWaveFunctionComponent
                                    std::vector<ValueType>& dlogpsi,
                                    std::vector<ValueType>& dhpsioverpsi)=0;
 
+  virtual void evaluateDerivativesWF(ParticleSet& P,
```
ye-luo: Need documenation

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
```diff
@@ -550,6 +550,7 @@ void MultiSlaterDeterminantFast::evaluateDerivatives(ParticleSet& P,
                                                      std::vector<ValueType>& dlogpsi,
                                                      std::vector<ValueType>& dhpsioverpsi)
 {
+  this->evaluateDerivativesWF(P, optvars, dlogpsi);
```
ye-luo: Remove `this->`. You are calling the local one. Please remove all similar use.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
```diff
@@ -644,25 +645,24 @@ void MultiSlaterDeterminantFast::evaluateDerivatives(ParticleSet& P,
...
-            cdet += CSFexpansion_p[cnt] * detValues_up[upC] * detValues_dn[dnC] * psiinv;
             q0 += (tmp1 * laplSum_up[upC] + tmp2 * laplSum_dn[dnC]);
             for (size_t l = 0, j = N1; l < NP1; l++, j++)
               v1 += tmp1 * static_cast<ValueType>(dot(P.G[j], grads_up(upC, l)) - dot(myG_temp[j], grads_up(upC, l)));
             for (size_t l = 0, j = N2; l < NP2; l++, j++)
               v2 += tmp2 * static_cast<ValueType>(dot(P.G[j], grads_dn(dnC, l)) - dot(myG_temp[j], grads_dn(dnC, l)));
             cnt++;
           }
-          dlogpsi[kk] = cdet;
+          //dlogpsi[kk] = cdet;
```
ye-luo: Remove!!!

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
```diff
@@ -644,25 +645,24 @@ void MultiSlaterDeterminantFast::evaluateDerivatives(ParticleSet& P,
...
           //convert(cdet, dlogpsi[kk]);
-          ValueType dhpsi = (RealType)-0.5 * (q0 - cdet * lapl_sum) - cdet * gg - v1 - v2;
+          ValueType dhpsi = (RealType)-0.5 * (q0 - dlogpsi[kk] * lapl_sum) - dlogpsi[kk] * gg - v1 - v2;
           //ValueType dhpsi =  -0.5*(tmp1*laplSum_up[upC]+tmp2*laplSum_dn[dnC]
```
ye-luo: Please also remove these commented lines.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.h
```diff
@@ -132,6 +132,10 @@ class MultiSlaterDeterminantFast : public WaveFunctionComponent
                            std::vector<ValueType>& dlogpsi,
                            std::vector<ValueType>& dhpsioverpsi);
 
+  void evaluateDerivativesWF(ParticleSet& P,
```
ye-luo: Please use override key word to explicitly overriding the virtual function. This should help us writing better code.
evaluateDerivativesWF() override
Only needed for the derived class function declaration. Not needed in the definition.
Please mark all the overriding functions in this class.

src/QMCWaveFunctions/Jastrow/DiffOneBodyJastrowOrbital.h
```diff
@@ -131,6 +131,36 @@ class DiffOneBodyJastrowOrbital : public DiffWaveFunctionComponent
                            const opt_variables_type& active,
                            std::vector<ValueType>& dlogpsi,
                            std::vector<ValueType>& dhpsioverpsi)
+  {
+    this->evaluateDerivativesWF(P, active, dlogpsi);
```
ye-luo: Remove "this->"

src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
```diff
@@ -163,6 +163,39 @@ class DiffTwoBodyJastrowOrbital : public DiffWaveFunctionComponent
                            const opt_variables_type& active,
                            std::vector<ValueType>& dlogpsi,
                            std::vector<ValueType>& dhpsioverpsi)
+  {
+    if (myVars.size() == 0)
+      return;
+    this->evaluateDerivativesWF(P, active, dlogpsi);
```
ye-luo: Remove "this->"

src/QMCWaveFunctions/TrialWaveFunction.cpp
```diff
@@ -552,10 +552,10 @@ void TrialWaveFunction::copyFromBuffer(ParticleSet& P, WFBufferType& buf)
   assert(buf.size() == buf.current() + buf.current_scalar() * sizeof(double));
 }
 
-void TrialWaveFunction::evaluateRatios(VirtualParticleSet& VP, std::vector<RealType>& ratios)
+void TrialWaveFunction::evaluateRatios(VirtualParticleSet& VP, std::vector<ValueType>& ratios)
 {
   assert(VP.getTotalNum() == ratios.size());
-#if defined(QMC_COMPLEX)
+//#if defined(QMC_COMPLEX)
```
ye-luo: Remove!!!!!

src/QMCWaveFunctions/TrialWaveFunction.cpp
```diff
@@ -566,25 +566,26 @@ void TrialWaveFunction::evaluateRatios(VirtualParticleSet& VP, std::vector<RealT
...
-  for (int i = 0, ii = NL_TIMER; i < Z.size(); ++i, ii += TIMER_SKIP)
-  {
-    myTimers[ii]->start();
-    Z[i]->evaluateRatios(VP, t);
-    for (int j = 0; j < ratios.size(); ++j)
-      ratios[j] *= t[j];
-    myTimers[ii]->stop();
-  }
-#endif
+//#else
```
ye-luo: Remove!!!!!!

src/QMCWaveFunctions/TrialWaveFunction.cpp
chrislzhao: It's now deleted.

src/QMCDrivers/QMCCostFunction.cpp
chrislzhao: Removed.

src/QMCDrivers/QMCCostFunction.cpp
chrislzhao: Removed.

src/QMCDrivers/QMCCostFunction.cpp
chrislzhao: Removed.

src/QMCDrivers/WaveFunctionTester.cpp
chrislzhao: removed.

src/QMCHamiltonians/NonLocalECPComponent.cpp
chrislzhao: Renamed.

src/QMCHamiltonians/NonLocalECPotential.deriv.cpp
chrislzhao: Removed.

src/QMCHamiltonians/NonLocalECPotential.deriv.cpp
chrislzhao: Removed.

src/QMCHamiltonians/NonLocalECPotential.deriv.cpp
chrislzhao: changed.

src/QMCHamiltonians/QMCHamiltonian.h
chrislzhao: removed.

src/QMCWaveFunctions/DiffWaveFunctionComponent.h
chrislzhao: Documentation added.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
chrislzhao: Removed.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
chrislzhao: Removed.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
chrislzhao: Removed.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.h
chrislzhao: marked with override keyword.

src/QMCWaveFunctions/Jastrow/DiffOneBodyJastrowOrbital.h
chrislzhao: removed

src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
chrislzhao: removed

src/QMCWaveFunctions/TrialWaveFunction.cpp
chrislzhao: removed

src/QMCWaveFunctions/TrialWaveFunction.cpp
chrislzhao: removed

src/QMCHamiltonians/NonLocalECPotential.deriv.cpp
chrislzhao: I think both way works, and I have changed it to follow the above call sequence.

chrislzhao:
@ye-luo I have made changes.

src/QMCHamiltonians/NonLocalECPotential.deriv.cpp
ye-luo: The exact call sequence is required. It works for you probably because your wavefunction is simple.

src/QMCWaveFunctions/TrialWaveFunction.cpp
```diff
@@ -566,25 +565,10 @@ void TrialWaveFunction::evaluateRatios(VirtualParticleSet& VP, std::vector<RealT
       r[j] *= t[j];
     myTimers[ii]->stop();
   }
-  RealType pdiff;
   for (int j = 0; j < ratios.size(); ++j)
   {
-    RealType logr = evaluateLogAndPhase(r[j], pdiff);
-    ratios[j]     = std::exp(logr) * std::cos(pdiff);
-    //ratios[j]=std::abs(r)*std::cos(std::arg(r[j]));
+    ratios[j] = r[j];
```
ye-luo: Get ride of r[j]

src/QMCWaveFunctions/TrialWaveFunction.cpp
```diff
@@ -667,6 +651,34 @@ void TrialWaveFunction::evaluateDerivatives(ParticleSet& P,
   }
 }
 
+void TrialWaveFunction::evaluateDerivativesWF(ParticleSet& P,
+    const opt_variables_type& optvars,
+    std::vector<ValueType>& dlogpsi,
+    bool project)
```
ye-luo: You are the only one using evaluateDerivativesWF. Get ride of project.

src/QMCWaveFunctions/TrialWaveFunction.h
```diff
@@ -144,11 +144,11 @@ class TrialWaveFunction : public MPIObjectBase
 
   /** functions to handle particle-by-particle update */
   RealType ratio(ParticleSet& P, int iat);
-  ValueType full_ratio(ParticleSet& P, int iat);
+  ValueType evaluateFullRatio(ParticleSet& P, int iat);
```
ye-luo: Adding a documentation for this function and note the difference to ratio

src/QMCWaveFunctions/TrialWaveFunction.cpp
```diff
@@ -552,10 +552,9 @@ void TrialWaveFunction::copyFromBuffer(ParticleSet& P, WFBufferType& buf)
   assert(buf.size() == buf.current() + buf.current_scalar() * sizeof(double));
 }
 
-void TrialWaveFunction::evaluateRatios(VirtualParticleSet& VP, std::vector<RealType>& ratios)
+void TrialWaveFunction::evaluateRatios(VirtualParticleSet& VP, std::vector<ValueType>& ratios)
 {
   assert(VP.getTotalNum() == ratios.size());
-#if defined(QMC_COMPLEX)
   std::vector<ValueType> t(ratios.size()), r(ratios.size(), 1.0);
   ;
```
ye-luo: remove ";"

ye-luo:
Okay to test

src/QMCHamiltonians/NonLocalECPotential.deriv.cpp
chrislzhao: I have changed it and made sure to use the exact same call sequence.

src/QMCWaveFunctions/TrialWaveFunction.cpp
chrislzhao: Removed.

src/QMCWaveFunctions/TrialWaveFunction.cpp
chrislzhao: This argument has been removed.

src/QMCWaveFunctions/TrialWaveFunction.h
chrislzhao: I have added documentation for this function.

src/QMCWaveFunctions/TrialWaveFunction.cpp
chrislzhao: r[j] has been removed.

chrislzhao:
@ye-luo Hi Ye do you know what would cause the check failure? The code compiles fine on my side with QMC_COMPLEX=1 for both AoS and SoA, however, it throws out a linking error with QMC_COMPLEX=0 and AoS (SoA is fine). The error is:

/global/software/sl-7.x86_64/modules/langs/python/3.6/lib/libxml2.so: undefined reference to `ucnv_setToUCallBack_58'
/global/software/sl-7.x86_64/modules/langs/python/3.6/lib/libxml2.so: undefined reference to `ucnv_setFromUCallBack_58'
/global/software/sl-7.x86_64/modules/langs/python/3.6/lib/libxml2.so: undefined reference to `UCNV_TO_U_CALLBACK_STOP_58'
/global/software/sl-7.x86_64/modules/langs/python/3.6/lib/libxml2.so: undefined reference to `ucnv_open_58'
/global/software/sl-7.x86_64/modules/langs/python/3.6/lib/libxml2.so: undefined reference to `ucnv_convertEx_58'
/global/software/sl-7.x86_64/modules/langs/python/3.6/lib/libxml2.so: undefined reference to `ucnv_close_58'
/global/software/sl-7.x86_64/modules/langs/python/3.6/lib/libxml2.so: undefined reference to `UCNV_FROM_U_CALLBACK_STOP_58'
collect2: error: ld returned 1 exit status

ye-luo:
@chrislzhao I'm solving the conflict now.

ye-luo:
Test this please

chrislzhao:
@ye-luo Do you know why the checks failed?

PR #1797, opened 2019-08-14
Add batched interfaces in SPOSet.

ye-luo:
Start to populate multi walker batching interfaces.
The prefix `mw_` is a tag indicating these interfaces are intended to operate on multiple walkers in a batch.
evaluateValue, evaluateVGL are chosen to avoid the old problem of calling all the functions evaluate.


prckent:
"avoid the old problem of calling all the functions evaluate." : Yes! Heartily agree.

PR #1798, opened 2019-08-15
Adding scrape of fermi energy to pwscf_analyzer.py

mcbennet:
This will include the fermi energy of pwscf into the analyzer image.

### DIFF
--- a/nexus/lib/pwscf_analyzer.py
+++ b/nexus/lib/pwscf_analyzer.py
@@ -188,6 +188,25 @@ class PwscfAnalyzer(SimulationAnalyzer):
         #end try
 
         try:
+            fermi_energies = []
+            for l in lines:
+                if l.find('Fermi energy')!=-1:
+                    fermi_energies.append( eval( l.split('is')[1].split()[0] ) )
+                #end if
+            #end for
+            if len(fermi_energies)==0:
+                self.Ef = 0.0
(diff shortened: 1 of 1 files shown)


mcbennet:
All tests of nxs-test pass.

nexus/lib/pwscf_analyzer.py
```diff
@@ -187,6 +187,25 @@ def analyze(self):
             #end if
         #end try
 
+        try:
+            fermi_energies = []
+            for l in lines:
+                if l.find('Fermi energy')!=-1:
+                    fermi_energies.append( eval( l.split('is')[1].split()[0] ) )
```
jtkrogel: Prefer "float" over "eval".  I try to avoid "eval" as much as possible because it can be used to execute arbitrary Python code from text.

mcbennet:
Okay, I agree that float is better. I was mimicking what is used to scrape the total energy. In that case, 'eval' is currently used. Should I switch both to 'float' on the next commit?

jtkrogel:
Yes, please do.  Thanks for the catch; the PwscfAnalyzer code is some of the oldest in the Nexus repository.

mcbennet:
All tests of nxs-test pass.

grahamlopez:
okay to test

PR #1800, opened 2019-08-16
Fix infinite loop related to SO ECP handling

### LINKED ISSUE #1779: BGQ clang and oxygen clang unit test Hamiltonian is failing
unit test Hamiltonian is failing but not for a wrong value.
https://cdash.qmcpack.org/CDash/testDetails.php?test=5740567&build=70990
This started 2 days ago and #1771 is the only PR get in that day.

ye-luo:
Closes #1779 
libc++ treat subnormal number conversion as badbit when operating istream. The badbit but not eof status caused infinite loop and timeout in the nightlies (bgclang and oxygen clang). With this fix, the infinite loop is avoided and the code prints
```
failbit detected when reading type (type_info::name) d, value 2.21025771677336e-308
```
file Zn.ccECP-SO.xml needs a fix to put subnormal number as zero. I also don't like tabs in that file.

Also removed a compiler warning for not providing a return.

### DIFF
--- a/src/QMCHamiltonians/SOECPComponent.h
+++ b/src/QMCHamiltonians/SOECPComponent.h
@@ -47,12 +47,12 @@ private:
   std::vector<RadialPotentialType*> sopp_m;
 
 public:
-  SOECPComponent(){};
 
-  ///destructor
-  ~SOECPComponent(){};
-
-  SOECPComponent* makeClone(const ParticleSet& qp){};
+  SOECPComponent* makeClone(const ParticleSet& qp)
+  {
+    APP_ABORT("SOECPComponent::makeClone not yet implemented");
+    return nullptr;
+  };
 
   ///add a new Spin-Orbit component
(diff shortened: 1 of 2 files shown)


src/QMCHamiltonians/SOECPComponent.h
```diff
@@ -47,12 +47,12 @@ class SOECPComponent : public QMCTraits
...
 public:
-  SOECPComponent(){};
 
-  ///destructor
-  ~SOECPComponent(){};
-
-  SOECPComponent* makeClone(const ParticleSet& qp){};
+  SOECPComponent* makeClone(const ParticleSet& qp)
+  {
+    APP_ABORT("SOECPComponent::makeClone not yet implemented");
```
PDoakORNL: Instead of APP_ABORT I've started throwing std::runtime_error in these cases. One big plus is this gives us a path to better error handling when concurrency is non trivial and you can check for the the throw in a catch test program.

PR #1801, opened 2019-08-16
Fix a few xml memory leaks

### LINKED ISSUE #204: xml string handling leaks memory (xmlNodeListGetString)
I was checking some of my code for thread safety and trying to ensure the xml parsing was thread safe, when I found that there are many calls to xmlNodeListGetString that are not matched to a call to xmlFree.  My understanding is that xmlNodeListGetString allocates a new character array and that it is up to the user to deallocate it.  A quick grep shows that this could lead to memory leaks in many files, including:

src/QMCApp/QMCMain.cpp
src/OhmmsData/libxmldefs.h
src/OhmmsData/OhmmsParameter.h

I did not run these down exhaustively, so if this is an issue it is probably present in many other files too...

ye-luo:
closes #204
Fix multiple memory leaks related to xml.
Add XMLNodeString class to handle xmlNode content to std::string conversion
Add XMLAttrString class to handle one xmlNode attribute to std::string conversion

So far, I used valgrind on a few unit tests and didn't find any leak due to xml. However valgrind cannot be a good way to monitor leaks because I saw leaks from MPI and OpenMP runtime libraries. They can be false alarm but indicating that valgrind is not suitable for tracking purpose.

Q: how to write the following code in one line?
```
const XMLParsingString node_string(cur);
std::istringstream stream(node_string);
```
I tried
```
// error, "std::istringstream(XMLParsingString)" was seen as non-class type
std::istringstream stream(const XMLParsingString(cur));
// working, ugly double parentheses
std::istringstream stream((const XMLParsingString(cur)));
// working, a bit C style
std::istringstream stream((const XMLParsingString)(cur));
```
What is a good C++ way to make this one line?


ye-luo:
Fixing xmlGetProp

prckent:
The problem with OpenMP and MPI leakage is exactly why serial builds are useful and desired. 

@correaa How is the MPI wrapper's serial mode? I believe AFQMC doesn't build without MPI at the moment, but this will always complicate use of other tools, so can't be supported long term.

PDoakORNL:
This is an example of
https://www.fluentcpp.com/2018/01/30/most-vexing-parse/

write
```C++
std::istringstream stream(XMLParsingString{cur});
```

prckent:
@PDoakORNL A good read. Even more reason to switch on all the clang warnings.

correaa:
> @correaa How is the MPI wrapper's serial mode? I believe AFQMC doesn't build without MPI at the moment, but this will always complicate use of other tools, so can't be supported long term.

I wrote a full implementation of the message passing part of MPI in https://gitlab.com/correaa/boost-mpi3/blob/master/fake/mpi.h .  The library has its own `mpirun` and compiler wrapper: https://gitlab.com/correaa/boost-mpi3/tree/master/fake
It was a lot of work with little feedback, so I move on to other things. 

If more features are necessary from the fake-MPI it can be requested here: https://gitlab.com/correaa/boost-mpi3/issues

What does it mean that AFQMC doesn't build without MPI? (@mmorale3) Maintaining to versions of the code is too much work (even if `ifdef`s are used for that). The whole point of fake-MPI is that the same MPI version of the code could run 1) in 1 single MPI process and 2) without an implementation of MPI available. Requiring that an MPI code compile without MPI is too much to ask for that code in my opinion. (it is like asking that a C++ program would compile without a C++ compiler).
As a matter of fact, AFQMC can be taken out of the picture here because if it is well written (using standard features of MPI) and it can run in a single MPI process, the program should be able to compile with fake-MPI. In other words, if we are taking serial (fake MPI) builds seriously then fake-MPI should be taken as its own MPI implementation, at the same level of openmpi or mpich and then report missing features of fake-MPI on the process. Can fake-MPI be added as compilation option in QMCPack? (@markdewing)

prckent:
Thanks Alfredo. Hopefully the AFQMC folks can comment here on the current situation. I recall that there was some SHM usage or similar that was tricky to fake.

I would like the "no MPI" configuration to take place behind the scenes, so if e.g. we need to switch include paths that should happen in our CMakeLists.txt

ye-luo:
Now the code should no more leak memory when using xmlNodeListGetString and xmlGetProp.

src/QMCWaveFunctions/MolecularOrbitals/AtomicBasisBuilder.h
```diff
@@ -239,7 +239,7 @@ typename AtomicBasisBuilder<RFB>::COT* AtomicBasisBuilder<RFB>::createAOSet(xmlN
     if (cname1 == "basisGroup")
     {
       radGroup.push_back(cur1);
-      int l = atoi((const char*)(xmlGetProp(cur1, (const xmlChar*)"l")));
+      const int l = std::stoi(XMLAttrString{cur1, "l"});
```
PDoakORNL: I want to get to using the c++ coversion functions but std::stoi will throw an exceptions if it can't parse the input. I don't think this is in a try catch scope.

src/QMCWaveFunctions/PlaneWave/PWOrbitalBuilder.cpp
```diff
@@ -69,9 +69,8 @@ bool PWOrbitalBuilder::put(xmlNodePtr cur)
     std::string cname((const char*)(cur->name));
     if (cname == "basisset")
     {
-      const xmlChar* aptr = xmlGetProp(cur, (const xmlChar*)"ecut");
-      if (aptr != NULL)
-        myParam->Ecut = atof((const char*)aptr);
+      const XMLAttrString a(cur, "ecut");
+      if (!a.empty()) myParam->Ecut = std::stod(a);
```
PDoakORNL: like std::stoi

src/QMCWaveFunctions/lcao/AOBasisBuilder.h
```diff
@@ -255,7 +255,7 @@ COT* AOBasisBuilder<COT>::createAOSet(xmlNodePtr cur)
     if (cname1 == "basisGroup")
     {
       radGroup.push_back(cur1);
-      int l = atoi((const char*)(xmlGetProp(cur1, (const xmlChar*)"l")));
+      const int l = std::stoi(XMLAttrString{cur1, "l"});
```
PDoakORNL: another std::stoi

src/QMCWaveFunctions/MolecularOrbitals/AtomicBasisBuilder.h
ye-luo: At the moment, I think it is no worse than the C functions
When I was changing atoi to stoi and atof to stod, I did ask myself what happens if the conversion fails. I think we do need to catch the exception which is good for the code. I thought about using istringstream and check failbit but the code gets lengthy.
Could you explain a bit more what you mean by  "I don't think this is in a try catch scope"?

PR #1805, opened 2019-08-16
Update oxygen nightlies scripts

prckent:
Spack python updates and PYTHONPATH fix

To free up time on oxygen the combination of builds tested will be updated in a later PR.


markdewing:
This PR also changes some Intel compiler and MKL versions, but the changes aren't described in the PR message.  In particular it appears to downgrade the MKL from 2019 to 2018.

prckent:
AFAIK we don't have a sensitivity to MKL versions. There are builds using MKL 2019 (the Intel2019 builds) and builds using MKL 2018.  If there are specific combinations that we need to test let me know since we are trying to cut down.

PR #1806, opened 2019-08-16
Nexus: add updates for August user meeting

jtkrogel:
User meeting is on the Structure class.  Updates fix a couple of rusty functions, complete file writing abilities, and add a unit test as a demo for others.


markdewing:
okay to test

PR #1810, opened 2019-08-19
Unified Driver: Simplifying setup and refactoring Input out of QMCDriverNew and VMCBatched 

PDoakORNL:
- Refactored much of the input reading out of QMCDriverNew and VMCBatch. Creating:
  - QMCDriverInput
  - VMCDriverInput
- Cut local_walker starting setup down to one location and one testable behavior.
- MCPopulation is beginning to collect its content.
- Fixed bug passing update mode to VMCFactoryNew
- Collected test_drivers XML (for new tests into common header to allow easier updating of valid driver section XML over time)
- Added skeleton of Crowd class (More action in next PR) 

I apologize for the epic DIffs, @ye-luo was correct so much was removed or moved from QMCDriver.(h/cpp) and VMC.(h/cpp) it would have been better to start from empty calls.


prckent:
@PDoakORNL What is the lisp used for? How would someone else benefit from it? (i.e. Should it be in the main repo or not?)

PDoakORNL:
It generates setters and getters from selected member variable declarations.

src/QMCDrivers/VMC/VMCFactoryNew.cpp
```diff
@@ -2,47 +2,50 @@
...
+  qmcdriver_input.readXML(input_node_);
+  VMCDriverInput vmcdriver_input;
+  vmcdriver_input.readXML(input_node_);
   QMCDriverInterface* qmc = nullptr;
-  if (VMCMode == 0 || VMCMode == 1) //(0,0,0) (0,0,1)
+
+  if (vmc_mode_ == 0 || vmc_mode_ == 1) //(0,0,0) (0,0,1)
   {
-    qmc = new VMCBatched(pop, psi, h, ppool, comm);
+    qmc = new VMCBatched(qmcdriver_input, vmcdriver_input, pop, psi, h, ppool, comm);
```
ye-luo: I feel better having vmcdriver_input derived from qmcdriver_input following the nature that vmcdriver is derived from qmcdriver.

src/QMCDrivers/VMC/VMCFactoryNew.h
```diff
@@ -27,11 +29,15 @@ class MCPopulation;
 class VMCFactoryNew
 {
 private:
-  int VMCMode;
-  xmlNodePtr myNode;
+  const int vmc_mode_;
+  // const ?
+  xmlNodePtr input_node_;
```
ye-luo: Yes. I think a QMC section input never gets touched.

src/QMCDrivers/tests/ValidQMCInputSections.h
```diff
@@ -0,0 +1,79 @@
...
+    <parameter name="timestep">             1.0 </parameter>
+    <parameter name="usedrift">              no </parameter>
+  </qmc>
+)",
+    R"(
+  <qmc method="vmc_batch" move="pbyp">
+    <parameter name="crowds">                 8 </parameter>
+    <estimator name="LocalEnergy" hdf5="no" />
+    <parameter name="walkers">                32 </parameter>
+    <parameter name="stepsbetweensamples">    1 </parameter>
```
ye-luo: I'd like to retire this input parameter.

src/QMCApp/QMCDriverFactory.cpp
```diff
@@ -247,7 +252,7 @@ std::unique_ptr<QMCDriverInterface> QMCDriverFactory::createQMCDriver(xmlNodePtr
   if (das.new_run_type == QMCRunType::VMC || das.new_run_type == QMCRunType::CSVMC)
   {
     //VMCFactory fac(curQmcModeBits[UPDATE_MODE],cur);
-    VMCFactory fac(das.what_to_do.to_ulong(), cur);
+    VMCFactory fac(das.what_to_do[UPDATE_MODE], cur);
```
markdewing: Is passing just the UPDATE_MODE bit the same as passing the entire bitset?   The  VMCFactory code appears to use the GPU_MODE bit for selecting the CUDA version.

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -2,68 +2,61 @@
...
+  // qmc_driver_mode.set(QMC_WARMUP, 0);
+}
 
-  prevSteps               = nSteps;
-  prevStepsBetweenSamples = nStepsBetweenSamples;
+VMCBatched::IndexType VMCBatched::calc_default_local_walkers()
+{
+  IndexType rw = vmcdriver_input_.get_requested_walkers_per_rank();
+  if (rw < num_crowds_)
+    rw = num_crowds_;
```
ye-luo: Changing user expectation is not good. If I have memory for only 6 walkers.
I believe num_crowds_ is default to the number of threads 7 for example.
The code bumps the number of walkers to 7 and the code cannot run.

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -2,68 +2,61 @@
...
+  walkers_per_crowd_      = (rw % num_crowds_) ? rw / num_crowds_ + 1 : rw / num_crowds_;
```
ye-luo: walkers_per_crowd should be a vector of a size equal to the number of crowds.
What you defined here seems to be the maximum number of walkers per crowd.

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -2,68 +2,61 @@
...
+  if (rw != vmcdriver_input_.get_requested_walkers_per_rank())
+    app_warning() << "VMCBatched driver has adjusted walkers per rank to: " << local_walkers << '\n';
+
+  if (vmcdriver_input_.get_samples() >= 0 || vmcdriver_input_.get_samples_per_thread() >= 0 ||
+      vmcdriver_input_.get_steps_between_samples() >= 0)
+    app_warning() << "VMCBatched currently ignores samples and samplesperthread\n";
+
+  // TODO: Simplify samples, samples per thread etc in the unified driver
+  // see login in original VMC.cpp
+  return local_walkers;
```
ye-luo: l thnk the current implementation calc_default_local_walkers is OK for experiment purpose but needs to be generalized.

src/QMCDrivers/VMC/VMCDriverInput.cpp
```diff
@@ -0,0 +1,35 @@
...
+// File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Laboratory
+//
+// File created by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+#include "QMCDrivers/VMC/VMCDriverInput.h"
+
+namespace qmcplusplus
+{
+VMCDriverInput::VMCDriverInput(int walkers_per_rank, const std::string& use_drift)
```
ye-luo: Why do we need walkers_per_rank and walkers_per_rank? I think read from XML is sufficient. I found them used for testing. If testing is the only purpose, this needs to be changed.

src/QMCApp/QMCMain.cpp
```diff
@@ -338,7 +338,7 @@ bool QMCMain::executeQMCSection(xmlNodePtr cur, bool noloop)
   return success;
 }
 
-/** validate the main document
+/** validate the main document and (read the walker sets !)
```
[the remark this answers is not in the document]
PDoakORNL: The HDF5 walker sets

src/QMCDrivers/VMC/VMCDriverInput.cpp
```diff
@@ -0,0 +1,35 @@
...
+    : requested_walkers_per_rank_(walkers_per_rank), use_drift_(use_drift)
+{}
+
+void VMCDriverInput::readXML(xmlNodePtr& node)
```
ye-luo: xmlNodePtr is a pointer. no need of using reference.

src/QMCApp/QMCDriverFactory.cpp
```diff
@@ -256,9 +261,10 @@ std::unique_ptr<QMCDriverInterface> QMCDriverFactory::createQMCDriver(xmlNodePtr
   }
   else if (das.new_run_type == QMCRunType::VMC_BATCH)
   {
-    VMCFactoryNew fac(das.what_to_do.to_ulong(), cur);
+    VMCFactoryNew fac(cur, das.what_to_do[UPDATE_MODE], qmc_common.qmc_counter);
```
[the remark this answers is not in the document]
PDoakORNL: I agree this way it was visible here instead of accessed inside the factory or worse the driver itself.

src/QMCDrivers/tests/ValidQMCInputSections.h
prckent: Agreed. Worth noting, but perhaps this is not the place to fix. Additionally it is a duplication of the already badly named substeps.

src/QMCDrivers/VMC/VMCBatched.cpp
prckent: @PDoakORNL  I think we need the ability to run empty crowds in the code and correctly do no work.  This will certainly be helpful for testing. e.g. running the code with one walker and a variable number of crowds will give a consistent result. 

Input logic can deal with reminders and possible enforcement of sensible settings.

src/QMCDrivers/VMC/VMCBatched.cpp
PDoakORNL: Yes its the number of walkers each crowd has.

src/QMCDrivers/VMC/VMCBatched.cpp
PDoakORNL: I'm definitely ok with abandoning the current behavior and having some crowds under capacity.  I think that's less confusing as well.

src/QMCDrivers/VMC/VMCDriverInput.cpp
PDoakORNL: walkers_per_rank and walkers_per_crowd?
Ideally you should be able to construct these input objects without XML although at this point the Constructors are not very systematic.

src/QMCDrivers/VMC/VMCBatched.cpp
PDoakORNL: As far as I can tell only VMC behaved like this.  Which is why I stuck this out in the derived class. If we can remove the "weird" behavior and get rid of it that would be good.

src/QMCApp/QMCDriverFactory.cpp
PDoakORNL: This will be addressed in PR's to follow

src/QMCDrivers/VMC/VMCFactoryNew.cpp
```diff
@@ -2,47 +2,50 @@
...
+                                          ParticleSetPool& ptclpool,
+                                          HamiltonianPool& hpool,
+                                          WaveFunctionPool& ppool,
+                                          Communicate* comm)
 {
-  int np = omp_get_max_threads();
-  //(SPACEWARP_MODE,MULTIPE_MODE,UPDATE_MODE)
+  QMCDriverInput qmcdriver_input(qmc_counter_);
+  qmcdriver_input.readXML(input_node_);
+  VMCDriverInput vmcdriver_input;
```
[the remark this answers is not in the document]
PDoakORNL: Fixed

src/QMCDrivers/VMC/VMCFactoryNew.cpp
PDoakORNL: We can continue to discuss this is just a first factoring of the Input out of the driver.

PDoakORNL:
test this please

src/QMCDrivers/VMC/VMCDriverInput.cpp
PDoakORNL: fixed this.

PR #1815, opened 2019-08-20
Fix Zn.ccECP-SO.xml

rcclay:
I set the floor for absolute values to be 1e-20, and anything smaller is zeroed out for the spin-orbit channels.  Also got rid of tabs and adopted the same 3 column format asfor the nonlocal channels.


rcclay:
Uggh.  Forgot to sync the new xml to the unit test directories.  Standby.

prckent:
On the positive side this teaches us that the unit tests are quite sensitive.

rcclay:
So I just built & tested the CPU version on Rhea, and it looks like it passes all the tests. :|

ye-luo:
Test this please.

PR #1818, opened 2019-08-22
adding emacs temp files to .gitignore

PDoakORNL:
I'm tired of `git status` scrolling off the top of the screen.
These are common elements in .gitignore of other projects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,14 @@
+# emacs temp files
+*~
+*#
+
+# python bytecode 
 *.pyc
+
+# excluded manual files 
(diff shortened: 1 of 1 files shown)


prckent:
I added the qmcpack_manual.pdf back to the ignores. We don't want that committed.

PDoakORNL:
Did not remove that one purpose.  @grahamlopez pointed out to me that while many projects do this you can have a user level ~/.gitignore that will do what I want.  you just need to do this:
```shell
 git config --global core.excludesfile ~/.gitignore
```

Issue #1769, opened 2019-07-29
Forces:  Include ion-ion force contribution to ACForce


rcclay:
Currently, ACForce does not appear to accumulate the ion-ion contribution to the total force on the ions.  I feel like this should be included by default.  Verification that this can be computed is needed for the PBC case, and implementation is definitely needed for the openBC case.

jtkrogel:
@rcclay We are targeting molecules in open boundary conditions first, so ion-ion force in OBC will be needed.  It seems like this should be pretty simple to add. Is this something you can do in the near future?

rcclay:
@jtkrogel Sure. It's the easiest force to compute, just need to sit down, implement, and unit test it.

tiihonej:
I can take a look at the ion-ion force in OBC, unless you already have it ready, @rcclay?

rcclay:
Feel free to take a stab at it.  It'll go in src/QMCHamiltonians/CoulombPotential.h .  If you check the constructor, there's a "if(!is_active)" block, indicating whether QMCPACK expects those particles to be moving or not.  If they don't, it currently computes the like-species energy in the constructor and returns it every time the estimator is queried so you don't have to recompute it at every electron step.  I would do forces the same way.  You can consult src/QMCHamiltonians/CoulombPBCAA.cpp to see how this was done in PBC's.  OpenBC's will be easier since you don't have to worry about ewald breakups.  

1.)  Implement bare like species forces.  Look at evaluateAA() in CoulombPotential to get an idea of how the charge and distance tables are accessed.  You'll be messing with \mathbf{dr} now instead of just scalar quantities for forces.  You can park it in its own evaluateForces routine or something for compartmentalization. 
2.)  ForceBase class provides a prefix variable (the name of the force quantity in scalar.dat), an addObservablesF() function for registering these to scalar.dat, and a "forces" vector for storing the forces.  You'll need to call the addObservablesF() and set the aforementioned variables, ideally with some logic guarding these operations (if ComputeForces==True, then do stuff...  otherwise don't).  See CoulombPBCAA  
3.)  To work with the ACForce estimator, you'll need to add an evaluateWithIonDerivs() function.  You'll need to set "hf_terms" to whatever forces you computed, and set "pulay_terms" to zero.  Likewise, see CoulombPBCAA.

Feel free to ask questions!

rcclay:
Closed by #1820

PR #1820, opened 2019-08-22
Add IonIon force observable to CoulombPotential in OBC

tiihonej:
In response to issue #1769 from @rcclay, I've added code to compute IonIon forces due to coulomb potential in OBC, when "forces=yes" attribute is present.  The value is accumulated with that of ACForce estimator.  I've tested that the code works for more than two ions, and checked the validity by hand.

### DIFF
--- a/src/QMCHamiltonians/CoulombPotential.h
+++ b/src/QMCHamiltonians/CoulombPotential.h
@@ -19,6 +19,7 @@
 #include "Particle/ParticleSet.h"
 #include "Particle/WalkerSetRef.h"
 #include "Particle/DistanceTableData.h"
+#include "QMCHamiltonians/ForceBase.h"
 #include "QMCHamiltonians/QMCHamiltonianBase.h"
 #include <numeric>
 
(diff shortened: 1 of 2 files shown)


jtkrogel:
Is there a reason to keep the forces="yes" requirement?  I would prefer for ion-ion to be included automatically anytime a force estimator is used.

rcclay:
@jtkrogel In the long run, no.  In the short run, this allows the force components to be printed separately to scalar.dat for checking purposes, and also handles some potential one-off initializations that might be needed for forces.  This will definitely need to be streamlined.

jtkrogel:
I think the total force and each sub-component should be written automatically every time.  I guess the conditional initialization issue requires more coordination between the force estimator and the hamiltonian components?

rcclay:
>  I guess the conditional initialization issue requires more coordination between the force estimator and the hamiltonian components?

That's the tricky bit.

tiihonej:
I added a unit test. Not sure if it's very classy, but it seems to do the job. I take it we are still sticking with the 'forces=yes' practice for now. Enabling/disabling the feature should be easy enough using a flag in the initialization, but the logic behind this flag should probably be re-designed in the future. I would vote for making ion-ion force the default, provided that there is a force estimator present, otherwise not.

PR #1821, opened 2019-08-22
Restore manual section on adding a wavefunction

markdewing:
It was accidentally removed while integrating the edits


prckent:
I'll merge this since the CI is down and this only touches the manual, which is not checked by the CI in any case...

PR #1852, opened 2019-08-29
Fix band read for hybrids

mcbennet:
In this PR, a fix is made that avoids band read failure when a hybrid
functional is used in the calculation. During a hybrid calculation,
band data is printed multiple times in the output (once for each scf
convergence). The changes here make sure that nfound, index, and the
band objects are re-initialized each time '- SPIN UP -' is seen in
the output.

### DIFF
--- a/nexus/lib/pwscf_analyzer.py
+++ b/nexus/lib/pwscf_analyzer.py
@@ -236,14 +236,17 @@ class PwscfAnalyzer(SimulationAnalyzer):
             read_rel      = False
             for i in xrange(len(lines)):
                 l = lines[i]
-                if '- SPIN UP -' in l:
-                    up_spin   = True
+                if 'End of self-consistent calculation' in l:
                     # Initialize each time in case a hybrid functional was used
                     nfound = 0
                     index = -1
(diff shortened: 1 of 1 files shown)


mcbennet:
100% of nxs-test tests pass.

markdewing:
okay to test

nexus/lib/pwscf_analyzer.py
```diff
@@ -241,6 +236,15 @@ def analyze(self):
             read_rel      = False
             for i in xrange(len(lines)):
                 l = lines[i]
+                if 'End of self-consistent calculation' in l:
```
jtkrogel: What about the non-self consistent case?  Shouldn't we also keep the original empty initialization of bands, etc?

mcbennet: Good catch. I have restored the original inits in the most recent commit.

mcbennet: Additionally, I added a condition to prevent the inits from being performed redundantly for the first iteration of scf convergences.

PR #1855, opened 2019-08-29
Unified driver VMC block setup

PDoakORNL:
Another push on the way to VMCBatched unified driver. This PR mostly involves effort to refactor the minimum from MCWalkerConfiguration, QMCUpdateBase, VMCUpdatePbyP, and CloneManager on the way to eliminating them from the code. These classes provide code to the current drivers through fairly complicated means.  The new classes aim to reduce the reliance on inheritance as a method to reuse code and mix name spaces and scopes while providing the useful functionality in a more understandable format.

- MCPopulation creates and holds the walkers all the walker scope "clones" CloneManager code has migrated to here.  It is owned by the driver. It has the possibilty of being released to the next driver.
- VMCBatched takes the functions from QMCUpdateBase and VMCUpdatePbyP.
 -MoveContext takes code and data from MCWalkerConfiguration most of which operated at the walker level, right now it is owned by the driver but it should be composed into the Crowd in a following PR.
- the Crowd holds references to walkers and clones in MCPopulation for the thread scope its owned by the driver.

I have also made an effort to reduce constructor arguments of ParticleSet type. I introduced a new type for this that encapsulates the indexes for the change of particle species in particle related lists. In many cases this was all that constructors taking a ParticleSet needed.  In general it would be nice to have constructors that expressed objects actually construction dependence. Most of these classes take ParticleSet's later as function arguments so maintaining all or some of a particle set as state is undesirable.


PDoakORNL:
Consider waiting until it actually passes CI. EstimatorManagerCrowd is probably best ignored for now.

ye-luo:
I feel MCPopulation needs to be owned by the QMCMain and drivers capture them by reference. This was the way you did. What is difference compared with moving ownership in and out of the driver?

ye-luo:
I got confused by the last paragraph
> I have also made an effort to reduce constructor arguments of ParticleSet type.
Did you change any constructor of ParticleSet? 
> I introduced a new type for this that encapsulates the indexes for the change of particle species in particle related lists.
What is the new type?
> In many cases this was all that constructors taking a ParticleSet needed. In general it would be nice to have constructors that expressed objects actually construction dependence. Most of these classes take ParticleSet's later as function arguments so maintaining all or some of a particle set as state is undesirable.
I have thought about that. For example, WaveFunctionComponent can hold a const reference of ParticleSet instead of passing it as a function argument. Now with batching. It is less clear who to do this. Let us finish the initial batching code and then find a solution. It will be easier when when we trim more legacy code.

PDoakORNL:
I added a number  of constructors but then had to revert because I collided with RPAJastrow. I haven't given up on using them in new constructors but I'm sure you already knew it wasn't going to be so easy with WFC constructors. 
The type is
```C++
using PtclGrpIndexes = std::vector<std::pair<int,int>>
```

src/QMCDrivers/VMC/VMCBatched.h
```diff
@@ -24,17 +25,46 @@ namespace qmcplusplus
  */
 class VMCBatched : public QMCDriverNew
 {
+public:
+  using PosType = QMCTraits::PosType;
+  /** To avoid 10's of arguments to runVMCStep
+   */
+  struct StateForThread
```
markdewing: This seems like it has everything except for thread-specific state?   It needs a better name.   QMCInput?  QMCSimulation?

src/QMCDrivers/MoveContext.h
```diff
@@ -0,0 +1,97 @@
...
+{
+
+class MCPopulation;
+class DistanceTableData;
+/** Thread local context for moving walkers
+ *
+ *  created on the stack once per driver per crowd
+ *  refactored out of ParticleSet, only the minimum
+ */
+class MoveContext
```
markdewing: I read 'Move' as a verb (even in the class name) and I expect this to have to do with moving data in the program (vs. moving walker positions in space).
ParticleMoveContext?  CrowdParticleMoveContext?  DanceContext? (when a bunch of particles move, we could call it a dance? :smiley: That's probably getting too far outside standard domain language, though.)

PDoakORNL: That hadn't occurred to me but now that's how I read it too. Will rename.

PDoakORNL:
@ye-luo 
> I feel MCPopulation needs to be owned by the QMCMain and drivers capture them by reference. This was the way you did. What is difference compared with moving ownership in and out of the driver?

As of now the there is the issue of MCPopulation being constructed in the VMCBatched branch of QMCDriverFactory. I'd like the factories to either be stateless or short lived so I don't want to "store" any new objects in them. But I can push the construction up into QMCMain where I think you are right that it belongs. Then pass it into QMCDriverFactory.

ye-luo:
Good that we agree on the final ownership of MCPopulation and we can fix it later.
I saw your intention to tear apart the ParticleSet but I think most pieces of it has some internal logic to be kept together although there are many pieces not really belonging to it and we will re-factor them out.
For example, P.G, P.L are the gradients and laplacian of WF.
P.Collectables/Observables should be owned by the EstimatorManager or something other responsible class.

PDoakORNL:
Depending on you mpi build and compiler `omp_get_max_threads() ` get report a variety of thread counts.  For testing I now force openmp to give 8 threads via an RAII override object.  This is only used in testing.

Issue #1853, opened 2019-08-29
short-li2_sto-sj_dmc crashing


ye-luo:
https://cdash.qmcpack.org/CDash/testDetails.php?test=6012652&build=76238
Caused by #1820 not improved by #1841 
@tiihonej could you have a look?

tiihonej:
I'll look into to it. There's another fix coming up, too, which is not affecting any existing tests. I'll extend the existing static force test a bit to catch that and post them asap.

ye-luo:
Good now. https://cdash.qmcpack.org/CDash/testDetails.php?test=6025245&build=76469

PR #1857, opened 2019-08-29
Fix errors in CoulombPotential cloning after forces implementation

tiihonej:
Fixing cloning and constructor problems that were introduced in earlier commits in special cases involving OMP simulations (#1853 ) and AA-Hamiltonian of multiple active species. Extended one of the unit tests in this respect.


markdewing:
okay to test

PR #1861, opened 2019-09-02
Refactor hdf hyperslab

ye-luo:
I wrote my understanding of src/io in src/io/README.md.
This PR makes a few changes in hdf_archive and its related template classes.

1. Since this PR, scalars are written as hdf5 scalars instead 1d array of size 1. hdf5 library can still read/write data on old file correctly but h5diff complains about incompatible objects although not showing an error. That is why all the calculation runs and diff with gold h5 files doesn't show error in the converter tests. Using scalars as multidimensional data with rank 0 (HDF SCALAR), simplify our logic and hence code.

2. @lshulen readHyperslab has been renamed as readSlabSelection. readSlabReshaped and writeSlabReshaped are added to support read/write container data from/to a multidimensional dataset(hyperslab) with arbitrary shape on file. See examples in the tests.  ` src/io/tests/test_hdf_manual_reorder.cpp` has been split into `test_hdf_hyperslab.cpp` and `test_hdf_reshape.cpp`

3. Better support for datatype with value semantics when used by containers. std::complex TinyVector and Tensor are supported. `Matrix<TinyVector<std::complex<T>>>` can be used without adding composite template specialization.

4. @mmorale3 @fdmalone This PR change the template specialization for Multi containers. Please have a look. Template specialization of multi in hdf_hyperslab.h has been moved to hdf_multi.h.
container_proxy_multi.h is renamed as container_traits_multi.h and gets simplified.
So far no problem with test_afqmc_walkers.

5. Reduce hdf_hyperslab constructors to 1 and check std::array  for global dimension, local dimension and offset values since hdf5 requires unsigned.

6. Change int to size_t in a few places to stop warning from GCC. In C++, an implicit conversion from signed to unsigned value in initializer list caused warning.


ye-luo:
@mmorale3 @fdmalone 
`AFQMC/Hamiltonians/HamiltonianFactory.cpp` and `AFQMC/Hamiltonians/KPFactorizedHamiltonian.cpp`
can not read Energies correctly.
In `HamiltonianFactory.cpp`
```C++
    ValueType NuclearCoulombEnergy(0);
    ValueType FrozenCoreEnergy(0);

    if(head) {
      std::vector<ValueType> Rdata(2);
      if(!dump.readEntry(Rdata,"Energies")) {
        app_error()<<" Error in HamiltonianFactory::fromHDF5(): Problems reading  dataset. \n";
        APP_ABORT(" ");
      }
      if(Rdata.size()>0)
        NuclearCoulombEnergy = Rdata[0];
      if(Rdata.size()>1)
        FrozenCoreEnergy = Rdata[1];
    }
```

the h5 file used for unit test contains
```
h5ls -d ham_chol_sc.h5/Hamiltonian/Energies
Energies                 Dataset {2}
    Data:
        (0) -15.3649862050191, 0
```
In a real build,
`std::vector<ValueType>` is `std::vector<float>` and expect 1D data.
in a complex build,
`std::vector<ValueType>` is `std::vector<std::complex<float>>` and expect 2D data and caught by the added error trap in this PR.
Should the energy always real? Are they used as complex later just for convenience?
In this case, we can define Rdata as `std::vector<RealType>`. I tried this change and all the tests pass. I'd like your confirmation if this is a correct fix before push the code change.

lshulen:
A design question.  Is it necessary to be so restrictive in the type for the shape argument in methods like writeSlabReshaped (it has to be a std::array of unsigned int type pod)?  If so, why not just lock it down completely to hsize_t or unsigned int and get rid of a lot of template complexity?  I also thought an alternative would have been to provide a public typedef, but since the dimension is part of the type this seems impractical at first....

ye-luo:
> A design question. Is it necessary to be so restrictive in the type for the shape argument in methods like writeSlabReshaped (it has to be a std::array of unsigned int type pod)? If so, why not just lock it down completely to hsize_t or unsigned int and get rid of a lot of template complexity? I also thought an alternative would have been to provide a public typedef, but since the dimension is part of the type this seems impractical at first....

I initially allow signed integer types and added error traps if sizes are negative. Then I thought why not asking developers to be precise and errors get caught at compile time? Then made a choice of requiring unsigned types. This adds a bit burden to developers but I think it may be worth it. Note that directly assigning int to size_t doesn't cause warning. Warnings only occur when using initializer list. I tried not to expose hdf5 native type hsize_t (unsigned long int, or something else) directly to people using hdf_archive. A template integer type with unsigned restrictions gives flexibility (any integer) with restrictions (unsigned). I think it is better than requiring unsigned int directly. If people find using unsigned integer is problematic, I can allow it. The code can be retrieved from history.

Regarding std::array for shape parameters. To construct a hyperslab, knowing the rank of the multidimensional array is necessary and I don't see a use case which needs runtime determination. So std::array defines the rank at compile time.

Because of the rank, writeSlabReshaped needs to be a template function and the template on integer type should not cost more.

markdewing:
I wish the int->size_t change and the TinyVector -> std::array change had been separate PR's.  They would have been trivial to review, and made this PR smaller.

ye-luo:
@markdewing I believe if I had known the need of changing size_t and std::array, I could probably do the change without breaking the code before the API change. Unfortunately this didn't happen. Only when I refactor the code, I realize the need of change in all the places. Now I cannot easily go back.

lshulen:
I'm sure you will figure out how you want this done.  I will point out that it seems a runtime check is still warranted.  In the case of reading: "Are the dimensions compatible with what is in the hdf file?"  This is only known at runtime.  In the case of writing: "Does data contain enough elements in order to write out what I am specifying?"  Similar things apply to hyperslab type operations.

I don't think requiring the client programmer to use an unsigned type really solves the problem.  However, I don't think it's my place hold things up over my reservations.  I will not object to this PR going in.

src/io/hdf_hyperslab.h
```diff
@@ -12,202 +12,164 @@
...
 
-template<typename CT, unsigned MAXDIM>
-struct h5data_proxy<hyperslab_proxy<CT, MAXDIM>>
-{
-  hyperslab_proxy<CT, MAXDIM>& ref_;
-  h5data_proxy(hyperslab_proxy<CT, MAXDIM>& a) : ref_(a) {}
-  inline bool read(hid_t grp, const std::string& aname, hid_t xfer_plist = H5P_DEFAULT)
+  /** checks if file_space, elected_space and offset are self-consistent
+   */
+  inline void checkUserRankSizes()
```
ye-luo: @lshulen runtime checks of filespace, selected space and offset are invoked in both reading and writing. You think more checks are needed?

src/io/hdf_hyperslab.h
```diff
@@ -12,202 +12,164 @@
...
-      if (!get_space(grp, aname, rank, ref_.slab_dims.data(), true))
-      {
-        std::cerr << " Disabled hyperslab resize with boost::multi::array_ref<gpu_ptr>.\n";
-        return false;
-      }
-      return h5d_read(grp, aname, ref_.data(), xfer_plist);
-    }
+    std::vector<hsize_t> sizes_file;
+    getDataShape<typename hyperslab_proxy<CT, RANK>::element_type>(grp, aname, sizes_file);
+    ref_.adaptShape(sizes_file);
```
ye-luo: @lshulen The above two steps first query file shape and then adjust the container if the size doesn't match. If the container doesn't know how to resize properly. The code throw runtime error.

src/io/hdf_hyperslab.h
```diff
@@ -12,202 +12,164 @@
...
+                    ref_.slab_offset.data(), hyperslab_proxy<CT, RANK>::SpaceType::get_address(container_traits<CT>::getElementPtr(ref_.ref_)),
+                    xfer_plist);
   }
+
   inline bool write(hid_t grp, const std::string& aname, hid_t xfer_plist = H5P_DEFAULT)
   {
-    std::cerr << " Disabled hyperslab write with boost::multi::array_ref<gpu_ptr>.\n";
-    return false;
+    ref_.checkUserRankSizes();
+    return h5d_write(grp, aname.c_str(), ref_.file_space.rank, ref_.file_space.dims, ref_.selected_space.dims,
```
ye-luo: @lshulen you are right. I need to check size before writing to see if sufficient size is provided in the container.

src/io/hdf_hyperslab.h
lshulen: My point is not that more runtime checks were needed.  It was simply that enforcing that the indices be > 0 by using an unsigned type doesn't really get you out of performing runtime checks if you want to make sure things are OK.  

Then, my logic was that if runtime checks are needed, than having one compile time check (indices are greater than 0) and several runtime checks seems a bit muddy.  Why not just make them all runtime checks at that point?

src/io/hdf_hyperslab.h
lshulen: If the type supports that check, then I think it would be a good thing.

src/io/hdf_hyperslab.h
ye-luo: @lshulen I tend to agree with you. Since the compiler doesn't always warning signed to unsigned conversion. A conversion from -1 to unsigned introduces a crazy number. I will remove the unsigned type and add back runtime checks.

ye-luo: The above hyperslab constructor throw runtime error when negative values are detected.

src/io/hdf_hyperslab.h
ye-luo: checkContainerCapacity now handles container size checking.

fdmalone:
> @mmorale3 @fdmalone
> `AFQMC/Hamiltonians/HamiltonianFactory.cpp` and `AFQMC/Hamiltonians/KPFactorizedHamiltonian.cpp`
> can not read Energies correctly.
> In `HamiltonianFactory.cpp`
> 
> ```c++
>     ValueType NuclearCoulombEnergy(0);
>     ValueType FrozenCoreEnergy(0);
> 
>     if(head) {
>       std::vector<ValueType> Rdata(2);
>       if(!dump.readEntry(Rdata,"Energies")) {
>         app_error()<<" Error in HamiltonianFactory::fromHDF5(): Problems reading  dataset. \n";
>         APP_ABORT(" ");
>       }
>       if(Rdata.size()>0)
>         NuclearCoulombEnergy = Rdata[0];
>       if(Rdata.size()>1)
>         FrozenCoreEnergy = Rdata[1];
>     }
> ```
> 
> the h5 file used for unit test contains
> 
> ```
> h5ls -d ham_chol_sc.h5/Hamiltonian/Energies
> Energies                 Dataset {2}
>     Data:
>         (0) -15.3649862050191, 0
> ```
> 
> In a real build,
> `std::vector<ValueType>` is `std::vector<float>` and expect 1D data.
> in a complex build,
> `std::vector<ValueType>` is `std::vector<std::complex<float>>` and expect 2D data and caught by the added error trap in this PR.
> Should the energy always real? Are they used as complex just for convenience in later use?
> In this case, we can define Rdata as `std::vector<RealType>`. I tried this change and all the tests pass. I'd like your confirmation if this is a correct fix before push the code change.

The two numbers should always be real. I may have misunderstood myself when adapting the integral generation scripts. As far as I know FrozenCoreEnergy is probably never used, so it was just a happy accident that writing the NuclearCoulombEnergy in complex number format worked. @mmorale should probably clairfy.

mmorale3:
Inside QMCPACK, the reference energies should be ValueType.
It is probably fine to keep them real in the file. The second energy is
meant to be a frozen core energy,
in case orbitals are frozen and the distinction of energies is necessary
later on.

ye-luo:
> Inside QMCPACK, the reference energies should be ValueType. It is probably fine to keep them real in the file. The second energy is meant to be a frozen core energy, in case orbitals are frozen and the distinction of energies is necessary later on.

Only change I/O. The number will be read as RealType and then always assigned to ValueType for later use.

ye-luo:
@mmorale3 any review comments in the multi related parts?

prckent:
Paging @mmorale3

Issue #1864, opened 2019-09-03
nexus test issue


ye-luo:
Since #1836
https://cdash.qmcpack.org/CDash/testDetails.php?test=6030711&build=77081
This seem caused by the introduce of new modules.
Test exception: "ImportError: cannot import name gcd"

prckent:
Perhaps bora needs a newer numpy?

If true, this also points to our need for a proper set of minimum versions & for nexus to check for them. Otherwise we get these errors that to the user will look like a bug of ours.

prckent:
Paging @jtkrogel  for comment.

ye-luo:
Yes. We need a policy.

jtkrogel:
The issue is the `gcd` function which was added to NumPy in version 1.15 (~July 2018).  Moving to wrap all numpy imports in the umbrella fashion, (i.e. `import numpy as np` then call `np.gcd` somewhere) would protect all the other parts of the code from failing here.

We do need to determine the minimum numpy version (and other modules) for Nexus as well as establish a policy for new additions.

For policy, I think we should hold at least to the "two year rule", so `gcd` is too new and should be replaced.  Suggestions for efficient ways to identify the minimum versions for library components imported into Nexus as it currently stands are welcome.

jtkrogel:
The Nox package looks like a good option for minimum version testing and/or enforcement:

https://nox.thea.codes/en/stable/

prckent:
- You can simply query the versions of the imported modules and quit if too old. Numpy seems to be the critical one - I believe it has been the source of *all* our python related versioning problems.
- For testing we can setup something using the minimum version.
- As a first step, we need to declare the minimum versions as we have for QMCPACK in a readme.md. Then we can update the CI and nightly tests appropriately.

prckent:
My reaction to nox: please no more packages or utilities. Lets do the simplest possible thing. We already know how to install exact versions of python and specific packages (e.g. with spack, we do this for oxygen nightlies). I don't want another dependency for something we have already solved.

jtkrogel:
I agree about query/test based on known minimum version.  If we have an easy way to scan over library (e.g. numpy) versions then we can determine what this is.  At present we do not even know what the minimum version is.

prckent:
1.13.2 was released  on 27 September 2017
1.13.1 was released on 6 July 2017

-> Given a release is upcoming,  see if 1.13.2 works

jtkrogel:
[...]

PR #1865, opened 2019-09-04
Nexus: localize gcd

jtkrogel:
Addresses #1864.  Impact of gcd use is localized to the `get_band_tiling` function.


ye-luo:
Okay to test

PR #1867, opened 2019-09-04
Fix printing of denominator.

fdmalone:
Print block specific denominator to hdf5 rather than all naverages.


markdewing:
okay to test

Issue #1768, opened 2019-07-29
Forces:  Make Ewald the default handler for derivatives


rcclay:
Currently, LRHandlerSRCoulomb handles derivatives of the long range coulomb potential.  I don't think this has been unit tested and there's enough gossip indicating that this probably doesn't work.  Need to shift over derivative computation to the Ewald3D handler (which is unit tested) until LRHandler or LRHandlerSRCoulomb are tested with derivatives.

rcclay:
Closed by #2019

PR #1872, opened 2019-09-05
[WIP] Temporarily fix Ewald handling of force calculations in PBC

tiihonej:
Currently the program crashes, if IonIon or PseudoPot forces are requested in a PBC simulation (OBC works since recently). Temporary solution has been to use a different Ewald handler, as suggested in #1768 . However, this requires more extensive testing, since currently the change of the handler leads to numerical issues and breakdown of many numerous unit tests.

This commit clarifies distinction between force and non-force calculations, enabling safer change of the handler, and consequently, safer usage of "forces=yes" attribute in PBC. The solution is suboptimal, because it involves a non-tested Ewald breakup, but it is strictly limited to force-calculations and generally a better alternative to crashing the program. What do you think, @rcclay ?


src/QMCHamiltonians/tests/test_force.cpp
```diff
@@ -225,7 +225,7 @@ TEST_CASE("Chiesa Force", "[hamiltonian]")
   std::cout << " Force = " << force.forces << std::endl;
 
   // Unvalidated externally
-  REQUIRE(force.forces[0][0] == Approx(3.186559306));
```
rcclay: Why did this number change?  This doesn't bode well.

src/QMCHamiltonians/CoulombPBCAB.cpp
```diff
@@ -181,7 +181,10 @@ CoulombPBCAB::Return_t CoulombPBCAB::evaluateWithIonDerivs(ParticleSet& P,
                                                            ParticleSet::ParticlePos_t& pulay_terms)
 {
   forces = 0.0;
-  Value  = evalLRwithForces(P) + evalSRwithForces(P) + myConst;
+  if (ComputeForces)
```
rcclay: evaluteWithIonDerivs exists to force computation of ion derivatives without the if(ComputeForces) flag.  This function should always call evalXwithForces.

src/QMCHamiltonians/tests/test_force.cpp
tiihonej: I agree, but I don't have a specific explanation. I know this is a result of changing from LRHandlerSRCoulomb to EwaldHandler3D, and thus, actual evidence that the handler case needs more testing.

src/QMCHamiltonians/CoulombPBCAB.cpp
tiihonej: Let me write it down, so that we get mutual understanding of the logic:

The "forces=yes" attribute indicates, whether we want to include force contributions from a particular hamiltonian term, such as IonIon, ElecElec, ElecIon, or PseudoPot. PseudoPot is considered in another class, and we do not concern ElecElec or ElecIon (bare) forces in this case. The main concern here is IonIon, which may contain AA and AB terms. So, adding "forces=yes/no" to IonIon should allocate an Ewald handler with/without gradients, respectively.

The ACForce estimator seems to be a separate player, calling evaluateIonDerivatives() irregardless of the forces-attribute. That means it is up to particular hamiltonians to control whether they are on/off, returning energy/energy+forces, respectively. (For convenience, I think every physically motivated hamiltonian should be on by default.)

The problem arises, when evaluateWithIonDerivatives() is requesting a gradient from a "forces=no"-hamiltonian, because none have been allocated. Possible fixed include

use logic to make handling of Ewald gradients foolproof inside hamiltonian classes (eg CoulombPBCAA/CoulombPBCAB)
always allocate Ewald gradients, when ACForce is present
The current proposition falls into category 1. I don't see a problem with it in IonIon (AB) case, but I agree there could be a problem with ElecIon hamiltonians. It is only logical that evalXwithForces be called every time, but then there should be extra care either in allocating the gradients or requesting them. Either that or we confirm that the gradients are unnecessary, if the the hamiltonian is "excluded" from the forces with "forces=no". Am I making sense?

To me this does not sound like an overly complicated problem, but I thought I'd ask your opinion before making the next move.

tiihonej: Okay, actually the problem is rather with PseudoPot hamiltonian than IonIon, but anyway there is something that needs fixing

PR #1874, opened 2019-09-05
Add more constructors to ComplexApprox

### LINKED ISSUE #1871: test_wavefunction failing for ComplexApprox issue with Intel
https://cdash.qmcpack.org/CDash/testDetails.php?test=6081090&build=77645
```
  REQUIRE( r_all_val == ComplexApprox(ValueType(1.653821746120792, 0.5484992491019633)) )
with expansion:
  (1.65382,0.548499)
  ==
  ComplexApprox( (0.548499,0.548499) )
```
The code is actually correct since the left side matches the reference value on the REQUIRE line.

A much simpler reproducer.
The following check works with RealType as double but not as float
```REQUIRE(std::complex<RealType>(1, 0) == ComplexApprox(std::complex<RealType>(1, 0)));```
The problem only occurs with Intel compiler. GNU and Clang work fine.

Since ComplexApprox is not part of catch, we probably need to have a look more carefully.
@markdewing Could you have a look since you contributed ComplexApprox?
I think we need tests for ComplexApprox. It is part of QMCPACK not Catch. If we have unit tests rely on it, itself must be tested. We don't have any unit tests for it so far.

markdewing:
Addresses an issue with the Intel compiler where a complex float constant would not be passed properly to a function accepting a complex double.

Also add constructors for non-complex double and float to avoid ambiguity errors with non-complex constants.

solves part of #1871

### DIFF
--- a/external_codes/catch/complex_approx.hpp
+++ b/external_codes/catch/complex_approx.hpp
@@ -9,7 +9,23 @@ namespace Catch {
 class ComplexApprox
 {
 public:
-    ComplexApprox(std::complex<double> value) : m_value(value), m_compare_real_only(false) {
+    ComplexApprox(const std::complex<double> &value) : m_value(value), m_compare_real_only(false) {
+      init_epsilon();
+    }
+
+    ComplexApprox(const std::complex<float> &value) : m_value(value), m_compare_real_only(false) {
+      init_epsilon();
+    }
+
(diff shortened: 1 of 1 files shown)


prckent:
Care needed if we ever update catch.

ye-luo:
@prckent ComplexApprox is not from catch but maintained by @markdewing

PR #1875, opened 2019-09-05
Rename QMCHamiltonianBase to OperatorBase

ye-luo:
This is a change by `sed`.
Rename QMCHamiltonianBase to ~HamiltonianComponent~ OperatorBase.
QMCHamiltonian holds a collection of QMCHamiltonianBase and it causes confusion.
It should be clear now.


jtkrogel:
If we want to consider for just a second, the use of "Hamiltonian" for these is a misnomer from the start.  Both Hamiltonian components (e.g. CoulombPotential) and observables that are clearly unrelated to the Hamiltonian (e.g. DensityEstimator) inherit from this as a base class, but all are quantum operators.  Perhaps "OperatorBase" is better?

jtkrogel:
I'll additionally note that the use of `UpdateMode[PHYSICAL]` to distinguish between Hamiltonian components and other observables is rather confusing from a nomenclature point of view (aren't all of these "physical"?).

ye-luo:
Q: What is the real benefit of Hamiltonian components and observables both derived from the same base class and both owned by `QMCHamiltonian H;`?
I was thinking to separate them completely.
An observable may need the whole Hamiltonian like forces.
HamitlonianComponent for those contribute to local energy
ObservableBase for operators not contribute to local energy.
If an operator can be used in both places, have both base classes. For example, MPC has a physical switch.

jtkrogel:
What you are talking about now really goes beyond this PR, but is fine for discussion.

All of these things are observables (quantum operators) and have basically the same needs in that respect, so I think they should share a base class.  If H components are special relative to other observables, maybe they get an intermediate class (HamitonianComponent) that derives from Observable.  There is no logical reason for Hamiltonian to own both sets.  In this respect it really behaves as some kind of ObservableManager, which can legitimately be split off.  Likely this would still own both general types but might only read from HC's and evaluate O's.

ye-luo:
I will change the name to OperatorBase. We have a name collision "Observable".
I think we can keep the same base class until a real need occurs. We may want HamiltonianComponent derived from OperatorBase to have tighter control of requirement.
A derivative may be required for HC but not OB.

PDoakORNL:
I guess many things are better than before, but this is not the interface of an operator.

PDoakORNL:
SuppliesVTableToUnionOfFunctionsUsedByPsiROperatorPsiRThings

prckent:
Confirming that the manual builds.

PR #1876, opened 2019-09-06
Batched implementation in DiracDeterminant

ye-luo:


ye-luo:
@prckent 
For a function overriding its base class virtual function, it is not necessary to put a documentation if the content is the same as the base class version. [Doxygen](https://docs.qmcpack.org/doxygen/doxy/d1/ded/a04035.html) can propagate the base class comment. So I think we can add documentation when it is necessary to mention any difference or implementation details not known to the base class.
I added documentation for ratioGrad_compute. All the mw_ interfaces should just follow the base class.

prckent:
Waiting on Peter

PR #1878, opened 2019-09-09
Add docs on distances and particles to the manual

markdewing:
Some development documentation on how to program ParticleSets, distance tables, and particle groups.

### DIFF
--- a/manual/distance_tables.tex
+++ b/manual/distance_tables.tex
@@ -10,8 +10,7 @@ In the future the interface may change to use functions to set and retrieve posi
 so the SoA transformation of the particle data can happen automatically.
 
 A particular distance table is retrieved with \texttt{getDistTable}.
-The first table in the list is the table for like particles (electron-electron or ion-ion), and is always present.
-Use \texttt{addTable} to add a different \texttt{ParticleSet} and return the index of the distance table.
+Use \texttt{addTable} to add a \texttt{ParticleSet} and return the index of the distance table.
 If the table already exists the index of the existing table will be returned.
 
 The mass and charge of each particle is stored in \texttt{Mass} and \texttt{Z}.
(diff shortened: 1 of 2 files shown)


manual/distance_tables.tex
```diff
@@ -0,0 +1,91 @@
...
+\subsection{ParticleSets}
+The \texttt{ParticleSet} class stores particle positions and attributes (charge, mass, etc).
+
+The \texttt{R} member stores positions.
+For calculations, the \texttt{R} variable needs to be transferred to the structure-of-arrays (SoA) storage in \texttt{RSoA}.   This is done by the \texttt{update} method.
+In the future the interface may change to use functions to set and retrieve positions
+so the SoA transformation of the particle data can happen automatically.
+
+A particular distance table is retrieved with \texttt{getDistTable}.
+The first table in the list is the table for like particles (electron-electron or ion-ion), and is always present.
```
ye-luo: I think this behavior has been removed. The first table only depends on the first call of addTable.

manual/distance_tables.tex
```diff
@@ -0,0 +1,91 @@
...
+The flag, \texttt{IsGrouped}, indicates if the particles are grouped or not.
+The particles will not be grouped if the elements are not grouped together in the input file.
+This ordering is usually the responsibility of the converters.
+
+Code can be written to only handle the grouped case, but put an assert or failure check if the particles are not grouped.  Otherwise the code will give wrong answers and it can be time-consuming to debug.
+
+\subsection{Distance tables}
+
+Distance tables store distances between particles.
+There are symmetric (AA) tables for distance between like particles (electron-electron or ion-ion) and asymmetric (BA) tables for distance between unlike particles (ion-electron)
```
ye-luo: I think electron-ion is more accurate for SoA.

manual/distance_tables.tex
```diff
@@ -0,0 +1,91 @@
...
+
+The \texttt{Distances} and \texttt{Displacements} members contain the data.
+Individual elements are accessed with with the \texttt{()} operator.
```
ye-luo: If I remember correctly, `[i][j]` works but `(i,j)` doesn't for SoA.

manual/distance_tables.tex
markdewing: Yes, this has changed.   Removed that sentence.

manual/distance_tables.tex
markdewing: Changed.

manual/distance_tables.tex
markdewing: The () operator works for Distances, but not for Displacements.   The [] syntax works for both - the example code was updated to use that instead.

PR #1880, opened 2019-09-09
Initial batched inferfaces in ParticleSet

ye-luo:
Review after #1876 Changes should be reduced.
Added interfaces are preliminary since we don't offload distance table computation now.
More changes may be needed in the near future.


prckent:
OK to merge once the oxygen CI passes.

PR #1881, opened 2019-09-09
[WIP] Unified driver calling multi

PDoakORNL:
There is a way to go on this yet but for discussion of RefVector type_traits/template_types.hpp.  I had it previous work already replaced some std::vector<T*> with std::vector<std::reference_wrapper<T>> which provides nearly the semantics of std::vector<T&>.


src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -16,6 +16,7 @@
 #include "Concurrency/Info.hpp"
 #include "Utilities/RunTimeManager.h"
 #include "ParticleBase/RandomSeqGenerator.h"
+#include <boost/range/combine.hpp>
```
markdewing: Is this used somewhere?

PDoakORNL: Good catch, the experiment with boost::combine didn't work as I liked.  std::transform for the most part accomplished what I needed.

PR #1882, opened 2019-09-09
Fix pbc scripts

fdmalone:
Some fixes for AFQMC pyscf PBC integral/wavefunction generation scripts following update for non-integer occupancies.
1. Fixes inconsistent multiplication of nkpts since nelectron is now the total number of electrons in the supercell.
2. Fixes bug when generating trial in MO basis
3. Fixes order of writing wavefunction and hamiltonian so that wavefunction and hamiltonian can be written to same file.


markdewing:
okay to test

PR #1883, opened 2019-09-09
Afqmc workflow tests

fdmalone:
This PR adds the AFQMC examples from the workshop in a slightly modified / updated form as examples and uses these to form a set of workflow tests.

Currently the workflow is:
1. Run pyscf.
2. Generate wavefunction / integrals.
3. Generate input file.
4. Run AFQMC.
5. Analyse result.

To avoid adding a lot of data to the repo any regression in the workflow is only tested in step 5 by analysing the output (assuming the previous steps don't fail entirely).

I decided to keep everything deterministic so I did add a reference scf checkpoint file from pyscf so that the same MO coefficients / transformation matrices could be used. These files are typically small. However, if people prefer no data, I can remove these at the cost of losing deterministic qmc tests.

Ideally these tests could be run with the nightlies and they can be extended to test more complicated input options such as on node parallelism. I needed to modify the xml input file generation routine slightly due to the updated estimators.


markdewing:
okay to test

PR #1884, opened 2019-09-09
Fix issue with lack of msd.

fdmalone:
Fixes bug in UHF PBC wavefunction if no partial occupancies are found.


ye-luo:
Okay to test

PR #1886, opened 2019-09-09
Update afqmc wavefunction converter

fdmalone:
Fixes some issues with conversion of real space MSD wavefunctions to AFQMC format. Mainly handles case when number of orbitals exceeds 64 and multiple bit strings are required for storage of the occupation strings.


prckent:
OK to test

PR #1891, opened 2019-09-11
Fix hdf reading boolean

ye-luo:
Reading bool in hdf must save initial value to prevent uninitialized values in case of missing intended dataset.
It should fix converter_test_LiH_pyscf failure on multiple platforms.
https://cdash.qmcpack.org/CDash/testDetails.php?test=6111561&build=78400

In addition, I add one more header inclusion. Hopefully it fixes the Clang 4.0 compilation error. I don't have clang 4.0 at hand for testing.


prckent:
Confirmed Clang 4.0.1 is working now. Converter tests also worked with this build.

PR #1893, opened 2019-09-11
Wrapper Class RotationHelper for Orbital Optimization 

spinedaflores:
I've made a new pull request. This has my RotationHelper wrapper class implemented to perform orbital optimization for both the MultiSlaterDeterminatFast and DiracDeterminant classes. Threading feature should work for both classes as well. I've also updated the wavefunction files for the test called short-H4-orb-opt


spinedaflores:
@ye-luo 
Hey Ye, This is a new pull request that is cleaner to deal with. I've updated the tests wfn files as I mentioned in the first comment. I'm happy that everything seems to work from some local tests I've done. If you think everything looks good then I should probably delete some now redundant code in the lcaoset class.

markdewing:
okay to test

src/QMCWaveFunctions/Fermion/DelayedUpdate.h
```diff
@@ -106,7 +106,7 @@ class DelayedUpdate
     }
     const T cone(1);
     const T czero(0);
-    const int norb     = Ainv.rows();
+    const int norb     = Ainv.cols();
```
ye-luo: Ainv is the inverse transpose of A. So norb = Ainv.rows() = A.rows().
@spinedaflores did you get any trouble if leaving this unchanged?

src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
```diff
@@ -44,7 +44,11 @@ void DiracDeterminant<DU_TYPE>::set(int first, int nel, int delay)
 {
   FirstIndex = first;
   ndelay     = delay;
-  resize(nel, nel);
+
+  resize(nel, Phi->OrbitalSetSize);
```
ye-luo: I believe you need to do this change to make the code working.
On the other hand, I feel we should not resize all the matrices in DiracDeterminant to Phi->OrbitalSetSize.
Could you describe what you experienced?

src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
```diff
@@ -44,7 +44,11 @@ void DiracDeterminant<DU_TYPE>::set(int first, int nel, int delay)
...
+
+  if(Optimizable)
+    Phi->buildOptVariables(nel);
```
ye-luo: I'm not sure if this is the best place to call this. Leave this message now so I don't forget.

src/QMCWaveFunctions/Fermion/DelayedUpdate.h
spinedaflores: You're right. It seems that we don't need that change. 

On my first attempt of doing the pull request I found that norb was being incorrectly set to number of electrons (when more than the minimal number of orbitals are being read in). I must have changed this when I was trying to figure out why earlier. But it looks like we don't need this change

src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
spinedaflores: If you leave the resize arguments to both be 'nel' the problem occurs when the evaluate functions inside LCAOrbitalSet.cpp are called. LCAOrbitalSet member variable OrbitalSetSize will cause lines such as **std::copy_n(temp.data(0), OrbitalSetSize, psi.data());**  to copy too many elements into psi.data(). So either all the matrices in DiracDeterminant need to be resized or the evaluate functions in LCAOrbitalSet need a different way of determining how many elements to copy into psi.data() argument

src/QMCWaveFunctions/Fermion/DelayedUpdate.h
ye-luo: @spinedaflores Please fix these changes.

ye-luo:
> I've also updated the wavefunction files for the test called short-H4-orb-opt

Could you explain more what do you mean? 

I tried to run the existing tests and got 
```
The following tests FAILED:
	284 - short-H2-orb-opt-16-1 (Failed)
	285 - short-H2-orb-opt-16-1-totenergy (Failed)
	305 - short-H4-orb-opt-16-1-totenergy (Failed)
	307 - short-H4-orb-opt-dmc-16-1-totenergy (Failed)
	309 - short-H4-orb-opt-dmc-1-16-totenergy (Failed)
```
Can we fix the code and pass the existing tests instead of changing tests?

src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
ye-luo: We need to fix the evaluate functions to respect result vector size, namely the number of SPOs needed can be smaller than the number of SPOs SPOSet can provide.

spinedaflores:
@ye-luo when I said that I updated the test files it was just the formatting of the input files so that variables would be read in correctly. The tests that aren't "total energy" pass for me. But I just realized that the failure of the total energy could be due to the fact that the original tests were probably restricted orbital optimization instead of the current unrestricted.

spinedaflores:
I forgot to mention that this was for short-H4-orb-opt. I didn't update the short-H2-orb-opt one because I didn't see it. Sorry about that I'll look at it now

spinedaflores:
@ye-luo Do you need me to redo reference files for short-H4-orb-opt test? Is there anything you need from me?

ye-luo:
@spinedaflores 
I corrected how the RotationHelper is connected to the builders and fixed all the deterministic test failing.

> The tests that aren't "total energy" pass for me.

It only means no crash, no error at run time. The energy check needs to be fixed.
The SoA code gets higher energy than the AoS code. The AoS was spin unrestricted when rotation is used.

Please recover the spin restricted input of wfs_3 although moving the "opt_var" from "determinant" to "sposet", we should make an effort toward correct restricted spin support.

Please clean up LCAOrbitalSet by removing all pieces which have been moved to RotationHelper already.

src/QMCWaveFunctions/SPOSetBuilderFactory.cpp
```diff
@@ -246,10 +247,12 @@ SPOSet* SPOSetBuilderFactory::createSPOSet(xmlNodePtr cur)
   std::string bname("");
   std::string sname("");
   std::string type("");
+  std::string rotation("no");
   OhmmsAttributeSet aAttrib;
   aAttrib.add(bname, "basisset");
   aAttrib.add(sname, "name");
   aAttrib.add(type, "type");
+  aAttrib.add(rotation, "optimize");
```
ye-luo: To be more precise, orbital optimization is also possible without using rotation. We may need more precise input paramters. I'm thinking of optimize="rotate" instead of just optimize="yes". To be discussed.

src/Optimize/VariableSet.cpp
```diff
@@ -274,7 +274,7 @@ void VariableSet::print(std::ostream& os, int leftPadSpaces, bool printHeader)
        << setw(max_value_len) << NameAndValue[i].second << " " << setw(max_type_len) << ParameterType[i].second << " "
        << setw(max_recompute_len) << Recompute[i].second << " ";
 
-    os << std::defaultfloat;
+//    os << std::defaultfloat;
```
ye-luo: Could you paste printout before and after this change? It helps us understanding the change.

src/QMCWaveFunctions/RotationHelper.h
```diff
@@ -0,0 +1,267 @@
+#ifndef ROTATION_HELPER_H
```
ye-luo: QMCPLUSPLUS_ROTATION_HELPER_H

src/QMCWaveFunctions/RotationHelper.cpp
```diff
@@ -0,0 +1,845 @@
+#include "QMCWaveFunctions/RotationHelper.h"
```
ye-luo: Missing license header in both cpp and h files.

src/QMCWaveFunctions/RotationHelper.h
```diff
@@ -0,0 +1,267 @@
...
+#define ROTATION_HELPER_H
+
+#include "QMCWaveFunctions/SPOSet.h"
+#include <memory>
+#include "QMCWaveFunctions/BasisSetBase.h"
+
+#include <Numerics/MatrixOperators.h>
+#include "Numerics/DeterminantOperators.h"
```
ye-luo: Please help me deleting these four inclusion which are no more needed after my change

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -137,15 +137,32 @@ class SPOSet : public QMCTraits
 #endif
 
   /// create optimizable orbital rotation parameters
+  // Single Slater creation
+  virtual void buildOptVariables(const size_t& nel) {}
```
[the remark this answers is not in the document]
spinedaflores: The Orbital Rotation Parameters for a Multi-Slater Jastrow wfn will depend on the determinants present in the wfn, while for the single Slater it is simply all occupied to virtual rotation parameters.

src/Optimize/VariableSet.cpp
spinedaflores: The only reason I commented this line out is because the compiler I am using doesn't seem to support this manipulator and I will get a compiler error 

the compiler I'm using  
intel/18.0.5/openmpi/1.10.1/bin/mpicc
intel/18.0.5/openmpi/1.10.1/bin/mpicxx

the error
/ascldap/users/sdpined/Desktop/qmcpack_pull_request_Ye/src/Optimize/VariableSet.cpp(277): error: namespace "std" has no member "defaultfloat"
      os << std::defaultfloat;

src/QMCWaveFunctions/RotationHelper.h
```diff
@@ -0,0 +1,267 @@
...
+                         const size_t NP2,
+                         const std::vector<std::vector<int>>& lookup_tbl);
+
+  void checkInVariables(opt_variables_type& active)
+  {
+    if (Optimizable && !IsCloned)
+    {
+      if (myVars.size())
+        active.insertFrom(myVars);
+      Phi->storeParamsBeforeRotation();
```
ye-luo: @spinedaflores it becomes clear to me setting deltaP zero should happen in checkInVariables since storeParamsBeforeRotation is called here.

src/Optimize/VariableSet.cpp
ye-luo: Could you type "icc -v" and report the output here? I think it is probably due a too low gcc version. It seems passing our C++14 check.

ye-luo:
```
$ numactl -N1 ctest -j16 -R orb-opt
Test project /sandbox/opt/cleanup/spinedaflores/build_real
      Start 308: short-H4-orb-opt-16-1
 1/14 Test #308: short-H4-orb-opt-16-1 .................   Passed   24.03 sec
      Start 286: short-H2-orb-opt-16-1
 2/14 Test #286: short-H2-orb-opt-16-1 .................   Passed    9.56 sec
      Start 314: short-H4-orb-opt-dmc-1-16
 3/14 Test #314: short-H4-orb-opt-dmc-1-16 .............   Passed    7.64 sec
      Start 312: short-H4-orb-opt-dmc-16-1
 4/14 Test #312: short-H4-orb-opt-dmc-16-1 .............   Passed    6.38 sec
      Start 306: short-H4-RHF-orb-opt-16-1
 5/14 Test #306: short-H4-RHF-orb-opt-16-1 .............   Passed   24.03 sec
      Start 310: short-H4-orb-opt-1-16
 6/14 Test #310: short-H4-orb-opt-1-16 .................   Passed   27.33 sec
      Start 332: short-H4-msd-orb-opt-16-1
 7/14 Test #332: short-H4-msd-orb-opt-16-1 .............   Passed   11.86 sec
      Start 307: short-H4-RHF-orb-opt-16-1-totenergy
 8/14 Test #307: short-H4-RHF-orb-opt-16-1-totenergy ...***Failed    0.06 sec
      Start 287: short-H2-orb-opt-16-1-totenergy
 9/14 Test #287: short-H2-orb-opt-16-1-totenergy .......   Passed    0.07 sec
      Start 309: short-H4-orb-opt-16-1-totenergy
10/14 Test #309: short-H4-orb-opt-16-1-totenergy .......   Passed    0.07 sec
      Start 313: short-H4-orb-opt-dmc-16-1-totenergy
11/14 Test #313: short-H4-orb-opt-dmc-16-1-totenergy ...   Passed    0.08 sec
      Start 315: short-H4-orb-opt-dmc-1-16-totenergy
12/14 Test #315: short-H4-orb-opt-dmc-1-16-totenergy ...***Failed    0.08 sec
      Start 333: short-H4-msd-orb-opt-16-1-totenergy
13/14 Test #333: short-H4-msd-orb-opt-16-1-totenergy ...***Failed    0.08 sec
      Start 311: short-H4-orb-opt-1-16-totenergy
14/14 Test #311: short-H4-orb-opt-1-16-totenergy .......***Failed    0.08 sec

71% tests passed, 4 tests failed out of 14
```
Issues:
1. The reference values of short-H4-msd-orb-opt-16-1 probably needs to be changed. It seem too low energy.
2. All the 1 MPI + 16 threads runs are not reliable.
3. RHF is giving a higher energy. It sounds like it should be but I still feel there might be bugs.

This PR is already better than the AoS code. Thus left issues can be dealt in later PRs if changes can not get in immediately. This PR breaks AoS single det orbital rotation but we will delete instead of fix.

ye-luo:
Hard abort for AoS+rotation has been added. Two issues are added to track next action items.

ye-luo:
lost connection to alcf-gnu6-mkl18. Ignore its CI result. The previous commit passed.

PR #1895, opened 2019-09-11
Add PYSCF_BIN variable to guard afqmc workflow tests.

fdmalone:
Should fix test breakages for nightlies due to a lack of pyscf.


markdewing:
okay to test

prckent:
Once this is proven working the manual and README.md will need updating to mention the new PYSCF_BIN. At that point please request the real space PySCF folks (@anbenali ?) to add a real space workflow check using PySCF.

PR #1899, opened 2019-09-12
Add statistical tests for diamondC_1x1x1 with bspline wavefunction converted from LCAO

Hyeondeok-Shin:
As @anbenali requested, statistical tests for periodic 1x1x1 diamond with real-space bspline wavefunction converted from periodic LCAO generated with Pyscf. The bspline converter @Paul-St-Young made is used.

 The same testsets with LCAO are added for comparison of performance and statistical result between LCAO and converted bspline. They should give the same statistical result, so they are using the same references.

 All tests were passed with Intel Xeon with gcc 6.2.0.

 Hyeondeok Shin


markdewing:
okay to test

prckent:
Q. What is the current state and timeline for the converter?

Hyeondeok-Shin:
@prckent Now we're testing converter to give flexibility for various templates in pyscf input. Our plan is to complete it and make PR before thanksgiving. Sorry for the delays.

Hyeondeok-Shin:
This PR will be closed and reopened with smaller h5 and converter when ready.

PR #1901, opened 2019-09-13
Nexus: remove dependence on dateutil

jtkrogel:
Although the `dateutil` module is widely available (every local install of Python I've ever seen), it is not part of the Python standard library.  This PR removes the dependency on `dateutil`.

Issue reported by @PDoakORNL.  Peter, please can you confirm that tests now pass with your specialized Spack install?


markdewing:
okay to test

markdewing:
This is a small change - I'm going to merge this now and @PDoakORNL can check it when he has time.

PR #1903, opened 2019-09-13
Nexus: cleanup unused and/or outdated code

jtkrogel:
Delete unused or outdated code ahead of the testing push.


markdewing:
okay to test

PR #1906, opened 2019-09-14
Add offload batched spline evaluation

ye-luo:
Support batched evaluation in SplineC2ROMP.

Note: current organization of SplineAdopter classes is suboptimal. I had to copy paste the fall back solution in every class. Better scheme needs to be explored.

PS: to workaround unit test failure. I put omp_get_max_threads() instead of hard-coded 8.


prckent:
What is the root cause of the unit test failure? Them or us?

prckent:
What is the logic behind similarly named mw_evaluateVGL and mw_evaluate_vgl? It would be better to have clearly distinct names.

prckent:
Is the plan to offload only full value-gradient-laplacian calls? i.e. Is value only offload not worth it, or is it simply not important enough at this time?

ye-luo:
@prckent 
>What is the root cause of the unit test failure? Them or us?

IBM runtime cap number of threads to 4, If there is no OMP_NUM_THREADS or OMP_THREAD_LIMIT specified. We either enforce 8 if we intended to test 8 threads (we did this in the driver tests) or change the test to used default max number of threads

>What is the logic behind similarly named mw_evaluateVGL and mw_evaluate_vgl? It would be better to have clearly distinct names.

mw_evauluateVGL was required by the base class.
mw_evaluate_vgl belongs to the adopor and it is not a virtual funciton specialization.
mw_evauluateVGL calls SplineAdopter::mw_evaluate_vgl internally.
Currently the single evaluation is called SplineAdopter::evaluate_vgl so I just add mw_.
Long term, prefer to reduce the number of layers and keep one name.

> Is the plan to offload only full value-gradient-laplacian calls? i.e. Is value only offload not worth it, or is it simply not important enough at this time?

single walker value evaluation doesn't gain on GPU. But batched value only evaluation will be added although it is not important at the moment. NLPP batched evaluation has its dedicated code path.

PR #1907, opened 2019-09-16
Nexus: tests for simple utilities

jtkrogel:
This PR adds tests for modules that perform simple utility functions.


markdewing:
okay to test

PR #1908, opened 2019-09-16
Initial running VMC batched driver

ye-luo:
Changes are mostly in the batched driver. I can get the code running and extra performance numbers. The statistics are not fully checked since the output to scalar.dat is still being worked. I only did preliminary printing checking.

Note: manual updated to reflect the real code, usedrift is on by default.

This PR mixes code from me and @PDoakORNL. Thus we need additional reviewer.


src/QMCDrivers/Crowd.h
```diff
@@ -28,8 +28,11 @@ class Crowd
 {
 public:
   using MCPWalker = MCPopulation::MCPWalker;
+  // This is getting a bit eh.
```
prckent: Decryption please. Not scalable? Ugly?

src/QMCHamiltonians/QMCHamiltonian.cpp
```diff
@@ -488,48 +488,59 @@ QMCHamiltonian::Return_t QMCHamiltonian::evaluate(ParticleSet& P)
...
-    for (int i = 0; i < H.size(); ++i)
+    for (int i = 0; i < H_list[0].get().H.size(); ++i)
     {
-      ScopedTimer local_timer(myTimers[i]);
+      ScopedTimer local_timer(H_list[0].get().myTimers[i]);
       const auto HC_list(extract_HC_list(H_list, i));
-      H[i]->mw_evaluate(HC_list, P_list);
+
+      // This lambda accomplishes two things
+      // 1. It makes clear T& and not std::reference<T> is desired removing need for gets.
```
markdewing: Should this be std::ref<T> or std::reference_wrapper<T>?

src/QMCDrivers/Crowd.h
PDoakORNL: I'm not happy with the MCPWalker name.  But Walker isn't a safe type name to use in the code yet.

src/QMCHamiltonians/QMCHamiltonian.cpp
PDoakORNL: should be reference_wrapper an important disambiguation here.

src/QMCWaveFunctions/TrialWaveFunction.h
```diff
@@ -341,29 +347,29 @@ class TrialWaveFunction : public MPIObjectBase
...
-  }
+  static std::vector<WaveFunctionComponent*> extract_WFC_list(const UPtrVector<TrialWaveFunction>& WF_list, int id);
 
   static std::vector<std::reference_wrapper<WaveFunctionComponent>> extract_WFC_list(
       const std::vector<std::reference_wrapper<TrialWaveFunction>>& WF_list,
       int id);
+  /** }@ */
+
+  // helper function for extrating a list of gradients from a list of TrialWaveFunction
+  std::vector<ParticleSet::ParticleGradient_t*> extract_G_list(const std::vector<TrialWaveFunction*>& wf_list) const;
```
markdewing: Having both extract_G_list and extract_g_list seems confusing.    Maybe extract_G_list_ptr and extract_g_list_ref, which would get a little more clue as to difference in function.
It looks like there is an (incompletely applied) naming convention that lowercase is for a reference vector and upper case is for a vector of pointers ???
It seems that the goal is to use references as much as possible and avoid the use of pointers (correct?)   Is there a remaining use for a vector of pointers, or is it for backwards compatibility in places?

src/QMCDrivers/Crowd.h
ye-luo: Do we have a rule for adding comments before or after? I prefer before the code.

src/QMCWaveFunctions/TrialWaveFunction.h
ye-luo: We will have only one in the end. Because we are at the transition period, both of them are used. To be accurate it is ref_list and ptr_list instead of list_ref and list_ptr.

PDoakORNL: I was updating the naming and changing to reference vectors while ye's lower layer code kept the old naming and needed the ptr lists, so this was convenient but as you point out not intelligible to anyone.  The intention is to toss the ptr_lists but especially with G and L it seems like we'll at least have a few months with both.  I think the rename refactor will be fast let me check.

src/QMCDrivers/Crowd.h
PDoakORNL: There is a rule and its before. To be honest I'm not 100% sure which type alias I was commenting on I will clean

PR #1909, opened 2019-09-16
Update group in mpi_wrapper from gitlab.

markdewing:
Also requires a change in AFQMC code to match an interface change.


mmorale3:
Mark, How did you handled the problem with the barrier calls inside shm_ptr_with_raw_ptr_dispatch.hpp?

markdewing:
The barrier call problem occurs in shared_window.hpp, which is not updated in the PR.

external_codes/mpi_wrapper/mpi3/shared_window.hpp
```diff
@@ -33,7 +33,9 @@ struct shared_window : window<T>{
 	{}
 	shared_window(shared_window const&) = default;
 	shared_window(shared_window&& other) : window<T>{std::move(other)}, comm_{other.comm_}{}
-	auto get_group() const{return group(*this);}
+		group get_group() const{
```
PDoakORNL: fomatting seems kind of random here.

PR #1910, opened 2019-09-16
Nexus: tests of testing module

jtkrogel:
Harden functions used for testing and add explicit tests for these functions.  This PR depends on #1907.


markdewing:
okay to test

PR #1911, opened 2019-09-16
QMCPACK to FCIDUMP

fdmalone:
Allows for QMCPACK Cholesky factorised Hamiltonians (both with and without k-point symmetry) to be written in FCIDUMP format. Helpful for interfacing to chemistry codes.


ye-luo:
Okay to test

PR #1915, opened 2019-09-17
Unified driver running2

PDoakORNL:
Unified driver running and collecting basic scalar esitmators.  Everything but accept ratio appearing in scalar.dat.   vmc and vmcbatched with crowds of 1 walker appear to produce equivalent output. A great deal of cleanup to do yet with estimator manager and the lighter weight estimator manager crowd.


ye-luo:
@prckent I restored the test to be the same as the others. We don't have substeps in the batched driver. So the statistics is not equivalent to others. We will need substeps back.

ye-luo:
@Hyeondeok-Shin after merge please create a deterministic test for det_qmc_short_vmcbatch.in.xml. Please replace the commented deterministic test which I didn't finish.

PR #1919, opened 2019-09-17
Fix writing of 4-fold symmetric FCIDUMPs

fdmalone:
Incorrect logic lead to symmetry equivalent integrals being printed. I added to the unit test to check for this.


markdewing:
okay to test

PR #1920, opened 2019-09-18
Quality of life improvements including memory leak fixes and boost header deprecations

PDoakORNL:
Several Fixes Here:

1.  LLVM address sanitizer works for all build types.
2. Some allocators return a pointer that is basically a[0] when you a = `allocator.allocate(0)`. This must still be freed or you leak 1 byte of memory.  This is standards compliant but makes VectorSoaContainer container leak memory.  The work around is not to allocate size zero vectors.
3. boost/random.h spews deprecated header warnings and I just couldn't take it anymore. The boost folks need to update their header.
4. I needed fast and small test of QMCDriverNew to work on sanitizer issues.  Required improvement of mock derived driver.
5. In case someone still calls Mallocator to allocate zero size memory, it is more logical to return a nullptr. What the underlying call is going to give us we can't be sure.


src/OhmmsPETE/OhmmsVector.h
```diff
@@ -220,6 +220,13 @@ class Vector
     {
       mAllocator.deallocate(X, nAllocated);
     }
+    // when n == 0 std::allocator's return value is unspecified.
+    if(n == 0)
```
markdewing: The values set if n==0 all get overwritten by the code following the if-test

PDoakORNL: This is why we have code review.  It's supposed to return.

PDoakORNL: Actually now that I look at this closer you wont make it to resize_imp if n == 0.

PDoakORNL: Some I'm reverting this.

ye-luo:
Will review end of day.

PDoakORNL:
Just for the record I don't require all requested reviews. But I plan to wait on @ye-luo

src/OhmmsSoA/VectorSoaContainer.h
```diff
@@ -47,7 +47,7 @@ struct VectorSoaContainer
   ///allocator
   Alloc myAlloc;
   ///default constructor
-  VectorSoaContainer() { setDefaults(); }
+  VectorSoaContainer() : nLocal(0), nGhosts(0), nAllocated(0), myData(nullptr) {}
```
ye-luo: I think we should fully replace all setDefaults. is there any left?

src/OhmmsSoA/VectorSoaContainer.h
```diff
@@ -75,13 +75,14 @@ struct VectorSoaContainer
   }
 
   ///move constructor
-  VectorSoaContainer(VectorSoaContainer&& in) : nLocal(in.nLocal), nGhosts(in.nGhosts)
+  VectorSoaContainer(VectorSoaContainer&& in) : nLocal(in.nLocal), nGhosts(in.nGhosts), nAllocated(in.nAllocated), myData(in.myData)
   {
-    nAllocated    = in.nAllocated;
-    myData        = in.myData;
+    free(myAlloc);
```
ye-luo: I feel bad using free in C++ code. Is the std::vector implemented in a same way?

src/OhmmsSoA/VectorSoaContainer.h
```diff
@@ -139,6 +140,15 @@ struct VectorSoaContainer
...
     nLocal     = n;
+    // Bail out to avoid some allocator implementations returning allocated ptr that
+    // must be freed, breaking the nAllocated scheme.
+    if(n == 0)
+    {
+      nAllocated = 0;
+      nGhosts = 0;
+      myData = nullptr;
+      return;
+    }
```
ye-luo: I feel better with if-else. The code logic is more clear. return in the middle of the code is sub-optimal.

src/simd/Mallocator.hpp
```diff
@@ -42,6 +42,8 @@ struct Mallocator
 
   T* allocate(std::size_t n)
   {
+    if (n == 0)
+      return nullptr;
```
ye-luo: My understanding for C++ allocator is allocate(0) is unspecified behavior.
All the containers we made must not rely on unspecified behavior.
I feel better throw an error if 0 is requested. This helps catching any bugs in our containers and the performance hit due to the check should be negligible. 
The check needs to be added in both allocate() and deallocate()
@PDoakORNL any thoughts?

PDoakORNL: I read it as unspecified but you aren't allowed to crash. I wasn't ready to go to a throw, but I think you're right. A 0  allocate shouldn't make it to Mallocator, it might not be safe with another allocator and we don't know what they will do. I will update.

src/OhmmsSoA/VectorSoaContainer.h
PDoakORNL: Agree will clean up.

src/OhmmsSoA/VectorSoaContainer.h
PDoakORNL: I'll check.

src/OhmmsSoA/VectorSoaContainer.h
PDoakORNL: Both gnu and llvm avoid use of malloc or free in their vector source. Worth changing it.

ye-luo: Probably just put in the initializer list of the move constructor `myAlloc(std::move(in.myAlloc))`

PDoakORNL: The use of free isn't limited to here unfortunately.  I want to put this off until some other work on OhmmsVector since it requires changes there as well.

src/OhmmsSoA/VectorSoaContainer.h
PDoakORNL: restructured this code.

src/simd/Mallocator.hpp
PDoakORNL: So Mallocator will throw now.

prckent:
@PDoakORNL Finished with updates?

PDoakORNL:
Yes

src/OhmmsSoA/VectorSoaContainer.h
```diff
@@ -119,41 +113,51 @@ struct VectorSoaContainer
...
        */
   __forceinline void resize(size_t n)
   {
-    static_assert(std::is_same<Element_t, typename Alloc::value_type>::value, "VectorSoaContainer and Alloc data types must agree!");
+    static_assert(std::is_same<Element_t, typename Alloc::value_type>::value,
+                  "VectorSoaContainer and Alloc data types must agree!");
+    if (n == 0)
+    {
+      free();
+      return;
```
markdewing: It looks like the code was restructured to use if/else, but the returns were not removed?

ye-luo:
Wait for my last changes.

PR #1921, opened 2019-09-18
Unified driver single walker per crowd QMCHamiltonian::flex_evaluate optimization

PDoakORNL:
Restored a working single walker path to the flex_evaluate call.  Checked that RVO actually occurs for returned vector, it does at least with clang 8.  Added force_rvo template which can be used as shown in 
commit 2d77d00


PDoakORNL:
Also put ContextForSteps on a diet. Going to try having a per walker RandomGen, if that works sensibly class will be eliminated.

ye-luo:
@PDoakORNL I'd like to keep things simple. I believe most compilers have RVO optimization. If not, the extra cost is very low because, the call counts of Hamiltonian evaluate is quite low.

src/QMCHamiltonians/QMCHamiltonian.cpp
```diff
@@ -532,16 +532,16 @@ std::vector<QMCHamiltonian::RealType> QMCHamiltonian::flex_evaluate(const RefVec
...
-    
-    //}
-  // else if (H_list.size() == 1)
-  // {
-  //   H_list[0].get().evaluate(P_list[0]);
-  // }
+  }
+  else if (H_list.size() == 1)
+  {
+    std::vector<RealType> local_energies(1,0.0);
```
ye-luo: Need resize, not declaration.

ye-luo: There is resize above. Seems need nothing here.

src/QMCHamiltonians/QMCHamiltonian.cpp
```diff
@@ -491,8 +490,9 @@ QMCHamiltonian::Return_t QMCHamiltonian::evaluate(ParticleSet& P)
 std::vector<QMCHamiltonian::RealType> QMCHamiltonian::flex_evaluate(const RefVector<QMCHamiltonian>& H_list,
                                    const RefVector<ParticleSet>& P_list)
 {
-  // if (H_list.size() > 1)
-  // {
+  std::vector<RealType> local_energies(H_list.size(),0.0);
```
ye-luo: Although I hate QMCHamiltonian::Return_t, Please use consistent data type as single evaluation.
Return_t is always on full precision. We want local energy not losing accuracy.

src/QMCHamiltonians/QMCHamiltonian.cpp
PDoakORNL: 😢

PR #1925, opened 2019-09-19
More granual timers in VMCBatched.  Fixed ccmake ENABLE_TIMERS issue.

PDoakORNL:


src/QMCDrivers/QMCDriverNew.h
```diff
@@ -85,6 +85,32 @@ class QMCDriverNew : public QMCDriverInterface, public MPIObjectBase
    */
   std::bitset<QMC_MODE_MAX> qmc_driver_mode_;
 
+  /** The timers for the driver.
+   *
+   * This cleans up the driver constructor, and a reference to this structure 
+   * Takes the timers into thread scope.
+   */
+  struct Timers
```
ye-luo: Is DriverTimers a better name?

PDoakORNL: Thinking about this even though its a nested type its less readable used as a argument to the threaded functions.  I'll change to `DriverTimers`

src/QMCDrivers/QMCDriverNew.h
```diff
@@ -304,7 +331,10 @@ class QMCDriverNew : public QMCDriverInterface, public MPIObjectBase
 
   /** }@ */
 
-        std::vector<RandomGenerator_t*> RngCompatibility;
+  std::vector<RandomGenerator_t*> RngCompatibility;
+
+  Timers timers_;
```
ye-luo: Is it under `public:`? I think it should be in protected.

src/QMCDrivers/QMCDriverNew.h
PDoakORNL: Done

src/QMCDrivers/QMCDriverNew.h
PDoakORNL: Done

markdewing:
@ye-luo I will look into using ScopedTimer with a reference.

PR #1927, opened 2019-09-20
Nexus: flexible testing of float equality

jtkrogel:
Treat native float and numpy.float64 on equal footing for the purposes of equality testing.


markdewing:
okay to test

PR #1932, opened 2019-09-22
Nexus: expand tests for structure module

jtkrogel:
Expand test for `structure` module.  A large number of functions remain untested (see `# test needed` in `structure.py`).  Tests for most of these are deferred beyond the Python 3 transition.  This PR depends on #1931.


prckent:
OK to test

PR #1934, opened 2019-09-23
Add Spin Variables to ParticleSet and Walker

rcclay:
For dynamical spin and non-collinear calculations, one needs to do MC on both the 3N electron coordinates **R** and N spin coordinates **s**, and they both need to be treated on par with one another.  The current solution I have permanently adds **s** to ParticleSet and Walker, which gets parsed and resized along with R, group ID's, etc.  

PRO:  there's very little code (~18 lines including comments, excluding unit test) to generalize the ParticleSet/Walker for spins, and parsing an XML block like this happens automatically:  
```
<particleset name="e">
    <group name="e" size="3"> 
    <parameter name="charge">-1</parameter> 
    <attrib name="position" datatype="posArray"> 
      -0.28   0.0225     -2.709 
      -1.28   1.0225     -1.709 
      -2.28   2.0225     -0.709 
    </attrib> 
    <attrib name="spins" datatype="scalarArray"> 
      1.0 
      0.2 
      3.0 
    </attrib> 
  </group> 
</particleset> 
```

CON: ParticleSet and Walker now grow by an N_elec double array for everyone.  While it would be nice to allocate this only when needed, the way that ParticleSet XML parsing->ParticleSet construction for attributes works, this will add quite a bit of code complexity that arguably might not be worth the slight increase in memory requirements.  

I'm inclined to add **s** as is, but does anyone have strong feelings on this?


PDoakORNL:
Starting question: right now the particle "groups" just seem to exist to have two boxes of particles, one up and one down. I think a more general solution here to be to template on the SpinType and then have a spin variable per particle instead of using whether a particles index is above or below some value. Possibly these spin types would be Bool(up/down), Scalar, Vector (Non collinear).  We could then specialize on the spin type to do the spin elements of the calculation.

ye-luo:
ParticleSet is not just for electrons but also for ions and point charges. Group is not only for up and down. In the case of ion, it mean species. It should be allowed to construct a ParticleSet with up/down electrons + protons all walking as quantum particles. So no matter what we do for spin, the groups can not be removed. We may replace some places using spins instead of groups.

ye-luo:
I do believe in the future R, R+S will be on its own layer beneath ParticleSet.
I was thinking about how to generalized QMCPACK to handle lattice model. As long as we abstract all the random walking operations, ParticleSet can handle lattice sites as well.
So to handle the three cases, scalar spin, vector spin, lattice sites, we will need to build better abstractions in ParticleSet.

ye-luo:
So far the change in ParticleSet is still light, I believe the future change in driver to and R and R+S case can become more ugly. All the comments I put didn't mean to block progress. If all of thus think a bit more, we might find a much better solution.

prckent:
I think the discussion should focus of any downsides of trying out this route and whether there are any simple improvements.

The refactoring of particleset is a longer discussion and should not slow down this work. e.g. As is well known, I am in favor of greatest simplicity and minimum number of layers, particular for critical classes such as this one.

ye-luo:
For trying out, S can be added next to RSoa.
@rcclay Will the random walking of spin pollute our driver?

Refactoring particleset is urgent because we need to fix R, RSoa, Rgpu problem. This is not very different from how to manage S.
Refactoring should happen in parallel with other developments.

rcclay:
Going from specific to high level:

- @ye-luo The addition of S shouldn't pollute the drivers.  I anticipate the only changes to the drivers required will be the addition of spin-specific movers, which will just manipulate that extra degree of freedom.
- @PDoakORNL  Some sort of templating and just better abstraction in general would definitely help out ParticleSet.   Rather than templating on the spin type however, it might be a good idea in the future to template ParticleSet and Walker based on the type of R, since it's R that contains all the "degrees of freedom" that need to be manipulated in a QMC calculation.  For example, R would stay the same for normal calculations, R->[(x0,y0,z0,s0),...(xN,yN,zN,sN)] for spin calculations, R could become a vector of integer positions for lattice calculations @ye-luo , etc (could easily think of different spatial dimensions).  Additionally, there's the 500lb Gorilla in the room concerning observables:  any observable with a spin operator in it will need to be treated differently between the spin-dependent and spin-conserved calculations.  Templating observables and wave functions on ParticleSet type seems better than having independent "SpinDensity" and "SpinDependentSpinDensity" classes for example.  By the way, there's a decent chance I will bite the bullet and template ParticleSet myself because of this very point, but probably when I get to that part of the code.

To the point that @prckent made though, these changes are not ideal, but they are small, unobtrusive, and would allow me to develop and unit test other (IMHO) more important kernels in mainline QMCPACK (spinor evaluation, spin-orbit hamiltonian, etc).  I would suggest merging this for now, and in the mean time, having a design discussion about what ParticleSet is not doing well currently and what we think it should do in the future, perhaps at an ECP meeting.  Off the bat:
1.)  Consistent initialization is nontrivial (@jptowns), and closely tied to either HDF or XML.  Makes unit testing a real pain.  
2.)  Adding or changing degrees of freedom is an all or nothing approach without a looot of code logic hacked in.  Also loops back to point 1 for initializing these new degrees of freedom.  
3.)  Everything takes ParticleSet as an argument.

markdewing:
I think this should go in to allow continued progress on the spin-dependent code.

BTW on point 2 about everything being public - @ye-luo and I have discussed some initial fixes for this and I will start implementing them.

PR #1935, opened 2019-09-24
Replace compare_real_only with std::real

markdewing:
As @ye-luo pointed out, std::real can be used instead of ComplexApprox().compare_real_only();

Also remove the 'compare_real_only' code from ComplexApprox.

On the surface this PR is a simple refactoring.   More significantly, the code changes were performed with a new code transformation tool called 'Comby'.

The homepage: https://comby.dev/

The transformation was performed with this command:
```
comby -i \
-matcher .c \
"REQUIRE(:[var1]==:[]ComplexApprox(:[stuff]).compare_real_only());" \
"REQUIRE(std::real(:[var1]) == Approx(:[stuff]));" 
```
After this ran there was one line (involving 'epsilon()') that I had to modify by hand.

There is a live editor that is useful for developing the matches.  This link should go to the page that I used to develop this transformation.
https://bit.ly/2kVgTk0


external_codes/catch/complex_approx.hpp
```diff
@@ -9,19 +9,19 @@ namespace Catch {
 class ComplexApprox
 {
 public:
-    ComplexApprox(const std::complex<double> &value) : m_value(value), m_compare_real_only(false) {
+    ComplexApprox(const std::complex<double> &value) : m_value(value) {
       init_epsilon();
     }
 
-    ComplexApprox(const std::complex<float> &value) : m_value(value), m_compare_real_only(false) {
+    ComplexApprox(const std::complex<float> &value) : m_value(value) {
```
ye-luo: Is it possible to reduce the four constructors to two template constructors, one for complex and one for real?

external_codes/catch/complex_approx.hpp
```diff
@@ -42,10 +41,7 @@ class ComplexApprox
     friend bool operator == (double const& lhs, ComplexApprox const& rhs)
     {
         bool is_equal = rhs.approx_compare(lhs, rhs.m_value.real());
-        if (!rhs.m_compare_real_only)
-        {
-          is_equal &= rhs.approx_compare(0.0, rhs.m_value.imag());
-        }
+        is_equal &= rhs.approx_compare(0.0, rhs.m_value.imag());
```
ye-luo: The comparison between real lhs value and a complex rhs is invalid. ComplexApprox should always check both the real and the complex part. Please also remove the next function mixing complex and real.

prckent:
Useful tool

external_codes/catch/complex_approx.hpp
markdewing: Maybe, but that feels like a change for a different PR.

external_codes/catch/complex_approx.hpp
markdewing: Fixed.

external_codes/catch/complex_approx.hpp
ye-luo: Could you try it locally? If many other changes are needed, then put it on a separate PR.

markdewing: I tried it locally, it works.   However, part of the reason for so many constructors is a bug in the Intel compiler.   I'm not really interested in doing the testing.   You can make the change if you feel it is important enough, and want to do the testing.

ye-luo: If I remember correctly, Intel compiler failed if there is only `ComplexApprox(std::complex<double> value)`. If we use `ComplexApprox(std::complex<T> value)`, it will probably work and we can save a few lines of code. However, the real issue of the ComplexApprox is that it is part of QMCPACK not Catch. If we have unit tests rely on it, itself must be tested. We don't have any unit tests for it so far.

ye-luo: For sure, adding tests should be a separate PR.

ye-luo: OK. This seems not critical. Let us tackle it when it becomes an issue or we need to work on this part of code. We may use it beyond just `float` and `double` .

PR #1936, opened 2019-09-24
Nexus: expand tests for physical system module

jtkrogel:
Expand existing set of tests for the `physical_system` module.  Also add a missing test for the `Structure` class.


markdewing:
okay to test

Issue #1888, opened 2019-09-10
Jastrow coefficients are frozen in optimization loops


tiihonej:
Since Aug 27 looping the Jastrow optimizer does not do anything, but the coefficients appear to be frozen. That is, the values remain the same on each loop iteration after the first. I'm attaching an archive with example input and output. I haven't tested thoroughly with different optimizers, but at least blocks of the following form seem affected:

`   <loop max="8">
      <qmc method="linear" move="pbyp" checkpoint="-1">
         <parameter name="warmupSteps"         >    300                </parameter>
         <parameter name="blocks"              >    200                </parameter>
         <parameter name="steps"               >    1                  </parameter>
         <parameter name="subSteps"            >    10                 </parameter>
         <parameter name="timestep"            >    0.3                </parameter>
         <parameter name="useDrift"            >    no                 </parameter>
         <parameter name="samples"             >    51200              </parameter>
         <parameter name="MinMethod"           >    OneShiftOnly         </parameter>
         <parameter name="minwalkers"          >    0.01               </parameter>
         <parameter name="nonlocalpp"          >    yes                </parameter>
      </qmc>
   </loop>`

[frozen_jastrow.zip](https://github.com/QMCPACK/qmcpack/files/3597754/frozen_jastrow.zip)

prckent:
This looks to be a real problem that we simply didn't notice in our regular testing:
<img width="799" alt="Screen Shot 2019-09-10 at 5 56 45 PM" src="https://user-images.githubusercontent.com/13683814/64653446-72063080-d3f4-11e9-8e2c-75db8e5682d3.png">

ye-luo:
#1830 didn't fully fix WF optimization.

PR #1937, opened 2019-09-24
Add deterministic test for Jastrow optimization

Hyeondeok-Shin:
Add deterministic test for Jastrow coefficient on the optimization process.
 This test is to check whether optimization process works normally via direct comparison of Jastrow coefficient with the reference. It will provide the easy way to find an issue in the optimization process (Ex. #1888).     

 He atom example is used as the test, and only one J2 coefficient is used as the test for simple test.

 The reference is generated with QMCPACK 3.6.0 with Intel18 compiler.

 Hyeondeok Shin


ye-luo:
Please remove qmc.xml cont.xml stat.h5 files.

Hyeondeok-Shin:
@ye-luo Yes, they are not necessary. Deleted.

tests/molecules/He_ae/check_opt.py
```diff
@@ -0,0 +1,116 @@
...
+    output = get_jub_b(prefix_file)
+    reference = get_jub_b(ref_file)
+
+    if len(output) != len(reference):
+       exit_fail('Number of coefficient in test({0}) does not match with the reference({1})'.format(len(output),len(reference)))
+
+    success = True
+   
+    for i in range(len(output)):
+       quant_success = float(output[i]) == float(reference[i])
```
jtkrogel: This floating point comparison should have a tolerance.  E.g. abs(float(output[i])-float(reference[i]))<1e-5.

Hyeondeok-Shin: @jtkrogel Yes, it's right. I added tolerance (1e-6) for the test, and now script prints tolerance and deviation from the reference.

prckent:
More comments to follow, but is this confirmed to pass with different compilers, recent versions etc.?

Hyeondeok-Shin:
@prckent 
1. Test was passed with QMCPACK develop version (gcc 6.2 and intel 18 on Intel Xeon) & QMCPACK 3.7 (BG/Q).
2. I'll add more commit for smaller optimization steps and the VMC run

Hyeondeok-Shin:
@prckent 

1. Now Optimization steps become smaller.
2. VMC run has added in the end, and deterministic test for the VMC energy has added to check whether optimized parameter is successfully stored.

Please let me know if additional changes are needed.

prckent:
OK to test

Issue #1897, opened 2019-09-11
IonOrbital need SoA


ye-luo:
https://cdash.qmcpack.org/CDash/testDetails.php?test=6114702&build=78458
Broken I believe since #1846.
When the code is build SoA but addTable requests AoS, the pointer to DT is nullptr and caused segmentation fault.
We need to add a check for SoA build and stop the code if AoS DT is requested.
@markdewing I assign you because you are more familiar about how to add the protection in createDT functions. 
@Paul-St-Young I assign you because you added the integration test and are more familiar with the class.

PS: I feel IonOrbital need to be renamed. The current name doesn't reflect its use.

Paul-St-Young:
OK, I implement SOA for this along with the lattice deviation estimator.
Regarding the name, how about `LatticeGaussianProduct`?
This wavefunction is a product of gaussians, each centered around a lattice site.

ye-luo:
Should be good. Unless other suggestions show up, just put what you suggested.

Issue #1845, opened 2019-08-28
Need unit test for kSpaceJastrow


ye-luo:
See discussion #1727

PR #1938, opened 2019-09-24
SOA versions of `latdev` and `ionwf`

Paul-St-Young:
Responding to #861 and #1897.

I added a better test for `LatticeDeviationEstimator`, then added SOA implementation for `LatticeDeviationEstimator` and `IonOrbital`.

### DIFF
--- a/tests/estimator/latdev/free/flatdev.py
+++ b/tests/estimator/latdev/free/flatdev.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+import h5py
+import numpy as np
+
+def corr(trace):
+  """ calculate the autocorrelation of a trace of scalar data
+
+  correlation time is defined as the integral of the auto-correlation
+   function from t=0 to when the function first reaches 0.
+
+  Args:
+    trace (list): should be a 1D iterable array of floating point numbers
+  Return:
+    float: correlation_time, the autocorrelation time of this trace of scalars
+  """
+  mu     = np.mean(trace)
+  stddev = np.std(trace, ddof=1)
+  if np.isclose(stddev, 0):  # easy case
+    return np.inf  # infinite correlation for constant trace
+  correlation_time = 0.
+  for k in range(1, len(trace)):
+    # calculate auto_correlation
+    auto_correlation = 0.0
+    num = len(trace)-k
+    auto_correlation = np.dot(trace[:num]-mu, trace[k:]-mu)
+    auto_correlation *= 1.0/(num*stddev**2)
+    if auto_correlation > 0:
+      correlation_time += auto_correlation
+    else:
+      break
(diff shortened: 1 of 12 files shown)


ye-luo:
Could you add unit tests for LatticeGaussianProduct?

Paul-St-Young:
@ye-luo I am willing to do it, but not right now. Can I add a unit test in a future PR?
I have included an integration test in this PR that runs in a second.

ye-luo:
@Paul-St-Young I think you are hitting your credit line. I found #1845 #1640.
Writing unit tests for LatticeGaussianProduct should be no more difficult than a Jastrow factor and it should take less than 1h. If you still insisting postponing the test, I can accept it if you create an issue. However, other reviewers may consider this not acceptable.

markdewing:
okay to test

tests/estimator/latdev/free/flatdev.py
```diff
@@ -0,0 +1,74 @@
...
+  zscore = abs(ym-expect)/ye
+  fail = np.any(zscore > 3.)
+  success = ~fail
+  return success, zscore
+
+if __name__ == '__main__':
+  # calculate <x^2> <y^2> <z^2>
+  nequil = 6
+  fh5 = 'mt.s000.stat.h5'
+  ym, ye = get_latdev_me(fh5, nequil)
```
markdewing: It is possible for qmca to read and compute the lattice deviations?
One goal would be to reduce the redundant code for computing autocorrelations and averages.
However, it would require more infrastructure for calling qmca and reading its output.  If possible, it's a change for a future PR.

ye-luo:
@Paul-St-Young @markdewing Please create an issue for the missing unit test.

Paul-St-Young:
@ye-luo I created #1945 to track of the unit test.
Sorry for being slow, I don't have a lot of bandwidth right now. I will make a point on my schedule to take care of #1640 and #1845 by October. I want to make unit tests for the absolute correctness of the code so it will take some time. Or do you just want to see simple regression tests?

tests/estimator/latdev/free/flatdev.py
Paul-St-Young: My impression was that `qmca` handles `scalar.dat` only, so I have always rolled my own analysis for `stat.h5`. This may have changed recently. Will keep in mind for a future PR

markdewing:
@Paul-St-Young  - Thanks for getting this change in.   I think having a test is more important than the specific form of the test (unit vs. integration).

ye-luo:
@Paul-St-Young Unit test for regression purpose should be good. You can leave a note next to the reference number by saying the numbers are not independently verified. We have quite a few unit tests made this way. Once you have time to independently verify the number, you just need to remove or update note.

PR #1939, opened 2019-09-25
Nexus: harden structure and physical system tests

jtkrogel:
This PR adds protection against scipy calls in some structure tests.  This was causing failures on CDash: https://cdash.qmcpack.org/CDash/testDetails.php?test=6227452&build=81102

Structure comparisons are also hardened.  If an atom is close to a box corner, tiling can result in different orderings of the final tiled atoms due to numerical sensitivity.  Fallbacks are added to check atomic species and position equality fully within the full minimum image convention.

### DIFF
--- a/nexus/bin/nxs-test
+++ b/nexus/bin/nxs-test
@@ -2593,7 +2593,7 @@ def structure():
 
     nunit('rwigner')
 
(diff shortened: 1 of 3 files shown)


markdewing:
okay to test

PR #1940, opened 2019-09-25
Nexus: add tests for nexus_base module

jtkrogel:
Add tests for the `nexus_base` module.  This PR depends on #1939.


markdewing:
okay to test

PR #1941, opened 2019-09-25
Nexus: QmcpackAnalyzer Warning

mcbennet:
Changing the initial value of "nblocks_exclude" to zero.
This suppresses the warning below when "nblocks_exclude"
is never re-assigned to a non-negative value.

  QmcpackAnalyzer warning:
    runtime exception encountered
      File "../nexus/lib/qmcpack_analyzer.py", line 482, in analyze
        QAanalyzer.analyze(self,force=force)
      File "../nexus/lib/qmcpack_analyzer_base.py", line 487, in analyze
        v.analyze(force)
      File "../nexus/lib/qmcpack_analyzer_base.py", line 494, in analyze
        self.analyze_local()
      File "../nexus/lib/qmcpack_result_analyzers.py", line 397, in analyze_local
        energies.append(dmc.scalars.LocalEnergy.mean)
    AttributeError: 'DmcAnalyzer' object has no attribute 'scalars'

Comment from Jaron regarding this:

When the user does not provide a value for nblocks_exclude, the analyzer is supposed to set the value by calling equilibration_length. ...

### DIFF
--- a/nexus/lib/qmcpack_method_analyzers.py
+++ b/nexus/lib/qmcpack_method_analyzers.py
@@ -79,7 +79,7 @@ class MethodAnalyzer(QAanalyzer):
             complete &= 'opt' in files
         #end if
         equil = request.equilibration
-        nblocks_exclude = -1
+        nblocks_exclude = 0
         if isinstance(equil,int):
             nblocks_exclude = equil
         elif isinstance(equil,(dict,obj)):


markdewing:
okay to test

markdewing:
A couple of questions (maybe more for @jtkrogel):
1. How did the situation arise that the analyzer did not call equilibration_length to set nblocks_exclude.   Should this situation be added as a regression test?
2. How does the code get from nblocks_exclude=-1 to the listed error message (no 'scalars' in DmcAnalyzer).    A brief look at the code seems to show that nblocks_exclude is used to extract a slice of the data (`sample[nblocks_exclude:]`), and so -1 for nblock_exclude will return an empty array.  Which I assume leads to nothing getting assigned to `scalars`?

jtkrogel:
Responding to question 1: I don't know.  Likely through a lack of testing in the past.  The QmcpackAnalyzer is too unwieldy to use/build upon in the future.  It will be deprecated in favor of a command line tool that may reuse some of its components.  In light of the coming deprecation, I think simply resetting the default value for nblocks_exclude is a reasonable stopgap.

mcbennet:
I have narrowed this down to the equilibration_length call.

`            try:
                nbe = equilibration_length(self.data[equilibration])
            except:
                print("An exception occurred")
`

results in the output of "An exception occurred". Given this, if the problem is stemming from numerics.py then I supposed we would want to patch that.

mcbennet:
The error is coming from within equilibration_length at the call to random.randint (imported from numpy) -- a print statement that I added after this call is never reached. Possibly there is a namespace conflict here, because when I try `import numpy as np` then use np.random.randint I see no exception

PR #1943, opened 2019-09-25
Nexus: update test for nexus imports

jtkrogel:
Be more explicit with the test for user facing imports from the `nexus` module.  This PR depends on #1940.


markdewing:
okay to test

PR #1947, opened 2019-09-26
Unified driver dmc batched

PDoakORNL:
First installment of DMCBatched driver.  Much the same pattern as the VMCBatched driver.  Unit tests to basic instantiation. Does not appear to break existing production code.  Debuggin in advanceWalkers.


PDoakORNL:
This is a good place to stop this PR as the next step is refactoring the branch engine and walker controllers which will be quite involved.

prckent:
Can you ("we") defer updating the branch engine and walker controllers or is this required to get DMC batched working and estimate a FOM + throughput vs system size?

PDoakORNL:
Can't be completely avoided due to heavy use of MCWalkerConfiguration in thos classes API's. Although often I find this whole mess of an object was passed in order to get access to just a few members in the inherited name space.

But I'm defering it as much as possible. My plan is a max speed refactor into new method functions that don't rely on DMCUpdate or MCWalkerConfiguration.

prckent:
Is the default walker count handling consistent with the old DMC drivers?

PDoakORNL:
Yes, but I think and A/B unit test is still in order.

prckent:
In the interest of development velocity and real consequential deadlines this needs to get merged. If no one comments within the hour that they are looking at this I will merge the PR ahead of tonight's nightlies.

prckent:
(Mark reports no comments)

PR #1948, opened 2019-09-27
L2 DMC

jtkrogel:
Putting up this WIP PR to gather feedback on the merge of new DMC projector features for L2 potentials into mainline.  I'm particularly interested in any issues spotted relative to #1947.

Please ignore the `DMCUpdatePbyPVMC` and `DMCUpdatePbyPL2VMC` classes.  These are development scaffolding and will be removed.  The main task is to merge `DMCUpdatePbyPL2` and `DMCUpdatePbyPWithRejectionFast` in a manner that is suitable for current mainline.

Other changes of note reside in `DriftOperators` and `QMCHamiltonian`.


markdewing:
What's an L2 potential?  Is the background published somewhere?

jtkrogel:
Sorry, this is a pseudo-Hamiltonian in the literature (but with the "a" term set to zero, keeping only the L^2, or "b", term).

See https://doi.org/10.1103/PhysRevLett.62.2088.

src/QMCDrivers/DMC/DMCUpdatePbyPL2.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+        // compute diffusion step 
+        H.computeL2D(W,iat,Dtmp);
+        D = Dtmp; // upcast for mixed precision
+        Dchol = cholesky(D);
+        dr_diff = dot(Dchol,dr_diff);
+        dr += sqrttau * dr_diff;
+
+        // reverse the intermediate drift move
+        W.rejectMove(iat);
```
PDoakORNL: How many of these early bail outs matter? How many just torture the branch prediction of the processor? 
Does it really make sense to check the move before its even scaled?

jtkrogel: Apart from the spurious `makeMoveAndCheck` at the beginning to ensure the temporary distances are populated, the bailout pattern is similar to the original DMC implementation (see code just below `!L2_proj`.  The check before scaling does not depend on the scaled drift in the original or the L2 algorithm.  I don't know the answer regarding necessity, but it should be the same for both algorithms.

PDoakORNL:
Main thing that I think is going to impact you is the Update layer is removed.  advanceWalker is replaced by a static function `advanceWalkers` See VMCBatched::advanceWalkers, DMCBatched::advanceWalkers.

They're a number of ways we could to deal with specialized updates, since in your case you're only overiding advanceWalkers it suggests that choosing an implementation for advanceWalkers at construction of the driver is sufficient.
```C++
class DMCBatched
{
...
using AdvanceWalkersFunction = std::function<void(const DMCBatched::StateForThread&, Crowd&,DriverTimers&,ContextForStep&,bool)>;
DMCBatched(...,advanceWalkers) : ... advanceWalkers_(advanceWalkers) {}
...
AdvanceWalkersFunction advanceWalkers_;
...
}

class DMCL2Advance {
void operator()(const DMCBatched::StateForThread& sft, Crowd& crowd, ContextForSteps& move_context, bool recompute)
{
...
}
```
I think its clear much of the advanceWalkers_ code is actually shared the specialized section is infact a subset of this, examination of different update code should make that clear in time.

jtkrogel:
This branch is now up to date with develop.  I'm going to add a deterministic test, clean up the code a bit, and then request a merge.

It will be good to have this functionality released prior to the move to batched drivers as it is a core technology for the Reboredo FWP.  Having tests, etc, merged will also facilitate the transition to batched drivers when they are ready.

jtkrogel:
Ready for merge.

ye-luo:
Okay to test

prckent:
@jtkrogel What are you thoughts on the mixed precision builds that are failing in the CI? Hopefully they are small fixes.

jtkrogel:
Mixed precision build issues should be resolved now.

ye-luo:
@jaron could you resolved the reorder warning. Basically the declaration order is not consistent with the constructor initialization order.

ye-luo (approved):
I fixed the GCC reorder warning. Get this in before the upcoming release.

Issue #1817, opened 2019-08-22
Segfault in sendWalkers


ashwinyes:
While trying NiO performance test with **Walkers=1** and **MPI Processes > 2** on a ThunderX2 ARM64 server, I was hitting a segfault in sendWalkers.

Digging deeper, looks like the cause for this is a wrong loop condition in the following line.
https://github.com/QMCPACK/qmcpack/blob/b7dc7acbde80cf19eb20f97c76b602664a9e9835/src/QMCDrivers/DMC/WalkerReconfigurationMPI.cpp#L119

The loop condition should have been **i <= maxIndex** .

With the current condition looks like all indices in **Zeta** is not getting filled and as a by product the  comparison **Zeta[ind] < tryp && Zeta[ind] >= wCur** in the subsequent while loop is not correctly done and **ind** index was going out of bounds and a segfault happens in the subsequent call to **sendWalkers**.

I am not a subject expert in QMCPACK. So I will leave it to the experts here to look into this and fix it.

ashwinyes:
I tried the below performance tests. All of them was showing the issue.

qmcpack-3.7.0/build/tests/performance/NiO/dmc-a4-e48-cpu/NiO-fcc-S1-dmc.xml
qmcpack-3.7.0/build/tests/performance/NiO/dmc-a8-e96-cpu/NiO-fcc-S2-dmc.xml
qmcpack-3.7.0/build/tests/performance/NiO/dmc-a16-e192-cpu/NiO-fcc-S4-dmc.xml
qmcpack-3.7.0/build/tests/performance/NiO/dmc-a32-e384-cpu/NiO-fcc-S8-dmc.xml
qmcpack-3.7.0/build/tests/performance/NiO/dmc-a64-e768-cpu/NiO-fcc-S16-dmc.xml

prckent:
@ye-luo Do you consider this fixed or worthy of a reinvestigation?

jtkrogel:
Never reproduced

PR #1952, opened 2019-09-28
Unified driver population control (partial)

PDoakORNL:
Traveling this weekend.  Want to get this in there before I get in car, more commits later.  Currently not compiling but I think the outline is clear. Expect to have in a running if not correct state soon.


src/QMCDrivers/MCPopulation.cpp
```diff
@@ -122,11 +136,49 @@ void MCPopulation::createWalkers(IndexType num_walkers)
...
+      createWalkerInplace(*it_walkers);
+      allocateWalkerStuffInplace(walker_index);
+      break;
+    }
+    ++walker_index;
+  }
+  // you need to add to all the vectors if no walkers have been killed
+  // for now we throw since this must be implemented
+  if (walker_index == walkers_.size())
+    throw std::runtime_error("Needed to add to population in MCPopulation but it is not implemented.");
```
prckent: Reminder to implement this before merging

prckent:
Note the reported bug https://github.com/QMCPACK/qmcpack/issues/1817

src/QMCDrivers/DMC/WalkerControlMPI.cpp
```diff
@@ -390,4 +447,163 @@ void WalkerControlMPI::swapWalkersSimple(MCWalkerConfiguration& W)
...
+  UPtrVector<MCPWalker> new_walkers;
+  std::vector<int> copies_of_each_new_walker;
+
+  // sort good walkers by the number of copies
+  std::vector<std::pair<int, std::reference_wrapper<MCPWalker>>> sorted_good_walkers;
+  for (int iw = 0; iw < ncopy_w.size(); iw++)
+    sorted_good_walkers.push_back(std::make_pair(adjust.copies_to_make[iw], adjust.good_walkers[iw]));
+  // Sort only on the number of copies
+  std::sort(sorted_good_walkers.begin(), sorted_good_walkers.end(),
+            [](auto& a, auto& b) { return a.first > b.first; });
```
ye-luo: I thought std::sort works with vector pairs by the first value.

src/QMCDrivers/DMC/WalkerControlMPI.cpp
```diff
@@ -390,4 +447,163 @@ void WalkerControlMPI::swapWalkersSimple(MCWalkerConfiguration& W)
...
+ *
+ *  2005 called and asked for its synchronous messaging back
+ *  I think MPI async is more than two years old
+ *
+ *  MCPopulation is sufficiently different from MCWalkerConfiguration that this is 
+ *  basically a rewrite.  I have not replicated the more ridiculous elements of the original.
+ *  For example:
+ *   * Sending duplicates to the same reciever rank but resorting good walkers if send-receive pair 
+ *     changes.
+ *   * Sending a tiny separate blocking message to tell the receiver how many copies
```
ye-luo: This is intended to avoid sending duplicates to the same receiver. The two items mentioned here are contradictory to each other.

src/QMCDrivers/DMC/WalkerControlMPI.cpp
```diff
@@ -390,4 +447,163 @@ void WalkerControlMPI::swapWalkersSimple(MCWalkerConfiguration& W)
...
+      WalkerMessage message = message_queue.front();
+      message_queue.pop();
+      send_requests.push_back(myComm->comm.isend_value(message.multiplicity, message.target_rank));
+      size_t byteSize    = message.walker.byteSize();
+      send_requests.push_back(myComm->comm.isend_n(message.walker.DataSet.data(), byteSize, message.target_rank));
+    }
+  }                                                    
+
+  std::vector<mpi3::request> receive_requests;
+  messages_incoming *= 2; // extra message sent for multiplicity
```
ye-luo: What is the magic number 2 here?

PDoakORNL: That was for the separate number of copies message, the next push to this PR will remove this.

src/QMCDrivers/DMC/WalkerControlMPI.cpp
PDoakORNL: https://en.cppreference.com/w/cpp/utility/pair/operator_cmp
TLDR
First compare lhs.first rhs.first if lhs.first == rhs.first then compare lhs.second, rhs.second.

I considered this oversorting since from my naive perspective sorting by walker IDs seemed if anything a potential bias.
However as concurrency, threaded and MPI, cause divergence this could add to it.

So while subtle this is a change.

ye-luo:
@prckent regarding your last point. The current production code does send out per-node weight and energy with allreduce. Because it is per-node and to avoid sending duplicated walkers, there is a small blocking near-neighbour send/receive to communicate info about walker duplication before non-blocking send/receive heavy walker data.

tests/solids/diamondC_2x1x1_pp/CMakeLists.txt
```diff
@@ -60,6 +60,15 @@ ENDIF()
                     1 DIAMOND2_DMC_SCALARS # DMC
                     )
 
+  QMC_RUN_AND_CHECK(short-diamondC_2x1x1_pp-dmcbatch_sdj
+                    "${CMAKE_SOURCE_DIR}/tests/solids/diamondC_2x1x1_pp"
+                    qmc_short
+                    qmc_short_dmcbatch.in.xml
```
ye-luo: This test uses the reference values of qmc_short_vmc_dmc.in.xml but input parameters are very different. Need a fix later.

PDoakORNL: The test is definitely invalid and I expect it to fail.  But it doesn't crash and produces scalar.dat output.  Population control is suppressed by forcing the no_branch call on WalkerControl.  Either before or after this PR is merged that is the next task for batched DMC.

PDoakORNL:
Their should be no significant change to the existing production code.  If the FOM can be run without population control this PR would be sufficient for DMCBatched. At this point I think we should review to merge for the small refactors to objects shared with the original drivers and limit divergence.  I agree that we should when "the dust settles" does a more holistic review of the Unified Driver.  The new population control code is not yet correct beyond the unit tests and I intend to be the subject of the next PR.

prckent:
Quick comment on the batched tests diamondC_2x1x1_pp, dmc & possibly also vmc. Needs to be fixed when the population control is fixed and the DMC batched driver is claimed to be correct:

The batched tests should be essentially exact copies of the existing short tests except with (presumably) better statistics so that the batching is meaningfully tested and there is still sufficient equilibration. The expected statistical error can be obtained from the data we already have for the long tests.

The current dmc_batch input specifies subSteps which is not meaningful in DMC. I would prefer that we ban the specification of this value in the DMC input block; at least values !=1 should be an abort since it signifies that the user does not know what they are doing :-)

PDoakORNL:
> specification of this value in the DMC input block; at least values !=1 should be an abort since it signifies that the user does not know what they are doing :-)

I will make the input and possibly check less rubbish when I figure out what is wrong with gcc and the new driver tests in the complex mixed precision case.

prckent:
Also happy for this to be merged as soon as the CI passes.

PDoakORNL:
The DMC substeps check did not make it in but I added it to my issues list.  Will be in next Unified Driver PR

PR #1953, opened 2019-09-30
Nexus: Add rotate and matrix_transform functions 

mcbennet:
This PR adds the function `rotate` to `structure` and allows the user to perform arbitrary rotations of the structure. Additionally, the function `matrix_transform` is added which allows the user to transform the structure in an arbitrary way (matrix need not be orthogonal).

Both `rotate` and `matrix_transform`, when given explicit matrices, are assumed to be in column-major form, i.e., they transform a vector v as v'=Tv


nexus/lib/structure.py
```diff
@@ -1457,21 +1551,149 @@ def stretch(self,s1,s2,s3):
...
-        kaxnew = dot(self.kaxes,inv(skew).T)
+        kaxnew = dot(self.kaxes,inv(A).T)
         self.pos     = dot(dot(self.pos,axinv),axnew)
         self.center  = dot(dot(self.center,axinv),axnew)
         self.kpoints = dot(dot(self.kpoints,kaxinv),kaxnew)
         self.axes  = axnew
         self.kaxes = kaxnew
         if self.folded_structure!=None:
-            self.folded_structure.skew(skew)
+            self.folded_structure.matrix_transform(A)
```
jtkrogel: Should this be `self.folded_structure.matrix_transform(A.T)` instead?

mcbennet: Yes, very good.

markdewing:
okay to test

markdewing:
@jtkrogel Are you okay with this PR?

jtkrogel:
Yes.

PR #1956, opened 2019-09-30
Nexus: fix use of random in equilibration_length

jtkrogel:
This is a small update to fix a namespace collision in the `equilibration_length` function in the `numerics` module.  A test has been added that failed prior to the fix.


markdewing:
okay to test

PR #1957, opened 2019-09-30
More mpi3 wrapper code copied from gitlab repo

markdewing:


prckent:
How close to parity will the MPI wrapper files be with the gitlab repo after this PR?

markdewing:
`communicator.hpp` needs to be synced.  I've tested it and it works.
`shared_communicator.hpp` and `shared_window.hpp` remain to by synced, and I think there will a design issue to be resolved regarding AFQMC and which side (AFQMC or the wrapper) provides some global (MPI) synchronization.

Issue #1946, opened 2019-09-26
PGI 19.7 compile-time error when build_type=RelWithDebInfo


naromero77:
@markdewing @prckent 

This occurs with QMCPACK 3.8.0 and PGI 19.7. The PGI_LOCALRC files is configured to use the GCC 6.2.0 using:
```
makelocalrc /soft/com/packages/pgi/19.7/linux86-64/19.7 -d /homes/naromero/ -gcc `which gcc` -gpp `which g++` -g77 `which gfortran` -x
```

I found that it was also necessary to add the following line to the PGI_LOCALRC file:
```switch -pthread is replace(-lpthread) positional(linker);```

Here is the problem: When `build_type=Release` which is the default, QMCPACK builds properly. However, `build_type=RelWithDebInfo` (which adds the `-g`) flag results in the following error:

```
     5618    "/nfs/software/linux-ubuntu_trusty_amd64/apps/packages/gcc/gcc-6.2.0/include/c+
  >> 5619    +/6.2.0/bitset", line 76: internal error: write_name_entry_to_file:
     5620              inconsistant number of debug entries
     5621    
     5622            _WordT         _M_w[_Nw];
     5623                               ^
     5624    
     5625    1 catastrophic error detected in the compilation of "/sandbox/naromero/tmp/spack-stage/qmcpack-3.8.0-sb4gezmijcge5lddcb5qlahoyjc5
             irvw/spack-src/src/QMCHamiltonians/tests/test_bare_kinetic.cpp".
     5626    Compilation aborted.
     5627    pgc++-Fatal-/nfs/software/linux-ubuntu_trusty_amd64/com/packages/pgi/19.7/linux86-64-llvm/19.7/bin/pggpp1-llvm TERMINATED by sign
             al 6
     5628    Arguments to /nfs/software/linux-ubuntu_trusty_amd64/com/packages/pgi/19.7/linux86-64-llvm/19.7/bin/pggpp1-llvm
     5629    /nfs/software/linux-ubuntu_trusty_amd64/com/packages/pgi/19.7/linux86-64-llvm/19.7/bin/pggpp1-llvm --llalign -Dunix -D__unix -D__
             unix__ -Dlinux -D__linux -D__linux__ -D__NO_MATH_INLINES -D__LP64__ -D__x86_64 -D__x86_64__ -D__LONG_MAX__=9223372036854775807L '
[...]

PR #1958, opened 2019-09-30
Nexus: Initialization of nblocks_exclude

mcbennet:
Setting initialization value to -1 (as it was before #1941),
given that #1956 resolved the underlying issue of the warning.


markdewing:
I'm a little concerned about reverting this, given how it failed last time.
It looks like -1 is used as a sentinel to indicate that nblocks_exclude was not set from the command line(?).   But it's also a valid array index.
Would an initial value of 'None' be better?  That would at least trigger an error at the point of use if the variable was not set differently.

jtkrogel:
I believe `nblocks_exclude` is set by calling `equlibration_length` in the event it is not set by the user.  See `DatAnalyzer.__init__` in `qmcpack_quantity_analyzers`.  If I recall correctly, this was the call that was originally problematic in #1946 

If desired, one could back this up by adding `assert(nbe>=0)` in a number of places where `nblocks_exclude` is accessed and used.

mcbennet:
The initial value of `nblocks_exclude` was not the cause of the failure that was reported in #1941 .
By chance I originally changed its initial value to zero and saw the the warning was suppressed, which led me to believe that the variable wasn't being set correctly after initialization. However, after further digging, it has been found that the code behaves correctly (in the case of `nblocks_exclude=-1`) and attempts to set `nblocks_exclude` through `equilibration_length` -- however, within `equilibration_length` the code was failing when attempting to call `random.randint` due to a scoping issue.

markdewing:
Maybe at least check nbe after it should be set in DatAnalyzer.__init__

prckent:
I await further comment here before merging

markdewing:
okay to test

mcbennet:
In hindsight, it appears that the assertion after `DatAnalyzer.__init__` is not needed.
`equilibration_length` can not return a value that is non-negative.

jtkrogel:
I think this is fine.  At this point we are speculating about the potential for buggy behavior.  I suggest we merge this PR and revisit if any such behavior is observed down the road.

markdewing:
To be more general, the assertion could be moved to after the if-block.  That way it would catch the case that `equilibration` is `None`.   In the language of contracts, this is ensuring a postcondition about nblocks_exclude after `DatAnalyzer.__init__` executes.

PR #1959, opened 2019-09-30
oxygen jenkins exits on configure/build failure

grahamlopez:
This should prevent the false-positive like we got with #1947


grahamlopez:
Erm, I guess I mean false negatives.

PR #1963, opened 2019-10-01
Small warnings

grahamlopez:
Remove some of the most trivial warnings from non-MPI gcc 7.4 build:
- clang-formatted all of `formic/utils/mantissa_exponent.h` to address `-Wmisleading-indentation`
- trivial changes for `-Wcomment`
- variable length arrays in einspline. I introduced a few raw `new [] / delete[]` pairs. Since einspline is kind of a "we don't touch it" situation IIUC, on top of being very c-like already, this seemed like the least intrusive route.


src/QMCWaveFunctions/tests/test_einset.cpp
```diff
@@ -112,9 +112,6 @@ TEST_CASE("Einspline SPO from HDF", "[wavefunction]")
 </tmp> \
 ";
 
-  // monoO
```
ye-luo: @markdewing Do we still need monoO? You probably has the best knowledge since you added these lines when this test was created. There are more monoO related lines although we don't have any value checking in the file. If we decide to delete monoO, I can take care of the rest since I'm working on this file to add more spline unit tests.

ye-luo:
@grahamlopez is -Wcomment turned on by default with gcc?

grahamlopez:
@ye-luo none of these are, except for `-Wvla`. I am (slowly) trying to work towards being able to turn on `-Wall`. Many more to come.

src/QMCWaveFunctions/tests/test_einset.cpp
markdewing: The monoO lines can be removed.

PR #1965, opened 2019-10-02
Add a completion message to pyscf_to_afqmc.py

jtkrogel:
The converter now prints `# Finished.` when it has completed.  This will make it easier for users and Nexus alike to catch unsuccessful execution.


markdewing:
okay to test

PR #1969, opened 2019-10-02
compiler warnings 2 of N: -Wmaybe-uninitialized

grahamlopez:
Ideally, these fixes would rework the logic to better use scoping and RAII for added safety, instead of depending on all cases being handled correctly by several `if` - `else if` - `else` blocks. I did make some simple changes like this, but when things got too complex or it seemed reasonable to do so, I applied the band-aid and moved on. I also didn't spend a lot of time in sections of the code that need larger consideration anyway (e.g. xml/builders).

### DIFF
--- a/src/QMCDrivers/QMCLinearOptimize.cpp
+++ b/src/QMCDrivers/QMCLinearOptimize.cpp
@@ -645,7 +645,7 @@ QMCLinearOptimize::RealType QMCLinearOptimize::getSplitEigenvectors(int first,
                                                                     bool& CSF_scaled)
 {
   std::vector<RealType> GEVSplitDirection(N, 0);
-  RealType returnValue;
+  RealType returnValue = 0;
   int N_nonlin = last - first;
   int N_lin    = N - N_nonlin - 1;
   //  matrices are one larger than parameter sets
--- a/src/QMCDrivers/QMCLinearOptimize.cpp
+++ b/src/QMCDrivers/QMCLinearOptimize.cpp
@@ -645,7 +645,6 @@ QMCLinearOptimize::RealType QMCLinearOptimize::getSplitEigenvectors(int first,
                                                                     bool& CSF_scaled)
 {
   std::vector<RealType> GEVSplitDirection(N, 0);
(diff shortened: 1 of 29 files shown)


markdewing:
Did this PR also run the code through clang-format?   It makes it hard to find the places where the code actually changed.

grahamlopez:
No, but looks like my editor got happy with correcting trailing whitespaces. I guess we still haven't formatted all of the code yet?

ye-luo:
What is the difference between `-Wmaybe-uninitialized` and `-Wuninitialized`?

grahamlopez:
the man page explains it better than I could here. FWIW, there are no `-Wuninitialized` warnings in the PR either.

src/QMCDrivers/QMCLinearOptimize.cpp
```diff
@@ -746,9 +746,8 @@ QMCLinearOptimize::RealType QMCLinearOptimize::getSplitEigenvectors(int first,
       CSF_scaled = true;
     else
       CSF_scaled = false;
-    returnValue = std::min(lowest_J_EV, lowest_CSF_EV);
+    return std::min(lowest_J_EV, lowest_CSF_EV);
   }
```
prckent: What should the return value be if CSF_Option=="TypO!" ?

grahamlopez: This is an unhandled case both before and after this PR. Maybe someone else can comment what should happen - `APP_ABORT()`?

prckent: I'll vote for that. (  Really, this points to a design problem: CSF_Option should be an enum or similar by this point so that no unknown setting is possible. Input handling would do the validation. Sadly I think there are a bunch of these throughout the code. )

grahamlopez: You are correct. I seem to remember several places where things are set in conditionals, with no base case fall through handling. As I said, I also shied away from going deeply into input/option handling since this has generated contentious discussion in the past, and would take quite some effort besides.

Issue #1945, opened 2019-09-26
Unit test needed for GaussianLatticeProduct


Paul-St-Young:
See #1938

Paul-St-Young:
closed by #1972

PR #1972, opened 2019-10-03
add unit test for LatticeGaussianProduct

Paul-St-Young:
Address #1945, I tested AOS and SOA on the CPU.

I think #1897 can be closed once this is merged.


Paul-St-Young:
Quick question: is there an easy way to `TrialWavefunction::getComponent` after `TrialWavefunction::addComponent`?

I want to directly verify the internal parameters in this wavefunction component, but cannot access it in the unit test because the builder directly puts this component into `TrialWavefunction`.

ye-luo:
We don't have a simple way to access individual WaveFunctionComponent(WFC) from TrialWaveFunction::Z[]. Nor do I think it is the right way to do. I looked at the WaveFunctionComponentBuilder which takes TrialWaveFunction(TWF) as one of the constructor argument and hold a reference inside. I don't see a clear a logic why it needs to get entangled.
I also see the notorious put() function. I'm consider to change bool WaveFunctionComponentBuilder::put(xml) to WFC* buildWaveFunctionComponent(xml) and get ridge of TWF entanglement. A derived class implements Jastrow* buildWaveFunctionComponent(xml) where Jastrow is a derived class of WFC. Once this is done, you can directly access the WFC without going through TWF.

ye-luo:
Okay to test

src/QMCWaveFunctions/tests/test_lattice_gaussian.cpp
```diff
@@ -0,0 +1,135 @@
...
+  xmlNodePtr jas1 = xmlFirstElementChild(root);
+
+  LatticeGaussianProductBuilder jastrow(elec_, psi, pp);
+  jastrow.put(jas1);
+  double width = 0.5;
+  double alpha = 1./(2*width*width);
+  //// check initialization. Nope, cannot access Psi.Z
+  //for (int i=0; i<2; i++)
+  //{
+  //  REQUIRE(psi.Z[0]->ParticleAlpha[i] == alpha);
```
markdewing: This will be checked by the wavefunction value below.   It's not as direct, but at least will get checked.

grahamlopez:
test this please

PR #1975, opened 2019-10-04
Avoid He opt deterministic test in CUDA build

Hyeondeok-Shin:
Because He_ae test is not supported on CUDA build, its deterministic test is always failing on CUDA build on cdash. So, He_ae opt deterministic test needs to be skipped on CUDA build.

Hyeondeok Shin


markdewing:
okay to test

prckent:
Please keep on cleaning the failing/unsupported tests like this. Thanks.

PR #1976, opened 2019-10-05
Updates for AFQMCTOOLs 

fdmalone:
Some updates for afqmctools:

1. Address unsafe file opening from converter.
2. Add __init__.py to test directories to enable unittest discover option.
3. Fix issue with write_xml_input for new observable input structure. I added a unit test to check for correctness.


markdewing:
okay to test

Issue #1979, opened 2019-10-08
scalar.h5 should be stat.h5


jtkrogel:
The AFQMC code currently writes heavy observables, like the 1RDM, into a `scalar.h5` file.  This is in addition to the `scalar.dat` file that actually contains scalars.

The double naming (scalar.h5 and scalar.dat) is confusing.  Additionally, the naming is confusing because the real space code uses the distinctly named `stat.h5` file for the type of heavy observables `scalar.h5` contains.

The `scalar.h5` file from the AFQMC code should be renamed `stat.h5` to standardize outputs and avoid confusion.

prckent:
Agreed.

prckent:
Also, the 1DRM is not a scalar...

fdmalone:
Ok. It's not really clear that 'stat' is an obvious name for things other than 'scalars' though is it?

jtkrogel:
Resolved by #1980.

PR #1980, opened 2019-10-08
Update HDF5 filename for heavy observables.

fdmalone:
Performs scalar.h5 -> stat.h5 change requested in #1979.


markdewing:
okay to test

PR #1981, opened 2019-10-08
compiler warnings 3 of N: -Wreorder

grahamlopez:
This PR partially addresses `-Wreorder` warnings. It covers somewhat over half of the compilation according to CMake. I will submit a follow-up to get the rest, and then enable the warnings by default.


prckent:
Thanks Graham. Happy to merge (sitting with @ye-luo  and @PDoakORNL ) once the CI passes.
We should merge soon so that conflicts do not build up.

PR #1992, opened 2019-10-10
Periodic gaussian complex k-point support

### LINKED ISSUE #1988: Periodic gaussian complex k-point support
This release to include the now-working but non-optimized support for arbitrary k-points. A PR is in preparation. The implementation will be optimized and improved at a later date.

anbenali:
Closes #1988 

This PR shows the implementation of evaluation of complex wavefunction with LCAO at any arbitrary Kpoint. 
Tiling to any shape of wavefunction is handled by converter (already implemented from prior versions) and working for any shape. 
      * Correct tiling from kpoint  and supertwist is handled by Nexus following QMCPACK school of 2019 example using Nexus. 

Provided 3 tests (short and long): 
  1- Carbon Diamond 1x1x1 at kpoint (1/3 1/3 1/3)
  2- Aluminum fcc 1x1x1 at random kpoint
  3 - Carbon Diamond 2x1x1 at random kpoint from tiling. 

All WF from Hartree Fock trial wavefunction and no Jastrow functions. The point is to compare directly to the HF wavefunction. 

Deterministic test will be added in a following PR by @mcbennet .


markdewing:
okay to test

prckent:
Please also provide a short test. The long ones are only run at weekends for selected builds and sometimes not even then.

Testing at a very low symmetry k-point (0.11,0.23,-0.34) is helpful but not critical. It must take a while to get good statistics compared to spline runs.

src/QMCWaveFunctions/lcao/SoaAtomicBasisSet.h
```diff
@@ -146,6 +146,17 @@ struct SoaAtomicBasisSet
     T r_new;
     // T psi_new, dpsi_x_new, dpsi_y_new, dpsi_z_new,d2psi_new;
 
+    PosType Tvec;
```
rcclay: What is this variable?  Comment please

mcbennet: Tvec is a translation vector. gendisp is the difference between the ion position and the electron position, while dr is the difference between the nearest image ion (relative to electron) and the electron position.

In order to reduce the number of images that need to be summed over when generating the AO,  Tvec is constructed such that it can be subtracted from the electron position to bring the electron closer to the primitive cell (Tvec is then necessarily added to the phase afterwards).
As it is coded, it actually brings the electron into a box defined by x,y,z-projections of the sum of the prim lattice vectors. This was easier to implement than bring the electron strictly into the primitive cell and also seemed to sufficiently bring down the number of images.

Tvec is not needed if there is already something floating around that tells us which cell the electron is in -- I was not sure if there was anything like this already.

prckent: This explanation can be put in comments

anbenali: addressed

src/QMCWaveFunctions/lcao/SoaAtomicBasisSet.h
```diff
@@ -135,7 +138,7 @@ struct SoaAtomicBasisSet
 
 
   template<typename LAT, typename T, typename PosType, typename VGL>
-  inline void evaluateVGL(const LAT& lattice, const T r, const PosType& dr, const size_t offset, VGL& vgl)
+  inline void evaluateVGL(const LAT& lattice, const T r, const PosType& dr, const size_t offset, VGL& vgl,std::vector <double> gendisp)
```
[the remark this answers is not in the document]
anbenali: Addressed and modified into PosType

anbenali:
Please do not merge yet As I might have found a case where it breaks (will be part of the tests)

ye-luo:
Once you think the PR is ready to go, please remove the WIP.

anbenali:
@ye-luo  @rcclay : This is ready to test (and merge) when approved.

anbenali:
The changes are small and non-invasive. Can we process this faster as I need to start the second PR on complex coefficients for the multideterminants in order to finish the paper.

anbenali:
> While fixing the merge problem, please update the periodic PySCF section in the manual. A minimum two sentence change would be to state that the implementation is general and indicate if there are any significant speed advantages to a gamma wavefunction.

@prckent : Addressed!

prckent:
The requested fixed appear to have been done. Unless any reviewer comments before, I think OK to merge once the CI passes & any conflicts are fixed.

rcclay:
I’m happy to have it  merged, once merge conflict is fixed.

anbenali:
@prckent @ye-luo @markdewing  I fixed the conflict. Anyone can merge now :)

ye-luo:
@anbenali The Al short test has the same input file as the long test, please have a look.
In later PR, we should have unit test which is even more important and useful than a deterministic test.

anbenali:
> @anbenali The Al short test has the same input file as the long test, please have a look.
> In later PR, we should have unit test which is even more important and useful than a deterministic test.

@ye-luo Fixed.. somehow busted... And yes, @mcbennet will add a deterministic test.

mcbennet:
> In later PR, we should have unit test which is even more important and useful than a deterministic test.

Hi @ye-luo, can you expand a bit more about what you are suggesting here?

prckent:
@mcbennet A unit test ideally invokes the smallest part of the code needed to test the functionality added here; ideally a single unit of the code. We have spline unit tests for example. These have the advantage of running instantly and conclusively point to a problem in the splines if there is one. This makes it dramatically easier for someone new to the project or unfamiliar with the methodology to contribute. A test program (unit test) that builds a very few orbitals from a very few basis functions would be helpful. We have noticed that an inability to write these programs without pulling in a large amount of the rest of QMCPACK highlights design issues where refactoring is needed...

PR #1993, opened 2019-10-10
compiler warnings 4 of N: -Wunused-variable (part 1)

grahamlopez:
This PR makes a first pass at removing the trivial `-Wunused-variable` warnings. Essentially, I removed all of them that "obviously" don't have side effects. Whoever reviews this should just check that they agree with me.

There are still a ton of warnings due to unhandled mpi, hdf5, and xml return codes. I will make another pass to generically handle these return values. I expect there to be a few more that should have made it to this PR, but got swamped by the noise of this second class.


grahamlopez:
sorry ya'll. too many build configurations for me to check. hang tight and I'll have a fix

src/QMCWaveFunctions/Fermion/BackflowTransformation.h
```diff
@@ -432,7 +432,6 @@ class BackflowTransformation //: public OrbitalSetTraits<QMCTraits::ValueType>
     activeParticle = iat;
     for (int i = 0; i < NumTargets; i++)
       oldQP[i] = newQP[i] = QP.R[i];
-    const auto& myTable = P.getDistTable(myTableIndex_);
```
markdewing: This looks like it won't compile with AoS.   This statement should be moved inside the #ifndef ENABLE_SOA

src/QMCWaveFunctions/Fermion/BackflowTransformation.h
```diff
@@ -740,7 +739,6 @@ class BackflowTransformation //: public OrbitalSetTraits<QMCTraits::ValueType>
       dr[1] = 0.05;
       dr[2] = -0.3;
       P.makeMove(iat, dr);
-      const auto& myTable = P.getDistTable(myTableIndex_);
```
markdewing: Same issue with AoS build

src/QMCWaveFunctions/Fermion/BackflowTransformation.h
grahamlopez: yep, exactly why CI failed. I didn't check AoS

src/QMCWaveFunctions/Fermion/BackflowTransformation.h
grahamlopez: should be fixed now and ready for review

PDoakORNL:
I really didn't think there were so many of these.

Issue #1983, opened 2019-10-09
Global Random found Harmful


PDoakORNL:
So 
```C++
NonLocalECPotential::NonLocalECPotential(ParticleSet& ions,
                                         ParticleSet& els,
                                         TrialWaveFunction& psi,
                                         bool computeForces,
                                         bool useVP)
    : IonConfig(ions),
      Psi(psi),
      UseTMove(TMOVE_OFF),
      myRNG(&Random),
```
Random is a global object and not threadsafe.  In the new DMCBatched driver NonLocalECPPotential::evaluate is called in a threaded context.  It does this
```C++
   PPset[ipp]->randomize_grid(*myRNG);
```
This rather randomly causes SIGSEGV

This took me quite some time to catch in a core dump. Enough with the global objects and unclear scopes.

jtkrogel:
What are the implications for current use of random numbers for different threads?

ye-luo:
The global RNG recorded in the constructor is only used by the CUDA code.
The CPU code path overwrites the recorded one by the per-thread RNG at a later stage by calling setRandomGenerator. So there is no problem in the production code.

prckent:
...and the CUDA code does not do threaded execution, so production use is safe even if ugly.

PDoakORNL:
I still have a segfault. CoreDump points here.

PDoakORNL:
This is only surfaced for the DMCBatched, I not implying current code isn't a special case where this somehow works.  But I'm not participating in further development impacted by this until I fix it and I don't much care about the state manipulation that makes it work in the legacy code.

ye-luo:
Yes. This is a valid bug report. We probably need to find a clean way to deal with per walker RNG need.

PR #1997, opened 2019-10-11
[WIP] Thread safety for global Random object.

PDoakORNL:
The additional concurrency in the batched drivers revealed a possible race on Random.  A surprising number of calls to Random() are sprinkled through the code despite it looking at first glance like we use one carefully seeded RNG per thread.


PDoakORNL:
DMCBatched no longer frequently crashes
patches but doesn't resolve #1983

prckent:
Is there any change to the random utilization in the existing omp based qmc drivers with this update or were they always successfully using distinct random sequences in a thread safe manner? (I think yes, but want to confirm).

PDoakORNL:
We need to look at the call stack at the time each of the Random() calls was made. I'll figure out how to get a tool to spit that out. My guess is there is not a race on it in the old drivers/ single evaluators since a rather large threads have been frequently run and this sort of crash not observed. The global random is init, saved and loaded along with the per thread local ones even though it is global instead of owned by the RandomNumberCotrol, so it was "part" of the simulation. My earlier alarm there is unfounded.

PDoakORNL:
Spoke to soon its still broken, I neglected to remember that access was through a pointer to base in NonLocalECPotential which is where it matters.

PDoakORNL:
This has gotten rolled in with my next PR as some of the issue was global but there was a specific bug related to the batched drivers with the NonLocalECPotentials  myRng

PR #1998, opened 2019-10-11
Fix batch NLPP evaluation in DMC and use thread local random number sequences

PDoakORNL:
So NLPP evaluation is now happening for DMCBatched and in a thread safe manner. The setNonLocalMoveHandler is probably just a temporary hack but I think it beats setting up an inheritance relationship.  I think by restructuring some calls across QMCMain, QMCDriverNew, and the specific algorithm drivers it can probably be eliminated.


src/Utilities/RandomGenerator.h
```diff
@@ -99,16 +98,61 @@ extern RandomGenerator_t Random;
...
 } // namespace qmcplusplus
 #endif
 #endif
 #endif
+
+
+namespace qmcplusplus
+{
+
+class RNGThreadSafe : public RandomGenerator_t
```
ye-luo: Seems not directly related to this PR.

src/QMCHamiltonians/NonLocalECPotential.h
```diff
@@ -60,6 +61,16 @@ class NonLocalECPotential : public OperatorBase, public ForceBase
    */
   void setNonLocalMoves(xmlNodePtr cur) { UseTMove = nonLocalOps.put(cur); }
 
+  void setNonLocalMoves(const std::string& non_local_move_option,
+                                        const double tau,
+                                        const double alpha,
+                                        const double gamma)
+  {
+    UseTMove = nonLocalOps.thingsThatShouldBeInMyConstructor(non_local_move_option,
```
ye-luo: All the Tmove handling changes can be in separate PR.

src/QMCDrivers/DMC/DMCBatched.cpp
```diff
@@ -51,12 +62,19 @@ QMCTraits::IndexType DMCBatched::calc_default_local_walkers(IndexType walkers_pe
   return local_walkers;
 }
 
+void DMCBatched::setNonLocalMoveHandler(QMCHamiltonian& golden_hamiltonian)
+{
+  golden_hamiltonian.setNonLocalMoves(dmcdriver_input_.get_non_local_move(), qmcdriver_input_.get_tau(),
```
ye-luo: Not understanding why only golden_hamiltonian gets set. This breaks the requirement that walker in a crowd must have identical state.

src/QMCDrivers/DMC/DMCBatched.cpp
```diff
@@ -417,10 +437,16 @@ void DMCBatched::runDMCStep(int crowd_id,
...
+                            UPtrVector<ContextForSteps>& context_for_steps,
+                            UPtrVector<Crowd>& crowds)
 {
-  Crowd& crowd  = *(crowds[crowd_id]);
+  Crowd& crowd                         = *(crowds[crowd_id]);
+  auto setThreadLocalRNGForHamiltonian = [](QMCHamiltonian& ham, ContextForSteps& step_context) {
+    ham.setRandomGenerator(&(step_context.get_random_gen()));
+  };
+  for (int iw = 0; iw < crowd.size(); ++iw)
+    setThreadLocalRNGForHamiltonian(crowd.get_walker_hamiltonians()[iw], *(context_for_steps[iw]));
```
ye-luo: Can we move this added part into crowd as a member function? This part has been copied many times.

src/QMCHamiltonians/QMCHamiltonian.h
```diff
@@ -271,12 +271,23 @@ class QMCHamiltonian
   /** set non local moves options
    * @param cur the xml input
    */
-  void setNonLocalMoves(xmlNodePtr cur)
+  [[deprecated]] void setNonLocalMoves(xmlNodePtr cur)
```
ye-luo: I'm not sure if we are ready to expose this type of warning to users.

src/QMCDrivers/DMC/DMCBatched.cpp
ye-luo: `context_for_steps[iw]` or `context_for_steps[crowd_id]`?
`crowd.setRNGForHamiltonian(context_for_steps[crowd_id]->get_random_gen());`

ye-luo: I really need to try this branch so just pushed the bug fix.

ye-luo:
I tried the branch. It made the NLPP evaluation using the thread local RNG. However, the DMC batched driver execution is still not quite right thus still wrong timing compared to the production driver.  Please update the PR title to reflect its actually change.

PDoakORNL:
I will have a chance to look at this again tomorrow morning

src/QMCDrivers/DMC/DMCBatched.cpp
```diff
@@ -27,7 +30,15 @@ DMCBatched::DMCBatched(QMCDriverInput&& qmcdriver_input,
                        QMCHamiltonian& h,
                        WaveFunctionPool& wf_pool,
                        Communicate* comm)
-    : QMCDriverNew(std::move(qmcdriver_input), pop, psi, h, wf_pool, "DMCBatched::", comm), dmcdriver_input_(input)
+    : QMCDriverNew(std::move(qmcdriver_input),
```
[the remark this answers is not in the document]
PDoakORNL: I agree, this is a preview of the behavior of clang-format in 9.x I'll protect it with `// clang-format off|on` but well its a bummer.  When I have time I'll look into whether we can tune it without breaking 7.x and 8.x clang-format.

prckent:
Everything works now :-) and the branch is current. Can we merge?

ye-luo:
LGTM

prckent:
Mark confirm no objections, so merging.

PR #1999, opened 2019-10-11
Update references for mixed-precision deterministic test

Hyeondeok-Shin:
So far, references for mixed-precision were same as the full-precision, so it was hard to guess statistics for the mixed-precision and some deterministic tests were failed in mixed precision build.

 To solve this problem, references for mixed precision deterministic test are updated as the average of observables estimated from various environments in order to capture variety of statistics in mixed-precision with different compilers or cpus.

 Here is list of environment used when averaging mixed-precision results.

 1. Non-CUDA
   - Intel Xeon - gcc6.2.0, Intel Xeon - intel18, Intel Xeon - clang7, cetus - clang, bora - intel18, hyperion - gcc, hyperion - intel18, oxygen - intel18

 2. CUDA
   - bora, summit, and oxygen

 All tolerances in mixed precision tests are reduced, and it will resolve a several test failure on mixed precision build on cdash. But CUDA build on oxygen will frequently have some failure in the test because localECP is strongly fluctuating with the CUDA in oxygen. I'm not sure why but it needs to keep eyes open with it.
 
 Hyeondeok Shin


prckent:
This update highlights the need to have automated methods to regenerate all this reference data. We will have many more run variants to support long term.

Since the deterministic tests check for consistency (vs correctness, which needs a long statistical run), it is OK to have different values between full and mixed precision. However the difference  again highlights that the mixed and full precision builds really do behave differently.

prckent:
OK to test

PR #2001, opened 2019-10-14
Update oxygen test configuration and fix setting passing for ctest

### LINKED ISSUE #1858: Clang6-CUDA-Full-Release build on Oxygen is not under full-precision.
Clang6-CUDA-Full-Release build on Oxygen is not under full-precision without -D QMC_MIXED_PRECISION=0 flag.

https://cdash.qmcpack.org/CDash/buildSummary.php?buildid=76347

-- defining the float point precision
   Base precision = double
   Full precision = double
   CUDA base precision = float
   CUDA full precision = double

 This build is using exactly same build command with Clang6-CUDA-Release build for the mixed precision (https://cdash.qmcpack.org/CDash/buildSummary.php?buildid=76336), so it needs to be checked.

  Hyeondeok Shin

### LINKED ISSUE #1962: Oxygen nighly AoS labeled builds are actually SoA
`-DENABLE_SOA=0` doesn't get added but the script in the repo seems to be able to add it.
Affecting GCC8-MKL-Complex-AoS-Release and GCC8-MKL-AoS-Release builds

prckent:
Closes #1962 
Closes #1858 

These settings bugs occurred because our ctest_script would in general not pass zero valued configuration flags through to cmake. With changes to our defaults this is now required.

Oxygen integration test and setup scripts are updated following the two year rule in readiness for new release. These need to run before additional fixes or improvements can be assessed. CUDA will need more work, but it could be that develop is broken with the compiler and library versions we are using.


CMake/ctest_script.cmake
```diff
@@ -238,16 +238,25 @@ IF ( QMC_COMPLEX )
   SET( CTEST_OPTIONS "${CTEST_OPTIONS};-DQMC_COMPLEX='${QMC_COMPLEX}'" )
 ENDIF()
 
+#Always explicitly enable/disable MPI in testing
 IF ( DEFINED QMC_MPI )
   SET( CTEST_OPTIONS "${CTEST_OPTIONS};-DQMC_MPI='${QMC_MPI}'" )
+ELSE()
+  SET( CTEST_OPTIONS "${CTEST_OPTIONS};-DQMC_MPI=0" )
```
ye-luo: This breaks other nightly testing scripts without specifying QMC_MPI and this is the opposite of our default.

CMake/ctest_script.cmake
```diff
@@ -238,16 +238,25 @@ IF ( QMC_COMPLEX )
...
   SET( CTEST_OPTIONS "${CTEST_OPTIONS};-DQMC_MIXED_PRECISION='${QMC_MIXED_PRECISION}'" )
+ELSE()
+  SET( CTEST_OPTIONS "${CTEST_OPTIONS};-DQMC_MIXED_PRECISION=0" )
 ENDIF()
 
+#Always explicitly enable/disable SoA in testing
 IF ( ENABLE_SOA )
   SET( CTEST_OPTIONS "${CTEST_OPTIONS};-DENABLE_SOA='${ENABLE_SOA}'" )
+ELSE()
+  SET( CTEST_OPTIONS "${CTEST_OPTIONS};-DENABLE_SOA=0" )
```
ye-luo: This is the opposite of our default.

CMake/ctest_script.cmake
prckent: ah yes, we need "if (defined var ) {} else { set default }" assuming a value =0 counts as defined in cmake speak

prckent: Fix pushed

PR #2004, opened 2019-10-15
Update ctest_script

ye-luo:
1. Put most flags under `IF ( DEFINED QMC_FLAG )`
2. Remove host specific configurations.
3. If a flag is a string, add single quotes.


prckent:
@PDoakORNL Are any ctest_scripts updates relevant to you? This should only affect machines reporting to cdash.

PDoakORNL:
LGTM

PR #2005, opened 2019-10-15
missing includes for NonLocalECPotential_CUDA

PDoakORNL:
It would appear that previously the CUDA implementation of the NonLocalECPotential recieved the declaration of NLjob and NonLocalECPComponent through some torturous include chain that was broken by a recent PR. This corrects that issue.


prckent:
With the DNS issue causing the ALCF test problem, can this be safely merged?

My vote is yes given the nature of the change.

PR #2008, opened 2019-10-16
Enable legacy cuda nightlies on summit

grahamlopez:
I think this is close, but I'm not sure yet that this is working completely right. I would like some input from anyone more familiar with how the nightlies/cdash works.

Another thing that is needed: actually building the code on the nodes via jsrun. There are cgroup limits on the login/batch nodes now, and once a load/memory threshold is crossed, that session gets severely throttled.

https://docs.olcf.ornl.gov/systems/summit_user_guide.html?highlight=cgroup#per-user-login-node-resource-limits


prckent:
LGTM. Given the resource limits & timeout, perhaps start with unit & deterministic only to see how long they take?

tests/test_automation/nightly_olcf_summit.sh
```diff
@@ -43,24 +43,20 @@ echo --- Hostname --- $(hostname -f)
...
-  git clone --depth 1 https://github.com/QMCPACK/qmcpack.git ${source_dir}
-else
-  echo --- Updating local QMCPACK git $(date)
-  cd ${source_dir}
-  git pull
-fi
+# refresh the source checkout
+mkdir -p ${base_dir}
+cd ${base_dir}
+rm -rf ${source_dir}
```
markdewing: A bare `rm -rf` feel dangerous - like a mistake in the script could delete things unintentionally.  Perhaps check that the `source_dir` variable is not empty and that that directory exists before doing the remove.

grahamlopez: both of these cases: `rm -rf    ` and  `rm -rf does_not_exist` result in a no-op with exit `0`

grahamlopez: I am happy for alternatives that allow it to start with a clean slate.

markdewing: I guess those cases are covered naturally, then.   I don't have any other ideas for checks.

ye-luo:
@prckent Do you want the summit CUDA nightly to run the legacy cuda code or the new GPU code with CUDA?

prckent:
@ye-luo We'll need to check both. My thoughts were that these can be added incrementally.

Note that we can't make a release without legacy CUDA working.

ye-luo:
legacy CUDA enabled in this PR

PR #2009, opened 2019-10-16
Ctest uses default site name if not specified

prckent:
Fix to default removed in #2004 resulting in most entries on cdash (empty).


ye-luo:
Thank you @prckent . My oversight in #2004. I need to move this line close to the use of HOSTNAME. Please review at #2010. If you agree with #2010, please close this PR.

Issue #1991, opened 2019-10-10
MPI job not running with error Signal code:  (-6)


afernandezody:
Hello,
I'm new to QMCPPACK and have compiled v3.8.0 with OMPI v4.0.2 and latest HDF5. The tests (ctest -R unit) passed with no failures. However, when I try to run an example, the system aborts and returns a signal code (-6). 
```
[centos@ip-172-31-44-183 coral2]$ mpirun -np 2 /home/centos/qmcpack-3.8.0/build/bin/qmcpack NiO-example.in.xml >out_2
Rank =    0  Free Memory =    70 MB
Rank =    1  Free Memory =    70 MB
[ip-172-31-44-183:30171] *** Process received signal ***
[ip-172-31-44-183:30171] Signal: Aborted (6)
[ip-172-31-44-183:30171] Signal code:  (-6)
[ip-172-31-44-183:30171] [ 0] /lib64/libpthread.so.0(+0xf5d0)[0x7f579ea165d0]
[ip-172-31-44-183:30171] [ 1] /lib64/libc.so.6(gsignal+0x37)[0x7f579e6702c7]
[ip-172-31-44-183:30171] [ 2] /lib64/libc.so.6(abort+0x148)[0x7f579e6719b8]
[ip-172-31-44-183:30171] [ 3] /lib64/libstdc++.so.6(+0x8b6cf)[0x7f579f0f46cf]
[ip-172-31-44-183:30171] [ 4] /lib64/libstdc++.so.6(+0x91b81)[0x7f579f0fab81]
[ip-172-31-44-183:30171] [ 5] /lib64/libstdc++.so.6(+0x91db4)[0x7f579f0fadb4]
[ip-172-31-44-183:30171] [ 6] /home/centos/qmcpack-3.8.0/build/bin/qmcpack[0x4399b4]
[ip-172-31-44-183:30171] [ 7] /home/centos/qmcpack-3.8.0/build/bin/qmcpack[0xa72274]
[ip-172-31-44-183:30171] [ 8] /home/centos/qmcpack-3.8.0/build/bin/qmcpack[0xaa0ee9]
[ip-172-31-44-183:30171] [ 9] /home/centos/qmcpack-3.8.0/build/bin/qmcpack[0xaa76ac]
[ip-172-31-44-183:30171] [10] /home/centos/qmcpack-3.8.0/build/bin/qmcpack[0xa1e410]
[...]

PR #2014, opened 2019-10-17
catch any remaining exception then call APP_ABORT

PDoakORNL:
Ideally exceptions would be handled closer to throw location where useful information could be added.


src/QMCApp/qmcapp.cpp
```diff
@@ -43,205 +44,220 @@ void output_hardware_info(Communicate* comm, Libxml2Document& doc, xmlNodePtr ro
...
-  myinput             = myinput.substr(0, myinput.size() - 4);
-  logname << myinput;
-
-  if (qmcComm->rank() != 0)
-  {
-    outputManager.shutOff();
-    // might need to redirect debug stream to a file per rank if debugging is enabled
   }
-  if (inputs.size() > 1 && qmcComm->rank() == 0)
+  catch (const std::exception& e)
```
markdewing: Does std::bad_alloc give a reasonable message in what(), or should there be a separate catch for the out-of-memory condition?

PDoakORNL: Well what() is all you have to call.  What is in it is implementation defined I believe.

ye-luo:
Can someone verify in #1991 the code actually throws an std::bad_alloc and it can be properly catched. If it does, I do believe we can move the try catch to the clone function and provide more helpful message. We only allocate vast memory in very few places and can provide very specific message.

PDoakORNL:
So I'm going to try and do this in a VM, Isn't a very nice thing to try and do to HPC a node.

prckent:
We already have exceptions so adding this catch seems important. Any objections?

ye-luo:
More precise, local and verbose catches are preferred than a global one.

prckent:
But is there a problem to add this in the interim? Are we closing this PR?
@PDoakORNL what is your preference?

jtkrogel:
Potentially dangerous if any raised exceptions are actually innocuous or incorrectly raised.  Broad scope includes all libraries too, right?  This is speculative, but will be interesting to see if the code becomes fragile due to this.

PDoakORNL:
An unhandled exception will currently cause the code to terminate without calling app_abort

prckent:
@jtkrogel The code is already fragile due to this. This PR will fix the problem.

ye-luo:
I'd like to see a confirmation that the catch works as we expected. @PDoakORNL you can simply put a large meshfactor in the input file to claim huge memory.

ye-luo:
@PDoakORNL is the major concern that std::terminate instead off MPI_ABORT may cause program  hangling?

PDoakORNL:
Yes

PDoakORNL:
ok to test

prckent:
Merging since the ALCF CI appears to have technical problems and not issues with this PR.

PR #2016, opened 2019-10-18
Fix CUDA runs due to NLPP RNG = nullptr

ye-luo:
Fix summit CUDA failures in determinsitic tests.
https://cdash.qmcpack.org/CDash/index.php?project=QMCPACK&date=2019-10-17
Bug introduced in #1998 not caught due to missing working CUDA tests in nightlies.


ye-luo:
Should fix summit CI missing CUDA module as well.

PR #2019, opened 2019-10-18
Fix PBCAA evaluation of Coulomb and Chiesa forces

tiihonej:
Evaluation of PBCAA forces has been partly broken due to various reasons:
- usage of incompatible LR handler (LRHandlerTemp::evaluateGrad not fully implemented)
- deprecated method to evaluate LR forces (ForceChiesaPBCAA::evaluateLR_AA)
- calling LRCoulombSingleton::getDerivHandler apparently breaks if called from multiple locations with potentially different particlesets (ForceChiesaPBCAA & CoulombPBCAA)

This commit fixes/gets around the aforementioned problems by choosing the LR handler more explicitly. Also, ForceChiesaPBCAA::evaluateLR_AA now uses the same technique as well-tested CoulombPBCAA::evalLRwithForces.

The default LR handlers can now be safely chosen as follows:
- Coulomb potential energy: LRHandlerTemp
- CoulombPBCAA forces: EwaldHandler3D
- ChiesaPBCAA forces: EwaldHandler3D
By safely, I also mean that the existing unit tests are passed, although the convergence of LR breakup in one of the existing force tests is changed.

I'm marking this WIP and opening discussion to ensure whether
- the explicit get*Handler is considered an elegant enough solution
- some other choice of defaults proves better in further testing
- there's reason to add more tests
- there's anything else important to consider (@rcclay , @Paul-St-Young ?)


tiihonej:
I amended the original PR, because the separation of LR handlers in CoulombPBCAA was incomplete.

I understand that while it's handy to explicitly choose LR handler for specific interactions, a nice optimization is lost. I'd like to hear your thoughts: do we prefer optimal handlers for energies/forces separately, or stick to one, which allows for a bit of simplicity and efficiency?

If latter, the choice for the moment is probably LRHandlerTemp. In this case, evaluateGrad or something similar needs to be implemented, so that IonIon forces can be accurately computed in all situations.

tiihonej:
After further review with @rcclay the changes could be simplified. Essentially, there no need for more that two LR handlers (energies & forces), but they must be explicitly separated within CoulombPBCAA. There's also no need to change the essence of ForceChiesaPBCAA::evaluateSR_AA, except to fix an obvious bug.

rcclay:
Ok to test.

tiihonej:
Added one more commit to fix #2031 and add a unit test for CoulombPBCAA IonIon forces

rcclay:
@tiihonej Does that commit fix the 64 atom H case?

tiihonej:
@rcclay Judging by the first couple of ions, yes, it does. However, according to @Paul-St-Young there might be further errors ion ElecElec and ElecIon forces, but then, they are a separate matter.

PR #2020, opened 2019-10-18
Add deterministic test for spline+no Jastrow for LiH

### LINKED ISSUE #2018: Need spline + no Jastrow deterministic test
It can be derived from short-LiH_solid_1x1x1_pp-gamma-vmc_hf_noj-1-16 .
See discussion in #2015

Hyeondeok-Shin:
Closes #2018 .

Add deterministic test for spline + no Jastrow for solid LiH.
References are obtained from QMCPACK 3.6 with intel18 on Intel Xeon, and tests are all passed on current developer version of QMCPACK on Intel Xeon (gcc & clang for non-CUDA) and Summit (CUDA).

Hyeondeok Shin


ye-luo:
Okay to test

Hyeondeok-Shin:
@prckent As pointed out, it seems due to seed number. Too different references between full & mixed precision may give confusion in future, so I'll update the references with different random seed to aviod it.

ye-luo:
I don’t get why we need different seeds. We would like to keep the minimal amount of input files.

Hyeondeok-Shin:
@ye-luo I mean update seed number & references for the current input file, not an additional one.

Hyeondeok-Shin:
Updated. Now everything seems good.

PR #2021, opened 2019-10-18
Complex Coefficients in Multideterminants

anbenali:
Implementation of complex coefficients in  multideterminant expansions. 

Fairly simple by only transforming RealType into ValueType and reading properly from the XML wavefunction.

Coefficients are still forced to be read in double precision. 

Contains one test with diamond 1x1x1 with complex wavefunction and complex Multideterminants

Note that in order to follow the occupation provided by the multideterminant section, the MO coefficients we read are stored as MO_Coeff_Kpoint_N, MO_Coeff_Kpoint_N+1 etc. This convention is easier when doing the sCI as it reduces the cost of sorting the excitations. Therefore the HDF5 coming from yscf always stores both cases; sorting the MO coeffs by kpoints (MSD case) or by Eigenvalue (SD case)

 
- Missing 1 test (Carbon2x1x1) which is in progress for the long reference run.


prckent:
OK to test

prckent:
Updating just to see if the ALCF CI passes while waiting for answer to the above question.

anbenali:
> Why is tests/solids/diamondC_1x1x1-Gaussian_pp_Tw_cplx_MSD/twf_input/MultiDet.h5 >1MB ?

We have 60k determinants...  hence 1.9M for the file. However, according to @ye-luo  having only 500 determinants is enough for the test... I will therefore redo the test with smallest number of determinants.

prckent:
Thanks. This reduces the size of the distribution. Best to remake the PR with a clean git history.

anbenali:
Replaced by PR #2062

PR #2026, opened 2019-10-21
Nexus: output corrected energy via qmca with MPC alone

jtkrogel:
Enable calculation of a finite size corrected energy value when only MPC is present.

@zenandrea, please can you confirm that this works for you as expected?  Usage: `qmca -q ce *scalar*`.


markdewing:
okay to test

ye-luo:
MPC can be a physical Hamiltonian component if the input is physical="yes". How do you distinguish that in the post process.

jtkrogel:
I don't think this case is handled currently.  I guess it would be marked by an absence of ElecElec along with presence of MPC.  There is nothing in the scalar.dat output that contains information related to `physical="yes"/"no"`.

PR #2029, opened 2019-10-22
Cleaning up SimpleFixedNodeBranch

PDoakORNL:
Made VParam enum scoped, Changed VParamType to std::array subtype.

### DIFF
--- a/src/QMCDrivers/SimpleFixedNodeBranch.h
+++ b/src/QMCDrivers/SimpleFixedNodeBranch.h
@@ -128,7 +128,7 @@ struct SimpleFixedNodeBranch : public QMCTraits
    *  Easy serialization is a relatively minor concern compared to the 
    *  annoyance this causes elsewhere.
    */
-  enum class SBVP
+  enum class SimpleBranchVectorParameter
   {
     TAU = 0,
     TAUEFF,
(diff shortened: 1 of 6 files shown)


src/QMCDrivers/SimpleFixedNodeBranch.h
```diff
@@ -116,30 +123,41 @@ struct SimpleFixedNodeBranch : public QMCTraits
...
+    FILTERSCALE,
+    VPARAM_MAX = 17 // four extra, why? Sloppy or undocumented hack?
   };
 
   /** controlling parameters of real type
    *
    * Mostly internal
    */
-  typedef TinyVector<FullPrecRealType, B_VPARAM_MAX> VParamType;
+  struct VParamType : public std::array<FullPrecRealType, static_cast<size_t>(SBVP::VPARAM_MAX)>
```
ye-luo:
```
using VParamType = std::array<FullPrecRealType, static_cast<size_t>(SBVP::VPARAM_MAX)>
```
Is this better?

PDoakORNL: I finished what I was refactoring toward. Removed the `operator[](int)` vParamType now has a type safe operator[] i.e. you must use a SBVP value. The vParam enum doesn't pollute the qmcplusplus namespace.

The other advantage is any C++ programmer knows how std::array works. Not so TinyVector.

Long term motive, I can now get the enum type the vParam uses so a pretty printer for the debugger can be written and I don't have to leave the declaration of the enum open.

ye-luo: TinyVector should be only used for tight math. All the generic use of array with known size at compile time should adopt std::array. Your way of wrapping std::array as VParamType still leak public accessors like `at()`, but we at least got some protection.

PR #2032, opened 2019-10-22
Fix the minimum tolerance for the deterministic tests

Hyeondeok-Shin:
We defined the minimum tolerance for the deterministic test as 10^-7, but there are some too lower tolerances causing failure on the test on cdash. (eg. https://cdash.qmcpack.org/CDash/testDetails.php?test=6461618&build=86251)

  This PR to fix the minimum tolerance for the test as 10^-7 as we promised. It will resolve the failure on the test.

 Hyeondeok Shin


markdewing:
okay to test

prckent:
Thanks @Hyeondeok-Shin

PR #2034, opened 2019-10-23
Fix DMCbatch timing

ye-luo:
The bug was caused by PhaseDiff which was not assigned in the batched interface of TWF.
PhaseDiff will be deleted in the near future because it can be directly access from WF ratios. There is no need to store it in TWF and later accessed by TWF.getPhaseDiff.

Adjust qmc_short_vmcbatch_dmcbatch to be as close to qmc_short_vmc_dmc as possible.


ye-luo:
The statistics are still slightly off the reference. We still need to fix the branching and take a close look if there is any other bugs.

ye-luo:
I can not understand why there were unit test failures in the gpu build. The failure happens in tests/test_coulomb_CUDA.cpp:199 randomly. It seems irrelevant to this PR. We need to keep an eye on gpu failures in the following days.

PR #2035, opened 2019-10-23
Fix g(r) for SoA

### LINKED ISSUE #1699: gofr estimator gives incorrect results for SoA
Related to #396 and #392 and #984.

AoS versions of the estimator on the HEG give expected results:
aos_n64_gofr

But SoA versions of the estimator do not have the appropriate normalization:
soa_n64_gofr

This behavior is not unique to the HEG, I've also seen it in FeO.
This was test on QMCPACK built on 28 June commit number
ba0e62ad98ece9c81555a8750de4e28612fc4dd2, using clang 3.9.0 on a small linux workstation.

markdewing:
The SoA version is missing the +1 offset for the pair_id index to the normalization factor.

Closes #1699


prckent:
Thanks Mark. By chance, have you thought of a way to test this? Perhaps a deterministic or short run and check of the normalization (large r limit), which should be very robust?

markdewing:
I emailed Josh a couple of days ago and he said he was working on a test.

prckent:
I made an issue so that we don't lose track of needing the test.

PR #2036, opened 2019-10-24
Serial hdf5 afqmc kp.

fdmalone:
Removes dependency on parallel hdf5 for kpoint Cholesky generation by implementing serial writing.

Serial write is now the default.


prckent:
OK to test

PR #2039, opened 2019-10-24
Fix serial write kpoint symmetry issue

fdmalone:
#2036 failed to correctly account for momentum transfer symmetry. This bug fix skips transfers that can't exist in the hdf5 file by symmetry.


markdewing:
okay to test

Issue #1786, opened 2019-08-08
Postprocessing tool for general estimators


jtkrogel:
There have been several requests for a tool that can post-process heavy observable data stored in stat.h5 files.  This also aligns with local needs at ORNL and needs in general to productionize use of the momentum distribution, for example.

Basic needs include the ability to twist average data appropriately.  This operation takes different forms for real space and k-space data.

Below I list a development plan for this tool (qstat).  Items will be added as appropriate.
- [ ] Basic ability to read stat.h5 files
- [ ] Postprocessing capability for densities
- [ ] Self-descriptive internal labeling within stat.h5 to identify estimator type
- [ ] Class to represent 3D functions over grids
- [ ] Ability to twist average n(k) data
- [ ] Ability to interpolate n(k) data and take derivatives

PR #2040, opened 2019-10-24
Enable density postprocessing

jtkrogel:
This PR includes the provisional `qdens` tool that can be used to obtain (twist averaged) up, down, up+down, and up-down densities from spin density data outputted by QMCPACK in `stat.h5` files.

This tool will have a limited lifetime as it will be superseded in the hopefully not too distant future by a general purpose tool to post process arbitrary quantities produced by QMCPACK in `stat.h5` files.  See also #1786.


jtkrogel:
@prckent If Jenny runs into any issues with the tool, she is welcome to contact me directly.

PR #2041, opened 2019-10-24
add unit tests for LRHandlers

Paul-St-Young:
No test currently exists in the LongRange folder.
This PR aims to check the correctness of the values produced by the short-range long-range breakup handlers `LRHandler`. Future PRs will use finite difference to check derivatives.

I think the `EwaldHandler3D` test is good to go. I mark this work in progress (WIP) because `LRHandlerTemp` needs a double-check. The sum of short-range and long-range pieces is correct, but the short-range piece spills outside the box. @rcclay can you take a look at my two comments that start with `??` in `test_temp.cpp` please?


markdewing:
okay to test

src/LongRange/tests/test_temp.cpp
```diff
@@ -0,0 +1,104 @@
...
+  dr = 5.0/nr;  // L/[# of grid points]
+  for (int ir=1;ir<nr;ir++)
+  {
+    r = ir*dr;
+    rinv = 1./r;
+    vsr = handler.evaluate(r, rinv);
+    vlr = handler.evaluateLR(r);
+    //// short-range part must vanish after rcut
+    //if (r>2.5) REQUIRE(vsr == Approx(0.0));
+    //// ?? YY: may it not?
```
rcclay: Uh.  I would think vsr should be zero here.  It's not??

src/LongRange/tests/test_temp.cpp
```diff
@@ -0,0 +1,104 @@
...
+  ref.Lattice = Lattice;  // !!!! crucial for access to Volume
+  ref.LRBox = Lattice;  // !!!! crucial for S(k) update
+  StructFact *SK = new StructFact(ref, Lattice.LR_kc);
+  ref.SK = SK;
+  LRHandlerTemp<EslerCoulomb3D<double>, LPQHIBasis> handler(ref);
+
+  handler.initBreakup(ref);
+  REQUIRE(handler.MaxKshell == 78);
+  REQUIRE(handler.LR_rc == Approx(2.5));
+  REQUIRE(handler.LR_kc == -1); // ?? YY: shouldn't the internal be updated?
```
rcclay: Oddly, doens't look like LRHandlerTemp sets this... rather, the optional argument is not passed in, so this gets set to -1.

src/LongRange/tests/test_lrhandler.cpp
```diff
@@ -0,0 +1,75 @@
...
+  //  the full Coulomb potential should be retained in kspace
+  for (int ish=0;ish<handler.MaxKshell;ish++)
+  {
+    int ik = SK->KLists.kshell[ish];
+    double k2 = SK->KLists.ksq[ik];
+    double fk_expect = fk(k2);
+    REQUIRE(handler.Fk_symm[ish] == Approx(norm*fk_expect));
+  }
+  // ?? cannot access base class method, too many overloads?
+  // handler.evaluate(SK->KLists.kshell, rhok1.data(), rhok2.data());
```
rcclay: ??

src/LongRange/tests/test_temp.cpp
rcclay: I will think about this.

rcclay: OK.  My knee jerk reaction is that LRHandlerTemp is playing fast and loose with the evaluation of short-stuff for r>r_c.  v_{lr}(r) = \sum c_i h_i(r) inside r_c where c_i is a coefficient, h_i(r) is a smooth basis function with compact support.  However, v_{lr}(r)=v(r) at r>=r_c (and the matching at r=r_c should be continuous).  This implies that the short-range piece, given by v_{sr}=v(r)-v_{lr}(r) is v(r)-\sum c_i h_i(r) for r<r_c (which is what the code does), and strictly 0 for r>r_c.  The fact that it does not return 0 in this case is an actual bug.  That this has not manifested in the mainline QMCPACK is because in PBC's, CoulombPBCAA and CoulombPBCAB spline the short-range piece only up to r<=r_c and evaluate to 0 for r>r_c.  This should definitely be fixed, because if I chose not to spline, I would get the wrong answers.

prckent:
@Paul-St-Young please remove the WIP if you are done with updates.

src/LongRange/tests/test_temp.cpp
rcclay: If you look at lrDf(), all real-space functions should have that logic.

Paul-St-Young: @rcclay are you saying the `LRHanlderTemp::evaluate` should have an if statement that sets it to zero when r>rc?

This effectively makes the short-range potential a piece-wise potential with a jump at L/2. Would this not cause any problems?

src/LongRange/tests/test_temp.cpp
Paul-St-Young: Should like set this in `LRHandlerTemp::initBreakup`?

src/LongRange/tests/test_lrhandler.cpp
Paul-St-Young: Yeah, not sure what happened here, but I did not want to put in the effort to debug `DummyLRHandler`...

Paul-St-Young:
@prckent OK, I will remove WIP, once I hash out the final form with @rcclay

Paul-St-Young:
The real-space functions were never intended to be evaluated outside of `rcut`, so the roles of SR and LR pieces swapped when I forced the evaluation
lrtemp3d-vsr-vlr

After putting in the if statements per @rcclay's comments, the real-space function can now be safely evaluated outside of `rcut`
lrtemp3d-vsr-vlr-withif

This PR is good to go.

src/LongRange/LRHandlerTemp.h
```diff
@@ -127,6 +131,7 @@ class LRHandlerTemp : public LRHandlerBase
   inline mRealType evaluateLR(mRealType r)
   {
     mRealType v = 0.0;
+    if (r>=LR_rc) return 1./r;
```
rcclay: This should be return myfunc(r,rinv).  1/r assumes coulomb and breaks the generic handling of myFunc.

Paul-St-Young: Great catch! I just pushed an update

Issue #2000, opened 2019-10-14
Evaluation of Finite size effects


zenandrea:
I would like to evaluate finite size effects.

Concerning the correction to the potential, I can use MPC.
If I set in input
    <pairpot name="ElecElec"  type="coulomb" source="e"    target="e"   />
    <pairpot type="MPC" name="MPC" source="e" target="e" ecut="30.0" physical="no"/>
then the local energy is calculated using the Ewald ElecElec interaction.
The correction is given my MPC-ElecElec.
I can use the tool qmca to have the value and associated error of MPC and ElecElec, separately, but not of the difference (which would have a much smaller error, because of the correlation between the MPC and ElecElec fluctuations).
So, if I want to compute the MPC-ElecElec error I should write a script to do that.
Am I correct? Or is there some way to tell qmca or some other script to do that?

Concerning the kinetic correction:
As far as I understood, I cannot get the kinetic correction or the structure function from the SoA version, because it not yet implemented. Thus, I should instead use the AoS version (which could have a bug in version v.3.8.0, issue #1995).
Which version of qmcpack is safe for those quantities (v.3.7.0 AoS)?

zenandrea:
I forgot to mention:
the manual reports about a tool `qmcfinitesize`
which I cannot find.
Does the tool (still/yet) exist?

aannabe:
If I am not mistaken, printed MPC will be only a correction if the above input is used.
3.7.0 AoS seems to have no bugs mentioned previously.

prckent:
I think that @jtkrogel has the best perspective of the situation but he is out for a few days. When he is back and has time we should discuss the situation here. Clearly MPC need to work since it is a standard (default?) observable for solid state calculations. I see two issues (1) How are the results output and analysed conveniently, including with error cancelation as mentioned above?  Some tweaks to qmca could handle this. (2) What is the status of MPC in the main develop branch, what is tested etc.? This is important given the upcoming transition to a single code path.

jtkrogel:
Currently `qmca` gives a corrected energy if both MPC and the Chiesa et al S(k) corrections are available.  I can put in a quick update to print a corrected energy in the presence of MPC alone.

[...]

PR #2042, opened 2019-10-25
Nexus: add more comprehensive finite size correction to qmca

jtkrogel:
This PR attempts to remedy deficiencies identified in #2000 regarding the application of MPC finite size corrections to the total energy.

The new approach first sums all non-electron-electron energy terms and then checks that the local energy is recovered when either ElecElec or MPC is added.  If the local energy can be recovered, MPC is added if present (ElecElec otherwise) and the Chiesa kinetic energy correction is added if present.

If the new approach is not successful, the code falls back to the old approach, which assumes ElecElec has been used as physical.


jtkrogel:
@zenandrea, please check whether this code handles all the cases you need.

nexus/bin/qmca
```diff
@@ -1842,16 +1845,55 @@ class DatAnalyzer(QBase):
...
+                        ce += data.KEcorr
+                    #end if
+                #end if
+            #end if
+            if ce is None:
+                ce = data.LocalEnergy.copy()
+                if 'MPC' in data and 'ElecElec' in data:
+                    ce += data.MPC - data.ElecElec
+                #end if
+                if 'KEcorr' in data:
```
ye-luo: `if 'KEcorr' in data` and the above one can be added at the end of the giant if block once by itself.

nexus/bin/qmca
```diff
@@ -1842,16 +1845,55 @@ class DatAnalyzer(QBase):
...
+                        ce += data.MPC
+                    elif 'ElecElec' in data:
+                        ce += data.ElecElec
+                    #end if
+                    if 'KEcorr' in data:
+                        ce += data.KEcorr
+                    #end if
+                #end if
+            #end if
+            if ce is None:
```
ye-luo: if condition based on `ce` as `none` or not good. Taking me 10 minutes to understand the logic makes me feel it is suboptimal here. Why not just define a flag MPC_is_physical if nee+MPC=le.

if MPC_is_physical 
  ce = le
else if MPC in data and ElecElec in data
  ce = le +MPC - ElecElec
end if

if KEcorr in data
  ce += KEcorr
endif

ye-luo: I also don't get why nee needs to computed based on Kinetic.
if ElecElec in data
  nee = LE - ElecElec
else
  nee = Le
endif

if MPC in data and (nee + MPC == LE)
  MPC_is_physical =true
endif

nexus/bin/qmca
jtkrogel: True.  I wanted to keep the two attempts complete and separate, but I don't care much either way.

nexus/bin/qmca
jtkrogel: The approach taken by the "if 'Kinetic' in data" block requires no special knowledge of the terms QMCPACK used internally to form the local energy.  It checks if it can recover the local energy by using either ElecElec or MPC as the e-e energy just to confirm that all the other terms included via self.non_ee_potentials comprehensively represents the non-e-e potential terms.  If it can, the corrected energy is always kinentic + non-e-e + MPC (if present, and +ElecElec otherwise).

jtkrogel: The 'if ce is None' is a fallback in case this approach does not work, e.g. a potential term is missing in the list.  The fallback represents the approach taken in the code prior to this PR, which is likely only correct if ElecElec is physical and MPC is not.

jtkrogel: See updates now with in-code comments.

ye-luo:
Okay to test

PR #2047, opened 2019-10-26
add unit tests for derivatives of LRHandlers

Paul-St-Young:
As promised in #2041. This PR adds finite difference tests for `LRHandler`s.

### DIFF
--- a/src/LongRange/LRHandlerSRCoulomb.h
+++ b/src/LongRange/LRHandlerSRCoulomb.h
@@ -282,6 +282,7 @@ private:
     LRBreakup<BreakupBasis> breakuphandler(Basis);
     //Find size of basis from cutoffs
     mRealType kc = (LR_kc < 0) ? ref.LR_kc : LR_kc;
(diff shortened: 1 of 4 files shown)


PDoakORNL:
Ok test this

prckent:
OK to test

Paul-St-Young:
@rcclay `LRHandlerTemp` derivatives are not functional at the moment. `gcoefs` are never initialized.

Paul-St-Young:
Also, I am not confident about the absolute correctness of `LRHandlerSRCoulomb`. It is correct to `~10^6` when compared to `Ewald3D`, which is fully validated against QE.

I have attached an example that shows the convergence of `LRHandlerSRCoulomb`.
[h2r190.zip](https://github.com/QMCPACK/qmcpack/files/3779433/h2r190.zip)
f4-h2-srcoul-ewald

prckent:
Suggestion: add hard aborts whenever something is unimplemented. We have been burned before.

PR #2048, opened 2019-10-28
Change log representation in WFC

ye-luo:
log(WFC) was stored as RealType LogValue and PhaseValue separately in WaveFunctionComponent.
This is is actually not the natural way of expressing log. Because WFC can be real with sign and complex, a united way is storing log(WFC) always in complex regardless of real or complex WFC.
After doing this change, converting back and forth between WFC value and its log becomes very simple and tons of #ifdef QMC_COMPLEX are not needed at all.

So far, I only changed WFC and derived classes and the next step will change TWF. However, this requires changes in drivers and I will do it after other driver cleaning work.

@markdewing I introduced PolarApprox similar to ComplexApprox because log(psi) values are allowed to be different by any amount of 2pi.


prckent:
Should our modifications to external codes such as catch be mixed in with the original sources?

markdewing:
The original catch source is a single include file and our modifications are small.  I think it's okay to keep this for now - I don't see a big benefit from changing it.

markdewing:
PolarApprox appears to overload the meaning of a complex number to be (magnitude,phase) for some comparisons.  Is this correct?   If so, it will be confusing.

external_codes/catch/polar_approx.hpp
```diff
@@ -0,0 +1,71 @@
+#ifndef CATCH_POLAR_APPROX
+#define CATCH_POLAR_APPROX
+
+#include <complex>
+#include <cmath>
+
+// Copy and modify the ComplexApprox class to handle complex numbers for polar coordinates
+
+namespace Catch {
+class PolarApprox : public ComplexApprox
```
PDoakORNL: I don't think polar approximation is a subtype of ComplexApprox it is a sibling. It's not good when inheritance is for reusing code and doesn't model the actual object relations.

ye-luo: Agreed. It is a sibling and should not use inheritance. Will change.

ye-luo:
@PDoakORNL apart from fixing the PolarApprox, this PR will not carry more changes since it is already hard to look at.

PDoakORNL:
That's fine, just file an issue if the next PR isn't coming soon.  That way we will get back at some point.

markdewing:
LogComplexApprox would definitely be a more precise name, and less confusing.

ye-luo:
Updating LogComplexApprox should be done today.

ye-luo:
There are a few FIXME need to be addressed. Will remove WIP once that is done.

PR #2049, opened 2019-10-28
abort if trying to use LRHandlerTemp derivatives

Paul-St-Young:
small amendment to #2047, abort is using `LRHandlerTemp` derivatives per @prckent suggestion.


markdewing:
okay to test

PR #2050, opened 2019-10-29
Enable Chiesa kinetic energy correction on SoA build & add deterministic test for KEcorr

### LINKED ISSUE #2025: Chiesa finite size correction not compatible with SoA
No version of this correction exists for SoA.  As this is currently our only two body finite size correction for the kinetic energy, a port to SoA should be made.

See also #1003.

Hyeondeok-Shin:
closes #2025. Thank you very much for @ye-luo for fruitful comments and help on this. 

 Enable Chiesa kinetic energy correction on SoA build. Function to estimate KEcorr is ported to SoA build, and a helper class has added on J2OrbitalSoA to activate KEcorr during also optimizing Jastrow on SoA.  KEcorr on SoA was tested on VMC & Jastrow optimization, and KEcorr is estimated normally on both SoA & AoS build now.

 In addition, deterministic test for KEcorr has added on deterministic-diamondC_2x1x1_pp-vmc_sdj. Reference is generated on AoS build, and it is passed on both real & complex build on SoA.

 Hyeondeok Shin


prckent:
OK to test

Hyeondeok-Shin:
@prckent I'm not sure if we need it because kecorr should be a constant in fixed Jastrows. So, if implementation on kecorr has changed, we'll be able to see that change on the deterministic test. If we need to check constancy on kecorr, short test sounds enough for me.

PR #2051, opened 2019-10-29
Nexus: tests for project manager module

jtkrogel:
Add tests for the `project_manager` module.  This includes actual execution of full workflows composed of simple test runs (in which each run just calls "echo run" instead of performing an actual simulation).


jtkrogel:
Note for reviewers: in addition to deleting some old/unused code in project_manager.py, the contents of this file was rearranged for better logical consistency.  This generates a lot of visible file changes that are actually just cosmetic.

jtkrogel:
Ready for merge.

markdewing:
okay to test

PR #2053, opened 2019-10-29
Spack nightly test script that runs on Ubuntu Desktop at Argonne.

naromero77:
This test script runs as a cron job on the managed Ubuntu 18.04 desktop at Argonne. The script builds with multiple compilers, multiple variants, with and without CUDA and tests the patch on QE converter. Note that it is incredibly time consuming to build each one of these combinations from scratch. There are over 36 combinations per compiler and we don't even test all the different variants.


prckent:
For reference, how long does this take to run and on which machine? (The oxygen spack setup script takes ~6h, an older 12 core Xeon)

prckent:
How confident are you that you will see the failure messages?

ye-luo:
Okay to test

PR #2054, opened 2019-10-30
Bug Fix  for PBC gamma

anbenali:
When running real code with PBC LCAO code, phase factor should be 1.0. A bug was introduce when implementing Complex LCAO wavefunctions as it was tested for molecules but not for PBC systems.

This PR should fix all failed tests on real PBC.


markdewing:
okay to test

PR #2057, opened 2019-10-30
update blue waters build script

Paul-St-Young:


ye-luo:
Okay to test

prckent:
Note that now titan is decommissioned we have no equivalent Cray machine to check builds or run tests on.

Paul-St-Young:
@prckent Thanks for the heads up. I will run my own short and long tests before production.

PR #2058, opened 2019-10-30
Afqmc occupancy threshold

fdmalone:
Add user controllable occupancy threshold parameter for wavefunctions generated from mean field calculations with non integer occupancies.


markdewing:
okay to test

PR #2055, opened 2019-10-30
Nexus: tests for bundle module

jtkrogel:
Add tests for `bundle` module.

PR #2059, opened 2019-10-30
Nexus: unify test framework

jtkrogel:
Place all tests in the same framework (pytest style).  Also specify test execution order following logical dependency order.  This PR depends on #2055.


markdewing:
okay to test

PR #2061, opened 2019-10-30
Nexus: support QMCPACK DLA input

jtkrogel:
Add ability to specify set `DLA="yes"` in QMCPACK input via Nexus.

To use do `qmc = generate_qmcpack(...,dla=True)`.


markdewing:
okay to test

PR #2064, opened 2019-10-31
DMCBatched branching

PDoakORNL:
Numerous bugs mainly involving the complex distributed state machine behind the legacy population control.  Much of it remains.  Some of it is still not being updated as the refactored code expects. So energies get lower and lower and more and more walkers spawn.


prckent:
Test this please

src/QMCDrivers/DMC/DMCBatched.cpp
```diff
@@ -232,19 +233,20 @@ void DMCBatched::advanceWalkers(const StateForThread& sft,
...
       twf_reject_list.clear();
       elec_accept_list.clear();
       elec_reject_list.clear();
 
-
       for (int iw = 0; iw < num_walkers; ++iw)
       {
         if ((!rejects[iw]) && prob[iw] >= std::numeric_limits<RealType>::epsilon() &&
-            step_context.get_random_gen()() < prob[iw] * std::exp(log_gb[iw] - log_gf[iw]))
+            step_context.get_random_gen()() < prob[iw])
```
ye-luo: The above unchanged epsilon check gets affected by the change of prob.
This is actually an inconsistency between VMC and DMC implementations. The non-batched drivers also have it. It is hard to say which is right and which is wrong since the epsilon check is only a safeguard. I personally like the DMC way. So how about changing all the VMC to be consistent with DMC in a later PR?
After the change in this PR. The VMC non-batch and batch are consistent. The DMC non-batch and batch are consistent. VMC and DMC are not consistent.

src/QMCWaveFunctions/TrialWaveFunction.cpp
```diff
@@ -790,8 +793,18 @@ void TrialWaveFunction::flex_updateBuffer(const RefVector<TrialWaveFunction>& wf
   {
     buf_list[iw].get().put(wf_list[iw].get().PhaseValue);
     buf_list[iw].get().put(wf_list[iw].get().LogValue);
-    assert(buf_list[iw].get().size() ==
-           buf_list[iw].get().current() + buf_list[iw].get().current_scalar() * sizeof(double));
+#ifndef NDEBUG
```
ye-luo: 1. I don't quite get the logic of `#ifndef`.
2. could you write a small helper function so we don't duplicate twice?

src/QMCDrivers/DMC/DMCBatched.cpp
```diff
@@ -232,19 +233,20 @@ void DMCBatched::advanceWalkers(const StateForThread& sft,
       std::transform(drifts.begin(), drifts.end(), log_gb.begin(),
                      [oneover2tau](auto& drift) { return -oneover2tau * dot(drift, drift); });
 
-      std::transform(ratios.begin(), ratios.end(), prob.begin(),
-                     [](auto ratio) { return std::norm(ratio); });
+      for (int iw = 0; iw < num_walkers; ++iw)
+      {
```
ye-luo: Could you remove the `{}` and save two lines when you make a commit for other changes..

src/QMCWaveFunctions/TrialWaveFunction.cpp
PDoakORNL: This shouldn't be checked in production because it should never be wrong. It used to be an assert which only do anything `#ifndef NDEBUG` but it didn't provide the information I needed to debug. I agree about the helper function will do.

src/QMCDrivers/DMC/DMCBatched.cpp
PDoakORNL: Well except the original DMC skips the epsilon comparison.  I don't think there is any point to that comparison so I think we should drop from both and make VMCBatched consistent.

src/QMCWaveFunctions/TrialWaveFunction.cpp
ye-luo: Thanks. The logic is correct. I was confused by the double negative in the macro condition and name.

src/QMCDrivers/DMC/DMCBatched.cpp
ye-luo: My original comment was incorrect and I updated it. @PDoakORNL was right there was check in the VMC driver only.
I still think this check is beneficial to kill bad sampling. Since random numbers are generated [0,1) there is still a chance that `0 < prob[iw]` gets valid. When prob gets very close to zero, the TWF is very singular and then all the numerics goes wrong. We get spikes in the energy and the Monte Carlo sampling cannot recover by itself. So I'd like to protect this case although the probability is low but it is possible.

PDoakORNL: I will leave the check but since what we ultimately compare is the std::norm(ratio) * std::exp(log_gb[iw] - log_gf[iw])) I think it should be fine to compare that to epsilon.

ye-luo: The check is intended to guard ratio close to zero namely the WF of the proposed move is very singular.

PDoakORNL:
On leconte (i.e. power9 with clang I now see this failure)
```shell
../src/LongRange/tests/test_srcoul.cpp:129: FAILED:
  REQUIRE( handler.lrDf(r) == Approx(dvlr) )
with expansion:
  1844197.971078356
  ==
  Approx( 1844097.9702885668 )
```

PR #2066, opened 2019-10-31
Added new unit test for PairCorrEstimator.

### LINKED ISSUE #396: Tests of gofr estimator needed
SoA implemenation is added in #395 but the correctness is not tested due to the lack of tests.

jptowns:
The test consists of 8 electrons (4 up, 4 down) on a B1 (NaCl) lattice of side-length 2.
g_uu(r), g_ud(r) and g_dd(r) are computed and compared to analytic results.

The test is part of test_hamiltonian. Currently, PairCorrEstimator fails this test because of two approximations in the assignment of norm_factor.

I have fixed this, and will include it in the following PR.

Closes #2035 
Closes #396


src/QMCHamiltonians/tests/test_paircorr.cpp
```diff
@@ -0,0 +1,197 @@
...
+      std::cout << std::setw(4) << i
+		<< "  " << std::setw(12) << i*0.10
+  		<< "  " << std::setw(12) << gofr[i]
+  		<< "  " << std::setw(12) << gofr[i+20]
+  		<< "  " << std::setw(12) << gofr[i+40] 
+  		<< "\n";
+    } 
+
+  // Verify against analytic result.
+  const RealType eps = 1E-01;  // tolerance
```
prckent: Is there are reason for this large tolerance (absolute difference for checks)? Would you be happy to make it stricter?

prckent:
Thanks Josh. Expect approval after we have discussed the eps question.

src/QMCHamiltonians/tests/test_paircorr.cpp
jptowns: The tolerance is small because about 0.05-0.01 disagreement between analytic and QMCPACK result, depending on the gofr bin. I'm not sure where the source of disagreement is coming from, as the norm_factor is now exact.

With the fix from Mark and the changes to norm_factor, a test on a more realistic system now gives expected results with the expected normalization. 256 electron HEG with rs=1.0 no jastrow (SOA + fixes):
gofr

jptowns: Same problem before fixes:
gofr old

prckent:
Test this please

prckent:
Clearly this PR is moving us in the right direction.

rcclay:
test this please

ye-luo:
Okay to test

ye-luo:
@jptowns I fixed some compilation issues (relying on CI to test all cases). The left failures are caused by the tests. Please have a look.

PR #2067, opened 2019-10-31
Pair corr estimator norm factor fix

jptowns:
Fixed PairCorrEstimator so that it will pass unit test.

The reason it failed was that norm_factor was built using two approximations:
1.) Small delta approximation: (r+delta)^3 - r^3 = 3*delta*r^2 + 3*r*delta^2 ~ 3*delta*r^2
2.) Large N approximation: N(N-1) = N^2 - N ~ N^2

Both approximations are badly violated for a small cell with a large delta and small number of particles, which is exactly the case for the unit test.

Because computing the normalization is not the computational bottleneck in a typical QMC calculation, I changed norm_factor to use the exact expressions.

### DIFF
--- a/src/QMCHamiltonians/tests/test_paircorr.cpp
+++ b/src/QMCHamiltonians/tests/test_paircorr.cpp
@@ -15,7 +15,6 @@
 #include "Lattice/ParticleBConds.h"
 #include "Particle/ParticleSet.h"
 #include "Particle/DistanceTableData.h"
-#include "Particle/SymmetricDistanceTableData.h"
 #include "QMCHamiltonians/PairCorrEstimator.h"
 #include "QMCApp/ParticleSetPool.h"
 
@@ -134,8 +133,13 @@ TEST_CASE("Pair Correlation", "[hamiltonian]")
 
   elec->get(std::cout); // print particleset info to stdout
 
-  // Set up the distance table
+  // Set up the distance table, match expected layout
+#ifdef ENABLE_SOA
+  const int ee_table_id = elec->addTable(*elec, DT_SOA, false);
+#else
   const int ee_table_id = elec->addTable(*elec, DT_AOS, false);
+#endif
+  
   const DistanceTableData& dii(elec->getDistTable(ee_table_id));
   elec->update(); // distance table evaluation here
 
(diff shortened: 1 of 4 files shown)


rcclay:
ok to test

jptowns:
I'm not sure I understand all the messages from the above builds, but it looks like complaints about AFQMC. Do I need to do something special in order to test my changes?

src/QMCHamiltonians/tests/test_paircorr.cpp
```diff
@@ -0,0 +1,197 @@
...
+// File created by: Mark Dewing, markdewing@gmail.com, University of Illinois at Urbana-Champaign
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+#include "catch.hpp"
+#include "OhmmsData/Libxml2Doc.h"
+#include "Lattice/ParticleBConds.h"
+#include "Particle/ParticleSet.h"
+#include "Particle/DistanceTableData.h"
+#include "Particle/SymmetricDistanceTableData.h"
```
markdewing: This should only be included for AOS.

markdewing:
This issue is included files for SOA vs. AOS.    Try removing the include of SymmetricDistanceTable.h.

src/QMCHamiltonians/tests/test_paircorr.cpp
prckent: Maybe we should put AoS_only in their name?

src/QMCHamiltonians/tests/test_paircorr.cpp
```diff
@@ -0,0 +1,197 @@
...
+  elec->R[6][1] = 0.0;
+  elec->R[6][2] = 1.0;
+  elec->R[7][0] = 1.0;
+  elec->R[7][1] = 1.0;
+  elec->R[7][2] = 1.0;
+
+  elec->get(std::cout); // print particleset info to stdout
+
+  // Set up the distance table
+  const int ee_table_id = elec->addTable(*elec, DT_AOS, false);
```
markdewing: This should have an ifdef for SOA vs. AOS
```
#ifdef ENABLE_SOA
 ....  addTable(*eloc, DT_SOA, false);
#else
  ...  addTable(*eloc, DT_AOS, false);
#endif

jptowns: Wow. Turns out if you run in SoA mode, you can still create a AoS distance table. For the unit test, this meant that PairCorrEstimator basically never went down the SoA path in PairCorrEstimator::evaluate() !

Fixed and pushing new commit now.

PR #2074, opened 2019-11-04
Nexus: tests for input classes

jtkrogel:
Add tests for all classes derived from `SimulationInput`.


markdewing:
okay to test

ye-luo:
Overriding alcf-gnu6-mkl18 CI which failed in noMPI build due to improper configuration.

PR #2077, opened 2019-11-05
Nexus: tests for analyzer classes

jtkrogel:
Add tests for all classes derived from `SimulationAnalyzer`.


markdewing:
okay to test

PR #2078, opened 2019-11-05
Power9 mass massv

PDoakORNL:
```shell
with mass/massv linked
      DMCBatched::MovePbyP                     2.7871     0.1506           2500       0.001114845
        ParticleSet::computeNewPosDT           0.0756     0.0756          40000       0.000001891
        ParticleSet::donePbyP                  1.7544     1.7544           5000       0.000350884

without mass/massv linked
      DMCBatched::MovePbyP                    10.9274     0.1589           2500       0.004370963
        ParticleSet::computeNewPosDT           0.0727     0.0727          40000       0.000001818
        ParticleSet::donePbyP                  9.8953     9.8953           5000       0.001979057
```

I think I've solved the sincos problem for power9, still need to try this on summit (these results are from leconte)
Additionally yesterday this came into llvm
https://github.com/llvm/llvm-project/commits/73c3137a82c96789f4a2b8ec9427d23fa73498d8/llvm/test/CodeGen/PowerPC/lower-massv.ll

FindMass.cmake is a bit hacky, since for whatever reason both `check_function_exists` and `check_cxx_symbol_exists` fail.


ye-luo:
Nice to get this fixed but don't get disappointed by the small gain in the full app.
The gain is from structure factor calling the following function which uses vector math underneath
https://github.com/QMCPACK/qmcpack/blob/19db7a775ea743646dba9450723406d0bc1cdba7/src/LongRange/StructFact.cpp#L124

prckent:
Good to fix and be consistent with our vector math obsession on x86.

Minor nits:
* Find vector math.cmake should print "Using IBM MASS library for vector math"
* I would prefer we call the file FindIBMMASS to make it clearer that it is IBM specific functionality in file listings. MASS is not well known.

prckent:
Notice that eval_e2iphi has an out of place implementation with MASS in Numerics/e2iphi.h  while the implementation with Intel MKL VML makes use of a library call that works in place. Perhaps MASS now has an appropriate call that would avoid the extra copy? simd/vmath.hpp is also incomplete and is missing custom versions of inv() that exist with Intel MKL.  I don't know if/where we call these.

Aside: I don't understand the reasoning behind separate simd and Numerics directories when Numerics also has to be simd aware. We could merge them.

PDoakORNL:
I have mass inv calls added in next push. And I’ll take a quick look at the organization.

ye-luo:
Regarding cmake organization, we need to simplify the top level complexity and enhance fexibility for other vendor scalar and vector math libraries. I have seen ARM need libamath.

So we need our own FindMATH it handles scalar math libraries.
Most of the time it should do nothing since -lm is added by compiler.
To use vendor optimized math library, we need to add IBM MASS or ARM amath library when using GCC or Clang. Vendor compilers may add such libraries for us.

We also need to expand FindVectorMath to support MASSV in a sane way.

prckent:
@PDoakORNL While cleanup is appreciated, no need to get it all done here.  Only Summit is particularly relevant at the moment in my opinion. If/When an additional platform needs to be supported we can tidy further.

prckent:
@PDoakORNL Please put comments in the added clang scripts to say that they are experimental as per the new build_olcf_summit_XL.sh . We don't want anyone running these unintentionally.

```
echo "----------------------- WARNING ------------------------------------"
echo "This is **not** production ready and intended for development only!!"
echo "Use config/build_olcf_summit.sh for production on Summit."
echo "----------------------- WARNING ------------------------------------"
```

Small consistency improvement: load_summit_modules.sh -> load_olcf_summit_modules.sh please add comment to say that these modules need to be loaded to run the executables built with build_olcf_summit_modules.sh, and vise versa. From now on I think we should do this on other platforms as the scripts are updated.

PDoakORNL:
I have done a force push to this branch so if you pulled it you have an invalid branch until you `git pull --force`

Clang build for Summit will appear in a future PR.

prckent:
@ye-luo Do you approve merging after your fixes, assuming the CI passes? Please mark your requested changes as addressed if so. Then others can merge the PR.

ye-luo:
After connecting the cmake variables to macros in the source code, there are some compilation issues which @PDoakORNL is looking at it now.

PR #2084, opened 2019-11-06
Avoid generating multi determinant in AFQMC.

fdmalone:
When generating wavefunctions from mean field calculations with partially occupied degenerate bands the wavefunction generation script will attempt to generate a MSD expansion to include all these states. This can generate very large expansions which aren't currently useful. This PR sets the default to just generate a single determinant trial and avoid the potentially costly determinant generation step.


markdewing:
okay to test

prckent:
Test this please

Issue #2085, opened 2019-11-06
Possible Integer Overflow for KList Construction/Update


rcclay:
For the construction of a set of k-vectors within a certain cutoff k_c, the function "long KContainer::GetHashOfVec(TinyVec<int> kvec, int hashparam)" is called.  This hash is given by kvec[2]+hashparam*(kvec[1]+hashparam*kvec[0]).

For a cubic cell with L=15bohr, LR_DIM_CUTOFF=15, the coulomb potential construction can initialize a kvector list of size 113080, which is then used for hashparam.  This causes an integer overflow for 32 bit integers for most reasonable choices of integer vectors kvec. 

Solution:  change the return type and math to be 64 bit integers (long long).

rcclay:
Closed by PR #2087

PR #2087, opened 2019-11-07
Change hash value and map key to long long type

rcclay:
Addresses issue #2085


src/LongRange/KContainer.cpp
```diff
@@ -306,7 +306,7 @@ void KContainer::BuildKLists(ParticleLayout_t& lattice, bool useSphere)
   //Now fill the array that returns the index of -k when given the index of k.
   minusk.resize(numk);
   // Create a map from the hash value for each k vector to the index
-  std::map<int, int> hashToIndex;
+  std::map<long long, int> hashToIndex;
```
PDoakORNL: instead of the overload set and all the casts in the header, since this is private and used only here what about a generic lambda?

```C++
auto getHashOfVec = [](const auto& inpv, int hashparam)->long long{
    long long hash = 0; // this will cause integral promotion below
    for(int i = 0; i < inpv.Size; ++i)
      hash += inpv[i] + hash * hashparam;
    return hash;
  };
```

rcclay: That's slick.  Doing it.

src/LongRange/KContainer.h
```diff
@@ -29,15 +29,16 @@ class KContainer : public QMCTraits
 {
 private:
   //Function to return a unique number for each kVector
-  inline long GetHashOfVec(const TinyVector<int, 3>& inpv, int hashparam)
+/*  inline long long GetHashOfVec(const TinyVector<int, 3>& inpv, int hashparam)
```
ye-luo: Do we still need these commented code? Or delete?

rcclay: Oops.  Removed.

PR #2088, opened 2019-11-07
Expand ENABLE_SOA region for future deletion

ye-luo:
No new items identified for the transition gap.


prckent:
Related topic: Are there many source files that are AoS specific that don't have AoS in their name? Renaming them would be a good readiness activity.

src/QMCWaveFunctions/Fermion/BackflowBuilder.cpp
```diff
@@ -21,7 +21,9 @@
 #include "QMCWaveFunctions/Fermion/Backflow_ee.h"
 #include "QMCWaveFunctions/Fermion/Backflow_ee_kSpace.h"
 #include "QMCWaveFunctions/Fermion/Backflow_eI.h"
+#ifndef ENABLE_SOA
 #include "QMCWaveFunctions/Fermion/Backflow_eI_spin.h"
+#endif
```
jtkrogel: As discussed in the ECP call, I think we should keep this one.

We can just comment out the AoS code contained in `Backflow_eI_spin.h` at the time of the transition.  This will facilitate later resurrection of the code and will not impact SoA function.  The abort you put in below will prevent any attempted execution of the code after the transition.

prckent: This was agreed upon.

An abort will be needed + comments.

PR #2090, opened 2019-11-07
Pick the correct optimal jastrow with nexus

tiihonej:
Looking closely to the optimizer performance it turns out that nexus is not picking the correct Jastrow coefficients. Nexus analyses the optimization results for the best performing series and picks the corresponding wavefunction to carry over to other calculations. However, the Jastrow coefficients producing the optimal results are not those associated with the same index but those one step earlier. This commit fixes this behavior.

If the Jastrow is well converged, this does not make a big difference (but should be fixed nonetheless). If the Jastrow still fluctuates a lot, so will the outcomes (especially forces, # 1847).

I marked this WIP, because I don't know how to unit test nexus, so I'd like to hear from @jtkrogel if this might break something elsewhere.


jtkrogel:
1. Please check the edge cases, e.g. what happens when the data from `*s000.scalar.dat` has the lowest energy/cost?

2. Are the QMCPACK code updates related to forces intended for this PR?

markdewing:
okay to test

PR #2091, opened 2019-11-08
Nexus: tests for executables

jtkrogel:
Add tests for both general purpose Nexus executables (`nxs-sim`, `nxs-redo`) and QMCPACK-specific executables (`qmca`, `qmc-fit`, `qdens`).

This PR depends on #2089.


markdewing:
okay to test

PR #2092, opened 2019-11-08
Add Summit XL OpenMP offload + CUDA build script.

ye-luo:


ye-luo:
@prckent print a warning with pointer.

PR #2094, opened 2019-11-11
Nexus: stabilize tests

jtkrogel:
Stabilize a few tests on more than one platform (comparisons against zero and other numerical stability issues).  Also fix test names to be compatible w/ our ctest approach (e.g. a test named "simulation" matched several others with -R, a name change fixes this).


ye-luo:
Okay to test

PR #2095, opened 2019-11-11
Improve manual and add branch cutoff description

### LINKED ISSUE #2065: branching_cutoff_scheme description needed in manual
(Noticed in #2063)

prckent:
* Compiles
* Added branching_cutoff_scheme section
* Added references to Zen and DMC algorithm papers
* Split DMC parameter table in two to workaround > 1 page table issue.

@ye-luo Please directly edit the branching_cutoff_scheme text if necessary to make it correct. 

Closes #2065


prckent:
I wonder if anyone knows the origin of the "classic" branch cutoff algorithm implemented in QMCPACK? Is there a reference - a publication or thesis?

jtkrogel:
I looked into this some months ago and if I remember correctly, most of it is in Ceperley-Reynolds 1982.  The velocity rescaling bit is from Umrigar 1993.  @ye-luo, does this match your understanding?

prckent:
@jtkrogel I could not find the reference to the classic formula, but hopefully I simply overlooked it. If you can find it again easily please note here. Otherwise we can update in future.

ye-luo:
@jtkrogel and @prckent I don't know where the "classic" formula comes from. When I need to add other options, I moved the old one as "classic". I just updated the formula in the manual to reflect the exact implementation. I do feel it is a strange formula with the sigma^2 instead of sigma. It will be very helpful if we can find the original formula to confirm the current one not being a bug. Having sigma^2 instead sigma basically introduce a bit loose control. For a reasonable WF, the difference between using sigma^2 and sigma should be negligible but for a poor quality WF, the population may be easier to blow up or die.

jtkrogel:
No, I don't know where the specific form of the classic formula comes from either.

In reference to checking correctness of zero timestep FN-DMC extrapolations, there is extensive data cited for LiH (many studies) in the following review paper:

https://doi.org/10.1080/01442359509353305

prckent:
The dimensions (units) of the classic formula are a little strange, but we do at least know that the algorithm is correct in the small tau limit. Classic is a suitable name for it.

I think we can proceed as is, since this PR just aims at describing what we have. However we will have to revisit the topic of what is implemented and our default algorithms

prckent:
@ye-luo Assuming you approve, please can you approve and retrigger the CI? It looks like we are suffering from unreliable GPU code.

ye-luo:
Test this please

PR #2099, opened 2019-11-13
Faster read/write afqmctools

fdmalone:
For larger systems writing/reading integrals became bottleneck due to unnecessarily unpacking lists. This PR removes this.


markdewing:
okay to test

PR #2101, opened 2019-11-15
Remove clang warning suppression and reduce warnings

ye-luo:
Mostly delete unused code and remove deprecated register keyword.


prckent:
How many warnings are left after removing these settings? (cdash has a ~500 limit before truncating)

ye-luo:
After removing these warning suppression, clang should in principle print more warnings previously suppressed but I resolved all these places.

PR #2103, opened 2019-11-18
Fix LCAO supertwist (0 0 0) with complex part

anbenali:
Fix case where SuperTwisit 0 0 0 still has Complex component - Was previously forced to put imaginary part to 0. (case of super cells 3x3x3 5x5x5 etc.. )


markdewing:
okay to test

prckent:
Simpler and more correct.

Issue #1770, opened 2019-07-30
Spin-Orbit Commit Plan


rcclay:
Hi all!  Here's my tentative PR schedule for getting spin-orbit estimators and spin DMC into QMCPACK:

- [x] Establish an XML format for QMCPACK PP's with SO interactions.  Have this format be parsed and "registered" by QMCPACK, meaning that skeleton classes are constructed where they should be and data ends up where it should.  (PR #1771 )
- [x] Augment the LCAO h5 and/or pwscf.h5 file formats to handle spinors. Parse these and check that data is ending up in correct skeleton classes. (PR #2104)
- [x] Implement spinor orbital class and SpinorSlaterDeterminant classes. (PR #2104).
- [x] Implement SO potential evaluation. (PR #2312, #2325)
- [x] Implement SOVMC and SODMC. (PR #2147, #2197)

prckent:
Please try to write tests as you go so that (as much as possible) the different components can be tested independently.

ye-luo:
@rcclay done? close?

rcclay:
Implementation is done.  We're tracking a bug in PBC's, but that should be a separate issue.

PR #2104, opened 2019-11-18
Define Spinor Set and Allow Parsing for Augmented ESHDF5 Format.

rcclay:
This is a beefy PR, but it basically does the following.
1.  Defines a SpinorSet, which is derived from SPOSet, 
2.  Defines a <spinorset_builder> syntax in XML and underlying class to build a spinor SPOSet.  
3.  Defines an EinsplineSpinorSetBuilder, which derives from EinsplineSetBuilder, and is responsible for parsing and building an einspline spinor set from the augmented ESHDF5 spinor format.  This is defined in PR #1783.
4.  Creates a unit test of this entire pathway for O2 in PBC's, but with an imposed non-collinear magnetic state.  
5.  Includes a complete QMCPACK xml file for building a Slater-Jastrow wave function built up of non-collinear spinors.

This should address points 2 and 3 in the Spin-Orbit commit plan laid out in Issue #1770.

Some general notes on the philosophical design of this...
- Principal of least resistance and maximal laziness is employed.  SpinorSets are basically SPOSets that hold two SPOSets, one for the up channel, and one for the down channel, and combine them in a way consistent with the Mitas/Melton paper.  This is so SpinorSets can be dropped into existing antisymmetric wavefunctions without any modifications, and also so that the current builders for SPOSets can be hijacked as much as possible.  
- EinsplineSetBuilder as I have complained to some before does many things...  parsing, construction, and error checking.  Because of the format of the Spinor HDF5 file, many of these checks will fail, but things are too entangled to separate the checks from overall parsing/construction.  Because of this, I've added a "skipChecks" flag to ReadOrbitals, OccupyOrbitals, etc., to let hard aborts slide if checks fail.


ye-luo:
Please post input example.

rcclay:
@ye-luo I've included an input example in this PR:  tests/solids/monoO_noncollinear_1x1x1_pp/qmc_short.in.xml

rcclay:
Wasn't as careful as I should have been with real vs. complex.  I'll fix tomorrow.

ye-luo:
Since EinsplineSpinorSetBuilder is derived from SPOSetBuilder and SpinorSet is derived from SPOSet, I don't see a reason to branch off current input style and parsing code.
```
   <sposet_builder> 
     <sposet> 
     </sposet> 
   </sposet_builder> 
```
I feel better with
```
   <sposet_builder name="A" type="spinorbspline" href="eshdf.h5" tilematrix="1 0 0 0 1 0 0 0 1" twistnum="0" source="ion0" size="12"> 
     <sposet name="myspo" size="12"> 
       <occupation mode="ground"/> 
     </sposet>
   </sposet_builder> 
```
Then we don't need SpinorSetBuilderFactory.cpp which mostly is a duplication of SPOSetBuilderFactory.

rcclay:
That’s a fine simplification. I’ll do that.

src/QMCWaveFunctions/SPOSetBuilderFactory.cpp
```diff
@@ -335,7 +348,7 @@ void SPOSetBuilderFactory::build_sposet_collection(xmlNodePtr cur)
   while (element != NULL)
   {
     std::string cname((const char*)(element->name));
-    if (cname == "sposet")
+    if (cname == "sposet" || cname == "spinorset")
```
ye-luo: This change is no more needed.

src/QMCWaveFunctions/SPOSetBuilderFactory.h
```diff
@@ -51,15 +51,15 @@ class SPOSetBuilderFactory : public MPIObjectBase
 
   ~SPOSetBuilderFactory();
 
-  SPOSetBuilder* createSPOSetBuilder(xmlNodePtr rootNode);
+  virtual SPOSetBuilder* createSPOSetBuilder(xmlNodePtr rootNode);
```
ye-luo: This change is no more needed.

src/QMCWaveFunctions/SPOSetBuilderFactory.h
```diff
@@ -51,15 +51,15 @@ class SPOSetBuilderFactory : public MPIObjectBase
...
 
   void loadBasisSetFromXML(xmlNodePtr cur) { last_builder->loadBasisSetFromXML(cur); }
 
   SPOSet* createSPOSet(xmlNodePtr cur);
 
   void build_sposet_collection(xmlNodePtr cur);
 
-private:
+protected:
```
ye-luo: This change is no more needed.

src/QMCWaveFunctions/SpinorSet.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include "QMCWaveFunctions/SpinorSet.h"
+
+namespace qmcplusplus
+{
+SpinorSet::SpinorSet():SPOSet(),className("SpinorSet"),spo_up(nullptr),spo_dn(nullptr)
+{
+
+}
+
+SpinorSet::~SpinorSet()
```
ye-luo: There is no need to implement empty destructor.

PR #2106, opened 2019-11-19
Error cmake if MKL is detected via LAPACK/BLAS.

### LINKED ISSUE #2098: Compilation breaks if MKL includes are not available but MKL library was detected
We assume HAVE_MKL=true implies that at least
#include <mkl_service.h>
#include <mkl_types.h>
are available, but this need not be so. Their availability is not guaranteed unless the include paths have been setup, which we do not check for currently.

mkl_service is used to check for nested threading support. 

mkl_types is used in formic for reasons that I do not understand.

ye-luo:
Closes #2098


prckent:
Note that I modified the error message to also mention MKL_ROOT. Conceivably someone has copied the static library but not the mkl specific include files.

PR #2107, opened 2019-11-19
Nexus: fix opt selection

jtkrogel:
@tiihonej will test if this adjustment is correct.


markdewing:
okay to test

tiihonej:
This does the job!

Issue #2105, opened 2019-11-18
Incorrect ion-ion energy in some periodic systems


prckent:
This is an important bug that exists in QMCPACK dating to at least mid-2015, likely earlier.

First noticed for strained phosphene and also reproduced in A-A stacked graphene, for some configurations the ion-ion coulomb energy is incorrectly computed in these periodic systems. The errors can be in eV. To date dense solids appear to be correct - and these are already checked nightly. We are investigating the exact cases when the problem occurs and will update this issue. The bug could be in "plumbing", implementation, or simply unexpected behavior of the implemented method (Natoli/Ceperley).

The ion-ion term should of course be exactly the same as reported by the code that generated the trial wavefunction e.g. Quantum Espresso. For the time being we recommend making cross-checks.

prckent:
Initial report with files https://groups.google.com/forum/#!topic/qmcpack/3Hkt5sWY_e4

prckent:
Fast to run tests were added for this bug in #2116 - bug-deterministic-grapheneC_1x1_pp-vmc_sdj_z10-1-1-ionion and bug-deterministic-grapheneC_1x1_pp-vmc_sdj_z30-1-1-ionion . Both fail. The reference value is currently the QE value.

rcclay:
I'm just going to use this as the dumping ground for ongoing tests:
**Test 1:  Convergence of Ewald ion-ion Interaction in QMCPACK vs. QE:**

Ewald in QE converges really quickly.  I'm doing config #3 from the forum--QE is initialized for a 100 atom cell corresponding to the 5x5x1 tiling...  this is to keep the comparison between the supercell QMCPACK calculation and QE's ewald summation apples-to-apples.  

Ecut (Ry) | Ewald Energy (Ry) | Ewald Energy (eV/cell)
5 | 3497.85849399 | 1903.63148311
10 | 3497.85849326 | 1903.63148271
20 | 3497.85849589 | 1903.63148414
40 | 3497.85849589 | 1903.63148414

This is in contrast to QMCPACK's Ewald3D and Optimized Breakup handlers:

r_c*k_c | Ecut (Ry) | Ewald 3D (eV/cell) | OptBreakup (eV/cell)
15 | 0.904562182 | 1903.61535656 | 1903.71679048
30 | 3.618248729 | 1903.79687315 | 1903.48366227
60 | 14.47299492 | 1903.84280423 | 1903.42547671
[...]

PR #2108, opened 2019-11-20
Add deterministic test for 1x1 2D graphene 

Hyeondeok-Shin:
Add deterministic test for 1X1 2D graphene to reproduce the issue claimed on #2105 . Two deterministic tests for graphene within 'p p n' boundary condition has added. 1. 10 A for z direction. 2. 20 A for z direction using 'vacuum' parameter.

Here are comments for this deterministic tests.

1.  References for ion-ion term are slightly difference between compiler used to build QMCPACK.
eg. intel18 ion-ion : 14.00302796   gcc 6.2 : 14.00303308

2. As reported on the issue, the current QMC ion-ion term shows much difference with the original DFT result. (10A vacuum graphene :  0.025 eV, 20A vacuum graphene : 0.889 eV ion-ion term difference) Therefore, the current references for the test may not right values, so they need to be changed when the issue has resolved.

3. I put references from QMCPACK 3.6, and DFT ion-ion term on CMakeList as the comment. They will be removed when the issue has resolved. Currently, tests for only full-precision non-cuda build have added. The test for the other build will be added also when the issue has resolved.

Hyeondeok Shin


Hyeondeok-Shin:
@prckent I'll add ion-ion term for the other build, and change bconds and references for 'p p p' if values are changed. The reason why I put references as the bad ion-ion term is to make CI pass (alcf-gnu6-mkl18) to merge. I can change the test name to merge. It will break our rule to make a title for the test but temporary until the issue has resolved.

prckent:
OK to test

Hyeondeok-Shin:
@prckent Now, test is available with all builds.
 I changed ion-ion term references to Q.E. output and change test name to issue-deterministic-grapheneC_1x1_pp-vmc_sdj. This name is temporary, and not detected with ctest -L deterministic, so CI will be passed. On the other hand, it is detected with ctest -R deterministic, so test can be seen on cdash for certain build using ctest -R deterministic. (eg. cetus)
  Please let me know if you need additional commit on this.

prckent:
This should end up under "nightly" on cdash, since this part of the dashboard catches tests that  don't match labels (EXCLUDE_LABEL "deterministic|performance|converter|unstable")

Hyeondeok-Shin:
@prckent based on results from @jtkrogel, ion-ion energy difference between Q.E. and QMCPACK becomes significantly larger as increasing z cell size for graphene. Actually, current test shows ~ 0.025 eV difference with 10 A z cell, but 20 A z cell (ppn) shows large 0.88 eV difference. 
 So, if we want to see larger difference than current test, we should increase z cell size, but I don't recommend it with ppp because h5 size will extremely large.  
 Is it good to revive only ion-ion test for p p n with vacuum?  It will give large difference without adding large h5 file. I don't see any significant difference on energy between ppp and ppn for 10 A Z cell, so I think 20 A z cell is the same.

prckent:
Lets focus on ppp since this is what QE calculates. 

We should have looked at the file sizes earlier. Please can you investigate the file sizes with (say) a 10Ry cutoff? The wavefunctions are irrelevant for these tests, so the full 120Ry is not needed. This should allow the current 10A and a new 20A one to have a sensible size. 

Regarding ppn, I would like the conversation around ppp and the 3D bug to proceed further before we decide what to do with mixed boundary conditions.

prckent:
Another option would be to make this a test that actually runs QE, as we do for the qe-atomO test. The reference ion-ion energy could be grabbed from the QE output. An example like this would make it very convenient to add more tests without inflating the size of the repository.

Hyeondeok-Shin:
@prckent I'll do with the smaller cut-off first and make an additional test for larger z size. It can be done quickly. We can talk about Q.E. test if we feel additional z sizes are needed after checking it.

prckent:
Please remake the PR from a fresh branch so we don't have big file in the git history.

Hyeondeok-Shin:
I'll close this PR and remake it when ready.

PR #2111, opened 2019-11-20
Particle set spins updates

camelto2:
This PR adds a few function calls in ParticleSet that are needed for the dynamic spin implementation. Essentially this just adds the capability to update the spin variables when makeMove and makeMoveAndCheck are called, and to accept the move if necessary.  This will be necessary for the spin-orbit ECP evaluation as well as the spin updates in VMC and DMC.


rcclay:
Ok to test.

src/Particle/ParticleSet.cpp
```diff
@@ -402,6 +402,14 @@ void ParticleSet::makeMove(Index_t iat, const SingleParticlePos_t& displ)
   computeNewPosDistTablesAndSK(iat, activePos);
 }
 
+void ParticleSet::makeMove(Index_t iat, const SingleParticlePos_t& displ, const RealType& sdispl)
```
ye-luo: Rename makeMoveWithSpin.
Please maximize code reuse. Instead of copy code.
{
  makeMove();
  activeSpin = spins[iat] + sdispl;
}
Please do similar change in makeMoveAndCheck.

src/Particle/ParticleSet.cpp
```diff
@@ -448,24 +446,10 @@ bool ParticleSet::makeMoveAndCheck(Index_t iat, const SingleParticlePos_t& displ
...
-    else
-    {
-      newRedPos = Lattice.toUnit(activePos);
-      if (!Lattice.isValid(newRedPos)) is_valid = false;
-    }
-  }
-  computeNewPosDistTablesAndSK(iat, activePos);
-  return is_valid;
+    makeMoveAndCheck(iat,displ);
+    activeSpin = spins[iat]+sdispl;
```
ye-luo: The check result needs to be propagated.
```
activeSpin = spins[iat]+sdispl;
return makeMoveAndCheck(iat,displ);
```

camelto2: oops, fixed

PR #2113, opened 2019-11-21
Nexus: Adding occupations card functionality

mcbennet:
This allows the user to set occupations as a string, list, or array.
When a string is input, the occupations variable in the SYSTEM card
is set to the value of the string. When a list or array is input,
then the occupations variable in the SYSTEM card is set to
'from_input' and an OCCUPATIONS card is added and the values from
the list/array are populated wihtin this card.


markdewing:
okay to test

PR #2115, opened 2019-11-21
Update TWF ratio handling

### LINKED ISSUE #2110: Wavefunction gradients broken after change to log representation
Various methods involving wavefunction gradients are broken since migration to log representation (PR #2048 ). This mainly affects force-related observables, such as Pulay terms in the ACForce estimator. For instance, the attached figure demonstrates that the ZV-force (no gradients) is quite okay, but the ZB-correction (has gradients) is just garbage.
CO_F_broken

A special case is the estimator of the NLPP force contributions, which is produced by adding 'forces=yes' attribute to a PseudoPot hamiltonian. In this case even the total energies will be contaminated by the bug, and thus, off by a large fraction.

The apparent fix is to update the gradient methods, such as psi.ratioGrad and whatever else is needed to take gradients of wavefunction components.

ye-luo:
In TrialWaveFunction, delete ratio and put calcRatio in place.
Replace ratioGrad with calcRatioGrad.
All the ratios are now in ValueType.

Fixes #2110 @tiihonej please try this branch.

In #2048 I changed the return ratio value of ratio() and ratioGrad().
The old code returns the absolute value of WFratio in complex and the WFratio value in real.
#2048 changed that behavior always returning the absolute value.
In many places, the ratio^2 is used and there was no problem but the NLPP force routine had a problem. The old diverging behavior between real and complex is not acceptable.
So I'm replacing ratio and ratioGrad with calcRatio and calcRatioGrad returning always ratio in ValueType.

- [x] add a test case when a move cross the node boundary.


tiihonej:
Works like a dream in my test cases. Thanks for the prompt fix!

src/QMCWaveFunctions/TrialWaveFunction.h
```diff
@@ -168,11 +168,6 @@ class TrialWaveFunction : public MPIObjectBase
                         ParticleSet::ParticleGradient_t& fixedG,
                         ParticleSet::ParticleLaplacian_t& fixedL);
 
-  /** functions to handle particle-by-particle update
-   * both ratio and full_ratio will be replaced by calcRatio which will handle ValueType.
-   */
-  RealType ratio(ParticleSet& P, int iat);
```
rcclay: This is removed, but the calcRatio documentation references it.  Can you clean this documentation up a bit?

src/QMCWaveFunctions/TrialWaveFunction.h
```diff
@@ -207,13 +202,13 @@ class TrialWaveFunction : public MPIObjectBase
                           TinyVector<ParticleSet::ParticleGradient_t, OHMMS_DIM>& grad_grad,
                           TinyVector<ParticleSet::ParticleLaplacian_t, OHMMS_DIM>& lapl_grad);
 
-  RealType ratioGrad(ParticleSet& P, int iat, GradType& grad_iat);
+  ValueType calcRatioGrad(ParticleSet& P, int iat, GradType& grad_iat);
```
rcclay: Probably need a documentation comment here.  What are the arguments & what's done to them (looking at grad_iat), and what's the return value.

ye-luo: Since after this change, the naming at TWF and WFC is different. Documentation is definitely needed. I'm working on another PR adjust the accumulation and return precision of calcRatio and calcRatioGrad, I will get all the documentation up.

src/QMCWaveFunctions/TrialWaveFunction.h
ye-luo: This is an oversight.

src/QMCWaveFunctions/TrialWaveFunction.h
prckent: @ye-luo please keen an eye for any missing unit test coverage while doing so.

PR #2116, opened 2019-11-22
Add deterministic test for 1x1 graphene

Hyeondeok-Shin:
Continue from PR #2108.

 Add deterministic tests for 1x1 graphene. Two tests are added for z=10 and z 30 A cell size.
 Through this test, we can check a large difference between QMC and DFT Ewald as claimed on #2105, and it becomes large as increasing z cell size for graphene. (Z =10 A : ~ 0.025 eV, Z = 30 A : ~ 29.47 eV) DFT outputs can be found on ./dft-inputs.

 Hyeondeok Shin


jtkrogel:
The ion-ion energies for the Z=30A graphene cell are revealing on their own:
```
QMCPACK GCC   6.2  ion-ion  82.68669281
QMCPACK Clang 7.0  ion-ion  82.68860463
QMCPACK Intel 18   ion-ion  82.68945173
Quantum Espresso   ion-ion  81.60597872
```
Some part of the ion-ion code in QMCPACK is clearly very sensitive; the ion-ion energies can vary by as much as 3 mHa depending on the compiler used.

jtkrogel:
@Hyeondeok-Shin, does QE's reported Ewald energy depend on the PW energy cutoff?  I notice you use a rather small one for the tests.  This is good for the sake of a small orbital file, but it might be worth checking that QE's reported Ewald energy is converged.

prckent:
@jtkrogel We'll file a bug in QE if the cutoff energy has anything to do with the ion-ion Coulomb energy :-)

Hyeondeok-Shin:
@jtkrogel I reproduced with 120 Ry cut-off, and gives almost same Ewald with 10 Ry. (1e-7 Ry difference on 30 Z size cell, but negligible.) So, we don't have to consider cut-off for ion-ion.

prckent:
OK to test

jtkrogel:
@Hyeondeok-Shin Thanks.  I did not expect there to be any relationship, but wanted to confirm.

PR #2117, opened 2019-11-22
[Hold] Nexus: clean executables

jtkrogel:
Merge only after the 3.9.0 release.

This PR removes code from the `qmca`, `qmc-fit` and `qdens` executables that was copied over from Nexus libraries.

The original intent of the copying was to make the executables stand-alone (no Nexus install needed), and thus increase the chances of successful use.

In moving toward the Python 3 transition, duplicate code is a liability and has been removed.  Instead, the required components are imported from Nexus.  In an attempt to restore the "stand-alone" nature of the executables, relative imports are attempted first.  These also do not require Nexus to be installed to succeed.

### DIFF
--- a/nexus/bin/qdens
+++ b/nexus/bin/qdens
@@ -1,50 +1,126 @@
 #! /usr/bin/env python
 
 import gc
-from numpy import ceil,sqrt,dot,ndarray,arange,savetxt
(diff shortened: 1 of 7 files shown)


jtkrogel:
Subsumed by #2118.  Closing.

Issue #1290, opened 2019-01-02
python3 support for NEXUS and QMCPACK 


prckent:
A transition is needed due to python 2.7 becoming unsupported soon. https://pythonclock.org/

How much work will it take?

- [x] Port core generic module to Python 3
- [ ] Get existing tests to work in Python 3
- [ ] Get a couple of existing workflows to work in Python 3
- [ ] Expand test set in Python 2
- [ ] Get expanded Python 2 tests to work in Python 3
- [ ] Careful pass over remaining code for known 2 vs 3 issues (e.g. division, sorting)
- [ ] Road test Python 3 version prior to release

See #158 (Nexus tests) for detailed breakdown of needed tests.

jtkrogel:
Yes, it does look like the time is indeed now: https://python3statement.org/.  I am in favor of making a clean break to full Python 3 support and dropping Python 2 as this is less work and there is little benefit to continue Python 2 support in the transition.

The changes are not small, e.g. the meaning of division changes and several core class functions (e.g. for dict's) have altered return types.  The extensive use of Nexus' obj class rather than dict may blunt the impact of some of the changes, but I will just need to dig in and try it.

The main work item, prior to full porting, is further rounding out the test set and then porting over the needed changes.  I will edit the initial comment to list work items.

jtkrogel:
I just tried this morning to port the core generic module to Python 3.  This is a good starting point because extensive unit tests exist for this module.

I was successful after a couple of hours, though some of the changes required were revealing.  For example, Python 3 sorting is no longer generic (it now simply falls over when sorting a list containing, say integers and strings).  This feature is relied upon heavily in some I/O and error reporting.  I ended up finding a python2-like fall back sort online that I will try to use instead.  

Another change is in NumPy's class hierarchy (may or may not be linked with Python 3), i.e. numpy integers are no longer derived from Python integers which makes dynamic type checking more complicated.  Perhaps there is a way to capture both without complicating the code, but it will take a little more searching.

jtkrogel:
Anyone who is interested can follow developments in my Python 3 branch of Nexus:

https://github.com/jtkrogel/qmcpack/tree/nx_py3

ye-luo:
@prckent close this? I think we have more recent issues to track this.

prckent:
And we have transitioned! Closing.

PR #2118, opened 2019-11-22
Nexus: Python 3 transition

jtkrogel:
Merge only after the 3.9.0 release.  ~~This PR depends on #2117.~~

With this PR, Nexus transitions totally to Python 3.  If the user attempts to execute Nexus with Python 2, the code aborts and prints instructions to use Python 3.

The process I followed in the transition was:

1. Update the code manually until all tests pass.  Main changes regard use of iteritems(), xrange(), sorted(), keys(), and values() and subtle differences between Python 2 and 3 regarding integer division (integer division now always returns real numbers) and string data types (strings are now all unicode, but some imported modules still proffer data only in ASCII). 

2. Inspect all recommendations made by the `2to3` automatic translation tool on the updated files and include a subset of the changes.  Not all recommendations were worthwhile (e.g. `2to3` is prone to over recommend conversion to lists during iteration), but a number of conversions missed in the manual pass were caught in untested portions of the code.

3. Change all executables and examples to invoke `#! /usr/bin/env python3` and update the versions module to reflect the change to Python 3.  This does not guarantee a successful first execution of Nexus in all environments, but there is no other option that does and at least the intent is abundantly clear.  Invoking a Python 3 interpreter manually with any input script should work fine.  During this process, an additional optional dependency on cif2cell was caught and added to `versions.py` as well as the manual.

4. Run the 2019 workshop examples and compare against the older reference data.  I could not run Quantum Package (the install on my laptop appears to have been broken sometime in the last 7 months), but all the other workflows run successfully.

Once 3.9.0 is in, merging this PR will allow everyone to road test Nexus with Python 3 for a while before the release of QMCPACK 4.0.0 and Nexus 2.0.0.


jtkrogel:
I should mention: though merging must wait for the v3.9.0 release, the PR is complete and is ready for review now.

markdewing:
What prevents this from running on python2?  Besides the explicit python3 interpreter in the #! lines?

jtkrogel:
In `versions.py`:
```
# Require Python 3
if python_version_tuple()<('3','0','0'):
    print('\nNexus is compatible only with Python 3.\n  You attempted to run with Python version {}.\n  Please rerun with Python 3.\n'.format(python_version()))
    sys.exit(1)
#end if
```

markdewing:
If that check is removed, what fails?

jtkrogel:
At least two things: (1) changes in binary/ascii representation of strings, (2) `print` is now used as a local keyword in places in `qmcpack_input` (it is not a global keyword in Python 3). 

(1) will cause at least runtime errors for many file I/O operations, including use of the `execute` function which is used extensively to launch the simulation codes.

(2) will cause the interpreter to fail with a syntax error when `qmcpack_input` is imported, which should happen anytime `nexus` is imported.

jtkrogel:
This should go in now, ahead of the new year.

@markdewing Please finish your review when you get a chance.

ye-luo:
Please update #1290 after merging

jtkrogel:
@prckent, is there a reason this PR should go on hold?

prckent:
We had discussed not switching to Python 3 before the release. For efficiency lets discuss the release on Slack.

prckent:
This will be decided on Monday when a few more people are back from holidays, but after discussing the situation it seems that the preferred route is to make this transition now, just ahead of the v3.9.0 release.

jtkrogel:
Merge conflicts are fixed now.

prckent:
Thanks.

While this PR doesn't have an approved review yet, based on the discussions above & elsewhere, I believe that there are no concerns with it. The PR will be merged >=2pm today unless a good reason for delay is presented.

People will likely need to update or reconfigure their software stack to install and run develop after this is in.

jtkrogel:
This PR covers the Nexus transition to Python 3.  There are likely other outstanding needs for QMCPACK configuration via Python 3 that need separate PR's.

prckent:
OK to test

PR #2121, opened 2019-11-23
[WIP] Debug dmc branching

PDoakORNL:
Population control now seems to work. Thanks to @ye-luo for suggesting defensive calls to ParticleSet::update and TrialWaveFunction::evaluateLog. This seem to be necessary because I've failed to reproduce the sequence of side effect and serialization and deseriatlization from Buffers as in the legacy driver.  I think actually changes in ParticleSet and TrialWaveFunction are a better idea that trying to perfect the simulation of the legacy state machine system.


ye-luo:
You rebased the code when merging the last PR but this branch is created before the rebase.
So please rebase on the develop.

PDoakORNL:
To much divergence creating synthetic PR's

PR #2126, opened 2019-11-27
Add Bspline converter for pyscf and statistical integration test for 1x1x1 diamond

Hyeondeok-Shin:
Continue from #1899. Worked with @TApplencourt.

 Add LCAO -> bspline converter for pyscf and its statistical integration test.
 Pyscf input template to use converter can be found in ./tests/solids/diamondC_1x1x1-Gaussian_pp/dft-inputs/diamondC_1x1x1_bspline.py. It is using the same template with the LCAO converter in pyscf input.

 Converter test is not added yet. H5 and xml are generated via python script directly, so I'd like think the best strategy to provide reasonable converter test and update for it with separated PR if sounds OK.
 Currently, bspline converter worked in the gamma point calculation, but implementation on k-point scf and multi-determinant is on the plan for the future.

 Hyeondeok Shin


prckent:
My preference would be that we add converter tests from the start, similar to QE tests or how the O atom test is done. Then we won't need any 5MB h5 files in the git (as here) and can add many more converter tests without running into size concerns.

If this is needed quickly we can find a workaround.

Hyeondeok-Shin:
@prckent My best scenario was to show LCAO and bspline is obviously in the same references in integration test, but I can replace it to the converter test. 
   I'll add pyscf test similar to the pw.x and pw2qmcpack.x test. This test will include both LCAO and bspline test to check whether h5 is normally generated through PyscfToQmcpack.py and they are in the same references.  
   With this manner, we'll able to check starting point for pyscf and we don't have to put large h5 to archive. I'll make additional commit for the pyscf test if it sounds good for you and @anbenali.

jtkrogel:
Quick question: the code here looks like it is Pyscf specific.  How hard would it be to instead convert the Gaussian h5 file to ESHDF (B-spline format only)?

prckent:
@jtkrogel Good question. This would e.g. open up use of Crystal wavefunctions.

Hyeondeok-Shin:
@jtkrogel Good point. I'm also thinking converting Gaussian h5 to ESHDF directly can be the final form of the converter. My plan is to work on it after finishing full implementation for Pyscf first. When I make sure the converter works for all possible LCAO template for pyscf (complex, multi-det, etc.), I'll be working to replace it to h5 converter. Now, the converter perfectly works on gamma-point calculation for pyscf, and I'd like to update it step-by-step.

src/QMCTools/PyscfToQmcpack_Spline.py
```diff
@@ -0,0 +1,947 @@
...
+##
+## File developed by: Thomas Applencourt, applencourt@anl.gov,  Argonne National Laboratory
+##                    Hyeondeok Shin, hshin@anl.gov, Argonne National Laboratory
+##                    Yubo "Paul" Yang, yubo.paul.yang@gmail.com, University of Illinois Urbana-Champaign
+##                    Anouar Benali, benali@anl.gov, Argonne National Laboratory
+##                  
+## File created by: Thomas Applencourt, applencourt@anl.gov, Argonne National Laboratory
+#######################################################################################
+
+from __future__ import print_function
```
markdewing: If this is python 3 only (as a check below indicates), this import is not needed.

src/QMCTools/PyscfToQmcpack_Spline.py
```diff
@@ -0,0 +1,947 @@
...
+  gvecs, eig_df = save_eigensystem(mf, save=False)
+
+  # generate wave function file
+  # ================================================
+  generate_pwscf_h5(loc_cell,gvecs,eig_df,h5_fname)  
+
+
+  # generate QMCPACK input file
+  # ================================================
+  from lxml import etree
```
markdewing: The lxml package is not a default python package.   The import should happen earlier, and maybe be checked to print a nicer error message.

prckent: Good catch. Can we not avoid this? It complicates installation, testing, user support etc. further.

src/QMCTools/PyscfToQmcpack_Spline.py
TApplencourt: You right, let me fix that

src/QMCTools/PyscfToQmcpack_Spline.py
TApplencourt: Honestly, I agree. 

This work was based on the good work of Yang, Paul. I just updated and fixing some bugs on this code but he did the majority of the work.
He was using already using `lxml`, so I decided to follow is steps.
I can try to see how hard it's to go directly from the native python `xml` parser. 

On the other hand, to use this script people need to install PySCF first, so they can just install `lxml` at the time :).   But I understand, for the qmcpack point of view, having one more dependency is tedious.

TApplencourt: Honestly, I agree.
This work was based on the good work of Yang, Paul. I just updated and fixing some bugs on this code but he did the majority of the work.
He was using already using lxml, so I decided to follow is steps.
I can try to see how hard it's to go directly from the native python xml parser.
On the other hand, to use this script people need to install PySCF first, so they can just install lxml at the time :slightly_smiling_face:. But I understand, for the qmcpack point of view, having one more dependency is tedious.

Reposted: because of bugs

prckent: Since it is a PySCF dependency it is not too bad.

Hyeondeok-Shin:
Now large bspline h5 is removed, and pyscf workflow test is added for LCAO and Bspline test.
 The workflow test will be automatically added when related modules (pyscf, QmcpackToPyscf, etc.) exist. (It wouldn't be added within Python2.)

 I think this PR should be reopened to make git history clean, so I'll do when @prckent and @markdewing are happy with current PR.

prckent:
I don't have any bandwidth to test this, but it LGTM. I assume it doesn't run or create significant new output when PySCF or dependencies are not available.

Hyeondeok-Shin:
This test won't be added and run if PySCF or dependencies are not available. I'll close this PR when @markdewing is ok.

src/QMCTools/PyscfToQmcpack_Spline.py
```diff
@@ -0,0 +1,936 @@
...
+
+def pyscf2qmcpackspline(cell,mf,title="Default", kpts=[], kmesh=[],  sp_twist=[]):
+  import sys, re
+
+  Restricted=True
+  PBC=False
+  Gamma=False
+
+  # Python version check
+  if sys.version_info < (3, 2):
```
markdewing: When loading this file, Python 2 gives a syntax error on the f-strings (around line 75).  This version check will never be executed.   f-strings were introduced in Python 3.6, so presumably the same thing will happen for python 3 versions 3.5 and lower (but I didn't check, only ran under Python 2).

TApplencourt: I can fix the "fstring" issue. 
Do you have a minimun python version for QMCPACK? 
I can either remove all the "fstring" (I will be tedious but I understand) or just fix the warning.

TApplencourt:
The code has been updated to remove "fstring".

markdewing:
Looks good.  Thanks for removing the fstring.   It makes formatting more tedious, but it allows the code to run under more versions of Python.

Hyeondeok-Shin:
This PR will be closed and reopened to make git history clean.

PR #2127, opened 2019-12-02
Add Wave function API's for spin moves and spin gradient computation.

rcclay:
This PR adds the following two methods to TrialWaveFunction

1. calcRatioSpinGrad:  the exact analogue of calcRatioGrad, but for the spin gradient.
2. evalSpinGrad:  exact analogue of evalGrad, but for the spin gradient.  

Default behavior is an abort due to missing implementation, but DiracDeterminant gets a nontrivial implementation for these methods, and J1OrbitalSoA, J2OrbitalSoA, and JeeIOrbitalSoA get trivial implementations under the assumption that they are spin-independent in normal use cases (which is the intended case).  

Back end work has been done to DiracDeterminant, SpinorSet, and ParticleSet to make sure these functions work properly, in addition to preserving the expected behavior of evaluateLog, ratio, and ratioGrad.  

Unit test is included.  This checks the correct evaluation of the values, position and spin gradients, and laplacians of a Slater determinant of spinors.  It also tests the ratio and gradient computations for a proposed single particle move with and without rejection.


ye-luo:
Better to have
evalGradWithSpin calling evalGrad + spinor operations.
calcRatioGradWithSpin calling calcRatioGrad + spinor operations.
At the driver level there is only a single call.

rcclay:
So there's a goofy thing going on in my mixed-precision build where the spinor determinant unit test in src/QMCWaveFunctions/tests/test_dirac_det.cpp segfaults.  If I run it through a debugger, it not only doesn't crash, but passes all the tests.  I think its something to do with line test_dirac_det.cpp:684 and DiracDeterminant's call to evaluate_spin(), but nothing obvious is jumping out at me. Nothing seems wrong with the full precision build. Any ideas?

rcclay:
Found it.  Pushing up a fix.

rcclay:
@ye-luo The reason for this choice of implementation in the Jastrows is I want the default behavior of WaveFunctionComponent to be a hard abort if it gets called in a spin computation, rather than to assume by default that the component is spin-independent.  It's my way of making the developer/user think before they start running with random WaveFunctionComponents.  That said, if you would rather move the Jastrow implementations to WaveFunctionComponent, I'll do it.

ye-luo:
Most of the WaveFunctionComponent derived classes are spinor agnostic Jastrow factors. Requiring a dummy implementation is a burden. If someone would like to try a Jastrow with the spin determinant. I feel it is better to let it go instead of requiring fixing the code by adding a dummy implementation. On the other hand, single det and multi det class need some protection. They are explicitly required to use spins if requested. I just noticed that you didn't touch DiractDeterminantBase which connects SlaterDet and DiracDeterminant. So Please implement interfaces and move the abort there.
Please also add abort in multi determinant which will be eventually removed once we implement this feature.

rcclay:
@ye-luo Default behavior is now to assume spin-independence for WaveFunctionComponent, unless it derives from DiracDeterminantBase, in which case it aborts in the absence of an implementation.  

I also ran clang-format on everything I touched.

Issue #2083, opened 2019-11-06
NLPP derivatives in optimization need SoA


jtkrogel:
Currently only supported in AoS.

jtkrogel:
It turns out that NLPP derivatives are sometimes crucial for correct optimization.  Below is a comparison plot of the change in the total energy for Li2 during Jastrow optimization with and without the NLPP derivatives on.

![energy_vs_series](https://user-images.githubusercontent.com/24547905/69453910-f2b9a500-0d32-11ea-8ef6-1c4d9d410d93.png)

For work on forces, we will have to use the AoS code for optimization until SoA is fixed for this.

jtkrogel:
Who is looking into fixing this issue?

ye-luo:
OK. I will take care of it next week.

prckent:
Still investigating, but I think another user might have been caught by this for a heavier system. 
Since SoA is the default, it means that new users will have worse wavefunction optimization than they should. Once implemented, I think we should make sure that NLPP derivatives are on by default in all examples, change the code default etc. Better to be slower but reliable.

jtkrogel:
I agree 100% that these should be on by default.

ye-luo:
1. This feature should be default.
2. implementing optimized virtual move code path is necessary to have painless 1.

PR #2128, opened 2019-12-04
Nexus: optimize with NLPP derivatives by default

jtkrogel:
Update Nexus to write optimization blocks with NLPP derivatives on by default.

This PR depends on the resolution of issue #2083.


tiihonej:
I believe the hold could be lifted, since PR #2179 has been merged already? I tested the update in my own projects and it seems to work correctly.

prckent:
OK to test

PR #2131, opened 2019-12-05
Add dense hamiltonian

fdmalone:
Add a dense implementation of  factorised ERI (3-index) tensor for real hamiltonians. Avoids O(N^2M^2) memory overhead of storing half-rotated (ik|jl) used in the sparse code at a slightly increased computational cost. Ideal for larger molecules where sparsity in ERI tensor in MO basis is typically small. Works on GPU.


src/AFQMC/HamiltonianOperations/Real3IndexFactorization_batched_v2.hpp
```diff
@@ -0,0 +1,629 @@
+//////////////////////////////////////////////////////////////////////
```
fdmalone: Why is there _v2 version for these files. Multiple versions? Can we just include one?

external_codes/mpi_wrapper/mpi3/shared_window.hpp
```diff
@@ -20,9 +20,9 @@ namespace mpi3{
 
 template<class T>
 struct shared_window : window<T>{
-//	shared_communicator& comm_;
-	shared_window(shared_communicator& comm, mpi3::size_t n, int disp_unit = alignof(T)) //: //sizeof(T)) : // here we assume that disp_unit is used for align
-		//window<T>()//, comm_{comm}
+	shared_communicator& comm_;
```
fdmalone: Please check merge conflict resolution.

src/AFQMC/Hamiltonians/RealDenseHamiltonian.cpp
```diff
@@ -0,0 +1,264 @@
+#include<cstdlib>
```
fdmalone: Has the bug for distributed Hamiltonians been fixed?

src/AFQMC/HamiltonianOperations/Real3IndexFactorization_batched_v2.hpp
mmorale3: Possibly. The original version is faster, but uses about 30% more memory. The v2 version is probably the one to keep, due to memory limitations on gpu. With c++17, we can keep both versions on a single file.

src/AFQMC/Hamiltonians/RealDenseHamiltonian.cpp
mmorale3: not yet

ye-luo:
Okay to test

PR #2136, opened 2019-12-06
Improve HDF5 cmake settings

ye-luo:
HDF5_PREFER_PARALLEL is used by FindHDF5 to select parallel HDF if both serial and parallel HDF5 library are found. HDF5_PREFER_PARALLEL was hard-coded to 1 for MPI builds. This is the right choice for good performance but it can cause problems.

On Ubuntu, both parallel and serial hdf5 library can be installed from package manager. However the parallel version is tided to OpenMPI or MPICH. So I need a way to disable parallel HDF5 searching because I'm not using the correponding MPI library.

So far, the default for MPI build is
HDF5_PREFER_PARALLEL=1, ENABLE_PHDF5=1
for non-MPI build is
HDF5_PREFER_PARALLEL=0, ENABLE_PHDF5=0
I add extra protections for wrong combination of flags. Both HDF5_PREFER_PARALLEL and ENABLE_PHDF5 are not allowed wtih non-MPI build.
MPI build always warns users for performance loss if parallel HDF5 feature is not enabled.


prckent:
Good to have this functionality. Another common use case is when you have access to both serial and parallel HDF5 but the parallel one is built with the wrong MPI library.

PR #2137, opened 2019-12-09
Increase tolerance for k-shell identification in LongRange potentials

rcclay:
This is a temporary fix for Issue #2105 until a more general solution can be discussed.


prckent:
OK to test

jtkrogel:
It would be very interesting to test the limits of the code following this change, e.g. find the size of cell for which the LR code fails against the reference.

src/LongRange/KContainer.cpp
```diff
@@ -250,10 +250,10 @@ void KContainer::BuildKLists(ParticleLayout_t& lattice, bool useSphere)
   //Update a record of the number of k vectors
   numk = kpts_tmp.size();
   std::map<int, std::vector<int>*> kpts_sorted;
-  //create the map: use simple integer with resolution of 0.001 in ksq
+  //create the map: use simple integer with resolution of 0.00000001 in ksq
   for (int ik = 0; ik < numk; ik++)
   {
-    int k_ind = static_cast<int>(ksq_tmp[ik] * 1000);
+    int k_ind = static_cast<int>(ksq_tmp[ik] * 100000000);
```
jtkrogel: Should we also change to `long int` here?  I don't see why we are limiting ourselves to `int` for an operation that has proved delicate.

prckent:
Well, this update might have already pushed the limits because the CI is failing in Coulomb related unit tests! Perhaps a cast is needed. I suggest a hard check for possible overflow as well.

Simply not using a integer map and using the actual k-vector length as a float would have much less risk. ("Premature optimization causes bug"). To be discussed later once we have some less wrong / more safe code in production.

jtkrogel:
We could definitely use a different algorithm to populate the shell table.  Simply sorting the `ksq_tmp` vector and then running a single pass over it to check the difference between adjacent values vs. a real valued tolerance should allow a more robust tabulation into shells.

Another issue is more fundamental: the radial difference between adjacent shells in any type of rectilinear grid will rapidly decrease with radius, potentially leading to precision problems.  It may or may not be an issue in practice, though.

src/LongRange/KContainer.cpp
prckent: Agree. Since numk can be huge, as can all the other integer indices, why not have long ints throughout?

prckent:
@rcclay Your update arrived as I was commenting. Use a long long for the numk loop as well? All the integer indices here can be large.

rcclay:
One thing I'd like to establish is whether there are any actual computational savings by doing this grouping by k-shell.  Looking at the actual long-range evaluation, this is done by a sum over rho_k where rho_k is not radially symmetric...  the code first sums over every rho_k in the shell, then multiplies the sum by v_|k|.  While this is fine, I'm not sure how much we save by doing it this way as opposed to summing over all k-points.

rcclay:
@prckent Here's my thinking:  provided that int is 32 bits, this will accommodate a k-space grid of roughly 1024x1024x1024 before overflow.  If we don't think this is big enough, I'll go ahead and change the loop variable to 64 bit long long.

prckent:
@rcclay If we can improve the code by simplifying it as you suggest that would be a big improvement. However beware that there could be some numerical reasons why things are done as they are. We can discuss after getting this fix in.

prckent:
@rcclay long long or error trap the too many k for range of int situation.

rcclay:
Rather expectedly, the number of k-shells is pretty sensitive to tolerance in mixed precision.  I will need to think about this.

ye-luo:
My two cents. 1. Need an immediate fix. 2. Need a long term fix with better thinking. Feel headache with the mapping. QE has g vector sorting which can be very slow. Parallel sorting is tricky. Need to reevaluate the cost and see if some current optimization is worthwhile.

ye-luo:
I don’t think the key problem is sorting but some unsafe assumptions made in the past.

rcclay:
OK.  Making the mixed precision build work with this change is going to be a bit nontrivial.  I think this workaround might be better implemented by a user rather than being pushed into the main code...  at least until a proper long-term fix is in place.

prckent:
Where is the problem? Can you upgrade the precision to be double/full throughout the setup code? I would prefer that the mixed precision code is slower but less dangerous in v3.9.0

rcclay:
Problem is that the KContainer is PosType, which means it's float in multiprecision.  If you make the determination of shell based on |k|^2, you don't really have the precision to resolve much past where the threshold is currently set.  This can be worked around by changing the types of KContainer and |k|^2, but it starts looking like a lot of work for what was supposed to be a quick workaround...

prckent:
Confirming that this builds with Intel2019.

The "bug" tests for z10 nearly pass while for z30 the errors are still large. Is this consistent with your runs? z30 seems very slow to converge even with changing LR_dim_cutoff.

Unless you have other suggestions I think we should merge this, putting mainline in a much safer state. 
```
[pk7@oxygen build_intel2019]$ ctest -VV -R "bug-deterministic-grapheneC_1x1_pp-vmc_sdj" |grep -A 10 "ion-ion energy"
1201:   Reference ion-ion energy: 14.002118126747
1201:   QMCPACK   ion-ion energy: 14.003381954628
1201:             ion-ion diff  : 0.0012638278801305
1201:             diff/atom     : 0.00063191394006523
1201:             tolerance     : 0.0003
1201:
1201: Please try increasing the LR_dim_cutoff parameter in the <simulationcell/>
1201: input.  Alternatively, the tolerance can be increased by setting the
1201: LR_tol parameter in <simulationcell/> to a value greater than 0.0003.
1201: If you increase the tolerance, please perform careful checks of energy
1201: differences to ensure this error is controlled for your application.
1201:
--
1203:   Reference ion-ion energy: 81.605978721118
1203:   QMCPACK   ion-ion energy: 82.688199774306
1203:             ion-ion diff  : 1.0822210531883
1203:             diff/atom     : 0.54111052659414
1203:             tolerance     : 0.0003
1203:
1203: Please try increasing the LR_dim_cutoff parameter in the <simulationcell/>
1203: input.  Alternatively, the tolerance can be increased by setting the
1203: LR_tol parameter in <simulationcell/> to a value greater than 0.0003.
1203: If you increase the tolerance, please perform careful checks of energy
1203: differences to ensure this error is controlled for your application.
1203:
Errors while running CTest
```

rcclay:
@prckent I've reproduced the failing tests.  It looks like for the Z30, I can't increase LR_DIM_CUTOFF much past 20 before the optimized breakup starts turning up singular values in the SVD solve for the breakup coefficients.  I will check this against the Ewald handler and see what happens.

rcclay:
The plot thickens.  Here's what I get when I swap out the Optimized Breakup with the Ewald3D handler.  This indicates to me that there might be another issue in the optimized breakup handler.  

LR_DIM_CUTOFF  ION-ION ENERGY
15  81.603089627494
30 81.605597915

prckent:
@rcclay I suggest we merge this now since it fixes at least one bug, if not the slow convergence of the optimized breakup. Do you agree?

rcclay:
I agree.

PR #2139, opened 2019-12-10
Update converter tests to be compatible with Python 3

markdewing:
Needed for #2126 to run correctly.

The spaces/tabs issue breaks any converter test running under Python 3 (or at least Anaconda Python 3.7.3, which is what I tested with)


jtkrogel:
This looks OK.  Do the tests still work with Python 2?

markdewing:
The code still works with Python 2.

PR #2140, opened 2019-12-10
Nexus: Default Hybrid Rep params

mcbennet:
This PR removes the setting of spline_radius and spline_npoints values
by nexus given that QMCPACK sets its own defaults for these quantities.

Thank you @jtkrogel for your assistance with this PR.


markdewing:
okay to test

PR #2144, opened 2019-12-11
Fix reference Ewald

### LINKED ISSUE #2142: Reference Ewald is incorrect
See NiO 32 atom cell. QE reported -478.59629996 Ry for 4 atom cell.
https://cdash.qmcpack.org/CDash/testDetails.php?test=6863360&build=94127
```
  Reference ion-ion energy: -1489.4328628572
  QMCPACK   ion-ion energy: -1914.3871767605
```
The correct value should be `-478.59629996 / 2 * 8 = -1914.38519984 Ha`

jtkrogel:
closes #2142.

The sums in the reference implementation were terminated whenever the contribution over a cubic shell was less than a specified tolerance.  In the cubic NiO series for some cell sizes, the contribution over a cubic shell can vanish.  This lead to the behavior documented in #2142.

The ion-ion energy per 4 atom NiO cell is listed below for the original reference and native QMCPACK implementations:
```
         Reference         QMCPACK        
 4 atom  -239.29814998004  -239.2981423734
 8 atom  -239.29814997992  -239.2981710054
16 atom  -239.29814997956  -239.2980872025
32 atom  -186.17910785715  -239.2983961978
64 atom  -217.33657911019  -239.2981926791
```

The fix is to instead sum the absolute value of each of the point contributions over the shell and terminate only when this absolute sum over the shell is below the tolerance.  Following the fix, the ion-ion energies become:
```
         Reference         QMCPACK        
 4 atom  -239.29814998008  -239.29814237331
 8 atom  -239.29814998008  -239.29817100545
16 atom  -239.29814998007  -239.29808720243
32 atom  -239.29817132488  -239.29839619770
64 atom  -239.29814998041  -239.29819267906
```

Notice that for the 32 atom cell, the reference value still deviates from the baseline somewhat, though it returns for the 64 atom case.  I can't tell whether this case is simply more sensitive or whether the deviation is due to some differences in internal numerics in the representation of the cell and/or atomic positions for this particular tiling.  QMCPACK also shows the largest deviation from the reference for this case:

IonIon_err

On a separate note, I observe that the reference Ewald check takes on the order of minutes for the 64 atom case and might become prohibitive for larger system sizes.

### DIFF
--- a/src/QMCHamiltonians/EwaldRef.h
+++ b/src/QMCHamiltonians/EwaldRef.h
@@ -199,11 +199,12 @@ class KspaceEwaldTerm
 template<typename T>
 RealType gridSum(T& function,bool zero=true,RealType tol=1e-11)
 {
-  RealType dv = 1e99;
-  RealType v  = 0.0;
-  IntType im  = 0;
-  IntType jm  = 0;
-  IntType km  = 0;
+  RealType dv  = 1e99;
+  RealType dva = 1e99;
+  RealType v   = 0.0;
+  IntType im   = 0;
+  IntType jm   = 0;
+  IntType km   = 0;
   IntVec  iv;
   // Add the value at the origin, if requested.
   if(zero)
@@ -212,9 +213,10 @@ RealType gridSum(T& function,bool zero=true,RealType tol=1e-11)
     v += function(iv);
   }
   // Sum over cubic surface shells until the tolerance is reached.
-  while(std::abs(dv)>tol)
+  while(std::abs(dva)>tol)
   {
-    dv = 0.0; // Surface shell contribution.
+    dva = 0.0;
+    dv  = 0.0; // Surface shell contribution.
     // Sum over new surface planes perpendicular to the x direction.
     im += 1;
     for(IntType i: {-im,im})
@@ -224,7 +226,9 @@ RealType gridSum(T& function,bool zero=true,RealType tol=1e-11)
           iv[0] = i;
           iv[1] = j;
           iv[2] = k;
-          dv += function(iv);
+          RealType f = function(iv);
+          dv  += f;
+          dva += std::abs(f);
         }
     // Sum over new surface planes perpendicular to the y direction.
     jm += 1;
@@ -235,7 +239,9 @@ RealType gridSum(T& function,bool zero=true,RealType tol=1e-11)
           iv[0] = i;
           iv[1] = j;
           iv[2] = k;
-          dv += function(iv);
+          RealType f = function(iv);
+          dv  += f;
+          dva += std::abs(f);
         }
     // Sum over new surface planes perpendicular to the z direction.
     km += 1;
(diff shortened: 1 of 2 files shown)


markdewing:
okay to test

jtkrogel:
The 32 atom cell really is just a hard case:
```
tol     value                  deviation from N=4 reference
1e-6   -239.295157742675       0.002992237405
1e-7   -239.2974222785         0.000727701580
1e-8   -239.29838682535       -0.000236845270
1e-9   -239.29807692635        0.000073053730
1e-10  -239.298171324875      -0.000021344795
1e-11  -239.2981440745         0.000005905580
1e-12  -239.2981515267875     -0.000001546707
1e-13  -239.2981505941375     -0.000000614057
```
The residual error decreases like the square root of the tolerance.

src/QMCHamiltonians/EwaldRef.h
```diff
@@ -199,11 +199,12 @@ class KspaceEwaldTerm
 template<typename T>
 RealType gridSum(T& function,bool zero=true,RealType tol=1e-11)
 {
-  RealType dv = 1e99;
-  RealType v  = 0.0;
-  IntType im  = 0;
-  IntType jm  = 0;
-  IntType km  = 0;
+  RealType dv  = 1e99;
```
ye-luo: if RealType is float, there will be a compiler warning. Use std::numeric_limits<RealType>::max Instead.

jtkrogel: Done.

jtkrogel:
@ye-luo Any reason not to use `double` instead of `QMCTraits::RealType` in EwaldRef?

ye-luo:
No reason not to use double but don't use double directly, use FullPrecRealType.

src/QMCHamiltonians/EwaldRef.h
```diff
@@ -43,7 +43,7 @@ enum
 /// Type for integers
 using IntType     = int;
 /// Type for floating point numbers
-using RealType    = QMCTraits::RealType;
+using RealType    = QMCTraits::FullPrecRealType;
```
ye-luo: I think this is a very confusing alias. Please use FullPrecRealType or FPRT if you prefer something short.

jtkrogel: I'm tempted to just use `double` so that EwaldRef is fully decoupled from the rest of the QMCPACK configuration.

If you prefer, I'll just use `int_t` and `real_t` to avoid overlap with the naming of the normal hierarchy of QMCPACK types.

ye-luo: That is better. Please do so. Because it is in the header and namespace, it can propagate in principle.

prckent: I like the logic of the non-overlap / independent route as well.

jtkrogel: Done.

jtkrogel:
I don't have further updates planned for this PR, so, if everyone is good, it can go in.

PR #2145, opened 2019-12-11
Spack Regression Test Update Dec 2019

naromero77:
The main change are:
- Bash function to make the code more compact and cleaner.
- Separate logging of Spack CONFLICTS and FAILURES.
- General hardening of script to account for bash idiosyncrasies.


markdewing:
okay to test

PR #2146, opened 2019-12-11
Afqmctools updates

fdmalone:
Minor updates for gamma point conversion script and FCIDUMP header format.


markdewing:
okay to test

PR #2147, opened 2019-12-11
VMC with dynamic spins

camelto2:
This PR adds spins as a quantum variable for VMC calculations with PbyP updates. This satisfies part of  Issue #1770, namely the VMC part. The DMC implementation will be nearly identical. 

Whether or not spin moves are used is simply added as a flag in the VMC xml input with the parameter SpinMoves="yes", and the dynamics of the spins can be determined by a spinmass also given in the input xml. This effectively changes the timestep for the spin variables relative to the spatial variables, as described in  Melton, Bennett, & Mitas J. Chem. Phys. 144, 244113 (2016); https://doi.org/10.1063/1.4954726 

This also corrects a type mismatch between function arguments in ParticleSet, which were missed by PR #2127 since activeSpinVal is Scalar_t

There is also a unit test making sure that the spin variable is appropriately changed with an accepted move.


markdewing:
okay to test

camelto2:
previous failed test was due to inconsistency in ParticleSet.h and ParticleSet.cpp

Should now build correctly in mixed precision.

src/QMCDrivers/VMC/VMC.cpp
```diff
@@ -49,6 +50,12 @@ VMC::VMC(MCWalkerConfiguration& w,
   m_param.add(UseDrift, "useDrift", "string");
   m_param.add(UseDrift, "usedrift", "string");
   m_param.add(UseDrift, "use_drift", "string");
+  m_param.add(SpinMoves, "spinMoves", "string");
+  m_param.add(SpinMoves, "spinmoves", "string");
+  m_param.add(SpinMoves, "spin_moves", "string");
```
ye-luo: We will need to set official input tag naming rules. Only pick one and error others as unidentified input. Not a problem of this PR, we will settle that latter when deleting drivers.

prckent: Full agreement on this one. Please pick a single tag.

camelto2: fixed

jtkrogel:
Mind using the same convention for "spinmass" and "spinMoves"?  I know there isn't consistency elsewhere, but it might be beneficial between these related variables.

camelto2:
> Mind using the same convention for "spinmass" and "spinMoves"? I know there isn't consistency elsewhere, but it might be beneficial between these related variables.

Done, everything is now consistent for triggering the spin moves and respective variables

Issue #2143, opened 2019-12-11
MPI wrapper and AFQMC breaks Intel complex build


prckent:
https://cdash.qmcpack.org/CDash/viewBuildError.php?buildid=94091

correaa:
It is strange, I can reproduce the error using the EDG front-end only when I *add* the move in `comm_{/*std::move(*/other.comm_/*)*/}` . In any case adding the `std::move` it is not a solution in my opinion.

This is in part why I don't like to hold a reference to the communicator in the `window` (and I don't actually https://gitlab.com/correaa/boost-mpi3/blob/master/shared_window.hpp#L23). 
Perhaps switching to a communicator pointer instead of a reference is the way to go in the custom version of MPI3.

mmorale3:
Alfredo, 
Did you fix the issue with the version of the constructor in communicator that blocked globally (the one that took a group as argument)?
I can't get rid of the reference (or pointer) to the communicator in shared_window with this problem.

correaa:
On Wed, Dec 11, 2019 at 10:56 AM mmorale3 <notifications@github.com> wrote:

> Alfredo,
> Did you fix the issue with the version of the constructor in communicator
> that blocked globally (the one that took a group as argument)?
>

yes. Indirectly, in the sense that the reference is not needed.

> I can't get rid of the reference (or pointer) to the communicator in
> shared_window with this problem.
>
Great. I think that is the sane approach, if there is an obvious
performance degradation we can keep looking for solution.

As I tried to explain before, my idea is that any performance cost of
creating a communicator is amortized by the fact that a barrier is needed
anyway (even if one has the communicator already) in the general case
(after any non-trivial operation).

I might be wrong but it is reasonable to think that the cost of a barrier,
of a broadcast or a communicator creation are all of the same order. Is it
also possible that the mpi environment could be caching some frequently
used communicator setup (especially here that there is an actual instance
of the communicator in existence)

Thanks,
Alfredo

mmorale3:
Alfredo,

You are still using MPI_Comm_create, this blocks on world.

correaa:
> Alfredo,
> 
> You are still using MPI_Comm_create, this blocks on world.

Use the communicator constructor (from group) with the extra integer.

https://gitlab.com/correaa/boost-mpi3/blob/master/communicator.hpp#L2938

[...]

PR #2153, opened 2019-12-16
removed need for comm& in shared_window

fdmalone:
Should address #2143


markdewing:
okay to test

prckent:
Thanks. Expected outcome: working intel2018+complex+afqmc builds on cdash.

prckent:
It worked: https://cdash.qmcpack.org/CDash/buildSummary.php?buildid=95485

PR #2154, opened 2019-12-16
Nexus: Add recenter option to add_kpoints

mcbennet:
In the case that the user would prefer not to recenter the set of kpoints


markdewing:
okay to test

PR #2157, opened 2019-12-17
Update coverage scripts to be Python 3 compatible

markdewing:
Most of the changes are print statement to print function.
The iterkeys and iteritems methods were changed to keys and items.   This has the potential to affect performance because these methods return a list rather than an iterator, but that doesn't seem to be an issue when running the coverage tests.

The updated code works with python 2 and python 3

Python 3 - https://cdash.qmcpack.org/CDash/viewCoverage.php?buildid=95216
Python 2- https://cdash.qmcpack.org/CDash/viewCoverage.php?buildid=95318

### DIFF
--- a/tests/coverage/compare_gcov.py
+++ b/tests/coverage/compare_gcov.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 
 import argparse
 from collections import OrderedDict, namedtuple, defaultdict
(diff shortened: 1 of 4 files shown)


prckent:
Thanks Mark for these updates and for the successfuly uploads. I'll debug the situation on oxygen once these changes are merged.

PR #2159, opened 2019-12-18
Fix dense hamiltonian

fdmalone:
Fixes bug with chunking in low memory algorithm and distributed hamiltonian for dense hamiltonian.


ye-luo:
Okay to test

PR #2160, opened 2019-12-18
Refactor spline adoptors

ye-luo:
All the changes are inside `src/QMCWaveFunctions/BsplineFactory`

The old code has the following structure
```C++
template<typename SplineAdoptor>
struct BsplineSet : public SPOSet, public SplineAdoptor
```
I was force to write pass through functions and it becomes hard to write fallback functions with a reasonable resolution.
Second, the template parameter TT brings little extra value and SplineAdoptorBase role is not clear
```C++
template<typename ST, typename TT>
struct SplineC2RSoA : public SplineAdoptorBase<ST, 3>
```
So I simply fuse both BsplineSet and SplineAdoptorBase and use directly inheritance.
```C++
struct BsplineSet : public SPOSet

template<typename ST>
struct SplineC2R : public BsplineSet
```

A lot of classes and files are renamed to be more understandable. SoA and adoptors in the names are all removed.
Rename list:
HybridAdoptorBase.h => HybridRepCenterOrbitals.h
HybridCplxAdoptor.h => HybridRepCplx.h
HybridRealAdoptor.h => HybridRepReal.h
SplineAdoptorBase.h => deleted
SplineAdoptorReaderP.h => SplineSetReader.h
SplineC2CAdoptor.h => SplineC2C.h
SplineC2RAdoptor.h => SplineC2R.h
SplineC2ROMP.h => SplineC2ROMP.h
SplineHybridAdoptorReaderP.h => HybridRepSetReader.h
SplineR2RAdoptor.h => SplineR2R.h


src/QMCWaveFunctions/BsplineFactory/BsplineSet.h
```diff
@@ -27,100 +28,98 @@
...
- * @todo Make SplineAdoptor be a member not the base class. This is needed
- * to make MultiBsplineSet (TBD) which has multiple SplineAdoptors for distributed
- * cases.
+/** BsplineSet is the base class for SplineC2C, SplineC2R, SplineR2R.
+ * Its derived template classes manage the storage and evaluation at given precision.
+ * BsplineSet also implements a few fallback routines in case optimized implementation is not necessary in the derived class.
  */
-template<typename SplineAdoptor>
-struct BsplineSet : public SPOSet, public SplineAdoptor
+struct BsplineSet : public SPOSet
```
PDoakORNL: This is a good chance to bring this in line with the coding standards:

> Do not use struct as a way to avoid controlling access to the class. Only in rare cases where a class is a fully public data structure struct is this appropriate. Ignore (or fix one) the many examples of this in QMCPACK.

src/QMCWaveFunctions/BsplineFactory/HybridRepCenterOrbitals.h
```diff
@@ -25,7 +25,7 @@
 namespace qmcplusplus
 {
 template<typename ST>
-struct AtomicOrbitalSoA
+struct AtomicOrbitals
```
PDoakORNL: Again not a `struct`

src/QMCWaveFunctions/BsplineFactory/HybridRepCplx.h
```diff
@@ -0,0 +1,218 @@
...
+  {
+    this->className = "Hybrid" + this->className;
+    this->KeyWord   = "Hybrid" + this->KeyWord;
+  }
+
+  virtual SPOSet* makeClone() const override { return new HybridRepCplx(*this); }
+
+  inline void resizeStorage(size_t n, size_t nvals)
+  {
+    SplineBase::resizeStorage(n, nvals);
```
PDoakORNL: I don't think splinebase and hybridbase should be inherited from. They make a diamond down to SPOSet which is in general a road to ruin. And conceptually they are elements of the mixed basis. 

One might expect the splinebase and hybridbase to be on somewhat equivalent footing but a splinebase has virtual functions for evaluation and then so you would expect of the hybrid base. However they just have similar/same function names which is certainly problematic enough, yikes.

src/QMCWaveFunctions/BsplineFactory/BsplineSet.h
ye-luo: Will update. Need to making builders friend classes. I wanted to move more code from this class to builders. So far these changes are not taking higher enough priority.

src/QMCWaveFunctions/BsplineFactory/HybridRepCplx.h
PDoakORNL: Ok I will look at this more.  I still see a class with a huge accessible namespace available to the methods. I don't see yet why data members needed by the owner can not be accessed from the contained object.

ye-luo: Could you explain a bit more? I don't quite understand what you mean in your last reply.

ye-luo:
There is actually no new files. Now I update all the classes from struct to class and put private, protected and public based my understanding so far. The template parameters are all CAPITAL now.

src/QMCWaveFunctions/BsplineFactory/BsplineSet.h
ye-luo: Done.

src/QMCWaveFunctions/BsplineFactory/HybridRepCenterOrbitals.h
ye-luo: Done.

src/QMCWaveFunctions/BsplineFactory/HybridRepCplx.h
PDoakORNL: I wish I had time to. We can discuss by skype after the Holidays. I don't think it should block the PR.

PDoakORNL:
Refactoring the class hierarchy further will have to wait for more discussion post NLPP and spline fall back.  The coding convention updates look good to me.

prckent:
OK, we'll merge once the branch update and CI completes.

PR #2161, opened 2019-12-18
Updated AFQMC Estimator Extraction

fdmalone:
Updated extraction + naive analysis routines for  new AFQMC estimators, including 1RDM, 2RDM and on-top pair density matrix. Added generic  extraction routine with specialised observable analysers.


markdewing:
okay to test

PR #2162, opened 2019-12-18
Enable back propagation with (potentially truncated) msd trial.

fdmalone:
Small update to allow MSD trials for back propagation.


markdewing:
okay to test

PR #2164, opened 2019-12-18
Add external potential

kryczko:
This is a request to be able to include a custom external potential that can be read in through the HDF5 format.


markdewing:
okay to test

src/QMCHamiltonians/GridExternalPotential.h
```diff
@@ -0,0 +1,77 @@
...
+#include <einspline/bspline.h>
+
+
+namespace qmcplusplus
+{
+struct GridExternalPotential : public QMCHamiltonianBase
+{
+  const ParticleSet& Ps;
+
+  UBspline_3d_d *spline_data;
```
ye-luo: I guess spline_data is shared by all the threads and the evaluation functions are thread safe. So let us use shared pointer instead of raw pointer here.

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
```diff
@@ -157,7 +157,7 @@ bool EinsplineSetBuilder::ReadOrbitalInfo_ESHDF()
       if (Super2Prim[i] < 0)
       {
         app_error() << "Supercell ion " << i << " not found in the primitive cell" << std::endl;
-        abort();
+        // abort();
```
ye-luo: checked in by mistake?

markdewing: This is a hack to make the external potential work with no ions.

kryczko: To be more specific, this is to avoid qmcpack from terminating if there are no ions read in from an HDF5 format file that holds SPOs from a Qbox calculation.

ye-luo: Then let us introduce an option to allow skipping the ion check.

tests/models/sho/vmc_sho.xml
```diff
@@ -34,7 +34,7 @@
     </wavefunction>
 
     <hamiltonian name="h0" type="generic" target="e">
-      <extpot type="HarmonicExt" mass="5.0" energy="0.3"/>
+      <extpot type="HarmonicExt" mass="1.0" energy="1.0"/>
```
ye-luo: The input file of a test is changed but there is no change of the testing results.
I guess it was for development and should be reverted.
A statistic test should not hard code random seed.

tests/models/sho/vmc_sho_hdf5/vmc_sho_hdf5.xml
```diff
@@ -0,0 +1,49 @@
+<simulation>
+  <project id="sho" series="0"/>
+  <random seed="13" />
```
ye-luo: Same here. A statistic test should not hard code random seed.

src/QMCHamiltonians/GridExternalPotential.cpp
```diff
@@ -0,0 +1,150 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2016 Jeongnim Kim and QMCPACK developers.
```
ye-luo: Please put `// Copyright (c) 2019 QMCPACK developers.`

src/QMCHamiltonians/GridExternalPotential.h
kryczko: How would a shared pointer be implemented here?

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
kryczko: What should the criteria for the skip be? I feel like if no ions are defined in the input file then this error should not be thrown.

tests/models/sho/vmc_sho.xml
kryczko: How should I proceed here?

tests/models/sho/vmc_sho_hdf5/vmc_sho_hdf5.xml
kryczko: I can remove this.

src/QMCHamiltonians/GridExternalPotential.cpp
kryczko: Will do.

src/QMCHamiltonians/GridExternalPotential.h
ye-luo: I mean using `std::shared_ptr<UBspline_3d_d> spline_data;`

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
ye-luo: I got confused. Does your run have no ions from input xml or no ions from h5 generated by qbox?
This check is inside a loop of `for (int i = 0; i < SourcePtcl->R.size(); i++)`. Since you hit the error check, your input xml file has ions. It seems that your h5 has no info about ions in the primitive cell, namely the qbox converter generated h5 is not fully compliant with the h5 generated by QE.
Is "IonPos.size()==0" in your case? IonPos records ion positions loaded from h5.
To workaround the check, you can change `if (Super2Prim.size() == 0` to
```C++
if (IonPos.size()==0)
{
  app_warning() << "Ion positions are missing in h5. Bypassing supercell to primitive cell electron position mapping check!" << std::endl;
}
else if (Super2Prim.size() == 0)
```
In this way, the mapping check is bypassed with a warning printed.

tests/models/sho/vmc_sho.xml
ye-luo: More of clarification needed here. Please explain
1. Why do you need to change parameters in an existing test? How wrong was the old value?
2. Why such changes doesn't impact the reference value?

kryczko: The external potential generated in Python for the HDF5 format file uses parameters mass = 1.0 and energy = 1.0. I can change the parameters in the Python script such that it matches the old testing parameters.

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
kryczko: there should be no ions in both cases.

src/QMCHamiltonians/GridExternalPotential.h
kryczko: I tried changing:
```
UBspline_3d_d *spline_data;
```
to 
```
std::shared_ptr<UBspline_3d_d> spline_data;
```
but now I get a compilation error:
```
/home/kryczko/git/qmcpack/src/QMCHamiltonians/GridExternalPotential.cpp:120:61: error: cannot convert ‘std::shared_ptr<UBspline_3d_d>’ to ‘UBspline_3d_d*’ for argument ‘1’ to ‘void eval_UBspline_3d_d(UBspline_3d_d*, double, double, double, double*)’
       eval_UBspline_3d_d(spline_data, r[0], r[1], r[2], &val);
```

ye-luo: `eval_UBspline_3d_d(spline_data.get(), r[0], r[1], r[2], &val);` will solve the problem.
https://en.cppreference.com/w/cpp/memory/shared_ptr

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
ye-luo: Then change line 127 to
`if (Super2Prim.size() == 0 && SourcePtcl->R.size() > 0)`

tests/models/sho/vmc_sho.xml
ye-luo: @Paul-St-Young since you contributed the test, could you advice which is the way to go?

src/QMCHamiltonians/GridExternalPotential.h
kryczko: Now I get the error:
```
/home/kryczko/git/qmcpack/src/QMCHamiltonians/GridExternalPotential.cpp: In member function ‘virtual bool qmcplusplus::GridExternalPotential::put(xmlNodePtr)’:
/home/kryczko/git/qmcpack/src/QMCHamiltonians/GridExternalPotential.cpp:87:61: error: no match for ‘operator=’ (operand types are ‘std::shared_ptr<UBspline_3d_d>’ and ‘UBspline_3d_d*’)
                                      BC, BC, BC, data.data());
```

ye-luo:
I fixed the compilation and merged current develop. However I cannot push to this PR branch probably because your fork is not a direct fork of QMCPACK main repo but Mark's fork.
Please merge
https://github.com/ye-luo/qmcpack/tree/add_external_potential
to this PR branch.

Assuming your are on your add_external_potential branch.
git remote add ye-fork https://github.com/ye-luo/qmcpack.git
git merge ye-fork/add_external_potential
git push

tests/models/sho/vmc_sho.xml
ye-luo: Your change in the input file so far breaks the existing test short-sho-vmc-1-16. This needs to be addressed.

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
ye-luo: I will need to have a look after you fix the test. Keep this change not commited when you are fixing the test.

ye-luo:
Please also move folder
tests/models/sho/vmc_sho_hdf5
to
tests/models/grid_external_potential
Leaving vmc_sho_hdf5 inside sho is quite confusing.
Although GridExternalPotential has some relation with HarmonicExternalPotential but they are independent features.

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
kryczko: i have changed the test to match the older parameters. I can confirm I get the same answer in both cases with the same parameters.

ye-luo:
I forgot to metion `git fetch ye-fork` before the merge command.
Just add me as a collaborator of your fork. Then I will be able to fix the rest issue.

ye-luo:
Giving CI a try now. I'm still working on it.

ye-luo:
1. I reverted all the the ion super cell to primitive cell mapping check change in this PR. It is an issue unrelated to this PR and current test doesn't reflect that needs. We need a separate PR to address that issue. So far an integration test using qbox WF is missing.
2. Please revise the reference error bar in tests/models/CMakeLists.txt for GEP. The one from SHO seems to small. The test has been connected and you can run with `ctest -R grid_ext_pot-vmc -j16`

Once point 2 is addressed, this PR can be merged.

kryczko:
I tried running the command given above in `qmcpack/tests/models` but I get the following error: 
```
Test project /home/kryczko/git/qmcpack/tests/models
No tests were found!!!
```

ye-luo:
> I tried running the command given above in `qmcpack/tests/models` but I get the following error:
> 
> ```
> Test project /home/kryczko/git/qmcpack/tests/models
> No tests were found!!!
> ```

Run that command in the build folder where you ran `make`. Manual section 3.9 describes how to use ctest.

kryczko:
I checked out the manual, but there is no description about how to set these values.  I have increased the size of the second number next to totenergy:
```
GEP_SCALARS "totenergy" "0.45   1e-1"
```
but still the test does not pass. I am confused because the analysis of qmca tells me that the localEnergy should lie within 0.45 ± 1e-1.

prckent:
I have been keenly watching the progress of this PR. What is needed to get it "unstuck" and merged in? We certainly need a test but can give help with that if needed. @kryczko are you able to configure and run tests now?

kryczko:
I am able to test, but I am getting a test failure with `totenergy`. I am not entirely sure what is going on for the test, but I am assuming we want the energy to be a certain value within fluctuation. If you look above I set the following, but still have a test failure. Do I need to recompile the code everytime I change a value in the test?

ye-luo:
@kryczko if you touch the reference value in the test file in the repo, you need to do `cmake ..` in your build folder to update the test file used by ctest.

ye-luo:
Is 1e-1 too loose?

kryczko:
oh yes it is way too loose, but I was just illustrating that I couldn't make it pass with something large. I didn't run cmake after changing the value though, so once I do that it should be good to go.

ye-luo:
Once you update the PR with appropriate reference numbers, this PR is complete.

tests/models/CMakeLists.txt
```diff
@@ -9,6 +9,16 @@ ENDIF()
   LIST(APPEND SHO_SCALARS "potential" "0.225 0.0028")
   LIST(APPEND SHO_SCALARS "latdev"    "1.0   0.013")
 
+# The errors on the total energy are numerical noise, not statistical errors
+IF(QMC_MIXED_PRECISION)
+  LIST(APPEND GEP_SCALARS "totenergy" "0.45  0.00000004")
+ELSE()
+  LIST(APPEND GEP_SCALARS "totenergy" "0.45  4.0e-14")
+ENDIF()
```
ye-luo: @kryczko 
The `IF(QMC_MIXED_PRECISION)` seems unnecessary. I copy pasted from SHO test.
Just generate a reasonable error bar for mixed precision build and keep one line for both cases.

kryczko:
I've made the tolerance 1e-6 and all tests have passed. All good on my end.

prckent:
Many thanks for this contribution @kryczko

kryczko:
thanks for all of the help getting this PR in! :-)

kryczko:
Hi @ye-luo, I just tried running one of my experiments and I still hit the error:
```
ERROR Supercell ion 0 not found in the primitive cell
```
If you want I can open another issue regarding this.

kryczko:
It seems like SourcePtcl->R.size() yields the number of electrons rather than the number of nuclei if that helps.

ye-luo:
I commented above
> 1. I reverted all the the ion super cell to primitive cell mapping check change in this PR. It is an issue unrelated to this PR and current test doesn't reflect that needs. We need a separate PR to address that issue. So far an integration test using qbox WF is missing.

kryczko:
Okay, i will start a new pull request.

ye-luo:
Please open an issue with a reproducer. Then we can assess what is the right fix instead of a duck tape.

PR #2165, opened 2019-12-18
Update afqmctools converters

fdmalone:
1. Adds converters between kpoint,sparse and dense hamiltonian file formats (when it makes sense).
2. Adds dense writing of Hamiltonian for chemistry problems.


markdewing:
okay to test

PR #2166, opened 2019-12-18
Allow pyscf density fitting to be used in place of Cholesky 

fdmalone:
If df integrals are present in chkfile the converter will use these instead of generating Cholesky.


markdewing:
okay to test

PR #2167, opened 2019-12-19
Add Bspline converter for pyscf and pyscf workflow test for 1x1x1 diamond

Hyeondeok-Shin:
Continue from #2126. 

 Add LCAO -> bspline converter for pyscf and pyscf workflow test for both LCAO and Bspline. They uses the same references.
 The workflow test will be automatically added when related modules (pyscf, QmcpackToPyscf, etc.) exist, and be able to check in the converter test.

 Hyeondeok Shin


markdewing:
okay to test

prckent:
Is this known to work with Python3?

Hyeondeok-Shin:
@prckent This test is specified for Python3, so works with Python3. In Python2, this test won't be added automatically.

prckent:
Q. How long do the diamondC_1x1x1_pp-vmc_gaussian_sdj-1-16 tests take to run ?

Hyeondeok-Shin:
@prckent With LCAO, it takes 148 sec in Intel Xeon while 5 sec is needed for b-spline.

Issue #2158, opened 2019-12-17
Ewald check is slow 


prckent:
Formalizing the results of some conversations:
1. It is worth to add a timer around the Ewald check code. Larger cells appear to be taking significantly longer on cdash.
1. Some simple OpenMP and MPI would be worthwhile. The former would most help with our tests, but for general calculations both are needed.

prckent:
Following #2169, MPI will be needed + timing printout in subsequent improvements.

prckent:
Still slow post OpenMP in #2169 , so we should consider possible efficiency improvements as well as MPI. 128 atoms takes O(2500) seconds with 16 threads.
<img width="868" alt="Screen Shot 2019-12-22 at 10 13 14 PM" src="https://user-images.githubusercontent.com/13683814/71333522-47a23400-2508-11ea-9bcc-abe2aba6763d.png">

prckent:
64 atoms are noticeably faster:
<img width="868" alt="Screen Shot 2019-12-22 at 10 17 33 PM" src="https://user-images.githubusercontent.com/13683814/71333689-f777a180-2508-11ea-85f8-5690a0cfd6bc.png">

prckent:
The Ewald check is now fast enough. Thanks @jtkrogel and @ye-luo.

**64 atoms** 
<img width="798" alt="Screen Shot 2020-01-03 at 10 21 41 AM" src="https://user-images.githubusercontent.com/13683814/71731493-dff6bd80-2e12-11ea-9573-762a6cadb399.png">
**128 atoms**
<img width="794" alt="Screen Shot 2020-01-03 at 10 28 40 AM" src="https://user-images.githubusercontent.com/13683814/71731879-d457c680-2e13-11ea-9563-fdb9c4024fcc.png">

PR #2169, opened 2019-12-20
Speed up reference Ewald sums with OpenMP

markdewing:
First attempt to address #2158

Also add a timer.   Some of the investigation was done with additional timers, and most of the time is spent in the k-space sum.

Convert the triangular loop to a square loop so the OMP collapse clause will work.

If the triangular loop is left in place, and the OMP is placed only on the outer loop, load balancing becomes an issue - the final iteration will be the most time consuming.

This has only been lightly tested.


markdewing:
I'm getting 4.5x speedup on 6 threads for the NiO a16-e192 problem.    Larger problems are taking too long to get timings for the serial code.

prckent:
Perhaps greater OpenMP speedups will be possible with a flattened loop and explicit "triangle indexing" to avoid the need for the dynamic schedule.

prckent:
@ye-luo OK. Earlier I had the following but did not check timings:

```
 real_t ve2 = 0.0;
 #pragma omp parallel for reduction(+:ve2)
 for(size_t k=0; k<N*(N-1)/2; k++)
   {
      size_t i = k/N, j = k%N;
      if(j<=i) i = N-i-2, j=N-j-1;
        {
          real_t qq = Q[i]*Q[j];
          ve2 += qq*ewaldSum(R[i]-R[j],a,tol/qq);
        }
   }
```

prckent:
@ye-luo OK

PR #2170, opened 2019-12-21
Fix parallel converter testing on Bora

ye-luo:
Two issues identified.
1. N_CONCURRENT_TESTS was not propagated properly when set as an environment variable.
2. The result check tests in the QE workflow tests are not tagged as converter tests.


prckent:
Woth noting that this change also wildcards all qe- tests as converter tests. We could do the same for other codes in case of similarly missed labels.

PR #2171, opened 2019-12-22
Workaround intel compiler bug

ye-luo:
The old workaround for "internal compiler error" cannot be further expanded and I put a new cleaner workaround here.
The order of the two lines of dot1 does matter with intel >=19 compilers.
Since intel is switching to clang based compiler, I doubt effort any will be put in the old compiler.


prckent:
Which compilers did you verify this change with? Given the history of ICE in the Intel Compilers please check with Intel 2018, 2019.

ye-luo:
Tested 2018, 2019, 2020 beta and oneAPI beta. 2020 is out but I don’t have it installed. 2020 beta is representative.

PR #2172, opened 2019-12-24
Optimize EwaldRef

ye-luo:
Further improve the situation of #2158, 2-4x speed up although it is still slow.
Changes are
1. Rearrange loops to reuse results.
2. Improve vectorization.

Now with 128 atom NiO, EwaldRef takes 297 sec on 16 core Xeon which should not be too bad for production runs.
If MPI is really needed, we can replace the current threading level with MPI and put threading in the inner levels. This will not help single MPI runs which most our tests do.
If we add MPI, the check code must be moved to the builder instead of the Coulomb class.

### DIFF
--- a/src/QMCHamiltonians/CoulombPBCAA.cpp
+++ b/src/QMCHamiltonians/CoulombPBCAA.cpp
@@ -52,9 +52,9 @@ CoulombPBCAA::CoulombPBCAA(ParticleSet& ref, bool active, bool computeForces)
   {
     update_source(ref);
 
-    ewardref::RealMat A;
-    ewardref::PosArray R;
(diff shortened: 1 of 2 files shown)


prckent:
Thanks. Please can you add a few comments on how the tiling, worklist, and threading is done? Someone modifying this code in future should not have to reverse engineer the algorithm.

prckent:
I think 5 minutes for 128 atons is OK for the 3.9 release, but long term we will want MPI parallelization as well. E.g. Summit has about 7 threads/GPU and MPI task.

ye-luo:
@prckent fixed namespace typo and add comments about the algorithm and potential changes in the future.

PR #2173, opened 2019-12-24
Included Rhea in machine.py (for Nexus)

zenandrea:
Included Rhea into the machines that Nexus can use


markdewing:
okay to test

prckent:
Thanks @zenandrea . I expect that this will be merged without trouble after the holidays. If you need it sooner I could merge it, but Jaron should really take a look.

PR #2178, opened 2020-01-01
Remove Blue Gene Q support

ye-luo:
ALCF BGQ is officially retired.
Delete all lines of source code and configurations related to BGQ.


ye-luo:
@prckent a bit more deleting is done.

PR #2179, opened 2020-01-01
Fix NLPP derivative in SoA

### LINKED ISSUE #2083: NLPP derivatives in optimization need SoA

ye-luo:
Needs #2177 merged in first.
Close  #2083.

I mostly relied on unit tests. The full simulation case I tried didn't make a significant difference using or not using NLPP derivatives.
@jtkrogel Could you check your Li2 case?


tiihonej:
After running similar I can reproduce the behavior in #2083, only this time I can use both SoA and AoS builds.

Li2_soa_nlpp_derivs
Li2_soa_nlpp_noderivs

So I would consider the Li2 test passed and #2083 fixed by this PR. Thank you! Also, PR #2128 or something similar is needed to manage nonlocalpp_derivs with nexus, although it's good to have them now enabled by default.

jtkrogel:
Thanks @tiihonej.  @ye-luo It's nice to have this fix.  This is the last update needed prior to AoS removal.

PR #2180, opened 2020-01-02
Make delayed update active only by request

ye-luo:
Needs #2177 merged in first.

The currently code activates delayed update on every acceptMove which is safe during p-by-p or T-move but algorithms making a move and a reverse move are not safe to use DU. developers are required to put a completeUpdates() after every acceptMove. This can cause potential bugs for new developers.
So I introduced a flag safe_to_delay in TWF and WFC acceptMove and default to false. So drivers known safe for delayed update have safe_to_delay = true requested, other places (NLPP forces and derivatives) where the safety is unknown, default SM-1 update will be called at every acceptMove to ensure the inverse matrix is safe to use.


prckent:
For the record: What is the assumption of the delayed update that is being violated?

(Presumably it currently assumes in-sequence electron moves and has no capability for rewind?)

ye-luo:
Delayed update doesn't require in-sequence electron moves. It can be arbitrary order.
However, there are two violations
1. Delayed update cannot delay moves of the same electron more than once.
2. The delayed update delays the updates of inverse Slater matrix. For forces and NLPP derivatives, the major reason to actually accept move is to use the inverse matrix. So the completeUpdate must be forced.

PR #2183, opened 2020-01-04
implemented rediag for nomsd

fdmalone:
Allows for rediagonalizing CI hamiltonian with non-orthogonal determinants. Useful for when codes generating expansions follow different conventions to those followed internally.


ye-luo:
Okay to test

PR #2184, opened 2020-01-06
Walker batching in NLPP step 1

ye-luo:
This PR implements walker batching in NonLocalECPComponent and NonLocalECPotential.
So far it relies on mw_evaluateRatios fallback and the specialization on the wavefunction side will be implemented as step 2.

1. short-diamondC_2x1x1_pp-vmcbatch-4b_sdj has been added to check the new code path.
2. some adjustments in VirtualParticleSet::makeMoves
3. Put const VirtualParticleSet in evaluateRatios. All the ParticleSet used by WFC should be const, it is easier to making the change after deleting AoS
4. Some changes in NonLocalECPComponent maybe break tracing. The trace deterministic tests seem passing.
5. I tried to use RefVector for newly added codes but I noticed that the transition from pointer to RefVector in WFC was paused and @PDoakORNL we need to continue the work.


src/QMCHamiltonians/NonLocalECPotential.cpp
```diff
@@ -306,6 +261,190 @@ void NonLocalECPotential::evaluate(ParticleSet& P, bool Tmove)
...
+    {
+      NonLocalECPotential& O(static_cast<NonLocalECPotential&>(O_list[iw].get()));
+      ParticleSet& P(P_list[iw]);
+      if (jobid < O.nlpp_jobs[ig].size())
+      {
+        const auto& job = O.nlpp_jobs[ig][jobid];
+        ecp_potential_list.push_back(O);
+        ecp_component_list.push_back(*O.PP[std::get<0>(job)]);
+        p_list.push_back(P);
+        iat_list.push_back(std::get<0>(job));
```
PDoakORNL: See comment in header.  If this isn't a Tuple but a struct this becomes less cryptic and error prone. Also looks like something in the debugger.

src/QMCHamiltonians/NonLocalECPotential.h
```diff
@@ -22,54 +22,56 @@
...
 public:
-  NonLocalECPotential(ParticleSet& ions,
-                      ParticleSet& els,
-                      TrialWaveFunction& psi,
-                      bool computeForces = false,
-                      bool useVP         = false);
+  /** tuple type for NLPP calculation of a pair of ion and electron
+   * ion id, electron id, electron position, ion electron distance, ion to electron displacement
+   */
+  using NLPPJob = std::tuple<int, int, PosType, RealType, PosType>;
```
PDoakORNL: I'd favor making this a structure.  
```C++
struct NLPPJob
{
  int ion_id;
  int electron_id;
  PosType elec_pos;
  RealType ion_elec_dist;
  PosType ion_elec_disp;
}
```

ye-luo: Good suggestion. I regretted after the tuple was coded. Will submit a separate PR to address this.

PR #2187, opened 2020-01-06
Replace python2 iteritems by items

prckent:
process_perf.py was failing due to legacy python 2 iteritems.

This PR allows the performance tests to work and report to cdash. I have verified that all the deterministic tests work and do not give python errors.


markdewing:
I don't think check_stats.py should be changed in this PR.    It has additional issues with Python 3 compatibility ('import cPickle').    The calls to '_iteritems' are not the builtins (though they may eventually map to them)

prckent:
Restored check_stats.py. I would assume that all these files have additional Python 3 trouble.

PR #2190, opened 2020-01-07
Add XML selection of LR Handler

rcclay:
First cut of this.  This PR adds the XML tag "LR_handler" to the lattice description and to the wave function specification when parsing ion sets and lattices from an ESHDF5 file.  

This will break with XML files which follow Juha's declaration of LR handler type in the Chiesa force estimator...  but hopefully this is an overall simplification.  

Example of a lattice declaration:  
`<simulationcell>
         <parameter name="lattice" units="bohr">
                  3.77945227        0.00000000        0.00000000
                 -0.00000000        3.77945227        0.00000000
                 -0.00000000       -0.00000000        3.77945227
         </parameter>
         <parameter name="bconds">
            p p p
         </parameter>
         <parameter name="LR_dim_cutoff"       >    15                 </parameter>
         <parameter name="LR_handler"       >    ewald                 </parameter>
      </simulationcell>
`
Supported types for LR_handler are:
1.  "esler" for current optimized breakup behavior.  The code will default to this if the option is not specified.
2. "ewald" for the Ewald3D handler
3. "natoli" for the original Natoli/Ceperley optimized breakup.

### DIFF
--- a/src/LongRange/LRCoulombSingleton.cpp
+++ b/src/LongRange/LRCoulombSingleton.cpp
@@ -30,6 +30,7 @@ namespace qmcplusplus
 //initialization of the static data
 LRCoulombSingleton::LRHandlerType* LRCoulombSingleton::CoulombHandler      = 0;
 LRCoulombSingleton::LRHandlerType* LRCoulombSingleton::CoulombDerivHandler = 0;
+LRCoulombSingleton::lr_type LRCoulombSingleton::this_lr_type = ESLER;
 /** CoulombFunctor
  *
  * An example for a Func for LRHandlerTemp. Four member functions have to be provided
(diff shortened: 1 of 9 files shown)


prckent:
Some thoughts on names only to begin with:

1. We should choose names and input parameters carefully to minimize likelihood of future updates.
1. "LR" is not a very clear abbreviation, so perhaps we should take this opportunity to change it? 
1. I am uncomfortable naming the methods after people since we don't have their blessing and can't be sure how close what we have in the code now is to what they originally intended.
1. For the record, what is the difference between "esler" and "natoli"? Is there a publication for the former?
1. How about opt_breakup, opt_breakup_esler, and ewald ?
1. Note that should #2182 be successful, I am keen that we delete **all** of these old implementations.

rcclay:
@prckent To answer some of your questions:
2. I can change that to "longrange_handler" or something.
3. I'm fine changing it to opt_breakup, opt_breakup_natoli, and ewald.  I originally named the options after people by pulling the first author off the paper or thesis the method appeared in.
4. The original Natoli opt breakup scheme "splines" the short-range piece of the potential and optimizes the spline coefficients to minimize the residual error.  Ken Esler's method splines the long-range piece instead and optimizes those coefficients. One can site his thesis or the qmcpack manual, section 26.2.8 :P.
5. see 3
6. Simplification is very much warranted.  Unless we're doing optimized breakups for other functional forms routinely (long range jastrows and what not), this level of generality is not worth the added code complexity IMHO.

prckent:
2. If was wondering if "coulomb_handler" or "coulomb_method" would be better... however I can live with LR_method for now. If the anisotropic code is proves sufficient all this will be moot.
3. opt_breakup, opt_breakup_original, and ewald ?

It appears that the CUDA build is broken - you should have an email.

Issue #2186, opened 2020-01-06
Nexus unit tests should check and print stderr from command execution


markdewing:
While running the Nexus unit tests, the 'qmc_fit' test failed (nexus/tests/unit/test_qmc_fit.py)  The issue was the qmc_fit executable was not found, because 'nexus/bin' was not on the PATH.

The issue would be faster to diagnose from test output if stderr from 'execute' calls is checked and output if not empty.    (The return code isn't checked, either, but the stderr output is more useful.)

prckent:
Should we rely on PATH being set? For testing and installation test_qmc_fit could use the relative or absolute path to nexus/bin . e.g. The QMCPACK tests don't require the qmcpack binary on the PATH.

jtkrogel:
`nexus/bin` should not need to be on PATH for the tests to work.  Do you get this behavior from `ctest` or from `nxs-test` directly?

markdewing:
It happened from both ctest and nxs-test.

markdewing:
I don't have '.' on the PATH, either.

PDoakORNL:
no one should have '.' in the PATH

PR #2192, opened 2020-01-07
Nexus: fix test path issues

jtkrogel:
This PR addresses #2186.  All Nexus executables now successfully run without Nexus being installed.  

All tests except the user examples pass when both `PATH` and `PYTHONPATH` are empty.  The user example tests pass provided Nexus is installed  (`PYTHONPATH=/your/path/to/nexus/lib`).  

Additionally, `nxs-test` can be used in query mode (e.g. `nxs-test --list` or `nxs-test --ctestlist`) without having a NumPy installation, as is needed to supply correct labels to `ctest` during QMCPACK configuration.

A small fix was applied to the QMCPACK reference optimization inputs to account for the updated defaults in #2128.

### DIFF
--- a/nexus/bin/nxs-sim
+++ b/nexus/bin/nxs-sim
@@ -2,7 +2,36 @@
 
 import os
 from optparse import OptionParser
-from generic import obj
+
+
+def find_nexus_modules():
+    import sys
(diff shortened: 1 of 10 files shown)


markdewing:
okay to test

PR #2194, opened 2020-01-07
Walker batching in NLPP step 2

ye-luo:
Review and merge after #2193.
Add the offload implementation SplineC2ROMP::mw_evaluateDetRatios
I also cleanup the unsafe use of pointers in SplineC2ROMP.

Performance numbers are recorded at the end of
https://github.com/ye-luo/qmcpack/wiki/Performance-analysis-on-small-problem-sizes

In this PR, BsplineAllocator taking three allocator template argument is not good. I only intended to make it working. I'm looking at ways to clean it up but more refactor will be needed.


prckent:
Terrific speedup consistent with the earlier analysis on that page. The speed is ~doubled compared to previous :-)

@ye-luo Are the timings https://github.com/ye-luo/qmcpack/wiki/Performance-analysis-on-small-problem-sizes#2020-01-07-update-nlpp-walker-batching-completed-and-mass-connected-both-bottleneck-1-and-2-are-resolved for Summit with 384 electrons? Please add comment to indicate. Also, how was the code compiled for the above data?

src/QMCWaveFunctions/BsplineFactory/SplineC2ROMP.cpp
```diff
@@ -594,56 +715,38 @@ void SplineC2ROMP<ST>::evaluateVGL(const ParticleSet& P,
...
+  const size_t num_pos = psi_v_list.size();
   const int ChunkSizePerTeam = 128;
   const int NumTeams         = (myV.size() + ChunkSizePerTeam - 1) / ChunkSizePerTeam;
   const auto padded_size     = myV.size();
-  if (offload_scratch.size() < padded_size * nwalkers * 10)
-    offload_scratch.resize(padded_size * nwalkers * 10);
-  const auto orb_size = psi_v_list[0]->size();
-  if (results_scratch.size() < orb_size * nwalkers * 5)
-    results_scratch.resize(orb_size * nwalkers * 5);
+  if (offload_scratch.size() < padded_size * num_pos * 10)
```
prckent: Where do these magic numbers come from?

ye-luo: Done. VGL needs 5, VGH needs 10.

ye-luo:
> Terrific speedup consistent with the earlier analysis on that page. The speed is ~doubled compared to previous :-)
> 
> @ye-luo Are the timings https://github.com/ye-luo/qmcpack/wiki/Performance-analysis-on-small-problem-sizes#2020-01-07-update-nlpp-walker-batching-completed-and-mass-connected-both-bottleneck-1-and-2-are-resolved for Summit with 384 electrons? Please add comment to indicate. Also, how was the code compiled for the above data? Same as the updated recipe (IBM XL + mass only)?

Update the wiki page indicating this is exactly the same simulation as the old case 2). The IBM XL recipe in the repo should be up-to-date. I made a change in an early PR using a compiler option specific to fixing XL C++14 warning. This also allows me to enable IBM MASS.

Issue #2188, opened 2020-01-06
Residual Python 2 in QMCPACK


prckent:
Although NEXUS was updated and the QMCPACK cmake and test configuration appears to work, there is residual python2 that needs to be checked for and updated. 

* Use of print as a statement print needs to be updated to function() usage. e.g. The skinetic_check.py fails due to this ( ctest -R skinetic -VV ).

There must be more updates beside print usage and #2187 - each script needs to be invoked and checked.

ye-luo:
Ubuntu 19.10 is totally python2 free but I don't have access to it.
On my local machine, I create a fake bash script named python and put it first on my PATH to detect python2 invocation.

markdewing:
Running with a python 3 on the path first should be sufficient to catch problems.

markdewing:
Unless someone is already working on them, I can work on scripts/check_deriv.py and scripts/check_grad_lap.py  which are used with the 'bccH_2x2x2_ae' test.

prckent:
Note that much of the current breakage on https://cdash.qmcpack.org from oxygen is NOT due to python 3 but due to configuration issues from the updated software stack (compilers, libraries, cmake etc.) built via spack as part of the python 3 updates. e.g. There is a problem with HDF5 in the nightlies.

prckent:
Can we update the checklist at the top? What is not run/updated/checked yet?

Other than PySCF checks it looks like everything is running on oxygen with Python3 now. Can everything be checked off?

ye-luo:
@markdewing is the coverage also OK?

jtkrogel:
Added `labs` to the checklist.

markdewing:
Coverage was updated in #2157 
Converters were updated in #2139

ye-luo:
@jtkrogel completed by #2270 ?

jtkrogel:
#2270 covers the set of labs we have given in workshops (1-4).  The excited states lab still needs review, and there may be some issues separate from Python 2/3 considerations.

PR #2198, opened 2020-01-07
Update check_stats.py for python 3

markdewing:
The division operator needs updating as the single slash returns a floating point number by default in python 3.  The double slash is needed for integer division.

The 'range' builtin now returns a range object and needs to be converted to a list to recover the python 2 behavior.   The 'xrange' builtin no longer exists, as the changes to 'range' make it unnecessary.

Partially addresses #2188.

This script should still work with python 2

### DIFF
--- a/tests/scripts/check_stats.py
+++ b/tests/scripts/check_stats.py
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
 
 from __future__ import print_function, division
 
--- a/tests/scripts/check_stats.py
+++ b/tests/scripts/check_stats.py
@@ -1,6 +1,6 @@
 #! /usr/bin/env python
 
-from __future__ import print_function
+from __future__ import print_function, division
 
 # Statical error checking code for use by testing framework (stat.h5 files)
 # Jaron Krogel/ORNL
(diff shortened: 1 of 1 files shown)


tests/scripts/check_stats.py
```diff
@@ -1,6 +1,6 @@
 #! /usr/bin/env python
```
ye-luo: On Ubuntu 19.10, there is no more python 2.7 installation by default installation. Do you think we should update all the scripts to force python3. Nexus has done this. On systems with both python 2.7 and python3 installed. python is pointing python 2.7.

markdewing: Official guidance here: https://www.python.org/dev/peps/pep-0394/

I would prefer to leave it as 'python' because that is the most generic, and these scripts still work with python 2.   However, runtime distributions have the option of not providing a generic 'python' command at all, which then doesn't work.

I think we should leave it as 'python' for now, and see how many people run into problems, and how many environments do not provide a generic 'python'.

prckent: We'll make python3 a hard requirement (many reasons). This PR need not change.

prckent: Reconsidering: it will save some time to update the shebang here. I'll make the one character edit.

PR #2199, opened 2020-01-08
Disabling OpenMP nested threading inside TasksOneToOneOpenmp sections

PDoakORNL:
At least with llvm's openmp on power9 nested threading results in
terrible performance.  This is not surprising since the maximum number
of core locked threads should already be in use for the crowds
so the nested threading can only cause contention.

### DIFF
--- a/src/Concurrency/TasksOneToOneOPENMP.hpp
+++ b/src/Concurrency/TasksOneToOneOPENMP.hpp
@@ -41,6 +41,9 @@ void TasksOneToOne<Threading::OPENMP>::operator()(F&& f, Args&&... args)
       throw std::runtime_error(nesting_error);
   int nested_throw_count = 0;
   int throw_count = 0;
+  // Nor should nested threading be used when a task block is active
+  int outside_nested = omp_get_nested();
+  omp_set_nested(0);
 #pragma omp parallel num_threads(num_threads_) reduction(+ : nested_throw_count, throw_count)
   {
     try
@@ -62,6 +65,7 @@ void TasksOneToOne<Threading::OPENMP>::operator()(F&& f, Args&&... args)
       ++throw_count;
     }
   }
+  omp_set_nested(outside_nested);
   if (throw_count > 0)
     throw std::runtime_error("Unexpected exception thrown in threaded section");
   else if (nested_throw_count > 0)


prckent:
Which LLVM? Open source version 9, the development version, or the installed IBM XL on Summit?

PDoakORNL:
I imagine this is a problem everywhere but this is with LLVM9 in the batched drivers. 

If you have x cores and x threads that are written to not block during the region (Which is my intention with TaskOneToOnes) then nested threading will always be harmful.  We can of course define an abstraction that allows a controlled amount of nested threading within a region, but right now it is uncontrolled except by turning it off and on with the env variable. 

When in a section concurrent on crowds these loops frequently end up being over the number of walkers in a crowd.  There is at least in LLVM9 using the batched drivers a big performance hit. I think I see openmp contexts setup and torn down to do 1 walkers worth of work, while crowd threads appear to get stopped to run these "optimized" nested loops.

There is a great fondness for just tossing directives around every loop in the code and assuming that will be useful, I was deleting them but since many are useless but this is a better solution for now. The legacy driver path does extract some advantage from these.

ye-luo:
I don't think this change is necessary. How to use properly use nested OpenMP are mentioned in the manual "4.5.1 Nested OpenMP threads". I verified that the current batched driver works with nested OpenMP on KNL, I can run 16 crowds and 4 walkers in each crowd with OMP_NUM_THREADS=16,4 on 64 cores. The crowds are distributed on the first level and walkers within each crowd are distributed in each second level thread team.

By default, nested OpenMP is not enabled. If you nest your abstraction layer, the second level should just get 1 thread. there is no need to explicitly turn on/off nested OpenMP.
If you enable nested openmp but only give one number to OMP_NUM_THREADS like
`OMP_NESTED=TRUE OMP_NUM_THREADS=16" The second level will enable 16 threads as well and oversubscription happens. This is a misuse of nested OpenMP.

ye-luo:
On the CPU OpenMP, intel and llvm both have good efficiency.

ye-luo:
I think `#pragma omp parallel num_threads(num_threads_)` is causing the problem because you explicitly request overriding the default behaviour of run-time.
When this abstraction was added we understood this limitation of TasksOneToOne abstraction layer and put a safeguard in line 40. How could your test bypass this check.

ye-luo:
@PDoakORNL could you give more information what was the code and how you where controlling OpenMP which caused over-subscription issue?

The current implementation in QMCPACK is like this 
```
TasksOneToOne<Threading::OPENMP>
loop over crowds
{
   if( multiple walkers)
   {
      #pragma omp parallel for
      loop over walkers
   }
   else (1 walker only ) 
   {
     function with 1 level of threading inside to handle extreme large problems.
   }
}
```

By default, no nested threading, only the outer level threading is enabled. The inner level has 1 thread and there is no penalty.
If nested threading is requested and there are multiple walkers in each crowd, both first and second level should work.
For very large systems, users can add 1 walker per crowd and activate the threading inside a walker.
Assuming properly controlling affinity, all the cases from small to large problems are covered performance should be decent.

ye-luo:
We can discuss this after the ECP call.

PDoakORNL:
It seems like I probably didn't have all the OMP environment variables set so the runtime would behave properly. However this does prevent behavior that this abstraction does not cover.

I'd like to see us actually express our concurrency strategies in code and have actually reasoned about them. If we want to support a strategy of one crowd and multiple threads that should be expressed in the code so data threadscope and synchronization can be explicitly understood.  I think QMCPACK is too complex a code to leave this to the interaction of pragma's and environment variables.

This doesn't mean openmp can't be used, it just needs to be abstracted so scopes are clear and important behavior is legible. After a few months debugging concurrency and data state issues in QMCPACK I think this benefit would be just as important as being able to use other threading and tasking methods.

src/Concurrency/TasksOneToOneOPENMP.hpp
```diff
@@ -41,6 +41,9 @@ void TasksOneToOne<Threading::OPENMP>::operator()(F&& f, Args&&... args)
       throw std::runtime_error(nesting_error);
   int nested_throw_count = 0;
   int throw_count = 0;
+  // Nor should nested threading be used when a task block is active
+  int outside_nested = omp_get_nested();
```
ye-luo: You need to be very careful with this change because it also affects also the first level.
Based on https://www.openmp.org/spec-html/5.0/openmpsu120.html about omp_get_nested,
this routine returns true if max-active-levels-var is greater than 1 for the current task.
If I remember correctly, quite a few widely used OpenMP runtimes default max-active-levels-var to some number like 2^32-1 if OMP_MAX_ACTIVE_LEVELS environment doesn't exist. Namely, what you add here will not take any effect.
If the default of max-active-levels-var is 1, then what you changed will turn off all the threading even the first (crowd) level. In addition omp_get_nested is deprecated in 5.0. Its use must be restricted.

ye-luo:
I think there is always a balance between full control and a simple code. As a C++ programmer, I would prefer fully precisely control threading environment. The tasks in the abstraction layer need to keep a state about its threading hierarchy. Then it is basically implementing something similar to OpenMP runtime and then instruct OpenMP runtime to behave exactly what our runtime requests. On the other hand, my scientist role persuades me to keep minimal threading impact on the source code, the runtime implementation should be done in the hands of CS experts. So far, the current threading arrangement works well and my scientist mind wins.

I'd like to check what was actually happening in your run and figure out what is the actual problem.
Discussing OpenMP vs other threading/tasking model is not really our needs. What is needed by science gives us priority.

If on CPU the current batched driver runs the same speed with
1. Ncore MPI and 1 thread per MPI.
2. 1 MPI and Ncore threads.
3. 1 MPI and Ncore threads on the first level and 1 thread
is this sufficient to say the current nested OpenMP usage is not penalize the performance?

PR #2200, opened 2020-01-08
Update CMakeLists and all shebangs to require Python3

prckent:
Search and replace. Does not imply that I checked the scripts actually work with python 3. Notice lots of example scripts are touched.


prckent:
check_stats.py is updated in #2198

prckent:
ALCF python3 does not have numpy
```
185/236 Test #566: deterministic-diamondC_1x1x1_pp-vmc-dmc-trace_scalars-1-1-check ............***Failed    0.09 sec

  File "/sandbox/jenkins-ci/workspace/QMCPACK-CI-GNU6-MKL18/label/skylake-only/type/Complex/tests/scripts/check_traces.py", line 19, in <module>

    import numpy as np

ImportError: No module named 'numpy'



```

ye-luo:
FindPython3 requires cmake 3.12 released on 2018-06-14 which is shorter than 2 years.
I personally prefer just bump up cmake version. I never difficulty to build cmake on any machine including old machines like BGQ in the past.

jtkrogel:
Thumbs down on adding a requirement to build new CMake on systems older than 18 months.

Trying to run `python3` and `python` on a short Python 3 script should suffice.

ye-luo:
Changed my mind. Add a cmake version check. If <3.12, run a simple python script.

prckent:
Good catch with FindPython3. Since it is very easy to check with our own code, I am in favor of the new solution and maintaining the two year rule.

ye-luo:
Test this please

ye-luo:
The last failure seems related to the random unit test failure with the cuda build. Try CI again.

ye-luo:
Test this please

prckent:
We should keep an eye on cdash for flaky CUDA build results. e.g. https://cdash.qmcpack.org/CDash/testDetails.php?test=7129211&build=99071 has deterministic-unit_test_particle_base failing.

prckent:
Second time lucky.

PR #2202, opened 2020-01-08
Python3 requirement in README.md

prckent:


ye-luo:
I moved this commit to #2200 to reduce a bit the CI pressure. #2200 now abort at cmake if python3 is not found which is consistent with this readme change.

PR #2203, opened 2020-01-08
CMake update supress expensive system tests support real mpi testing.

PDoakORNL:
I found the edit-compile-debug loop far too long with the current RUN_QMC_APP tests, this adds the ability to noop those expensive and dependency unaware custom commands.

Additionally it allows one to setup mpi unit tests that actually use multiple ranks.


CMake/macros.cmake
```diff
@@ -141,6 +141,26 @@ FUNCTION( RUN_QMC_APP TESTNAME SRC_DIR PROCS THREADS TEST_ADDED TEST_LABELS ${AR
...
+# far exceed even the feature and compiler checks for time consumed when rerunning cmake.
+#
+# On leconte with rerun of cmake with
+# slow tests 27 secs to run cmake
+# with them disable 8 secs
+
+SET( QMC_DISABLE_SLOW_TESTS FALSE CACHE BOOL "Disable the slow cmake system test additions")
+
+IF (QMC_DISABLE_SLOW_TESTS)
+  FUNCTION(QMC_RUN_AND_CHECK BASE_NAME BASE_DIR PREFIX INPUT_FILE PROCS THREADS SHOULD_SUCCEED)
```
markdewing: I think the arguments can be removed from these empty functions.  That avoids inconsistencies in the argument lists if the functions are changed in the future (or avoids the need to update two places).
CMake requires functions to be called with the same number or more arguments, so the calls will work fine.

PDoakORNL: Nice trick, done.

CMake/macros.cmake
```diff
@@ -141,6 +141,26 @@ FUNCTION( RUN_QMC_APP TESTNAME SRC_DIR PROCS THREADS TEST_ADDED TEST_LABELS ${AR
...
 
+# this disables the extremely slow setup of full system run tests
+# these prevent a reasonable compile debug loop if CMakeLists.txt files are touched since they
+# far exceed even the feature and compiler checks for time consumed when rerunning cmake.
+#
+# On leconte with rerun of cmake with
+# slow tests 27 secs to run cmake
+# with them disable 8 secs
+
+SET( QMC_DISABLE_SLOW_TESTS FALSE CACHE BOOL "Disable the slow cmake system test additions")
```
ye-luo: 1. please move this line to the CTEST section in the top level cmakelist.txt. Keep only functions in this macro.cmake.
2. We use "integration test" instead of "system test" in our documentation and manual. Please align the vocabulary.
3. "slow tests" sounds like the tests are slow to run but here is for tests slow to configure. So better call it QMC_DISABLE_TESTS_CONFIGURED_SLOWLY. Please leave a note before this line saying this will also disable part of deterministic tests and most users should not set this flag specifically.

CMakeLists.txt
```diff
@@ -890,9 +890,11 @@ ENDIF()
 SUBDIRS(src)
 
 IF(NOT QMC_BUILD_SANDBOX_ONLY)
-  SUBDIRS(tests)
-  SUBDIRS(examples)
-  IF(IS_DIRECTORY "${PROJECT_SOURCE_DIR}/nexus")
+  IF( NOT QMC_DISABLE_SLOW_TESTS )
```
ye-luo: Here and the above redefining functions as print are not disabling tests in a consistent way but more "ad hoc".
I'm OK with your current choice of both places and we may learn what minimal testing to keep over time.
I personally would like to keep all the deterministic integration tests in the minimal set.

CMake/macros.cmake
PDoakORNL: Maybe we should update the manual.

Integration tests are usually done without the entire system up, and between a well defined subset of object/modules. QMCPACK has a good bit "accidental" integration testing in the unit tests but very little that was written on purpose.
https://en.wikipedia.org/wiki/Integration_testing

It's hard to be precise but I wouldn't generally consider anything feeding input files to the main executable an integration test.

CMakeLists.txt
PDoakORNL: It's a workaround, the commands are "broken" from a developer productivity standpoint. But its a workaround that anyone improving the build (outside of those commands), debugging, or refactoring will save a ton of time with. I'll update the description of the option to explain it is for development.

A better workaround could take TEST labels into account.

prckent:
Use of test labels is a good idea medium term provided it is technically possible to skip the test construction. 

Since this is a developer productivity improvement that is not user visible I think this version can go in. Clearly a cleaner implementation is desirable, but larger updates to our cmake are really needed first.

prckent:
@ye-luo Good question. @PDoakORNL Is one of these sufficient? Do both disable the configuration? Is one more targeted at slow "cmake" (cmake tests+python+filesystem+...) configuration and the other is runtime?  If this is the case, how about QMC_DISABLE_SLOW_TEST_CONFIGURATION and QMC_DISABLE_SLOW_TESTS ?

I would like to understand the pain points here so that we improve the right part of "cmake" in future.

PDoakORNL:
I messed up updating the option name, their should only be one option, will fix soon

prckent:
Thanks Peter. Would appreciate the note on the pain points as well, but not a blocker on the PR. (We'll be talking with Kitware soon so the info is needed.)

PDoakORNL:
Removed the second option.  Since this is "short" term workaround we don't want that much granularity.  I will make an issue that details the cmake "problem" we would like solved instead.

prckent:
Final selected option is  QMC_NO_SLOW_CUSTOM_TESTING_COMMANDS

prckent:
Seems safe to merge given the prior comments and approvals.

PR #2208, opened 2020-01-10
Fix test_particle

ye-luo:
The original test was made sensitive to PBC. The displacement of the pair of electrons used for checking was exactly half of one of the basis vectors. So numerical sensitivity can actually reverse the direction which is still valid. So I need to drag the electron a bit away from the high symmetry point to avoid this situation.
Not sure if this is related to https://cdash.qmcpack.org/CDash/testDetails.php?test=7129211&build=99071


ye-luo:
On ALCF CI machine, I also have a few nightly builds. Clang 6 was failing from 6 days ago.
https://jenkins-ci.mcs.anl.gov/me/my-views/view/QMCPACK/job/QMCPACK-Clang6-noMPI-MKL19/

PR #2209, opened 2020-01-10
Replace tuple with NLPPJob

ye-luo:
NLPPJob struct is more verbose than tuple.


PDoakORNL:
That’s part of the point. Instead of many lines of comments and cryptic code and debugger code it is more self documenting. It’s not as if it a one time return struct

prckent:
@PDoakORNL OK to merge or would you like to run some checks or make updates?

PR #2210, opened 2020-01-10
Nexus: protect against returned bytes from Popen

jtkrogel:
This PR protects against proliferation of bytes objects through the execution path (and subsequent type mismatch errors caused by them).

@tiihonej will check that this code works.


markdewing:
okay to test

tiihonej:
Based on a few tests, this fix works on Cades (job manager) and locally on workstations (no manager). 

There is another unrelated issue, however: the old nexus pickle files created in python 2.7.x are not loaded correctly due to another encoding mismatch. I'll post an issue for details and discussion.

PR #2212, opened 2020-01-10
Fix distributed dense

fdmalone:
Fixes distributed dense hamiltonian, and adds some fixes for read/write of NOMSD.


markdewing:
okay to test

PR #2213, opened 2020-01-10
Constant size walker properties

PDoakORNL:
In the new drivers the aim is to throw as much of the rambling legacy state machine out as possible.  Part of it that I found while debugging the MPI population control was the rather unfortunate interaction of the Walker "Properties" the DataSet buffer and the mpi send and receives. Until the Walker DataSet is completely purged from the design choosing a constant size for the Walker::Properties seems the most robust fix.  I've added a minimal replacement for the inappropriate use of OhmmsMatrix for the properties. Using this as the properties container abusers of this Walker data member will no longer be able to cause the difficult to diagnose memory corruption bugs via Properties. Objects that make a copy of it and then assign back and forth from Walkers are also forced to use this new type.

Initially I attempted to use the simpler ConstantSizeVector to replace the Properties however RMC and other dusty legacy drivers assume a table of properties.  When I get to rewriting Walker I imagine the Properties will probably be a templated type. I don't like the forcing of multidimensional Properties on every driver implementation because a couple of complex drivers need it.

The size of the Properties allocation is configurable at build time.


PDoakORNL:
I am aware that ForwardWalking.cpp has some real issues

PDoakORNL:
Fix is simplified and prevents crash with mpi and DMCBatched.  Correct mpi population control for DMCBatched and hang caused by same fixed in following PR's. What I think is a legacy CUDA issue remains to be dealt with before merge.

src/AFQMC/CMakeLists.txt
```diff
@@ -67,6 +67,7 @@ IF(ENABLE_CUDA)
   target_link_libraries( afqmc ${CUDA_curand_LIBRARY} )
 ELSE(ENABLE_CUDA)
   ADD_LIBRARY(afqmc ${AFQMC_SRCS})
+  TARGET_LINK_LIBRARIES(afqmc libgfortran.a)
```
ye-luo: checked in by mistake?

src/QMCDrivers/WalkerProperties.h
```diff
@@ -0,0 +1,49 @@
...
+ * \todo: this enum and the handling of "Properties" through the class hierarchy
+ * is hot garbage. Replace, for now we're just making it safe.
+ *
+ * LOCALPOTENTIAL being defined last of the Walker "Properties" is special there
+ * is likely to be some code lurking that depends on it. Keep until this mess is removed.
+
+ * Believe it or not,you need to modify ParticleSet::initPropertyList to match the walker enum if you add to it.
+ * _Don't_, actually work on fixing this.
+ */
+  enum Indexes : int16_t
```
ye-luo: Why not
```
enum class WalkerProperties
{
}
```

src/AFQMC/CMakeLists.txt
PDoakORNL: Yes but I do need this to build with afqmc is on.  Separate issue.

src/QMCDrivers/WalkerProperties.h
PDoakORNL: Requires a `static_cast<int>` for every use as an index. Since other "users" of Walker::Properties use arbitrary int indexes into it for >= WP::NUMPROPERTIES we can't define a helper function that forces accessor to use type WalkerProperties and do the cast inside the accessor.

Once we retire MCWC with the legacy drivers. I'd like to refactor walker and the use of PooledMemory in it. This could allow the Properties issue to be handled in a different manner.

ye-luo: Maybe this could work? Anyway not critical.
```
enum class WalkerProperties : int
{
}
```

ye-luo:
Overall we are moving in good directions. Some details will be discussed and settled after getting the code up and running correctly first. I suggested merging once libgfortran.a gets removed in CMake and CUDA build gets fixed.

src/AFQMC/CMakeLists.txt
prckent: This will break the build on systems without gfortran. Can not merge this line.

PR #2215, opened 2020-01-13
Remove resetTargetParticleSet in BackflowTransformation.h

ye-luo:
resetTargetParticleSet is an unnecessary state machine. The cleaning work needs to be done from the leaves of the call tree.


prckent:
A deterministic backflow test would give more confidence in these changes. Do you have one?

ye-luo:
We don't have a determinstic. I ran the HEG sjb test 16-1, 4-4, 1-16 all pass.

prckent:
There are lots of calls to resetTargetParticleSet in the code outside of backflow. I suspect many are there more out of a concern for safety and not understand/respecting the state machine(s). What is your plan for them?

ye-luo:
All the resetTargetParticleSet should be removed. So far some pieces of code still need it, for example AoS LCAO. The cleaning work needs to be done via the dependence chain gradually SPO->WF->Ham.

prckent:
Keep chopping.

PR #2216, opened 2020-01-13
Enable Theta (limited) testing

ye-luo:
Also clean up scripts of retired machines.


ye-luo:
Unfortunately, Theta is like Titan having a very slow file system. I tried to add only one a128 performance test and got timeout right during performance tests. Probably will try a different queuing strategy.

PR #2218, opened 2020-01-13
Add distance table offload

ye-luo:
This part of code overall is still WIP but I'd like to merge the current status due to other high priority work. This is also good to get some feedback at the current stage. It is a good time to pause and think before more changes.

I made the original distance table offload code in Autumn and I have to drop the branch due to a few significant issues. After more thinking and experimenting, this branch was created to address a few design issues after getting more experience with the initial code.

In this branch, I'm experimenting with QuantumVariables to replace the current RSoA (may be also R in the future). This abstraction allows me to abstract RSoA with and without offload.
Consider this is an attempt to build an abstraction layer in ParticleSet. So I still use ParticleSet to manage the state machine but put the data under the abstraction layer.

This PR also adds ParticleSet::flex_update interface for future implementation for batching. Currently only single walker distance table offload is supported. This can be requested via input but not activated by default even if you build the code with offload.

~~Merge after #2215~~
I still need to add documentation before merging but would like to try CI first.


markdewing:
What does the name 'QuantumVariables' refer to?  Does it relate to the ParticleSet quantum_domain?

prckent:
The fact that this question needs to be asked indicates that the name is bad.

prckent:
Can you do the offload without touching other structures too much?

This PR looks like a complication and the benefits are not clear to me. If we are going to modify the abstractions (needed), why is this the way to go? What are the alternatives?

ye-luo:
> What does the name 'QuantumVariables' refer to? Does it relate to the ParticleSet quantum_domain?

I have no idea about `quantum_domain` does any body know what it is used for? I never touched it over the years. My quick guess is that ion ParticleSet quantum_domain = classic, electron ParticleSet quantum_domain is quantum? So this is just a labelling of ParticleSet?

ye-luo:
> Can you do the offload without touching other structures too much?
> 
> This PR looks like a complication and the benefits are not clear to me. If we are going to modify the abstractions (needed), why is this the way to go? What are the alternatives?

I chose QuantumVariables when I was thinking only electron ParticleSet in mind. So the quantum variables are electron postions or positions+spins with/without offload. Now I just realized that ParicleSet is also used for ions. So maybe DynamicVariables are more suitable considering ions are not quantum?

I need to keep a copy of RSoA on GPU but I don't want to put RSoA_gpu next to RSoA and have R, RSoA, RSoA_gpu in ParticleSet directly. QuantumVariable was introduce to abstract all kinds of dynamic variables in the ParticleSet. We discussed about template ParticleSet for spin, however, this results in template TWF and Hamiltonian. We are not there yet. Using QuantumVariable at least help us identify what abstraction is need in ParticleSet and make change to template easier in the future.

RSoA is used only within ParticleSet calculating SoA distance tables. So this is the only member variable touched in ParticleSet. The whole code relies more on R instead of RSoA. So touching RSoA is the minimal change I can do to introduce the offload code without touch the whole code base (if touching R).

ye-luo:
The abstract change is in the first part of this PR. So I created a separate PR #2221. with half of the commits from this PR. Should help reviewing

prckent:
quantum_domain should probably be deleted in a separate PR. I am asking around but it appears to be of archaeological interest only.

jtkrogel:
Particles are in two quantum domains: "quantum" (electrons) and "classical" (ions).  These are particle related labels that are stored in QMCHamiltonianBase objects.

With "QuantumVariables" are you just looking to weaken API dependencies on the full ParticleSet?  Why is there a need to cluser/label some of the variables in ParticleSet rather than use their individual types.

ye-luo:
> With "QuantumVariables" are you just looking to weaken API dependencies on the full ParticleSet? Why is there a need to cluser/label some of the variables in ParticleSet rather than use their individual types?

1. I'm not looking forward to weaken WF and Hamiltonian API dependency on full ParticleSet. Do you have any motivation and idea to explore in this direction?
2. We need to find a way to stop increasing APIs. makeMove and makeMoveWithSpin + makeMoveWithGPU... the number of API blows up and also the number of drivers blows up. If we can build an abstraction properly. There is only a single makeMove and one driver.

ye-luo:
Test this please

ye-luo:
Will recreate a new PR with better description reflecting the current status and it will be easier for reviewing.

PR #2221, opened 2020-01-14
Introduce DynamicCoordinates abstraction

ye-luo:
In this branch, I'm experimenting with DynamicCoordinates to replace the current RSoA (may be also fusing R in the future).
```C++
public:
VectorSoaContainer<RealType, DIM> RSoA;
```
to
```C++
protected
std::unique_ptr<DynamicCoordinates> coordinates_;
```
This abstraction allows me to abstract RSoA with and without offload.

Consider this is an attempt to build an abstraction layer inside ParticleSet. So I still use ParticleSet to manage the state machine but put the data under the abstraction layer.

I don't have a good idea of replacing the object name RSoA so far. Suggestions are welcome.


ye-luo:
Q from @markdewing 
> What does the name 'QuantumVariables' refer to? Does it relate to the ParticleSet quantum_domain?

I have no idea about `quantum_domain` does any body know what it is used for? I never touched it over the years. My quick guess is that ion ParticleSet quantum_domain = classic, electron ParticleSet quantum_domain is quantum? So this is just a labelling of ParticleSet?

I chose QuantumVariables when I was thinking only electron ParticleSet in mind. So the quantum variables are electron postions or positions+spins with/without offload. Now I just realized that ParicleSet is also used for ions. So maybe DynamicVariables are more suitable considering ions are not quantum?

ye-luo:
Q from @prckent 
> Can you do the offload without touching other structures too much?
> 
> This PR looks like a complication and the benefits are not clear to me. If we are going to modify the abstractions (needed), why is this the way to go? What are the alternatives?

I need to keep a copy of RSoA on GPU but I don't want to put RSoA_gpu next to RSoA and have R, RSoA, RSoA_gpu in ParticleSet directly. QuantumVariable was introduce to abstract all kinds of dynamic variables in the ParticleSet. We discussed about template ParticleSet for spin, however, this results in template TWF and Hamiltonian. We are not there yet. Using QuantumVariable at least help us identify what abstraction is need in ParticleSet and make change to template easier in the future.

RSoA is used only within ParticleSet calculating SoA distance tables. So this is the only member variable touched in ParticleSet. The whole code relies more on R instead of RSoA. So touching RSoA is the minimal change I can do to introduce the offload code without touch the whole code base (if touching R).

PDoakORNL:
ParticleCoordinates. Right now we have position and spin. Coordinates are ordered vectors on a particular basis. If they were actual types correctness of code would be easier to reason about and perhaps assured at compile time.

I’ll need a while to review this and I need to get to batched mpi fix in first.

prckent:
> I’ll need a while to review this and I need to get to batched mpi fix in first.
This is the correct priority imo.

markdewing:
So, if I understand correctly, QuantumVariables generalizes positions to encompass spin variables as well.    Are there likely to be any other variables this needs to handle, or does position and spin cover it?  
Would ParticlePositionsOrSpins (or ParticlePositionsAndSpins) be a reasonable name?  (I'm not sure I like the names, mostly trying to provoke thought)

Another issue: QuantumVariable is a domain abstraction (generalizing position/spin), or at least the name is a domain abstraction, but QuantumVariableKind is an implementation abstraction (offload or not)

ye-luo:
@markdewing thank you for the feedback. So far `quantum` receives most complaints.
I tend to choose `DynamicCoordinates` for the base class. It means coordinates of particles which may be moved. The derived class can be `RealSpacePositionsAndSpins` in the spin case. For the variable, probably just `coordinates`.

I'm not dealing with spin right now. So QuantumVariableKind only has offload. I'm adding the minimal change to support offload without if-else and ifdef.

How to incorporate spin needs to be experimented because consistent changes will be needed in WF and Ham to update return values. I'm not 100% sure the abstraction I added is useful for incorporating spin but an abstraction is needed regardless how to implement the abstraction.

ye-luo:
Maybe I was over ambitious. At the moment, the real use of this abstraction is to hide offload code from ParticleSet. Let me know if there are any concerns other than the bad name I chose initially. I'd like to get this PR settled soon.

ye-luo:
I updated the class and object name. This PR is ready for review.

src/Particle/DynamicCoordinates.h
```diff
@@ -0,0 +1,64 @@
...
+#include "Configuration.h"
+#include "OhmmsSoA/Container.h"
+
+namespace qmcplusplus
+{
+/** enumerator for DynamicCoordinates kinds
+ */
+enum class DynamicCoordinateKind
+{
+  QV_POS, // SoA positions
```
markdewing: Prefix is still QV from previous name.   Should be DC now?

src/Particle/DynamicCoordinates.h
```diff
@@ -0,0 +1,64 @@
...
+//
+// File developed by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//
+// File created by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+/** @file DynamicCoordinates.h
+ */
+#ifndef QMCPLUSPLUS_QUANTUM_VARIABLE_H
```
markdewing: include guard name should be updated

prckent:
DynamicCoordinates is definitely a better name than QuantumVariables, and I don't have a better name to offer currently.

prckent:
I think this is a step in the right direction and propose we merge Monday evening unless other comments are received.

prckent:
Test this please

prckent:
Based on the previous ORNL and ALCF CI runs passing and current ALCF CI pass, this is safe to merge.

PR #2232, opened 2020-01-17
Mpi walker transmission fix

PDoakORNL:
putting this here mostly to get CI, but nearly there.

### DIFF
--- a/CMake/ClangCompilers.cmake
+++ b/CMake/ClangCompilers.cmake
@@ -22,14 +22,14 @@ ENDIF(QMC_OMP)
 # Set clang specific flags (which we always want)
 ADD_DEFINITIONS( -Drestrict=__restrict__ )
 
-SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -fomit-frame-pointer -fstrict-aliasing")
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fomit-frame-pointer -fstrict-aliasing -D__forceinline=inline")
+SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -fstrict-aliasing")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fstrict-aliasing -D__forceinline=inline")
 
 # Set extra optimization specific flags
-SET( CMAKE_C_FLAGS_RELEASE     "${CMAKE_C_FLAGS_RELEASE} -ffast-math" )
-SET( CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -ffast-math" )
-SET( CMAKE_C_FLAGS_RELWITHDEBINFO     "${CMAKE_C_FLAGS_RELWITHDEBINFO} -ffast-math" )
-SET( CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -ffast-math" )
+SET( CMAKE_C_FLAGS_RELEASE     "${CMAKE_C_FLAGS_RELEASE} -fomit-frame-pointer -ffast-math" )
+SET( CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fomit-frame-pointer -ffast-math" )
+SET( CMAKE_C_FLAGS_RELWITHDEBINFO     "${CMAKE_C_FLAGS_RELWITHDEBINFO} -fomit-frame-pointer -ffast-math" )
+SET( CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -fomit-frame-pointer -ffast-math" )
 
 # Set extra debug flags
 SET( CMAKE_C_FLAGS_DEBUG     "${CMAKE_C_FLAGS_DEBUG} -fno-omit-frame-pointer -fstandalone-debug" )
--- a/src/QMCDrivers/Crowd.cpp
+++ b/src/QMCDrivers/Crowd.cpp
@@ -41,15 +41,8 @@ void Crowd::addWalker(MCPWalker& walker, ParticleSet& elecs, TrialWaveFunction& 
 
 void Crowd::loadWalkers()
 {
-  auto it_walker       = mcp_walkers_.begin();
-  auto it_walker_elecs = walker_elecs_.begin();
-  //flex walkers here
-  while (it_walker != mcp_walkers_.end())
-  {
-    (*it_walker_elecs).get().loadWalker(*it_walker, true);
-    ++it_walker;
-    ++it_walker_elecs;
-  }
+  for(int i = 0; i < mcp_walkers_.size(); ++i)
+    walker_elecs_[i].get().loadWalker(mcp_walkers_[i], true);
 }
 
 void Crowd::setRNGForHamiltonian(RandomGenerator_t& rng)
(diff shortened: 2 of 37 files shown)


CMake/ClangCompilers.cmake
```diff
@@ -22,14 +22,14 @@ ENDIF(QMC_OMP)
 # Set clang specific flags (which we always want)
 ADD_DEFINITIONS( -Drestrict=__restrict__ )
 
-SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -fomit-frame-pointer -fstrict-aliasing")
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fomit-frame-pointer -fstrict-aliasing -D__forceinline=inline")
```
ye-luo: Was this causing problems in debug build?

src/QMCDrivers/Crowd.cpp
```diff
@@ -41,15 +41,8 @@ void Crowd::addWalker(MCPWalker& walker, ParticleSet& elecs, TrialWaveFunction&
...
-  auto it_walker_elecs = walker_elecs_.begin();
-  //flex walkers here
-  while (it_walker != mcp_walkers_.end())
-  {
-    (*it_walker_elecs).get().loadWalker(*it_walker, true);
-    ++it_walker;
-    ++it_walker_elecs;
-  }
+  for(int i = 0; i < mcp_walkers_.size(); ++i)
+    walker_elecs_[i].get().loadWalker(mcp_walkers_[i], true);
```
ye-luo: Like this much better than iterators.

src/QMCDrivers/tests/test_DMCBatched.cpp
```diff
@@ -30,7 +30,7 @@ TEST_CASE("DMCBatched::calc_default_local_walkers", "[drivers]")
 
 
   auto testWRTWalkersPerRank = [&dtest](int walkers_per_rank) {
-                                 DMCBatched dmc_batched(std::move(dtest()));
```
ye-luo: No need to fix here if it works now, but it is unsafe. When I was working on https://github.com/QMCPACK/qmcpack/pull/2221, I noticed the MCPopulation move constructor and move assign operator cannot be compiler generated by default because ParticleSet cannot be moved. The reason is that the compiler tries to use copy assign operator of ParticleSet which should be either deleted or implemented just like the copy constructor.

src/QMCHamiltonians/OperatorBase.cpp
```diff
@@ -42,7 +42,6 @@ OperatorBase::OperatorBase() : myIndex(-1), Dependants(0), Value(0.0), tWalker(0
  */
 void OperatorBase::mw_evaluate(const RefVector<OperatorBase>& O_list, const RefVector<ParticleSet>& P_list)
 {
-#pragma omp parallel for
```
ye-luo: Do not change this. This is needed so far. I verified that this works very well with OMP_NUM_THREADS=16 and 1 crowd in the input.

CMake/ClangCompilers.cmake
PDoakORNL: Yes it appears that the first -fomit -fno-omit wins.

src/QMCHamiltonians/OperatorBase.cpp
PDoakORNL: So now this is OMP_NUM_THREADS=1,16 behavior but induced by forcing one crowd? I'll put this back in for now (as I expect this has to do with AMD support), but  we need to continue to abstract threading and the actual runtime thread setup needs to be handled in one place and not via a proprietary colleciton of ENV. The adhoc nature of OMP thread safety and its opaqueness when reading the source and compiling time is going to continue to be an issue until we do.

ye-luo: If running 1 crowd is requested, the first OpenMP level is an inactive threading level, there is no need to set nested threading. When using a list of thread numbers like OMP_NUM_THREADS=1,16, they refer to active threading levels. I used this on Summit to compare with the CUDA code which is conceptually 1 crowd.

CMake/ClangCompilers.cmake
ye-luo: CMAKE_C_FLAGS is added to all builds.
For a build release CMake invokes both $CMAKE_C_FLAGS $CMAKE_C_FLAGS_RELEASE.

PDoakORNL: Not sure what you mean the CMAKE_C(XX)_FLAGS should not contain `-f-omit-frame-pointer` if you don't want that in the flags for the debug build.  This fixes that. While leaving this optimization in the default RelWithDebugInfo and Release.

src/QMCDrivers/tests/test_DMCBatched.cpp
PDoakORNL: Since most of this PR has been long exploration of the unintended consequences of "unsafe" things I'd like to just leave this fixed.

src/QMCHamiltonians/OperatorBase.cpp
PDoakORNL: I get what you mean about the semantics of openmp but this throws threading control back to implicit openmp threading.  In application threading model we should be using to reason about thread scopes and synchronization this is a 1,16

prckent:
General comment: Getting MPI working with DMC is blocking the release. Once this is in the focus should be on getting correct statistics. Threading philosophy and design discussions will need to be done, but later.

src/QMCDrivers/DMC/DMCBatched.cpp
```diff
@@ -457,11 +457,13 @@ void DMCBatched::setMultiplicities(const DMCDriverInput& dmcdriver_input,
   auto setMultiplicity = [&dmcdriver_input, &rng](MCPWalker& walker) {
     constexpr RealType onehalf(0.5);
     constexpr RealType cone(1);
-    RealType M = walker.Weight;
+    RealType M;
```
ye-luo: I submitted #2243 to follow up the remaining issue here.

src/QMCHamiltonians/OperatorBase.cpp
ye-luo: We can discuss in person with more details. This raw OpenMP implementation doesn't ignore the crowd level concurrency. Both crowd level and walker level are needed.

ye-luo:
Not sure if this should fix the non-MPI case as well or not.
I still get short test failure
```
Command: "/scratch/opt/cleanup/PDoakORNL/tests/scripts/check_scalars.py" "--ns" "3" "--series" "1" "-p" "qmc_short_vmcbatch_dmcbatch" "-e" "2" "--le" "-21.844975 0.00779"
Directory: /scratch/opt/cleanup/PDoakORNL/build_gnu_real/tests/solids/diamondC_2x1x1_pp/short-diamondC_2x1x1_pp-vmcbatch_dmcbatch_sdj-1-16
"short-diamondC_2x1x1_pp-vmcbatch_dmcbatch_sdj-1-16-totenergy" start time: Jan 23 21:26 CST
Output:
----------------------------------------------------------
Tests for series 1
  Testing quantity: LocalEnergy
    reference mean value     : -21.84497500
    reference error bar      :   0.00779000
    computed  mean value     : -21.68150807
    computed  error bar      :   0.01288663
    pass tolerance           :   0.02337000  (  3.00000000 sigma)
    deviation from reference :   0.16346693  ( 20.98420103 sigma)
    error bar of deviation   :   0.01505820
    significance probability :   1.00000000  (gaussian statistics)

Test status: fail
<end of output>
```

ye-luo:
I ran the same input file with 2 MPI and the result is still far from reference value.
```
qmc_short_vmcbatch_dmcbatch  series 0  -21.645694 +/- 0.000000   1.215875 +/- 0.000000   0.0562 
qmc_short_vmcbatch_dmcbatch  series 1  -21.681356 +/- 0.011070   0.988482 +/- 0.012590   0.0456 
```

Issue #2224, opened 2020-01-15
Error thrown when no ions are found in HDF5 file when converting XML output from QBox.


kryczko:
I have an 8 electron system, with 4 SPOs from Qbox. I am using a custom external potential which is read in from an HDF5 file. This is the following output of qmcpack:
```
$ qmcpack qmcpack_in.xml 
--------------------------------------------------------------------------
[[62075,1],0]: A high-performance Open MPI point-to-point messaging module
was unable to find any relevant network interfaces:

Module: OpenFabrics (openib)
  Host: thundercracker

Another transport will be used instead, although this may result in
lower performance.

NOTE: You can disable this warning by setting the MCA parameter
btl_base_warn_component_unused to 0.
--------------------------------------------------------------------------
Rank =    0  Free Memory = 170503 MB
  Input file(s): qmcpack_in.xml 

=====================================================
                    QMCPACK 3.8.9

       (c) Copyright 2003-  QMCPACK developers

                    Please cite:
 J. Kim et al. J. Phys. Cond. Mat. 30 195901 (2018)
      https://doi.org/10.1088/1361-648X/aab9c3

  Git branch: develop
  Last git commit: 915d81efd172bc521ae5a997735f5d6683e5ff5a-dirty
  Last git commit date: Wed Jan 15 09:13:28 2020 -0500
  Last git commit subject: Merge pull request #2223 from prckent/oxygen_setup
=====================================================
  Global options 

  Total number of MPI ranks = 1
  Number of MPI groups      = 1
  MPI group ID              = 0
  Number of ranks in group  = 1
  MPI ranks per node        = 1
  OMP 1st level threads     = 48
  OMP nested threading disabled or only 1 thread on the 2nd level

  Precision used in this calculation, see definitions in the manual:
  Base precision      = double
  Full precision      = double

  Structure-of-arrays (SoA) optimization enabled

  Input XML = qmcpack_in.xml

  Project = qmc
  date    = 2020-01-15 09:56:09 EST
  host    = thundercracker.sims.nrc.ca

 Random Number
 -------------
  Offset for the random number seeds from input file (mod 1024): 13

  Range of prime numbers to use as seeds over processors and threads = 47-307


 Lattice
 -------
  Simulation cell radius   = 10.235000 bohr
  Wigner-Seitz cell radius = 10.235000 bohr

  Lattice (bohr):     20.4700000000      0.0000000000      0.0000000000
                       0.0000000000     20.4700000000      0.0000000000
[...]

PR #2234, opened 2020-01-21
Allowing the ability to have no ions and checking ion counts in HDF5 and XML.

kryczko:
These changes are result of issue #2224, to support having no ions. This includes making checks between an HDF5 file that holds orbitals calculated from Qbox and the XML input file.

### DIFF
--- a/src/Particle/ParticleSet.cpp
+++ b/src/Particle/ParticleSet.cpp
@@ -2,7 +2,7 @@
 // This file is distributed under the University of Illinois/NCSA Open Source License.
 // See LICENSE file in top directory for details.
 //
-// Copyright (c) 2016 Jeongnim Kim and QMCPACK developers.
+// Copyright (c) 2020 QMCPACK developers.
 //
 // File developed by: Ken Esler, kpesler@gmail.com, University of Illinois at Urbana-Champaign
 //                    Luke Shulenburger, lshulen@sandia.gov, Sandia National Laboratories
@@ -12,6 +12,7 @@
 //                    Ye Luo, yeluo@anl.gov, Argonne National Laboratory
 //                    Mark A. Berrill, berrillma@ornl.gov, Oak Ridge National Laboratory
 //                    Mark Dewing, markdewing@gmail.com, University of Illinois at Urbana-Champaign
+//                    Kevin Ryczko, kryczko@uottawa.ca, University of Ottawa
 //
 // File created by: Jeongnim Kim, jeongnim.kim@gmail.com, University of Illinois at Urbana-Champaign
 //////////////////////////////////////////////////////////////////////////////////////
@@ -152,7 +153,7 @@ void ParticleSet::resetGroups()
   if (nspecies == 0)
   {
     // APP_ABORT("ParticleSet::resetGroups() Failed. No species exists");
-    app_log() << " There are no particles in this particle set." << std::endl;
+    app_warning() << "ParticleSet::resetGroups(): There are no particles in this particle set." << std::endl;
   } else {
     int natt = mySpecies.numAttributes();
     int qind = mySpecies.addAttribute("charge");
--- a/src/Particle/ParticleSet.cpp
+++ b/src/Particle/ParticleSet.cpp
@@ -151,71 +151,73 @@ void ParticleSet::resetGroups()
   int nspecies = mySpecies.getTotalNum();
   if (nspecies == 0)
   {
-    APP_ABORT("ParticleSet::resetGroups() Failed. No species exisits");
-  }
-  int natt = mySpecies.numAttributes();
-  int qind = mySpecies.addAttribute("charge");
-  if (natt == qind)
-  {
-    app_log() << " Missing charge attribute of the SpeciesSet " << myName << " particleset" << std::endl;
-    app_log() << " Assume neutral particles Z=0.0 " << std::endl;
-    for (int ig = 0; ig < nspecies; ig++)
-      mySpecies(qind, ig) = 0.0;
-  }
-  for (int iat = 0; iat < Z.size(); iat++)
-    Z[iat] = mySpecies(qind, GroupID[iat]);
-  natt        = mySpecies.numAttributes();
-  int massind = mySpecies.addAttribute("mass");
-  if (massind == natt)
-  {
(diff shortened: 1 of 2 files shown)


markdewing:
okay to test

src/Particle/ParticleSet.cpp
```diff
@@ -151,71 +152,73 @@ void ParticleSet::resetGroups()
...
-  else
-    app_log() << "  Distinctive masses for each species " << std::endl;
-  for (int iat = 0; iat < Mass.size(); iat++)
-    Mass[iat] = mySpecies(massind, GroupID[iat]);
-  std::vector<int> ng(nspecies, 0);
-  for (int iat = 0; iat < GroupID.size(); iat++)
-  {
-    if (GroupID[iat] < nspecies)
-      ng[GroupID[iat]]++;
+    // APP_ABORT("ParticleSet::resetGroups() Failed. No species exists");
```
markdewing: Delete the old code.

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
```diff
@@ -124,106 +126,119 @@ bool EinsplineSetBuilder::ReadOrbitalInfo_ESHDF(bool skipChecks)
...
-    Super2Prim.resize(SourcePtcl->R.size(), -1);
-    std::vector<int> prim_atom_counts;
-    prim_atom_counts.resize(IonPos.size(), 0);
-    for (int i = 0; i < SourcePtcl->R.size(); i++)
+  // Check the number of ions in the XML and 
+  // HDF5. Skip the checks in some situations but 
+  // not others. See Issue #2224.
+
+  if (SourcePtcl->R.size() != IonPos.size()) {
+    app_error() << "The ion sizes differ between the XML and HDF5.\n";
```
markdewing: This error message should print the sizes, to make debugging issues faster  (test_einset is failing).

kryczko: I have now modified it to print out the different sizes.

kryczko:
any ideas on why the tests are failing?

markdewing:
Compile failure in EinsplineSetBuilderESHDF.fft.cpp, line 136.   There's a semicolon ending the line that shouldn't be there.

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
```diff
@@ -124,106 +126,121 @@ bool EinsplineSetBuilder::ReadOrbitalInfo_ESHDF(bool skipChecks)
...
+    app_error() << "The ion sizes differ between the XML and HDF5.\n" 
+                << "Number of ions in XML : " << SourcePtcl->R.size() << std::endl
+                << "Number of ions in HDF5: " << IonPos.size() << std::endl;
+    abort();
+  }
+
+  if (SourcePtcl->R.size() == 0 && IonPos.size() == 0) {
+    app_log() << "There are no ions in XML nor HDF5. No further checks will be performed.\n";
+  } else {
+    if (Super2Prim.size() == 0)
```
markdewing: You might also use the previous check to set a variable (checkCellConsistency)
```
bool checkCellConsistency = true;
if (SourcePtcl ....)
   // message
   checkCellConsistency = false;
}
if (Super2Prim.size() == 0 && checkCellConsistency)
{
...
```

That would avoid indenting a bunch of code.   Unfortunately, the diff algorithm is not recognizing this block of code is the same as before with added indentation, and it makes the changes look more extensive than they really are.

kryczko: yeah, I agree. This is what I will do.

markdewing:
The unit tests are failing in test_einset.cpp for the diamond_2x1x1 test.
```
ERROR The ion sizes differ between the XML and HDF5.
Number of ions in XML : 4
Number of ions in HDF5: 2
```
Because of tiling, it seems the number of electron may not match between HDF and XML input (??? Someone with more knowledge will have to answer?)
This is also probably why the Super2Prim check is so complicated - it's checking that all the ions map correctly, not that they match one-to-one.

The easiest fix would be to remove the check that the XML and HDF ions numbers match.

kryczko:
I think the check has to be done elsewhere where I am hitting Fatal Error. Aborting at Einspline needs the source particleset

kryczko:
any ideas where this error is being printed from? @markdewing

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
```diff
@@ -124,8 +126,23 @@ bool EinsplineSetBuilder::ReadOrbitalInfo_ESHDF(bool skipChecks)
...
+  // Check the number of ions in the XML and 
+  // HDF5. Skip the checks in some situations but 
+  // not others. See Issue #2224.
+
+  if (SourcePtcl->R.size() != IonPos.size()) {
+    app_error() << "The ion sizes differ between the XML and HDF5.\n" 
+                << "Number of ions in XML : " << SourcePtcl->R.size() << std::endl
+                << "Number of ions in HDF5: " << IonPos.size() << std::endl;
+    abort();
+  }
```
markdewing: This check should be removed.    The QMC wavefunction can be created by tiling the DFT wavefunction, in which case the number of ions will not match.    I suppose this could be checked with (# of tiles)*(DFT Ions) == (QMC ions), but that's not necessary now.

markdewing: If I remove this check, the wavefunction unit tests pass for my build.

kryczko: I'm confused about how to proceed, as mentioned above, I don't even get to this error now. I hit another error.

kryczko: With the new change made by @ye-luo, do I need to define an empty particle set?

prckent:
@kryczko Mark will comment here soon.

ye-luo:
superseded by #2259

PR #2236, opened 2020-01-22
Bypass createESParticleSet check if target has been built

ye-luo:
#2224 hit this code path but I cannot see a good reason.
I added the desired behaviour in the source comment.

After this fix, #2224 will hit
```
Fatal Error. Aborting at Einspline needs the source particleset
```
and we can implement the desired behaviour discussed in #2224.


kryczko:
can someone review this merge please so I can start another PR as discussed #2224

PR #2239, opened 2020-01-22
Nexus: follow Cades transition to Slurm

jtkrogel:
Cades at ORNL is transitioning to Slurm.  Part of the machine still uses Moab.  Support both for now.


ye-luo:
Okay to test

PR #2241, opened 2020-01-23
Nexus: Allow use of `dr` in `SpinDensityAnalyzer`

mcbennet:
Currently, the `SpinDensityAnalyzer` will not accept the case when
the user specifies `dr` rather than `grid`. This PR allows users
to specify either.


markdewing:
okay to test

Issue #2238, opened 2020-01-22
Force estimator broken by recent commits


tiihonej:
QMCPACK runs with the ACForce estimator crash at launch with the latest code (since around Jan 16). The error message is something like the following:

>   Resetting Properties of the walkers 1 x 50
ERROR You cannot resize a constant size matrix beyond its initial max dimensions
Fatal Error. Aborting at Unhandled Exception
*** The MPI_Abort() function was called after MPI_FINALIZE was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
[psi3:13270] Local abort after MPI_FINALIZE started completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!

I take suggestions to better pinpoint where this is coming from. Sample inputs are attached:
[vmc_force_jan2020.tar.gz](https://github.com/QMCPACK/qmcpack/files/4099192/vmc_force_jan2020.tar.gz)

jtkrogel:
Sounds like a good target for addtional unit/deterministic/integration tests.

prckent:
Ack! A simple view of this would be that we need a simple integration test and preferably a deterministic integration test that uses ACForce. However, better would be a unit or small scale integration test that can catch this.

tiihonej:
#2242 Did not fix it for me. I could write a simple integration test for ACForce, @prckent . Actually, the above inputs slightly modified would probably work?

prckent:
@tiihonej Please modify an existing test, e.g. LiH dimer. This will avoid introducing new files for the wavefunctions and potentials. Of course, this assumes that LiH triggers the test. I suspect that parsing the force to check the numerical results will require writing some new python code in the test system, but for now that is not essential: the key test is whether the code crashes or not.

tiihonej:
@prckent Modifying existing file should be alright and, and LiH already covers a great deal of force-related features (the only exception probably being the derivatives of some higher-order WF-components). Testing against the values of the Pulay terms should be straightforward and catch all there is to catch regarding the actual performance.

PDoakORNL:
@tiihonej Do you mean you get the same resize error?

tiihonej:
@PDoakORNL Nevermind that, there was something else wrong with my build. I did a fresh checkout and everything works. I'll push an integration test and then close this issue.

tiihonej:
There are a few issues with adding an integration test for the current ACForce estimator.

[...]

PR #2242, opened 2020-01-24
Force estimator fix

PDoakORNL:
Additional fixes addressing issues with #2238 
many of the evalSource... functions show assumption that the ions reference held as state and the ions passed in are nearly equivalent, at least from forces they are not and segmentation faults result.  Of course some of this may have appeared to work in Release builds, but it wasn't.  I still have crashes with the legacy driver, forces, and OMP_NUM_THREADS > 0, but they are caused by accesses within the omp sections (race condition?) and I leave that to someone else.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,6 +155,10 @@ SET (EXECUTABLE_OUTPUT_PATH ${qmcpack_BINARY_DIR}/bin CACHE PATH "Single output 
 
 ######################################################################
 # Set the compiler-time parameters
+# WALKER_MAX_PROPERTIES max number of observables + 12 or so standard
+#   props. Things like forces are per particle so you may need this to
+#   rather large.  Each property increases walker size by
+#   sizeof(FULLPRECREALTYPE)
 # OHMMS_DIM =  dimension of the problem
 # OHMMS_INDEXTYPE = type of index
 # OHMMS_PRECISION  = base precision, float, double etc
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -163,7 +167,7 @@ SET (EXECUTABLE_OUTPUT_PATH ${qmcpack_BINARY_DIR}/bin CACHE PATH "Single output
 # QMC_MPI =  enable MPI
 # QMC_OMP = enable OMP
 ######################################################################
-SET(WALKER_MAX_PROPERTIES 32 CACHE STRING "Maximum number of properties tracked by walkers")
```
ye-luo: I assume requiring recompiling the code for using more estimators is just a temporary situation.
Cmake doesn't like int https://github.com/QMCPACK/qmcpack/pull/2231

prckent:
@ye-luo I believe that previous use of an int here resulted in cmake warnings across a large number of cmake versions, hence a string is currently used. We do this for many other variables. Any change should be verified to not introduce warnings on cdash.

CMakeLists.txt
PDoakORNL: Yes this should be temporary.  It should be properly addressed in refactoring the PooledMemory use in Walker,  Estimator and Hamiltonian architecture.

ye-luo: Some background info. There are PooledMemory and PooledData used in different places. In the old days, PooledData was the only one used. Since any change on them affects many classes, we'd better discuss what is the new design before making big PRs.

PR #2244, opened 2020-01-24
Working Population Control for DMCBatched driver

### LINKED ISSUE #2243: Uninitialized value M in DMCBatched::setMultiplicities
GCC warning detects this in #2232 
The original code in QMCUpdateBase::setMultiplicity sets M = walker.Weight.
DMCBatched::setMultiplicities is not doing the same way.
Please explain the change and fix the warning as well.

PDoakORNL:
fixes #2243


PDoakORNL:
DMCBatching population control gives the appearance of working well.
 
One more commit is in coming with a multiple rank test for short vmcbatch_dmcbatch test
it passes the qmc_short_vmcbatch_dmcbatch 1 16 test already.

ye-luo:
short-diamondC_2x1x1_pp-vmcbatch_dmcbatch_sdj works in both 1x16 and 2x16 runs if I build the code with GNU. However, I noticed failing unit tests deterministic-unit_test_drivers and deterministic-unit_test_drivers_mpi as well as the diamondC integration test if I build the application with Intel compiler.

prckent:
Excellent to finally have this.

Regarding deterministic tests with multiple MPI tasks, we don't have many of these yet for the legacy code since we concentrated on threaded tests. There can be plenty of sensitivities in the MPI e.g. since no guarantees are made about the order of operations. Possibly the legacy code has sensitivities we need to look at as well as the code here. Proving out the population control is best done via unit tests.

PDoakORNL:
I will take a look at Intel,  the unit tests must pass.

PDoakORNL:
So right now there is a problem with the statistics of the multi rank batched walker population.  It no longer crashes or goes nuts, but the error is ~30 sigma.  Exploring.  Perhaps this PR should go as it stands because this is much closer to working than what is in develop.

prckent:
Agreed. Better to have more small PRs.

In the current PR, is the 1 MPI * many threads result still correct?

PDoakORNL:
yes with my current compiler mix

PDoakORNL:
I am building with intel now to look at @ye-luo 's reported unit test situation.

PDoakORNL:
Took a long time to get a working intel environment off of oxygen. I see no unit test failures with intel. So this is ready to go.

markdewing:
The return value in doNotBranch was set to use the weight in commit 9e0b9808392ffb67313db08ab4a7936b57ff7ab3

In this PR it's changed to the population.   I'm curious about the context of how the return value is used.
Presumably, doNotBranch is only called during the warmup stage, so the value doesn't matter much.  Although, it's being changed here so the return value must have some effect.

PDoakORNL:
In SimpleFixedNodeBranch it is assigned to a pop_now, I will change it back for legacy although I noticed no difference and guessed this was just lucky and since it violated the naming and documentation of the caller.

In the new driver It is expected to be the population. I have found the fog over weight, walker existence and multiplicity to be of great annoyance in refactoring/rewriting the population control.

PDoakORNL:
@markdewing I've reverted the changes to legacy code in WalkerControlBase.cpp.

ye-luo:
doNotBranch is called at the first step of DMC. In the first step, the trial energy is from VMC and to high. So the weight will increase a lot. If we do the branching right away, the population explodes. Instead "doNotBranch" only calculate the total weight (total population if branching is executed) which is used to adjust the weight calculation in the next step. Then there is no population explosion.

So I think returning `curData[WALKERSIZE_INDEX]` is bad also for the new code.

ye-luo:
On another machine, unit tests pass but got
889 - short-diamondC_2x1x1_pp-vmcbatch_dmcbatch_4r_sdj-4-4 (Failed)

PDoakORNL:
Sadly that one still fails everywhere

PR #2250, opened 2020-01-27
Nexus: enable radial density analysis

jtkrogel:
This PR contains tool to support analysis of the momentum distribution.  Just basic reading/plotting capabilities so far.  More to follow.


jtkrogel:
This should be merged now.  It is already used in production analysis for the ABOx project.

Moreover some of the symmetry related updates are needed for new development related to RMG support.

ye-luo:
Do we have documentation or example for users?

prckent:
@ye-luo Good question. We can't keep adding features to Nexus with no docs or mention.

prckent:
I made https://github.com/QMCPACK/qmcpack/issues/2716 for docs discussion.

ye-luo (approved):

ye-luo:
Okay to test

PR #2251, opened 2020-01-27
Use H5Eset/get_auto version 2 APIs

brtnfld:
Changed from using H5Eset_auto and H5Eget_auto to H5Eset_auto2 and H5Eget_auto2. The version 2 APIs were introduced in HDF5 1.8.0. QMCPACK already specifies version 2 APIs (H5Dcreate2), so this introduces no new requirements in terms of minimum HDF5 version required.

These changes are needed because the HDF5 DAOS VOL connector does not have the version 1 APIs implemented and hence will not work DAOS when using HDF5.


markdewing:
okay to test

PR #2255, opened 2020-01-28
Nexus: harden boolean rendering

jtkrogel:
Only allow true/false (or equivalent) values and always accept ("yes"/"no") when rendering from internal Python values to QMCPACK inputs for booleans.


markdewing:
okay to test

jtkrogel:
Hold please.  I need to look into some test failures related to the processing of NoneType.

jtkrogel:
Ready now.

PR #2257, opened 2020-01-29
Nexus: Allow for atom symbol indexing

mcbennet:
Previously, if get_atomic_numbers is called after the user supplies an indexed atomic symbol for an element in structure.elem (e.g., elem=['Si1', 'Si2'] etc.), the code would abort. The following PR, allows the user to use indexed labels.

### DIFF
--- a/nexus/lib/structure.py
+++ b/nexus/lib/structure.py
@@ -4989,7 +4989,13 @@ class Structure(Sobj):
     def get_atomic_numbers(self):
         an = []
         for e in self.elem:
-            an.append(ptable[e].atomic_number)
+            if e[1:].isdigit():
+                an.append(ptable[e[0:1]].atomic_number)
+            elif e[2:].isdigit():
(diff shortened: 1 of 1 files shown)


markdewing:
okay to test

nexus/lib/structure.py
```diff
@@ -4989,7 +4989,13 @@ def get_number_of_atoms(self):
     def get_atomic_numbers(self):
         an = []
         for e in self.elem:
-            an.append(ptable[e].atomic_number)
+            if e[1:].isdigit(): # Index concatenated to one-character atomic symbol (e.g., C1)
```
jtkrogel: You should use the `is_element` function from `periodic_table.py`.

Something like:
```
iselem,e = is_element(e,symbol=True)
if not iselem:
    # some kind of error message
else:
    an.append(ptable[e].atomic_number)
#end if
```

PR #2259, opened 2020-01-30
Allow empty particleset in the input

markdewing:
To handle #2224, allow an empty ParticleSet to be created if it has the special name "empty".
Replaces #2234 

Example input:
```
<particleset name="empty" random="no">
    <group name="u" size="0">
      <parameter name="charge">0</parameter>
    </group>
  </particleset>
```

### DIFF
--- a/manual/particleset.tex
+++ b/manual/particleset.tex
@@ -77,6 +77,8 @@ The \ixml{particleset} blocks specify the particles in the QMC simulations: thei
 \begin{itemize}
 \item \ixml{name}/\ixml{id} \\
 Unique name for the particle set. Default is ``e" for electrons. ``i" or ``ion0" is typically used for ions. 
+For special cases where an empty particle set is needed, the special name ``empty" can be used to bypass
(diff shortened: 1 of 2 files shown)


prckent:
The group is always required?

markdewing:
I believe having one group is required.  I didn't try fixing the code so that no groups would work.

kryczko:
When I try to run now I get the following error:
```
Fatal Error. Aborting at ParticleSet::resetGroups() Failed. ParticleSet 'empty' has group 'u' containing 0 particles. Remove this group from input!
```

ye-luo:
After modifying your input by adding "empty" particleset and source="empty", I'm able to run with this branch. I'm afraid you were using a wrong branch. Once this PR gets merged, you can pull the develop. It should be easier for you.

Here is the modified input. You need to change txt to xml.
[qmcpack_in.txt](https://github.com/QMCPACK/qmcpack/files/4138823/qmcpack_in.txt)

PR #2260, opened 2020-01-30
Update HDF5 not found message.

markdewing:
The suggested HDF5 version is out of date.  Version 1.8 or newer is needed, or the code will not compile.  The HDF5_VERSION variable (from FindHDF5) does not get set with 1.6 - checking the version is not going to be useful.  The last 1.6 version was released in 2009, so people running into this version is unlikely to be an issue anyway.

No suggested version is given, since the version choice affects parallel IO performance, not correctness.

Prompted by #2251


prckent:
Good catch. Ideally the cmake FindHDF5 would allow a minimum version to be set https://cmake.org/cmake/help/latest/module/FindHDF5.html , but this is not yet possible.

PR #2262, opened 2020-01-30
update Golub

Paul-St-Young:
Trivial update to the campus cluster at University of Illinois Urbana-Champaign

PR #2264, opened 2020-01-31
update taub to golub in test_machines

Paul-St-Young:
address test failure from #2262


prckent:
Test this please

prckent:
OK to test

PR #2265, opened 2020-01-31
Add nexus ntest to jenkins_oxygen_cpu.sh

### LINKED ISSUE #2263: Nexus needs to be tested by CI
If this is duplication, please let me know.
#2262 indicates more automation is needed.

prckent:
Full precision real build only since nexus ntest tests are independent of QMCPACK.

Closes #2263


ye-luo:
I expected the initial CI to fail because #2264 fix comes after.
I have doubts the coverage of added nexus tests in CI.
@jtkrogel do you think we need to add more nexus tests than `ctest -R ntest`?

prckent:
My suggestion is to merge this and then to a test PR that deletes an important part of NEXUS.

PR #2266, opened 2020-01-31
Nexus: add plotting capability for non-local pseudopotentials

jtkrogel:
This PR implements 2D polar colormap plots to highlight the sign structure of the non-local operator component of semi-local pseudopotentials.


markdewing:
okay to test

PR #2267, opened 2020-01-31
Nexus: add interoperability support for old (Python 2) workflow data

jtkrogel:
After the move to Python 3, old Python 2 workflows could not be read or expanded upon with the updated code.  A number of projects are mid-stride (straddling Python 2 and Python 3 Nexus through the execution of the project) and users will benefit by being able to load archived workflows created during the Python 2 era.

This PR adds a simple fix based on how pickle files are loaded by trying to load with an encoding compatible with the Python 2 files if the original load fails.

### DIFF
--- a/nexus/bin/nxs-redo
+++ b/nexus/bin/nxs-redo
@@ -30,6 +30,16 @@ if help:
 #end if
 
 paths = dict()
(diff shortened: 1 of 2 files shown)


markdewing:
okay to test

PR #2268, opened 2020-01-31
update blue waters build script

Paul-St-Young:
Update Blue Waters configure and build script in response to recent software updates.
Both @kkly1995 and myself have been able to build the QMCPACK executable using this script.

This year marks the last hurrah for Blue Waters, so this script can be removed after Dec. 31 2020.


markdewing:
okay to test

prckent:
@Paul-St-Young We have no way to test this, so will take it at your word.

Paul-St-Young:
@prckent thanks for the heads up. Yes, we intend to run our own integration tests for every feature we use.

PR #2269, opened 2020-01-31
[HOLD] Fix multithread ctest qmc runs

PDoakORNL:
A number of things prevent threaded system i.e. running the qmcpack executable (historically integration tests) run via ctest from running on the expected number of cores.  This doesn't always happen but these work arounds make it actually possible to run the short tests in my development environment and I imagine many others.

1. openmpi by default binds one or two rank runs to core.  So all threads will run on one core. I have changed the default MPIEXEC_PREFLAGS to set --bind-to socket.  This will limit the threads to one socket for one test.  This works well for 16 thread jobs and 4x4 jobs on most of our testing targets.

2. While I'm not 100% sure on the cause setting ctest PROC_AFFINITY TRUE will force a threaded test to run on one thread.  I think this only happens when the openmp library is mismatched for ctest and the test executable.  This is even worse than item 1. Default is to have TRUE but it can now be set FALSE via QMC_CTEST_PROC_AFFINITY. 

3. openmp makes bad choices by default for distributing and binding the threads so I have added OMP_PLACES and OMP_PROC_BIND to the environment for test invocations.

If I've done this right it should be unnoticeable to anyone not interested in using the new QMC_CTEST_PROC_AFFINITY cmake cache option.


PDoakORNL:
Thats a good point Paul, looking at the FindMPI module more closely its pretty worthless. I also see I was wrong to make those various MPI variables cache they come in from FindMPI.

ye-luo:
1. OpenMPI is problematic. See our manual entry 3.9.5 how to workaround the slow testing issue if you are the admin of a node.
2. Can we improve it? My recent experience shows if you pass an mpi compiler wrappers to CMAKE_LANG_COMPILER, findMPI doesn't add anything to the flags and link line. I have not tried to see if it is able to return MPI library vendor info. If so, we can make better use of it.
3. OpenMP affinity control. I think your OpenMPI caused issue to OpenMP. All our unit test runs with 1 MPI + 1 thread by ctest. No problem there. For other tests, please use -j with ctest to allow more cores and parallel testing. The default affinity is spread and logic cores (hyperthreads) good for most people with 1 socket CPU and few cores. For advanced developers using powerful workstations. They can control the OpenMP affinity via environment variable. There is no need to hard code it in our cmake. For multi socket workstation, I found numactl is very helpful.

Here is what I do on my workstation (2 socket with 16 cores each and 2 hyperthreads).
If I don't care about performance. "ctest -R deter -j 32". If you are using power9 probably "OMP_PLACES=cores ctest -R deter -j 32" is better. The logic core layout is different between IBM and Intel.
if I need to run on socket 1 and avoid hyperthreading, "OMP_PLACES=cores numactl -N 1 ctest -R short -j 16"

Let me know if this is not enough for your case. Overall, it is better to reduce complication in our cmake if we can easily do it outside.

PDoakORNL:
> All our unit test runs with 1 MPI + 1 thread by ctest.
This isn't true, in order to properly unit test threading and mpi I have unit tests that must run on multiple threads and multiple ranks. 
Additionally the short tests aren't short if they are running on one core.

Your command seems to fix the Ctest affinity lockup so I'm going to close this. It's contrary to the ctest man page I'm looking at which says -j # is but this is how it should work. I'm going to close this in favor of hopefully a PR from you with better Cmake vs MPI handling over all.

PR #2275, opened 2020-02-04
[Do not merge] Test breaking change to NEXUS to test CI

prckent:
If nexus is tested in the CI, removing nexus.py should cause the CI to fail.


prckent:
=> The CI is not testing nexus

PR #2277, opened 2020-02-04
Catch wrong number of ranks to MPI unit tests.

PDoakORNL:
I also introduce a very simple abstraction Utilities/MPIExceptionWrapper.hpp to wrap unit test code using MPI exceptions thrown therein aren't lost and APP_ABORT is called.  This will help make sure we aren't chasing ghosts because someone has called the test directly with the "wrong" number of ranks or somehow overridden the rank count set by ADD_MPI_UNIT_TEST.

### DIFF
--- a/src/QMCDrivers/tests/test_WalkerControlMPI.cpp
+++ b/src/QMCDrivers/tests/test_WalkerControlMPI.cpp
@@ -2,7 +2,7 @@
 // This file is distributed under the University of Illinois/NCSA Open Source License.
 // See LICENSE file in top directory for details.
 //
-// Copyright (c) 2019 QMCPACK developers.
+// Copyright (c) 2020 QMCPACK developers.
 //
 // File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Laboratory
 //
@@ -20,7 +20,9 @@
 #include "QMCDrivers/MCPopulation.h"
 #include "QMCDrivers/QMCDriverInput.h"
 #include "QMCDrivers/tests/test_WalkerControlMPI.h"
+#include "Utilities/MPIExceptionWrapper.hpp"
 #include "Utilities/OutputManager.h"
+
 
 //#include "Concurrency/Info.hpp"
 //#include "Concurrency/UtilityFunctions.hpp"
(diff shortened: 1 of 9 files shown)


src/QMCDrivers/tests/test_WalkerControlMPI.cpp
```diff
@@ -192,6 +201,10 @@ TEST_CASE("MPI WalkerControl population swap walkers", "[drivers][walker_control
     std::vector<int> count_after{1, 2, 2};
     test.testPopulationDiff(count_before, count_after);
   }
+                   };
```
ye-luo: Needs formatting.

src/Utilities/tests/CMakeLists.txt
```diff
@@ -23,3 +23,16 @@ TARGET_LINK_LIBRARIES(${UTEST_EXE} qmcutil ${QMC_UTIL_LIBS} ${MPI_LIBRARY})
...
 #ADD_TEST(NAME ${UTEST_NAME} COMMAND "${QMCPACK_UNIT_TEST_DIR}/${UTEST_EXE}")
 ADD_UNIT_TEST(${UTEST_NAME} "${QMCPACK_UNIT_TEST_DIR}/${UTEST_EXE}")
+
+IF(HAVE_MPI)
+  SET(UTEST_EXE test_${SRC_DIR}_mpi)
+  SET(UTEST_NAME deterministic-unit_test_${SRC_DIR}_mpi)
+  #this is dependent on the directory creation and sym linking of earlier driver tests
+  SET(MPI_UTILITY_TEST_SRC test_mpi_exception_wrapper.cpp)
+  ADD_EXECUTABLE(${UTEST_EXE} ${MPI_UTILITY_TEST_SRC})
+  USE_FAKE_RNG(${UTEST_EXE})
```
ye-luo: I don't think we need this RNG stuff here.

PDoakORNL: Will delete that

src/Utilities/MPIExceptionWrapper.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include "Message/Communicate.h"
+
+namespace qmcplusplus
+{
+
+class MPIExceptionWrapper
+{
+public:
+  MPIExceptionWrapper() {};
```
ye-luo: In this case, I think there is no need of explicit empty constructor.

ye-luo: Please also delete the std::cout in `~MCPopulation` and probably just delete the whole empty destructor.

prckent:
@ye-luo Are you OK with these changes?

PR #2279, opened 2020-02-05
[WIP] Modernising cmake and clean up dependency

ye-luo:
I initially wanted to update our MPI handling but the change propagates to cleaning up all the internal and external dependency of our CMake.

0. Need cmake 3.10 which is released on Nov 20th 2017.

1. MPI part now always invoke FindMPI. There is no behaviour change. If compilers are actually MPI wrappers, it mostly does nothing. If compilers passed in are not MPI wrappers, it looks for MPI library and we issue a warning. It becomes possible to add library specific settings, I'm able to detect MVAPICH2 and issue warnings for #1760 
PS: MPIEXEC is deprecated and use MPIEXEC_EXECUTABLE instead.

2. Clean up internal target dependency:
```
  qmcutils -> qmcbase -> qmcwfc -> qmcham -> qmcdrivers -> qmc -> qmcpack.
                einspline -> ^      formic_utils-> ^          afqmc -> ^
                                                    qmcutils -> ^
```
Unit test should depend on the appropriate level.

3. We have QMC_UTIL_LIBS collecting all the libraries. It has bad dependency resolution.
a). Define Math::FFTW3, Math::BLAS_LAPACK, Math::scalar_vector_functions. Vendor ones are looked for first, if not define fallback. FindMKL, FindIBMMASS are all updated.
b). Define or use IO::HDF5, LibXml2::LibXml2, Boost::boost targets
c). Not changed so far: CUDA, mpi_wrapper, catch, boost_multi

4. Our config.h.cmake.in contains all the flags. It causes heavy dependence. When I build the targets for library, I tried to integrate the macro into COMPILE_DEFINITIONS. Only when a library is needed by a target, then the -D flags are added.

I think further cmake modernization effort will be needed. Some details I did here may not be optimal and require CMake expert advise.


markdewing:
This PR contains several logically separate changes.   These should be separate PR's.

ye-luo:
I can take make a PR for all the source code change to reduce the scale of this PR. Let me do this first.
Splitting cmake changes will not be easy because it may need to temporarily break the code compilation. @markdewing if you have any idea which part may potentially be separated from others, let me know.

prckent:
Very simply there is too much in this PR. It needs to be split now. It is not reviewable safely, and if anything goes wrong it will be painful to bisect the changes. Better to make changes incrementally.

Note: Don't force use of a newer cmake version unless absolutely necessary, and indicate why. If there is an important feature in the newer cmake, fine.

ye-luo:
Keep this PR alive at the moment to track piece by piece PR and code diffs.

ye-luo:
changes landed by #2286 #2289 #2295 #2296

PR #2286, opened 2020-02-06
Refine cmake dependence

ye-luo:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

In modern CMake, dependency between targets are critical. This PR reorganize the dependency of QMCPACK internal targets.
```
MISC -v
  qmcutils -> qmcbase -> qmcwfc -> qmcham -> qmcdrivers -> qmc -> qmcpack.
                einspline -> ^    formic_utils-> ^          afqmc -> ^
                                                qmcutils -> ^
```
MISC is a temporary solution, will be reduced to nothing after further cleaning.
It is still not perfect and will need to rearrange source files in the future.

## What type(s) of changes does this code introduce?
_Put an `x` in the boxes that apply._

- [ ] Bugfix
- [ ] New feature
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

### Does this introduce a breaking change?

- [ ] Yes
- [x] No

## What systems has this change been tested on?
My workstation w/ w/o MPI and CUDA
## Checklist

_Put an x in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- [x] this PR is up to date with current the current state of 'develop'
- [ ] code added or changed in the PR has been clang-formatted
- [ ] this PR adds tests to cover any new code, or to catch a bug that is being fixed
- [ ] documentation has been added (if appropriate)

Issue #1703, opened 2019-07-05
Memory alloction/deallocation is not safe with MVAPICH


ye-luo:
After investigating 'segmentation fault ' on Cooley at ALCF, I found that the [Mallocator](https://github.com/QMCPACK/qmcpack/blob/develop/src/simd/Mallocator.hpp) was causing the problem but I think this is not our fault but MVAPICH.

Mallocator uses aligned_alloc whenever available. If aligned_alloc is not available, posix_memalign is used. On Cooley, aligned_alloc is available from the OS.
In the libmpi.so of MVAPICH, it provides definition of both posix_memalign and free but not aligned_alloc. At linking, the aligned_alloc is picked up from OS and free is picked up from libmpi.so. The free was not able to delete the memory allocated by aligned_alloc safely and caused 'segmentation fault'.

Solution 1.
add `if defined(MVAPICH2_VERSION)`. This doesn't work when mpi.h is not included.

Solution 2.
Users add `-D QMC_EXTRA_LIBS=-lc` when configuring cmake, -libc is added explicitly before libraries supplied by the MPI compiler wrapper.

Solution 3.
Just don't use aligned_alloc and stick to posix_memalign. posix_memalign is linux only, aligned_alloc is a language standard.

I think it is the fault of MVAPICH overriding system(libc/glibc) malloc/posix_memalign/free but aligned_alloc was missed.
I verified that MPICH(Intel/Cray) and OpenMPI doesn't override these routines.

I tend to use solution 2 and push MVAPICH to fix the problem. Any thoughts?

PDoakORNL:
I think that is best, they should not break libc.

markdewing:
Is solution 2 guarded to only be used if using MVAPICH?

ye-luo:
We don't `add -D QMC_EXTRA_LIBS=-lc`. Users using MVAPICH needs to apply this workaround when they type cmake.

prckent:
I vote for none of the above. 

Solution: Detect mvapchi2 during our cmake configuration and put in the link time workaround.

harisubramoni:
Dear, Ye.

Thank you for bringing this to our attention. We appreciate it. So far, we have not received any issues about users wanting to use “aligned_alloc”. We will see how to handle it in our code and get back to you.

Some history about this feature is given below.

[...]

PR #2289, opened 2020-02-07
Adopt more FindMPI and requires CMake 3.10

ye-luo:
## Proposed changes
~~#2286 is needed for this PR.~~
This PR use FindMPI even if the CMAKE_C/CXX_COMPILER are already compiler wrappers.
1. FindMPI is significantly improved in cmake 3.9 and 3.10 deprecated MPIEXEC and prefers MPIEXEC_EXECUTABLE. So this PR requires cmake 3.10 which is released on Nov 20th 2017.
2. Users configure and compile as usual. The logic how FindMPI finds MPI is provided in https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindMPI.cmake#L101
Most user should hit case 2 and we issue warning in case 3 as before.
3. It becomes possible to add library specific settings, I'm able to detect MVAPICH2 and issue warnings for #1703. We can also add OpenMPI specific flags to fix the slow testing problem with threads in the future.
4. FindMPI defines MPI::MPI_CXX target (modern cmake style) I use it in our dependency tree.

_Put an `x` in the boxes that apply._

- [ ] Bugfix
- [ ] New feature
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [x] Documentation content changes
- [ ] Other (please describe):

- [ ] Yes
- [x] No

ALCF Theta Intel
NERSC Cori Intel
OLCF Summit GNU, XL, CUDA
My workstation, GNU no MPI, GNU finds MPI not via compiler wrapper.

## Checklist

_Put an x in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask.  This is

- [x] this PR is up to date with current the current state of 'develop'
- [ ] this PR adds tests to cover any new code, or to catch a bug that is being fixed
- [x] documentation has been added (if appropriate)


ye-luo:
@prckent updated the manual and added backward compatibility with warning.

ye-luo:
Test this please

CMakeLists.txt
```diff
@@ -540,30 +495,52 @@ IF (BUILD_AFQMC AND NOT APPLE)
...
+  IF(${MPI_CXX_LIBRARY_VERSION_STRING} MATCHES "MVAPICH2" AND NOT ${MPI_CXX_LIBRARY_VERSION_STRING} MATCHES "disable-registration-cache")
+    MESSAGE(FATAL_ERROR "MVAPICH2 with registration cache enabled breaks QMCPACK. "
+                        "Use a different MPI library or build MVAPICH2 with --disable-registration-cache configure option.")
   ENDIF()
 
-  IF ( NOT DEFINED MPIEXEC_PREFLAGS )
-    SET( MPIEXEC_PREFLAGS "" )
+  IF(NOT ${CMAKE_CXX_COMPILER} STREQUAL ${MPI_CXX_COMPILER})
+    SET(MPI_WARNING_LIST
+        "Building MPI version without using MPI compiler wrappers.\n"
```
markdewing: I think the preferred way to use CMake with MPI is to set the the compiler variables to the underlying compiler, and let FindMPI find all the include paths and libraries, etc.   Is there something in particular that we know will break QMCPACK (like not all the include or libraries paths are specified in a way that will pick up the MPI files?).     I don't think anything should be done on this PR, just pointing it out here.

prckent: The safe way to use MPI with cmake is to use the MPI compiler wrappers. This give a guaranteed outcome and the correct libraries will be linked. We recommend this route in the manual. Mileage may vary the other way, particularly on a "legacy" machine with multiple dusty MPI installations.

prckent:
I tested a subset of the nightlies using Intel MPI, OpenMPI, and no MPI. These all configured and ran indicating that the likely use cases for the FindMPI implementation are functional.

CMakeLists.txt
ye-luo: IMO, the preferred way is passing MPI compiler wrappers as compilers. FindMPI knows how to deal with this and keep the compile/link line clean. If the compilers passed in are not compiler wrappers, FindMPI goes to the folder containing MPIEXEC_EXECUTABLE and look for compiler wrappers of all the needed languages. Although it probably can find all the libraries and set proper flags and include directories based on querying MPIEXEC_EXECUTABLE options like -echo/-showme, the resulting compile and link lines get extremely long and error prone.

PR #2290, opened 2020-02-07
Nexus: Allow singlet/triplet excitation types

mcbennet:
## Proposed changes

This PR will allow the user to specify an excitation which is a pure singlet or pure triplet.
Presently, the PR is a work in progress but I thought I would share in order to gather thoughts
from @jtkrogel and others.

The current status is as follows. When the Nexus user specifies an excitation like so

```
qmc = generate_qmcpack(
    ...,
    excitation=['singlet','-35 +36'], 
    ...,
) 
```

Nexus will then generate a multideterminant WF that contains a single two-determinant CSF corresponding to the singlet or triplet excitation. As of now, only the orbital (state) index notation is allowed by the user when using this feature -- a future commit will allow for (ti,bi) notation. Furthermore, a block will need to be added in the case that either the singlet/triplet state is not possible with just a single-particle excitation and in the case that the desired singlet/triplet state can be constructed with only one determinant (in the case the user will be directed to use the 'up' or 'down' type)

- New feature

- No

## Checklist

- Yes. this PR is up to date with current the current state of 'develop'
- N/A. code added or changed in the PR has been clang-formatted
- No. this PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. documentation has been added (if appropriate)


prckent:
Given #2753, what are the plans + timeline for proceeding from WIP status? Is there a lot left to do?

mcbennet:
> Given #2753, what are the plans + timeline for proceeding from WIP status? Is there a lot left to do?

I recall that there were a few edge cases that needed to be addressed. I will block off a bit of time to wrap it up.
One week should be plenty.

mcbennet:
I am pretty close to getting this out of WIP state. I have tested this in the LCAO case and it works as expected.
I am moving on to PW tests. If it works there as well, I believe this will be ready for review.

mcbennet:
This PR is ready for review. PW case also works well. I imagine more testing is needed here -- perhaps as users play with this, the need for additional guards, etc., will reveal itself.

My tests involved the singlet/triplet excitation in N2. Both from a gamess route and pwscf route.

nexus/lib/qmcpack.py
```diff
@@ -239,7 +239,12 @@ def incorporate_result(self,result_name,result,sim):
                 qs  = input.simulation.qmcsystem
                 oldwfn = qs.wavefunction
                 newwfn = res.qmcsystem.wavefunction
+                if hasattr(oldwfn.determinantset,'multideterminant'):
+                    del newwfn.determinantset.slaterdeterminant
+                    newwfn.determinantset['multideterminant'] = oldwfn.determinantset['multideterminant']
```
jtkrogel, line 244: Use `.multideterminant` syntax rather than `['multideterminant']

nexus/lib/qmcpack.py
```diff
@@ -239,7 +239,12 @@ def incorporate_result(self,result_name,result,sim):
                 qs  = input.simulation.qmcsystem
                 oldwfn = qs.wavefunction
                 newwfn = res.qmcsystem.wavefunction
+                if hasattr(oldwfn.determinantset,'multideterminant'):
+                    del newwfn.determinantset.slaterdeterminant
+                    newwfn.determinantset['multideterminant'] = oldwfn.determinantset['multideterminant'] 
+                    newwfn.determinantset['sposets'] = oldwfn.determinantset['sposets']
```
jtkrogel, line 245: Similar for sposet.

nexus/lib/qmcpack_input.py
```diff
@@ -4233,6 +4233,8 @@ def bundle_inputs(self,quantity,values,input):
         for value in values:
             inp = input.copy()
             qhost = inp.get_host(quantity)                               
+            print(inp)
```
jtkrogel, line 4236: Remove print statements

nexus/lib/qmcpack_input.py
```diff
@@ -4811,15 +4813,121 @@ def generate_determinantset_old(type           = 'bspline',
...
-            det = dset.get('updet')
+            sdet = dset.get('updet')
         elif spin_channel=='down':
-            det = dset.get('downdet')
+            sdet = dset.get('downdet')
+        elif spin_channel=='singlet' or spin_channel=='triplet':
+
+            # Is multi-det WF appropriate?
+            if elns.down_electron.count != elns.up_electron.count:
+                QmcpackInput.class_error('The \'singlet\' and \'triplet\' excitation types assume number of up and down electrons is the same. Otherwise, one should use \'up\' or \'down\' types.\nFor your system: Nup={} and Ndown={}'.format(elns.up_electron.count,elns.down_electron.count))
```
jtkrogel, line 4826: Doesn't this requirement depend on the reference state (e.g. if it is already singlet-like or triplet-like)?

jtkrogel (changes requested):

nexus/lib/qmcpack.py
mcbennet: Resolved.

nexus/lib/qmcpack.py
mcbennet: Resolved.

nexus/lib/qmcpack_input.py
mcbennet: Resolved.

nexus/lib/qmcpack_input.py
mcbennet: Presently, the code assumes that the ground state is a closed shell singlet and the user wants to create a pure singlet or pure triplet excited state

jtkrogel: OK for now, can expand in the future.  Change the message to "types currently assume".

nexus/lib/qmcpack_input.py
```diff
@@ -4841,7 +4948,7 @@ def generate_determinantset_old(type           = 'bspline',
                 QmcpackInput.class_error('excitation with vb-cb band format works only with special k-points')
             #end if
             
-            vb = int(det.size / abs(linalg.det(tilematrix))) -1  # Separate for each spin channel
+            vb = int(sdet.size / abs(linalg.sdet(tilematrix))) -1  # Separate for each spin channel
```
jtkrogel, line 4951: This looks like overuse of find/replace.  `linalg.sdet` needs to remain `linalg.det` to calculate the determinant of the tiling matrix.

mcbennet: Resolved.

mcbennet: Good catch

nexus/lib/qmcpack_input.py
mcbennet: Resolved.

nexus/lib/qmcpack_input.py
jtkrogel: I'm not seeing the change here.  Committed and pushed?

mcbennet: My change was pushed in commit 1153c8d with message "Fix sdet typo".
Though I kept `sdet.size`. I think only the `linalg.sdet` should be fixed, right?

jtkrogel (approved):
LGTM

nexus/lib/qmcpack_input.py
jtkrogel: Yes

ye-luo:
Test this please

PR #2292, opened 2020-02-08
Bug fix for molecular linear dependency issue #2291

fdmalone:
## Proposed changes

Fix for AO->MO transformation matrix which has lost MOs due to linear dependencies. Addresses  #2291.
I added a test for this and in the process needed the ability to recompute the HF energy. The HF energy can be printed by the used if the verbosity is set to high (-vvv).
  
## What type(s) of changes does this code introduce?
_Put an `x` in the boxes that apply._

- [X] Bugfix
- [ ] New feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

### Does this introduce a breaking change?

- [ ] Yes
- [X] No

## What systems has this change been tested on?
Unit tests and problematic case raised in issue. Tested on macbook pro python3.7.4, pyscf v1.7.0

## Checklist

- [X] this PR is up to date with current the current state of 'develop'
- [ ] code added or changed in the PR has been clang-formatted
- [X] this PR adds tests to cover any new code, or to catch a bug that is being fixed
- [X] documentation has been added (if appropriate)

PR #2294, opened 2020-02-09
Fix linear dependency issue 

fdmalone:
## Proposed changes
Follow up for #2292 which deals with case when overlap is singular and we want to work in orthogonalised AO basis. Replaces inverse with pseudo inverse in this case as transformation matrix is rectangular. 
Also tidied some help message in converter.

- [X] Bugfix

- [ ] Yes
- [X] No

Laptop

## Checklist

- [X] this PR is up to date with current the current state of 'develop'
- [X] this PR adds tests to cover any new code, or to catch a bug that is being fixed
- [ ] documentation has been added (if appropriate)


ye-luo:
Okay to test

PR #2302, opened 2020-02-10
Removed the use of const h5_observable_type

brtnfld:
## Proposed changes

(1) the datatype should have been hid_t, not hsize_t

(2) In C++, for global objects, the constructor is called before main() is called. This results in H5open() getting called before MPI_init, which is not good practice in HDF5 usage. This behavior is a result of H5T_NATIVE_DOUBLE not being a const parameter in HDF5, but is instead determined dynamically. 

no. 2 causes failure in the DAOS VOL connector call daos_cont_global2local. This failure is a result of the DAOS vol connector being initialized in serial and then later used in parallel. This case might eventually be handled by the HDF5 DAOS VOL connector.

Changed to pass H5T_NATIVE_DOUBLE to H5Dwrite instead.

_Put an `x` in the boxes that apply._

- [ ] Bugfix
- [ ] New feature
- [ ] Code style update (formatting, renaming)
- [ x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

- [ ] Yes
- [ x] No

DAOS cluster

## Checklist

_Put an x in the boxes that apply. You can also fill these out after creating
the PR. If you're unsure about any of them, don't hesitate to ask.  This is

- [ x] this PR is up to date with current the current state of 'develop'
- [ ] this PR adds tests to cover any new code, or to catch a bug that is being fixed
- [ ] documentation has been added (if appropriate)


ye-luo:
Approve this PR. Merging will be resumed after the release tomorrow.

ye-luo:
Okay to test

PR #2312, opened 2020-02-14
Spin-Orbit ECP Component Evaluation

### LINKED ISSUE #1770: Spin-Orbit Commit Plan

camelto2:
## Proposed changes

This PR implements the evaluation of the Spin-orbit ECP components, complete with a unit test and the code to generate the reference data. Additionally, this lays the framework for the full SOECPotential, but it is not currently implemented and will come in a later PR to keep this one smaller and easier to review. In the next PR, the SOECPotential will be implemented, and will then finish up #1770. 

With this PR, we can close #1815 by removing the Zn.ccECP-SO.xml file entirely. Since the SO terms were not in the correct format, I replaced the SOECP reading unit test to use so_ecp_test.xml  added here which is in the correct format.

Additionally, the SOECP radial potential was not splined in the form V(r), but was kept as r*V, whereas in the NonLocal radial potential was just V(r). This fixed the spline to be of the form V(r) to be consistent. 

Documentation will be added once the entire workflow is sorted out to explain how to do a calculation. 

- New feature

- No

my laptop (Macbook Pro  Mojave OS)

## Checklist


rcclay:
OK to test.

camelto2:
@ye-luo, this does handle both cases. All you need to do is add a hamiltonian component with an xml file. In that xml file, you will have your normal nonlocal terms which get registered to ECPotential as normal, and if you have spin-orbit terms they will be created. So the nonlocal terms will still be used an evaluated as normal. Only if the xml file includes spin-orbit terms, indicated by <vps_so> will anything happen with this code.

@prckent It is definitely possible to have ECPotential handle all of the work I was planning for SOECPotential to have, which would minimize any divergence. In that case there would only be the ECPotential, and it would manage both the nonlocal and spin-orbit components

The initial reason for creating a separate SOECPotential that is would be similar to ECPotential was because we may want cases where we do not treat the SO terms as physical in the hamiltonian (only want an expectation value in scalar.dat, but not inside the local energy). Having a separate Hamiltonian component seemed to be the path of least resistance since if I included the spin-orbit components inside ECPotential, then without a workaround they would always be included as physical in the Hamiltonian. The case I am thinking of is using normal wave functions but doing a perturbative estimate of the SO interaction, similar to some of the recent work by Lucas Wagner and company.  I'll have a look at how to accomplish this only inside of ECPotential

For this PR though, the important part is the SOECPComponent which handles the evaluation correctly. If a change from SOECPotential into ECPotential, that can be handled in the next PR

PR #2322, opened 2020-02-27
New setup for oxygen CI

PDoakORNL:

## Proposed changes

New pipeline based CI config for jenkins. This will prepare us to run CI on multiple machines from a common server.
The amount of repeated elements in the build scripts are greatly reduced and a better summary of the CI jobs can be seen on the server.  (Getting at least maintainers access to the server is for a following PR). Further simplification will follow but we should get the CI back up and running.

- Build related changes

- No

## Checklist


prckent:
All these files will need ornl (site name) and/or oxygen (machine name) added to their file names. The test automation directory covers many sites. Jenkinsfile_oxygen -> Jenkinsfile_ornl_oxygen  etc.

PDoakORNL:
Ok until we have a firewall exception for oxygen we will not be able to use the Github Authentication so I have turned on the user signup for now.

ye-luo:
We had three cases in the past. oxygenCI/cuda with -DENABLE_CUDA=1 case is missing.

prckent:
I think now would be a good time to add some documentation about the GPU flags.

QMC_CUDA is the legacy real space CUDA support.
ENABLE_CUDA is ?
ENABLE_OFFLOAD is ?

PDoakORNL:
I will add that case @ye-luo.

PDoakORNL:
So I added the enable_cuda CI variant and greatly reduced the amount of duplication in the CI code. More files can be removed once this is merged. There is now a private jenkins shared pipline library at QMCPACK/qmcpack_shared_jenkins.  The common pipeline code and the list of default emails lives in that library.

PDoakORNL:
Also you now get an email if your build is Aborted by someone via Jenkins. These do not go to the default addressees.

ye-luo:
@PDoakORNL 
Q1. Could you list what flag combinations are used in each CI pipline?
In the old setting, we had
CPU/GPU/CUDA each has 4 builds: Real + AoS + FP, Real + SoA + MP, Cplx + AoS + FP, Cplx + AoS + FP.
Q2. What version is GCC is used now? We were using GCC 6.

prckent:
I would like to get something merged and in use that we can learn from and iteratively improve. Thus this version or something very close to it should be merged. 

@PDoakORNL Did you write a matching install script to the spack versions.sh? I can attempt to use those in the nightlies with time, and we can encourage others to use them. 

Regarding software versions, I think the ideal would be that the CI uses exactly the same versions as used for one/a few variants as used in the nightlies. This was not the case in the past, and should be a topic for the future, not this PR.

Note that python and packages also need to be installed to test nexus. This can be done later. 
```
echo --- Python setup for NEXUS
spack install py-numpy@1.18.0
spack install py-h5py^py-numpy@1.18.0
spack install py-pandas@0.25.1^py-numpy@1.18.0
spack install py-scipy@1.4.1^py-numpy@1.18.0
spack activate py-numpy@1.18.0
spack activate py-h5py^py-numpy@1.18.0
spack activate py-pandas@0.25.1^py-numpy@1.18.0
spack activate py-scipy@1.4.1^py-numpy@1.18.0
```
Then 'spack load python'

PDoakORNL:
All builds are do a product of ‘number type = [real, complex] and precision = [ full, mixed ]  
Which feed the nspace and precision shell variables.

They should be all SOA we’re never going to have another release with AoS as I understand

For cpu
QMC_IMMUTABLE_FLAGS="-DBUILD_AFQMC=1"

For gpu
QMC_IMMUTABLE_FLAGS="-DQMC_CUDA=1"

For enable_cuda
QMC_IMMUTABLE_FLAGS="-DENABLE_CUDA=1"

cmake ${QMCNSPACE_FLAG} ${QMCPRECISION_FLAG} -DENABLE_SOA=1 -DCMAKE_C_COMPILER="mpicc" -DCMAKE_CXX_COMPILER="mpicxx" ${QMC_IMMUTABLE_FLAGS} -DQMC_NO_SLOW_CUSTOM_TESTING_COMMANDS=1 ../.. 2>&1 | tee cmake.out

PDoakORNL:
ornl_oxygen_setup.sh Should match 
spack_supported_package_versions.sh
Everything should just source spack_supported_package_versions.sh
But in the interest of getting this merged I didn’t want to also edit all of the nightly scripts many of which I can’t test
I think the nexus tests should be a shared pipeline step included in only 1 or a few pipelines. That’s a separate PR

prckent:
@PDoakORNL Good idea to only test NEXUS in a very limited number of pipelines.

PR #2324, opened 2020-03-02
Extract sample stack

markdewing:
## Proposed changes

Extract the storage of samples to a separate class SampleStack.  These samples are used later in wavefunction parameter optimization.

@PDoakORNL suggest that the samples should be stored in SoA form for better memory performance.  The MCSample class should be isolated behind the SampleStack interface, so it should be straightforward to change the underlying storage.

- No

laptop - also CUDA build

## Checklist


markdewing:
For the name 'Stack' - I just copied the existing term without thinking about it.   I thought of in the colloquial sense of  'a pile' rather than the CS container.

Maybe the name should not contain 'stack'?  SampleStorage?  FreezeDriedSamples? ConfigurationSamples?

prckent:
I don't have a better suggestion at the moment. We can see what this grows into. e.g. Depending on how the data for non-local reevaluation during optimization is stored.

src/Particle/SampleStack.h
```diff
@@ -0,0 +1,91 @@
...
+
+  int getMaxSamples() const { return MaxSamples; }
+
+  void setMaxSamples(int max_samples) { MaxSamples = max_samples; }
+
+  bool empty() const { return sample_stack_.empty(); }
+
+  void putSample(unsigned int i, const Walker_t& w);
+
+  void getSample(unsigned int i, Walker_t& w) const;
```
PDoakORNL: The semantics of this are bad to me why break encapsulation why void with output argument.  This is going to make the refactor I have planned for walker more painful. 
Rather than hide the MCSample type this would make much more sense to me

```C++
MCSample SampleVector::getSample(size_t sample_index)
{
  return MCSample(sample_statck_[sample_index];
}

Walker walker;
walker.loadSample(sample_stack.getSample());
```

src/Particle/SampleStack.h
```diff
@@ -0,0 +1,91 @@
...
+namespace qmcplusplus
+{
+class MCWalkerConfiguration;
+class HDFWalkerOutput;
+class MCSample;
+
+class SampleStack
+{
+public:
+  typedef ParticleSet::Walker_t Walker_t;
```
PDoakORNL: use using

src/Particle/SampleStack.h
```diff
@@ -0,0 +1,91 @@
...
+  void getSample(unsigned int i, Walker_t& w) const;
+
+  //@{save/load/clear function for optimization
+  inline int getNumSamples() const { return CurSampleCount; }
+  ///set the number of max samples
+  void setNumSamples(int n);
+  ///save the position of current walkers to SampleStack
+  //void saveEnsemble();
+  ///save the position of current walkers
+  void saveEnsemble(walker_iterator first, walker_iterator last);
```
PDoakORNL: Again why create a dependence on the walker type for this.  Perhaps an iterator over std::vector<MCSample&>.

src/Particle/SampleStack.h
```diff
@@ -0,0 +1,91 @@
...
+   * @param np number of processors
+   * @return true with non-zero samples
+   */
+  bool dumpEnsemble(std::vector<MCWalkerConfiguration*>& others, HDFWalkerOutput* out, int np, int nBlock);
+  ///clear the ensemble
+  void clearEnsemble();
+  //@}
+
+private:
+  int TotalNum;
```
PDoakORNL: total_num_ etc.

src/Particle/SampleStack.h
markdewing: Some piece of code has to know about both Walker_t and MCSample.   I can change it so that translation code is not here, and expose MCSample outside the class.

src/Particle/SampleStack.h
markdewing: That requires writing more code to convert from Walker_t to MCSample.   Which is fine, but it makes the PR larger.  I am trying to make this extraction as minimally intrusive as possible, especially on code outside of MCWalkerConfiguration.

markdewing:
I've updated some of the code to reflect coding standards.   Can I change the MCSample interface in a separate PR?

src/Particle/SampleStack.h
```diff
@@ -0,0 +1,91 @@
...
+  int MaxSamples;
+  int CurSampleCount;
+
+  std::vector<MCSample*> sample_stack_;
```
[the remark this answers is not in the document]
markdewing: Changed to `sample_vector_`
(`samples_` was another choice, but that's used in MCWalkerConfiguration.h for the SampleStack.)

src/Particle/SampleStack.cpp
```diff
@@ -0,0 +1,144 @@
...
+  clearEnsemble();
+  MaxSamples = n;
+  //do not add anything
+  if (n == 0)
+    return;
+  sample_stack_.reserve(n);
+  int nadd = n - sample_stack_.size();
+  while (nadd > 0)
+  {
+    sample_stack_.push_back(new MCSample(TotalNum));
```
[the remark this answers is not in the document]
markdewing: My goal is that the current PR is pure refactor to extract the sample handling code.  As such it should leave the memory behavior unchanged (for better or worse).   I can make this change in the future, but I would like it to be a separate PR.  Same with adding the destructor (which will be unnecessary if this is changed)

PDoakORNL:
```shell
/dev/shm/jenkins/enable_cuda/src/Particle/SampleStack.h: In member function 'int qmcplusplus::SampleStack::setTotalNum(int)':

/dev/shm/jenkins/enable_cuda/src/Particle/SampleStack.h:41:60: warning: no return statement in function returning non-void [-Wreturn-type]

   int setTotalNum(int total_num) { total_num_ = total_num; }
```

PR #2325, opened 2020-03-02
SOECPotential implementation

camelto2:
## Proposed changes

This PR implements the evaluation of the Spin-Orbit Potential, that was introduced in  #2312 . Just like the ECPotential Builder manages and creates the Local and NonLocalPotentials, it now creates another potential for the total Spin-Orbit contributions if it notices that any of the pseudo xml files include SO terms. 

An input flag is also given to determine whether the spin-orbit terms should be physical or not, i.e. <pairpot type="pseudo" ... physicalSO="yes/no">. This would allow for perturbative style treatment similar to [arXiv:1809.04133](https://arxiv.org/abs/1809.04133), where the SO contribution is not included in the total energy. If physicalSO="yes", it will be included in the local energy. 

- New feature

- No

my workstation

## Checklist


ye-luo:
Okay to test

prckent:
Test this please

PR #2328, opened 2020-03-03
Detect OpenMPI and add --bind-to none by default.

### LINKED ISSUE #2326: Unable to specify thread binding via MPIEXEC_PREFLAGS due to excess CMake quoting
**Describe the bug**

OpenMPI will by default bind threads of MPI programs with a single task to a single core. This is always the wrong choice for QMCPACK (or any properly threaded application) and will always slow the tests. e.g.

```
export OMP_NUM_THREADS=128
mpirun -n 1 qmcpack input.xml
```
will not do what we want, despite being a reasonable invocation on, say, an AMD Rome processor with lots of cores.

Users have to specify "--bind-to none" for the desired behavior:
```
mpirun --bind-to none -n 1 input.xml
```

For our tests to use this flag, ideally we would use the MPIEXEC_PREFLAGS='--bind-to none' cmake setting, but when when invoked cmake put excess quotes, which confuses mpirun's parser.

```
mpirun "-n" "1" "--bind-to none" "input.xml"
```

The upshot of this bug ("feature") is that many of our tests take 16x longer to run than they should and we waste test hardware. 

Does anyone have a workaround to fix the quoting? This seems to be a known cmake problem but I could not find a workaround. I am specifically looking for a user invocable fix to the thread binding situation and not an edit to an OpenMPI configuration option. e.g. For the case where a user installed OpenMPI via spack or has no root access.

**To Reproduce**

Latest release or develop. Build and run the tests. Check top to see process mapping.

**Expected behavior**

Use the expected number of cores.

When I use --bind-to none by hand, I was able to get 12800% utilization on nitrogen.ornl.gov.

**System:**

OpenMPI 4.0.2, typical nightly build configurations. oxygen or nitrogen.

**Additional context**

ye-luo:
## Proposed changes
Fix #2326
When OpenMPI is detected, add "--bind-to none" if MPIEXEC_PREFLAGS is empty.
If MPIEXEC_PREFLAGS is not empty, issue an warning saying --bind-to is not specified.

- Build related changes

- No

Workstation bora.

## Checklist


CMakeLists.txt
```diff
@@ -500,6 +500,19 @@ IF(QMC_MPI)
                         "Use a different MPI library or build MVAPICH2 with --disable-registration-cache configure option.")
   ENDIF()
 
+  IF(${MPI_CXX_LIBRARY_VERSION_STRING} MATCHES "Open MPI")
+    IF(NOT MPIEXEC_PREFLAGS)
+      SET(MPIEXEC_PREFLAGS "--bind-to;none")
+    ELSE(NOT MPIEXEC_PREFLAGS)
+      IF(NOT "${MPIEXEC_PREFLAGS}" MATCHES "--bind-to")
+        MESSAGE(WARNING "Default Open MPI thread affinity may be bad for QMCPACK. "
```
markdewing: Should be more precise about what 'bad' means.  It means all threads would be bound to one core, and give very poor QMCPACK performance.

prckent:
LGTM and previous comments were addressed.

PR #2332, opened 2020-03-06
Nitrogen nightly scripts

prckent:
## Proposed changes

Nightly scripts and spack install for nitrogen (dual AMD 7542 2.9GHz 32 core running RHEL8). Uses amdblis for BLAS. Does not use amdlibflame or amdfftw since these are not yet in spack. QE is built for gcc compilers.

These scripts are an initial starting point and in future versions will be modified to use spack_supported_software_versions, to add LLVM 10, use more optimized libraries etc.

(Openblas was found to be unreliable and to crash in the QMCPACK performance tests due to memory allocation and thread issues internal to Openblas. )

- Other (please describe): Integration tests

- No

Nitrogen

## Checklist

### DIFF
--- a/ornl_nitrogen_setup.sh
+++ b/ornl_nitrogen_setup.sh
@@ -1,0 +1,155 @@
+#!/bin/bash
+
+#
+# Installs compilers & libraries for QMCPACK testing via SPACK 
+#
+#          *** DESTROYS EXISTING INSTALLATION ***
+#
+#
+
+echo --- START setup script `date`
+
+rm -r -f $HOME/apps/spack $HOME/.spack
+
+mkdir $HOME/.spack
+
+cat >$HOME/.spack/config.yaml<<EOF
+config:
+
+  build_stage:
+    - /scratch/$USER/spack_build_stage
+    - /home/$USER/apps/spack/var/spack/stage
+
+  build_jobs: 128
+
+EOF
+# Use flash /scratch for builds
+rm -r -f /scratch/$USER/spack_build_stage
+mkdir /scratch/$USER/spack_build_stage
+
+cd $HOME/apps
+git clone https://github.com/spack/spack.git
+
+cd $HOME/apps/spack
+# For reproducibility, use a specific version of Spack
+# Use tagged releases https://github.com/spack/spack/releases
+# git checkout v0.13.3
+#git checkout b9dc263801ab8b9ce46e83adec8002c299fe2e44
+#Author: Justin S <3630356+codeandkey@users.noreply.github.com>
+#Date:   Fri Jan 3 15:52:59 2020 -0600
+#
+#    py-intervaltree: new package at 3.0.2 (#14277)
+#git checkout v0.13.4
+module() { eval `/usr/bin/modulecmd bash $*`; }
+
+cd bin
+./spack bootstrap
+
+export SPACK_ROOT=$HOME/apps/spack
+. $SPACK_ROOT/share/spack/setup-env.sh
+
+echo --- Spack list
+spack find
+echo --- Spack compilers
+spack compilers
+echo --- Spack compiler add
+spack compiler find
+echo --- Spack compilers
+spack compilers
+echo --- Modules list
+module list
+echo --- End listings
+
+#
+# Versions should be consistent with setup script
+#
+
+# GCC
+# Dates at https://gcc.gnu.org/releases.html
+gcc_vnew=9.2.0 # 2019-08-12
+
+#Zen2 optimziations are only in gcc 9.1+, with improved scheduling in 9.2+
+#For now, only use newer compilers
+
+# For CUDA toolkit compatibility
+gcc_vcuda=8.3.0 #  2019-02-22
+
+# LLVM 
+# Dates at http://releases.llvm.org/
+llvm_vnew=9.0.0 # 2019-09-19
+#Zen2 scheduleing optimization are only ing LLVM 10+
+
+# HDF5
+hdf5_vnew=1.10.5 # Releeased 2019-02-28
+hdf5_vold=1.8.19 # Released 2017-06-16
+
+# CMake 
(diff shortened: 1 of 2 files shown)


ornl_nitrogen_setup.sh
```diff
@@ -0,0 +1,155 @@
...
+spack install netlib-lapack%gcc@${gcc_vnew} # Use netlib since it will work and we have wasted enough time already
+
+#spack install openblas%gcc@${gcc_vnew} # Will crash in Performance tests
+
+echo --- Convenience
+spack install git
+echo --- Python setup for NEXUS `date`
+
+spack install py-numpy^openblas%gcc@${gcc_vnew}
+spack install py-mpi4py^openmpi@${ompi_vnew}%gcc@${gcc_vnew} fabrics=auto
```
PDoakORNL: This could be done for the oxygen script as well.  Using that I ended up with with an additional openmpi build triggered by that script since spack defaults to openmpi@3.1.5.

prckent: Yes, that would be part of the cleanup. These scripts for nitrogen and more specific than those for oxygen. We might be able to use the same script with only a few variations based on the hostname.

Unfortunately there are spack bugs to work around. e.g. py-numpy is built with openblas despite my comments about it being unreliable in QMCPACK because 
```
spack spec py-numpy^amdblis
Input spec
--------------------------------
py-numpy
    ^amdblis

Concretized
--------------------------------
==> Error: Package py-numpy does not depend on amdblis
```
whereas
```
spack spec py-numpy
Input spec
--------------------------------
py-numpy

Concretized
--------------------------------
py-numpy@1.18.1%gcc@9.2.0+blas+lapack arch=linux-rhel8-zen2
    ^openblas@0.3.9%gcc@9.2.0~consistentFPCSR~ilp64+pic+shared threads=none arch=linux-rhel8-zen2
    ^py-cython@0.29.14%gcc@9.2.0 arch=linux-rhel8-zen2
        ^py-setuptools@41.4.0%gcc@9.2.0 arch=linux-rhel8-zen2
            ^python@3.7.6%gcc@9.2.0+bz2+ctypes+dbm~debug+libxml2+lzma~nis~optimizations+pic+pyexpat+pythoncmd+readline+shared+sqlite3+ssl~tix~tkinter~ucs4~uuid+zlib arch=linux-rhel8-zen2
                ^bzip2@1.0.8%gcc@9.2.0+shared arch=linux-rhel8-zen2
                    ^diffutils@3.7%gcc@9.2.0 arch=linux-rhel8-zen2
                        ^libiconv@1.16%gcc@9.2.0 arch=linux-rhel8-zen2
                ^expat@2.2.9%gcc@9.2.0+libbsd arch=linux-rhel8-zen2
                    ^libbsd@0.10.0%gcc@9.2.0 arch=linux-rhel8-zen2
                ^gdbm@1.18.1%gcc@9.2.0 arch=linux-rhel8-zen2
                    ^readline@8.0%gcc@9.2.0 arch=linux-rhel8-zen2
                        ^ncurses@6.1%gcc@9.2.0~symlinks+termlib arch=linux-rhel8-zen2
                            ^pkgconf@1.6.3%gcc@9.2.0 arch=linux-rhel8-zen2
                ^gettext@0.20.1%gcc@9.2.0+bzip2+curses+git~libunistring+libxml2+tar+xz arch=linux-rhel8-zen2
                    ^libxml2@2.9.9%gcc@9.2.0~python arch=linux-rhel8-zen2
                        ^xz@5.2.4%gcc@9.2.0 arch=linux-rhel8-zen2
                        ^zlib@1.2.11%gcc@9.2.0+optimize+pic+shared arch=linux-rhel8-zen2
                    ^tar@1.32%gcc@9.2.0 arch=linux-rhel8-zen2
                ^libffi@3.2.1%gcc@9.2.0 arch=linux-rhel8-zen2
                ^openssl@1.1.1d%gcc@9.2.0+systemcerts arch=linux-rhel8-zen2
                    ^perl@5.30.1%gcc@9.2.0+cpanm+shared+threads arch=linux-rhel8-zen2
                ^sqlite@3.30.1%gcc@9.2.0~column_metadata+fts~functions~rtree arch=linux-rhel8-zen2
```
i.e. spack py-numpy or amdblis are broken and for now we get an unwanted openblas installation.

An additional hdf5 library is also created somewhere in the python dependencies.

Sadly spack does not try very hard to minimize the amount of freshly installed software: without all the ^openmpi@4 type entries, spack would merrily install openmpi v3.x even when we have a valid instance of v4 installed.

I am hopeful that some of these bugs go away during the next couple of weeks while we wait for the final LLVM 10.0 release and spack support. There has been considerable improvement in the last few weeks.

ye-luo:
it is quite strange that OpenBLAS doesn't work for you. On my machine, OpenBLAS built with USE_OPENMP=1 TARGET=ZEN  works well with QMCPACK.

prckent:
OpenBLAS is probably failing due to some variation of https://github.com/xianyi/OpenBLAS/issues/539. I am not going to spend any more time on it, since blis is AMD's preferred linear algebra library. Using https://github.com/amd/aocl-spack looks to be the best low maintenance in the medium term, provided it works with the latest spack. The aocl-spack overlays blis and libflame, and should solve the py-numpy dependency issue.

ye-luo:
> OpenBLAS is probably failing due to some variation of [xianyi/OpenBLAS#539](https://github.com/xianyi/OpenBLAS/issues/539). I am not going to spend any more time on it, since blis is AMD's preferred linear algebra library. Using https://github.com/amd/aocl-spack looks to be the best low maintenance route in the medium term, provided it works with the latest spack. The aocl-spack overlays blis and libflame, and should solve the py-numpy dependency issue.

My guess is that the default threading option of OpenBLAS picked by spack is not compatible with OpenMP. AMD finally did something right shipping all the libraries together hopefully it works out of box.

PR #2333, opened 2020-03-06
Change the default walkers per rank to one

markdewing:

## Proposed changes
If `walkers_per_rank_` defaults to zero, a vmc_batch section without 'walkers' specified will divide by zero in `calc_default_local_walkers`. The `num_crowds_` variable gets set to zero, and there is a division by `num_crowds_`.

This also makes the code behave consistent with the manual (minimum of one walker per rank)

- Bugfix

- No

laptop

## Checklist

### DIFF
--- a/src/QMCDrivers/QMCDriverInput.h
+++ b/src/QMCDrivers/QMCDriverInput.h
@@ -78,7 +78,7 @@ protected:
   input::PeriodStride config_dump_period_;
   IndexType starting_step_              = 0;
   IndexType num_crowds_                 = 0;
-  IndexType walkers_per_rank_           = 0;
+  IndexType walkers_per_rank_           = 1;
   IndexType requested_samples_          = 0;
   IndexType sub_steps_                  = 1;
   // max unecessary in this context


PDoakORNL:
Default walkers per crowd for vmc should be ~ total_walkers / num_threads.  For DMC it should be ~ target_walkers / num_threads. If the value isn't set it is 0 and determined programmatically.  If the input is invalid it can't be and should be caught and an appropriate error issued.

If we default to the old configuration that is all that will ever be tested and mostly what will be used.

src/QMCDrivers/QMCDriverInput.h
```diff
@@ -78,7 +78,7 @@ class QMCDriverInput
   input::PeriodStride config_dump_period_;
   IndexType starting_step_              = 0;
   IndexType num_crowds_                 = 0;
-  IndexType walkers_per_rank_           = 0;
```
PDoakORNL: This must be set to an invalid input option since it should come only from input or calculation.  We have talked about dropping it from the input spec(i.e. just set total_walkers = your_desired_walkers_per_rank * ranks).  Since the repored max_num_threads can be complete nonsense depending on the threading framework, including on Summit if the hardware threads are visible to openmp, I don't want to automatically set total_walkers = max_threads. Input validation should be added instead.

I think the ambiguous 'walkers' should have no place in the input spec. By all means we can start a section of the manual for the "New" input. But I favor breaking immediately with the old input spec when its ambiguous or nested in such a way to create XML node ownership ambiguity and the old defaults should be eliminated when they don't support the batched concurrency we're working toward which will most performant with the most common size of system.

markdewing:
If I read this correctly, 'walkers' needs to be a required input option for VMC.

prckent:
Commentary:

For the batched drivers -- which approximately no one is using yet -- I suggest hard abort on anything nonsensical and the smallest amount of code+defaults possible to keep them running. I don't have a strong opinion on how to do this other than that very little time be spent doing so. The default of 1 seems simple. We should also reserve the right to change what we do in future for these developmental drivers. 

The motivation behind these comments is that we should not build up fresh input logic for the "old" input parameters since these are going to be abandoned soon. We should put effort into new parameters that e.g. specify the normalization. The historical inconsistencies are a source of confusion for people new to the code and QMC. We will hard abort if any new parameters are given in combination with any "old" parameters so that we have a chance to get the logic correct + consistent. Of course, we need to find some time to actually do this (+update all the converters consistently), but that is the goal.

PDoakORNL:
@markdewing 
I'm guilty of remembering some things that are only in my population control branch as being already merged into develop.  In that branch I revamped handling of the walker count to follow the behavior of the current driver when total_walkers or walkers is specified in DMC and target_walker or walkers is specified in DMC.  I'm making a PR of just that part of my recent mpi population control fixes now.  I believe that code avoids this whole issue entirely as it no longer takes simply walkers_per_rank to determine the default walkers. Various unit tests are updated or added to check it as well.

PDoakORNL:
My PR is proving difficult to break up. I also needed to change the way initial allocation of walkers worked for DMC.  So I'm just going to prep the full PR.  So I have no trouble with this being merged temporarily but the default is going to get switched back to 0 in that PR.  At the same time the bug is fixed.

PR #2336, opened 2020-03-11
Cades build

mcbennet:
## Proposed changes

 Updating ORNL CADES build script

1. An import of python is needed given recent transition to python3
2. Builds with and without AVX are included for use with skylake and
batch nodes, respectively.

- Build related changes

- No

ORNL CADES

## Checklist


config/build_ornl_cades.sh
```diff
@@ -37,47 +37,50 @@ CMAKE_FLAGS="-DCMAKE_C_COMPILER=mpicc \
...
 
-# Configure and build cpu real AoS
+# Configure and build cpu real SoA
 echo ""
 echo ""
 echo "building QMCPACK for cpu AoS real for CADES SHPC Condo"
-mkdir -p build_cades_cpu_real
-cd build_cades_cpu_real
-cmake -DENABLE_SOA=0 $CMAKE_FLAGS ..
+mkdir -p build_cades_cpu_real_SoA
```
jtkrogel: Probably better to call these "_skylake" rather than "_SoA".

config/build_ornl_cades.sh
```diff
@@ -37,47 +37,50 @@ CMAKE_FLAGS="-DCMAKE_C_COMPILER=mpicc \
...
+CMAKE_FLAGS="-DCMAKE_C_COMPILER=mpicc \
+             -DCMAKE_CXX_COMPILER=mpicxx"
+
+# Configure and build cpu real SoA without AVX
 echo ""
 echo ""
 echo "building QMCPACK for cpu SoA real for CADES SHPC Condo"
-mkdir -p build_cades_cpu_real_SoA
-cd build_cades_cpu_real_SoA
+mkdir -p build_cades_cpu_real_SoA_noavx
```
jtkrogel: Better to name as "_batch" rather than "_SoA_noavx".

mcbennet: Just to confirm, remove both "SoA" and "noavx" qualifiers and add "batch" ?

ye-luo: In this case, you still get avx2. noavx512 is more accurate

prckent:
Since there are many queues and various generations of hardware on CADES, it would be beneficial to very clearly state which queues and CADES units this targets in comments at the top of the file. e.g. CNMS nodes are haswells and broadwells.

jtkrogel:
There is only one queue that benefits from the AVX instructions (skylake).  All the others have AVX removed.  Correct?

Perhaps best then to have "qmcpack_cpu_real" and "qmcpack_cpu_real_skylake"?

prckent:
As the main users of the script, pick what you find logical. Should other people express an interest we can update.

config/build_ornl_cades.sh
jtkrogel: Keep skylake.

config/build_ornl_cades.sh
jtkrogel: Remove all qualifiers here.

jtkrogel:
Sounds good.

markdewing:
okay to test

Issue #2330, opened 2020-03-05
Final DMC energy varies with equilibration strategy


jtkrogel:
The final DMC energy for a 20 atom cell of SrCoO3 differs when QMCPACK is run in two ways that should be equivalent:

1. A single DMC run with a timestep of 0.002/Ha.
2. DMC run first with a timestep of 0.02/Ha followed by 0.002/Ha.

In this case, a difference of 6.6(7) mHa/f.u. is observed in the final DMC energy using the CPU code on Theta (case 1) and CADES (case 1 and 2) with Theta and CADES agreeing for case 1.

A plot of the 2-timestep data with differing numbers of equilibration blocks removed, seems to show a very long and slow transient.  For comparison, a full equilibration from VMC to the DMC 0.002/Ha equilibrium occurs within 100 blocks.  Starting from the 0.02/Ha equilibrium, the new 0.002/Ha equilibrium is not reached even after 1000 blocks.

![GS4_ediff](https://user-images.githubusercontent.com/24547905/76005030-75c4cc80-5ed8-11ea-8b55-b6892bed5f52.png)

Given the rapid equlibration from the high energy VMC state, it is unclear what mechanism could be responsible for the lack of equilibration from the lower energy, and presumably closer, DMC 0.02/Ha state.  More investigation of this effect is warranted.

The code's inability to relax to the same minimum from different initial states casts doubt on the equilibrium found from standard single shot calculations performed in production.

Issue identified by @mcbennet.

prckent:
Hoping to narrow this bug/feature: if you do a restart, does the problem persist?

prckent:
Attention @camelto2 . Until resolved, this issue would preclude a smarter equilibration strategy.

camelto2:
> Attention @camelto2 . Until resolved, this issue would preclude a smarter equilibration strategy.

Agreed. I've been chatting with @mcbennet about this offline. @lshulen and @rcclay have pointed out a number of issues that could be the culprit similar to what you mention above.

jtkrogel:
This is for T-moves version 0.  Trying a restart is a good idea.  Right now, we are performing reruns on CADES with slightly smaller errorbars for 20, 10, and 5 atom cells as initial runs gave ambiguous results for the smaller cells.  This should tell us if the issue is reproducible and whether it appears for smaller systems as well.

The 20 atom runs take about a day on 8 Skylake nodes.

jtkrogel:
[...]

PR #2349, opened 2020-03-18
[WIP] Fix for Issue #2330

mcbennet:
## Proposed changes

This PR contains a proposed fix to Issue #2330.
Further testing is needed to understand what builds
continue to show bias and why.

- Bugfix

- No

Workstation and NERSC Cori

## Checklist


PDoakORNL:
The whole resetRun thing for SimpleFixedNodeBranch was added in
a24d9a77ffc30c921fed82887ad66f7fec09d545
as a "bug fix" 😭 

It was called in DMC::resetComponents(xmlNodePtr cur)
I deleted it in 982ac99b8 when I tried to kill reuse of QMCDriver subtypes and removed the call to resetComponents for DMC.
Looking at whether any other damage might have been done to the legacy driver there.
Or if any state manipulation was dropped there.  I don't think so since I try to bring just the "state" hiding in SFNB into the unified driver.

For DMC legacy CUDA it is still called via another code path.

mcbennet:
@PDoakORNL something I have noticed that might be helpful as you look around is that when I run a deterministic test of the form VMC->DMC_timestep1->DMC_timestep2, and compare to 3.8.0, I see that all VMC quantities match to all decimals in the scalar files but only AoS gives matching quantities for  DMC_timestep1 (while SoA does not give matching quantities).

After combing through my outputs, I noticed that for AoS, DMC_timestep1 starts with the same trial energy between develop and 3.8.0 while SoA produces two different starting trial energies for 3.8.0 and develop.

PDoakORNL:
Looking at QMCDriver.cpp since the driver fishes the previous drivers EstimatorManagerBase point out of the SFNB `Estimators->reset()` needs to be called there since that was in resetComponents as well.

PDoakORNL:
@mcbennet Apologies for pushing directly to your branch. I intended to push to a branch on my fork but was a little to fast on the keyboard.

PR #2366, opened 2020-04-06
Clang linker workaround 

ye-luo:
## Proposed changes

1. Workaround clang linker bug for offload using object target. The code compiles now but not fully running yet due other bugs.
2. Split qmcwfs target into qmcwfs and qmcwfs_cuda to facilicate 1.
3. introduce spir64 for intel next gen compiler.

- Build related changes

- No

Ryzen-box

## Checklist


markdewing:
Item #3 should be a separate PR.  It looks like the files are completely disjoint from the files in the other two points, so it shouldn't be that hard to split out.

src/CMakeLists.txt
```diff
@@ -243,14 +243,14 @@ ENDIF()
...
-  ENDIF(OHMMS_DIM  MATCHES 3)
+  ENDIF(OHMMS_DIM MATCHES 3)
 
   #Need to add this if nothing else but miniapps is compiled
-  if(QMC_BUILD_SANDBOX_ONLY)
-    set(PARTICLE  ${PARTICLE} QMCWaveFunctions/WaveFunctionComponent.cpp)
-  endif()
+  IF(QMC_BUILD_SANDBOX_ONLY)
+    SET(PARTICLE ${PARTICLE} QMCWaveFunctions/WaveFunctionComponent.cpp)
+  ENDIF(QMC_BUILD_SANDBOX_ONLY)
```
markdewing, line 253: Adding the condition in 'endif' is for backwards compatibility (according to the 'endif' docs).     It probably shouldn't be added when cleaning up CMake files.

ye-luo: I think we can be flexible here. Sometimes putting more details in endif helps reading.

ye-luo:
> Initial comments without looking at the code:
> 
>     1. I agree with Mark, please split.
> 
>     2. Please specify which clang you are referring to. Does IBM XL on Summit work (main concern)? 9.x or 10.0?

1. I removed the housekeeping change. 
2. IBM is not affected. Now I'm able to develop code without the slow XL. Clang develop is what I need. 9 and 10 are outdated to me.

prckent:
It is good that you will be able to develop faster with this change. I would still like to know which compiler versions build QMCPACK with this change. Will release clang10.0 with the openmp target backend compile QMCPACK successfully? or does this require a change made after the 10.0 release branch? Does AOMP still work? Nightlies will be added using these compiler shortly.

ye-luo:
XL and AOMP always run. The fix for math header lands after Clang 10 unfortunately.

prckent:
Merging ahead of nightlies since CI is temporarily offline.

PR #2369, opened 2020-04-08
Fix for Issue #2330 

mcbennet:
## Proposed changes

This PR contains a proposed fix to Issue #2330.

I created a clean PR that does not contain the commits from @PDoakORNL, since I have only performed tests (shared in #2330 thread) using my commits.

- Bugfix

- No

Workstation and NERSC Cori. Workstation.

## Checklist


markdewing:
Can we get a description of why this call to resetRun is important for correct behavior?

prckent:
A consequence of this fix should be the bug-deterministic_diamondC_1x1x1_pp-vmc-dmc tests passing. These pass for real AoS only currently, and should at least pass for real SoA (default). I suspect that the test data is bad for complex / fixed-phase, so in a later update we should not run the test for complex runs or add complex reference data.
 
https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=bug-deterministic-diamondC_1x1x1_pp-vmc-dmc-lg_ts_sdj-1-1-2-totenergy&date=2020-04-08

mcbennet:
In #1704, @PDoakORNL intended to remove the re-use of sub-types of QMCDriver. Prior to this, when the DMC type was re-used, there was a call to resetComponents which called resetRun. After #1704, the call to resetComponents was removed. Likely there is state data that is not being properly passed to the new DMC driver when chaining runs and a call to resetRun mends this.

mcbennet:
I am noticing that the develop branch with SoA does not even pass the bug-deterministic_diamondC_1x1x1_pp-vmc-dmc for the first series (first time step). Is this expected?

prckent:
Paging @Hyeondeok-Shin

Hyeondeok-Shin:
@mcbennet Yes, it is expected. These were passed up to 3.8.0 and failed on 3.9.1. I first thought failure on the initial DMC on large time step affected on failure on the next DMC section as well.

prckent:
OK to test

PR #2373, opened 2020-04-09
Fix SOQMC missing assignment of spin in ParticleSet functions

camelto2:
#2300  Proposed changes

missing spins update in ParticleSet::loadWalker, which caused significant VMC bias when walkers/mpi > 2*OMP_NUM_THREADS. 
Also, missing spins in ParticleSet::assign which is used in ParticleSet constructor. 

- Bugfix

- No

linux workstation

## Checklist


ye-luo:
Okay to test

PR #2376, opened 2020-04-11
Better MKL cmake handling

ye-luo:
## Proposed changes
Set up MKL clean when it is provided by the system.
I noticed that the old behaviour adding -L/usr/lib/x86_64-linux-gnu and rpath caused compiler picking up OS installed libraries instead of compiler shipped libraries.
The new behavior doesn't add -L and rpath and leave the library linking path implicit.
So the logic is if the library is found implicitly, we leave the linking path implicit.

Note: If build from an existing folder, delete MKL_LINK_DIRECTORIES entry in CMakeCache.txt and cmake again before make.

- Build related changes

- No

Bora ubuntu 18.04, GNU + MKL(standalone), Intel + MKL(compiler -mkl)
Laptop ubuntu 19.10. GNU + MKL(system package)
Theta Cray. CC(Intel) + MKL(craype module)

## Checklist


prckent:
Please can you give more details about the combinations that you tested. e.g. With GNU compilers, with Intel, with MKL available, with openblas/blis etc.

ye-luo:
> Please can you give more details about the combinations that you tested. e.g. With GNU compilers, with Intel, with MKL available, with openblas/blis etc.

Updated the description. No impact if MKL is not used.

ye-luo:
The ALCF CPU CI uses MKL and GPU CI runs use system lapack/blas to ensure maximal diversity.

PR #2377, opened 2020-04-12
Fix last bit of #2330

### LINKED ISSUE #2330: Final DMC energy varies with equilibration strategy

ye-luo:
## Proposed changes
complete fix of #2330 
#2177 changed from
```
W.setActive(iat);
if (rr > m_r2max)
{
  ++nRejectTemp;
  continue;
}
W.makeMoveAndCheck(iat, dr)
```
to
```
if (rr > m_r2max)
{
  ++nRejectTemp;
  continue;
}
W.makeMoveAndCheck(iat, dr)
```
by shifting all the computation in setActive into makeMoveAndCheck.
So the early continue caused problem. The fix moves (rr > m_r2max) check after makeMoveAndCheck.
Now the failing deterministic test passes in real build.
The complex build still fails due to diverging behaviour between fix node and fix phase approximation. I have some thoughts but not fully hammered, so I will leave the resolution later.

This bug should only affect systems with very small cell (a few atoms <10) and very large time step.

- Bugfix

- No

Bora workstation.
## Checklist


src/QMCDrivers/DMC/DMCUpdatePbyPFast.cpp
```diff
@@ -83,13 +83,11 @@ void DMCUpdatePbyPWithRejectionFast::advanceWalker(Walker_t& thisWalker, bool re
       dr += sqrttau * deltaR[iat];
       RealType rr = tauovermass * dot(deltaR[iat], deltaR[iat]);
       rr_proposed += rr;
-      if (rr > m_r2max)
+      if (!W.makeMoveAndCheck(iat, dr) || rr > m_r2max)
```
prckent, line 86: No function calls like this in if statements please. The execution order is not obvious. Split.

src/QMCDrivers/DMC/SODMCUpdatePbyPFast.cpp
```diff
@@ -85,13 +85,11 @@ void SODMCUpdatePbyPWithRejectionFast::advanceWalker(Walker_t& thisWalker, bool
       ds += std::sqrt(tauovermass/spinMass)*deltaS[iat];
       RealType rr = tauovermass * dot(deltaR[iat], deltaR[iat]);
       rr_proposed += rr;
-      if (rr > m_r2max)
+      if (!W.makeMoveAndCheckWithSpin(iat, dr, ds) || rr > m_r2max)
```
prckent, line 88: No function calls like this in if statements please. The execution order is not obvious. Split.

src/QMCDrivers/RMC/RMCUpdatePbyP.cpp
```diff
@@ -153,13 +153,11 @@ void RMCUpdatePbyPWithDrift::advanceWalkersVMC()
       dr += sqrttau * deltaR[iat];
       RealType rr = tauovermass * dot(deltaR[iat], deltaR[iat]);
       rr_proposed += rr;
-      if (rr > m_r2max)
+      if (!W.makeMoveAndCheck(iat, dr) || rr > m_r2max)
```
prckent, line 156: No function calls like this in if statements please. The execution order is not obvious. Split.

src/QMCDrivers/RMC/RMCUpdatePbyP.cpp
```diff
@@ -284,13 +282,11 @@ void RMCUpdatePbyPWithDrift::advanceWalkersRMC()
       //RealType rr=dot(dr,dr);
       RealType rr = tauovermass * dot(deltaR[iat], deltaR[iat]);
       rr_proposed += rr;
-      if (rr > m_r2max)
+      if (!W.makeMoveAndCheck(iat, dr) || rr > m_r2max)
```
prckent, line 285: No function calls like this in if statements please. The execution order is not obvious. Split.

src/QMCDrivers/DMC/DMCUpdatePbyPFast.cpp
prckent: Rationale: many people will not know if makeMoveAndCheck is always called (or not).

ye-luo: Actually to me
1. makeMoveAndCheck is solid.
2. rr > m_r2max is a hack. Not clear what is this actually for. This type of hack always sounds like to make bad WF caused bad moves to continue.

prckent: Getting rid of the maximum displacement check is a discussion for another day.

Interesting that the intent to do something smart about large moves -- which should not happen in healthy runs -- might have precipitated a far more serious bug that affected actually healthy runs.

prckent: Just pull the call() out of the if statements for now.

ye-luo: Done. All pulled out.

PR #2381, opened 2020-04-15
Update multiple time step deterministic test

mcbennet:
Increasing initial timestep.

With an initial timestep of 0.6/Ha, the develop branch prior to #2377 fails for both DMC series. The error is also larger for the first series. 

## Proposed changes

Changing initial timestep of tests/solids/diamondC_1x1x1_pp/det_qmc_short_vmc_dmc_lg_ts.in.xml from 0.3/Ha to 0.6/Ha

- No

Workstation

## Checklist


ye-luo:
I'd like to remove bug tag ASAP so we can track it regularily.
I know that the complex build fails due to fixed phase vs fixed node. We either need to add reference number for complex or skip the test.
Also mixed precision needs to be checked.
@Hyeondeok-Shin please advice.

mcbennet:
> I know that the complex build fails due to fixed phase vs fixed node. We either need to add reference number for complex or skip the test.

I can add these to the present PR

prckent:
My understanding is that the larger timestep simply makes the test more sensitive, since old versions of the code with the bug were passing even with the shorter timestep. Good.

@mcbennet 

Thanks for the update.

I suggest to rename
bug-deterministic-diamondC_1x1x1_pp-vmc-dmc-lg_ts_sdj
as
deterministic-diamondC_1x1x1_pp-vmc-dmc-multi_ts_sdj

or similar. The key thing we are testing is varying timesteps.

Please modify this to only run for builds where it has a chance of passing. Ensure CUDA, complex and mixed precision are all skipped.

ye-luo:
Skipping CUDA, complex, mixed precision is an OK solution to me as well. The purpose is not to confuse people with an unintended test failing in these builds.

Hyeondeok-Shin:
@ye-luo I agree with skipping those tests for now.

ye-luo:
Okay to test

mcbennet:
SoA and AoS appear to diverge for 0.6/Ha case -- for both develop and 3.8.0.
Though I have confirmed that across 3.8.0 and Develop, SoA and AoS agree individually, e.g., for AoS case I see that 3.8.0 and Develop agree and same is true for SoA.

prckent:
The AoS code failed due to this divergence:

```
        Start 821: deterministic-diamondC_1x1x1_pp-vmc-dmc-multi_ts_sdj-1-1-2-totenergy
219/279 Test #821: deterministic-diamondC_1x1x1_pp-vmc-dmc-multi_ts_sdj-1-1-2-totenergy .....***Failed    0.08 sec
Tests for series 2
  Testing quantity: LocalEnergy
    reference mean value     : -10.20986762
    reference error bar      :   0.00000100
    computed  mean value     : -10.23546172
    computed  error bar      :   0.00000000
    pass tolerance           :   0.00000300  (  3.00000000 sigma)
    deviation from reference :  -0.02559410  (-25594.09700000 sigma)
    error bar of deviation   :   0.00000100
    significance probability :   1.00000000  (gaussian statistics)

Test status: fail
```

Since the AoS code is not long for the world, you could skip it. Alternatively, add distinct test data, or try playing with the input parameters to avoid the divergence between AoS and SoA (if possible). Your call.

PR #2385, opened 2020-04-16
Initialization of particle spins and modify spin drift by UNR modifier

camelto2:
## Proposed changes

This PR fixes a bug about spin initialization. If a user didn't input explicit spin variables in the xml, then there was a significant bias in the energy when using spin drift as all spins were intialized to zero. This adds a spin randomization in all the places I could think of where ParticleSetPool first initializes the electron ParticleSet. also adds to a unit test to make sure the spin variable isn't 0 when calling ParticleSetPool::randomize.

I also added a small refactor to the spin drift in the VMC and DMC Updates. This just moves the spin drift move calculations into the Greens Function UNR modifier, just to make it more analogous to how the spatial moves are done. 

- Bugfix

- No

linux workstation. intel and clang compilers

## Checklist


ye-luo:
Okay to test

camelto2:
@ye-luo after this is merged, I have a fix to the spindrift data type ready to submit

PR #2393, opened 2020-04-18
Fix mpi_wrapper dependency on libboost headers.

ye-luo:
## Proposed changes
The missing depdency breaks ryzen-box.
```
/home/yeluo/opt/qmcpack/external_codes/mpi_wrapper/mpi3/../mpi3/../mpi3/../mpi3/../mpi3/detail/datatype.hpp:9:9: fatal error: boost/serialization/strong_typedef.hpp: No such file or directory
 #include<boost/serialization/strong_typedef.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

- Build related changes

- No

ryzen-box

## Checklist


ye-luo:
boot headers may be found in system include folder if libboost-dev packages is installed on Ubuntu.
In this case, if the explicit boost header path is not added on the command line, the system one will be picked up via the implict paths.

On ryzen-box, I didn't install the system boost package and spotted this issue.

PR #2394, opened 2020-04-21
Fully adopt RefVector in WFC and SPOSet

ye-luo:
## Proposed changes
Replace std::vector<T*> with RefVector<T> for clear ownership.

- No

Bora

## Checklist


ye-luo:
@PDoakORNL initiated the change but we didn't manage to finish the transition. I'm adding new APIs due to omp needs. So it is better to complete the transition sooner than later. It also helps reducing the scale of my upcoming changes when reviewing.

PR #2398, opened 2020-04-22
Introduce DiracDeterminantBatched

ye-luo:
## Proposed changes
Introduce DiracDeterminantBatched stage 1. Use SM-1 algorithm.
1. Introduce Platforms/OpenMP/ompBLAS. Implements (batched) GEMV and GER in OpenMP. Interface mimics rocBLAS.
2. API change Fuse flex/mw_acceptMove and flex/mw_restore into flex/mw_accept_rejectMove.
3. Add SPOSet::mw_evaluateVGLandDetRatioGrads. It has fallback. SplineC2ROMP contains specialization. Orbital values, gradients, laplacian stay in the device.
4. SPOSet::useOMPoffload to identify offload implementation from outside.
5. test_TrialWaveFunction_diamondC_2x1x1.cpp and test_DiracDeterminantBatched.cpp unit tests to track newly added classes and code paths. The former one is sensitive to precision and machine, will keep an eye in the nightly tests in a wide range of machines.

Stage 2 will move more memory to device to avoid expensive transfer.
Stage 3 will connect batched matrix inversion.

- New feature

- No

Ryzen-box, Summit
## Checklist
- Yes. This PR adds tests to cover any new code


prckent:
(Ghost definition is already there)

PR #2399, opened 2020-04-22
Nexus: Resolve issue due to underlying xml library

mcbennet:
## Proposed changes

In some cases, expat seems to randomly insert "\n" when parsing xml input -- likely a result of a "chunking" prcoedure that is used. By setting `buffer_text = True`, this is avoided.

- Bugfix
- Update to xmlreader unit test

- No

Desktop

## Checklist


ye-luo:
Okay to test

PR #2400, opened 2020-04-23
Protect OpenMP user-defined reduction with macro.

ye-luo:
## Proposed changes
Protect OpenMP user-defined reduction with macro for compilers incapable of handling it.
user-defined reduction is not an offload-only feature but a generic one in OpenMP.

- Build related changes

- No

Ryzen-box

## Checklist


ye-luo:
@prckent please verify if it fixes your issue. If the compiler doesn't have built-in support for reduction of complex numbers, we are forced to find nasty solutions. So far GCC, Intel, Clang, XL has no complain.

PR #2410, opened 2020-04-27
DiracDeterminantBatched stage 2

ye-luo:
## Proposed changes
1. Make more data resident on accelerator and reduce data transfer.
2. Connect batched driver TrialWaveFunction::flex_completeUpdates
3. Some bug fix as well.

- Bugfix
- Code optimization

- No

Summit, Ryzen-box

## Checklist


ye-luo:
> Thoughts on future refactoring:
> 
>     * isOMPoffload() won't scale as (if?) more backend implementations are added.
> 
I think the actually meaning is whether certain data is available on the device or host only.
As we develop more, we can refine its definition which is not very clear in my mind.
On the other hand, if we no more do virtual SPOSet and use template, this may get avoided from the root. This is not easy decision to make right now.
>     * lots of ugly packing & unpacking. Not sure how this can be avoided.
We need to find a nicer way for packing and unpacking send/receive buffers. This is a performance optimization regardless of what programming model to use. We have some buffer code on the CPU but I'm not attempting it right now because more analysis will be needed to see if that buffer code is appropriate or not to use for device.

prckent:
I agree with both comments. For now I think we keep developing and then take care to analyze what we have once the offload is decently performing.

PR #2422, opened 2020-05-01
Adjust CheckQuadratureRuleReal tolerance

ye-luo:
## Proposed changes
Adjust tolerance to pass gcc8 and 50 quadrature point rule check in mixed precision on CI nodes.

- Bugfix

- No

ALCF CI nodes.

## Checklist


prckent:
It would be smarter for us to scale the allowable error by the size of the grid.

PR #2423, opened 2020-05-01
Add MatrixUpdateCUDA

ye-luo:
## Proposed changes
Add MatrixUpdateCUDA which is parallel to MatrixUpdateOMP. It has significantly more asynchronicity inside the class which behaves very similar to DelayedUpdateCUDA.
SM-1 update is implemented so far.

The benchmark system 128 atom problem 224 walker per GPU. has no J2 and ElecElec to bypass the e-e distance table impact.

legacy CUDA running SM-1
```
  DMCcuda                                 123.1690     2.9774              1     123.168981343
    DMCcuda::Drift_Diffuse                100.2203    91.7546             25       4.008810757
    DMCcuda::Hamiltonian                   15.6763     0.0002             25       0.627052712
```
Offload + CUDA.
```
  DMCBatched                                   209.3373     0.0247              1     209.337344183
    DMCBatched::RunSteps                       200.9368     2.9490             25       8.037471924
      DMCBatched::Hamiltonian                   76.3376     0.0009             25       3.053505973
      DMCBatched::MovePbyP                     114.5481     0.6045             25       4.581925263
```
It is 100 vs 114 seconds now in the p-by-p move portion. Still doing more analysis.

Cautious. The code is not production ready. I have to by pass a few transfer to have fair comparison. The recompute has not been correctly implemented in the unified driver. So there is still numerical instability in the mixed precision code.

- New feature

- No

Summit

## Checklist


prckent:
Our manual CI found no problems, not that offload is enabled with these builds yet.

ye-luo:
>     1. What is the motivation for this new implementation? My understanding is that this is a freshly written backend to the batched driver that uses some newly written CUDA code to help clarify issues with the OpenMP offload performance and also hopefully get an acceptably enough implementation on Summit in the interim. Correct? 

It is intended to sort out a few implementation details, particularly desired asynchronous behaviour and vendor library interoperability. The MatrixUpdateCUDA started from MatrixUpdateOMP and replaced a few implementation that can only be done directly with CUDA. So this class works only if we have both ENABLE_CUDA and ENABLE_OFFLOAD. Having two implementation in parallel also helps me to sort out a proper abstraction around MatrixUpdateXXX class and have a generic DiracDeteriminantBatched class. The API refactoring is still on going, I'm working on MatrixDelayedUpdateCUDA and certain abstraction is still evolving. I believe it supersedes MatrixUpdateCUDAeventually. We may also finally combine batching MatrixDelayedUpdateCUDA and non-batching DelayedUpdateCUDA. It takes a few steps to reach the end. Handwritten kernels will be OpenMP and BLAS/LAPACK calls vendor libraries.

>     2. So far the CUDA introduced is only for batched BLAS2-like functionality. Do you anticipate any more will be needed to make an acceptably fast code on Summit at this time? e.g. In the non-local evaluation or distance tables?
>
Unfortunately, almost all the batched BLAS functions are missing in cuBLAS. My hand written ones are not only intended to make the code faster but to allow drop-in replacement with other GPU vendor libraries. I can fuse a few kernels to get even faster but I prefer to stay as BLAS interface. I do need two extra BLAS1 for delayed update but not more BLAS2 needed.
>     3. You implemented some missing cuBLAS features in a file cuBLAS_inhouse. I think it is important to call these what they are since "inhouse" might be interpreted that we are reimplementing parts of cuBLAS, when really we (you) are implementing not-yet provided functionality that we would prefer not to have to do. I suggest renaming CUBLAS_inhouse -> CUBLAS_missing_functions or something similar. A bit longer but the whole files will be deletable if/when NVIDIA implements a full batched interface as AMD have already done ( https://docs.nvidia.com/cuda/cublas/index.html#blas-like-extension vs e.g. https://rocblas.readthedocs.io/en/latest/api.html#rocblas-type-gemv-batched  )
> 
Yes. I can update the name but please allow me not changing it in this PR. as I'm working on the other PR and this inhouse file has been touched as well. You will see the renaming in my next PR.

>     4. Define all the functions in CUBLAS_missing_functions. "This is a batched implementation of BLAS2 gemv assuming contiguous/non-strided access" etc.
> 
No problem, in the next PR with the renaming, all the functions BLAS1/2 in the header will properly documented.

>     5. Please discuss how this code is built (which options are needed) and eventually document. 

When you have ENABLE_CUDA and ENABLE_OFFLOAD, this code path will be added and as default.

>     6. I intend to insist that switching between these implementations be done in the input. This is an ideal time to work this out. We will need the ability to run either of these implementations and we should not continue the legacy policy of ever more specialized executables. They are too much of a burden.
> 
Yes. This is supported in the template class approach. I have not sorted out the input names, combinations and default. I have at least three implementations OMP, CUDA and CUDA + delayed update working with batched driver. It adds difficulty to arrange input tags. Please allow me to finish the last one and I may not need the CUDA SM-1 case any more.

> 
> Looking at SlaterDetBuilder.cpp I see
> 
> ENABLE_CUDA and ENABLE_OFFLOAD goes via MatrixUpdateCUDA
> ENABLE_OFFLOAD (only) goes via MatrixUpdateOMP
> 
> We need a flag (poll the group?) to select the implementation "offload" vs "cuda" vs "hip" etc.
>
We may need a wider discussion about the input names.
We have gpu=yes/no. I don't expect users to call out  a specific implementation and pollute input files. gpu=yes should go to the case available on their platform. "offload" is also very unique because it works both on CPU and GPU.
For users, only batch=yes/no, gpu=yes/no are relevant and they don't need to selecting cuda/hip/offload. More logically, it should be CPU or Offload+X (X=cublas/rocBLAS/MKL) eventually.

>     1. When do you plan to add a test?
Unit test is in-place. test_TrialWaveFunction_diamondC_2x1x1.cpp. One integration test is missing but depending on settling down input tags. Once I completed MatrixDelayedUpdateCUDA this situation should be more clear about input tag and then setup an integration test.

ye-luo:
Test this please

ye-luo:
Test this please

src/Platforms/CUDA/cuBLAS_inhouse.cu
```diff
@@ -0,0 +1,293 @@
...
+                                   const int incy,
+                                   const int batch_count)
+{
+  //return gemv_batched_impl(handle, trans, m, n, (const cuDoubleComplex*)alpha, (const cuDoubleComplex* const*)A, lda, (const cuDoubleComplex* const*)x, incx, (const cuDoubleComplex*)beta, (cuDoubleComplex* const*)y, incy, batch_count);
+  return cudaSuccess;
+}
+
+
+template<typename T, int ROWBS, int COLBS>
+__global__ void ger_batched_kernel(const int m, // number of columns in row major A
```
PDoakORNL, line 180: This is one important operation than MAGMA lacks, otherwise I'd argue we should use it.  Unfortunately it can also be a bit of a POS to build on some systems so...

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
```diff
@@ -107,7 +107,7 @@ void DiracDeterminantBatched<DET_ENGINE_TYPE>::mw_evalGrad(const RefVector<WaveF
   for (int iw = 0; iw < nw; iw++)
   {
     auto& det = static_cast<DiracDeterminantBatched<DET_ENGINE_TYPE>&>(WFC_list[iw].get());
-    invRow_list[iw] = det.psiMinv_dev_ptr + NumOrbitals * WorkingIndex;
+    invRow_list[iw] = det.psiMinv_dev_ptr + psiMinv.cols() * WorkingIndex;
```
PDoakORNL, line 110: Sometimes you use NumOrbitals sometimes a dimension of the data object that is over NumOrbitals.  If mixed naming is purposeful make a explanatory comment explaining the deviation (for instance psiMinv.cols() might be distributed or a subset?

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
```diff
@@ -156,16 +156,16 @@ void DiracDeterminantBatched<DET_ENGINE_TYPE>::mw_ratioGrad(const RefVector<Wave
     auto& det = static_cast<DiracDeterminantBatched<DET_ENGINE_TYPE>&>(WFC_list[iw].get());
     phi_list.push_back(*det.Phi);
     if (Phi->isOMPoffload())
-      psiMinv_row_dev_ptr_list[iw] = det.psiMinv_dev_ptr + NumOrbitals * WorkingIndex;
+      psiMinv_row_dev_ptr_list[iw] = det.psiMinv_dev_ptr + psiMinv.cols() * WorkingIndex;
```
PDoakORNL, line 159: see above NumOrbitals vs. psiXXX.cols()

src/QMCWaveFunctions/Fermion/MatrixUpdateCUDA.h
```diff
@@ -0,0 +1,314 @@
...
+#include "Platforms/OpenMP/ompBLAS.hpp"
+#include "Platforms/OpenMP/ompReduction.hpp"
+#include <cuda_runtime_api.h>
+#include "Numerics/CUDA/cuBLAS.hpp"
+#include "Platforms/CUDA/cuBLAS_inhouse.hpp"
+#include "QMCWaveFunctions/Fermion/matrix_update_helper.hpp"
+
+namespace qmcplusplus
+{
+/** implements dirac matrix update using OpenMP
```
PDoakORNL, line 30: Should say MatrixUpdateCUDA implements a specialized type of the generic DET_ENGINE type.

src/QMCWaveFunctions/Fermion/MatrixUpdateOMP.h
```diff
@@ -19,11 +19,9 @@
...
 #include "Platforms/OpenMP/ompBLAS.hpp"
+#include "Platforms/OpenMP/ompReduction.hpp"
 
 
-PRAGMA_OFFLOAD("omp declare reduction(+: std::complex<float>: omp_out += omp_in)")
-PRAGMA_OFFLOAD("omp declare reduction(+: std::complex<double>: omp_out += omp_in)")
-
 namespace qmcplusplus
 {
 /** implements dirac matrix update using OpenMP
```
PDoakORNL, line 27: Document as for MatrixUpdateCUDA

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -474,31 +474,46 @@ bool SlaterDetBuilder::putDeterminant(xmlNodePtr cur, int spin_group)
     return true;
 
   DiracDeterminantBase* adet = 0;
-  {
+
 #if defined(QMC_CUDA)
-    adet = new DiracDeterminantCUDA(psi, firstIndex);
+  adet = new DiracDeterminantCUDA(psi, firstIndex);
```
PDoakORNL, line 479: This deserves a \todo comment (make this logic readable).  minimum verbosity is not a goal for this sort of mess, ideally the design shouldn't require it.

src/Platforms/CUDA/cuBLAS_inhouse.cu
prckent: We can potentially ask the MAGMA folks to develop this via ECP as a hedge against incomplete vendor accelerated BLAS implementations.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
ye-luo: psiMiv.cols() is the stride. Need to revisit whether padding the NumOrbitals is necessary or not.

src/QMCWaveFunctions/Fermion/MatrixUpdateCUDA.h
ye-luo: Updated comments.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
ye-luo: Added a TODO.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
ye-luo: see above reply.

ye-luo:
Is it OK to merge now?

src/Platforms/CUDA/cuBLAS_inhouse.cu
prckent: https://icl.cs.utk.edu/projectsfiles/magma/doxygen/group__magma__ger__batched.html ?

PR #2427, opened 2020-05-05
Remove sys/stat.h inclusion

ye-luo:
## Proposed changes
sys/stat.h seems causing problem on nitrogen.
Trying just removing them.

- Bugfix

- No

Bora

## Checklist

### DIFF
--- a/src/AFQMC/HamiltonianOperations/tests/test_hamiltonian_operations.cpp
+++ b/src/AFQMC/HamiltonianOperations/tests/test_hamiltonian_operations.cpp
@@ -23,9 +23,6 @@
 #undef APP_ABORT
 #define APP_ABORT(x) {std::cout << x <<std::endl; exit(0);}
 
-#include <sys/stat.h>
-#include <unistd.h>
-#include <stdio.h>
 #include <string>
 #include <vector>
 #include <complex>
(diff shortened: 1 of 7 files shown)


src/AFQMC/HamiltonianOperations/tests/test_hamiltonian_operations.cpp
```diff
@@ -23,9 +23,6 @@
 #undef APP_ABORT
 #define APP_ABORT(x) {std::cout << x <<std::endl; exit(0);}
 
-#include <sys/stat.h>
-#include <unistd.h>
```
fdmalone, line 27: PR says remove sys/stat.h. Why are you removing the other headers?

ye-luo: I don't see a good reason to keep these lines. So just clean them on the way.
They are C or OS related headers. So I prefer removing them unless there is a real need.

ye-luo: @markdewing you introduced these .h file inclusion. Do you remember why they are needed when first introduced?

markdewing: I think I'm listed in 'blame' because I reorganized some large check-in of the AFQMC code.   I don't think I originated it.

ye-luo: @markdewing thanks for the confirmation that there is no specific reason to keep them.

PR #2436, opened 2020-05-08
3 body Jastrow forces

rcclay:

## Proposed changes
This adds finite difference calculation of the wave function ion gradients required to use 3 body Jastrow factors with ZVZB force calculations.  This will remain until analytic derivatives are derived and implemented.  
  
- New feature

- No

Intel Xeon Haswell
## Checklist

NOTE:  There isn't a unit test for this yet, but WaveFunctionTester can and has been used to validate this feature.


rcclay:
Looks like I was sloppy with MP types.  I'll fix tomorrow.

src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
```diff
@@ -969,6 +986,106 @@ class JeeIOrbitalSoA : public WaveFunctionComponent
...
+      P.update();
+      LogValueType log_m = evaluateLog(P, Gm, Lm);
+      QTFull::RealType log_p_r(0.0), log_m_r(0.0);
+      
+      log_p_r=log_p.real();
+      log_m_r=log_m.real();
+      dG=Gp-Gm;
+      dL=Lp-Lm;
+      //symmetric finite difference formula for gradient.
+      g_return[iondim] = c1 * (log_p_r - log_m_r);
```
ye-luo, line 1077: This formulation is only correct without node-crossing. I noticed this pattern in wftest and we need certain protection there as well.

ye-luo: OK. It is fine here since Jastrow is bosonic. But not a general case in wftest.

ye-luo:
Test this please.

src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
rcclay: What are you thinking?  Check for a sign difference and abort if that happens?  Abort if log_p or log_m gets really really negative?

rcclay: It's moot for the Jastrow, but a valid concern for wftester that I hadn't considered.

ye-luo: No problem here in the Jastrow.
In the generic case, the sign of psi matters and so as the phase of log(psi) not the sign of log(psi).

rcclay: Gotcha.

Issue #2434, opened 2020-05-07
Build broken with Intel2019.1 and Clang10 due to Multi allocator.hpp due to missing destroy function


prckent:
**Describe the bug**

Build is broken with Clang10 and AFQMC enabled
https://cdash.qmcpack.org/CDash/viewBuildError.php?buildid=115274

e.g.
```
qmcpack/external_codes/boost_multi/multi/memory/allocator.hpp:62:34: error: no matching member function for call to 'destroy'

undefined                allocator_traits::destroy(ctor_, p);
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
undefined
```

**To Reproduce**

Run latest develop with Clang10 and AFQMC enabled (implies an MPI build)
e.g. as configured at https://cdash.qmcpack.org/CDash/viewConfigure.php?buildid=115274

**Expected behavior**

QMCPACK should build

**System:**

sulfur, but should not matter

**Additional context**

(none)

correaa:
Before touching the library I would like to see if the allocator for `shm::shm_ptr_with_raw_ptr(_dispatch?)` defines (or deletes) a `destroy`.

prckent:
Thanks @correaa 

Intel2019.1 also fails

mmorale3:
This should be fixed now.

PR #2440, opened 2020-05-08
Mixed precision improvements for AFQMC

fdmalone:
## Proposed changes

Some improvements to mixed precision handling for AFQMC. Mostly removing ifdefs.

- No

quartz (intel Xeon) + lassen (~summit)

## Checklist


ye-luo:
I tend towards blocking this PR until AFQMC unit test abort fix comes in.

prckent:
I also think we should hold off AFQMC updates until there is a fix for https://github.com/QMCPACK/qmcpack/issues/2434 (@mmorale3 , @correaa ). I suspect that this will be one-few line fix. If it turns out to be complex and will take a long time we can revisit.

Rationale: We shouldn't risk building up more problems that will then be harder to chase down.

ye-luo:
Okay to test

PR #2445, opened 2020-05-11
Remove bug tag on ion-ion deterministic test for a graphene.

Hyeondeok-Shin:

## Proposed changes

As the related ion-ion bug had fixed (See #2137), tag on bug-deterministic-grapheneC needs to be rearranged. Two tests (z10 & z30) are now passing ewald check with larger LR_dim_cutoff, so the bug tag has removed and they moves to the deterministic test section.

This deterministic tests are now testing ion-ion for two different LR handler. z10 is for the default optimized break up and  z30 is for the Ewald3D handler. (See #2190)
Ion-ion references are generated in QMCPACK 3.9.1, and they're all passing with the intel 18 compiler on Intel Xeon.    

- Other (please describe): Adding test.

- No

 Intel Xeon

## Checklist


prckent:
Test this please

prckent:
deterministic-grapheneC_1x1_pp-vmc_sdj_z10-1-1-ionion failed.

No deterministic tests can fail.

Hyeondeok-Shin:
This test is still passing in my branch. I'm now working to find the reason.

Hyeondeok-Shin:
It seems CPU dependency has increased as increasing LR_dim_cutoff. This table shows ion-ion references from the same gcc 8.2 compiler but different computing node.


LR_dim_cutoff | login node | compute node | Diff
-- | -- | -- | --
10 | 14.003313 | 14.00331301 | 1E-08
20 | 14.00447033 | 14.00411253 | 0.0003578
30 | 14.00274806 | 14.00274287 | 5.19E-06
40 | 14.00270343 | 14.00270671 | 3.28E-06

We can see that this values are from the same binary but different in large LR_cutoff. So tolerance needs to be increased to do deterministic test for the large cutoff.

prckent:
Strange to only have 5 digits of accuracy for LR_dim_cutoff=20. Perhaps there is some incorrect usage of single precision maths?

prckent:
Also strange to have different results of course! I think this points to a bug.

Hyeondeok-Shin:
@prckent It could be. I'd like to keep this PR as the blocked until receiving additional comments from the other reviewers. I'm not sure whether increased tolerance makes sense or not.

Hyeondeok-Shin:
As talked in #2472, deterministic test with opt-breakup for quasi 2D system is not available because of numerical instability. We do not know the exact reference of the deterministic test for opt-breakup for now.
So opt-breakup is replaced to the Ewald3D and test is passing with different CPUs now.

Hyeondeok-Shin:
@prckent I think it is good to merge if CI passes.

prckent:
No need to hold the PR (now approved, sorry for delay), but one missing determinstic test is that we check the code correctly gives an error when opt-breakup is used with bad parameters &/or at an unstable geometry. i.e. This test would have expect fail set , as per the check for real code with complex twists.

prckent:
Test this please

PR #2456, opened 2020-05-14
Intial linear algebra text for real space QMC

ye-luo:
## Proposed changes
Documentation for linear algebra.

- No

ryzen-box

## Checklist


prckent:
Thanks Ye. I think we can expand this section in future but we can certainly merge now.

PR #2457, opened 2020-05-14
Label all statistical test result checks as unstable

jtkrogel:
## Proposed changes

Move all statistical tests into the unstable category.  All tests should still run and report on CDash based on non-failure of QMCPACK execution.

- No

laptop


ye-luo:
Okay to test

Issue #2448, opened 2020-05-12
AFQMC crashes with unexpected complex/real integral file


prckent:
**Describe the bug**

Per user report on Google groups and subsequent discussion ( https://groups.google.com/forum/#!topic/qmcpack/2gGphvbBPyk ), the AFQMC code will crash if a real-valued compilation encounters an integral file with complex numbers. The AFQMC code should detect this (handle the HDF5 error or enquire of the file before reading) and give an appropriate error.  If the complex code can not load real valued files similar error handling should be added.

This is a bug more than a feature request because it is a very easy mistake to make and it should be handled gracefully. e.g. The real space code gives advice if a complex wavefunctions are needed for a specific twist.

**To Reproduce**

Develop. Example HDF5 file on https://groups.google.com/forum/#!topic/qmcpack/2gGphvbBPyk

**Expected behavior**

A meaningful error message and abort.

**System:**

Any

**Additional context**

None

mmorale3:
How hard is it to get the dimensions of the dataset without actually reading it?
I think there is a function right now that does it by reading the dataset, which
is too inefficient I think. If we can get dimensions easily, we can check before reading them
and would give us a chance to abort gracefully.

fdmalone:
The functions that exist throw errors which could be caught checkShapeConsistency

ye-luo:
It seems that you want a way to query the rank and size of each rank of a dataset. This can be exposed. We have internal check codes which never read the dataset but the metadata only.

My concern is that relying on querying the dimension of a particular dataset is not an optimal safey strategy.

mmorale3:
We can just catch the exceptions, this should be good enough for better error messages.
Having access to the actual dimensions could also be useful. We could read real data in the complex build if we know ahead of time that it is real. Can you expose them? We'll use them with caution.

fdmalone:
Ye's original suggestion to add a flag in the HDF5 would work too?

mmorale3:
We should do both. We can't guarantee that our tools are used to build the
input files.

fdmalone:
Good point.

mmorale3:
On the other hand, we can make it a requirement of the format.

ye-luo:
Multiple kinds of exception may occur, it is not safe to blindly thrust an error was caused by real/complex difference.

Reading real and then used as complex doesn't require you to know the actually rank of dataset. The rank difference is caused by data type which is managed by hdf_archive not outside.

[...]

PR #2460, opened 2020-05-15
Check integral type for AFQMC 

fdmalone:
## Proposed changes

An attempt at fixing #2448 . I check the ComplexIntegral flag initially if present. Following this I added a little helper function to check the dimensions of the data set to be read. This will abort if QMC_COMPLEX=0 and complex integrals are found, or copy the real integrals to their internally used complex arrays otherwise.

I ignored THC because of Miguel's rewrite of this, and assume the kpoint code always will read complex integrals. Currently the dense code is only build for real builds, and so here I just check we aren't trying to read complex integrals.

The handling of this was a little hacky so suggestions are welcome.

- Bugfix

- No

xeon / summit

## Checklist


ye-luo:
Okay to test.

Issue #2458, opened 2020-05-14
AFQMC Hamiltonian needs sanity tester script.


fdmalone:
**Is your feature request related to a problem? Please describe.**
Not currently.

**Describe the solution you'd like**
An external script to test externally generated AFQMC hdf5 hamiltonian file.

**Describe alternatives you've considered**
None.

**Additional context**
Don't want externally generated hamiltonians causing runtime crashing unnecessarily.

fdmalone:
Closed following merge of #2461

PR #2461, opened 2020-05-16
AFQMC hamiltonian sanity checker

fdmalone:
## Proposed changes

Following up on the ECP call, it was mentioned that it would be useful to have a script to check the AFQMC file format. I added test_afqmc_input.py which does this for the Hamiltonian at least. I already had a function to return the data sets as a dictionary so I just call this with some better exception handling + error messages. Addresses #2458.

Fatal exceptions are raised and silent ones (for datasets which aren't used further in constructing the dictionary) are checked after the fact. @jtkrogel mentioned this could be useful for the nightlies.

- New feature

- No

python 3.7.7 macbook pro

## Checklist


ye-luo:
Okay to test

PR #2473, opened 2020-05-21
AFQMC kernel unit tests

fdmalone:
## Proposed changes

Add unit tests to cover kernels before portability work.

- Other (please describe): Tests

- No

intel xeon + summitish

## Checklist


fdmalone:
I need to remove some functions and enable one or two tests before merging.

fdmalone:
Good to go now.

ye-luo:
Okay to test

fdmalone:
I think it should be good now.

fdmalone:
yikes, give me a minute.

fdmalone:
is __CUDA_ARCH__ < 600 for alcf-gnu-CUDA. The mkl failure looks unrelated to this PR.

ye-luo:
CUDA_ARCH default is sm_35. Could you reproduce the failure if you leave CUDA_ARCH unset and build on your machine? The alcf-gnu-mkl18 failure is just a glitch.

fdmalone:
Yeah. I was building with sm_70. I need to undo some things.

ye-luo:
CUDA documentation mentions it only available for double since 600. Could you adopt the workaround provided in the doc?

fdmalone:
It was there originally. I removed it in this PR.

mmorale3:
The workaround is stupidly slow, makes the entire code useless if it is actually needed.
In my opinion, the AFQMC code does have the constraint CUDA_ARCH >= 600

ye-luo:
Leave that workaround so we can still use sm_35 but we may raise the default value to sm_60 or sm_70 to prevent people hitting there by default.

mmorale3:
Why are we doing the CI with sm_35? Do we actually support Kepler?
Even Maxwell is a real stretch. I don't see how any of the newer code will be able to maintain this.
Can you do offload to Kepler or Maxwell?

ye-luo:
Yes. I can offload to kepler. For people trying on kepler, slow code is better than not build.
If we find too many workaround needed for kepler, we can consider drop it.
I also think we should raise the default arch to at least pascal.

prckent:
I strongly vote in favor of a Pascal or Volta default and forgetting previous generations (it is 2020!). The current default is also clearly bad and I was thinking of PRing it to Volta in any case. We can discuss elsewhere as needed.

fdmalone:
Sorry for the delay. Had to walk dog.

PR #2485, opened 2020-05-26
Remove AFQMC statistical test

fdmalone:
## Proposed changes

AFQMC statistical tests were developed when there was no CUDA code. These are failing on the cdash but shouldn't be run. 

- Bugfix

- No

lassen

## Checklist

### DIFF
--- a/tests/solids/CMakeLists.txt
+++ b/tests/solids/CMakeLists.txt
@@ -24,7 +24,7 @@ IF (QMC_COMPLEX)
    SUBDIRS("diamondC_1x1x1-Gaussian_pp_Tw_cplx_MSD")
    SUBDIRS("Al-1x1x1-Gaussian_pp_ShiftedTwist")
 ENDIF()
-IF (BUILD_AFQMC AND QMC_COMPLEX)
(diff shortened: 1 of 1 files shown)


prckent:
Why doesn't this work with CUDA?

mmorale3:
The tests should work, but they are not properly setup right now.
For example, they are not being submitted using the same number of (active) gpus and mpi tasks per node, which is a requirement in afqmc. This just needs some cleaning.

fdmalone:
Some of the Hamiltonians aren't implemented on GPUs (sparse). AFQMC also requires the number of MPI tasks to equal the number of devices, so I think they will never work as currently configured (I think they use 16 tasks if I remember properly).

In connection with this, assuming they did run, the number of walkers would differ from the benchmarks (there would be fewer walkers), so I wouldn't necessarily trust the statistics, at least it would be slightly different than what was expected for the CPU runs.

mmorale3:
I asked Fionn to turn them off until we can fix this.

prckent:
Sounds very reasonable.

prckent:
Test this please

Issue #2487, opened 2020-05-27
deterministic-unit_test_drivers consistently failing with GCC on sulfur


prckent:
**Describe the bug**

All GCC builds fail in deterministic-unit_test_drivers on sulfur (Xeon 6248R).
https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=deterministic-unit_test_drivers&date=2020-05-27

MPI, No MPI, MKL and Blis+Netlib LAPACK, real and complex all fail. However builds with clang and Intel compilers are OK. The problem dates before 7 May and the beginning of the cdash history for this machine.

This may be related to #2271  because the tests fail in the VMC acceptance check
```
/scratch/pk7/QMCPACK_CI_BUILDS_DO_NOT_REMOVE/qmcpack/src/QMCDrivers/tests/test_vmc_driver.cpp:121: FAILED:
  REQUIRE( ar == Approx(1.0) )
with expansion:
  0.0 == Approx( 1.0 )
```

**To Reproduce**
1. release version or git commit hash being built
2. cmake command
3. full program/test invocation command
4. additional steps

Build latest develop. Reported on cdash.

**Expected behavior**
Test should pass.

**System:**
 - system name [e.g. fusiont5, summit]
 - modules loaded [e.g. output of `module list`]
 - other systems where this is reproducible [e.g. "my laptop", "none"]
sulfur configured as per nightly script.

**Additional context**

ye-luo:
It is all green now.
https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=deterministic-unit_test_drivers&date=2020-06-18

Issue #2271, opened 2020-02-04
Wrong CloneManager::acceptRatio in certain builds


ye-luo:
The summation loop in CloneManager::acceptRatio seems causing problems.

On my workstation with Xeon Skylake and Ubuntu 18.04, if I build with GCC 6-8 (beyond 8 not checked), test_drivers fails. The default gcc option is -march=native. If I force -march=skylake, it passes. Instead if I force -march=skylake-avx512, it fails. If I change -O3 to -O0, it always pass.

Intel compiler works fine on this node.

On a server with Xeon Skylake and RHEL 7.5, gcc 6-9 are all fine.

So I have no conclusion where the issue is and how to solve it. Leave this issue open to collect more evidence.

ye-luo:
Workaround applied. Won't fix.

PR #2489, opened 2020-05-28
Workaround for a potential GCC bug

ye-luo:
## Proposed changes
For #2271 and #2487
To me it seems like a GCC bug related to AVX512. Add a sentinel code which prints out values if doesn't match.
So far the issue is only in acceptance ratio calculation. So I do not force a stop but just overwrite the wrong value and move on.
acceptRatio is only called per block. Performance hit should be negligible.

- Other (please describe): compiler bug workaround

- No

bora

## Checklist


prckent:
We'll have to think + discuss more before any merge, but this is successfully triggered (gcc9+complex+mkl+mpi)
```
WARNING a potential compiler bug detected! Overwriting nAcceptTot wrong value 0 with correct value 192. Overwriting nRejectTot wrong value 1644167168 with correct value 0.

```

PR #2492, opened 2020-05-28
Adjust RNG life cycle in QMCDriverNew

markdewing:
## Proposed changes
For optimization (and other cases where the driver is reused), returning the RNG pointers to global control in the destructor is not the correct lifecycle.

Add 'endSection' to QMCDriverInterface to allow drivers to release resources, etc, that have been allocated or set up in earlier functions (most likely in 'put' or 'process')
The QMCMain::runQMC routine contains the sequence of calls that create the driver, set it up, run it, etc.

Increase the size of the RngCompatibility vector to the number of threads in order to use it with old drivers (for the optimizer - to mix the new driver to generate the samples with the old driver to evaluate the samples

- Preemptive bugfix for a new feature

- No

laptop

## Checklist


PDoakORNL:
I would like to review this and will start on it tomorrow.

markdewing:
After a conversation with @ye-luo and @PDoakORNL yesterday, I see this needs a different approach.

PR #2493, opened 2020-05-28
Collect samples in VMCBatched driver

markdewing:
## Proposed changes
Compute the samples to collect and associated quantities in calcSamples. The number of steps and the steps between samples might get adjusted to collect the required number of samples and to load balance the samples.

Because the number of steps may be adjusted, a method to set the max steps was added to QMCDriverInput.

Pass a SampleStack structure into the VMCBatched driver.

This branch was created from the branch in PR #2492 

- New feature

- No

laptop

## Checklist


PDoakORNL:
I would like to review this I can get started on it tomorrow.

markdewing:
The focus on backwards compatibility for computing the number of samples is not the right way forward.   As @ye-luo pointed out, the new code should use nblocks, nsteps, and nsubsteps to control the number of samples collect.    Backwards compatibility can be addressed later, if necessary.

PR #2494, opened 2020-05-28
[WIP] Add QMCOptimizeBatched driver

markdewing:
## Proposed changes

The QMCOptimize driver was copied to QMCOptimizeBatched and converted to work with batched VMC for the collection of samples.  This is intended to have the minimal number of changes needed to compile and run without crashing.
    
This code only works for one thread.  The old VMC driver collected samples on each thread.  The new VMC driver collects a single vector of samples.  A future commit will distribute the samples to the per-thread structure to enable threaded evaluation of the cost function.
    
There are a couple of TODO items in QMCOptimizeBatched::generateSamples where the calls on vmc engine for the new batched driver are commented out.  The effects of those calls should be investigated further to determine if they need replacement functionality.

This was branched from PR #2493 (and contains the changes for #2492).  The "WIP" part is added because it gets a little confusing to see all the changes from the previous PR's also included.

- New feature

- Yes

laptop.  The code builds and it runs one NiO test without crashing.

## Checklist

- Ye. This PR is up to date with current the current state of 'develop'

### DIFF
--- a/src/QMCApp/QMCMain.cpp
+++ b/src/QMCApp/QMCMain.cpp
@@ -585,6 +585,7 @@ bool QMCMain::runQMC(xmlNodePtr cur, bool reuse)
     qmc_driver->run();
     t1->stop();
     app_log() << "  QMC Execution time = " << std::setprecision(4) << qmcTimer.elapsed() << " secs" << std::endl;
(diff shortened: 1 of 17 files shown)


markdewing:
Superseded by #2497

PR #2501, opened 2020-06-01
Fix unit hamiltonians

ye-luo:
## Proposed changes
1. overlapping ions caused inf in Ewald.
2. ensure error message is sent through. Still need a fix a the upper level. No worse than the current situation.

- Bugfix

- No

My ideapad.

## Checklist


prckent:
Quick comment on the error messages/warnings:

I think it is helpful if these are easily greppable, so we should ensure to use a starting prefix or set of words, preferably all caps or something that will be seen in the output. I think e.g. A lower case sentence warning is unlikely to be seen due to the insufficiently friendly human output we have.

ye-luo:
So far the expectation capture add ERROR via app_error.
https://github.com/QMCPACK/qmcpack/blob/864f53d90bfaa2a0f0893f47e638794a6840f0f0/src/QMCApp/qmcapp.cpp#L260
Unfortunately it is printing using a sick pattern. Will revisit this after we settle the presentation.

PR #2502, opened 2020-06-01
Reorganize qmcbase CMake target and rename it to qmcparticle

ye-luo:
## Proposed changes
Simplely rearrange files. Move all the qmcbase target related files under Particle.
Rename qmcbase to qmcparticle, a more intuitive name.
Get these simple change in before trying more complicated cmake changes.

- Build related changes

- No

Bora

## Checklist


ye-luo:
We may do some cleaning in scalar_traits. What did you mean "Post C++11 they seem redundant"?
Currently the file holds very important qmcplusplus::conj function which workaround std::conj limitations.

PDoakORNL:
I don't know about about Particle as the root name but the hierarchy works better and the library should relate to the directory in some simple way.

Issue #2508, opened 2020-06-03
FDLR code and test needs removal


jtkrogel:
The FDLR tests are currently failing, as expected with the removal of AoS:

https://cdash.qmcpack.org/CDash/testDetails.php?test=8433087&build=120036
https://cdash.qmcpack.org/CDash/testDetails.php?test=8433101&build=120036

The failures don't occur until VMC, indicating that some FDLR code is still lurking in the wavefunction files.  This code should be removed, a guard placed to warn against attempted FDLR input, and the tests removed.

ye-luo:
The manual entry has not been removed.

PR #2512, opened 2020-06-03
Remove FDLR code (only compatible with removed AoS code)

jtkrogel:
## Proposed changes

Remove the FDLR wavefunction from the code as it is compatible only with the AoS code version, which was recently deprectated.  Also remove corresponding tests.  Addresses #2508.

Cleanup.

- No

Laptop.

## Checklist


ye-luo:
Okay to test

PR #2516, opened 2020-06-04
Checking number of kpoints for requested super twist

mcbennet:
## Proposed changes

The code currently checks each supertwist for the correct number of kpoints for tiling. In this PR, I propose that only the requested supertwist be sanity checked in this way.
Possibly, there are repercussions here that I am not seeing?

- Bugfix

- No

CADES/ORNL

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/EinsplineSetBuilderCommon.cpp
+++ b/src/QMCWaveFunctions/EinsplineSetBuilderCommon.cpp
@@ -539,39 +539,31 @@ void EinsplineSetBuilder::AnalyzeTwists2()
   // Now check to see that each supercell twist has the right twists
   // to tile the primitive cell orbitals.
   int numTwistsNeeded = std::abs(det(TileMatrix));
-  for (int si = 0; si < numSuperTwists; si++)
+  // First make sure we have enough points
+  if (superSets[TwistNum].size() != numTwistsNeeded)
   {
-    // First make sure we have enough points
-    if (superSets[si].size() != numTwistsNeeded)
+    char buf[1000];
+    snprintf(buf, 1000, "Super twist %d should own %d k-points, but owns %d.\n", TwistNum, numTwistsNeeded,
+             static_cast<int>(superSets[TwistNum].size()));
+    app_error() << buf;
+    APP_ABORT("EinsplineSetBuilder::AnalyzeTwists2");
+  }
+  // Now, make sure they are all distinct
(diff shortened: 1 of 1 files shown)


prckent:
These checks exist to stop users from using inconsistent input files.

Please can you explain how the current situation is a problem for you? Would the benefits of the weaker check outweigh the consequences of potentially letting more mistaken runs through?

mcbennet:
> Would the benefits of the weaker check outweigh the consequences of potentially letting more mistaken runs through?

I was mistaken before about the code exiting for a super twist that is not requested with `twistnum` -- it turns out that the code only exists if the correct number of kpoints is not present for the user's _requested_ twist.

As it turns out the code was exiting due to insufficient memory, however, when I initially glanced at the error file, the end of it showed lines like the following:
`ERROR Super twist 1 should own 4 k-points, but owns 1.`
Though I asked for `twistnum=0` for which I knew the orbital file had the relevant kpoints.
Since nothing was printed about the memory, I assumed this was the issue.

Given that the code allows me to get my run through after resolving my memory issue, then in my opinion there is no problem. That being said, perhaps it is worth considering whether or not the above case should be printed as a Warning rather than an Error.

prckent:
Do you know how your run ran out of memory? If it was on one of the big allocations (orbitals, determinant) then we might be missing an error check.

mcbennet:
> Do you know how your run ran out of memory? If it was on one of the big allocations (orbitals, determinant) then we might be missing an error check.

It was the determinant. No memory issues were seen for a smaller cell with the same orbitals, but after tiling the same orbitals to a larger cell I saw the break.

mcbennet:
Correction, I suspect it could have been the orbitals after tiling, though I am not sure the exact culprit. I can try to reproduce.

prckent:
If you still have the run, otherwise don't worry - if it comes up again we'll know it is important.

mcbennet:
Attaching .out and .err files. It appears the breakage occurred when generating the splines. However, there is no informative printout -- though the "error" from the super twists is visible.  

[opt.err.txt](https://github.com/QMCPACK/qmcpack/files/4735829/opt.err.txt)
[opt.out.txt](https://github.com/QMCPACK/qmcpack/files/4735830/opt.out.txt)

prckent:
There is also a mismatch in the twists reported.

I think we can close this for now. If is reoccurs for you or others we should investigate.

PR #2521, opened 2020-06-05
Fix segfault when not recomputing CI coefficients.

fdmalone:
## Proposed changes
Fixes a segfault which was only triggered on two-determinant trial. Also increase logging and flush to help find where code is crashing.

- Bugfix

- No

lassen cpu

## Checklist


src/AFQMC/Wavefunctions/WavefunctionFactory.cpp
```diff
@@ -1076,7 +1079,7 @@ void WavefunctionFactory::computeVariationalEnergyPHMSD(TaskGroup_& TG, Hamilton
     }
   }
   TG.Node().barrier();
-  if(TG.Node().root())
+  if(TG.Node().root() && recompute_ci)
     TG.Cores().all_reduce_in_place_n(to_address(H.origin()),H.num_elements(),std::plus<>());
   TG.Global().all_reduce_in_place_n(energy.origin(),2,std::plus<>());
   app_log() << " - Variational energy of trial wavefunction: " << std::setprecision(16) << energy[0] / energy[1] << "\n";
```
prckent, line 1085: put a std::endl here as well? why not replace all your "\n"s?

prckent: Line 1085

prckent:
OK to test

PR #2524, opened 2020-06-06
Move src/simd to src/Platforms/CPU/SIMD

ye-luo:
## Proposed changes
Move `src/simd` to `Platforms/CPU/SIMD` and introduce `platform_cpu` subtarget of `platform_device`.
`platform_cpu` will be expanded to include al CPU BLAS/LAPACK and vector math handling.
`platform_cpu` is different `platform_host`. The former is intended to using CPU as computational device. The later is intended for more OS related.
An empty container interface target has been introduced to make the current change build. It will be expanded to host all the container folders at the src level.

- Build related changes

- No

Bora
## Checklist


prckent:
This is a good direction. It also cleans up the top level directory.

PR #2527, opened 2020-06-07
Move BLAS/LAPACK/Vector math to Platforms/CPU

ye-luo:
## Proposed changes
Move BLAS/LAPACK/Vector math files to Platforms/CPU.

- No

Bora

## Checklist


src/QMCWaveFunctions/EinsplineSet.cpp
```diff
@@ -15,7 +15,7 @@
 //////////////////////////////////////////////////////////////////////////////////////
 
 
-#include <Numerics/e2iphi.h>
+#include <CPU/e2iphi.h>
```
prckent, line 18: "CPU/e2iphi.h"?

ye-luo: In the past, <> is used for non-QMCPACK code. "" is used for QMCPACK code. We include header files based on a relative path to src.
In the spirit of organizing olders into CMake targets, a header file from a different top level folder is considered external.

QMCWaveFunctions/EinsplineSet.cpp belongs to qmcwfs target while 
Platforms/CPU/e2iphi.h belongs to platform_device target.
So the inclusion is considered external. "-Isrc/Platforms" is passed by target dependency.
That is why <> is used and there is no Platforms on the path.

prckent: I suspect more cleanup is needed to satisfy this. No need to hold this up though.

PR #2529, opened 2020-06-08
Workaround Clang offload linker issue

ye-luo:
## Proposed changes
Use CMake object targets to workaround clang linker not being able to handle fat binary.
Due to the CMake design choice of object targets. Indirect object files are not propagated. So all the dependency must be explicitly listed. It is ugly but at least all protected by USE_OBJECT_TARGET.

I'm able to past most deterministic test using clang master built on Jun 4th.
1. deterministic-unit_test_new_drivers has some CUDA runtime error. To be investigated.
2. deterministic-diamondC_1x1x1_pp-vmc_sdj_kspace-1-1 failed due to missing evaluateRatios implementation in kspace Jastrow.

I'm able to run 1 batch of 16 walker and pass statistical test.
4 batch of 4 walker has some memory corruption which is observed in all compilers using LLVM libomptarget.

- Build related changes

- No

ryzen-box

## Checklist


ye-luo:
> Please add a comment at first use of USE_OBJECT_TARGET stating what it is for.

USE_OBJECT_TARGET was introduced earlier. We have the following in the top level CMakeLists.txt
```
OPTION(ENABLE_OFFLOAD "Enable OpenMP offload" OFF)
CMAKE_DEPENDENT_OPTION(USE_OBJECT_TARGET "USE CMake object target" OFF "ENABLE_OFFLOAD" OFF)
```
Is this what you were asking for?

PDoakORNL:
I'd say since this is new CMake knowledge to most maybe more like this at its first appearance in the make CMakeLists.txt

# Use CMake object targets to workaround clang linker not being able to handle fat binary.
# which blah blah... (linking PTX in as a backup with the assembled device I think)
# CMake does not propagate indirect object files by design
# So the dependency must be explicitly specified with USE_OBJECT_TARGET.

Clang might change this someday and then this won't be cryptic.

ye-luo:
@PDoakORNL thanks for the suggestion. Updated comments in CMakeLists.txt

prckent:
Thanks @PDoakORNL , @ye-luo . This is exactly what I was thinking.

PR #2531, opened 2020-06-09
Initial AFQMC HIP Port

fdmalone:
## Proposed changes

Initial port of AFQMC for AMD machines. I will open a separate PR for my cmake hacks to enable compilation. Surprisingly the unit tests mostly pass, only some missing lapack calls and a few sparse errors are causing issues. I haven't timed / run full simulations yet, but as it's a fairly big PR I think it's a good point to check in.

Some points:

- I'm currently using the ENABLE_HIP flag.
- There are some naming inconsistencies due to a mismatch between hipX and roc_X libraries. Here, I initially went with hipBlas etc due to porting ease, however, some reatures in rocBlas aren't yet available in hipBlas and there is no hipSolver so I had to use roc solver throughout. In addition, I had some issues with hiprand (header file location in my version which is a known issue) so I had to go with rocrand instead. This isn't such a big deal right now but we'll have to revisit in the future if/when these libaries get updated.

- New feature

- No

MI60

## Checklist


fdmalone:
I should say I tested on ROCM/3.3.0

prckent:
Why was it important to use roc/hiprand at all?

prckent:
Which card(s) have you been testing on?

fdmalone:
One of the cuda kernels involves generating normally distributed random numbers. @mmorale3 can comment more.

I've compiled / tested on the MI60 (https://hpc.llnl.gov/hardware/platforms/corona)

mmorale3:
We generate the random numbers on the GPU. AFQMC consumes several orders of magnitude more random numbers than real space (you need about ~5M-10M per walker per step, where M is the number of single particle orbitals). We have optimized everything else in the code to the point that generating them on the CPU and moving them over would become a significant expense.

prckent:
Test this please

prckent:
Since it doesn't disturb the rest of the code and we are very much in a learning mode, I am fine to merge this provided it passes the CI. 

Anyone else who plans to comment should indicate so ASAP.

ye-luo:
We have our first bite on HIP and portability and we will learn more. CI failure needs to be addressed before merging.

PDoakORNL:
I have a few quick comments please hold till end of day.

PDoakORNL:
It certainly a big step toward AMD support (I think).  I'm fine with this going and considering the cuda side of this is already part of the code base. Although this does copy paste that design and I'd like a real review of that some time.

So abbreviated comment:
The globals involved make me sad and they are breeding.  RAII treatment of the GPU resource would be better and then we wouldn't have to check if handles were initialized etc.

CUDA has pretty good documentation and I imagine HIP and ROC will too one day. 

But the device_pointer stuff going on here on the other hand doesn't and this PR reinforces my impression that general purpose libraries are being developed inside of AFQMC and that our application review and testing process isn't a great framework for doing that.

ye-luo:
Okay to test

prckent:
Noting my agreement with Peters comments.

PR #2533, opened 2020-06-11
Enable sample collection in VMCBatched driver

markdewing:
## Proposed changes

Updated version of #2493 

As pointed out by @ye-luo, we don't need to keep the specification for the number of samples backwards compatible (for now).  This simplifies the logic.

Now the number of samples is controlled by the number of blocks, number of steps, number of walkers, and the number of nodes.   The number of substeps can be used to adjust the number of steps between each sample.

- New feature

- No

Laptop

## Checklist


ye-luo:
Could you document added function and variable in VMCBatched.h? The rest looks good to me

PR #2536, opened 2020-06-13
Leverage CrayLinuxEnvironment provided by CMake

ye-luo:
## Proposed changes
Relying on $ENV{CRAYPE_VERSION} on Cray machine is error prone.
It doesn't allow us to built without CRAYPE impact on a Cray machine. With new changes in Cray Environment and Cray compiler becomes Clang derived. It gets hard to infer the real intention of a user. With this PR, it is required to specify -DCMAKE_SYSTEM_NAME=CrayLinuxEnvironment via cmake (supported since v3.5) if you are using Cray compiler wrapper. It is clean to set this cached variable via command line instead of inside the CMakeLists.txt after detecting CC. If a user uses CC without specifying CrayLinuxEnvironment, CMake stops and remind of adding it.

Additional changes.
1. Add a script for GPU offload builds on Tulip.
2. Restrict `build_*` gitignore to only top level and folder only.
3. Add OPENMP_NO_COMPLEX internal macro to protect std::complex functions in ompBLAS for early stage compilers not being able to handle std::complex in OpenMP offload region.

Q: @prckent I need to update the `installation` for Theta and Cori in our manual, rst or tex which one should I touch now? This part has been converged.

- Build related changes

- No

Bora, Ryzen-box, Tulip(Cray), Theta(Cray), Cori(Cray)

## Checklist


prckent:
@ye-luo You should modify the rst in docs
https://github.com/QMCPACK/qmcpack/blob/develop/docs/installation.rst

ye-luo:
> Please check and report that with this change the installation stills works at NERSC. Update the docs as appropriate.

Expected your request ;) Cori(Cray) was listed under "What systems has this change been tested on?"

All scripts and instructions updated.

ye-luo:
Test this please

prckent:
Did you consider adding a check to see if we might be on a Cray machine and giving a warning if CrayLinuxEnvironment is not set? It would need to not give false positives (which might be the answer to my question).

ye-luo:
Here is my understanding.
CrayLinuxEnvironment is not equal to Cray Compiler Wrapper (CC).
CrayLinuxEnvironment is more of telling cmake that it is cross-compiling.
Cray Compiler Wrapper (CC) is the compiler for cross compiling.

So when CC is used, CrayLinuxEnvironment is required. If user forgot to add CrayLinuxEnvironment, a fatal error will be printed by our cmake setting.

If a user passes another compiler with or without CrayLinuxEnvironment. It is not clear what the intention is. Compiling for the host or cross compiling without using CC? I think we don't need to worry much about power user.
Instead if a basic user is not using CC, a warning should help. There is no harm for us to do so. So I added a warning.

prckent:
We can't require "CC" == Cray compiler. Uppercase cc for c++ is not unheard of, so if I understand the logic correctly, the abort needs to be removed and just print a warning "If the CC compiler is a Cray compiler you should rerun CMake with...".

PR #2540, opened 2020-06-15
qmcfinitesize parse stat.h5 files

camelto2:
## Proposed changes

This adds functionality to the qmcfinitesize tool so that one can directly parse the data in the stat.h5 file. This removes the need for the user to separately post-process the stat.h5 files and pass the S(k) data as an ascii file, which was the only way to do it previously.

In the case of hdf5 parsing, since it includes block data in the stat.h5, this will first construct the fluctuating structure factor from the data in stat.h5 for each k and block. Then for each value of k, it will perform a block average using a simple equilibration metric to throw out unequilibrated S(k) data. The resulting equilibrated S(k) is then passed on for the FS corrections. 

- New feature

- No

Linux box, intel 19 compiler

## Checklist


src/QMCTools/QMCFiniteSize/SkParserHDF5.h
```diff
@@ -0,0 +1,24 @@
...
+#include "io/hdf_archive.h"
+#include <vector>
+
+namespace qmcplusplus
+{
+class SkParserHDF5 : public SkParserBase
+{
+public:
+  SkParserHDF5();
+  ~SkParserHDF5();
```
ye-luo, line 15: Please document this class.
No need of empty constructor and destructor.

src/QMCTools/qmcfinitesize.cpp
```diff
@@ -73,12 +73,18 @@ int main(int argc, char** argv)
       skparser = new SkParserScalarDat();
       skparser->parse(anxt);
     }
+    else if (a == "--hdf5")
+    {
+      skparser = new SkParserHDF5();
```
ye-luo, line 78: Please use unique_ptr and make_unique instead of raw pointer and new.

src/QMCTools/QMCFiniteSize/SkParserHDF5.h
camelto2: removed, and I documented the various other classes used in the qmcfinitesize tool

src/QMCTools/qmcfinitesize.cpp
camelto2: done

ye-luo:
Okay to test

src/QMCTools/QMCFiniteSize/FSUtilities.h
```diff
@@ -17,7 +17,9 @@ void get_gridinfo_from_posgrid(const std::vector<PosType>& posgridlist, //list o
                                RealType& dx,                            // the grid spacing
                                IndexType& Nx);                          // the number of grid points
 
-void getStats(const std::vector<RealType>& vals, RealType& avg, RealType& err);
+void getStats(const std::vector<RealType>& vals, RealType& avg, RealType& err, int start=0);
```
rcclay: Documentation needed here.

src/QMCTools/QMCFiniteSize/FSUtilities.h
```diff
@@ -17,7 +17,9 @@ void get_gridinfo_from_posgrid(const std::vector<PosType>& posgridlist, //list o
...
+
+int estimateEquilibration(const std::vector<RealType>& vals, RealType frac=0.75);
```
rcclay: And here.

src/QMCTools/QMCFiniteSize/SkParserHDF5.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include <cmath>
+
+namespace qmcplusplus
+{
+SkParserHDF5::SkParserHDF5(){};
+SkParserHDF5::~SkParserHDF5(){};
+
+void SkParserHDF5::parse(const string& fname)
+{
+  std::cout << "Here: " << fname << std::endl;
```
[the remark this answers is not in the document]
camelto2: Yep, this was removed already

src/QMCTools/QMCFiniteSize/FSUtilities.h
camelto2: done

src/QMCTools/QMCFiniteSize/FSUtilities.h
camelto2: done

PR #2541, opened 2020-06-16
Add optimizer with batched VMC

markdewing:
## Proposed changes
Integrates the gradient optimizers with the batched VMC for generating the samples.

Better version of #2497

The previous PR had some TODO items related to the function of some commented-out code.  It was determined that such code had no effect, and has been removed.

The lifetime of the RNG's was an issue in the previous PR, and has been fixed by adjusting the code to call 'process' only once after vmcEngine is created.

This code only works for one thread. The old VMC driver collected samples on each thread. The new VMC driver collects a single vector of samples. A future commit will distribute the samples to the per-thread structure to enable threaded evaluation of the cost function.

- New feature

- No

laptop

## Checklist


ye-luo:
I believe eventually we will have `QMCOptimizeBatched` derived from QMCDriverNew. But we should first get the current one in and working and then refactor. I'm OK with this PR.

src/QMCDrivers/WFOpt/QMCOptimizeBatched.h
```diff
@@ -0,0 +1,116 @@
...
+class MCPopulation;
+
+/** @ingroup QMCDrivers
+ * @brief Implements wave-function optimization
+ *
+ * Optimization by correlated sampling method with configurations
+ * generated from VMC.
+ */
+
+class QMCOptimizeBatched : public QMCDriver
```
PDoakORNL, line 45: QMCDriver will not be retained after the shift to unified drivers is done.  This should be refactored to not rely on inheriting from it. Considerable confusion will definitely result.

PR #2553, opened 2020-06-22
Workaround data race when offloading with Clang 

ye-luo:
## Proposed changes
1. A clang developer has identified the root cause of the data race and is working on it.
With a bit workaround, I'm able to run Clang offload with multiple host threads.
2. I also fixed a conflict between MASS and math header wrappers in Clang.
By testing MASS and MASSV separately in CMake, I made MASSV used by Clang.
3. CMake doesn't put optimization level on ppc64le and results in slow code. Fixed here.

- Bugfix
- Build related changes

- No

Ryzen-box, JLSE P9 node. Summit still has a very strange CUDA issue.

## Checklist


prckent:
Does this still work on Summit's current software stack with only 1 host thread? What is the "very strange CUDA issue"? Have you filed a ticket?

CMake/ClangCompilers.cmake
```diff
@@ -79,6 +79,13 @@ ELSEIF(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64" OR CMAKE_SYSTEM_PROCESSOR MATCHES
     SET(CMAKE_C_FLAGS     "${CMAKE_C_FLAGS} -mcpu=native")
     SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mcpu=native")
   endif() #(CMAKE_CXX_FLAGS MATCHES "-mcpu=" OR CMAKE_C_FLAGS MATCHES "-mcpu=")
+
+  IF(CMAKE_SYSTEM_PROCESSOR MATCHES "ppc64")
```
prckent, line 83: Add a note to indicate this is a workaround

ye-luo:
> Does this still work on Summit's current software stack with only 1 host thread? What is the "very strange CUDA issue"? Have you filed a ticket?

It is not related to threading. The error is 'invalid device ordinal happens' when the OpenMP library tries to create a context. Writing a ticket to OLCF is in the pipeline.

prckent:
What happens without this PR though? Has something changed on Summit? We need to be able to run on Summit.

ye-luo:
> What happens without this PR though? Has something changed on Summit? We need to be able to run on Summit.

XL has no problem. It is Clang. Not sure if something changed or Clang offload never fully works for us on Summit.

PR #2559, opened 2020-06-27
Remove Clang bug workaround and remove hard-coded CUDA_PROPAGATE_HOST_FLAGS for Clang

ye-luo:
## Proposed changes
1. Remove clang workaround since the compiler has been fixed 3 days ago.
2. Remove hard-coded CUDA_PROPAGATE_HOST_FLAGS=ON for Clang. The default value is changed to OFF regardless of compilers. Passing -ffast-math caused problem in the past at least for GNU.
3. In the old setting, using normal variable is problematic and causes conflict with the cached variable of the same name in FindCUDA. With this change, CUDA_PROPAGATE_HOST_FLAGS is always a cached variable. If a user is sure CUDA_PROPAGATE_HOST_FLAGS=ON is necessary and safe, just pass it to CMake and the effect is persistent. I actually need it OFF since my CUDA host compiler is often gcc not Clang. Propagating (clang offload) flags fail gcc. Current setting constantly set it back to ON.

- Build related changes

- No

Bora

## Checklist


prckent:
=> Put propagate flags in the Clang compiler file, or is this also no longer required?

ye-luo:
> => Put propagate flags in the Clang compiler file, or is this also no longer required?

CUDA_PROPAGATE_HOST_FLAGS is a cached variable default to ON inside findCUDA.

ye-luo:
Hold the merge. Let me look at the non-Clang case more carefully.

ye-luo:
Updated description. My change is good now.

PR #2565, opened 2020-07-01
Add tolerance option to check_traces.py

Hyeondeok-Shin:

## Proposed changes

 Current deterministic test for traces check is failing in mixed-precision build because current tolerance (1e-8) is too small to handle numeric sensitivity on mixed precision build.
 This PR is to test for mixed precision build with separated tolerance of 1e-6, which is small enough as much to work as deterministic test. 

- No

 Intel Xeon

## Checklist

### DIFF
--- a/tests/scripts/check_traces.py
+++ b/tests/scripts/check_traces.py
@@ -1274,9 +1274,9 @@ if __name__=='__main__':
                       default='0',
                       help='Exclude a number of DMC steps from being checked.  This option is temporary and will be removed once a bug in the DMC weights for the first step is fixed (default=%default).'
                       )
-    parser.add_option('--mixed_precision',dest='mixed_precision',
-                      action='store_true',default='False',
-                      help='Increase tolerance for mixed precision test (default=%default).'
+    parser.add_option('--tol',dest='tolerance',
+                      default='1e-8',
+                      help='Tolerance to check (default=%default).'
                       )
 
     opt,paths = parser.parse_args()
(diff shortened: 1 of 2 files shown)


prckent:
It is not obvious to me that this is the best route. It adds build type dependencies to the test scripts. Why not increase the allowable error bar in each of the mixed precision tests? This requires changes in more places but keeps the comparison explicit, and provided the test is defined near the full precision one, it will be easy to see how it differs. Did you consider this option?

Hyeondeok-Shin:
@prckent Do you mean separated tolerances for each of test for the mixed precision build? Actually, I didn't think about it because uniform 1e-6 tolerance seems enough to measure deterministic for me. If we need to see exact how it differ from the full precision I thought it would be enough to print actual difference whatever test failed or not. There is also another option increase tolerance as 1e-6 for both full and mixed precision build if we need to remove build type dependency in the test as much as possible. We need to hear from @jtkrogel about his preference to deal with the mixed precision.

prckent:
@Hyeondeok-Shin Yes, lets get @jtkrogel view. If he thinks that this PR is fine, we can merge. This is not a big change and we can revise in future should it be necessary.

jtkrogel:
All I would change is to swap out the added `--mixed_precision` flag for a `--tol` flag that defaults to 1e-8.  That way we just pass 1e-6 for mixed precision.

Hyeondeok-Shin:
Now flag is replaced to --tol and ready for review.

ye-luo:
Okay to test

PR #2567, opened 2020-07-01
Add linear optimizer version that uses VMCBatch

markdewing:
##  Proposed changes

Convert linear optimizer to use VMCBatch for generating samples.  Similar to #2541.

Fixes from #2562 are included where appropriate.

## What type(s) of changes does this code introduce

- New feature

- No

laptop

## Checklist


src/QMCDrivers/DriverTraits.h
```diff
@@ -18,7 +18,8 @@ enum class QMCRunType
   WF_TEST,
   VMC_BATCH,
   DMC_BATCH,
-  OPTIMIZE_BATCH
+  OPTIMIZE_BATCH,
```
ye-luo, line 21: Is OPTIMIZE_BATCH actually used?

ye-luo: I mean we use linear not opt. The same as we should only use linear_batch and there is no need to expose opt_batch even we need the class.

markdewing: For now, it seems necessary to distinguish between the batched and non-batched versions.   After the non-batched versions get removed, this enum could be simplified.    But I'm not sure if I understood your comment correctly?

src/QMCDrivers/QMCDriverFactory.cpp
```diff
@@ -92,7 +93,11 @@ QMCDriverFactory::DriverAssemblyState QMCDriverFactory::readSection(int curSerie
   if (curName != "qmc")
     qmc_mode = curName;
   int nchars = qmc_mode.size();
-  if (qmc_mode.find("linear") < nchars)
+  if (qmc_mode.find("linear_batch") < nchars)
+  {
+    das.new_run_type = QMCRunType::LINEAR_OPTIMIZE_BATCH;
```
ye-luo, line 98: I mean the enum is used to label drivers here. If we don't need "opt/opt_batch" but only "linear/linear_batch", no need to connect "opt/opt_batch". Anyway, should not block this PR.

PR #2570, opened 2020-07-02
Use RESOURCE_LOCK to run GPU performance tests exclusively

ye-luo:
## Proposed changes
Use RESOURCE_LOCK to run GPU performance tests exclusively at a time.
This feature is available since the minimal cmake version we support.

PS: minor update in Bora nightly script.

- Other (please describe): testing

- No

Bora

## Checklist


prckent:
Am I correct that with this fix we should be able to remove -DN_CONCURRENT_TESTS=1 from the GPU nightlies?

prckent:
Why not set for all GPU tests? Long term NiO is clearly too local a place to set this.

ye-luo:
> Am I correct that with this fix we should be able to remove -DN_CONCURRENT_TESTS=1 from the GPU nightlies?

It is intended for NiO performance tests only because of the memory contention and benchmark accuracy.
So it must be strictly 1 test at a time.
There is no reason to restrict small tests and N_CONCURRENT_TESTS=4 works fine on Bora.

prckent:
OK. We get crashes with N_CONCURRENT_TESTS>1 on sulfur and nitrogen and their V100 cards.

PR #2575, opened 2020-07-06
Fix crowds input tag and implement solid walker count input handling in batched drivers

PDoakORNL:
## Proposed changes

This PR addresses the issue in #8. 
### Walker counts can be specified for the unified drivers as follows:
* total_walkers are what we used to call target walkers for a dmc section or total walkers for a vmc section. Sometimes this is what you got from using just "walkers" as the parameter name.
* walkers_per_rank can be specified.  If you don't have total_walkers specified they are calculated as walkers_per_rank * ranks at runtime.
* crowds specifies the number of crowds per rank, if its specified you get that number of crowds per rank even if it is more than the number of walkers_per_rank.  If it is not specified you get num_crowds == max threads or num_crowds == walkers_per_rank whichever is smaller.
* If you specify no total, no per rank and no number of crowds.  total_walkers = num_ranks * max_threads (1 walkers per crowd)
* Conflicting walker count inputs result in a fatal runtime_error before the section starts.
* crowds can have no walkers. (empty ranks probably don't work with `WalkerControlBase`.

This calculation is now handled by the base `QMCDriverNew` class and the special determination of default walker counts for derived classes is removed.

Its an issue resolution and refactoring to the unified drivers only.  Legacy will not see a resolution to this issue.

- Documentation will be needed as part of the unified driver rollout.

- Yes but only for the unified/batched driver which is not used for production.

Leconte.

## Checklist


src/QMCDrivers/QMCDriverNew.cpp
```diff
@@ -485,37 +468,60 @@ std::ostream& operator<<(std::ostream& o_stream, const QMCDriverNew& qmcd)
...
-    if (awc.global_walkers % ranks)
+    awc.walkers_per_rank = fairDivide(required_total, num_ranks);
+    if (walkers_per_rank != 0)
     {
-      awc.walkers_per_rank = awc.global_walkers / ranks + 1;
-      awc.global_walkers   = awc.walkers_per_rank * ranks;
-      app_warning() << "TargetWalkers not divisible by number of ranks, TargetWalker increased to "
-                    << awc.global_walkers << '\n';
+      if (std::any_of(awc.walkers_per_rank.begin(), awc.walkers_per_rank.end(),
+                      [walkers_per_rank](IndexType num_walkers) { return num_walkers != walkers_per_rank; }))
```
ye-luo, line 486: We can simply check "required_total / walkers_per_rank * walkers_per_rank == required_total"?

src/QMCDrivers/QMCDriverNew.cpp
```diff
@@ -485,37 +468,60 @@ std::ostream& operator<<(std::ostream& o_stream, const QMCDriverNew& qmcd)
...
-  awc = this->calcDefaultLocalWalkers(awc);
-
-  if (awc.walkers_per_rank * ranks != awc.global_walkers)
+  // if we found a way to determine walkers per rank above
+  if (awc.walkers_per_rank.size() > 0)
   {
-    awc.global_walkers = awc.walkers_per_rank * ranks;
-    app_warning() << "Walkers per rank number of crowds, TargetWalker increased to " << awc.global_walkers << '\n';
+    if (num_crowds == 0)
+      num_crowds = std::min(static_cast<unsigned int>(awc.walkers_per_rank[rank_id]), Concurrency::maxThreads<>());
```
ye-luo, line 504: I prefer num_crowds not depend on walkers_per_rank[rank_id] but only input value or  Concurrency::maxThreads<>(). https://github.com/ye-luo/qmcpack/wiki/Population-size-input-handling#step-1-reduce-the-relation-map-by-first-handling-num_crowds
I may start the run with 6 walkers_per_rank but 7 crowds.

Could you follow step 1,2,3 order. The order of (a-d) in step 2 doesn't really matter.

src/QMCDrivers/QMCDriverNew.h
```diff
@@ -254,17 +277,15 @@ class QMCDriverNew : public QMCDriverInterface, public MPIObjectBase
...
   ///the number of saved samples
   IndexType target_samples_;
 
   /// the number of blocks between recomptePsi
   IndexType nBlocksBetweenRecompute;
 
-  IndexType walkers_per_rank_;
-  IndexType walkers_per_crowd_;
+  //IndexType walkers_per_rank_;
+  //IndexType walkers_per_crowd_;
```
ye-luo, line 288: Remove?

src/QMCDrivers/QMCDriverNew.h
```diff
@@ -382,12 +403,11 @@ class QMCDriverNew : public QMCDriverInterface, public MPIObjectBase
...
-  void makeLocalWalkers(int nwalkers, RealType reserve, const ParticleAttrib<TinyVector<QMCTraits::RealType, 3>>& positions);
+  void makeLocalWalkers(int nwalkers,
+                        RealType reserve,
+                        const ParticleAttrib<TinyVector<QMCTraits::RealType, 3>>& positions);
 
-  virtual AdjustedWalkerCounts calcDefaultLocalWalkers(QMCDriverNew::AdjustedWalkerCounts awc) const = 0;
-  int get_num_crowds() { return num_crowds_; }
-  void set_num_crowds(int num_crowds, const std::string& reason);
-  void set_walkers_per_rank(int walkers_per_rank, const std::string& reason);
+  // void set_walkers_per_rank(int walkers_per_rank, const std::string& reason);
```
ye-luo, line 410: Remove?

src/QMCDrivers/QMCDriverNew.h
```diff
@@ -420,6 +440,8 @@ class QMCDriverNew : public QMCDriverInterface, public MPIObjectBase
 
   friend class qmcplusplus::testing::VMCBatchedTest;
   friend class qmcplusplus::testing::DMCBatchedTest;
+  friend class qmcplusplus::testing::QMCDriverNewTestWrapper;
+  ;
```
ye-luo, line 444: Remove ;?

src/QMCDrivers/QMCDriverInput.cpp
```diff
@@ -21,6 +21,19 @@ namespace qmcplusplus
 {
 QMCDriverInput::QMCDriverInput(int qmc_section_count) : qmc_section_count_(qmc_section_count) {}
 
+void QMCDriverInput::checkNumCrowdsLTNumThreads(const int num_crowds)
```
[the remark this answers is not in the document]
PDoakORNL: I went back and forth with that.  I'll move it to QMCDriverNew

src/io/StlPrettyPrint.hpp
```diff
@@ -0,0 +1,32 @@
+//////////////////////////////////////////////////////////////////////////////////////
```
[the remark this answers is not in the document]
PDoakORNL: ok

src/QMCDrivers/QMCDriverNew.cpp
PDoakORNL: This doesn't prevent 6  walkers_per_rank and 7 crowds if you specify the num_crowds only if you want to control number of crowds using OMP_NUM_THREADS.  It does prevents the default case on the power 9 development node I use where not setting OMP_MAX_THREADS would result 6 walkers per rank and 176 crowds or with std::thread where the default max_threads is on order of 10k. Since I've done this many times I imagine users will run into this more than they want more crowds than the number of walkers per rank.

Yet it's an extra behavior...  I'll think about this as I take care of other changes.

prckent:
Quick comment: since this is developmental code that (re)implements required features I don't have strong opinions on the defaults, and I don't think we should spend much time on getting them "correct" now. Shortly we will want sensible defaults for when the user specifies an absolute minimum, such as total_walkers or total_walkers_per_mpi, and to not have to worry about other details. An actual concern is that there is a dedicated PR that adds tests of the behavior, so that we can't accidentally break Peter's work again. e.g. Run with 3 crowds specified in the input, parse the output to verify this is actually the case when run etc.

src/QMCDrivers/QMCDriverInput.cpp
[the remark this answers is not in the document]
PDoakORNL: done

src/io/StlPrettyPrint.hpp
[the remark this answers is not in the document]
PDoakORNL: done

src/QMCDrivers/QMCDriverNew.cpp
```diff
@@ -188,33 +181,23 @@ void QMCDriverNew::setStatus(const std::string& aname, const std::string& h5name
...
-  int num_threads(Concurrency::maxThreads<>());
-  if (num_crowds_ > num_threads)
-  {
-    std::stringstream error_msg;
-    error_msg << "Bad Input: num_crowds (" << qmcdriver_input_.get_num_crowds() << ") > num_threads (" << num_threads
-              << ")\n";
-    throw std::runtime_error(error_msg.str());
-  }
-}
+// void QMCDriverNew::set_num_crowds(int num_crowds, const std::string& reason)
```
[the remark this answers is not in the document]
PDoakORNL: done

src/QMCDrivers/QMCDriverNew.cpp
PDoakORNL: I'd prefer to leave this in unless it really prevents you from doing what you want to with the number of crowds. It prevents a node overloading loading behavior when every hardware thread is part of the OMP_NUM_THREADS default.

src/QMCDrivers/QMCDriverNew.h
PDoakORNL: Done

src/QMCDrivers/QMCDriverNew.h
PDoakORNL: Done

src/QMCDrivers/QMCDriverNew.h
PDoakORNL: Done

src/QMCDrivers/QMCDriverNew.cpp
ye-luo: The current logic is still over-complicated. As you pointed out, I cannot run walkers_per_rank = 6 and OMP_MAX_THREADS=7 with 7 crowds and allow the population to increase from 6 to 7 without a performance cliff which 6 crowds have.

PDoakORNL: I think you should specify num_crowds in that case. I'm not sold on openmp as the one true way, other threading systems do not use environment variables to configure the concurrency environment. Certainly our documentation should not privlege OMP_MAX_THREADS over num_crowds. But I will remove the complexity.

PDoakORNL: Done

src/QMCDrivers/QMCDriverNew.cpp
PDoakORNL: Makes sense to me. Done

src/QMCDrivers/QMCDriverNew.cpp
ye-luo: I meant Concurrency::maxThreads instead of exactly OMP_MAX_THREADS. For any change of the scheme. It is better to figure it out on paper like the wiki and then implemented it exactly. I will move the wiki to the manual since now it has been fully implemented in the batched driver.

ye-luo:
@PDoakORNL the syntax is broken.
I will have a second review later. Hopefully nothing major stands out.

PR #2578, opened 2020-07-09
Add separated tolerance for r^2 == dr^2 check for the mixed precision…

Hyeondeok-Shin:
## Proposed changes

 Currently wavefunction unit test is failing for mixed precision build on nitrogen and sulfur because double precision tolerance for r^2 == dr^2 check in test_hybridrep.cpp is too small for the mixed precision.
 This PR is to add separated tolerance for mixed precision build as the number between double and float. 

- Bugfix

- No

Intel Xeon

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/tests/test_hybridrep.cpp
+++ b/src/QMCWaveFunctions/tests/test_hybridrep.cpp
@@ -243,7 +243,11 @@ TEST_CASE("Hybridrep SPO from HDF diamond_2x1x1", "[wavefunction]")
   elec_.getDistTable(0).get_first_neighbor(0, r, dr, false);
   std::cout << std::setprecision(14) << "check r^2 against dr^2. "
             << "r = " << r << " dr = " << dr << std::endl;
+#if defined(MIXED_PRECISION)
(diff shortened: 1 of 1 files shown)


prckent:
If I am reading the failing tests correctly (e.g. clang10 complex mixed) we are getting "if (0<0)". Did you consider putting <= &/or using the single precision definition for epsilon?

ye-luo:
How about hold this PR for a moment.
The current error is "0.0f < 0.0"
How about print "std::abs(r * r - dot(dr, dr))" and "std::numeric_limits<double>::epsilon()"?

Hyeondeok-Shin:
@prckent I simply discussed with @ye-luo before making this PR and we agreed to use the number between single and double rather than the single precision. It is still hard to guess which number will be the best for nitrogen and sulfur so it seems the best to start with the number between single and double.

Putting <= instead of < sounds more safe to deal with the mixed precision. I'll change it.

Hyeondeok-Shin:
std::abs(r * r - dot(dr, dr)) is checked for nitrogen and sulfur. its value is larger than I expected but still between single and double precision. Now mixed-precision tolerance is adjusted for them and ready to review.

prckent:
Rather close to single precision epsilon, but I think we can run with it.

prckent:
OK to test

Issue #934, opened 2018-07-18
DMC densities have incorrect weight (CPU code)


prckent:
Per the report in #925 , the weight  in the first block of the DMC AoS code is incorrect. This is a definite bug and needs to be fixed in AoS and SoA. A similar fix may be needed in the CUDA code, but an additional bug is likely present there.

jtkrogel:
I wonder if this is related to the remaining first step weights issue in #789.  See the last "Trace of trial energy" plot there.  The spikes right at the beginning of each DMC could relate to improper handling of the weights as is seen in the density estimator.

jtkrogel:
In regards to what might be special about the first DMC block, I noticed the following during some data analysis.

The number of walkers fluctuates little in the first block, and much more in other blocks:
![nwalkers_trace](https://user-images.githubusercontent.com/24547905/54221741-7b0c5880-44ca-11e9-81de-26b6eb55a2aa.png)

The weight varies even less (almost none) than the walker count in the first block (look for the short horizontal line at the beginning of each series):
![weight_trace](https://user-images.githubusercontent.com/24547905/54221841-9f683500-44ca-11e9-9c4c-714c4b0b01d3.png)

The trial energy is absolutely bonkers for this block, likely reflecting the extreme population control needed to constrain the weights as above:
![trial_energy_trace](https://user-images.githubusercontent.com/24547905/54221891-b9097c80-44ca-11e9-91a6-3b0554988684.png)

Conclusion: extreme population control in the first block to blame for the incorrect weights in that block?  Thoughts @ye-luo?

PR #2579, opened 2020-07-09
EstimatorManager refactor and mpi population fix

PDoakORNL:
## Proposed changes

This was originally just a refactor partial rewrite of EstimatorManager to support correct averaging. 
However by the time it was ready changes that broke DMCBatched had been introduced into develop.
These are also address here, the amount of effort that will be required to pull these apart in this not yet production code is not worth it from my perspective.

- Bugfix: averaging is no longer done at multiple levels (this could result in errors due to variable number of walkers per crowd and rank.) In practice with a high walker count these errors should be very small (i.e. previous science does not include large errors from this.)
- Code style update (formatting, renaming) Since much of this was chasing subtle bugs whenever clarity would help, changes were made.
- Build related changes: Additional tests were added in appropriate files that are added to the build.
- For Debug builds:  a number of and `#ifndef NDEBUG`  sections were added to trap bad numerical situations and invalid states.

- No (or not to code that was actually working)

Leconte

## Checklist

### DIFF
--- a/src/Containers/OhmmsPETE/OhmmsVector.h
+++ b/src/Containers/OhmmsPETE/OhmmsVector.h
@@ -109,8 +109,12 @@ public:
   // Attach to pre-allocated memory
   inline void attachReference(T* ref, size_t n)
   {
-    if (nAllocated)
-      throw std::runtime_error("Pointer attaching is not allowed on Vector with allocated memory.");
+    if (nAllocated) {
+      free();
+      std::cerr << "Allocated OhmmsVector attachReference called.\n" << std::endl;
+      // Nice idea but "default" constructed WFC elements in the batched driver make this a mess.
+      //throw std::runtime_error("Pointer attaching is not allowed on Vector with allocated memory.");
+    }
     nLocal     = n;
     nAllocated = 0;
     X          = ref;
@@ -151,6 +155,9 @@ public:
   ///clear
   inline void clear() { nLocal = 0; }
 
+  ///zero
+  inline void zero() { std::fill_n(X, nAllocated, 0); }
+  
   ///free
   inline void free()
   {
--- a/src/Containers/OhmmsSoA/VectorSoaContainer.h
+++ b/src/Containers/OhmmsSoA/VectorSoaContainer.h
@@ -164,7 +164,11 @@ struct VectorSoaContainer
   __forceinline void attachReference(size_t n, size_t n_padded, T* ptr)
   {
     if (nAllocated)
-      throw std::runtime_error("Pointer attaching is not allowed on VectorSoaContainer with allocated memory.");
+    {
+      free();
+      std::cerr << "OhmmsVectorSoa attachReference called on previously allocated vector.\n" << std::endl;
+      // Nice idea but "default" constructed WFC elements in the batched driver make this a mess.
+    }
     nLocal  = n;
     nGhosts = n_padded;
     myData  = ptr;
--- a/src/Estimators/EstimatorManagerBase.cpp
+++ b/src/Estimators/EstimatorManagerBase.cpp
@@ -2,9 +2,10 @@
 // This file is distributed under the University of Illinois/NCSA Open Source License.
 // See LICENSE file in top directory for details.
 //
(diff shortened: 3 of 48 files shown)


prckent:
Thanks Peter! I think it important to stress that we believe that the previous averaging was incorrect, even if not obviously so and very subtly so under most circumstances. So this PR tackles an additional important bug besides the MPI.

jtkrogel:
Does this affect estimators measured with the non-batched drivers?

prckent:
@jtkrogel It is a good question whether and how this affects existing behavior or not. We definitely don't do sufficient MPI testing yet (historically this has not been a major source of bugs). 

How can we get better at DMC and observable testing?

jtkrogel:
We have two MPI DMC estimator tests that work:

 short-diamondC_1x1x1_pp-dmc-estimator-density-4-4-check
 short-diamondC_1x1x1_pp-dmc-estimator-spindensity-4-4-check

The full and partial sum tests (statistical, removes equilibration equilibration) all pass:

https://cdash.qmcpack.org/CDash/testDetails.php?test=8965672&build=127786

The tests also correctly detect that the first DMC block is wrong (see issue #934), and so the test as a whole fails:

```
  Fixed sum per block tests:
    d 0 4.000010681199691!=4.0
    u 0 4.000010681199691!=4.0
    some per block sums do not match the reference
    status of this test: fail
```

These tests should be run for the current PR and the test output manually checked to verify that the full and partial sum tests still pass with these code changes.

jtkrogel:
Actually, it would be ideal if Peter would just post here the full test output similar to what the cdash link above shows.

ye-luo:
I think it is better to split the EstimatorManager refactor with MPI fix. EstimatorManager can be independently validated with VMC.

jtkrogel:
Separated or not, each can affect estimators so I would like to see DMC tests of estimators for both.

ye-luo:
> Separated or not, each can affect estimators so I would like to see DMC tests of estimators for both.

We need EstimatorManager with VMC test validation.
Then Fix DMC population control in batched driver. I'm not confident with the current fix.
Then fix DMC estimator tests.

PDoakORNL:
This should be no change for legacy code.  I would rather not split this PR again.  The only reason the mpi update is with it is that the mpi was completely broken by other code added while I prepared the estimator manager refactor.  My code has been based on this change to estimator manager for the unified driver for quite some time.  If I split this again it must be accepted first. I will not restore the clearly incorrect averaging behavior of the legacy code to the unified driver again.  I'm not even sure I could easily replicate it without just rolling all this back and starting again.

prckent:
Someone needs to run the short tests on CPU and with legacy CUDA for this PR. Who has bandwidth? (I don't, at least not for a few days, and we should get this merged sooner if we can.)

PDoakORNL:
I was going to try to in a background thread but I owe other projects some time. I'd like to see it correct for develop first. I get a link error with the current develop and clang 7.9.0 need to figure out whether that is real or not. Then I will run it for this PR. Probably not today or tomorrow.

ye-luo:
Let us do the "split" in a less burden way. in QMCAPCK repo, make a new branch Staging_Fix based on the current develop. Then split the touch files into several topics and make PRs to Staging_Fix branch. It is not required to Staging_Fix to build when we are dealing with individual PRs. Once all these PRs are merged to Staging_Fix, we make PR Staging_Fix to develop at that point, CI is fully required.

PDoakORNL:
With i.e. clang 7.9.0 and the CPU build both the debug and release version of the dmc-estimator-density-4-4-check crash.

prckent:
That estimator currently runs reliably with gcc, Intel, Clang e.g. https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=short-diamondC_1x1x1_pp-dmc-estimator-density-4-4&date=2020-07-15

PDoakORNL:
I'm builidng a current gcc stack to build it and will go try it on x86 when I can.  But don't be alarmed and I'll be closing this PR soon to begin trying to PR it as "topics"

PR #2581, opened 2020-07-10
Fixes to sphinx build and some cleanup

jbatson5:
## Proposed changes

This PR cuts down some of the warnings returned when building for HTML (or LaTeX pdf) output. These also result in some minor cosmetic fixes. 

- No

## Checklist

- Yes/No. This PR is up to date with current the current state of 'develop'. Yes. 
- Yes/No. Code added or changed in the PR has been clang-formatted. N/A.
- Yes/No. This PR adds tests to cover any new code, or to catch a bug that is being fixed. N/A. 
- Yes/No. Documentation has been added (if appropriate). Yes.

### DIFF
--- a/docs/LCAO.rst
+++ b/docs/LCAO.rst
@@ -254,7 +254,7 @@ cell.
 
 This file contains information related to the trial wavefunction. It is identical to the input file from an OBC calculation to the exception of the following tags:
 
-*.wfj.xml specific tags:
+*\**.wfj.xml specific tags:
 
   +---------------+--------------+-------------+--------------------------------------------------------------------+
   | **tag**       | **tag type** | **default** | **description**                                                    |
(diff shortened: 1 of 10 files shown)


jbatson5:
@prckent Sadly, I wasn't able to do much with the duplicate citation warnings. It looks like, ironically, by trying to fix the earlier problem (i.e., making separate bibs for each rst file with citations), it's now a lot more difficult to employ the fixes here: https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html. Some combination of label and key prefixes might work, but I can't wrap my head around it as of now. There may just be too many instances of certain entries in multiple bibs to consolidate.

prckent:
LGTM.

The remaining warnings are all for duplicate labels. As we discussed before, I don't think we should worry about these. The duplicate references also occur  multiple times in the LaTeX bibliography. Since these are still correct and our primary documentation is the online version, again we choose to live with the situation.

prckent:
OK to test

PR #2582, opened 2020-07-10
Add flex version of evaluateDerivatives

markdewing:
## Proposed changes

Make it clear that this function evaluates derivatives of the variational parameters (vs. spatial derivatives) by adding 'Parameter' to the name.

This only implements the first level of the flex_* routine.  Deeper levels need to be implemented to run in nested parallel or with offload.  However, this will allow the optimizer code to change.
    
The original version works on one configuration and returns an array with one element per variational parameter (well, two arrays).   The flex version will evaluate several configurations, and needs to return an array of arrays.  Possible solutions are to use a Matrix, use nested std::vector, or create a new data structure.   We use the last approach by creating the RecordArray class.  It should provide some abstraction between the structure and the storage used.

There is an incidental change for the unit tests to make open boundary conditions work in complex builds. The unit test uses atomic orbitals, and fails with the complex build in EvalPeriodicImagePhaseFactors with the "Attempting to run PBC LCAO with no HDF5 support" error.
For actual runs, the complex build should not be used with LCAO and open boundary conditions (??).  If we want to enforce this, it would be nice to have an error message stating that.

- New feature

- No

laptop

## Checklist


prckent:
@anbenali Please can you study the state of LCAO with complex and create issues as needed?. There are previous reports of breakage. However I am not sure if this is new breakage or a gap in implementation.

src/Containers/MinimalContainers/RecordArray.hpp
```diff
@@ -0,0 +1,94 @@
...
+//
+// File developed by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
+//
+// File created by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+#ifndef QMCPLUSPLUS_RECORDARRAY_H
+#define QMCPLUSPLUS_RECORDARRAY_H
+
+#include <stdio.h>
```
ye-luo, line 15: Is this really needed? If yes, ue C++ header <cstdio> is better.

src/QMCWaveFunctions/TrialWaveFunction.h
```diff
@@ -318,6 +319,12 @@ class TrialWaveFunction : public MPIObjectBase
                            std::vector<ValueType>& dhpsioverpsi,
                            bool project = false);
 
+  void flex_evaluateParameterDerivatives(const RefVector<TrialWaveFunction>& wf_list,
```
ye-luo, line 322: Please make this static like other flex_ functions.

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
```diff
@@ -904,6 +904,13 @@ void LCAOrbitalBuilder::EvalPeriodicImagePhaseFactors(PosType SuperTwist,
 void LCAOrbitalBuilder::EvalPeriodicImagePhaseFactors(PosType SuperTwist,
                                                       std::vector<std::complex<RealType>>& LocPeriodicImagePhaseFactors)
 {
+  // For open boundary conditions
+  if (PBCImages[0] == 0 && PBCImages[1] == 0 && PBCImages[2] == 0) {
+    LocPeriodicImagePhaseFactors.resize(1);
+    LocPeriodicImagePhaseFactors[0] = 1.0;
+    return;
+  }
+
```
ye-luo, line 913: I feel better not using early return.
How about protect hdf5 read with
```
if  (!(PBCImages[0] == 0 && PBCImages[1] == 0 && PBCImages[2] == 0))
```

src/QMCWaveFunctions/tests/test_TrialWaveFunction.cpp
```diff
@@ -358,4 +359,167 @@ TEST_CASE("TrialWaveFunction_diamondC_1x1x1", "[wavefunction]")
 
 }
 
+void setup_He_wavefunction(Communicate* c, ParticleSet** elec_ptr, ParticleSet** ions_ptr, TrialWaveFunction** psi)
```
ye-luo, line 362: How about move this test to a separate file?
```
test_TrialWaveFunction_He.cpp
```

src/QMCWaveFunctions/tests/test_TrialWaveFunction.cpp
```diff
@@ -358,4 +359,167 @@ TEST_CASE("TrialWaveFunction_diamondC_1x1x1", "[wavefunction]")
...
+{
+  ParticleSet* elec = new ParticleSet;
```
ye-luo, line 364: Can we avoid using pointers in this file?

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
markdewing: I'm concerned that avoiding the early return is going to complicate the logic and make it harder to read.

src/QMCWaveFunctions/tests/test_TrialWaveFunction.cpp
markdewing: Fixed.    Used std::unique_ptr for WaveFunctionFactory, so at least now it will get deleted when the test ends.

src/Containers/MinimalContainers/RecordArray.hpp
markdewing: removed

src/QMCWaveFunctions/TrialWaveFunction.h
markdewing: Fixed.

src/QMCWaveFunctions/lcao/LCAOrbitalBuilder.cpp
ye-luo: To me adding early return creates two fully divergent code paths and complicates the logic. If we don't need to distinguish no-PBC and PBC, it is better not to do it. However, we need to add proper comment saying PBC case needs to read lattice vectors.

ye-luo: I will leave it to others to express their preferences.

markdewing: I worked through the TransX logic in the loop below the HDF file read and verified it gives the same result, and I found a way to add the extra logic.   I will update with the early return removed as soon as the local build works.

prckent:
@anbenali Comments?
@markdewing Make an issue for Anouar if we don't hear. Lets merge and continue.

PR #2586, opened 2020-07-17
Complete conversion of nexus documentation for rtd

jbatson5:
## Proposed changes

This PR completes the conversion of the Nexus guide for RTD output (this time in the correct directory). It also makes the highlighting "background color" grey rather than light green, per Jaron's request. I also tidied up a few typos and updated a few links. Devs may choose to customize conf.py and makefile. Both latexpdf and html outputs build without error. 

- No

## Checklist

- Yes/No. This PR is up to date with current the current state of 'develop'. Yes.
- Yes/No. Code added or changed in the PR has been clang-formatted. N/A. 
- Yes/No. This PR adds tests to cover any new code, or to catch a bug that is being fixed. N/A. 
- Yes/No. Documentation has been added (if appropriate). Yes.


prckent:
Confirming that I can build this documentation. LGTM. Thanks John!

There is one important issue that needs to be discussed:

This PR removes the autodoc configuration that was previously used to generate the developer documentation, some 20+ files in source. This is most evident in the Makefile change.

I was able to partially restore the developer documentation functionality but discovered a few more general issues that will need fixing before the developer section could be made available via readthedocs.

- I tried adjusting the makefile.  I had to modify the extensions line in conf.py `extensions = ["sphinxcontrib.bibtex", "sphinx_rtd_theme", "sphinx.ext.autodoc"]`
- This was only partially successful - I hit a docstring related AttributeError. This might be module version related, i.e. updates to Nexus are desired.
- For autodoc to run, all the required python modules must be available for import. For this to work in the readthedocs virtualenv, this requires a requirements.txt to exist. This would make installing Nexus easier in any case. By specifying specific module versions, the above problem might be bypassable, but upgrades to the code are preferable.

Therefore, before the developer documentation can be make available via readthedocs, at least a couple of updates would be required to Nexus. These look to be the kind of things that are either trivial (1h) or will take a whole day or two to sort out. I think it might be better to make this user documentation available now, i.e. merge the PR as is and setup readthedocs, and then work on expanding the functionality and any needed fixes/upgrades.

@jtkrogel Thoughts?

jtkrogel:
I'm in favor of merging this now and making improvements as needed.

Following the merge, perhaps I should simply add a requirements.txt and then we can see what remains after that.

prckent:
OK to test

prckent:
https://nexus-workflows.readthedocs.io/en/latest/

PR #2037, opened 2019-10-24
add more unit tests for ForceChiesa

Paul-St-Young:
After #2019 and #2033, the `ForceChiesaPBCAA` estimator now produces the correct VMC forces. Huge shout-out to @tiihonej and @rcclay for fixing these bugs! This PR turns the found bugs into unit tests to secure this bare force estimator.

For prosperity, I have included a test that shows the current results agree with those in PRB **93**, 035121 (2016) [rs1.25-i5.zip](https://github.com/QMCPACK/qmcpack/files/3768594/rs1.25-i5.zip)
f2-1023-fix-clone

@rcclay , does this fix #331?

PR #2588, opened 2020-07-17
Resurrect stress estimator

Paul-St-Young:
## Proposed changes
Resurrect stress estimator originally implemented by @rcclay 
I added a unit test and performed a regression integration test using the same example as the one in PR  #2037 
[f28g_rs1.25-i5.zip](https://github.com/QMCPACK/qmcpack/files/4940618/f28g_rs1.25-i5.zip)

I'll dig out the documentation and look into clang formatting. This requirement is new to me.

- New feature

- No

workstation with Xeon E5-2620 v2 and intel/19.0.4 compiler

## Checklist


src/QMCHamiltonians/StressPBC.cpp
```diff
@@ -0,0 +1,368 @@
...
+    }
+  }
+#endif
+
+  return res;
+}
+
+SymTensor<StressPBC::RealType, OHMMS_DIM> StressPBC::evaluateSR_AA(ParticleSet& P)
+{
+  const auto& d_aa = P.getDistTableSelf();
```
ye-luo, line 156: I think we can avoid adding this function in ParticleSet.
ee_table_index and ii_table_index are both available in this class. call getDistTable(id).
If StressPBC::evaluateSR_AA is called both for Ion and elec in this class, then you may modify the API to pass the index.

src/QMCHamiltonians/tests/test_stress.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  ions.resetGroups();
+
+  // The call to resetGroups is needed transfer the SpeciesSet
+  // settings to the ParticleSet
+  elec.resetGroups();
+
+  TrialWaveFunction psi(c);
+
+  LRCoulombSingleton::CoulombHandler = new EwaldHandler3D(ions);
```
ye-luo, line 94: Use unique_ptr

src/QMCHamiltonians/StressPBC.h
```diff
@@ -0,0 +1,111 @@
...
+  SymTensor<RealType, OHMMS_DIM> stress_eI_const;
+
+  // need wave function for kinetic stress
+  TrialWaveFunction& Psi;
+
+  ///source particle set
+  ParticleSet& PtclTarg;
+  ParticleSet& PtclA;
+  ///long-range Handler
+  LRHandlerType* AA;
```
ye-luo, line 37: Can we leverage unique_ptr or shared_ptr for all the raw pointers in the added class?

src/QMCHamiltonians/StressPBC.h
```diff
@@ -0,0 +1,111 @@
...
+  std::vector<RealType> Qat;
+  ///Zspec[spec] charge for the spec-th species of A
+  std::vector<RealType> Zspec;
+  ///Qspec[spec] charge for the spec-th species of B
+  std::vector<RealType> Qspec;
+  //Constructor
+  bool firstTimeStress;
+  StressPBC(ParticleSet& ions, ParticleSet& elns, TrialWaveFunction& Psi);
+
+  Return_t evaluate(ParticleSet& P);
```
ye-luo, line 68: Please use override keyword to mark all the functions overriding virtual functions. Once you introduce one, clang warns you all the rest. So check with clang is helpful.

src/QMCHamiltonians/StressPBC.cpp
```diff
@@ -0,0 +1,368 @@
...
+ 
+  return res;
+}
+
+SymTensor<StressPBC::RealType, OHMMS_DIM> StressPBC::evaluateSR_AB(ParticleSet& P)
+{
+  const auto& d_ab = P.getDistTable(ei_table_index);
+  SymTensor<RealType, OHMMS_DIM> res = 0.0;
+  //Loop over distinct eln-ion pairs
+#ifndef ENABLE_SOA
```
ye-luo, line 127: AoS has been deleted.

src/QMCHamiltonians/StressPBC.cpp
```diff
@@ -0,0 +1,349 @@
...
+      res += Zat[iat]*e*AA->evaluateSR_dstrain(drijs[iat], rijs[iat]);
+    }
+  }
+  return res;
+}
+
+SymTensor<StressPBC::RealType, OHMMS_DIM> StressPBC::evaluateSR_AA(ParticleSet& P)
+{ // !!!! HACK to obtain AA distance table using firstTimeStress
+  int itabSelf = ee_table_index;
+  if (firstTimeStress) itabSelf = ii_table_index;
```
ye-luo, line 143: It is better to pass the table index by an argument explicitly than relying on a state firstTimeStress.

src/QMCHamiltonians/StressPBC.cpp
Paul-St-Young: OK. Addressed.

src/QMCHamiltonians/tests/test_stress.cpp
Paul-St-Young: I am not familiar with unique_ptr. I tried `std::unique_ptr<LRCoulombSingleton::CoulombHandler>` and `dynamic_cast` the RHS. Neither compiled. Looking around the code base, I do not see an example for using unique_ptr with `new`. Could you point me to an example please?

src/QMCHamiltonians/StressPBC.h
Paul-St-Young: override keywords added

src/QMCHamiltonians/StressPBC.cpp
Paul-St-Young: AoS code deleted.

src/QMCHamiltonians/StressPBC.cpp
Paul-St-Young: OK. Addressed.

Paul-St-Young:
@rcclay is there documentation for the stress estimator hidden somewhere? I cannot find it in previous git versions.

prckent:
@Paul-St-Young I don't want to pile on, but have you thought about making a deterministic integration test for this? It is good to have the unit test, but integration tests are also very helpful. A deterministic one could be added to our continuous integration and protect the code from being accidentally broken. In view of the substantial refactoring that is ongoing, adding  deterministic tests is a key way we have to protect functionality. 

(Deterministic tests are currently 1 thread only, a few Monte Carlo steps only, and a fixed seed. We can discuss further if needed.)

src/QMCHamiltonians/tests/test_stress.cpp
ye-luo: I refactored some existing code in #2593. The change pattern should be quite the same.

rcclay:
General documentation?  No.  Pretty much everything on paper about this thing is in my thesis, and it's all equations.

prckent:
@rcclay Yes. Expand the unit test first.

src/QMCHamiltonians/StressPBC.cpp
prckent: @Paul-St-Young Probably the rationale is obvious, but the use of unnecessary use of state machines or reliance on particular state information has proven to be one of the major pain points with the existing code (understanding, maintenance, source of bugs, accessibility to new folks...). So  if you find yourself writing new code, please avoid relying on state wherever you can.

src/QMCHamiltonians/tests/test_stress.cpp
Paul-St-Young: @ye-luo Thanks a lot for #2593 . I updated this PR to use unique_ptr.

src/QMCHamiltonians/StressPBC.h
Paul-St-Young: Yes, addressed.

Paul-St-Young:
@rcclay I should mention that the off-diagonal components of the ion-ion stress is correct.
Mostly likely the Coulomb contribution to the off-diagonal stress is OK.

The current road block is testing the off-diagonal of the kinetic stress.
If I stretch the cell without changing the Bspline `SuperLattice`, then VMC diverges.

src/QMCHamiltonians/StressPBC.cpp
Paul-St-Young: @prckent good point. Will avoid the state machine as much as possible in the future.

Paul-St-Young:
@rcclay I added input documentation for the stress estimator.
Can you add in the theory part please?
Probably just copy and paste a few key excerpts from your thesis.
Hopefully you still have the .tex

Paul-St-Young:
I think this PR is ready to be merged now.
I hope to more unit tests and a deterministic test once we figure out how to the the kinetic components properly.
@prckent Is this OK with you?

Paul-St-Young:
@prckent sorry I missed the migration from tex to rst.
I hope the tex files will be deleted soon to avoid confusion in the future.

ye-luo:
Okay to test

prckent:
We have renamed the manual directory today. The announcement was previously made on the QMCPACK Google Groups. Sorry for the hassle.

PR #2589, opened 2020-07-18
Fix issue with multithreaded CI coefficient optimization

markdewing:
## Proposed changes
The `CI_Optimizable` variable was introduced in https://github.com/QMCPACK/qmcpack/pull/879 and
Commit: 0df840610785a92b8e9e91739ef80c993b61ddf6

However, it was not copied in `makeClone`.   As a consequence, multithreaded optimization runs would only compute parameter derivative information on the main thread.  Other threads would return zero.

This affects both derivative methods and the linear method (Hamiltonian/overlap matrices) when optimizing CI coefficients.

This bug must not drive the CI coefficient optimization completely off the rails, otherwise it would have been noticed before now.    If only one of the threads is actually accumulating data, then fewer samples are used than expected.   The other effect is that all the matrix entries for the CI coefficients are going to be smaller than they should be (the normalizing values are the full number of samples).   However they might still be roughly the same relative to one another - so the direction to move the parameters might be correct, but the step size would be smaller than it should be (this last effect is just speculation).

- Bugfix

- No

laptop

## Checklist


prckent:
😢 

Thanks Mark. 

We should discuss next week but clearly

- We need some tests to avoid this ever happening again. Since this is a threading issue, do you have any ideas for how to reliably catch it in the present design?
- If we didn't know already, this bug tells us that the existing cloning design and implementation is very difficult to work with...

ye-luo:
The makeClone function is intended to be the virtual copy construction and it should be as simple as just calling the copy constructor. We should implement proper copy constructor with value semantics instead of putting stuff in makeClone.
In the case here, the constructor is used instead of a proper copy constructor and thus the states have to be copied manually.

prckent:
I was thinking something similar. It would seem that if we use copy constructors for cloning, makeClone can be completely deleted. It will still be possible to have errors in the constructor, but at least the design will be clearer.

prckent:
Additionally, the naming is wrong. clones are identical. That is the dictionary definition. ours are not.

ye-luo:
Here it mean indistinguishable copies. If we make clones from 1 to N and then keep any one of them by destroying all the rest, there should be no side effect.

markdewing:
I have plans for some code that will let parameter gradients be tested.    Running those tests with multiple threads would have caught this issue.

PR #2592, opened 2020-07-20
Add code to help test parameter derivatives

markdewing:
## Proposed derivatives
The code compares analytic and finite difference derivatives.  It can also write the derivative values to a file compatible with qmca.
I started with the 'testDerivOptimization' file and removed code that did actual optimization.

- New feature

- No

laptop

## Checklist


src/Optimize/OptimizeBase.h
```diff
@@ -47,6 +47,8 @@ class CostFunctionBase
 
   virtual Return_t Params(int i) const = 0;
 
+  virtual std::string getName(int i) const = 0;
```
ye-luo, line 50: Try to be verbose, "getParamName"? can the return type "const std::string&"?

markdewing: I changed the name of the function to getParamName.   I don't think returning a std::string reference will be any better than the value, due to Return Value Optimization.

ye-luo: Even with RVO, there is one copy made while consistently using "const std::string&" since the initial return should ensure no copy. I will stop insist on this since this part is not performance critical.

PR #2597, opened 2020-07-22
Nexus: fix matplotlib import in qmca

jtkrogel:
## Proposed changes

Update the matplotlib import pattern in the `qmca` and `qmc-fit` executables.  The matplotlib library in the past sometimes required cycling through backend options to find a working solution.  This functionality has been subsumed within matplotlib itself since the Nexus imports were made, and the specialized imports are no longer necessary.  

This PR removes the additional backend search and instead of aborting when a matplotlib import error is caught, a warning is instead printed:  
```
Warning:
    Python library matplotlib raised uncaught exceptions during import.  Plots may or may not work.
```
This is motivated by experience that has shown that matplotlib may still successfully generate plots even when such exceptions are thrown.  The warning can be suppressed via an additional command line flag to `qmca`.  

- Bugfix

- No

Laptop

## Checklist

### DIFF
--- a/nexus/bin/qmc-fit
+++ b/nexus/bin/qmc-fit
@@ -17,17 +17,6 @@ except ImportError:
 #end try
 
 try:
-    import matplotlib
-    gui_envs = ['GTKAgg','TKAgg','Qt4Agg','WXAgg']
-    for gui in gui_envs:
-        try:
-            matplotlib.use(gui,warn=False, force=True)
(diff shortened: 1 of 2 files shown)


prckent:
Confirmed working with macports installed matplotlib. I was able to analyze the H2O molecule example and see the trace.

prckent:
OK to test

PR #2598, opened 2020-07-23
Rename LaTeX manual

prckent:
## Proposed changes

Rename the legacy manual LaTeX directory as legacy_manual to reduce chance of unintended edits. The current manual is in rst and in docs. Some of the original graphics files will need to be moved to docs/ after which the legacy_manual contents can be deleted.

- Bugfix

- No

None.

## Checklist


ye-luo:
I guess the auto generation script be broken for after this file name change.
should we remove https://docs.qmcpack.org/qmcpack_manual.pdf on the server?

prckent:
Server PDF is removed. Scripts are disabled. Website pages for both QMCPACK and Nexus point to readthedocs. Let me know if you spot a missed update.

PR #2601, opened 2020-07-24
Restore energy density estimator

jtkrogel:
## Proposed changes

Restore functioning of the energy density estimator.  With these updates, uniform and voronoi grids work in periodic boundary conditions, as confirmed by statistical tests as well as deterministic checksums (integral of each energy density sample must reproduce the scalar energy for each block).  Checks have been performed in both the single and multi-threaded cases.  Existing in-situ tests were invaluable for the restoration.

Some issues remain to be handled: 
* Adjustments need to be made to some of the statistical tests, which have mis-estimated errorbars (by a clean factor of two).  These were likely errors in test creation relating to walker count.  
* Systems with open boundary conditions have yet to be tested fully (later PR).
* Data routing issues via TraceManager remain.  An easy workaround is to switch traces on in the input file (without writing to disk).  Given relative priorities, this remnant issue is much lower priority (perhaps never in a PR).

- Bugfix

- No

Laptop

## Checklist


src/Particle/LongRange/StructFact.cpp
```diff
@@ -24,7 +24,7 @@ namespace qmcplusplus
 {
 //Constructor - pass arguments to KLists' constructor
 StructFact::StructFact(ParticleSet& P, RealType kc)
-    : DoUpdate(false), StorePerParticle(false), SuperCellEnum(SUPERCELL_BULK)
+    : DoUpdate(false), StorePerParticle(true), SuperCellEnum(SUPERCELL_BULK)
```
jtkrogel, line 27: @ye-luo What do you recommend to do with this?  Enabling this storage is necessary for streaming single particle energy information as needed for energy density and EKT.

ye-luo: Call ParticleSet::turnOnPerParticleSK() by the feature related consumer class constructor?

jtkrogel: I will give it a try.

jtkrogel: This just results in a segfault as if `StorePerParticle` is `false`.  Same result if called at beginning of each evaluate.

jtkrogel: Perhaps the resize is wrong?

jtkrogel: In `StructFact::resize`: `eikr_r.resize(nptcl, nkpts);`

jtkrogel: In `StructFact::turnOnStorePerParticle`: `eikr_r.resize(nptcl, KLists.numk);`

jtkrogel: Apparently not the problem.  Suggestions?

PDoakORNL: build debug
gdb your_unit_test

jtkrogel: Thanks.  I know exactly what is causing the segfault.

jtkrogel: The underlying issue (again) relates to state machines and order of operations, specifically the order of calls to `StructFact::resize` and `StructFact::turnOnStorePerParticle`.

jtkrogel: Found the issue.  See latest commit.

jtkrogel:
Ready now.

jtkrogel:
NB: this should go in ahead of #2599 in case of any interactions.

ye-luo:
Okay to test

ye-luo:
Could you fix `docs/readthedocs.org:nexus-workflows`?

jtkrogel:
What is the problem with it?

ye-luo:
> Could you fix `docs/readthedocs.org:nexus-workflows`?

CI was showing failure but it seems fine now.

PDoakORNL:
I'm not sure how there would be interactions between this and #2599 since the legacy estimators and drivers have not been altered. I could just as easily say #2599 should go in first because that work might be affected by #2601

PR #2602, opened 2020-07-25
Fix test_TrialWaveFunction_He unit test segfault

### LINKED ISSUE #2600: wavefunction deterministic unit test is broken
**Describe the bug**

SEGV in test_TrialWaveFunction_He.cpp in deterministic-unit_test_wavefunction with most compilers and builds (75% failure rate, https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=deterministic-unit_test_wavefunction&date=2020-07-24 )

e.g. https://cdash.qmcpack.org/CDash/testDetails.php?test=9113820&build=129979 shows failure in sulfur (Intel Xeon) and with Intel compiler

Since this missed our CI we can assume some illegal pointer, memory, C++ (etc.) usage that doesn't show on all platforms or all compilers.

**To Reproduce**

Run test on current develop.

**Expected behavior**

Test passes

**System:**

All of them

**Additional context**

markdewing:
## Proposed changes
Fixes #2600

The SPOSetBuilderFactory contains a static map that needs to be cleared between tests.

This commit also removes a potential lifetime issue with particle_set_map.

- Bugfix

- No

laptop, Intel 19.0 compiler

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/tests/test_TrialWaveFunction_He.cpp
+++ b/src/QMCWaveFunctions/tests/test_TrialWaveFunction_He.cpp
@@ -29,7 +29,7 @@ void setup_He_wavefunction(Communicate* c,
                            ParticleSet& elec,
                            ParticleSet& ions,
                            std::unique_ptr<WaveFunctionFactory>& wff,
-                           std::unique_ptr<WaveFunctionFactory::PtclPoolType>& particle_set_map)
+                           WaveFunctionFactory::PtclPoolType& particle_set_map)
 {
   std::vector<int> agroup(2);
   int nelec = 2;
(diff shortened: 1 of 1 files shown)


markdewing:
The failure would only reproduce if the two tests ran together, and in a particular order (--order flag to the unit tests can help with this).   That indicated it was something from one test interfering with another.
I vaguely remembered that something in some unit tests needed to be cleared out at the end of the tests, but I couldn't remember what it was, and didn't find it quickly looking through the tests.  If I had found it, the process would have been much easier.     As it was, I had to add print statements to track down the problem (the debug info wasn't very reliable - often saying <optimized out>, even for debug builds).

PR #2605, opened 2020-07-28
Fix energy density for periodic grids

jtkrogel:
## Proposed changes

Fix energy density estimator for periodic grids and add a feature to output ion energies separately as a point set, rather than including them directly in the histogramming grid as before.

- Bugfix
- New feature

- No

Laptop

## Checklist


ye-luo:
Okay to test

PR #2596, opened 2020-07-22
Add flex_evaluateDeltaLog

markdewing:
## Proposed changes 

Add flex_evaluateDeltaLog  as part of moving all the calls in QMCCostFunction::checkConfigurations to have flex versions.

Question: The flex_evaluateLog function has a special case for a single element wavefunction/particleset list.  Is this important?  The flex_evaluateDeltaLog code does not have this special case.

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- New feature

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
laptop

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- Yes  This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #2606, opened 2020-07-29
Add second flex_evaluateDeltaLog

markdewing:
## Proposed changes
This the the function to be called during correlated sampling.

Note that compared to evaluteDeltaLog, the dummy G and L are passed in as parameters rather than allocated in the function.  This pushes the temporary allocations to the caller of this function, and avoids performing memory allocation here.

- New feature

- Yes

laptop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/TrialWaveFunction.cpp
+++ b/src/QMCWaveFunctions/TrialWaveFunction.cpp
@@ -192,22 +192,7 @@ void TrialWaveFunction::recompute(ParticleSet& P)
   }
 }
 
-/** evaluate the log value of a many-body wave function
- * @param P input configuration containing N particles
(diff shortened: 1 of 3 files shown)


ye-luo:
This PR has the same title as #2596. Confused.

markdewing:
I change the title, so at least that is different.  Part of the issue is that there were two functions called 'evaluateDetaLog' and they were distinguished by their type signature.   One computes the fixed/varying values, and the other computes just the varying values.
I tried to fix this in the previous PR by add 'Setup' to the function name ("flex_evaluateDeltaLogSetup").   The PR name did not get changed to reflect this.

prckent:
Thanks for renaming the functions. Of course, functions with different functionality should have different names.

PR #2612, opened 2020-08-03
Add ion-ion forces to cep estimator in OBC

tiihonej:
## Proposed changes

Add ion-ion forces to the OBC version of the CCZ force estimator, for streamlined usage of the forces. This makes it compatible with the other force estimator. Also remove an obsolete variable.

- New feature

- No

## Checklist


tiihonej:
Modified an existing unit test to consider the added ion-ion terms

src/QMCHamiltonians/ForceCeperley.cpp
```diff
@@ -36,11 +36,30 @@ ForceCeperley::ForceCeperley(ParticleSet& ions, ParticleSet& elns)
...
-  forces_ShortRange.resize(Nnuc);
-  forces_ShortRange = 0.0;
+  ions.update();
+  evaluate_IonIon();
   ///////////////////////////////////////////////////////////////
 }
 
+void ForceCeperley::evaluate_IonIon()
+{
+  forces_IonIon = 0.0;
```
ye-luo, line 46: Please pass `forces_IonIon` as a function argument (reference). Avoid the same name as the class member.
Then we can mark this function const. This should give us cleaner code.

src/QMCHamiltonians/ForceCeperley.cpp
```diff
@@ -62,8 +81,11 @@ void ForceCeperley::InitMatrix()
 
 ForceCeperley::Return_t ForceCeperley::evaluate(ParticleSet& P)
 {
-  forces                               = forces_IonIon;
-  forces_ShortRange                    = forces_IonIon;
+  if (addionion == true)
+    forces = forces_IonIon;
+  else
+     forces = 0.0;
+  //forces                               = forces_IonIon;
```
ye-luo, line 88: Please remove commented line. One line above the indent is not correct. If you have clang-format around, could you run it on this file?

src/QMCHamiltonians/ForceCeperley.h
```diff
@@ -57,6 +56,8 @@ struct ForceCeperley : public OperatorBase, public ForceBase
 
   void resetTargetParticleSet(ParticleSet& P) {}
 
+  void evaluate_IonIon();
```
ye-luo, line 59: Could you document this function?

src/QMCHamiltonians/ForceCeperley.cpp
tiihonej: The current way is how it's done in the ForceChiesaPBCAA. I'd rather be consistent for now, and do a complete makeover of the force estimators later, if needed. Actually ForceBase would be the most ideal place to do the ion-ion terms, but it's currently missing. What do you think?

ye-luo: I don't think you can move it to ForceBase, PBC and non PBC are different.
I was not asking to move forces_IonIon out of the base.
I mean change the call from `evaluate_IonIon()` to `evaluate_IonIon(forces_IonIon)`.
define `evaluate_IonIon(ParticleSet::ParticlePos_t& forces)` instead of `evaluate_IonIon()`

tiihonej: Not quite as simply of course, but combining the two: non-PBC from ForceCeperley and PBC from ForceChiesaPBCAA, because the ion-ion terms should remain the same regardless of the estimator.
I understood what your meaning, but I was trying to be consistent with the current functions in ForceChiesaPBCAA. anyway, the direction is good, so I (hopefully) made the changes you requested.

src/QMCHamiltonians/ForceCeperley.h
```diff
@@ -57,6 +56,9 @@ struct ForceCeperley : public OperatorBase, public ForceBase
 
   void resetTargetParticleSet(ParticleSet& P) {}
 
+  // Compute ion-ion forces at construction to include in the total forces
+  const void evaluate_IonIon(ParticleSet::ParticlePos_t& forces);
```
ye-luo, line 60: I mean
```
void evaluate_IonIon(ParticleSet::ParticlePos_t& forces) const;
```

ye-luo:
Okay to test

PR #2625, opened 2020-08-10
Fix determinant tool for negative number

TApplencourt:
## Proposed changes

This PR fix the `determinant_tool` which analyses the determinants stored in HDF5 to print information about them.
This tool was broken for a negative number (and the bug was ofc in the only non-tested function...). 

In HDF5, for legacy reasons, we store the determinant as a list of signed integers in the HDF5. All the algorithms in this script assume unsigned integers, so we need to convert them prior to any analysis.  

The bugs were in the mismatch of data-type. Some of the value was converted to Python native data-type where others were still numpy data-type. This PR fixes this issue.  The only change is `s -> int(s)` in line 34 / 48.

- Bugfix

- No

Python3. Didn't add new dependency. 

## Checklist

-  This PR is up to date with current the current state of 'develop'

### DIFF
--- a/utils/determinants_tools.py
+++ b/utils/determinants_tools.py
@@ -30,8 +30,22 @@ from collections import namedtuple
 Determinants = namedtuple('Determinants', ['alpha', 'beta'])
 CSF = namedtuple('CSF', ['double', 'single'])
 
-def sanitize(det_spin, bit_kind_size):
-     return [s + (1 << bit_kind_size) if s < 0  else s for s in det_spin]
+def sanitize_numpy(det_spin, bit_kind_size):
+     '''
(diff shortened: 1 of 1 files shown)


ye-luo:
Okay to test

kgasperich:
The function was renamed to `sanitize_numpy`, so it should either keep the same name (`sanitize`) or later calls to `sanitize` should be changed to reflect the new name:
https://github.com/QMCPACK/qmcpack/blob/337e8f40aaf5d42a37b2de41fa635b9769bbb5de/utils/determinants_tools.py#L161-L162

TApplencourt:
You right! Fixed

prckent:
Test this please

ye-luo:
Test this please

PR #2628, opened 2020-08-13
Update AFQMC for HIP Port

fdmalone:
## Proposed changes

Fill in missing blanks for AFQMC HIP port. Only critical missing feature is missing hipblas_csrmm operations which there is a temporary workaround for.

- New feature

- No

amd MI60, rocm+gcc8.1.0

## Checklist


src/AFQMC/Numerics/detail/HIP/Kernels/dot_wabn.hip.cpp
```diff
@@ -9,12 +9,10 @@
...
 ////////////////////////////////////////////////////////////////////////////////
 
-
-#include<cassert>
 #include <complex>
-#include<hip/hip_runtime.h>
 #include <thrust/complex.h>
-#include<hip/hip_runtime.h>
+#include <hip/hip_runtime.h>
+#include "Platforms/CUDA_legacy/uninitialized_array.hpp"
```
ye-luo, line 15: Do not use this.
`#include <thrust/system/cuda/detail/core/util.h>`

ye-luo:
Okay to test

src/AFQMC/Numerics/detail/HIP/Kernels/dot_wabn.hip.cpp
fdmalone: Ok. Why is it defined in CUDA_legacy?

ye-luo: `uninitialized_array` was thought to be removed in CUDA by the CUDA_legacy code. So a copy of the file was put directly. But it is actually hidden instead of being removed. So in the new CUDA code, this header file copy is not used and the one from CUDA toolkit is used. ROCM has the same file.

fdmalone: Thanks!

prckent: CUDA legacy will be deleted.

prckent:
(Since we are developing and exploring, provided this does not cause trouble for the existing code and it is logically structured, not duplicative etc., then I am OK with it.)

src/AFQMC/Numerics/detail/HIP/Kernels/dot_wabn.hip.cpp
fdmalone: This header causes errors with the hip code. fatal error: 'cuda_occupancy.h' file not found. I can make the swap for the CUDA kernels only.

fdmalone:
We decided to isolate unitialized_array into external codes as the fix suggested does not work for HIP right now. This will happen separately. For the moment I just made the header swap for the CUDA kernels in AFQMC.

ye-luo:
Okay to test.

PR #2636, opened 2020-08-14
LCAO Spinor Builder

camelto2:
## Proposed changes

This PR adds the framework for having LCAO spinors in OBC, in particular for an interface to the DIRAC code. This adds a new LCAOSpinorBuilder, derived from the LCAOrbitalBuilder. It requires the spinors to be provided by an HDF5 file, and the spinor information is stored in KPTS_0 following the format from the OBC pyscf to hdf5 format. The key difference is that the spin up components of the spinor are stored in eigenset_0, and eigenset_0_imag, whereas the down components are stored in eigenset_1 and eigenset_1_imag. 

A unit test is added using a simple spinor and compares to reference values provided in the lcao_spinor_test.py script.

- New feature

- No

my workstation, gcc9.2.0

## Checklist


prckent:
Test this please

src/QMCWaveFunctions/LCAO/LCAOSpinorBuilder.h
```diff
@@ -0,0 +1,41 @@
...
+{
+public:
+  /** constructor
+     * \param els reference to the electrons
+     * \param ions reference to the ions
+     */
+  LCAOSpinorBuilder(ParticleSet& els, ParticleSet& ions, Communicate* comm, xmlNodePtr cur);
+  SPOSet* createSPOSetFromXML(xmlNodePtr cur);
+
+private:
```
rcclay, line 36: Docs

camelto2:
All of the failures seem to be git checkout issues

FATAL: Could not checkout 33c7a7c6b5ef052616e99bb2375aa38df5d62686
hudson.plugins.git.GitException: Command "git checkout -f 33c7a7c6b5ef052616e99bb2375aa38df5d62686" returned status code 128:
stdout: 
stderr: fatal: reference is not a tree: 33c7a7c6b5ef052616e99bb2375aa38df5d62686

camelto2:
forgot to add the check on the spin gradient to unit test. Will update with that and requested documentation from @rcclay

ye-luo:
Since you made LCAOSpinorBuilder derived from the LCAOrbitalBuilder, could you say a bit about what is reused from LCAOrbitalBuilder?

```
  //Sposet for the up and down channels of our spinors.
  std::shared_ptr<SPOSet> spo_up;
  std::shared_ptr<SPOSet> spo_dn;
```
are you intended to share an SPOSet between up and down?

camelto2:
> Since you made LCAOSpinorBuilder derived from the LCAOrbitalBuilder, could you say a bit about what is reused from LCAOrbitalBuilder?

@ye-luo So to build a SpinorSet, we need two independent SPOSets (different MO coefficients for both the spin up and spin down portions...similar to an unrestricted calcualtion). So for LCAO spinors, the LCAOSpinorBuilder needs to create two individual LCAOrbitalSets and register them to a SpinorSet. The main reason I derived LCAOSpinorBuilder from LCAOrbitalBuilder is due to the constructor, and the fact that there are member variables that they both need (like the ParticleSet, the basisSet, the Occupation list, etc).

In the LCAOrbitalBuidler constructor, it sets things like the h5_path, whether or not to transform to MultiQuinticSplines for the basis set, etc. Then at the end it builds the basis set from the hdf5 file. There are a few function calls associated with that, and stores that in a basis set. Then the LCAOrbitalBuilder registers that basis set with the LCAOrbitalSet constructor. So rather than reimplement the basis set building into a LCAOSpinorBuidler, I just derived from it so I get access to those functions and can then register that same basis set to both the up and down LCAOrbitalSets.

camelto2:
>   //Sposet for the up and down channels of our spinors.
>   std::shared_ptr<SPOSet> spo_up;
>   std::shared_ptr<SPOSet> spo_dn;
> ```
> 
> are you intended to share an SPOSet between up and down?

The reason I have std::shared_ptr<SPOSet> spo_up, spo_dn is because SpinorSet::set_spos needs std::shared_ptr<SPOSet> as arguments

prckent:
OK to test

ye-luo:
> > //Sposet for the up and down channels of our spinors.
> > std::shared_ptr spo_up;
> > std::shared_ptr spo_dn;
> > ```
> > 
> > are you intended to share an SPOSet between up and down?
> > ```
> 
> The reason I have std::shared_ptr spo_up, spo_dn is because SpinorSet::set_spos needs std::shared_ptr as arguments

I'm actually reviewing the SpinorSet and would like to know the motivation of `std::shared_ptr<SPOSet> spo_up` at the first place.
SpinorSet::set_spos having `std::shared_ptr<SPOSet>` is just a result of that.

camelto2:
I suppose it should be using unique_ptr, as SpinorSet should be the sole owner of the up and down pointers. I don’t think there is anything else that would need to have  ownership of the up and dn components. 
 
@rcclay, did you have a use case in mind for requiring shared_ptr in the SpinorSet design?

rcclay:
@ye-luo @camelto2 unique_ptr should be fine here.

camelto2:
is it preferred to go ahead and merge this and follow up with a separate PR modifying the SpinorSet and various builders to use unique_ptr, or should I just make the changes and add them to this PR?

src/QMCWaveFunctions/LCAO/LCAOSpinorBuilder.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include "Message/CommOperators.h"
+
+namespace qmcplusplus
+{
+LCAOSpinorBuilder::LCAOSpinorBuilder(ParticleSet& els, ParticleSet& ions, Communicate* comm, xmlNodePtr cur)
+    : LCAOrbitalBuilder(els, ions, comm, cur)
+{
+  if (h5_path == "")
+  {
+    APP_ABORT("LCAOSpinorBuilder only works with href");
```
ye-luo, line 26: use https://github.com/ye-luo/qmcpack/wiki/Error-handling
method 4. myComm is available in builders.

ye-luo:
I made a prototype in #2641

ye-luo:
#2641 has been merged. Please update this PR accordingly.

src/QMCWaveFunctions/LCAO/LCAOrbitalBuilder.h
```diff
@@ -96,5 +96,8 @@ class LCAOrbitalBuilder : public SPOSetBuilder
   void EvalPeriodicImagePhaseFactors(PosType SuperTwist,
                                      std::vector<std::complex<RealType>>& LocPeriodicImagePhaseFactors);
 };
+
+
+void readRealMatrixFromH5(hdf_archive& hin, const std::string& setname, Matrix<LCAOrbitalBuilder::RealType>& Creal);
```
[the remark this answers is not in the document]
ye-luo: Could you move this free function as a const member function of LCAOrbitalBuilder?
Better to be more restrictive.

Issue #2635, opened 2020-08-14
Change check_scalars.py input options


markdewing:
Currently a check_scalar.py invocation looks something like
```
check_scalars.py --le "-2.8 0.1" -p He_opt
```
The quantities to compare are hard-coded into the script.   Consequently, adding a new quantity requires changing check_scalars.py. 
As a minor issue, the argument to the comparison quantity takes two numbers separated with a space.  This is a little awkward with most shells, requiring quotes to escape the space.

As an alternate, what about a set of three options, `--name`, `--ref-value`, `--ref-error`?
The invocation would become
```
check_scalars.py --name LocalEnergy --ref-value -2.8 --ref-error 0.1
```

As for input validation, check_scalars issues a readable error message if the quantity is not found in the scalar.dat file.

There is a check_scalars feature that it can check multiple values on one command line.   This doesn't seem used much in the tests?   If it's important, it could replicated by allowing multiple instances of the name,ref-value,-ref-energy options.

The rationale for this change is testing parameter gradients.   Adding a bunch of parameter gradient names to check_scalars.py doesn't seem like a good solution.

This proposal is only to add these options to existing check_scalars functionality, not to change the existing hard-coded values.

jtkrogel:
This sounds reasonable to me.

prckent:
A good suggestion. Please be sure to return an error code if something goes wrong.

markdewing:
Implemented in #2640

PR #2640, opened 2020-08-17
Add check_scalars input options

markdewing:
## Proposed changes
Add --name, --ref-value, and --ref-error to check arbitrary named
columns in a scalar.dat file (versus adding the names to 'quantites'
list.

Implements the change suggested in #2635

This change only accepts a single value to test. If someone wants to test
multiple values in a single invocation, the 'action' value on add_option
could be set to 'append', and then the variables would become a list.

- New feature

- No

laptop

## Checklist


prckent:
OK to test

PR #2641, opened 2020-08-17
Change to unique_ptr in SpinorSet

ye-luo:
## Proposed changes
To make the ownership clean in SpinorSet.

- No

ryzen-box

## Checklist

### DIFF
--- a/src/QMCHamiltonians/tests/test_ecp.cpp
+++ b/src/QMCHamiltonians/tests/test_ecp.cpp
@@ -514,15 +514,15 @@ TEST_CASE("Evaluate_soecp", "[hamiltonian]")
   k2dn.resize(nelec);
   k2dn[0] = -dot(kdn[0], kdn[0]);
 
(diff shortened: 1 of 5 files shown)


rcclay:
@ye-luo on second thought, can you pass the unique_ptr in by value instead of by &&?

rcclay:
@ye-luo Non-critical but curious, and more out of my unfamiliarity with unique_ptr semantics.  Is there a reason you decided to pass the unique_ptr in by r-value reference instead of by value?  It seems like the latter would be more restrictive...

ye-luo:
> @ye-luo Non-critical but curious, and more out of my unfamiliarity with unique_ptr semantics. Is there a reason you decided to pass the unique_ptr in by r-value reference instead of by value? It seems like the latter would be more restrictive...

copy/assign by value is not allowed by unique_ptr. That is why move semantics are needed. Yes, the extra restriction is desired.

PR #2646, opened 2020-08-20
Initial implementation of batched cost function

markdewing:
## Proposed changes
Create versions of checkConfigurations and correlatedSampling that call batch code.

The code only works with one thread.  Some loops have been modified to be hardcoded to one thread, some have not.

The gradient and linear optimizers should work (in theory), as they only depend on quantities computed in checkConfigurations and correlatedSampling.  It's only been lightly tested on two examples.

Because this code creates vectors of ParticleSet, TrialWaveFunction, and QMCHamiltonian that are the length of the number of samples, memory usage may be an issue.   It is likely that a future change will need to process the samples in smaller batches.

There's a few other fixes
- set samples in vmcbatch correctly
- add destructor to SampleStack to free memory

- New feature

- No

laptop

## Checklist


prckent:
Test this please

prckent:
@PDoakORNL Are there "do not use" comments / can you put some "do not use" comments in clonemanager etc.?

markdewing:
I could close this PR and open another one that cleans up more of the legacy issues?

PDoakORNL:
That's up to you Mark.  If my concerns are addressed in a future PRs that's fine with me. I'm not sure whether folks going through a process like Mark is will see those comments since one might not even look into the headers for the dropped or refactored classes. I think it might be better if I write a quick document of which classes are out and if they have a replacement what it is.  Where should we put something like that now?

prckent:
Put the notes in the manual in the development section. (Since manual edits are easy and we have version control, it is a superior solution to the GitHub wiki.)

markdewing:
I'll add additional changes to this PR

markdewing:
All the legacy threading should be removed.  QMCCostFunctionBatched no longer inherits from CloneManager.

prckent:
This LGTM. @PDoakORNL OK to merge at this point or would you like to see this branch matured more?

PR #2651, opened 2020-08-24
Dirac cartesian basis ordering

camelto2:
## Proposed changes

This PR adds an additional AO basis ordering for cartesian basis functions. This follows the ordering for the DIRAC code, which will be necessary for spin-orbit testing for molecular systems. 

For reference, for cartesian gaussians the ordering is defined in Numerics/CartesianTensor.h and are hard coded in the Gamess ordering, which is 

x^L y^0 z^0
x^0 y^L z^0
x^0 y^0 z^L
x^{L-1} y^1 z^0
x^{L-1} y^0 z^1
x^1 y^{L-1} z^0
x^0 y^{L-1} z^1
...

For the DIRAC ordering, it instead uses
x^L y^0 z^0
x^{L-1} y^1 z^0
x^{L-1} y^0 z^1
x^{L-2} y^2 z^0
x^{L-2} y^1 z^1
x^{L-2} y^0 z^2
...,

This ordering is added in AOBasisBuilder::expandYlm(). Implementing the ordering here is much simpler than modifying the spinors themselves to be in gamess order. 

- New feature

- No

my workstation, gcc9.2

## Checklist

- No. This PR adds tests to cover any new code, or to catch a bug that is being fixed. 


rcclay:
OK to test

ye-luo:
Please add a unit test which validate the dirac specific code paths.

src/QMCWaveFunctions/LCAO/AOBasisBuilder.h
```diff
@@ -138,6 +139,16 @@ bool AOBasisBuilder<COT>::put(xmlNodePtr cur)
     addsignforM = 0;
   }
 
+  if (Morder == "Dirac")
+  {
+    expandlm    = DIRAC_CARTESIAN_EXPAND;
+    addsignforM = 0;
+    if (sph != "cartesian")
+    {
+      APP_ABORT(" Error: expandYlm='Dirac' only compatible with angular='cartesian'. Aborting\n");
```
ye-luo, line 148: Please take the chance to replace all the APP_ABORT in this file.

camelto2: done

camelto2:
> Please add a unit test which validate the dirac specific code paths.

Since this addition is only about the AO ordering, I added a cartesian gaussian unit test. It just creates a single orbital that is a linear combination of gaussians up to g functions, and compares the value of the orbital for both the normal gamess ordering and the new dirac ordering. 

I think this is sufficient since the gradients and laplacians of the gaussians are tested elsewhere, so this just makes sure that the ordering is correct. There is a reference python code added. 

This could also be added to in order to check the order of the spherical gaussians for various orderings as well

ye-luo:
> > Please add a unit test which validate the dirac specific code paths.
> 
> Since this addition is only about the AO ordering, I added a cartesian gaussian unit test. It just creates a single orbital that is a linear combination of gaussians up to g functions, and compares the value of the orbital for both the normal gamess ordering and the new dirac ordering.
> 
> I think this is sufficient since the gradients and laplacians of the gaussians are tested elsewhere, so this just makes sure that the ordering is correct. There is a reference python code added.
> 
> This could also be added to in order to check the order of the spherical gaussians for various orderings as well

You did exactly what we need for a unit test.

PR #2652, opened 2020-08-24
Include spins in walker buffers

camelto2:
## Proposed changes

This PR adds tracking of the electron spin variables across various walker related classes such as MCSample, MCPopulation, Walker, etc. This fixes a bug with SODMC due to load balancing with MPI when spins are necessary by adding spins to the walker buffers. Essentially, spins are added everywhere that the positions are in the walker buffers and makes sure spin is tracked when copying a walker, creating walkers, registering/copying from/updating a walker buffer. 
 
Some in-depth tests by @rcclay show that these changes make the DMC with spins agree with a single-core case where no MPI load balancing occur.

- Bugfix

- No

workstation

## Checklist


rcclay:
OK to test.

rcclay:
For what Cody was talking about.  The attached plots are DMC energy vs. spin mass for two wavefunctions:  1.) a mean field spinor calculation with the magnetization aligned along Z, denoted by "X-pol", and 2.) a mean field calculation with the magnetization aligned 45 degrees off the z-axis, denoted by "X-45deg" (this yields a nontrivial spinor).  We are doing DMC with 32 walkers on a single core, denoted by "sc-X", and with 4 mpi ranks, so ~8 walkers per core, denoted by "mpi-X".  For reference, we have a spin-polarized fixed-node DMC calculation denoted by "ref", shown as a horizontal line with error bars.  

**BEFORE BUFFER FIX**
image

**AFTER BUFFER FIX**
image

PR #2654, opened 2020-08-25
Fixing tests for HDF5, molecular tests and convert4qmc for UHF/UKS

anbenali:

## Proposed changes

This PR is to fix multiple small issues (incomplete implementations and miss-labeling of tests) 

- Bugfix (misslabeling of LiH dimer tests.. not dimers)
- New feature (UHF) 
- Other (please describe): Use of QP2 as a base code, implies using generated H5 file from QP2 (instead of txt file and convert to HDF5). Effectively has all codes use HDF5 except for Gamess. 
 

- No

Laptop / Theta

## Checklist


anbenali:
Waiting for qmc-ref test to run.

prckent:
OK to test

anbenali:
Added O single atom in UHF with triplet state to stress the UHF case more appropriately.

src/QMCTools/QMCGaussianParserBase.cpp
```diff
@@ -1925,11 +1939,11 @@ void QMCGaussianParserBase::dumpStdInputProd(const std::string& psi_tag, const s
         xmlAddChild(initopt, parameter(initopt, "warmupSteps", "2"));
         xmlAddChild(initopt, parameter(initopt, "timestep", "0.5"));
         xmlAddChild(initopt, parameter(initopt, "walkers", "1"));
-        xmlAddChild(initopt, parameter(initopt, "samples", "8000"));
+        xmlAddChild(initopt, parameter(initopt, "samples", "80000"));
```
ye-luo, line 1942: Do we really need to changes these parameters in the generated input files?

anbenali: yes... Unfortunately I find myself changing those all the time! In fact there are 2 type of users, the ones who use the files generated by convert4qmc (that would be only me for molecules) and the ones using nexus (the majority of people including me for solids) 

Nexus does not care about those. And for me, I use 80000 samples all the time as 8000 often is not enough.

prckent: No fixed sample count can apply to all systems, even all "small" ones, so I view this as further indication that "we are doing it wrong".  Until there are simpler inputs that specify target energy errors etc we will continue to have this problem.

I don't might what we merge here.

anbenali:
> If an old h5 file is used with this new code. What is the code behaviour?
> Is there any simple signature we can detect and ask user to either generate new files or use older releases?

@ye-luo : The code has backward compatibility. If old H5 is used the code will still deal with it. However, because the converter generates the new format, I had to update most of the H5 files in the test suite.  (the main difference is storing the MO coeffs in "Super_twist" instead of "KPTS_0" KPTS_0 did not make sens at all after finishing all implementations. 

Moreover, the previous behavior was the following: 
QP2 and Pyscf generate an hdf5 that is ompletely self contained. 
Gamess (and now all the batteries of codes we support through molden) generate a txt file then we make the hdf5. That HDF5 was made at the same time as the input files (xml) for qmcpack. a few sections already in the xml were then not added in the H5 file. Now the orb.h5 file generated by convert4qmc is identical to the pyscf and QP2 format and can be used to generate new inputs.

src/QMCTools/QMCGaussianParserBase.cpp
anbenali: Keep in mind that this is only generated with the flag "-production" as a way to provide a very "conservative" starting point. but I agree that having less parameters to control and a dynamic sample population that can increase or decrease with the complexity of the problem would be great.  the choice of "increasing" the samples is based solely on the type of calculations I have been running lately and obviously will not work for all class of problems

prckent:
Since these parameter changes are hidden behind the production/"I am Anouar" flag, no new users will be directly affected. OK to merge in my opinion.

anbenali:
From the manual: 

-production

Without this option, input files with standard optimization, VMC, and DMC blocks are generated. When the “-production” option is specified, an input file containing complex options that may be more suitable for large runs at HPC centers is generated. This option is for users who are already familiar with QMC and QMCPACK. We encourage feedback on the standard and production sample inputs.

ye-luo:
> > If an old h5 file is used with this new code. What is the code behaviour?
> > Is there any simple signature we can detect and ask user to either generate new files or use older releases?
> 
> @ye-luo : The code has backward compatibility. If old H5 is used the code will still deal with it. However, because the converter generates the new format, I had to update most of the H5 files in the test suite. (the main difference is storing the MO coeffs in "Super_twist" instead of "KPTS_0" KPTS_0 did not make sens at all after finishing all implementations.
> 
> Moreover, the previous behavior was the following:
> QP2 and Pyscf generate an hdf5 that is ompletely self contained.
> Gamess (and now all the batteries of codes we support through molden) generate a txt file then we make the hdf5. That HDF5 was made at the same time as the input files (xml) for qmcpack. a few sections already in the xml were then not added in the H5 file. Now the orb.h5 file generated by convert4qmc is identical to the pyscf and QP2 format and can be used to generate new inputs.

If the backward compatibility is supported, we need at least 1 test for it. Use any one of the small tests.

src/QMCTools/QMCGaussianParserBase.cpp
ye-luo: I don't mind this change. My overall thinking about such cases are
If some changes don't matter much, not changing is better than changing because any change has related questions like why? impact?

anbenali: ah.. I understand now. It is just  a bit annoying for me to always change the sample size but it is not the end of the world. This particular change is really not important to me, I added it since there was a bunch of small changes.. if you want to revert it it is fine with me.

anbenali:
> If the backward compatibility is supported, we need at least 1 test for it. Use any one of the small tests.

Added old short test.

ye-luo:
> > If the backward compatibility is supported, we need at least 1 test for it. Use any one of the small tests.
> 
> Added old short test.

Could you pick a small tests? FeCO6 has the fattest h5

anbenali:
@ye-luo : How about this?

ye-luo:
> @ye-luo : How about this?

It seems like an xml input not an h5 input. Could you double check?

anbenali:
> > @ye-luo : How about this?
> 
> It seems like an xml input not an h5 input. Could you double check?

... dumb me... fixed... with almost all symbolic links except for the hdf5 this way it is even less heavy.

PR #2657, opened 2020-08-25
Include ompReduction.h in SplineC2COMP.cpp

ye-luo:
## Proposed changes
ompReduction.h header was missing in SplineC2COMP.cpp. It doesn't matter for Clang which supports complex reduction with and without openmp user defined reduction (UDR).

In addition, this PR put the README.md aligned with installation.rst and update nightly test script on Bora.

- Bugfix

- Yes. I noticed IBM XL fails on complex build. With UDR, I got linker error. Without UDR, I got compiler ICE.

Summit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ make -j 8
  * Key QMC build options
 
 ```
-     QMC_CUDA            Enable CUDA and GPU acceleration (1:yes, 0:no)
+     QMC_CUDA            Enable legacy CUDA code path for NVIDIA GPU acceleration (1:yes, 0:no)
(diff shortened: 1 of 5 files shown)


prckent:
Does XL only give an error for offload complex builds or even for non-offload complex builds (cpu or cuda)?

ye-luo:
> Does XL only give an error for offload complex builds or even for non-offload complex builds (cpu or cuda)?

offload real build and CPU only builds are all OK. We never build the legacy CUDA code with XL.

prckent:
Other people have built the legacy CUDA with XL though. Fortunately we only need gcc to work in this case as you point out.

PR #2665, opened 2020-08-30
Symlink all the files from test folders to the build folder

ye-luo:
## Proposed changes
The scheme copying the test source folder to the build folder is suboptimal. First, the cmake step takes long on a slow filesystem.
Second, a lot of copied files are actually unused, specially qmc-ref.
In this PR, all the files from the test source folder will be symlinked. All the subdirectories are ignored.
If a needed file is outside the test source folder or in a subdirectory, making a soft link to the test source folder resolves the issue.

On my desktop with an SSD, the cmake time change is not noticeable but the whole build folder drop from 800MB to only 40MB.

Copying precisely only the needed files has been attempted but more complicated than I thought. We do know the exact input file but not other files still used in the calculation/check.

- Build related changes

- No

Ryzen-box

## Checklist

### DIFF
--- a/CMake/macros.cmake
+++ b/CMake/macros.cmake
@@ -22,14 +22,19 @@ FUNCTION( COPY_DIRECTORY SRC_DIR DST_DIR )
     EXECUTE_PROCESS( COMMAND ${CMAKE_COMMAND} -E copy_directory "${SRC_DIR}" "${DST_DIR}" )
 ENDFUNCTION()
 
-# Function to copy a directory using symlinks for the files. This saves storage
-# space with large test files.
+# Function to copy a directory using symlinks for the files to save storage space.
+# Subdirectories are ignored.
 # SRC_DIR must be an absolute path
 # The -s flag copies using symlinks
-# The -T ${DST_DIR} ensures the destination is copied as the directory, and not
-#  placed as a subdirectory if the destination already exists.
+# The -t ${DST_DIR} ensures the destination must be a directory
 FUNCTION( COPY_DIRECTORY_USING_SYMLINK SRC_DIR DST_DIR )
-    EXECUTE_PROCESS( COMMAND cp -as --remove-destination "${SRC_DIR}" -T "${DST_DIR}" )
+    FILE(MAKE_DIRECTORY "${DST_DIR}")
+    # Find all the files but not subdirectories
+    FILE(GLOB FILE_ONLY_NAMES LIST_DIRECTORIES FALSE "${SRC_DIR}/*")
+    FOREACH(F IN LISTS FILE_ONLY_NAMES)
+      #MESSAGE("Creating symlink from  ${F} to directory ${DST_DIR}")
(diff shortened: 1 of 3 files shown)


prckent:
Q. What happens if this function is rerun? Does FILE(MAKE_DIRECTORY "${DST_DIR}") need some protection against DST_DIR already existing, as was implicit in the old code?

(Unlikely to be able to test this today, someone else is welcome to check+review)

ye-luo:
> Q. What happens if this function is rerun? Does FILE(MAKE_DIRECTORY "${DST_DIR}") need some protection against DST_DIR already existing, as was implicit in the old code?
> 
> (Unlikely to be able to test this today, someone else is welcome to check+review)

There is no behaviour change in this regard. The folder and all its files are overwritten every time cmake is triggered. Overwriting is the desired behaviour. So no need of protection needed I think.

markdewing:
This also means the input files can be edited in either the build directory or the source directory, which I think was also one of the motivations for the change?

Otherwise, the choices when working on tests were:
1. Edit the source directory, rerun cmake setup, then rerun the tests
2. Edit in the build directory, rerun tests, but need to make sure changes get moved to source directory before cmake setup is run again.

CMake/macros.cmake
```diff
@@ -22,14 +22,19 @@ FUNCTION( COPY_DIRECTORY SRC_DIR DST_DIR )
...
-#  placed as a subdirectory if the destination already exists.
+# The -t ${DST_DIR} ensures the destination must be a directory
 FUNCTION( COPY_DIRECTORY_USING_SYMLINK SRC_DIR DST_DIR )
-    EXECUTE_PROCESS( COMMAND cp -as --remove-destination "${SRC_DIR}" -T "${DST_DIR}" )
+    FILE(MAKE_DIRECTORY "${DST_DIR}")
+    # Find all the files but not subdirectories
+    FILE(GLOB FILE_ONLY_NAMES LIST_DIRECTORIES FALSE "${SRC_DIR}/*")
+    FOREACH(F IN LISTS FILE_ONLY_NAMES)
+      #MESSAGE("Creating symlink from  ${F} to directory ${DST_DIR}")
+      EXECUTE_PROCESS( COMMAND cp -ds --remove-destination -t . "${F}" WORKING_DIRECTORY ${DST_DIR})
```
markdewing, line 36: Is there any difference between "-t ." with the working directory set to ${DST_DIR}
and "-t ${DST_DIR}" (and no working directory set)?    I suppose if DST_DIR is empty, the command might be interpreted differently?

ye-luo:
> This also means the input files can be edited in either the build directory or the source directory, which I think was also one of the motivations for the change?

Exactly. Any input change in the source folder will be reflect in the build folder. There is no need to copy back and forth between build folder and source folder.

CMake/macros.cmake
ye-luo:
```
yeluo@ryzen-box:~$ cp -ds --remove-destination -t temp abc.cpp
cp: temp/abc.cpp: can make relative symbolic links only in current directory
```

PR #2667, opened 2020-09-02
Fix accept ratio and block cpu time in batch vmc

markdewing:
## Proposed changes
The acceptance ratio does not appear to be correct in the output.   The larger the value of "steps" in the input, the smaller the acceptance ratio that is reported in the scalar.dat file.
Items in the estimators are divided by the block weight in `EstimatorManagerNew::makeBlockAverages`.   The acceptance ratio
should not be divided by the block weight before putting it into the estimator manager.
The block CPU time needs to protected from the "divide by block weight" by multiplying by the block weight before putting it into the estimator manager.

- Bugfix

- No

laptop

## Checklist

### DIFF
--- a/src/QMCDrivers/QMCDriverNew.cpp
+++ b/src/QMCDrivers/QMCDriverNew.cpp
@@ -548,7 +548,7 @@ void QMCDriverNew::endBlock()
   estimator_manager_->collectScalarEstimators(all_scalar_estimators);
   cpu_block_time /= crowds_.size();
 
-  estimator_manager_->stopBlockNew(total_accept_ratio, total_block_weight, cpu_block_time*total_block_weight);
+  estimator_manager_->stopBlockNew(total_accept_ratio, total_block_weight, cpu_block_time * total_block_weight);
 }
 
 } // namespace qmcplusplus
--- a/src/QMCDrivers/QMCDriverNew.cpp
+++ b/src/QMCDrivers/QMCDriverNew.cpp
@@ -545,11 +545,10 @@ void QMCDriverNew::endBlock()
     total_accept_ratio += crowd->get_accept_ratio() * crowd->get_estimator_manager_crowd().get_block_weight();
     cpu_block_time += crowd->get_estimator_manager_crowd().get_cpu_block_time();
   }
-  total_accept_ratio /= total_block_weight;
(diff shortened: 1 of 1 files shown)


src/QMCDrivers/QMCDriverNew.cpp
```diff
@@ -545,11 +545,10 @@ void QMCDriverNew::endBlock()
     total_accept_ratio += crowd->get_accept_ratio() * crowd->get_estimator_manager_crowd().get_block_weight();
     cpu_block_time += crowd->get_estimator_manager_crowd().get_cpu_block_time();
   }
-  total_accept_ratio /= total_block_weight;
```
ye-luo, line 548: It seems the old code is wrong in many regards.
1. blockweight != the number of walkers in a block.
2. energies and observables should be averaged by weight but acceptance ratio should be averaged by walker counts.
3. Block CPU time, should we even average? or just collect the rank 0 time?

markdewing: Or, for Block CPU time, take the max value over the crowds?

ye-luo: max value is desired but it may add extra synchronization.

prckent: A hazard of poorly defined terms!

* Acceptance ratio is computed over #walkers with no weight contribution.
* Blockweight is meant to be the total weight of walkers in the block. For a healthy load balanced run it will be similar to the total number of walkers.
* Block CPU is just a timer. I suggest we not worry about computing the max value for now and just report the average to avoid extra syncs. For a healthy run this is good enough.

PDoakORNL: The accept ratio is handled properly in #2672.  Statistics when walkers are transferred are still somewhat off.
BlockCPU is currently the total, could be average easily,

prckent: @PDoakORNL Thanks. If we merge this will the conflicts tolerable or would you prefer we not merge this?

PDoakORNL: Please don't merge this. It conflicts and the ratio normalization is wrong its better not in the history.

PR #2671, opened 2020-09-02
Fix link error in debug build with RunTimeManager

markdewing:
## Proposed changes
Add explicit template instantiations for RunTimeManager.

The release build works without these, but the debug build fails with a linker error.

- Bugfix

- No

laptop, gcc and clang

## Checklist

### DIFF
--- a/src/Utilities/RunTimeManager.cpp
+++ b/src/Utilities/RunTimeManager.cpp
@@ -22,6 +22,9 @@ namespace qmcplusplus
 {
 RunTimeManager<CPUClock> run_time_manager;
 
+template class RunTimeManager<CPUClock>;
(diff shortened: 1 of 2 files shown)


ye-luo:
Could you fix also the header file `extern template` lines? It seems RunTimeManager are repeated but RunTimeControl are missing.

markdewing:
It looks like we have no Debug builds in the CI or in the nightly testing?  That would have caught this issue.

markdewing:
The `extern template` lines have been added for RunTimeControl, and the ones for RunTimeManager have been moved to immediately follow the class, as seems to be the pattern.

prckent:
@markdewing Correct. There are no debug builds currently. What would be the most useful combination in your opinion?

PR #2674, opened 2020-09-02
Use parallel for in TasksOneToOneOPENMP and rename abstraction from Threading to Executor

ye-luo:
## Proposed changes
I'd like to change the TasksOneToOne from prescriptive to descriptive.
TasksOneToOne is used to run num_task concurrently regardless the hardware/thread details.

In the OpenMP case, it means running num_task with threads from a thread pool.
This is actually very close to std::async which doesn't specifying what the thread pool is.

This change allows me to reserve a larger thread pool than the num_task  for doing tasking later.
I have not thought much about restricting num_task <= thread_pool_size so far.

Rename the abstraction Threading to Executor.

- New feature

- No

ryzen-box

## Checklist

### DIFF
--- a/src/Concurrency/Info.hpp
+++ b/src/Concurrency/Info.hpp
@@ -2,13 +2,12 @@
 // This file is distributed under the University of Illinois/NCSA Open Source
 // License.  See LICENSE file in top directory for details.
 //
-// Copyright (c) 2019 QMCPACK developers.
+// Copyright (c) 2020 QMCPACK developers.
 //
-// File developed by:
-// Peter Doak, doakpw@ornl.gov, Oak Ridge National Lab
+// File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Lab
+//                    Ye Luo, yeluo@anl.gov, Argonne National Laboratory
 //
-// File created by:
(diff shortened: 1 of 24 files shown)


ye-luo:
I end up with renaming the abstraction to Executor which has the capability to run tasks in parallel.
Functions handled to Executor ends up like a (CUDA) kernel programming style.

ye-luo:
Probably need a better name than TasksOneToOne. ParallelExecutor ?

PDoakORNL:
I'm not sure whether this really harmonizes with the C++ executor concept as it currently stands... But it seems better to me.

ye-luo:
I think our abstraction is very lightweight and used in very few places. Adapting it to a different need should be very easy. The goal here is to consolidate our own abstraction.

ye-luo:
> I'm not sure whether this really harmonizes with the C++ executor concept as it currently stands... But it seems better to me.

Are you referring to the name "ParallelExecutor"? We could also consider ParallelLauncher or ParallelTaskLauncher.

ye-luo:
Done. changed to ParallelExecutor with extra note added about C++ executor.

PDoakORNL (approved):
This looks good to me.  Considerably more flexible now.

PR #2682, opened 2020-09-07
Add getApproximateEnergyVariance

ye-luo:
## Proposed changes
Add getApproximateEnergyVariance and remove getCurrentStatistics and getEnergyAndWeight.
"Approximate" and documentation makes it clear that it is not an average over weighted samples.
Before this change getCurrentStatistics only averages over the population of the current rank.
The statistical error on the variance is too large to be useful for safeguards.

- New feature

- No

Bora

## Checklist


prckent (commented):
Do you understand why we make use of an approximate energy / variance at all? Can you see the advantage over computing correct statistics? Obviously it was done historically, so this PR is an improvement, but approximate anything anywhere is a recipe for trouble.

ye-luo:
Since the old approximate function is intended for quick estimates of the energy and variance and print out to the standard output. It was never used in any of our statistical data files. I don't see any advantage over exact values and putting an update approximate function is only a temporal solution because I cannot quickly dig out the exact energy and variance from the estimator manager classes. I do feel the need to fix the estimator manager overall to separate the notion of weights and numbers of samples. Unfortunately it cannot be done quickly.

prckent:
Agreed.

I think in this case the best future for this function is its removal...

prckent (approved):

PR #2684, opened 2020-09-07
Fix converter uninitialized variables

ye-luo:
## Proposed changes
Recent change in convert4qmc surfaced a bug. It is caused by uninitialized variables.
I don't know why gcc didn't catch it by -Wmaybe-uninitialized -Wuninitialized

This PR also adds converter output printing to files.

- Bugfix

- No

Ye-Dell-Laptop

## Checklist


prckent (approved):
Excellent. We can review the results tomorrow.

(I agree with leaving the extra stdout/stderr logging in converter_test.py)

PR #2686, opened 2020-09-07
Enable Clang -Wall warning with a few exceptions

ye-luo:
## Proposed changes
Review and merge after #2685 
Instead off adding warnings one by one, add -Wall with a few exceptions  #2687 which should be gradually resolved.
The majority of warnings resolved in this PR are miss matching forward declaration struct vs class.

- Build related changes

- No

ryzen-box

## Checklist


prckent:
I think that this is a good approach, since it makes it explicit what the problems are.

prckent (approved):
Worth noting that many (all?) of the forward declared classes that are structs should really be classes with proper accessors etc.

ye-luo:
@mmorale3 @fdmalone we will need a fix in AFQMC on all the clang warning of std::move() temporal objects once this PR is merged.

Issue #2650, opened 2020-08-24
Scripts assume executable named python is available


prckent:
**Describe the bug**

Some of the testing and setup scripts assume an executable named python is available. This might not be the case even if python3 is installed, where there can be only an executable named python3. The current situation results in a scattering of errors since most of our scripting refers to python3. The easiest fix would be to refer to python3 consistently throughout.

e.g. tests/performance/NiO/sample/generate_all_variants.sh runs "python ../../adjust_qmcpack_input.py"
tests/converter/CMakeLists.txt:    ADD_TEST(NAME converter_${test_name} COMMAND python converter_test.py

**To Reproduce**

Configured and run the tests from develop.

**Expected behavior**
Success if python3 executable is available but python executable is not

**System:**
My laptop

**Additional context**

We could also error trap this situation.

PR #2688, opened 2020-09-08
Fix converter python

ye-luo:
## Proposed changes
On sulfur and nitrogen, python 3.8.5 is tricky. Also noticed that hard-code python picks up python2.7, use shebang instead.

- Bugfix

- No

laptop with spack

## Checklist


prckent:
Thanks. We need to write python3 everywhere &/or ensure to call scripts with the correct shebang. (#2650)

prckent (approved):

PR #2691, opened 2020-09-08
Clean up direct python invocation

### LINKED ISSUE #2650: Scripts assume executable named python is available

### LINKED ISSUE #2689: Potential python 2 mix
**Describe the bug**
Although we demand python3, I still found many direct invocation of "python" which could potentially point to python 2.
`grep -R "COMMAND python"` shows all the places which need to use shebang to enforce python3.
See #2688 as an example of fix.

found in
CMake/compareGCOV.cmake
tests/afqmc/workflow/CMakeLists.txt
tests/performance a few places

ye-luo:
## Proposed changes
Clean up direct python invocation
Fix #2689
Fix #2650 

- Bugfix

- No

Bora

## Checklist


prckent (approved):
Thanks.

This is all I recall seeing as well.

PR #2694, opened 2020-09-09
allow complex build for multideterminants with xml input

camelto2:

## Proposed changes
This is a simple fix to allow the multideterminant code to actually detect the CI coefficients from xlm input. Currently, all CI coefficients would be found as zero, because the ci coefficient was set before a put for the real and imaginary parts, so it was never correctly set. 

- Bugfix

- No

my workstation

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
+++ b/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
@@ -512,12 +512,14 @@ bool SlaterDetBuilder::putDeterminant(xmlNodePtr cur, int spin_group)
       if (useGPU == "yes")
       {
         app_log() << "  Using DiracDeterminantBatched with MatrixDelayedUpdateCUDA engine" << std::endl;
-        adet = new DiracDeterminantBatched<MatrixDelayedUpdateCUDA<QMCTraits::ValueType, QMCTraits::QTFull::ValueType>>(psi, firstIndex);
+        adet = new DiracDeterminantBatched<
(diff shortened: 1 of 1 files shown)


src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -1197,6 +1198,9 @@ bool SlaterDetBuilder::readDetList(xmlNodePtr cur,
         confAttrib.add(tag, "id");
         confAttrib.put(cur);
 
+#ifdef QMC_COMPLEX
+        ValueType ci(ci_real, ci_imag);
```
ye-luo, line 1202: I could not understand why this change makes a difference. Could you explain a bit more?

camelto2: inside the ifdef QMC_COMPLEX, ci_real and ci_imag aren't actually being updated until the confAttrib.put(cur) is called. So where the call was previously, the ci is being initialized as 0+i0 instead of the desired ci_real + i ci_imag. The ValueType ci(ci_real,ci_imag) has to happen after ci_real and ci_imag are actually read in

camelto2: Notice in the code above (around line 1000) when usingCSF is true, the correct behavior is obtained because ValueType ci(ci_real, ci_imag) is called after the put. This only effects the case when usingCSFs is false

ye-luo: I see it now. Thanks.

ye-luo (approved):

ye-luo:
Okay to test

PR #2697, opened 2020-09-10
No need of app_error().flush()

ye-luo:
## Proposed changes
1. error stream/file in C/C++ is unbuffered. No need to flush.
2. our app_error() always print "ERROR " as long as it is invoked. app_error().flush() triggered this and misled users.

- Bugfix
- No

Bora
## Checklist


PDoakORNL (approved):
LGTM

PR #2700, opened 2020-09-11
Change base of QMCOptimizeBatched to QMCDriverNew

markdewing:
List of changes
 * qmcdriver_input_ is part of QMCDriverNew, remove duplication
 * IsQMCDriver is used in old QMCDriver::putQMCInfo, but not in new
   driver
 * QMC_OPTIMIZE field in qmc_driver_mode is not used anywhere
 * NumOfVMCWalkers is only used to change the number of walkers in W,
   which is not used in new driver.
 * MyCounter - not used in batched VMC.  Was used in old VMC to
    distinguish the first run of a driver from subsequent runs when
    setting steps for sample collection.
 * the 'aroot' parameter to getConfigurations is not used, so it doesn't
   matter what we pass.  It is an overriden virtual function from
   QMCCostFunctionBase, which makes changing the signature a bit harder

Unfortunately, MCWalkerConfiguration needed to be keep, since it's
required to construct QMCCostFunctionBase.

Other optimization methods removed from factory.  Remaining methods are
 -  "test" - to output parameter derivatives, and compare analytic
        derivatives with finite difference derivatives
 -  "anneal" - gradient optimized using damped dynamics.  Left as an
        example.

- No

laptop

## Checklist


prckent (approved):
LGTM. Wait until Monday afternoon to allow for more comments &/or prior agreement before merging.

prckent:
No extra comments, so merging.

PR #2701, opened 2020-09-11
Use ParticleSet::flex_update in batched cost function

markdewing:
In the batched cost function, use ParticleSet::flex_update instead of calling 'update' on each particle set individually.

- No

laptop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
@@ -281,7 +281,6 @@ void QMCCostFunctionBatched::checkConfigurations()
     {
       ParticleSet* wRef = new ParticleSet(W);
       samples_.loadSample(wRef->R, iw);
-      wRef->update();
       p_ptr_list_[iw]  = wRef;
       wf_ptr_list_[iw] = Psi.makeClone(*wRef);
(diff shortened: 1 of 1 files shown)


src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -293,6 +292,8 @@ void QMCCostFunctionBatched::checkConfigurations()
     RefVector<ParticleSet> p_list        = convertPtrToRefVector(p_ptr_list_);
```
ye-luo, line 292: auto can make the code more brief.

markdewing: auto can also make the code harder to reason about types.  Maybe not so much in this case because of the function name.

markdewing: This would be better addressed in #2702, which touches those lines.

ye-luo: Exactly. Thanks to the function name. No ambiguity here. It is like using auto with make_unique.

markdewing: Changed in #2702

ye-luo (approved):

PDoakORNL (approved):
good conversation there

PR #2702, opened 2020-09-11
Remove some raw pointers from batched cost function

markdewing:
## Proposed changes
This PR follows #2701 

Remove some raw pointers from the code and replace with unique_ptr.

- No

laptop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
@@ -37,10 +37,7 @@ QMCCostFunctionBatched::QMCCostFunctionBatched(MCWalkerConfiguration& w,
 
 
 /** Clean up the vector */
-QMCCostFunctionBatched::~QMCCostFunctionBatched()
-{
-  delete_iter(RngSaved.begin(), RngSaved.end());
-}
+QMCCostFunctionBatched::~QMCCostFunctionBatched() { delete_iter(RngSaved.begin(), RngSaved.end()); }
 
 void QMCCostFunctionBatched::GradCost(std::vector<Return_rt>& PGradient,
                                       const std::vector<Return_rt>& PM,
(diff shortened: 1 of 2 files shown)


src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -281,17 +273,18 @@ void QMCCostFunctionBatched::checkConfigurations()
     {
```
PDoakORNL, line 273: Let's not use iw when it's not an index over walkers. I'm probably guilty of it as well but it might as well be i if it doesn't carry any information.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -281,17 +273,18 @@ void QMCCostFunctionBatched::checkConfigurations()
...
       ParticleSet* wRef = new ParticleSet(W);
       samples_.loadSample(wRef->R, iw);
-      wRef->update();
-      p_ptr_list_[iw]  = wRef;
-      wf_ptr_list_[iw] = Psi.makeClone(*wRef);
-      h_ptr_list_[iw]  = H.makeClone(*wRef, *wf_ptr_list_[iw]);
-      h0_ptr_list_[iw] = H_KE_Node->makeClone(*wRef, *wf_ptr_list_[iw]);
+      p_ptr_list_[iw].reset(wRef);
```
PDoakORNL, line 276: this conversion to unique_ptr looks ok to me.

PDoakORNL (approved):
This can go in as is but it would be nice to stop using iw as an index over not walkers. Although it is an index over walkers made to I think evaluate samples to calculate derivatives. They don't do any walking. What causes every sample to need it's own fat walker?

This is a topic for modernizing the Optimizers not this PR of course.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
markdewing: One of the next steps will be to operate on batches (rather than the full sample list).  The index variables can be changed then.

PR #2703, opened 2020-09-11
Nexus: fix twist averaging with single batch prefix

jtkrogel:
## Proposed changes

Fix twist averaging when analyzing only a single calculation with twist_info files.

- Bugfix

- No

Laptop

## Checklist


mcbennet:
For the case of a single batch prefix, I have checked between averaging by hand and 'qmca -a' with the changes here.
I also checked two different meshes for a single test system. For both meshes, I see agreement to all decimals between the two averages.

jtkrogel:
Ready for merge.

ye-luo (approved):

ye-luo:
Okay to test

Issue #2584, opened 2020-07-14
Batched driver error handling need to use myComm instead of throw


ye-luo:
**Describe the bug**
In #2575, error is handled by throw but myComm is accessible at the driver level. It is cleaner to use myComm->barrier_and_abort()

**Expected behavior**
Stop the code clean.

PR #2704, opened 2020-09-11
fix for #2584 Consistent fatal error handling in unified Drivers

PDoakORNL:

## Proposed changes

This address #2584
In the case of fatal errors that will occur on all ranks we can handle the mpi abort more cleanly.
This PR introduces a way to consistently handle fatal errors at runtime by throwing an exception but indicate that this is an error that will occur uniformly across ranks. It does not rely on the mpi communicator being available in the local scope. 
I've updated the unit test to check QMCDriverNew throws this exception when such an error is induced in QMCDriverNew.

this facility should be used in #2700 and for all such cases in the new drivers. 

- Bugfix
- New feature

- No

## Checklist


ye-luo (changes requested):
You assumed the local MPI scope is the same as the global MPI scope.
This is very dangerous.
Please catch at the first level where you have access to the communicator instead of the top.
UniformMPIError is very helpful. Please add the documentation that it must be capture at the inner most MPI scope otherwise it may hang the code.
Please remove the capture at qmcapp.

PDoakORNL:
So if this is thrown in a reduced MPI scope  but not caught until the global scope we have the nightmare hang. So at the qmcapp.cpp level we want to fall through to the abort without at the app scope.

I will also add caution about that to the class doc.

ye-luo:
> So the danger is if this is thrown in a reduced MPI scope but not caught until the global scope we have the nightmare hang. So at the qmcapp.cpp level we want to fall through to the abort without a barrier.
> 
> Will revise.

All the MPI implementations guarantee that an MPI_ABORT anywhere on any communicator will stop the whole run. Don't panic.

PDoakORNL:
I got that its why the std::exception catch in qmcapp.cpp is fine except for occasionally losing the message. But the catch(UniformMPIError) in qmcacpp is pretty likely to do that as soon as there is split a communicator.  Worse my doc example shows using the global comm which would do that even for a catch local with the split comm.

prckent:
A potential test of this would be an intentionally broken ensemble run.

src/QMCApp/QMCMain.cpp
```diff
@@ -551,56 +552,64 @@ bool QMCMain::runQMC(xmlNodePtr cur, bool reuse)
   std::string prev_config_file = last_driver ? last_driver->get_root_name() : "";
   bool append_run              = false;
 
-  if (!population_)
+  // Within this scope if you split myComm you must catch UniformMPIError for your split MPI comm scope.
+  try
   {
```
ye-luo, line 557: Is it possible to narrow this try region? I feel better to have it round where the throw may occur. Including the heavy run is less favourable to me.

src/QMCApp/QMCMain.cpp
```diff
@@ -551,56 +552,64 @@ bool QMCMain::runQMC(xmlNodePtr cur, bool reuse)
   std::string prev_config_file = last_driver ? last_driver->get_root_name() : "";
   bool append_run              = false;
 
-  if (!population_)
+  // Within this scope if you split myComm you must catch UniformMPIError for your split MPI comm scope.
```
ye-luo, line 555: Please move this message to UniformMPIError.
> you must catch UniformMPIError for your split MPI comm scope.

Better to say the exception must be caught at the first accessible communicator from the top of the call stack.

src/Message/UniformMPIError.h
```diff
@@ -0,0 +1,42 @@
...
+ *    my_local_comm->barrier_and_abort(ue.what());
+ *  }
+ *
+ *  insures that the error message actually makes it out on the head
+ *  rank before the MPI processes are aborted.
+ *  
+ *  If even one rank does not experience the issue you may hang the
+ *  job until the walltime expires.
+ */
+class UniformMPIError : public std::runtime_error
```
markdewing, line 33: I'm not really excited about the name, since it describes implementation details.     More precisely, it's the use of this class without creating a better named derived class.     The name makes me think there is something wrong related to MPI.

prckent: The problem here is that C++ and MPI have their own distinct error mechanisms.

I suggest UniformError for the name. When MPI is not available we can still safely call them.

Comments could be expanded to indicate that these functions are the preferred way to give errors since they are exception & MPI aware.

ye-luo: `UniformErrorForTopCommunicatorInCallStack`?

PDoakORNL: I think of the call stack going down like most debuggers I've used, so that name would be confusing to me. 

I think `UniformErrorInCommunicateScope` could work. Since Mark is right It does match the abstraction level of Communicate. Which I didn't initally notice since I have the bad habit of replacing Communicate with MPI mentally. 

This issue doesn't play out for thead level concurrency the same so I don't think `UniformError` is a good idea. For the thread level we already need to catch any exception thrown in a section run by a `ParallelExecutor` (PE) (at least for the std::thread and openmp implementations) similarly. Since if you don't catch the exception in the thread scope both of those "runtimes" call terminate, and then the rank crashes without calling MPI_Abort (I think). There is an implicit barrier at the close of the PE::operator(...) and after that barrier is passed an exception is thrown at the enclosing thread scope (generally the master thread) if any of the threads encountered and exception. That exception is caught by a catch that calls comm->abort(). 

We could easily write a slightly more complicated implementation for PE to handle uniform errors as a special case since we could pass out the exception information for every thread. This removes the burden from the developer of decided whether an exception would be uniform or not at the thread scope everywhere. At the communicate level without more complex comm code (more than the 10 lines or so it would be for PE) we don't have the information about other ranks exceptions, but the developer might have a good idea. The user experience is improved by seeing the error message everytime not just when rank 0 won the race.

I realize we want to treat the UniformErrorInCommunicateScope differently if thrown inside a PE, right now they would not result in same exception being thrown in the enclosing thread scope which is bad if you decide to say use paralellism parsing input.

src/QMCApp/QMCMain.cpp
PDoakORNL: I'm not sure I understand, do we have a compiler/platform combination that doesn't do 'zero cost' exceptions? There are a number of considerations about the scope at which you want to catch an exception, but I think it is more expensive to developers to catch them in more places than necessary. Ideally when we logically acquired a comm scope we would enter a scope like we do with a ParallelExecutor so the catch scope and the comm context are matched.  Since that is not the way we implement Communicate concurrency right now this looked to me to be as large as safe but inside of all comm->splits.

src/Message/UniformMPIError.h
ye-luo: We should not parsing input in any PE scope. The parsing of non-local move input in the batched driver results tons of printing messages. Even if we can stop printing on non-zero thread but I think the right fix is to parse the input outside PE and than handle an intermediate structure to inside PE. So we don't need to worry about inside PE case, we just don't support that trouble case at all.

markdewing:
The implementation of barrier_and_abort issues the message on rank 0, and then calls barrier and abort on all ranks.   Would it be sufficient to issue the message **and call abort** on rank 0, and then call barrier on all the other ranks?
```
barrier_and_abrt(msg) {
   if (!rank()) {
      std::cerr << msg << std::endl;
      Communicate::abort();
  } else {
     Communicate::barrier(); // Prevent other ranks from calling  abort first
 }
}
```

prckent:
Does MPI guarantee an abort will work if other tasks are in a barrier? Is the standard that tightly written and do the implementations follow it? Obvious risk here is that we hang a whole machine job.

src/QMCApp/QMCMain.cpp
ye-luo: I'm just concerned about UniformMPIError potentially being abused, if someone didn't diligently add proper catch and assumed this one is safe to do all the catch. There is no way to prevent this kind of abusing from high level and such abuse may cause the code to hang. That is why I'd like to be extremely cautious about using UniformMPIError for large region. It doesn't need to be as large as possible. If the catch is missed, the generic catch of runtime_error still works. Having it large instead is dangerous. In your case, protect process() is probably enough.

markdewing:
IMO, this is somewhat of a distraction from trying to get the batched drivers working.    Coming up with better error handling should take some analysis of the code and the creation of better categories - XML input errors, for instance.    Using APP_ABORT for now is very localized and easy to find and upgrade in the future (replace APP_ABORT with the appropriate throw).

ye-luo:
> The implementation of barrier_and_abort issues the message on rank 0, and then calls barrier and abort on all ranks. Would it be sufficient to issue the message **and call abort** on rank 0, and then call barrier on all the other ranks?
> 
> ```
> barrier_and_abrt(msg) {
>    if (!rank()) {
>       std::cerr << msg << std::endl;
>       Communicate::abort();
>   } else {
>      Communicate::barrier(); // Prevent other ranks from calling  abort first
>  }
> }
> ```

The key challenge is how to prevent incorrect use of barrier_and_abort()
1. good case. All the ranks in myComm hit barrier_and_abort. rank 0 prints.
2. bad case. Not all the ranks in myComm hit barrier_and_abort. If rank 0 hits, error is printed. The whole communicator hangs.

In @markdewing example,
1. good case, remains good. abort will stop whole comm regardless of what other ranks doing.
2. bad case not all the ranks hit barrier_and_abort. If rank 0 hits, error is printed, the whole communicator stops. If rank 0 doesn't hit, the whole communicator hangs.
we may get a chance to get the code stop even in bad case but we cannot tell of barrier_and_abort is abused or not.

@markdewing because this is a separate issue from the batched driver refactoring. It can be handled as deferred github issue just like the current PR.

src/Message/UniformCommunicateError.h
```diff
@@ -0,0 +1,47 @@
...
+ *    my_local_comm->barrier_and_abort(ue.what());
+ *  }
+ *
+ *  Which insures that the error message actually makes it out on the head
+ *  rank before the MPI processes are aborted.
+ *  
+ *  If even one member of the comm i.e. my_local_comm does not experience the issue
+ *  you will HANG the job until the walltime expires.
+ *  
+ *  If you split a communicator it is your responsibility to add a try catch block in your
```
ye-luo, line 34: I feel this is an unnecessary burden. `UniformCommunicateError` can be restricted to places where a communicator is not accessible and must be caught properly wherever a communicator is available. If we restrict the use case narrow to this. We don't need to worry about accidental hanging by mishandling of exceptions.

PDoakORNL:
I don't think my_comm_ should be available from some many places.  A particular communicator should have a knowable scope and it should be an argument to functions that may need comm. I have found very few places that it wasn't stored just so my_comm_ could be used "conveniently." But since my_comm_ can be split and many objects may live through a split communication context its not something that should be stored.

I don't think there should be any access to it in most places you would throw a UniformComunicateError. Also the above comment looks like I support the direct calling of `barrier_and_abort` at the location of the error. I don't. This returns us to some fatal errors are exceptions, some are Macros, some are calls to some global or nearly global pointer.

src/Message/UniformCommunicateError.h
PDoakORNL: It doesn't effect AFQMC since they don't use the error.  I'm happier with the status quo with std::run_time exception than assuming my_comm_ is in scope or having many try catch blocks that are just there because of fear.
If you want to put a split over existing code, or you want to use UniformCommunicateError under your split in new code you need to understand this.

ye-luo:
Every communicator should has a valid scope. That is why you don't see myComm available everywhere. So far most myComm are defined in builders and drivers where the communicator must be available to handle parallel I/O or algorithms.

I saw the source of problem of UniformComunicateError is that it doesn't carry info about the communicator. It is a more of a developer promise and this promise is only valid at the nearest accessible communicator. The exception design doesn't work well with MPI. I don't think we can solve it even the community cannot.

In your last debugger point, you can break on barrier_and_abort().

PDoakORNL:
What I found appealing about UniformCommunicateError is I don't have to have a my_comm_, if I'm not in a communicate context I don't have to treat it special at all. The scope that owns the Communicate handles its own "special" error handling needs, the throwing code is just giving it the detail encoded as a derived type that every call in logical concurrency is expected to throw as well.

ye-luo (approved):

PR #2707, opened 2020-09-12
Save the RNG for correlatedSampling during optimization

markdewing:
Create RNG objects to use in each copy of QMCHamiltonian.
Save the state so the RNG objects can be reset at the beginning of each call to correlatedSampling during optimization.

This will use the same RNG sequence for each sample when re-evaluating the energy.  This affects the non-local ECP evaluation.  The same rotated grid will be used for each sample.  Will this cause a problem?

- Bugfix

- No

laptop

## Checklist


prckent (approved):
LGTM. Developmental changes in the batched code.

PDoakORNL (approved):
LGTM

PR #2711, opened 2020-09-14
Uses batches in the batched cost function

markdewing:
Process the samples in smaller batches while computing the cost function in optimization.  This saves memory, and controls the vector sizes passed to the flex_* routines.

The `opt_batch_size_` parameter is set to one by default in the constructor.  A future PR will allow specifying it from the input file.

Loops over an individual batch use `ib` as the index, and indexes into the samples use `is`.   Loops that were untouched in this PR still use `iw` for samples.

The variables `dLogPsi` and `d2LogPsi` are defined in QMCCostFunctionBase.h.  Their length is the number of samples.  However, for the call to `flex_evaluateDeltLogSetup`, only a batch-sized slice is needed.   The fix is to use the conversion from pointer vector to reference vector to extract a subset.

- New feature

- No

laptop

## Checklist


ye-luo:
Can we use a different terminology other than batch? we have "batch" meaning different things in several places and it adds up confusion. My suggestion is wavefront. All independent items within wavefront but sequential from one wavefront to another.

prckent:
Quick comment: Perhaps group? If we are to create a term it should have a clear definition. What sets the size of the group in practice? The "per group" memory usage? Is the intent to create the slater matrices (N^2) for each group when we need to recompute the non-local pseudopotential terms?

markdewing:
The name gets tricky because the batch size matches the size of the inputs to the batched interfaces. So the concepts are not completely independent.   The idea of breaking up a larger set into smaller sets for processing is precisely the meaning of 'batch'.  I wonder if it would be better to use a name that adds a modifier to batch, rather than a separate name.

ye-luo:
A general recommendation is batch size =  walkers_per_rank.

markdewing:
The closest concept to the batch size is the number of walkers in batched VMC.   I think they'll have the same memory usage (or pretty close).

PDoakORNL:
Why worry about batch size your going to divide the total work over the worker threads you have.  At the top level that's going to be equal to the number of cores the rank owns (or it should be). 
The number of walkers for the crowds is derived from that. If you're not going to use the same as the VMC total then that should have something to do with how many of the walker elements the GPU can fit in total, there'd be no point in more than that.
The samples should be large enough that each "crowd" can have essentially the same size to work on.

In the batched VMC if a user wants to be dumb and have a number of walkers that doesn't divide nicely over their number of Crowds i.e. TopLevelThreads we just let them waste time with the imbalance, I'd do the same here. Although if you are going to just make  your own Population you can just make it equal for each thread.

Then I think @ye-luo thinks the computation would do best with a list(s) of walkerelement(s) and samples of equal size? That's where I might like an adjustable batch size or maybe better to just force all the samples down into the lower API. Is this what your batch size is?  I didn't think so because it would have nothing to do with the number of crowds.

So a group can be evaluated for each call from each thread to the lower level API. If this happened in a synchronized manner those would collectively be a wavefront, but we don't want to do that the GPU performs suboptimally if everyone transers and then everyone computes.  So there should if at all possible be no synchronization across the wavefront.  It should be like number of crowds lines of gunpowder of num walker_elements width burning toward the final synchronization.

ye-luo:
@PDoakORNL Here a wavefront is more like a walker population in VMC batched. Elements in a wavefront all have dedicated memory spaces. So they have the potential to run by a parallel device (CPU threads, GPU SM/CUs). Then the wavefront may be subdivided like the crowds in a VMC driver but we are **not** talking about the subdivided batches here.

> Why worry about batch size. You are going to divide the total work over the worker threads you have. At the top level that's going to be equal to the number of cores the rank owns (or it should be).

We are not dealing with threads or GPU streams here but to figure out maximum concurrency restricted by the device memory which is the wavefront size.

> Then I think @ye-luo thinks the computation would do best with a list(s) of walkerelement(s) and samples of equal size?

We are not dealing with efficiency here. The efficiency is constrained by the maximal concurrency. Once the maximum is exposed, we can than deal with them by threads or GPU streams like the crowds in VMC.

ye-luo:
To reduce synchronization, we actually needs to implement three levels.
800 samples.
wavefront size 100, 8 waves
crowd size 20, 5 crowds.

loop crowds 0-4 // parallel
  loop waves 0-7 // serial
    loops elements of wavefronts within this crowd, 0-19 // parallel

prckent:
(We should avoid using wavefront since it collides with AMD terminology)

ye-luo:
> (We should avoid using wavefront since it collides with AMD terminology)
If there is a better name, we can avoid wavefront. The reason I chose wavefront was to express the concurrency naturally.
'group' is not a good choice in this aspect.

PDoakORNL:
@ye-luo That's excellent that you have the information for that already. How can we determine that for an arbitrary calculation?  Is it as simple as
number of threads/streams to get decent desync from accellerator == 5 for V100
mem for walkers / mem per walker / number of crowds = 20
size of coefficients reasonable transfer and launch a kernel for on stream = 100 samples
number of grid blocks per kernel = 20

markdewing:
Once this PR gets in, optimizer calculations using the new driver should be possible, in theory (as long as they don't use the non-local PP part in the optimization - parameter derivatives related to the Hamiltonian are not implemented yet).
For this PR, I suspect the batch size should be set to the # of cores for CPU calculations, and as large as possible for GPU calculations.

The `flex_evaluateParameterDerivatives` function does not have an underlying parallel implementation (call to mw_*).  However, in the tests I've run so far - NiO with J2 (no J3) up to 768 electrons, this function doesn't take very much time.  I suspect this will change when scaling up in the number of parameters.

The next step is to create crowd-like parallelism above this batch level.   That would involve pulling p_ptr_list, etc into a separate structure.   Then put most of the bodies of checkConfigurations and correlatedSampling into a lambda to be called by a ParallelExecutor.    This has the following benefits

1. Running on GPU: Multiple crowds allow overlapped communication/computation between different crowds
2. Running on CPU:
  a. Allocations (of p_ptr_list and other structures) stay local to a core
  b. Assuming # of crowds is the number of cores, and the batch size is one, the lack of parallel flex_evaluateParameterDerivatives doesn't matter
  c. Larger grained parallelism means less synchronization and less serial work (that is, pretty much the entire body of checkConfiguration becomes parallel work, compared to just the flex_* calls being parallel work)

With the new batched drivers, there are more knobs that control performance.    This is something we will need to address.

PDoakORNL (approved):
I'm approving this but I don't think the flex functions are called with the breakup of resources and data I'd like to see.  That of course could be much different by just messing with various loops to get different partitions.  But this could be a first batched implementation.

The batch size here is actually the population size? I think that is very specific to thinking of the number of "walkers" being the number of samples that way down stream could be computed at once.  

All these little loops around thinks make it unclear to me what the concurrent context is, actually there seem to be potentially many. The whole scope of BatchedCostFunction object seems in play the whole time. 

I'm looking for one where the 
#pragma omp for
would go over the "walker resouce" to produce the possible to desynchronize input to the lower level calls I don't see it.

I see many loops that might or might not have a pragma dropped over them and be concurrent over different partitions of data and resources. I don't think there is much work in many of these loops. But without some hard thought they of course require synchronization if the next partitioning could be different and is dependent on the outcome of the previous loop.  

I think dividing the samples/work and walkers/memory spaces into well defined partitions distributed over well defined scopes (crowds) will make the data flow and dependencies explicit, making data races less tricky and increase the amount of time threads are independent. Resource contention and synchronizations are expensive in hybrid architectures. 

Obviously this is could be done much better in VMCBatch/DMCBatch as well.  For DMC especially I think this is how we get to the point where different algorithms for population control and synchronization become feasible to work on. 

I hope this all moves in that direction, I think that should be done before we're tuning the size of each level of concurrency.

PDoakORNL:
Mark my approval and your comment crossed, your comment above addressed a number of my concerns so... I should have refreshed before I submitted the approval.

ye-luo:
flex_XXX is always intended for a secondary parallelism below crowds which is the desynchronized scope.
Before WFC supplying mw_XXX for actually implementation need, flex_XXX can be the stop-gap handling the loop of walkers.

ye-luo:
Let us settle a name for the batch of walkers processing all samples. My reqeust doesn't need to block the current PR but needs to replace the terminology "batch" in the next PR. batch can be used as generic term for describing things.

markdewing:
As for a replacement for the batch name: Maybe 'chunk'?   I believe this is used in OpenMP to describe the amount work (loop iterations) handled on each thread.  The parameter could be called 'chunk_size', instead of 'batch_size'?

PDoakORNL:
Generally I'd see it as the size of the vector of resources a context is iterating over. That would be self documenting to a degree because you could then see what the chunked call is limited over.  If the resources weren't distributed from an enclosing scope but created in the context. I think it would be less confusing to use a name that made what they are obvious.  Perhaps num_evaluation_walkers here.

Generic variable names that require writing and reading documentation are nice to avoid.

markdewing:
This is the size of the items being passed to the flex_* functions, so maybe flex_batch_size or num_flex_elements?

ye-luo:
> This is the size of the items being passed to the flex_* functions, so maybe flex_batch_size or num_flex_elements?

This is **not** the size of the items being passed to the flex_* functions. It is 100 in my example, the size passed to flex_XXX is after crowd division, namely 20.

markdewing:
In the current PR, `opt_batch_size_` is the size of the items being passed to the flex_* functions.   I think this is what Ye calls 'crowd size' in his example.   I would write the loops the same way, but I think about the decomposition differently.

It also depends on which parameters are inputs, and which are derived.  I view the {crowd size|batch size|size of items passed to flex_* routines} as an input parameter, since it controls the efficiency on different hardware and it controls memory usage. 
The number of crowds is also an input parameter, since each crowd can be run on separate thread.
The number of samples is an input parameter.

Once these are fixed, then all the other parameters can be derived.


```
Samples: 800
loop over crowds (0-4)   // (5 crowds, 160 samples per crowd to process)
    loop over wavefront/batches (0-7)  // (crowd size/batch size is 20, need 8 iterations to process all the samples for this crowd)
         call flex_* (which loops over crowd size/batch size)
```

I guess I don't understand the importance of the wavefront size (= crowd size * number of crowds).   I suppose it is the total number of walkers resident in memory, and so forms a limit to be respected.   I get the abstract view that it is the parallel work to be done, but in practice is it is defined by other parameters more directly related to mapping to the hardware.

PR #2713, opened 2020-09-15
More organized printout

ye-luo:
## Proposed changes

@mcbennet ask me how to check if feature is using OpenMP offload or not. I ended up cleaning out printout a bit.
This PR adds printing
```
   Single particle orbitals (SPO)
   ------------------------------
    Name:    Type: spline   Builder class name: EinsplineSetBuilder
```
and
```
   Single Slater determinant
   -------------------------


     Determinant
     -----------
      Name: updet   Spin group: 0   SPO name: updet

      Reusing a SPO set updet
      Setting delay_rank to default value 1
      Using rank-1 Sherman-Morrison Fahy update (SM1)


     Determinant
     -----------
      Name: downdet   Spin group: 1   SPO name: downdet

      Reusing a SPO set downdet
      Setting delay_rank to default value 1
      Using rank-1 Sherman-Morrison Fahy update (SM1)
```
More cleaning work will still be needed for all SPOSet buildres.
The printout is still very messy.
This will be done together with input style cleaning.

- New feature

- No

ryzen-box

## Checklist


prckent (changes requested):
Thanks. Where in the output does is say whether offload is used?

I hit a variant of this problem but accident this morning when I noticed that the "offload" builds from AOMP on cdash weren't offload enabled due to ctest configuration issues.  (In that case we don't print anything and should print "Compiled for CPUs only" or similar.) 

For the current issue we should specifically print "Running on cpu" or "Running on accelerator via OpenMP offload" for each compute object.  If we are also using CUDA that should be indicated etc.

src/QMCApp/QMCMain.cpp
```diff
@@ -117,11 +120,23 @@ QMCMain::QMCMain(Communicate* c)
                 << std::endl;
 
   // Record features configured in cmake or selected via command-line arguments to the printout
-#ifdef ENABLE_OFFLOAD
-  app_summary() << "\n  OpenMP target offload to accelerators enabled" << std::endl;
+  app_summary() << std::endl;
+#if !defined(ENABLE_OFFLOAD) && !defined(ENABLE_CUDA) && !defined(QMC_CUDA) && !defined(ENABLE_ROCM)
+  app_summary() << "  CPU only build" << std::endl;
```
ye-luo, line 125: Added printout for "CPU only build"

src/QMCWaveFunctions/BsplineFactory/createComplexDouble.cpp
```diff
@@ -37,44 +37,62 @@ BsplineReaderBase* createBsplineComplexDouble(EinsplineSetBuilder* e, bool hybri
...
-      app_log() << "OpenMP offload has not been enabled with hybrid orbital representation!"
-                << " Running on the host." << std::endl;
+      app_summary() << "OpenMP offload has not been implemented to support hybrid orbital representation!"
+                    << " Running on CPU." << std::endl;
+      app_summary() << "    Using hybrid orbital representation." << std::endl;
       aReader = new HybridRepSetReader<HybridRepCplx<SplineC2C<double>>>(e);
     }
     else
+    {
+      app_summary() << "    Running on an accelerator via OpenMP offload." << std::endl;
```
ye-luo, line 53: For spline SPO, print "Running on an accelerator via OpenMP offload." here and "Running on CPU" below

ye-luo (commented):
Overall printout from SPOSetBuilder is still extremely messy.

ye-luo:
Output from a batched determinant implementation with CUDA and OpenMP offload.
```
     Determinant
     -----------
      Name: updet   Spin group: 0   SPO name: updet

      Reusing a SPO set updet
      Setting delay_rank to default value 1
      Using rank-1 Sherman-Morrison Fahy update (SM1)
      Using walker batching.
      Running on an NVIDIA GPU via CUDA acceleration and OpenMP offload.
```

ye-luo:
Spline SPO with OpenMP
```
    Using complex valued spline SPOs with complex single precision storage (C2C).
    Running on an accelerator via OpenMP offload.
```
This message is still immersed in other printout of EinsplineSetBuilder

prckent (approved):

prckent:
Thanks for answering my question and also for doing the cleanup - very good to have less public data, even if just the names.

PR #2717, opened 2020-09-17
Add AFQMC Timer Tool qmc-afqmc-performance

fdmalone:
## Proposed changes

Adds a timer tool for some of the most expensive operations in AFQMC. These are mostly linear algebra calls. Will write timing data to file and can be extended in the future as we spot more problematic code. Pretty helpful for timing on different devices / machines. Code is in Numerics/performance/performance.cpp, it will create a binary in build/bin/qmc-afqmc-performance.

In the process I fixed a compilation error for HIP which unfortunately got mixed in with the PR.

- New feature

- No

V100/MI60/xeon 

## Checklist


mmorale3 (approved):

ye-luo:
Can we use the same timer_manager as real space code?

src/AFQMC/Numerics/performance/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+#// File developed by: Fionn Malone malone14@llnl.gov, Lawrence Livermore National Laboratory
+#//
+#// File created by: Fionn Malone malone14@llnl.gov, Lawrence Livermore National Laboratory
+#////////////////////////////////////////////////////////////////////////////////////////////
+
+MESSAGE("Building AFQMC performance executable ")
+
+SET(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${qmcpack_BINARY_DIR}/bin)
+
+ADD_EXECUTABLE(afqmc_performance performance.cpp)
```
ye-luo, line 16: We try to converge executable names. `qmc-afqmc-performance`?

fdmalone: updated

fdmalone:
> Can we use the same timer_manager as real space code?

I could. The timer I used is used elsewhere in the AFQMC code.

ye-luo (changes requested):
> > Can we use the same timer_manager as real space code?
> 
> I could. The timer I used is used elsewhere in the AFQMC code.

Let us make the effort to reduce duplicated code.

markdewing:
At the very least, it would be nice to have the timer code in src/Utilities rather than src/AFQMC/Utilities.   It would be more obvious there are multiple timers.  Requirements for different uses can be divergent, and timers are easy to write, so it might be okay to have multiple implementations.  Having them in the same place would make it easier to spot similarities and make them more uniform.

fdmalone:
I didn't write the timer (@mmorale3) but it's apparently much simpler to use for my purposes that the timer_manager is, I'm not using the logging features and so on as I just want matrix dimensions vs time for easy plotting.

prckent:
Lets use the same timer manager as the real space code. If there is any missing functionality we can add to add. Besides avoiding duplication we will get uniform printing/reporting.

fdmalone:
I'm working on this now. Why was the PR closed?

ye-luo:
If you get the timer updated, just reopen the PR.

ye-luo:
If you are not using the log feature. use Utilities/Timer.h to get the timing. For places need the log feature, use time_manager.

prckent:
@fdmalone I assumed you were going to push a fresh PR. Reopen if easier. Sorry for the surprise.

fdmalone:
I made the change to use Utilities/Timer which had what I needed.

ye-luo:
Okay to test

ye-luo:
@fdmalone please update the PR title to be more verbose and also update the description.

src/AFQMC/config.h
```diff
@@ -28,8 +28,6 @@
 #include "multi/memory/fallback.hpp"
 
 #include "Utilities/TimerManager.h"
-#include "AFQMC/Utilities/myTimer.h"
-extern myTimer Timer;
```
ye-luo, line 32: Where is the definition of "Timer" correspond to this extern?
AFQMCFactory.cpp?

fdmalone: Yes. I believe this is legacy code and myTimer is no longer used globally in the AFQMC code, only for debugging with definite scope. I had to remove the extern to avoid name clash with Timer.h

fdmalone: Although given the build errors this assumption was wrong. Let me fix these.

ye-luo: I mean if the global one is not needed any more. The definition needs to be removed as well.
In addition, please fix the CI failures.

ye-luo (approved):
Assuming CI passes.

mmorale3 (dismissed):

fdmalone:
I fixed the issues with the real builds. Some cuda build were erroring from move errors?

ye-luo:
> I fixed the issues with the real builds. Some cuda build were erroring from move errors?

std::move are just warnings. Probably need to search error if the output is massive.

ye-luo:
fixing the std::move clang warning in a separate PR will be also beneficial.

ye-luo (approved):

PR #2723, opened 2020-09-22
Remove two SPOSet map in SlaterDetBuilder and SlaterDet

ye-luo:
## Proposed changes
The two maps owned and maintained in SlaterDetBuilder and SlaterDet are not necessary any more as SPOSet will eventually be unique_ptr no more shared by determinants as we are approaching clean spo determinant input style and memory clean code. In a spin restricted case, the sposet object for spin up and down no more share the same pointer but share coefficients only. So we are not loosing anything in this change.

Additional changes:
1. Moved RotatedSPO interception from SPOSetBuilderFactory to SPOSetBuilder which is more natural.
2. SPOSetScanner moved from SlaterDetBuilder to SPOSetBuilderFactory.

- No

Bora. Tested all short tests.

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/DiffWaveFunctionComponent.cpp
+++ b/src/QMCWaveFunctions/DiffWaveFunctionComponent.cpp
@@ -67,20 +67,20 @@ void NumericalDiffOrbital::evaluateDerivatives(ParticleSet& P,
     int jj = ind_map[j];
     if (jj < 0)
       continue;
-    LogValueType plus=0.0;
-    LogValueType minus=0.0;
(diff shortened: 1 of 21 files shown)


PDoakORNL (approved):
This looks good as a step toward a cleaner ownership and coupling between sposets and determinants.

Testing could really improve at least in clarity (but that predates this PR).

ye-luo:
I was working on changes in a large scope and found it too big. So getting this part first. More tests are on the way for input styles. Will address the CI failure soon.

PDoakORNL (approved):
LGTM

PR #2726, opened 2020-09-25
Fix LCAO initialization precision loss in mixed precision build

ye-luo:
## Proposed changes
1. The `transform' option in LCAO converts gaussian and slater orbitals to splines. However the initial value before conversion is not calculated in full precision in a mixed precision build. This is fixed by respect the template type parameter.
2. This PR also removes OptimizableFunctorBase from SlaterCombo and GaussianCombo since we are not using this feature at the moment.
3. Restore a few unit tests in mixed precision build and restore deterministic he_opt test in complex builds.

- Bugfix

- No

Bora

## Checklist


src/Numerics/GaussianBasisSet.h
```diff
@@ -15,19 +15,18 @@
...
 {
 template<class T>
-struct GaussianCombo : public OptimizableFunctorBase
+struct GaussianCombo
 {
   // Caution: most other code assumes value_type can only be real
   // but maybe it can be different precision
   // Possibly one of these types is the full precision and the other reduced precision
-  typedef T value_type;
+  typedef T real_type;
```
PDoakORNL, line 29: Later on this class appears as a template argument to Transform2GridFunctor and needs to discover T. Being able to see at a glance what is templated type in the class itself and that it is assumed real means I'd much prefer
```c++
template<typename REAL>
class GaussianCombo
{
 static_assert(std::is_floating_point<Real>::value,
                  "Other code assumes REAL is a real type");
 using Real = REAL; //Don't use this for member declarations;
...
```

src/Numerics/SlaterBasisSet.h
```diff
@@ -14,17 +14,16 @@
...
-#include "Numerics/OptimizableFunctorBase.h"
 #include "Numerics/SlaterTypeOrbital.h"
 #include "OhmmsData/AttributeSet.h"
 #include "io/hdf_archive.h"
 
 namespace qmcplusplus
 {
 template<class T>
-struct SlaterCombo : public OptimizableFunctorBase
+struct SlaterCombo
```
PDoakORNL, line 24: See GaussianCombo comment

src/Numerics/SlaterTypeOrbital.h
```diff
@@ -54,9 +53,9 @@ struct STONorm
  * where n-1 is the number of nodes.
  */
 template<class T>
-struct GenericSTO : public OptimizableFunctorBase
+struct GenericSTO
```
PDoakORNL, line 56: See GaussianComboComment

PDoakORNL (changes requested):
I think this is a good chance to do a little better

PDoakORNL (approved):

Issue #2070, opened 2019-11-01
Separate SPOSet tag and determinant tag needed


ye-luo:
Leave the LCAO a side at the moment since the handling of MO coefficients is more complicated but it carries the same problem.

Let us focus on the spline case. We have three input styles.

Style 1. using sposet_builder outside determinantset
```
<wavefunction name="psi0" target="e">
   <sposet_builder type="bspline" href="./pwscf.pwscf.h5" tilematrix="1 0 0 0 1 0 0 0 1" twistnum="0" source="ion0" version="0.10" meshfactor="1.0" precision="float" truncate="no">
      <sposet type="bspline" name="spo_ud" size="4" spindataset="0"/>
      <sposet type="bspline" name="spo_dm" index_min="4" index_max="8" spindataset="0"/>
   </sposet_builder>
   <determinantset>
      <slaterdeterminant>
         <determinant id="updet" group="u" sposet="spo_ud" size="4"/>
         <determinant id="downdet" group="d" sposet="spo_ud" size="4"/>
      </slaterdeterminant>
   </determinantset>
</wavefunction>
```
spo related info is 1 level above sposet tag.
Q1. why do we need attribute size on the determinant line. It should be derived from the number of 'u'. Or it means something else?

Style 2. using sposet inside determinantset
```
<wavefunction name="psi0" target="e">
   <determinantset type="bspline" href="./pwscf.pwscf.h5" tilematrix="1 0 0 0 1 0 0 0 1" twistnum="0" source="ion0" version="0.10" meshfactor="1.0" precision="float" truncate="no">
      <sposet type="bspline" name="spo_ud" size="4" spindataset="0"/>
      <sposet type="bspline" name="spo_dm" index_min="4" index_max="8" spindataset="0"/>
      <slaterdeterminant>
         <determinant id="updet" group="u" sposet="spo_ud" size="4"/>
         <determinant id="downdet" group="d" sposet="spo_ud" size="4"/>
      </slaterdeterminant>
   </determinantset>
</wavefunction>
```
In this case, the info on the sposet_builder are moved to determinantset and sposet are moved inside determinantset.
spo related info is 1 level above sposet tag.
What is index_min, index_max? we need documentation.

Case 3. Let me call it mixed style.
```
<wavefunction name="psi0" target="e">
[...]

PR #2727, opened 2020-09-25
Add unit tests for sposet input styles

ye-luo:
## Proposed changes
Step 1 of #2070
Introduced unit tests for current and desired sposet input styles. Some within determinants. Those styles with sposet outside determinantset are the only ones we are going to support in the future.

Next PRs will settle
1. make LCAO h5 basis initialization controlled by input and respect name tag. RMG needs two basissets for different spins.
2. Add deprecation warning to all deprecated styles. This will go to the next release and then we can delete those deprecated styles in the release after next.

- Bugfix
- Unit tests

- No

Bora

## Checklist


ye-luo:
Ready for review.

prckent:
I tweaked the error message which has needed improving for some time. However, would it ever be advisable to not abort in this scenario?

ye-luo:
> I tweaked the error message which has needed improving for some time. However, would it ever be advisable to not abort in this scenario?

This part of code was touched because of formatting. It was added for catching badbit caused by de-normal number and I think we should throw an exception. Will change it in a separate PR after more testing.

src/OhmmsData/libxmldefs.h
```diff
@@ -128,8 +128,8 @@ inline bool putContent(std::vector<T>& a, const xmlNodePtr cur)
       b.push_back(t);
     else if (!stream.eof() && stream.fail())
     {
-      std::cerr << "failbit detected when reading type (type_info::name) "
-                << typeid(T).name() << ", value " << t << std::endl;
+      std::cerr << "Error parsing XML for type (type_info::name) " << typeid(T).name() << ", value " << t
```
PDoakORNL, line 131: I assume this is the error which is going to become an exception?

src/QMCWaveFunctions/SPOSetBuilderFactory.cpp
```diff
@@ -301,37 +300,42 @@ void SPOSetBuilderFactory::build_sposet_collection(xmlNodePtr cur)
...
-  app_summary() << "   ------------------------------------------" << std::endl;
+  app_summary() << "   Single particle orbitals (SPO) collection" << std::endl;
+  app_summary() << "   -----------------------------------------" << std::endl;
   app_summary() << "    Name: " << collection_name << "   Type input: " << collection_type << std::endl;
   app_summary() << std::endl;
 
   // create the SPOSet builder
   SPOSetBuilder* bb = createSPOSetBuilder(cur);
+
+  // going through a list of basisset entries
```
PDoakORNL, line 311: basis set is two words I think the cname should be "basis_set"

Not addressable here should we start to make a list of input names to update?

PDoakORNL (approved):

src/OhmmsData/libxmldefs.h
prckent: Yes

src/QMCWaveFunctions/SPOSetBuilderFactory.cpp
ye-luo: I vote for not changing. It is short enough. Long ones need breakup. If change, we need proper protections.

PR #2732, opened 2020-09-29
Catchable APP_ABORT

PDoakORNL:
## Proposed changes

The APP_ABORT macro creates a situation where you can't simply break on any fatal error in QMCPACK. In gdb and lldb you won't catch at breakpoint set to a line in a macro definition. So if you want to catch all fatal errors in the debugger APP_ABORT is a problem. The macro is wide spread and tracking down where it actually was called from can be a pain.

This PR creates an actual function called by APP_ABORT so a breakpoint can be set to catch calls to APP_ABORT with the calling context intact.

- New feature

- No

Leconte

## Checklist

### DIFF
--- a/src/Message/AppAbort.cpp
+++ b/src/Message/AppAbort.cpp
@@ -11,9 +11,10 @@
 
 #include "Message/AppAbort.h"
 #include <iostream>
+#include "config.h"
 
 #ifdef HAVE_MPI
-#include "mpi3/environment.hpp"
+#include <mpi.h>
 
 void breakableAppAbort(const std::string& str_msg)
 {
--- a/src/Message/AppAbort.cpp
+++ b/src/Message/AppAbort.cpp
@@ -13,15 +13,17 @@
 #include <iostream>
 
 #ifdef HAVE_MPI
-void breakableAppAbort(std::string str_msg)
+#include <mpi.h>
+
+void breakableAppAbort(const std::string& str_msg)
 {
-  std::cerr << str_msg << std::endl;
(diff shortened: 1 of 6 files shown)


ye-luo:
I think it will better we move all the abort macro and your added routines to a separate header file out of Communicate.h

markdewing:
To solve the original problem, can you break on a call to MPI_Abort or exit?   
In any case, funneling APP_ABORT through a single function is a reasonable change.

PDoakORNL:
Sadly MPI_Abort doesn't work and that's the case I care about.

PDoakORNL:
I think the file split is reasonable since I would like to see APP_ABORT eventually disappear.  And since the argument to use it is lack of access to comm and not wanted to throw an exception it doesn't belong in Communicate. Will do tomorrow AM.

src/Message/AppAbort.cpp
```diff
@@ -0,0 +1,27 @@
...
+// File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Lab
+//
+// Refactored from: Communicate.cpp
+////////////////////////////////////////////////////////////////////////////////
+
+#include "Message/AppAbort.h"
+#include <iostream>
+
+#ifdef HAVE_MPI
+void breakableAppAbort(std::string str_msg)
```
ye-luo, line 16: Mixing Include mpi.h.
Please also use `const std::string&`

PDoakORNL:
Looking into issue last commit caused with AFQMC

ye-luo (approved):

PR #2733, opened 2020-10-01
Spin-orbit documentation

camelto2:

## Proposed changes

This PR adds some of the needed documentation in order to carry out spin-orbit calculations in QMCPACK for VMC/DMC. I added information in each of the relevant sections (Hamiltonian and Observables, Quantum Monte Carlo methods, etc) as well as a dedicated Spin-Orbit calculations in QMC section. I documented each of the relevant changes in the qmcpack input files in order to carry out one of these calculations, as well as provide some of the theory since this is likely unfamiliar to many users. 

- No

macbook

## Checklist

### DIFF
--- a/docs/bibs/methods.bib
+++ b/docs/bibs/methods.bib
@@ -105,3 +105,33 @@
   volume        = {132},
   year          = {2010}
 }
+
+@article{Melton2016-1,
+  title = {Spin-orbit interactions in electronic structure quantum Monte Carlo methods},
+  author = {Melton, Cody A. and Zhu, Minyi and Guo, Shi and Ambrosetti, Alberto and Pederiva, Francesco and Mitas, Lubos},
+  journal = {Phys. Rev. A},
+  volume = {93},
+  issue = {4},
+  pages = {042502},
+  numpages = {5},
+  year = {2016},
+  month = {Apr},
+  publisher = {American Physical Society},
+  doi = {10.1103/PhysRevA.93.042502},
+  url = {https://link.aps.org/doi/10.1103/PhysRevA.93.042502}
+}
+
+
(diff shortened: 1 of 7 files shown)


rcclay:
Ok to test.

ye-luo:
No need to block this PR but after reading the documentation, I feel the current way is error-prone and need to simplify the input.
1. the input of tag of spinor needs to be at particleset input. variable spins can be unique_ptr ParticleScalar_t initialized when activated.
2. spo input doesn't need to have special type="spinorbspline", if sposet can be used to tell spinor.
3. the same logic as 2 applies to hamiltonian and driver.

camelto2:
.

> No need to block this PR but after reading the documentation, I feel the current way is error-prone and need to simplify the input.
> 
> 1. the input of tag of spinor needs to be at particleset input. variable spins can be unique_ptr ParticleScalar_t initialized when activated.
> 2. spo input doesn't need to have special type="spinorbspline", if sposet can be used to tell spinor.
> 3. the same logic as 2 applies to hamiltonian and driver.

I agree that making some simplifications in the input specification is probably needed in the future. Moving the information into particleset seems reasonable, as we can just activate the appropriate builders in the code if the flag is present in particleset as opposed to reading it in multiple places from the xml input. 

The only thing I would like to keep in the Hamiltonian is the physicalSO flag, because there are some use cases where we would want to accumulate the spin-orbit as an observable but not have it in the local energy (like the perturbative/downfolding approach of Wagner). I think that particular parameter is more appropriate in the hamiltonian input.

ye-luo:
@rcclay approve?

rcclay (approved):
LGTM

ye-luo (approved):

PR #2734, opened 2020-10-02
Noncollinear and spin-orbit implementation for k-point Hamiltonian in AFQMC

mmorale3:

The non-collinear implementation of AFQMC is completed in this PR.
In non-collinear calculations, a spin-orbit hamiltonian is now allowed.
Propagation and energy evaluation are implemented.
Only the k-point hamiltonian (both CPU and GPU) works with non-collinear.
Observables (other than the energy) are not yet enabled with non-collinear.
Future PRs will complete the missing features.

New feature: Noncollinear and spin-orbit interactions in AFQMC.

No

Quartz, Lassen. GCC and Clang compilers.

## Checklist


external_codes/boost_multi/multi/memory/allocator.hpp
```diff
@@ -49,7 +49,8 @@ class allocator{
 	bool operator!=(allocator const& o) const{return mp_ != o.mp_;}
 	using void_pointer = typename std::pointer_traits<decltype(std::declval<memory_type*>()->allocate(0, 0))>::template rebind<void>;
 	pointer allocate(size_type n){
-		return static_cast<pointer>(static_cast<void_pointer>(mp_->allocate(n*sizeof(value_type), alignof(T))));
+		return static_cast<pointer>(static_cast<void_pointer>(mp_->allocate(n*sizeof(value_type), 16)));
```
fdmalone, line 52: why 16?

fdmalone (approved):
LGTM

fdmalone:
This only allows for spin dependent one-body Hamiltonian for k-point calculations right? What format  is expected in the input wavefunction / hamiltonian.

external_codes/boost_multi/multi/memory/allocator.hpp
mmorale3: This is a temporary fix until we update multi.
Cuda is incompatible with alignof() for std::complex<double>, so using alignof(T) in general
causes alignment issues when compiled in double precision for gpus.
With 16 all the regular types work fine. (Notice that cudamalloc aligns to 256).

mmorale3:
The one body hamiltonian as to be 2M*2M in non-collinear calculations.
Only the kpoint hamiltonian works now. For any other hamiltonian, the code should abort at some point
when it realizes the calculation is non-collinear. I will complete the missing parts soon.

fdmalone:
For the kpoint code this is 2*m x 2*m right? The HDF5 format doesn't change?

external_codes/boost_multi/multi/memory/allocator.hpp
ye-luo: Should we update multi?

ye-luo (changes requested):

mmorale3:
Updating multi is possibly a lot of work.
I prefer to push this first and update multi after.

ye-luo (approved):
Preferably, we should update multi and then update our code. I tried updating multi but found an internal conflict in multi. So let us use the workaround from @mmorale3 at the moment.

PR #2736, opened 2020-10-06
Use fixed width 64 bit integer for binary string conversion

ye-luo:
## Proposed changes
Fix nightly test failure on raspberry-pi.
"long int" is only 32 bit on ARM aarch64.

- Bugfix

- No

Ye-raspberrypi

## Checklist


prckent:
Did you miss a change?
```
Real-Mixed-enable-CUDA/src/QMCTools/QMCGaussianParserBase.cpp:962:26: error: type-id cannot have a name
    std::vector<unsigned int64_t> Val_alpha, Val_beta;
                         ^~~~~~~
```

ye-luo:
> Did you miss a change?
> 
> ```
> Real-Mixed-enable-CUDA/src/QMCTools/QMCGaussianParserBase.cpp:962:26: error: type-id cannot have a name
>     std::vector<unsigned int64_t> Val_alpha, Val_beta;
>                          ^~~~~~~
> ```

It seems showing up only on the CI machine with Clang. Anyway, it was unused code and just got removed.

ye-luo:
@prckent readthedocs seems having issue fetch the repo but I think we can just ignore since no docs are touched by this PR. Do you know how to trigger readthedocs to retry a build if it happens in the future? Our usual way only affects Jenkins.

prckent (approved):

prckent:
rtd issue is odd since the qmcpack manual was ok but the nexus one was not. Hopefully a transient issue at rtd.

PR #2737, opened 2020-10-06
Remove std::move warning in AFQMC.

ye-luo:
## Proposed changes
Resolves clang warning about using std::move on temporary objects.

- Bugfix

- No

Ryzen-box

## Checklist


prckent:
(Happily rtd is working again on this PR)

mmorale3:
This should only be done when we officially move to c++17.
This is not a warning in c++14.
As long as we support c++14, I prefer to use a macro instead of eliminating the std::move.

src/AFQMC/Estimators/BackPropagatedEstimator.hpp
```diff
@@ -191,7 +191,7 @@ class BackPropagatedEstimator : public EstimatorBase
 
     // 1. check structures
     if (Refs.size(0) != wset.size() || Refs.size(1) != nrefs || Refs.size(2) != nrow * ncol)
-      Refs = std::move(mpi3CTensor({wset.size(), nrefs, nrow * ncol}, Refs.get_allocator()));
+      Refs = mpi3CTensor({wset.size(), nrefs, nrow * ncol}, Refs.get_allocator());
```
ye-luo, line 194: No change of behavior. still creating the temporal object (prvalue) and picked the move assign operator.

src/AFQMC/Estimators/BackPropagatedEstimator.hpp
```diff
@@ -191,7 +191,7 @@ class BackPropagatedEstimator : public EstimatorBase
 
     // 1. check structures
     if (Refs.size(0) != wset.size() || Refs.size(1) != nrefs || Refs.size(2) != nrow * ncol)
-      Refs = std::move(mpi3CTensor({wset.size(), nrefs, nrow * ncol}, Refs.get_allocator()));
```
ye-luo, line 194: In this case, temporary mpi3CTensor type object is created and moved to Refs via move assign operator.

src/AFQMC/Estimators/FullObsHandler.hpp
```diff
@@ -111,20 +111,20 @@ class FullObsHandler : public AFQMCInfo
       std::transform(cname.begin(), cname.end(), cname.begin(), (int (*)(int))tolower);
       if (cname == "onerdm")
       {
-        properties.emplace_back(Observable(std::move(full1rdm(TG, info, cur, walker_type, nave, block_size))));
```
ye-luo, line 114: In this case, temporal full1rdm type object is created and then calling move constructor.

src/AFQMC/Estimators/FullObsHandler.hpp
```diff
@@ -111,20 +111,20 @@ class FullObsHandler : public AFQMCInfo
...
+        properties.emplace_back(Observable(full1rdm(TG, info, cur, walker_type, nave, block_size)));
```
ye-luo, line 114: compilers support copy elision(optional as an optimization prior C++17 and mandatory since C++17) will construct full1rd inplace into Observable.
If the compiler doesn't support copy elision, it first construct the temporal object (prvalue) and then invoke the move constructor which is the old code does.
I don't think it hurts us in C++14.

src/AFQMC/Estimators/FullObsHandler.hpp
ye-luo: In addition, if std::move is needed on the full1rdm. std::move is needed on Observable as well.

ye-luo (commented):
Clang just print a warning when std::move(temporal object) is detected. It actually doesn't check if the code needs the copy elision or not.
In AFQMC, std::move(temporal object) is used in two patterns,
1) in a constructor. This case is eligible for copy elision.
2) in a move assign operator, it has nothing to do with copy elision.
After analyzing these two patterns, I don't think there is any risk.
C++17 just made copy elision mandatory which is an optional optimization in C++14.

ye-luo:
Although I prefer to see less warning, I'm OK to defer the code change after 3.10 release which should be in just a few weeks.

prckent:
@ye-luo Please can you rename this PR a WIP or make it a draft so that it is clearly not a candidate for merging currently?

mmorale3:
I changed my mind, let's merge this PR since the number of warnings with c++17 compilers is problematic.
I'll fix any issues that appear with old compilers.

mmorale3 (approved):

PR #2738, opened 2020-10-08
Cleanup WF and Ham factory and pool

ye-luo:
## Proposed changes
1. Clean up WaveFunctionPool,  WaveFunctionFactory, HamiltonianPool and HamiltonianFactory.
2. Remove driver dependence on WaveFunctionPool and HamiltonianPool
3. Remove MPI comunicator in TrialWaveFunction. It was introduced for "convenience". Finally no more needed.
4. Caught a bug affecting unit tests. We need to enforce the big five rule.

- Bugfix

- No

Bora

## Checklist


src/QMCHamiltonians/HamiltonianFactory.h
```diff
@@ -30,39 +30,14 @@ class HamiltonianFactory : public MPIObjectBase
...
-  OrbitalPoolType& psiPool;
-  ///input node for a many-body wavefunction
-  xmlNodePtr myNode;
-
-  ///name of the TrialWaveFunction
-  std::string psiName;
-
-  ///list of the old to new name
-  std::map<std::string, std::string> RenamedProperty;
+  typedef std::map<std::string, WaveFunctionFactory*> PsiPoolType;
```
PDoakORNL, line 33: prefer 
```c++
using PsiPool = std::map<std::string, WaveFunctionFactory*>;
```

Although considering the existing typedefs perhaps a cleanup stage 2.

PDoakORNL (approved):
Especially happy to see the drop inheritance from MPIObjectBase in TrialWaveFunction.  But this is a significant improvement to this whole area of code.

src/QMCHamiltonians/HamiltonianFactory.h
ye-luo: Will change it when I resolve conflict with a later PR.

PR #2741, opened 2020-10-08
Add initial parameter gradient test

markdewing:
## Proposed changes
The system is a Helium atom with a bspline e-e jastrow with 4 parameters.  The gradient wrt those parameters is tested.

The test uses a new QMC_RUN_AND_CHECK_CUSTOM_SCALAR function.
This test uses the --name, --ref-value, --ref-error options to check_scalars.py added in #2640 in order to test arbitrarily named values.

The new function uses named arguments that are handled with
[cmake_parse_arguments.](https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html). This should make it easier to add new functionality via new named arguments in the future.   We might consider redoing QMC_RUN_AND_CHECK in the same style.  It might be possible to combine these functions (SCALAR_VALUES vs. CUSTOM_SCALAR_VALUES arguments?)

The first four arguments (BASE_NAME, BASE_DIR, PREFIX, and INPUT_FILE) aren't optional.   We could make them positional.  However, part of the motivation for named parameters is to remember what each parameter does w/o looking at the function source).    The function could also be updated to issue an error if any these are undefined.

- New feature

- No

laptop

## Checklist


prckent (commented):
How did you determine the error bars for the tests? How statistically reliable do you expect these to be?

markdewing:
I ran the problem 1000x as long and scaled the error bars.    (Originally the long run was 100x as long as the test run, but then I realized the test run took too much time, and scaled it back by 10x)
I don't know how reliable the statistics will be for parameter derivatives.

ye-luo:
The title says initial parameter gradient test. Will the testing capability be enhanced? Just to be sure that the functional to be added will be different from wavefunctionTester which should deal with per walker data. In the optimizer, testing could target statistical data.

prckent (commented):
This seems like a good route to extensibly test parameter derivatives as seen by the optimizer. Minor comments/questions, none critical for this PR:

Since we are using the opt_batch driver, is it necessary to specify samples or stepsbetweensamples?

I thought for the former we had agreed on one sample per actual step as a simplification. I don't recall discussion on stepsbetweensamples, but it feels like one more setting we could simplify and remove. substeps already exists for decorrelation purposes.

What do you think?

markdewing:
@prckent The samples and stepsbetweensamples parameters are unused with the batch optimization driver.  I removed them from the input file.

@ye-luo This PR contains the CMake infrastructure for statistical tests of parameter derivatives plus one test.  The 'initial' part referred to the number of tests of this type - more parameter derivative tests can and should be added.

prckent (approved):

PR #2743, opened 2020-10-08
Fix J2 KEcorr in multi threaded runs.

### LINKED ISSUE #2742: KEcorr is not constant in DMC
**Describe the bug**
The KEcorr in qmcpack should be constant for a fixed jastrow unless I am mistaken. In VMC, the correct behavior is obtained. However, in a DMC run I do not see this behavior. Since it is just a function of the FT of J2, which is fixed in both VMC and DMC, I would expect the KEcorr value in the DMC scalar.dat to be the same as the VMC. I will note that every few blocks the DMC agrees identically to the VMC, but not always as I naively expect. 

**To Reproduce**
1. checkout develop
2. go to tests/solids/diamond_2x1x1_pp/det_qmc_short.in.xml and add a simple dmc section following the vmc
3. Observe KEcorr inside the VMC and DMC .scalar.dat files.

**Expected behavior**
The KEcorr should be constant for a fixed jastrow

**System:**
my workstation, intel19.0.5

**Additional context**

ye-luo:
## Proposed changes

Fix #2742. @camelto2 I checked in VMC only.

- Bugfix
- No

Bora

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Jastrow/J2OrbitalSoA.h
+++ b/src/QMCWaveFunctions/Jastrow/J2OrbitalSoA.h
@@ -446,6 +446,7 @@ WaveFunctionComponentPtr J2OrbitalSoA<FT>::makeClone(ParticleSet& tqp) const
         fcmap[F[ij]] = fc;
       }
     }
+  j2copy->KEcorr = KEcorr;
   j2copy->Optimizable = Optimizable;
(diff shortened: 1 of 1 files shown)


camelto2 (approved):
I confirmed this fixes the problem in both VMC/DMC as I described in #2742

Issue #2433, opened 2020-05-06
AFQMC xxx_buffer_generators are global


PDoakORNL:
**Is your feature request related to a problem? Please describe.**
A great deal of trouble and annoyance is still being causes in the realspace QMCPACK because the decision was made to make many important objects globals. It's actually worth it to make an explicit decision about such practices early in a project.

Currently std::shared_ptr is used to create set of pointers to global xxx_buffer_generators.
I don't think this is better practice that simply using a raw pointer since you should be able to reason about the order of your object destruction and this implies users of this shared_ptr may need the global objects destruction delayed past the end of main.

I would argue they should/must be owned by a particular scope and not the "runtime." The consumer of the allocator should be able to control and document as code what the scope of the allocator is.

Global objects cause issue both for unit testing and for concurrent programming.

**Describe the solution you'd like**
Removal reliance on global shared_ptrs.  Choose a scope and lifetime for the allocators.

mmorale3:
I thought  about this for a long time and I decided to stick with global objects.
A new PR changes slightly the access pattern to the allocator generators, I now use a 
monostate-type pattern, but the objects are still global. 
The changes needed to provide a scoped access to these objects would be too great and
I'm not willing to go that route. I need this allocators at both the top level of the code and 
at the lowest level during dispatch calls to cuda libraries (cutensor, cublas and cusparse now
all require workspace, which is a pain to deal with).
Anyway, just letting you know that I made a decision and that it is addressed in a recent PR.

mmorale3:
I'm going to close the issue for now. You can open it again if you wish.

PR #2744, opened 2020-10-09
Introduces new access pattern for stack memory resources in AFQMC.

mmorale3:

## Proposed changes

Modifies the access pattern of stack memory resources in AFQMC. The code now uses a monostate-like pattern
to control access to the global memory resources. The actual memory resources are still global, but now
accessed (a single global state) through generic instantiations of objects of the manager classes.
Partially addresses issues raised in #2433 .
These changes are needed before I can fix issues with cuda11 support, since cusparse's new interface
for sparse matrix multiplication requires workspace arrays now. These changes will also be needed
for an upcoming PR which adds support for cuda's new features/libraries, like cutensor and mixed precision
routines in cublas11, both of which also require workspace arrays deep in the dispatch tree.

- New implementation of stack memory resource access.

- No

Quartz, Lassen. 

## Checklist


ye-luo:
Missing Memory/buffer_managers.cpp file

fdmalone (approved):

PR #2749, opened 2020-10-12
Async ratio and grad computation + Require Intel >=19

ye-luo:
## Proposed changes
add WFC::mw_ratioGradAsync and WFC::ratioGradAsync to support leverage asynchronous OpenMP tasking.
This is a very very preliminary implementation.
Would like to merge this to test more compilers from nightly tests.
`<wavefunction tasking="yes">` triggers the async code path.

Add a safeguard for Intel compiler and raise minimal requirement to 19.0

- New feature

- No

Bora

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/TrialWaveFunction.cpp
+++ b/src/QMCWaveFunctions/TrialWaveFunction.cpp
@@ -612,7 +612,7 @@ void TrialWaveFunction::flex_calcRatioGrad(const RefVector<TrialWaveFunction>& w
   const int num_wf = wf_list.size();
   grad_new.resize(num_wf);
   std::fill(grad_new.begin(), grad_new.end(), GradType(0));
-  grad_new.resize(num_wf);
+  ratios.resize(num_wf);
   std::fill(ratios.begin(), ratios.end(), PsiValueType(1));
 
   if (wf_list.size() > 1)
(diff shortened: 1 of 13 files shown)


ye-luo:
Test added. Ready for review.

prckent:
const bool tasking in the constructor?

Which compilers, CPUs and GPUs have you tested this with? Where is it expected to work? (We'll see tomorrow).

I think that this is the extent of the async interface that will be needed. Do you agree?

While this is preliminary and developmental, I don't think the input tags matter, so I won't hold this up for discussion. However I think async_execution or similar would be a better name if the tag will persist beyond a development period. Eventually I would prefer that we remove the input tag and disable the feature internally automatically for unsupported architectures.

ye-luo:
> const bool tasking in the constructor?
yes. this feature must be selected when the TWF object is constructed.

> Which compilers, CPUs and GPUs have you tested this with? Where is it expected to work? (We'll see tomorrow).
> 
It is pure CPU thing for the moment, GNU and Clang family should work. GPU code will be connected to tasking with "target nowait".

> I think that this is the extent of the async interface that will be needed. Do you agree?
> 
A short answer is yes, but It is very preliminary. After I settle the async distance table and Jastrow with dependency. It will be more clear.

> While this is preliminary and developmental, I don't think the input tags matter, so I won't hold this up for discussion. However I think async_execution or similar would be a better name if the tag will persist beyond a development period. Eventually I would prefer that we remove the input tag and disable the feature internally automatically for unsupported architectures.

When the async code gets closer to production, we will have a better understanding and we can pick a better name.

src/QMCWaveFunctions/TrialWaveFunction.cpp
```diff
@@ -38,14 +38,15 @@ typedef enum
 
 static const std::vector<std::string> suffixes{"V", "VGL", "accept", "NLratio", "recompute", "buffer", "derivs"};
 
-TrialWaveFunction::TrialWaveFunction(const std::string& aname)
+TrialWaveFunction::TrialWaveFunction(const std::string& aname, bool tasking)
```
prckent, line 41: const?

ye-luo: const matters more for reference type.
Here is just a scalar and passed in by value and local to the function, I don't really feel the need to add const.
We have many integer arguments in the code, I don't feel the need to update them by adding this type of const.

prckent (approved):

PR #2752, opened 2020-10-13
Add an option to turn off branching in DMC batched driver

ye-luo:
## Proposed changes
Mostly needed for benchmark purpose.
`<parameter name="branching"> no </parameter>` to turn off branching.

- New feature

- No

Bora

## Checklist


prckent (changes requested):
This is a dangerous developer/debug only option and should be labeled as such, e.g. "debug_disable_branching" (notice I flipped the meaning yes/no). It should also go in the documentation so that the documentation remains a faithful description of working input.

ye-luo:
Use string type for the flag at the moment. Due to #2758.
@Hyeondeok-Shin has started consolidating the input handling at the bottom level.

prckent (approved):

ye-luo:
We will need a "no branching DMC" feature with correct energy implemented but it is different from "debug_disable_branching".

PR #2754, opened 2020-10-14
Create qmcio_hdf and qmcio_xml cmake targets

ye-luo:
## Proposed changes
Create qmcio_hdf and qmcio_xml cmake targets by splitting qmcutils which is overcrowded.
qmcio_hdf is for source files previously at folder src/io. Files are now moved to src/io/hdf
qmcio_xml is for source files previously at folder src/OhmmsData. Files are now moved to src/io/xml
qmcio target is just an interface target including both qmcio_hdf and qmcio_xml.

Additional cleaning at src/Platforms
Platforms/accelerators.cpp has been moved to QMCApp because it is part of the app logic not the platforms building block.
tests in src/Platforms/XXX/tests are moved to src/Platforms/tests/XXX.

- Build related changes

- No

Bora

## Checklist


src/Platforms/tests/CPU/test_e2iphi.cpp
```diff
@@ -14,7 +14,7 @@
 
 #include <stdio.h>
 #include <string>
-#include "e2iphi.h"
+#include <CPU/e2iphi.h>
```
prckent, line 17: Quotes here?

src/Platforms/tests/OpenMP/test_class_member.cpp
```diff
@@ -15,8 +15,8 @@
 #include <memory>
 #include <vector>
 #include <iostream>
-#include "omp.h"
-#include "OMPallocator.hpp"
+#include <omp.h>
+#include <OpenMP/OMPallocator.hpp>
```
prckent, line 19: Quotes again?

src/QMCDrivers/CloneManager.cpp
```diff
@@ -21,7 +21,7 @@
 #include "Message/Communicate.h"
 #include "Message/OpenMP.h"
 #include "Utilities/IteratorUtility.h"
-#include <qmc_common.h>
+#include <Utilities/qmc_common.h>
```
prckent, line 24: Quotes again? How is this different from the usage in qmcapp.cpp

prckent (commented):
Seems like a good direction.

A number of our includes are now included via <dir/file.h> and not "dir/file.h". What is your logic?

src/Platforms/tests/CPU/test_e2iphi.cpp
ye-luo: Will be fixed in a separate PR.

src/QMCDrivers/CloneManager.cpp
ye-luo: Will be fixed in a separate PR.

src/Platforms/tests/OpenMP/test_class_member.cpp
ye-luo: `omp.h` is from compiler or OS. So use angle bracket.
`OpenMP/OMPallocator.hpp` is from QMCPACK, so should be quotes.
I fixed this place in this PR but need to fix other files in the repo using a script in a separate PR.

prckent (approved):
OK, good to further improve in a later PR. No need to iterate on this one.

PR #2759, opened 2020-10-15
Change QMCPACK source #include from angle brackets to quotes

ye-luo:
## Proposed changes
1. Change QMCPACK source #include from angle brackets to quotes
2. When XXX.h and XXX.cpp are in the same folder (they should), XXX.cpp should `#include "XXX.h"` without any path.

- No

Bora

## Checklist


prckent (approved):
LGTM

Hopefully we won't need the script again

src/Estimators/CSEnergyEstimator.cpp
```diff
@@ -14,7 +14,7 @@
 //////////////////////////////////////////////////////////////////////////////////////
 
 
-#include "Estimators/CSEnergyEstimator.h"
+#include "CSEnergyEstimator.h"
```
PDoakORNL, line 17: "Estimators/CSEnergyEstimator.h"

src/Estimators/TraceManager.cpp
```diff
@@ -14,7 +14,7 @@
 #if !defined(DISABLE_TRACEMANAGER)
 
 
-#include <Estimators/TraceManager.h>
+#include "TraceManager.h"
```
PDoakORNL, line 17: Estimators/TraceManager.h

src/AFQMC/Drivers/AFQMCDriver.cpp
```diff
@@ -7,10 +7,10 @@
 #include "OhmmsData/ParameterSet.h"
 #include "OhmmsData/libxmldefs.h"
 #include "Configuration.h"
-#include <Utilities/qmc_common.h>
+#include "Utilities/qmc_common.h"
 
 #include "AFQMC/config.h"
-#include "AFQMC/Drivers/AFQMCDriver.h"
+#include "AFQMCDriver.h"
```
PDoakORNL, line 13: AFQMC/Drivers/AFQMCDriver.h

src/AFQMC/Drivers/DriverFactory.cpp
```diff
@@ -7,7 +7,7 @@
 #include "mpi3/communicator.hpp"
 
 #include "AFQMC/Utilities/taskgroup.h"
-#include "AFQMC/Drivers/DriverFactory.h"
+#include "DriverFactory.h"
```
PDoakORNL, line 10: AFQMC/Drivers/DriverFactory.h

PDoakORNL (changes requested):
Maybe a change in the regular expression is needed?

src/Estimators/CSEnergyEstimator.cpp
ye-luo: This is on purpose.
CSEnergyEstimator.cpp should include CSEnergyEstimator.h without any directory if they are in the same directory.

PDoakORNL: https://qmcpack.readthedocs.io/en/develop/developing.html#includes
That's not our published coding standard.
All my new code explicitly qualifies internal include location under src

PDoakORNL: I believe "" are looked for in the entire non "system" include path set, this standard is usually thought to remove sloppy dependence on that.

ye-luo: https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html
"" looks for the directory of the cpp file first and guarantees matching h an cpp.
I think we need to update the coding standard.

PDoakORNL: I am now in agreement especially re goal of cmake targets determining include paths.

PR #2761, opened 2020-10-16
Remove unused multi walker API

ye-luo:
## Proposed changes
They are not needed in near term and easy to reintroduce. So delete them to reduce the maintenance.

- Other (please describe): delete things not needed.

- No

Bora

## Checklist


PDoakORNL (approved):
Looks good to me.

PR #2764, opened 2020-10-18
Update cray-libsci handling in CMake

ye-luo:
## Proposed changes
FindLAPACK handles correctly BLAS/LAPACK passed in via compiler wrappers by marking the search successful and leaving entries just empty. So we don't need to write bypass logic in our cmake. By removing that, we gain the capability of overriding BLAS/LAPACK from cray compiler wrappers. Due to Cray moving to newer Modules environment, the old way of unload cray-libsci becomes not feasible.

This PR also includes a small optimization beneficial overall and actually seems to fix issues of using Cray compiler.

- Optimization and compiler fix
- Build related changes

- No

Tulip.
Cori. Theta is not tested due to maintenance but the software stack is very close to Cori. In general, we use ENABLE_MKL anyway.
Bora.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -434,11 +434,6 @@ ELSE(CMAKE_TOOLCHAIN_FILE)
   ENDIF(ENABLE_MKL)
   IF ( HAVE_MKL )
     # Skip adding Math::BLAS_LAPACK target which has been imported by FindMKL.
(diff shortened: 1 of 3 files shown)


prckent (approved):
I assume the unrelated pinned allocator change was intentional.

PR #2769, opened 2020-10-22
Deterministic spin density tests

jtkrogel:
## Proposed changes

Add first deterministic tests for estimators.  Spin density is the first target.

This PR also fixes a bug introduced with the blanket use of symlinks in tests.  When those were introduced, symlinks were created only for files, but not directories.  Estimator tests depend on directory contents and so would generally fail.

- Tests

- No

Laptop

## Checklist


markdewing:
okay to test

prckent:
Q. How far do we think this is from being mergeable? We need the file linking fix for the release. Some deterministic density tests would also be excellent but we will need a few days to verify they run correctly on all our supported platforms.

jtkrogel:
This can be merged now.  It contains the first test.  Others can be added later.

CMake/macros.cmake
```diff
@@ -30,10 +30,13 @@ ENDFUNCTION()
 FUNCTION( COPY_DIRECTORY_USING_SYMLINK SRC_DIR DST_DIR )
     FILE(MAKE_DIRECTORY "${DST_DIR}")
     # Find all the files but not subdirectories
-    FILE(GLOB FILE_ONLY_NAMES LIST_DIRECTORIES FALSE "${SRC_DIR}/*")
+    FILE(GLOB FILE_ONLY_NAMES LIST_DIRECTORIES TRUE "${SRC_DIR}/*")
     FOREACH(F IN LISTS FILE_ONLY_NAMES)
-      #MESSAGE("Creating symlink from  ${F} to directory ${DST_DIR}")
-      EXECUTE_PROCESS( COMMAND cp -ds --remove-destination -t . "${F}" WORKING_DIRECTORY ${DST_DIR})
+      GET_FILENAME_COMPONENT(FL ${F} NAME)
+      IF(NOT EXISTS "${DST_DIR}/${FL}")
```
ye-luo, line 36: The old code has overwriting capability. Skipping if already exist changes the intended behaviour. Need to check if the existing softlink is the intended or not. If it is, skip is safe. Otherwise, need to overwrite the old one.

ye-luo (changes requested):

CMake/macros.cmake
jtkrogel: Under what circumstances could an existing softlink created by CMake be something other than what is intended?  How would you detect this?

ye-luo: The pre-existing file can be a manually created file instead of a softlink created by cmake. When I create a test, I usually work on the input file in a test folder under build folder. Once everything is correct, I add it to the source folder. Then the collision may happen. In the setting, overwriting happens.

ye-luo: google shows 'readlink' can return the source of the softlink.

jtkrogel: So the logic you want is:
```
if not path_exists
    create_symlink
elif path_is_not_a_symlink
    remove_file_at_path
    create_symlink  
``` 
?

ye-luo: > So the logic you want is:
> 
> ```
> if not path_exists
>     create_symlink
> elif path_is_not_a_symlink
>     remove_file_at_path
>     create_symlink  
> ```
> 
> ?

should be good for me.

prckent: I don't understand the concerns here and would like to:

ye-luo: > I don't understand the concerns here and would like to:

This PR also softlink folders and need special treatment to avoid softlink of folders inside folders (qmc_ref/qmc_ref) in a second cmake invocation.

prckent: There are a few scenarios for runninng/rerunning cmake.

1. In a fresh run (empty build directory), we need to make fresh links. I believe "ln -sf" is fine for this.

2. If cmake is run again in an existing directory, it would be ideal to do nothing. e.g. if link for file or directory exists, skip, otherwise make new link. (We are trying to minimize slow writes on the filesystem)

3. If someone has changed a link in the build directory to a local file or has made a different link, then they are running some risk. This person is clearly a developer and can cope with whatever we do provided we are consistent. No standard user will hit this scenario.

4. I did not think of a 4th scenario.

In bash this is only:

if [ ! -e $the_file ];    # only check if file exists, independent of type
ln -sf $sourcedir/$the_file .
fi

@ye-luo , @jtkrogel Are you OK with these behaviors? Are there other scenarios?

ye-luo: point 3. the desired behavior is to have cmake ensuring self consistency as much as possible. Invoking `cmake .` should restore test files generated by cmake. Invoking cmake is much cheaper than deleting the build folder for developers. I don't expect developers aware of this special 'skip' treatment deeply hidden. So the `elif` logic is desired. cmake has `if(IS_SYMLINK` support.

ye-luo: Version 1 overwriting symlink as well.
```CMake
      IF(EXISTS "${DST_DIR}/${FL}" AND NOT IS_SYMLINK "${DST_DIR}/${FL}")
        FILE(REMOVE_RECURSE "${DST_DIR}/${FL}")
      ENDIF()
      EXECUTE_PROCESS( COMMAND ln -sf "${F}" "." WORKING_DIRECTORY ${DST_DIR})
```
Version 2 skip overwriting if symlink. less secure but faster.
```CMake
      IF(NOT EXISTS "${DST_DIR}/${FL}")
        EXECUTE_PROCESS( COMMAND ln -sf "${F}" "." WORKING_DIRECTORY ${DST_DIR})
      ELSEIF(NOT IS_SYMLINK "${DST_DIR}/${FL}")
        FILE(REMOVE_RECURSE "${DST_DIR}/${FL}")
        EXECUTE_PROCESS( COMMAND ln -sf "${F}" "." WORKING_DIRECTORY ${DST_DIR})
      ENDIF()
```
@jtkrogel using '.' as target avoids 'qmc_ref/qmc_ref' problem.

ye-luo: It seem that we only need the following and no additional case handling is needed.
```
EXECUTE_PROCESS( COMMAND ln -sf "${F}" "." WORKING_DIRECTORY ${DST_DIR})
```
if the pre-existing name in DST_DIR is a normal file, it is overwritten.
If the pre-exisitng name in DST_DIR is a directory, An error is printed 'cannot overwrite directory'

jtkrogel: I'm leery of introducing `rm -rf` anywhere.  I will do this via system `rm` for single files, if that is OK.

prckent: OK. We can test whatever is done. There should be errors on initial or subsequent invocation.

ye-luo: > It seem that we only need the following and no additional case handling is needed.
> 
> ```
> EXECUTE_PROCESS( COMMAND ln -sf "${F}" "." WORKING_DIRECTORY ${DST_DIR})
> ```
> 
> if the pre-existing name in DST_DIR is a normal file, it is overwritten (-f behavior).
> If the pre-exisitng name in DST_DIR is a directory, An error is printed 'cannot overwrite directory'

I also prefer not using `rm` at all, this last solution doesn't it.

prckent: @jtkrogel Please do something you are happy with - perhaps close to what you initially put :-)  I agree with no rms. Please avoid errors on rerun of cmake. Acceptance of errors is something not to be normalized.

jtkrogel: Sounds good.  I will post final updates first thing tomorrow.

jtkrogel: Done.

ye-luo (approved):

PR #2772, opened 2020-10-28
Improve MKL auto-detection

ye-luo:
## Proposed changes
The convenient option `-mkl` in Intel compiler is not portable to other compilers. Better to use more standardized way of handling header file inclusion and linking. This PR brings the MKL library file handling by FindLAPACK. Once MKL library files are detected, our FindMKL is used to complete header file detection, FFT and vector math.

I was frequently stopped by our cmake setting when mkl is detected but ENABLE_MKL was not set. Now there is no more need of ENABLE_MKL option. Using alternative BLAS/LAPACK can be handled -DBLA_VENDOR provided by FindBLAS and FindLAPACK.

[WIP] for manual update needed.

- Build related changes

- No

Bora with external MKL, Dell laptop with ubuntu 20.04 apt shipped MKL.
Theta with CrayPE.

## Checklist


PDoakORNL (approved):
This looks good to me especially in elimination of so much branching.  The general modernization of build options is a good addition to.  I will see how this works with my toolchain as that's really the proof.  Is some cmake version dependence for findLAPACK?

ye-luo:
FindLAPACK from cmake 3.10 is new enough for our needs. FindLAPACK has better ARM support since 3.18 but we don't have anything setup for ARM.

prckent (changes requested):
Adding a protective block to make sure we don't merge this without careful consideration of the upcoming release.

This generally looks good to me but since it is a configuration related change we need to carefully check: gcc+mkl gcc+mkl with other libraries present, cori/Cray wrappers, and we also need to check what happens with a 2 year old cmake as well as the latest version

ye-luo:
> Adding a protective block to make sure we don't merge this without careful consideration of the upcoming release. And the docs updates are essential.
> 
> This looks good to me but since it is a configuration related change we need to carefully check: gcc+mkl gcc+mkl with other libraries present, cori/Cray wrappers, and we also need to check what happens with a 2 year old cmake as well as the latest version. I think these should all be fine, but we need to check.

I checked cmake down to 3.10 (3.12 in 2 year window).
Intel/Clang/GNU+MKL threaded works well both correctness and performance.
Cray systems Theta and Cori are tested.

What are "other libraries" with gcc+MKL? Did you mean using another BLAS/LAPACK even MKL is present in the environment?

before this PR, even with ENABLE_MKL=0 FindLAPACK may still find MKL and then abort and ask users to opt-in MKL. However, there is no simple way to opt out. Users have to remove MKL from search paths.
With this PR, users may use -DBLA_VENDOR=OpenBLAS or -DBLA_VENDOR=Generic to opt out MKL even it is on the search path.

prckent (commented):
Minor Q. Does cmake findlapack select MKL preferentially ahead of (say) OpenBLAS?  If not we'll need to specify use of MKL via BLAS_VENDOR in the test scripts.

ye-luo:
> Minor Q. Does cmake findlapack select MKL preferentially ahead of (say) OpenBLAS? If not we'll need to specify use of MKL via BLAS_VENDOR in the test scripts when using non-Intel compilers.

If you source mkl script which sets envIronment variable MKLROOT, MKL will be found ahead regardless of what you have in the system. I cleaned up test scripts already.

prckent:
Thanks Ye. I am extra paranoid about changes like this near a release. All good.

Is there anything you want to do before removing the WIP?

ye-luo:
> paranoid

I'm on your side. I had the same consideration as you have. It is your call to decide postpone merging after release or do it before. Since I striped out all the ENABLE_MKL in the repo including CMake, docs and scripts. This PR is ready to go.

prckent (approved):

prckent:
I think due diligence has been done on this. There is the L2 PR and perhaps some of the new deterministic tests to go in ahead of the release, so we will have some time to catch problems.

PR #2779, opened 2020-11-05
Raise cmake minimum to 3.12

ye-luo:
## Proposed changes
3.10 and 3.11 doesn't work well with object target. 3.12 works fine. 3.12 is the older version within 2 year window. So requires 3.12 now.

- Build related changes
- No

ryzen-box

## Checklist


ye-luo:
Hold a second, let me test more carefully 3.12.

ye-luo:
Good to go. Checked 3.12 to 3.14.

prckent (approved):
This version change is consistent with prior discussions. The "old" cmake used in the nightlies has been 3.12.3 for a couple of weeks.

PR #2788, opened 2020-11-12
Enable C++17 by default

ye-luo:
## Proposed changes
Raise CXX version to 17.
GCC minimum is raised to 7.

- [x] an explicit check of C++17 standard library.
- [x] update documentation.

@mmorale3 Do we need C++17 for nvcc?

- Build related changes

- No

ryzen-box

## Checklist


CMakeLists.txt
```diff
@@ -110,20 +110,7 @@ MESSAGE( STATUS "CMAKE_BUILD_TYPE is ${CMAKE_BUILD_TYPE}")
...
-# XYZ_HOME/lib
-# IF XYZ library is found, XYZ_FOUND is set to true
-#
-# find_package is used with cmake 2.8 and higher
-######################################################################
-#if(APPLE)
-#  set(QMC_INCLUDE_PATHS ${QMC_INCLUDE_PATHS} /sw/include)
-#  set(QMC_LIBRARY_PATHS ${QMC_LIBRARY_PATHS} /sw/lib)
-#endif(APPLE)
+SET(QMC_CXX_STANDARD 17 CACHE STRING "QMCPACK C++ language standard")
```
prckent, line 113: Unless you have a specific use for this, please no more variables.

ye-luo: I really prefer this an option rather than having it hard-coded.
I still need 14 for IBM XL for testing purpose.
Also experiment 20 when the time comes.
Editing CMakeLists.txt is too much interruption.
I can add a warning saying this there is no support if the value is not 17.

ye-luo:
Test this please.

mmorale3:
I do not need c++17 for nvcc, keeping nvcc requirements at 14 is fine.

ye-luo:
Take this PR offline because AFQMC + multi needs some fix and current CI brings the CI node fully offline.

PR #2794, opened 2020-11-17
More traps for bad combination of Intel and libstdc++

ye-luo:
## Proposed changes

Based on #2790 investigation.

- Build related changes

- No

CI node

## Checklist


prckent (approved):
Thanks for chasing these dependencies.

PR #2796, opened 2020-11-18
Clean up LCAOrbitalSet and fixes hdf5 io error handling

ye-luo:
## Proposed changes

1. Put most part of LCAOrbitalSet under private and add more checks.
2. fixed hdf5 io error handling, runtime_error was not thrown.

- Bugfix

- No

ryzen-box

## Checklist


anbenali (commented):
Because you remove completely NCA, NEA NCB and NEB as well as the number of orbitals in the readDetListH5(), maybe it would be nice to have that information printed in the output (?). For now (and maybe because it is still in debug use) it is extra reassuring to see that I am having the correct states I am supposed to get...

src/QMCWaveFunctions/LCAO/AOBasisBuilder.h
```diff
@@ -171,8 +171,9 @@ bool AOBasisBuilder<COT>::putH5(hdf_archive& hin)
     hin.read(Normalized, "normalized");
     hin.read(Morder, "expandYlm");
     hin.read(basisName, "name");
-    hin.read(basisType, "type");
-    hin.read(addsignforM, "expM");
+    // Ye to Anouar: do we need these inputs? They are missing in current h5
+    //hin.read(basisType, "type");
```
anbenali, line 175: Not really (for now). But it is just because there is specialization for GAMESS and Pyscf... I mirrored the XML. If future codes (outside of the ones from molden, have something else we can always revisit. So It can be removed for now.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -1214,10 +1244,6 @@ bool SlaterDetBuilder::readDetListH5(xmlNodePtr cur,
   std::string Dettype = "DETS";
   ValueType sumsq     = 0.0;
   OhmmsAttributeSet spoAttrib;
-  spoAttrib.add(NCA, "nca");
```
anbenali, line 1217: Because you remove completely NCA, NEA NCB and NEB as well as the number of orbitals in the readDetListH5(), maybe it would be nice to have that information printed in the output (?). For now (and maybe because it is still in debug use) it is extra reassuring to see that I am having the correct states I am supposed to get...

src/QMCWaveFunctions/LCAO/RadialOrbitalSetBuilder.h
```diff
@@ -240,8 +240,9 @@ bool RadialOrbitalSetBuilder<COT>::addGridH5(hdf_archive& hin)
     ri = tt;
     hin.read(tt, "grid_rf");
     rf = tt;
-    hin.read(tt, "rmax_safe");
-    rmax_safe = tt;
+    // Ye to Anouar: do we need this input? It is missing in current h5
+    //hin.read(tt, "rmax_safe");
+    //rmax_safe = tt;
```
anbenali, line 245: No this should actually be moved outside. no grid should be hard coded in HDF5 as there is no way of setting it properly without hard coding it.

anbenali (approved):
I

src/QMCWaveFunctions/LCAO/AOBasisBuilder.h
ye-luo: OK. I will remove them.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
ye-luo: NCA, NEA NCB and NEB input are only kept for XML only case. They are removed in H5 case.
I will add a printing out in the h5 case.

src/QMCWaveFunctions/LCAO/RadialOrbitalSetBuilder.h
ye-luo: I need to change it as TODO. This PR has not touched adding back the grid control in the h5 route.

prckent:
@ye-luo Please add TODO as an issue.
@anbenali OK otherwise?

ye-luo:
#2797 opened to track the residual work.

anbenali (approved):

PR #2802, opened 2020-11-24
Nexus: fix module overwrite bug in qmca

jtkrogel:
## Proposed changes

The `--save_average` option in `qmca` was broken due to a local namespace collision with the globally assigned name for the `numpy` module (imported as `np`).  The fix is to use a different local variable name.

- Bugfix

- No

Laptop

## Checklist


ye-luo:
Okay to test

ye-luo (approved):

PR #2803, opened 2020-11-24
Add MultiSlaterDeterminant unit test

ye-luo:
## Proposed changes
A preliminary test. Covering both slow and table method.

- New feature

- No

Ryzen-box

## Checklist


prckent (commented):
This is more of a mid-scale integration test than a unit test. It uses a sizable multideterminant LCAO wavefunction. If we used a very simple wavefunction and small multideterminant expansion we could potentially check it "by hand".

Is there a way to create a small multideterminant wavefunction with less setup? What is missing or blocking?

ye-luo:
It does use slightly larger MSD data set with 1487 dets but it is still unit test. Testing the WaveFunctionCompoment level individual calls. The h5 file is already there and thus needs less work to setup.

1. Creating a further smaller MSD WF may help us check against hand calculated numbers but might create too simple test case.
2. Invoking the builder requires less setup in the unit test.
3. Although the current won't guarantee 100% correctness but should be robust against code changes.

prckent:
I agree that this is more of a mid-level deterministic test. Certainly useful to have.

prckent (changes requested):
1. Please document "input style 1". Should be descriptive.

2. Where did the check values come from? (Ideally we would have non-table method and table method results in agreement...)

ye-luo:
>     1. Please document "input style 1". Should be descriptive.
> 
>     2. Where did the check values come from? (Ideally we would have non-table method and table method results in agreement...

1. Input style is the preferred input style. I need to review all the documentation about SPOSet input style. It will be a separate task.
2. both table and non-table method are checked in this unit test. They already agree here.

prckent:
Please update the text here to state that the orbitals and determinant sets are separate. I don't wish for us to get into the business of defining multiple input styles when we have a plan to only support one in future.

prckent (approved):

PR #2807, opened 2020-12-07
Guard Clang C++17 and AFQMC in cmake.

ye-luo:
## Proposed changes
Guard a bad triple. Upcoming 11.0.1 will get the issue fixed.

- Build related changes

- No

Bora

## Checklist


ye-luo:
Based on #2790 investigation.

prckent (approved):

PR #2816, opened 2020-12-16
Refactor MCWalkerConfiguration

ye-luo:
## Proposed changes
MCWalkerConfiguration was overloaded. It is both a particleset and a list of lightweight walkers. The later role has been moved to WalkerConfigurations which is the object carried between QMC sections and restart.
the batched drivers will only interact with WalkerConfigurations instead of the whole MCWalkerConfiguration.

- No

Ryzen-box

## Checklist

### DIFF
--- a/src/Particle/WalkerConfigurations.h
+++ b/src/Particle/WalkerConfigurations.h
@@ -1,0 +1,201 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2020 QMCPACK developers.
+//
+// File developed by: Jordan E. Vincent, University of Illinois at Urbana-Champaign
+//                    Ken Esler, kpesler@gmail.com, University of Illinois at Urbana-Champaign
+//                    Jeremy McMinnis, jmcminis@gmail.com, University of Illinois at Urbana-Champaign
+//                    Jeongnim Kim, jeongnim.kim@gmail.com, University of Illinois at Urbana-Champaign
+//                    Cynthia Gu, zg1@ornl.gov, Oak Ridge National Laboratory
+//                    Raymond Clay III, j.k.rofling@gmail.com, Lawrence Livermore National Laboratory
+//                    Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//                    Mark A. Berrill, berrillma@ornl.gov, Oak Ridge National Laboratory
+//
+// File created by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+/** @file WalkerConfigurations.h
+ * @brief Declaration of a WalkerConfigurations
+ */
+#ifndef QMCPLUSPLUS_WALKERCONFIGURATIONS_H
+#define QMCPLUSPLUS_WALKERCONFIGURATIONS_H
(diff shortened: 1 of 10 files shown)


src/Particle/WalkerConfigurations.h
```diff
@@ -0,0 +1,201 @@
...
+ *
+ * The quantities are shared by all the nodes in a group
+ * - NumSamples number of samples
+ * - Weight     total weight of a sample
+ * - Energy     average energy of a sample
+ * - Variance   variance
+ * - LivingFraction fraction of walkers alive each step.
+ */
+template<typename T>
+struct MCDataType
```
PDoakORNL, line 41: Postfix Type is unnecessary naming convention tells us MCData is a type.

src/Particle/WalkerConfigurations.h
```diff
@@ -0,0 +1,201 @@
...
+  T LivingFraction;
+};
+
+/** A set of light weight walkers that are carried between driver sections and restart
+ */
+class WalkerConfigurations
+{
+public:
+  /// walker type
+  using Walker_t         = Walker<QMCTraits, PtclOnLatticeTraits>;
```
PDoakORNL, line 60: I'd rather see WCWalker or similar instead of the trailing _t

src/Particle/WalkerConfigurations.h
```diff
@@ -0,0 +1,201 @@
...
+  using const_iterator = WalkerList_t::const_iterator;
+
+  /** starting index of the walkers in a processor group
+   *
+   * WalkerOffsets[0]=0 and WalkerOffsets[WalkerOffsets.size()-1]=total number of walkers in a group
+   * WalkerOffsets[processorid+1]-WalkerOffsets[processorid] is equal to the number of walkers on a processor,
+   * i.e., W.getActiveWalkers().
+   * WalkerOffsets is added to handle parallel I/O with hdf5
+   */
+  std::vector<int> WalkerOffsets;
```
PDoakORNL, line 76: The idea of "offsets" is not used in the new population control. But that can be dealt with when this is used with the unified driver.

src/Particle/WalkerConfigurations.h
```diff
@@ -0,0 +1,201 @@
...
+  void destroyWalkers(int nw);
+
+  ///clean up the walker list and make a new list
+  void resize(int numWalkers, size_t numPtcls);
+
+  ///return the number of active walkers
+  inline size_t getActiveWalkers() const { return WalkerList.size(); }
+  ///return the total number of active walkers among a MPI group
+  inline size_t getGlobalNumWalkers() const { return GlobalNumWalkers; }
+  ///return the total number of active walkers among a MPI group
```
PDoakORNL, line 124: This collides with some of MCPopulations responsibilities and I think that the GlobalNumWalkers state is out of scope for this class which I would like to see as a clean data object of rank scope.  We shouldn't be shy about cleaning this up in the future.

PDoakORNL (approved):
Review comments should be dealt with in the future but this can go in as it is.  This is a nice refactor from MCWalkerConfigurations of what is needed and useful.

ye-luo:
I will deal with the renaming within a follow-up PR today.
Currently the added class holds a list of local walkers and the info about the mount of walkers globally. I agree that they are two set of info although they are often used together when walkers are manipulated. We may consider split them when a need shows up.

ye-luo:
Tried renaming and a lot of files are touched. So the renaming will be deferred to a separate PR.

PR #2817, opened 2020-12-17
Move MCPopulation ownership into drivers

ye-luo:
## Proposed changes
Following #2816, walker configurations (lightweight walkers) are owned beyond drivers.
Fat walkers (MCPopulation) which are driver dependent. Current MCPopulation is good for VMC and DMC need. So MCPopulation ownership is moved into batched VMC/DMC drivers.
MCPopulation has a reference to the list of lightweight walkers once being constructed.
When a driver, as well as the owned MCPopulation, is destructed, the coordinates and weights are stored back to walker configuration list.

- No

Ryzen-box

## Checklist


PDoakORNL (approved):
Looks good to me.

PR #2818, opened 2020-12-17
Fix call to GradCost in complex build

markdewing:
In the complex build, the call to `GradCost` in `testDerivOptimization` was calling the empty base class version `QMCCostFunctionBase` due to a parameter mismatch.  The parameter gradients would come out as all zero.
Add 'override' to catch the issue, fix the signature, and add more override's to keep clang happy.

Change a parameter to const and pass-by-reference in `testDerivOptimization`.
Increase the number of digits printed out in the parameter gradient file.

Add const to the `print` function in `VariableSet` (from debugging a Jastrow optimization issue).

- Bugfix
- 

- No

laptop

## Checklist

### DIFF
--- a/src/Optimize/VariableSet.cpp
+++ b/src/Optimize/VariableSet.cpp
@@ -233,7 +233,7 @@ void VariableSet::setDefaults(bool optimize_all)
     Index[i] = optimize_all ? i : -1;
 }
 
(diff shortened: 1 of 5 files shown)


prckent (approved):
LGTM.

OK to merge provided the CI passes.

Interesting that this didn't result in an error in the past. Can you run clang with enough warnings enabled to catch any other instances?

PR #2820, opened 2020-12-18
Restrict use of last_driver and clean up the life cycle of driver objects

ye-luo:
## Proposed changes
Once a QMC section execution is completed, the driver is saved in last_driver object.
On a following QMC section, a new driver is created and last_driver is only release when it is being overwritten again.
In this PR, last_drive is only kept when a QMC input section is in a loop and gets cleaned up when the loop finishes.
If not a loop, a QMC driver object is destructed once that section execution is completed.

- Build related changes

- No

ryzen-box

## Checklist

### DIFF
--- a/src/QMCApp/QMCMain.cpp
+++ b/src/QMCApp/QMCMain.cpp
@@ -331,7 +331,7 @@ void QMCMain::executeLoop(xmlNodePtr cur)
       tcur = tcur->next;
     }
   }
-  // Clean up the last driver. No further reuse needed.
+  // Destroy the last driver at the end of a loop with no further reuse of a driver needed.
   last_driver.reset(nullptr);
 }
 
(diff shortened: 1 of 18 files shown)


src/QMCApp/QMCMain.cpp
```diff
@@ -331,6 +331,8 @@ void QMCMain::executeLoop(xmlNodePtr cur)
       tcur = tcur->next;
     }
   }
+  // Clean up the last driver. No further reuse needed.
```
PDoakORNL, line 334: I think it would be better to say. Last driver is destroyed, necessary state having been transfered.  Or something like that.

src/QMCWaveFunctions/Fermion/DiracDeterminantCUDA.h
```diff
@@ -149,22 +149,32 @@ class DiracDeterminantCUDA : public DiracDeterminantBase
 
 public:
   // safe-guard all CPU interfaces
-  DiracDeterminantCUDA* makeCopy(SPOSet* spo) const { APP_ABORT("Calling DiracDeterminantCUDA::makeCopy is illegal!"); }
+  DiracDeterminantCUDA* makeCopy(SPOSet* spo) const
+  {
+    APP_ABORT("Calling DiracDeterminantCUDA::makeCopy is illegal!");
```
PDoakORNL, line 154: As per #125 I think this should be a throw. As well as the following APP_ABORTS.

PDoakORNL (changes requested):
Looks like another good step.
Just don't like to see new use of the APP_ABORT macro.

src/QMCApp/QMCMain.cpp
ye-luo: Updated.

src/QMCWaveFunctions/Fermion/DiracDeterminantCUDA.h
ye-luo: Changed.

PDoakORNL (approved):

PR #2821, opened 2020-12-21
Add a scoped profiler to allow profiling target specific QMC drivers

ye-luo:
## Proposed changes
Review after #2820. Since #2820, a driver life cycle is better managed and we can introduce scoped profiling to target specific drivers.
See more detail in the doc update about `profiling="yes"` feature introduced for QMC drivers.

check Nsight Systems for GPU profiling.
Update VTune for CPU profiling. VTUNE_ROOT is introduced in CMake.

- New feature
- Build related changes

- No

epyc-server

## Checklist


PDoakORNL (approved):
LGTM

PR #2822, opened 2020-12-22
Fix NLPP derivatives

markdewing:
The `mw_evaluateWithParameterDerivatives` code needs to handle `dlogpsi` as an input, and `dhpsioverpsi` as an output (more precisely, an input that needs to be updated)

For the batch cost function in `checkConfigurations`, the `includeNonLocalH` value needs to be passed to the lambda so the NLPP value can be removed from the fixed energy when present.  Unfortunately, most of the diff is due to formatting changes from clang-format.

For the correlated sampling part, `compute_nlpp` needs to be passed to the lambda, and the branch that calls `QMCHamiltonain::flex_evaluateValueAndDerivatives` added.

- Bugfix

- No

laptop

## Checklist


prckent (approved):
LGTM

PR #2823, opened 2020-12-22
Fix memory leak in slaterdet

ye-luo:
## Proposed changes
The driver unit test was failing on graphic card with 8 GB memory, the root cause was that the DiracDeterminant objects are not freed in properly in SlaterDet. The second change is reducing the number of walkers + reserve in unit test from 32+16 to 8+2, the reduction of memory usage allows GPUs with 4GB memory to pass that unit test and it should help out cdash to be more green.

More clean up fix by changing raw pointer to lvalue unique_ptr will come after. Much more affected files and more testing needed.

- Bugfix

- No

## Checklist

### DIFF
--- a/src/QMCDrivers/MCPopulation.h
+++ b/src/QMCDrivers/MCPopulation.h
@@ -199,7 +199,7 @@ public:
 
   UPtrVector<MCPWalker>& get_walkers() { return walkers_; }
   const UPtrVector<MCPWalker>& get_walkers() const { return walkers_; }
-  UPtrVector<MCPWalker>& get_dead_walkers() { return dead_walkers_; }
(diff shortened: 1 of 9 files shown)


prckent (approved):
Two good improvements here.

OK to merge subject to CI passing.

ye-luo:
FYI: Each cuBLAS handle reserves 8.4MB and each cuSolver handle reserves 33.6MB. It needs to managed more carefully and cannot be allocated one per walker if we batch 1k walkers.

prckent:
Good point. 
e.g. We run 1K walkers with the legacy CUDA on small problems.
Need the handles at the crowd level or higher.

PR #2826, opened 2020-12-23
Smart pointers of SPOSet in singledet and multidet

ye-luo:
## Proposed changes
Use smart pointers of SPOSet in single det and multi dets. Reduce potential memory leak.
1. single det class uses shared_ptr. The sharing aspect is only intended for a list of single dets in multi-det slow code path. In most cases, it should be used as unique_ptr. Comments added in the source code.
2. Multi det uses unique_ptr.

- Bugfix

- No

Bora

## Checklist


prckent (approved):

prckent:
(OK to merge provided the CI passes. e.g. Perhaps some issues with CUDA?)

ye-luo:
> (OK to merge provided the CI passes. e.g. Perhaps some issues with CUDA?)

The legacy CUDA code seems fine when I ran deterministic tests by hand. There are performance test failure from last night and I will take a look in the following days.

PR #2827, opened 2020-12-23
Delete pointers owned by ParticleSetPool

ye-luo:
## Proposed changes
Make sure ParticleSet gold copy owned by ParticleSetPool got freed properly.

- Bugfix

- No

bora

## Checklist

### DIFF
--- a/src/Particle/ParticleSetPool.cpp
+++ b/src/Particle/ParticleSetPool.cpp
@@ -32,7 +32,7 @@
 namespace qmcplusplus
 {
 ParticleSetPool::ParticleSetPool(Communicate* c, const char* aname)
-    : MPIObjectBase(c), SimulationCell(nullptr), TileMatrix(0)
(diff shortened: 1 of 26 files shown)


src/Particle/ParticleSetPool.cpp
```diff
@@ -32,7 +32,7 @@
 namespace qmcplusplus
 {
 ParticleSetPool::ParticleSetPool(Communicate* c, const char* aname)
-    : MPIObjectBase(c), SimulationCell(nullptr), TileMatrix(0)
+    : MPIObjectBase(c), TileMatrix(0)
```
prckent, line 35: SimulationCell is no longer initialized with this constructor. Why?

ye-luo: SimulationCell is unique_ptr and initialized as nullptr by default.

prckent (approved):

PR #2830, opened 2020-12-29
Nexus: fix type check for reblock_factors in qmc-fit

jtkrogel:
## Proposed changes

Expand integer type check of reblock_factors input to qmc-fit to include numpy integer types.  This enables correct application of a single user provided reblocking factor to each timestep data series.

This issue was reported by Gopal Iyer.

- Bugfix

- No

Laptop

## Checklist


prckent:
OK to test

ye-luo (approved):

PR #2832, opened 2020-12-31
Add include limits C++ header

ye-luo:
## Proposed changes
As `std::numeric_limits<T>` is used, `<limits>` header file should be included.

- Bugfix

- No

Bora

## Checklist

### DIFF
--- a/external_codes/catch/complex_approx.hpp
+++ b/external_codes/catch/complex_approx.hpp
@@ -2,6 +2,7 @@
 #define CATCH_COMPLEX_APPROX
 
 #include <complex>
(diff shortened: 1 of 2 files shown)


prckent:
(Worth noting that this problem is in our extended catch functionality, not something in upstream).

prckent (approved):
Q. How did this problem surface?

ye-luo:
> Q. How did this problem surface?

It was surfaced when I was playing with GCC master branch. It affects both miniQMC and QMCPACK, it seems to be exposed by changes on the GCC side.

PR #2833, opened 2020-12-31
Fix evaluation of non-optimizable components in flex_evaluateDeltaLog.

markdewing:
If recompute is true (which it is for NLPP derivatives), need to call mw_evaluateLog to ensure the matrix inverses are computed
prior to the NLPP derivative computation.   This needs to done for wavefunction components that were *not* updated in the first part of the function.

The test case (NiO) that shows the problem has a determinant part with no variational parameters.

The cost function has two parts that compute the derivatives - the first one does setup and derivative computation (`checkConfigurations`).  The second one does correlated sampling (`correlatedSampling`), and only computes parts of the wavefunction that change (i.e that have optimizable variational parameters).
The problem doesn't manifest in `checkConfiguations` because evaluateLog is called on all wavefunction components.  The problem shows up in `correlatedSampling` because evaluateLog is not called on all components.  This logic fix ensures evaluateLog is called on all components in the NLPP case.

- Bugfix

- No

laptop, desktop

## Checklist


prckent (approved):
Good catch.

PR #2834, opened 2021-01-01
Two-body jastrow optimization bug fix : Set opt offset to be equal to first active variable

### LINKED ISSUE #2814: Freezing two-body Jastrow parameters does not work
At @anbenali's request, I am opening an issue. Many thanks to @markdewing for helping with the troubleshooting.

**Describe the bug**
When a optimizing a two-body Jastrow, setting optimize="no" does not work.
When optimize is yes for both parameters, the optimization works.
If optimize="no" comes before optimize="yes", the optimization fails.
If optimize="no" comes after optimize="yes", the optimization works.

**To Reproduce**
1. release version or git commit hash being built
`a6aaf95d3461e13751f7d38b5dc1ed52ca3618e6-dirty`
2. cmake command
` cmake -DENABLE_MKL=1 ..`
3. full program/test invocation command
`qmcpack Default.qmc.in-wfj.xml`
4. additional steps
Full input files [here](https://gist.github.com/shivupa/08cce8c1383132d5f8f84cb48183861d)

An H2O example with a two-body Jastrow u-u, u-d, d-d = u-u with two parameters each (starting at 0.0, but this doesn't matter):

```
    <jastrow name="J2" type="Two-Body" function="Bspline" print="yes">
      <correlation rcut="10" size="2" speciesA="u" speciesB="u">
        <coefficients id="uu" type="Array" optimize="no" >0.0 0.0 </coefficients>
      </correlation>
      <correlation rcut="10" size="2" speciesA="u" speciesB="d">
        <coefficients id="ud" type="Array" optimize="yes" >0.0 0.0 </coefficients>
      </correlation>
    </jastrow>
```

Results in (after several opt cycles `s014.opt.xml`):

```
    <jastrow name="J2" type="Two-Body" function="Bspline" print="yes">
      <correlation rcut="10" size="2" speciesA="u" speciesB="u">
        <coefficients id="uu" type="Array" optimize="no">0.0 0.0 </coefficients>
      </correlation>
      <correlation rcut="10" size="2" speciesA="u" speciesB="d">
        <coefficients id="ud" type="Array" optimize="yes"> 2.251239143 0</coefficients>
      </correlation>
    </jastrow>
```

**Expected behavior**
I would expect both the parameters of the second set to be updated. I have tested with more parameters and only the first parameter is only updated.

**System:**
Laptop Intel  i7-7700HQ

**Additional context**
Following Mark's suggestions, I think I have some idea of where the bug is happening so I'll post what I have found so far.

I think the problem is in `src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h` at line 149, the `OffSet` variable is pointing to the wrong indices:
```
    optimize? (uu/ud):     OffSet:
    ---------------------------------------------------------------
    yes/yes                i 0 OffSet[i].first 0 OffSet[i].second 2
                           i 1 OffSet[i].first 2 OffSet[i].second 4
                           i 2 OffSet[i].first 2 OffSet[i].second 4
                           i 3 OffSet[i].first 0 OffSet[i].second 2
    ---------------------------------------------------------------
    no/yes                 i 0 OffSet[i].first 0 OffSet[i].second 2
                           i 1 OffSet[i].first 1 OffSet[i].second 3
                           i 2 OffSet[i].first 1 OffSet[i].second 3
                           i 3 OffSet[i].first 0 OffSet[i].second 2
    ---------------------------------------------------------------
    yes/no                 i 0 OffSet[i].first 0 OffSet[i].second2
                           i 1 OffSet[i].first -1 OffSet[i].second1
                           i 2 OffSet[i].first -1 OffSet[i].second1
                           i 3 OffSet[i].first 0 OffSet[i].second2
```
The first and third case optimize correctly, but the second case does not. Even for the 1 parameter that is being updated I am not sure if the correct derivatives are being used to update that parameter.

shivupa:
Detailed disussion on issue #2814

## Proposed changes
The two body jastrow optimization would fail when the first optimizable parameters were not the first parameters (e.g. if the u-u terms were frozen). This implements @markdewing's suggestions from #2814, which fixes the offset variable to point at the correct set of parameters. 

Closes #2814 

- Bugfix

- Yes

Laptop Intel i7-7700HQ
Desktop Intel i7-2600S

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
+++ b/src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
@@ -139,7 +139,13 @@ public:
         lapLogPsi[i]  = new ValueVectorType(NumPtcls);
       }
       OffSet.resize(F.size());
-      int varoffset = myVars.Index[0];
+      int varoffset = -1;
+      for (int i = 0; i < myVars.size(); i++)
+      {
+        varoffset = myVars.Index[i];
+        if (varoffset != -1)
+          break;
+      }
       for (int i = 0; i < F.size(); ++i)
       {
         if (F[i] && F[i]->myVars.Index.size())
@@ -172,7 +178,7 @@ public:
         continue;
       if (active.recompute(kk))
         recalculate = true;
-      rcsingles[k] = true;
+      rcsingles[kk] = true;
     }
     if (recalculate)
     {
@@ -181,9 +187,9 @@ public:
         int kk = myVars.where(k);
         if (kk < 0)
           continue;
-        if (rcsingles[k])
+        if (rcsingles[kk])
         {
-          dhpsioverpsi[kk] = -RealType(0.5) * ValueType(Sum(*lapLogPsi[k])) - ValueType(Dot(P.G, *gradLogPsi[k]));
+          dhpsioverpsi[kk] = -RealType(0.5) * ValueType(Sum(*lapLogPsi[kk])) - ValueType(Dot(P.G, *gradLogPsi[kk]));
         }
       }
     }
@@ -270,9 +276,9 @@ public:
         int kk = myVars.where(k);
         if (kk < 0)
           continue;
-        if (rcsingles[k])
+        if (rcsingles[kk])
         {
-          dlogpsi[kk] = dLogPsi[k];
+          dlogpsi[kk] = dLogPsi[kk];
         }
         //optVars.setDeriv(p,dLogPsi[ip],-0.5*Sum(*lapLogPsi[ip])-Dot(P.G,*gradLogPsi[ip]));
       }


prckent:
OK to test

ye-luo (changes requested):
Unfortunately, the added change breaks optimization tests like short-H4-opt-linear-16-1 and short-H4-opt-OneShiftOnly-1-16.

prckent:
@shivupa Since this is more complicated than we were expecting, should I find someone else to continue this? I know you have your own research to get on with, and on our side this might be the catalyst to setup a broader set of quick to run deterministic optimization tests, check if this error pattern is repeated in other jastrows etc.. Cheers, Paul

shivupa:
Thanks. I tried to take a look, but it is definitely more complicated than I have the bandwidth for right now! I'd appreciate that.

PR #2837, opened 2021-01-04
Reset the RNG state during cost function evaluation

markdewing:
Set the RNG state to be the same during checkConfigurations and correlatedSampling.

The "test" optimize method compares the analytic and finite difference parameter gradients.  If NLPP is involved, and the RNG is not reset before correlatedSampling, the gradients can differ in the comparison by 1-10%.   If everything is correct, and the RNG is reset properly, the gradients differ on the order of 1e-8 to 1e-6.

- Bugfix

- No

laptop, desktop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
@@ -325,6 +325,19 @@ void QMCCostFunctionBatched::checkConfigurations()
       for (int ib = 0; ib < curr_crowd_size; ib++)
       {
         samples.loadSample(p_list[ib].get().R, base_sample_index + ib);
+
+        // Set the RNG used in QMCHamiltonian.  This is used to offset the grid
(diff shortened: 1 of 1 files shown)


prckent (commented):
Can we think of an extreme case where this might be dangerous / give wrong results? e.g. 1 to a few electrons, atom, single node run. Note that the existing code by default uses much denser integration grids than strictly necessary so potential problems due to picking an unlucky grid will be camouflaged. Given that all the electrons will be in different positions, this seems OK as you note.

Right now this seems safe to me.

prckent (approved):
If the batch and crowd size is 1, does this preserve the non-batched old CPU code behavior or is the behavior still different? (This is an example of a place where the new code might reasonably deviate from the non-batched old code, just as the drivers are subtly different.)

markdewing:
Yes, with batch and crowd size of 1, the parameter gradient is the same between the new batch code and the old code (using the same RNG seed.)

PR #2840, opened 2021-01-06
Port spin density observable

PDoakORNL:
## Proposed changes

This is a port of the legacy SpinDensity estimator from the legacy architecture. It is the beginning of the strip down refactor of OperatorBase as well. SpinDensityNew derives from OperatorEstBase which manages a generic buffer of QMCTraits::RealType for the derived OperatorEstimators. This replaces the use of Collectables as the data store for these estimators.

This sort of estimator is now instantiated under the "control" of EstimatorManagerNew and is accumulated when other estimators are accumulated not when the Hamiltonian is evaluated.

Right now each SpinDensity new gets its own copy of the accumulator data structure, a pending PR will address the issue of large SpinDensities.

MPI reduction occurs between ranks 1 time per estimator per block for SpinDensity.

I will provide additional documention of the new estimator design soon. I'm beginning work on the DensityMatrix estimator now. But this is as much as I think should be reviewed at once.

@jtkrogel Right now qdens will not work with the hdf5 output because (as far as I can tell) it needs to be able to understand the input.xml which it does not.  I tried to modify nexus to allow this but there is something I'm not grasping about qmcpack_input.py and qdens.

port of SpinDensity from QMCHamiltonian based legacy driver architecture to unified batched driver architecture and management by EstimatorManagerNew and not QMCHamiltonian.

- No

Leconte

## Checklist

- Yes This PR adds some tests to cover some new code, or to catch a bug that is being fixed
- Yes Some documentation has been added (if appropriate)


src/Estimators/EstimatorManagerCrowd.cpp
```diff
@@ -31,10 +31,23 @@ EstimatorManagerCrowd::EstimatorManagerCrowd(EstimatorManagerNew& em)
   for (int i = 0; i < em.Estimators.size(); i++)
     scalar_estimators_.push_back(em.Estimators[i]->clone());
   MainEstimator = scalar_estimators_[EstimatorMap[MainEstimatorName]];
-  if (em.Collectables)
-    Collectables = em.Collectables->clone();
+  std::for_each(em.operator_ests_.begin(), em.operator_ests_.end(),
```
ye-luo, line 34: Really don't like `begin/end` and range-based loops are better alternatives without the need of any lambda.

src/Estimators/EstimatorManagerNew.cpp
```diff
@@ -198,8 +200,10 @@ void EstimatorManagerNew::start(int blocks, bool record)
     h_file = H5Fcreate(fname.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
     for (int i = 0; i < Estimators.size(); i++)
       Estimators[i]->registerObservables(h5desc, h_file);
-    if (Collectables)
-      Collectables->registerObservables(h5desc, h_file);
+    for (int iop = 0; iop < operator_ests_.size(); ++iop)
```
ye-luo, line 203: range-based loop over `operator_ests_`?

src/Estimators/EstimatorManagerNew.cpp
```diff
@@ -394,6 +464,15 @@ bool EstimatorManagerNew::put(QMCHamiltonian& H, xmlNodePtr cur)
...
         app_log() << "  Adding a default LocalEnergyEstimator for the MainEstimator " << std::endl;
       }
+      else if (est_name == "spindensity_new")
+      {
+        // Eventually this should be getting read before Estimator Manager New is constructed
+        // EMN will be constructed with a vector of EstimatorInputs
+        estimator_inputs_.emplace_back(EstimatorInput{std::in_place_type<SpinDensityInput>});
+        SpinDensityInput& spdi = std::get<SpinDensityInput>(estimator_inputs_.back());
+        spdi.readXML(cur);
+        operator_ests_.emplace_back(new SpinDensityNew(spdi, pset.mySpecies));
```
ye-luo, line 474: std::make_unique instead of new

src/Estimators/OperatorEstBase.h
```diff
@@ -0,0 +1,163 @@
...
+  std::string myName;
+
+  QMCT::FullPrecRealType walkers_weight_;
+
+  QMCT::FullPrecRealType get_walkers_weight() const { return walkers_weight_; }
+  ///constructor
+  OperatorEstBase();
+  OperatorEstBase(const OperatorEstBase& oth);
+  ///virtual destructor
+  virtual ~OperatorEstBase() {}
```
ye-luo, line 104: I think `virtual ~OperatorEstBase() = default;` is better than `virtual ~OperatorEstBase() {}`

src/Estimators/SpinDensityInput.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+  if (have_dr && have_grid)
+  {
+    APP_ABORT("SpinDensity::put  dr and grid are provided, this is ambiguous");
+  }
+  else if (!have_dr && !have_grid)
+    APP_ABORT("SpinDensity::put  must provide dr or grid");
+
+  if (have_corner && have_center)
+    APP_ABORT("SpinDensity::put  corner and center are provided, this is ambiguous");
```
ye-luo, line 86: exception?

src/Estimators/SpinDensityNew.h
```diff
@@ -0,0 +1,81 @@
...
+  //data members
+  SpinDensityInput input_;
+  SpeciesSet species_;
+
+  // this is a bit of a mess to get from SpeciesSet
+  std::vector<int> species_size_;
+
+  //constructor/destructor
+  SpinDensityNew(SpinDensityInput& sdi, const SpeciesSet& species);
+  ~SpinDensityNew() {}
```
ye-luo, line 52: Not needed line.

src/Estimators/SpinDensityNew.h
```diff
@@ -0,0 +1,81 @@
...
+  SpinDensityNew(const SpinDensityNew& sdn);
+  
+  //standard interface
+  OperatorEstBase* clone();
```
ye-luo, line 56: Please add override keyword to functions overriding base virtual functions.

src/Estimators/tests/test_SpinDensityInput.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    SpinDensityInput sdi;
+    sdi.readXML(node);
+    CHECK(sdi.get_npoints() == 1000);
+    TinyVector<int, SpinDensityInput::DIM> grid(10, 10, 10);
+    CHECK(sdi.get_grid() == grid);
+    TinyVector<int, SpinDensityInput::DIM> gdims(100, 10, 1);
+    CHECK(sdi.get_gdims() == gdims);
+  };
+  std::for_each(testing::valid_spin_density_input_sections.begin(),testing::valid_spin_density_input_sections.end(), xml_test);
```
ye-luo, line 41: range-based loop?

src/Estimators/tests/test_manager_mpi.cpp
```diff
@@ -45,12 +45,11 @@ bool EstimatorManagerNewTest::testMakeBlockAverages()
   //   - 2 with 1 sample 1
   //   - 1 with 2
   double block_weight = 0;
-  std::for_each(estimators_.begin(),estimators_.end(),[&block_weight](auto& est){
-                                                       block_weight += est.scalars[0].count();
-                                                     });
+  std::for_each(estimators_.begin(), estimators_.end(),
```
ye-luo, line 48: range-based loop.

src/Estimators/tests/test_SpinDensityNew.cpp
```diff
@@ -0,0 +1,105 @@
...
+{
+  Libxml2Document doc;
+  bool okay = doc.parseFromString(testing::valid_spin_density_input_sections[0]);
+  REQUIRE(okay);
+  xmlNodePtr node = doc.getRoot();
+  SpinDensityInput sdi;
+  sdi.readXML(node);
+  SpeciesSet species_set;
+  int ispecies                      = species_set.addSpecies("C");
+  int iattribute                    = species_set.addAttribute("membersize");
```
ye-luo, line 38: What is the version of clang-format you used? This issue should have been fixed.

src/Particle/ParticleSet.h
```diff
@@ -37,7 +37,7 @@ class DistanceTableData;
 class StructFact;
 
 /** Specialized paritlce class for atomistic simulations
- *
+ *oo
```
ye-luo, line 40: unintended change

ye-luo (changes requested):
I had a discussion with Peter over the phone yesterday. variant and C++17 will be avoided.
I left other issues in-source review here.

src/Estimators/tests/test_SpinDensityNew.cpp
PDoakORNL: I tried version 7,10,12 (develop)
They all do this :(

PDoakORNL: Although other instances in the same file turn out fine

jtkrogel:
Please send me an email with your input and output files.  I will make a separate PR to address the needs of qdens.

jtkrogel:
How are you testing the correctness of the new implementation?

src/Estimators/EstimatorManagerCrowd.cpp
```diff
@@ -31,10 +31,25 @@ EstimatorManagerCrowd::EstimatorManagerCrowd(EstimatorManagerNew& em)
...
   MainEstimator = scalar_estimators_[EstimatorMap[MainEstimatorName]];
-  if (em.Collectables)
-    Collectables = em.Collectables->clone();
+  for(UPtr<OperatorEstBase>& upeb : em.operator_ests_)
+  {
+    operator_ests_.emplace_back(upeb->clone());
+  }
 }
 
+void EstimatorManagerCrowd::accumulate(int global_walkers, RefVector<MCPWalker>& walkers, RefVector<ParticleSet>& psets)
```
PDoakORNL, line 40: Moved from header.  Implementation code should not be in the header unless it is a huge issue to put in in the .cpp.

PDoakORNL (commented):
I think this is ready to go.  Short of starting over and doing a minimal change version still attached to QMCHamiltonian and Collectables there is only the skeleton of the datalocality handling which I'd like to stay there since I'm going to immediately be using them for the density matrices and the huge spindenisty implementation.

PDoakORNL:
As far as testing that this produces the same result as the original out side of running the full app and comparing a statistical result it is not easy since SpinDensity has absolutely no unit testing and depends on the assumption of a tight coupling to a single walker state machine as is the norm for the estimators. 

What should be trivial, accumulating once and checking the data structure is quite painful. But I expect I will have to write a couple of unit tests for the original to be sure.  Outside of that a statistical test with the full app seems like all that can be done.

This PR is just an iterative step so I can stop rebasing against develop and review is not delayed by to much code to review.

ye-luo (dismissed):

prckent:
Are there any outstanding questions about this PR?

ye-luo (approved):
Just re-approve due to rebase.

ye-luo:
@jtkrogel Does this PR look good to you?

prckent:
(I understand this PR is a step along the way, so provided there are no major objections on the direction, I hope we can merge.)

PR #2841, opened 2021-01-08
Hack for correct VMC timer ordering

markdewing:
The VMC timers for these objects have the same scope as the object lifetime (that is - timer starts in the constructor, and stops in the destructor).
The call to make_unique creates the new object first, and then deletes the old object.  This causes an incorrect ordering in the timers. (The new timer starts before the old timer stops.)

This results in an error message - "Timer stack pop not matching push". The printed timer results also show VMC nested under VMC.

The fix is to force the old object to destroyed before the new object is created.

- Bugfix

- No

laptop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimize.cpp
+++ b/src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimize.cpp
@@ -566,6 +566,8 @@ bool QMCFixedSampleLinearOptimize::processOptXML(xmlNodePtr opt_xml,
   NumOfVMCWalkers = W.getActiveWalkers();
 
 
+  // Hack to correctly order timer with object lifetime scope
+  vmcEngine.reset(nullptr);
   // create VMC engine
   // if (vmcEngine == 0)
   // {
--- a/src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimize.cpp
+++ b/src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimize.cpp
@@ -566,7 +566,7 @@ bool QMCFixedSampleLinearOptimize::processOptXML(xmlNodePtr opt_xml,
   NumOfVMCWalkers = W.getActiveWalkers();
 
 
-  // Hack to correctly order timer with object lifetime scope
+  // Destroy old object to stop timer to correctly order timer with object lifetime scope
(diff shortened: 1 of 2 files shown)


ye-luo:
I won't call it hack. It seems more of a bug fix.
We should always remember overwriting a unique_ptr is not side effect free.
It complicates timer and may introduce temporal increase of memory consumption.
So overwriting must be used with care. So please update the documentation by replacing "Hack" with an explanation of the actual intention.

markdewing:
I tried to think of a better solution.  The best I have is a version of `make_unique` with the desired semantics - delete first, then allocate/construct/assign.

markdewing:
I found a another issue with the object lifetime scoped timers.   It assumes that VMC is the only thing running while the object is alive.    In the optimizer, that is not true.   The result is all the time in the cost function evaluation is a child of VMC, which is not desired.  One solution is to delete the vmcEngine pointer right after the samples are generated, and the VMC portion is finished.

ye-luo:
> I found a another issue with the object lifetime scoped timers. It assumes that VMC is the only thing running while the object is alive. In the optimizer, that is not true. The result is all the time in the cost function evaluation is a child of VMC, which is not desired. One solution is to delete the vmcEngine pointer right after the samples are generated, and the VMC portion is finished.

delete vmcEngine is actually preferred. I expect it making the code logic a lot cleaner and it avoids having objects with vague ownership between VMC and Optimization Drivers.

ye-luo (approved):

PDoakORNL:
I think the issue is with NewTImer the string mapping + side effect based design. I don't think unique_ptr gives any guarantee about the order of operations at all. At the end of the call you could expect the old object to be gone but during it you can't make assumptions like that.

PDoakORNL:
In fact for this code 
```c++
vmcEngine = std::make_unique<VMCcuda>(W, Psi, H, myComm, false);
```
100% I'd expect from C++ semantics.
since this call is really
std::unique_ptr::operator=(this, std::make_unique<VMCcuda>(...))

markdewing:
@PDoakORNL I don't think the problem is with the timer implementation.   The issue is the assumption/requirement that all the timers are strictly nested.  I don't think changing that requirement is worthwhile, partly because it creates difficulties in interpreting the results and summarizing and displaying the data.

PR #2843, opened 2021-01-13
fixing classic heap overflow

PDoakORNL:
## Proposed changes

Use the right index into the right vector. Otherwise it will probably work until it doesn't. test_estimators and test_estimators_mpi now pass llvm ASAN.

- Bugfix

- No

## Checklist

### DIFF
--- a/src/Estimators/EstimatorManagerNew.cpp
+++ b/src/Estimators/EstimatorManagerNew.cpp
@@ -377,27 +377,27 @@ void EstimatorManagerNew::reduceOperatorEstimators()
     operator_recv_buffer.reserve(nops);
     for (int iop = 0; iop < operator_ests_.size(); ++iop)
     {
-      auto& estimator = *operator_ests_[iop];
-      auto& data = estimator.get_data_ref();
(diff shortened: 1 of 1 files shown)


src/Estimators/EstimatorManagerNew.cpp
```diff
@@ -377,27 +377,27 @@ void EstimatorManagerNew::reduceOperatorEstimators()
...
     for (int iop = 0; iop < operator_ests_.size(); ++iop)
     {
-      auto& estimator = *operator_ests_[iop];
-      auto& data = estimator.get_data_ref();
+      auto& estimator      = *operator_ests_[iop];
+      auto& data           = estimator.get_data_ref();
       size_t adjusted_size = data.size() + 1;
       operator_send_buffer.resize(adjusted_size);
       operator_recv_buffer.resize(adjusted_size);
-      auto cur   = operator_send_buffer.begin();
```
PDoakORNL, line 385: These three lines are the fix.  Everything else is clang-format changing its mind about assignment alignment.

ye-luo (approved):

PR #2844, opened 2021-01-14
Spin density low crowd memory

PDoakORNL:
## Proposed changes

Addition of a new accumulation method for SpinDensityNew at the crowd context.
These estimators collect what amounts to the "writes" to the rank data structure during the block which are then written at block end under the control of a single thread. In the case of a fine spin density grid these write queues are much smaller than a copy of the spindensity grid for each crowd.

Currently this is turned on by add the `save_memory` attribute for example:
```xml
     <estimator name="SpinDensityNew" type="spindensity_new" report="yes" save_memory="yes">
       <parameter name="grid">
         200 200 200
       </parameter>
       <parameter name="center">
         0.0 0.0 0.0
       </parameter>
       <parameter name="cell">
                  3.37316115        3.37316115        0.00000000
                  0.00000000        3.37316115        3.37316115
                  3.37316115        0.00000000        3.37316115
       </parameter>
     </estimator>
```
## What

- New feature
   In order to make testing with random sequences clearer and not have to deal FakeRandom RandomGenerator_t RandomNumberControl the reliance on the obsolete `boost/random.hpp`
   I added RandomForTest which uses a wrapper based on <random> I've had around for awhile.
- Documentation changes

- No

Leconte

## Checklist

- Some documentation has been added (if appropriate)


src/Estimators/EstimatorManagerCrowd.cpp
```diff
@@ -31,32 +31,35 @@ EstimatorManagerCrowd::EstimatorManagerCrowd(EstimatorManagerNew& em)
...
+  for (int i = 0; i < operator_ests_.size(); ++i)
+    operator_ests_[i]->accumulate(walkers, psets);
+}
 
 
 void EstimatorManagerCrowd::startBlock(int steps)
 {
   crowd_estimator_timer_.restart();
+  for (auto& uope : operator_ests_)
+  {
```
ye-luo, line 57: More of my taste it is not worth adding {} around a single line.

ye-luo (approved):

ye-luo:
@prckent you self-scribed as a reviewer, it is your call to have additional review or merge.

prckent:
Minor questions: When is use of this feature expected? 200^3 is a fairly standard grid, maybe even the small size. What are examples of the speed difference? ( Our experiences indicate that we should be hesitant about adding options and take care to set defaults optimally. )

src/Estimators/EstimatorManagerCrowd.cpp
PDoakORNL: I agree with you on that sometimes I just drop the braces and forget to clean up.

PDoakORNL:
1000x1000x1000 I'd think, some benchmarking needs to be done.

prckent (approved):

jtkrogel:
Just to note here: standard density grids are typically comprised of grid cells with an extent of about 0.25 Bohr on a side.  This might help to ballpark the type of applied systems where a particular grid size would be appropriate.

PDoakORNL:
I'll look at the test systems and make another file like qmc_spindens_vmcbatch_short.in.xml but with a system large enough that a larger grid is appropriate, this will get the particles vs. gridpoints relation more realistic.

ye-luo:
Unified driver supposed to work with CPU with many crowds as well. On KNL, with 400x400x400 grid with 128 walkers(crowds), single precision takes 15GB which is a huge NO. So storing one grid per crowd is not a good solution and this PR addressed this I think.

PR #2845, opened 2021-01-14
Deterministic test for parameter gradients

markdewing:
## Proposed changes

Add a deterministic test for parameter gradients for one test case.
Not every value is checked, but a few values from each component of the wavefunction.

The errors bounds were picked arbitrarily.  They were not derived from examining the variance in values from different compilers and machines.  It should be good enough to catch gross errors.

Add an option to QMC_RUN_AND_CHECK_CUSTOM_SCALAR to adjust the equilibration value to check_scalars (-e).   Otherwise the default value (2) will skip the first two values (and so require at least 3 blocks of output).

- No

laptop (gcc,clang)   Cooley(llvm)

## Checklist


prckent (approved):

prckent:
I'll merge this so we can start using it, but I wonder if an update will be needed for complex and/or mixed precision.

markdewing:
I tested it with the complex and mixed precision builds, and it worked.  The test error bounds are pretty large.

PR #2848, opened 2021-01-15
Add evaluateGL apis to TWF

ye-luo:
## Proposed changes
In order to clean up buffers, I need this new set of APIs to do the compute work used to be in updateBuffer.
Slow fallback is in-place to ensure correctness.

- New feature

- No

Bora

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/WaveFunctionComponent.cpp
+++ b/src/QMCWaveFunctions/WaveFunctionComponent.cpp
@@ -38,16 +38,35 @@ WaveFunctionComponent::WaveFunctionComponent(const std::string& class_name, cons
     throw std::runtime_error("WaveFunctionComponent ClassName cannot be empty!");
 }
 
-// WaveFunctionComponent::WaveFunctionComponent(const WaveFunctionComponent& old):
-//   Optimizable(old.Optimizable), UseBuffer(old.UseBuffer),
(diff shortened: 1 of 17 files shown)


src/QMCWaveFunctions/WaveFunctionComponent.cpp
```diff
@@ -38,16 +38,35 @@ WaveFunctionComponent::WaveFunctionComponent(const std::string& class_name, cons
...
-//   //if(dLogPsi.size()) dLogPsi.resize(dLogPsi.size());
-//   //if(d2LogPsi.size()) dLogPsi.resize(d2LogPsi.size());
-//   //if(dPsi) dPsi=old.dPsi->makeClone();
-// }
+void WaveFunctionComponent::mw_evaluateLog(const RefVector<WaveFunctionComponent>& WFC_list,
+                                           const RefVector<ParticleSet>& P_list,
+                                           const RefVector<ParticleSet::ParticleGradient_t>& G_list,
+                                           const RefVector<ParticleSet::ParticleLaplacian_t>& L_list)
+{
+#pragma omp parallel for
```
PDoakORNL, line 46: We need an abstraction for these nested levels of concurrency and they need to be carefully reasoned about.

This case its clean and organized enough with respect to the crowd threads that I think it's a good target case so I think they should stay.

PDoakORNL (approved):
This looks good, I'm looking forward to see how you deal with moving twf off of updateBuffer.

PR #2849, opened 2021-01-18
Drop WF buffers used in batched drivers

ye-luo:
## Proposed changes
Using a buffer to fuse data transfer is efficient but error prone. This PR removes almost all buffer operations in batched drivers leveraging #2848.
1. This PR also reduced interaction with the list of lightweight walkers owned by MCPopulation. The eventual goal is to get them removed. the list of lightweight walkers is only intended to used outside drivers or carry data between drivers not inside drivers.
2. Restore the recompute capability for determinants.
3. DMC without MPI is fixed as tests reporting pass. Consider the test check was not made tight. It can be buggy. With MPI still getting bad memory access.

Since the buffer motion is cleaned up, the walker exchange part is my next target.

- No

Bora

## Checklist

### DIFF
--- a/src/Particle/MCSample.h
+++ b/src/Particle/MCSample.h
@@ -36,6 +36,9 @@ struct MCSample
   ParticleSet::ParticleLaplacian_t L;
   ParticleSet::RealType LogPsi, Sign, PE, KE;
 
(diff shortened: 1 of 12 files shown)


markdewing (approved):

PDoakORNL (approved):
Looks saner already.  And yes the ranged for idiom is more readable.

PR #2852, opened 2021-01-20
Enhance TimerManager and fix a missing timer stop() call in VMCbatched driver

ye-luo:
## Proposed changes

1. Some error traps in NewTimer class are moved to TimerManager. Now if an error was detected, the code will stop instead of just print an error. Usually an error in the timer indicates pathological call flow, it is better to do a hard stop. Because exception can be thrown via TimerManager::pop_timer() called by ScopedTimer destructor, it is better to print the full error message right away instead of via runtime_error object which may get lost due to destructors being implicitly noexcept by default.
2. A bug was fixed in timer unit test.
3. The fix to unpaired timer call is trivial.

- Bugfix

- No

Epyc-server

## Checklist

### DIFF
--- a/src/Utilities/TimerManager.cpp
+++ b/src/Utilities/TimerManager.cpp
@@ -62,6 +62,18 @@ void TimerManager<TIMER>::initializeTimer(TIMER& t)
 }
 
 template<class TIMER>
+void TimerManager<TIMER>::throwErrorMessageOnceAndPrintOnlyAfter(const std::string& msg)
+{
+  if (stop_throw_already_in_bad_state)
+    std::cerr << msg << std::flush;
+  else
+  {
+    stop_throw_already_in_bad_state = true;
+    throw std::runtime_error(msg);
+  }
+}
+
+template<class TIMER>
 TIMER* TimerManager<TIMER>::createTimer(const std::string& myname, timer_levels mytimer)
 {
   TIMER* t = nullptr;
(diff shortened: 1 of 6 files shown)


src/Utilities/TimerManager.cpp
```diff
@@ -74,6 +75,40 @@ TIMER* TimerManager<TIMER>::createTimer(const std::string& myname, timer_levels
...
+template<class TIMER>
+void TimerManager<TIMER>::push_timer(TIMER* t)
+{
+  // current_timer() can be nullptr when the stack was empty.
+  if (t == current_timer())
+  {
+    std::cerr << "Timer " << t->get_name()
+              << " instance is already at the top of the stack. "
+                 "start() is being called again. This often happens when stop() is not paired properly with start(). "
+              << "ScopedTimer uses RAII and managers timer start/stop much safer." << std::endl;
```
markdewing, line 87: Possible change: "ScopedTimer uses RAII and manages timer start/stop more safely."

src/Utilities/TimerManager.cpp
```diff
@@ -74,6 +75,40 @@ TIMER* TimerManager<TIMER>::createTimer(const std::string& myname, timer_levels
...
+  TIMER* stack_top = current_timer();
+  if (stack_top == nullptr)
+  {
+    std::cerr << "Timer stack pop failed on an empty stack! Requested \"" << t->get_name() << "\"." << std::endl;
+    throw std::runtime_error("TimerManager pop_timer error!");
+  }
+  else if (t != stack_top)
+  {
+    std::cerr << "Timer stack pop not matching push! "
+              << "Expect \"" << t->get_name() << "\" but \"" << stack_top->get_name() << "\" on the top." << std::endl;
```
markdewing, line 106: Possible change: "Timer stack pop does not match push"
            "Expecting ... but ... is on the top"

ye-luo:
@markdewing thanks. I updated the printouts.

markdewing (approved):

prckent (approved):
LGTM

PR #2853, opened 2021-01-21
Change base class of linear optimizer to QMCDriverNew

markdewing:
## Proposed changes

The counterpart to #2700 for the linear optimizer.   Mostly changing names and moving a few things around to match QMCDriverNew.

Two changes that are not among the above
- `QMCLinearOptimizeBatched::put` was removed (rather than rename to process).  It is not called, and confusingly duplicates code in `QMCFixedSampleLinearOptimizeBatched`.
- Rng handling in the QMCDriverNew destructor.   The previous code leads to a segfault (null Rng pointer) when VMC follows an optimizer run.   See the comment for more details.

The test deterministic-He_ae-opt_vmc-batch-1-1-totenergy fails.  I think the value just needs to be updated, but I need to look into it more.

- New feature

- No

laptop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.h
+++ b/src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.h
@@ -51,16 +51,16 @@ public:
                                       Communicate* comm);
 
   ///Destructor
-  ~QMCFixedSampleLinearOptimizeBatched();
+  virtual ~QMCFixedSampleLinearOptimizeBatched();
 
   ///Run the Optimization algorithm.
-  bool run();
+  bool run() override;
   ///preprocess xml node
-  bool put(xmlNodePtr cur);
+  void process(xmlNodePtr cur) override;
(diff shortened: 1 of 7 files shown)


src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.h
```diff
@@ -51,16 +51,16 @@ class QMCFixedSampleLinearOptimizeBatched : public QMCLinearOptimizeBatched,
                                       Communicate* comm);
 
   ///Destructor
-  ~QMCFixedSampleLinearOptimizeBatched();
+  virtual ~QMCFixedSampleLinearOptimizeBatched();
```
ye-luo, line 54: Is virtual necessary?

markdewing: Probably not.  It was a change I made to try fix the Rng crash issue.

ye-luo (approved):

prckent (approved):

PR #2854, opened 2021-01-23
Output linear matrices

markdewing:
## Proposed changes

Output linear method matrices (overlap and Hamiltonian) in scalar.dat format.
They can be processed with qmca and check_scalars.py, and compared between the batched and non-batched versions of the code.
A separate option (`freeze_parameters`) disables updating the parameters so the output can be used for statistics.

Enable by adding the parameters `output_matrices` and `freeze_parameters` to the qmc block
```
   <qmc method="linear_batch" move="pbyp" checkpoint="-1">
      <parameter name="output_matrices">yes</parameter>
      <parameter name="freeze_parameters">yes</parameter>

```

- New feature

- No

laptop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/OutputMatrix.cpp
+++ b/src/QMCDrivers/WFOpt/OutputMatrix.cpp
@@ -27,7 +27,7 @@ void OutputMatrix::init_file(const std::string& root_name, const std::string& na
   // Assume that the root_name has the series suffix (".s000").
   // Remove the series suffix to get the project id
   std::string fname = root_name.substr(0, namelen - 5) + "." + name + ".s000.scalar.dat";
-  app_log() << "Output file: " << fname << std::endl;
+  app_log() << "Output matrix file: " << fname << std::endl;
 
   output_file_.open(fname);
   output_file_ << "# Index ";
--- a/src/QMCDrivers/WFOpt/OutputMatrix.cpp
+++ b/src/QMCDrivers/WFOpt/OutputMatrix.cpp
@@ -1,0 +1,60 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2021 QMCPACK developers.
+//
+// File developed by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
+//
+// File created by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+/** @file OutputMatrix.cpp
+ * @brief Output the Hamiltonian and overlap matrices from linear method
+ */
+
+#include "QMCDrivers/WFOpt/OutputMatrix.h"
+
+
+namespace qmcplusplus
+{
+OutputMatrix::OutputMatrix() : index_(0) {}
+
+void OutputMatrix::init_file(const std::string& root_name, const std::string& name, int N)
+{
+  int namelen = root_name.length();
+  // Assume that the root_name has the series suffix (".s000").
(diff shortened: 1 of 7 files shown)


markdewing:
Attached is a script to compare the output of qmca for two runs.

[compare_qmca.zip](https://github.com/QMCPACK/qmcpack/files/5860774/compare_qmca.zip)

prckent:
Very useful to have.

Just one concern: I think that the freezing/non-updating of parameters also should be well advertised. Perhaps output_matrices should be output_matrices_freeze_parameters or output_matrices_no_update? debug_matrices? A warning around line 1294 of QMCFixedSampleLinearOptimize would also help.

markdewing:
If the effect of freezing the parameters is a concern, it could be moved to a separate attribute - "freeze_parameters".

jtkrogel:
Why an xml attribute vs a `<parameter>`?  The qmc elements have all reached such a size that line-listed parameters are the common, and IMO appropriate, case.

markdewing:
Using an attribute was lazily extending the existing code. A `<parameter>` is probably better.

prckent:
@markdewing Separate parameters is a good.

prckent (approved):

src/QMCDrivers/WFOpt/OutputMatrix.cpp
```diff
@@ -0,0 +1,60 @@
...
+//
+// File created by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+/** @file OutputMatrix.cpp
+ * @brief Output the Hamiltonian and overlap matrices from linear method
+ */
+
+#include "QMCDrivers/WFOpt/OutputMatrix.h"
```
ye-luo, line 17: OutputMatrix.cpp needs to include "OutputMatrix.h". Related header files should be included without any path.

ye-luo (commented):
> A separate option (freeze_parameters) disables updating the parameters so the output can be used for statistics.

What statistics? I'm confused.

markdewing:
The output of the matrix files is in scalar.dat format that can be post-processed with qmca or check_scalars.py.    Each iteration of the "loop" construct in the input adds another line to the file - the freeze_parameters option keeps the parameters from changing at each iteration.

ye-luo:
If it is frozen, why running multiple loop iterations is needed? What is the intention?

markdewing:
The intention is to compare between the old,non-batched code and the batched code, and for regression tests.  It started with parameter gradients.  The existing 'test' input (in Optimize/testDerivOptimization.cpp) would print out the analytic gradient and numerical gradient.   I extended it to also output the analytic gradients to a scalar.dat file.   This can be used for regression testing, for instance, in tests/molecules/He_param.   In order for check_scalars.py to generate statistics, there needs to be multiple lines in the *.scalar.dat input file (of parameter gradients).    The gradient evaluation is carried out once per optimizer loop iteration.  In order to get multiple lines in the parameter gradient file, I hacked it to print one line per optimizer loop.

The input file to the He_param tests has the lines
```
  <loop max="10">
    <qmc method="opt_batch" move="pbyp" checkpoint="-1" gpu="no">
      <optimize method="test" output_param_file="yes"/>
```
This produces 10 lines of parameter gradients in the output file (He.param.s000.scalar.dat).
It's a hack.  Maybe there's a better way of doing this?

I extended the same pattern to the H and S matrices for the linear method.

ye-luo:
Now I get the point, even if parameters are fronzen, samples are changing and so the statistics are changing.
1. loop should update s00x. So you can get one file per iteration.
2. Checking the average of 10 loop iteration is more for convenience, it doesn't correspond to any real needs.
3. run_qmc can check scalar.dat at every QMC section. So loop=2 and check two scalar.dat even with just one line each should be sufficient to cover the real use case, isn't it?

PR #2862, opened 2021-01-29
Update Clang and GNU offload flags.

ye-luo:
## Proposed changes

GNU setting is new. Clang setting only has -Wno-unknown-cuda-version for less warning.

- Build related changes

- No

Epyc-server

## Checklist


prckent (approved):
We'll eventually need some dependencies for nvidia/amd/intel offloading

ye-luo:
So far 

> We'll eventually need some dependencies for nvidia/amd/intel offloading

We may have to do that eventually but so far the clang setting works for NVIDIA, AMD and Intel.

PR #2869, opened 2021-02-02
Use correct constructor for VMCDriverInput.

### LINKED ISSUE #2867: Wrong call to VMCDriverInput constructor
**Describe the bug**
VMCDriverInput(bool use_drift);
https://github.com/QMCPACK/qmcpack/blob/aa867a262d83da1594a6375d37ac7316b5250752/src/QMCDrivers/WFOpt/QMCWFOptFactoryNew.cpp#L22
https://github.com/QMCPACK/qmcpack/blob/aa867a262d83da1594a6375d37ac7316b5250752/src/QMCDrivers/WFOpt/QMCWFOptFactoryNew.cpp#L41
@markdewing I think this constructor is not absolutely needed. So I think it is better to just delete it to prevent miss use.

VMCDriverInput(0) is called and use_drift_ is set false.
VMCDriverInput::readXML reads use_drift = "yes" and
`if (use_drift == "no") use_drift_ = false;`  is not applied and use_drift_ remains false.
And the wrong VMC sampling is used.

Please update deterministic-diamondC_1x1x1_pp-param-grad-1-1 reference values as well.

markdewing:
Fixes #2867, which can cause the wrong setting for use_drift.

Also update deterministic-diamondC_1x1x1_pp-param-grad-1-1 reference values.

- Bugfix

- No

laptop

## Checklist


ye-luo (approved):

prckent (approved):
LGTM

OK to merge provided all the CI passes. Looks like slightly larger tolerances are needed.

prckent:
OK to test

ye-luo (approved):

PR #2871, opened 2021-02-03
Fix legacy CUDA DMC breakage

ye-luo:
## Proposed changes
Fix breakage #2855 and protect inaccessible code pathes which their tests skipped.

- Bugfix

- No

epyc-server

## Checklist


prckent (approved):
Thanks! Some of my tests are still running, but at least some of the previously crashing DMC short tests are now passing while the deterministic tests are still good.

I assume we didn't catch this in the deterministic tests because the population is small and not sufficiently dynamic.

Issue #2851, opened 2021-01-19
AFQMC converter failures


prckent:
**Describe the bug**

Many of the AFQMC converter tests are failing in the nightlies, e.g. https://cdash.qmcpack.org/CDash/viewTest.php?onlyfailed&buildid=177579

There could be configuration issues here (which we should trap) or underlying bugs. Pyscf may be at fault for a few. Possibly the converter scripts need updating for h5py 3?

**To Reproduce**

Develop version, gcc 10 mkl complex and real. python 3.8.7, h5py 3.1.0...., all installed via spack. I can push the exact nightly script and spack hash etc. if needed, but they are essentially the canonical ones.

**Expected behavior**
Tests pass

**System:**

Our nightly test systems sulfur, nitrogen

**Additional context**

prckent:
@Paul-St-Young Created as requested. Very interested if you can reproduce any of these. Suggestive of version dependencies/missed updates are errors such as https://cdash.qmcpack.org/CDash/testDetails.php?test=11508899&build=177579

Paul-St-Young:
@prckent Thanks! I will try to fix these. Gotta learn spack first

prckent:
Knowing whether they run on a random e.g. Ubuntu would also help. There shouldn't be anything spack specific here.

prckent:
Note that h5py v3 is needed to support the latest HDF5 v1.12.0

Paul-St-Young:
Yes, I was able reproduce many breakages on a fresh install of Fedora 33 (gcc 10.2.1, python 3.9.1, h5py 3.1.0 )

07 and 08 breakages are trivial and due to change of default mode of `h5py.File` from "r+" to "r". Solution is to specify "r+" mode `h5py.File('scf.chk', 'r+')`.

06 breakage is also trivial and due to removal of "fmt" keyword from `plt.savefig` in matplotlib 3.3. Solution is to remove this keyword.

I cannot reproduce problem with 01 using pyscf 1.7.5 from pip install. This appears to be a PySCF problem anyways.

05 is only real problem. I'll look into this tomorrow

Paul-St-Young:
`SparseTensorIO.hpp::writeSparseTensor` under the flag `write_hdf` in `FactorizedSparseHamiltonian.cpp` is the source of the breakage for example 05.

I got "ERROR HDF5 write failure in hdf_archive::write dims".
My guess is an analogous default mode changing from "r+" to "r" when opening an HDF5 file.

[...]

PR #2872, opened 2021-02-03
Update AFQMC converter examples

Paul-St-Young:

## Proposed changes

Address #2851 

The only change worth a double-check is the added write permission to wavefunction HDF5 file
```
Wavefunction WavefunctionFactory::fromHDF5
-  if (!dump.open(filename, H5F_ACC_RDONLY))
+  if (!dump.open(filename, H5F_ACC_RDWR))
```

- Bugfix

- No

i7-8750H, Fedora 33 (gcc 10.2.1, python 3.9.1, h5py 3.1.0 )

## Checklist


ye-luo:
Could you say a bit more about H5F_ACC_RDONLY -> H5F_ACC_RDWR? if the change is not added, what is the behavior of the code? I'm wondering if some write failure error was not captured properly.

Paul-St-Young:
Before the change, `afqmc.h5` is not edited during the run, so yes I agree there are write failures not being caught.
After H5F_ACC_RDONLY -> H5F_ACC_RDWR?, afqmc.h5/Wavefunction/PHMSD gets a new entry HamiltonianOperations/SparseTensor.

ye-luo:
@mmorale3 what is your assessment on the write error capture issue?

mmorale3:
An oversight on my end. Writing restart for hamiltonian objects was needed
at first, but by now is nearly irrelevant with the more
efficient choices. Only the sparse hamiltonian has a slow setup time and
could benefit for this. I might just disable this entirely.

prckent:
We know from painful experience that an H5 opened for writing is at risk for fatal corruption if there is crash or similar issue ahead of the file being closed. Therefore e.g. mixing H5 inputs and outputs is not wise since you can lose input data as well as output data.

Should we live with this current RW situation and merge?

Paul-St-Young:
Alternatively, we could disable Hamiltonian checkpoint simply by forcing `write_hdf=0` and revert to RDONLY for safety. Since @mmorale3 wants to disable this feature eventually, is it then acceptable to drop it from the tests?

prckent:
Sounds good to me. User impact is tiny / zero  as far as I know.

ye-luo:
Need a reproducer to look at the slipped error. Could you upload one?

Paul-St-Young:
@ye-luo `examples/afqmc/05-N2_phmsd` is the example that hits the write failure.

ye-luo:
Okay to test

ye-luo:
@mmorale3 looks good to you?

mmorale3:
Paul is working on a cleaner fix.

ye-luo (changes requested):
Blocked as we are waiting for a cleaner fix.

Paul-St-Young:
Instead of disabling `write_hdf` in each Hamiltonian object, I now disable it from `WavefunctionFactory.cpp`.
By passing an unopened restart file to `Hamiltonian::getHamiltonianOperations`, `write_hdf` is set to `false` by internal logic.

Paul-St-Young:
After more discussions with @mmorale3, we think the entire wave function restart design needs a revamp. Right now, the user is supposed to set "restart_file" in the input, run once to generate HamiltonianOperators, then remove "restart_file" and change wavefunction file to the previously set "restart_file" in order to skip generation of the HamiltonianOperators. We want to move to a design where the user need not change the input in between runs.

Given the considerable amount of work that would entail, I move to close this PR; make a new one with just the Python changes; then create a dedicated PR for the restart issue.

Does this sound good to everyone?

ye-luo:
You should always be able to repeat the "initial" run as you need.
wf_file="a.h5" precomptue_file="b.h5"
A run should check the existence of b.h5, if is not there, redo the precompute. If it is there, load it.
the a.h5 should not be touched.

Paul-St-Young:
yes, that's the plan

prckent (commented):
Closing this PR and then making a new one that just tackles the converter issues at e.g. https://cdash.qmcpack.org/CDash/viewTest.php?onlyfailed&buildid=185059 would be good.

PR #2873, opened 2021-02-03
[WIP] DPCPP port of DelayedUpdate

kgasperich:
## Proposed changes
This PR provides a DPCPP version of DelayedUpdate (DelayedUpdateSYCL).

The initial implementation was based on DelayedUpdateCUDA, and a few issues still exist due to reused code that worked for the CUDA implementation but are incorrect for the SYCL one (most notably, the error handling for lapack calls).

new classes/structs:
- `SYCLAllocatorBase`: base struct for all SYCL allocators; ensures a single shared queue (should be changed to a single context allowing multiple queues)
- `SYCLManagedAllocator`: allocator for SYCL unified memory (this is the only SYCL allocator implemented that has been implemented so far)
- `DelayedUpdateSYCL`: direct translation (CUDA->DPCPP) of `DelayedUpdateCUDA` .  The `getrs` call in the double-precision version of `invert_transpose` is incorrect (the mixed-precision one is ok).  Error handling is not done correctly for any of the `getrs`,`getrf` calls (`info` is not an output parameter anymore, need to implement clean exception handling).

new functions:
- everything in `QMCWaveFunctions/detail/SYCL/delayed_update_helper.h` is a direct CUDA->DPCPP translation from `CUDA/delayed_update_helper.cu`

## TODO

- [ ] make sure `CMakeLists.txt` is correct and put something like `#ifdef ENABLE_SYCL` (maybe also use `ENABLE_OFFLOAD`?) where appropriate
- [ ] fix comments that were copied from CUDA code
- [ ] fix exception handling for oneMKL lapack calls
- [ ] cleaner queue/context handling for allocators

- [x] Yes
- [ ] No

- code will not build without oneapi until `#ifdef`s are put back in
- only the mixed-precision version will build (and some tests still fail)

## Checklist

- [x] This PR is up to date with current the current state of 'develop'
- [x] Code added or changed in the PR has been clang-formatted


ye-luo:
More things needs to be sorted out first. Please do a PR on your fork repo not on the qmcpack main repo.

prckent:
Note that I am not terribly keen to have any SYCL in mainline QMCPACK ahead of broader discussions including testing, rationale, ability to support long term etc. etc. A better destination would be miniqmc.

ye-luo:
We will need at least the allocator part for pinned memory for optimal OpenMP offload performance.

PR #2875, opened 2021-02-04
Fix recompute_between_blocks default value in batched drivers

ye-luo:
## Proposed changes
Review after #2870.
1. Fix recompute_between_blocks default value in batched drivers. It was incorrect in mixed precision build.
2. Restore timers in WalkerControl.

- Bugfix

- No

epyc-server

## Checklist


PDoakORNL (approved):
LGTM

PR #2876, opened 2021-02-05
Initialize values to zero on device only memory

ye-luo:
## Proposed changes
Part of the fix for NaN issue when offload+CUDA is used. If we don't assign 0, the value can be NaN from CUDA device memory allocation.
I was expecting 0 * anything = 0 but this is not the case for NaN. So now always fill GPU array with 0.
There is another NaN I'm looking at and I will PR separately once it is fixed.

CUDAfill_n is made explicit template. Filling zero by byte is good for int, size_t, float, double, complex<float>, complex<double>.
Other unknown types may need customed fill_n.

@PDoakORNL I expanded the evalGrad check, 0 value usually doesn't occur in simulation but it is expected for unit test with FakeSPO.
 
- Bugfix

- Yes

Epyc-sever

## Checklist


PDoakORNL (approved):
LGTM hopefully not many more sources of NaN to go.

ye-luo:
The residual NaN seems coming out of T-moves. I'm looking at it closely now.

PR #2878, opened 2021-02-06
Clean up and optimize table method

ye-luo:
Review after #2755
## Proposed changes
Worked with @anbenali on cleaning and optimizing the MSD table method code path.

1. reduce hard-coded spin up and down with an array. Only updated routines affected and others can adapt to the same pattern. Open doors to up/down/positron case.
2. More unique_ptr to reduce memory leak.
3. Introduce WFC::prepareGroup() for handling computational.
4. The updated algorithm benefits all computation. Significant visible gain at 1M dets.

XML input tag Fast="yes" remains the pre-existing table method code-path
Fast="new" enables optimized code-path. This is the new default.
The divergence of these two are small so kept in one class.

- New feature

- No

epyc-server

## Checklist


prckent:
Please add a test for the old algorithm if all our tests will now go through the new path, or vise versa.

ye-luo:
@prckent unit test explicitly test both cases.
C2-H5-msd_no_precompute has been added to cover the old code-path.

prckent:
(@ye-luo If we can we always want both unit tests and a final integration test. The latter is what the user sees, the former help developers)

prckent (changes requested):
1. test_C4_MSD_excited/orbitals - remove empty file

1. As regards input tags, in general across the application we should use algorithm=meaningfulname rather than new/opt/old etc.

1. If we use "new" then one day we might have "newer" as well as "new".

1. Please change the input tag to 
algorithm
and use "legacy" for the old algorithm (or something similar.

1. Please pick a name other than "new" for the new algorithm, e.g. "precompute". One day it won't be new.

1. Please ensure there is a test using the same gold data with both the legacy and precompute algorithm. Can be stochastic initially. These might be present but there are so many files touched here I missed them in the GitHub interface.

1. Is it possible have these files in the modern input style with the SPO set created outside the determinant set? The reference outputs contain many warnings.

1. Since we already had a MSD test, what is unique about the new tests added here? Should we keep the old one?

ye-luo:
>     1. test_C4_MSD_excited/orbitals - remove empty file
This is from #2755 not part of this PR and related to convert4qmc. I can answer to my best knowledge, the empty file is to trick converter reading orbital from h5. I don't think it is a sane solution and a better solution should be used.

>     2. As regards input tags, in general across the application we should use algorithm=meaningfulname rather than new/opt/old etc.
> 
>     3. If we use "new" then one day we might have "newer" as well as "new".
> 
>     4. Please change the input tag to
>        algorithm
>        and use "legacy" for the old algorithm (or something similar.
> 
>     5. Please pick a name other than "new" for the new algorithm, e.g. "precompute". One day it won't be new.

Expect this discussion.
I don't even like the existing input xml tag "Fast" which refers to table method.
I would propose new input table_method
mapping relations.

| Syntax      | Description |
| ----------- | ----------- |
| Header      | Title       |
| Paragraph   | Text        |

| input tag |  no table method  | table method no precomputing | table method with precoputing |
| --- | --- | --- | --- |
| Fast | no | yes | - |
| table_method | no | no_precomputing | precomputing/yes |

yes will be floating input pointing to the best option.

> 
>     6. Please ensure there is a test using the same gold data with both the legacy and precompute algorithm. Can be stochastic initially. These might be present but there are so many files touched here I missed them in the GitHub interface.
> 

Yes. identical reference values are used.

>     7. Is it possible have these files in the modern input style with the SPO set created outside the determinant set? The reference outputs contain many warnings.
> 

I think @anbenali started updating the converter. I think it will be ready soon for updating those input files.

>     8. Since we already had a MSD test, what is unique about the new tests added here? Should we keep the old one?
The added excited state tests are from #2755.

prckent:
Re input tag. I prefer having "algorithm" since we can use that for other functionality and it will therefore reduce the multiplication of input tags and provide some consistency.

algorithm=all_determinants   - slow method, no tables
algorithm=table_method - our original table method implementation
algorithm=precomputed_table_method - new faster table method implementation

If not specified the default should be precomputed_table_method, i.e. the new code.

@anbenali ? Is this clear? Preferences?

( Another tag we could use widely would be "dispatch", with dispatch=cpu, cpu_batched, offload_batched, fpga_batched etc. )

anbenali:
@prckent I think I understand. will change converter accordingly and propagate the tags in the builder.

ye-luo:
Updated input tags.
Need to basically create the manual subsection #2907

ye-luo:
@prckent could you do a second round of review?

src/Particle/ParticleSet.h
```diff
@@ -562,16 +562,8 @@ class ParticleSet : public QMCTraits, public OhmmsElementBase, public PtclOnLatt
...
-  {
-    if (iat < 0 || iat >= TotalNum)
-      throw std::runtime_error("ParticleSet::getGroupID iat is out of range.");
-    int id = 0;
-    while (iat >= SubPtcl[id + 1])
-      id++;
-    return id;
-  }
+  ///return the group id of a given particle in the particle set.
+  inline int getGroupID(int iat) const { return GroupID[iat]; }
```
prckent, line 566: Removed the error check to get inlining and good optimization?

Perhaps add an assert that would be active during DEBUG?

src/io/OhmmsData/OhmmsParameter.h
```diff
@@ -46,10 +46,11 @@
...
 enum class TagStatus
 {
-  OPTIONAL, // there is a default value
-  REQUIRED, // no default value, input required.
+  OPTIONAL,   // there is a default value
+  REQUIRED,   // no default value, input required.
   DEPRECATED, // deprecated input tag. Input becomes optional
-  DELETED // deleted input tag. Hard stop if detected.
+  DELETED,    // deleted input tag. Hard stop if detected.
+  NOSUPPORT   // not supported input tags due to incompatible changes. Hard stop if detected.
```
prckent, line 53: Prefer UNSUPPORTED here, it is a little clearer imo.

prckent (changes requested):
Minor comments to address as entered.

src/Particle/ParticleSet.h
ye-luo: Added assert which is effective in a debug build.

src/io/OhmmsData/OhmmsParameter.h
ye-luo: Changed.

prckent (approved):

prckent:
Ran most of the non-long tests, llvmdev cpu. Looks to be OK

Issue #2874, opened 2021-02-03
Crash in VMC batched code (cusolver init) with not very large walker counts


prckent:
**Describe the bug**

Can not run many walkers before hitting a cusolver init error in the batched code.

e.g.  1024 walkers, 16 omp threads is bad:

8 electron water molecule / splines. llvmdev+cuda 11.2.0+openblas.

```
=========================================================
  Start VMCBatched
  File Root hf_vmc_1024w_batched_offload.s001
=========================================================
VMCBatched Driver running with target_walkers = 1024
                               walkers_per_rank = [1024]
                               num_crowds = 16
                    on rank 0, walkers_per_crowd = [64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64]

cusolverAssert: CUSOLVER_STATUS_INTERNAL_ERROR, file git_QMCPACK_prckent/qmcpack/src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h , line 111
cusolverCreate failed!
```

With 1 omp thread and therefore crowd, 16-257 walkers are OK. 511, 512, and 1024 walkers are bad. => there is a subtle limit somewhere. If us, we should fix, if CUDA, we should chunk over some reasonable walker count to bypass.

Added: Note that the  legacy CUDA code is fine with these walker counts.

**To Reproduce**

1. development version 
2. offload build
3. cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DQMC_MPI=0 -DCMAKE_CXX_FLAGS=-Wno-unknown-cuda-version -DENABLE_OFFLOAD=ON -DUSE_OBJECT_TARGET=ON -DENABLE_CUDA=ON -DCUDA_ARCH=sm_70 -DCUDA_HOST_COMPILER=`which gcc` -DQMC_DATA=/scratch/QMC_DATA_FULL ../qmcpack/

**Expected behavior**
Large walker counts should run successfully. For small electron count systems, large walker counts are needed for performance, as per the legacy CUDA code.

**System:**

 nitrogen (AMD+NVIDIA V100), nightly testing setup.

**Additional context**

ye-luo:
This is a memory issue I mentioned to you via slack. The reason is we cannot afford creating one cublas handle per walker. Each cublas handle needs 8MB and 1024 walker takes a lot. If you are not using batched slater determinant, you also need 33 MB per cusolver per walker.

[...]

PR #2880, opened 2021-02-08
Manage crowd shared resource

ye-luo:
## Proposed changes
Add Resource and ResourceCollection for managing shared resource of a set of walkers (crowd/population).
They are unit tested.
introduce CUDALinearAlgebraHandles resource to MatrixDelayedUpdateCUDA. No more create one per walker.

Also add more mw_ routines specialization in DiracDeterminantBatched and spline SPO.
Ready for connecting batched matrix inversion.

Address #2874 VMCBatched is ready.

- New feature

- Yes. DMC batch is broken but fixed by a follow-up PR. better to review separately.

Epyc-server

## Checklist

### DIFF
--- a/src/QMCDrivers/Crowd.cpp
+++ b/src/QMCDrivers/Crowd.cpp
@@ -62,26 +62,30 @@ void Crowd::initializeResources(const ResourceCollection& twf_resource)
     twfs_shared_resource_ = std::make_unique<ResourceCollection>(twf_resource);
 }
 
-void Crowd::lendResources()
+void Crowd::lendResources(size_t receiver)
 {
   if (walker_twfs_.size() > 0)
   {
     if (twfs_shared_resource_->is_lent())
       throw std::runtime_error("Crowd::lendResources resources are out already!");
+    if (receiver >= walker_twfs_.size())
+      throw std::runtime_error("Crowd::takebackResources receiver out of bound!");
     twfs_shared_resource_->lock();
     twfs_shared_resource_->rewind();
-    walker_twfs_[0].get().acquireResource(*twfs_shared_resource_);
+    walker_twfs_[receiver].get().acquireResource(*twfs_shared_resource_);
   }
 }
 
-void Crowd::takebackResources()
+void Crowd::takebackResources(size_t receiver)
 {
   if (walker_twfs_.size() > 0)
   {
     if (!twfs_shared_resource_->is_lent())
       throw std::runtime_error("Crowd::takebackResources resources was not lent out!");
+    if (receiver >= walker_twfs_.size())
+      throw std::runtime_error("Crowd::takebackResources receiver out of bound!");
     twfs_shared_resource_->rewind();
-    walker_twfs_[0].get().releaseResource(*twfs_shared_resource_);
+    walker_twfs_[receiver].get().releaseResource(*twfs_shared_resource_);
     twfs_shared_resource_->unlock();
   }
 }
--- a/src/QMCDrivers/Crowd.cpp
+++ b/src/QMCDrivers/Crowd.cpp
@@ -9,9 +9,15 @@
 
 #include "Crowd.h"
 #include "QMCHamiltonians/QMCHamiltonian.h"
+#include "ResourceCollection.h"
 
 namespace qmcplusplus
 {
+
+Crowd::Crowd(EstimatorManagerNew& emb) : is_shared_resource_lent(false), estimator_manager_crowd_(emb) {}
+
+Crowd::~Crowd() = default;
+
 void Crowd::clearWalkers()
 {
   // We're clearing the refs to the objects not the referred to objects.
@@ -48,6 +54,34 @@ void Crowd::setRNGForHamiltonian(RandomGenerator_t& rng)
 {
   for (QMCHamiltonian& ham : walker_hamiltonians_)
     ham.setRandomGenerator(&rng);
+}
+
+void Crowd::initializeResources(const ResourceCollection& twf_resource)
+{
+  if (!twfs_shared_resource_)
+   twfs_shared_resource_ = std::make_unique<ResourceCollection>(twf_resource);
+}
+
+void Crowd::lendResources()
+{
+  if (walker_twfs_.size() > 0)
+  {
+    if (is_shared_resource_lent)
+      throw std::runtime_error("Crowd::lendResources resources are out already!");
+    is_shared_resource_lent = true;
+    walker_twfs_[0].get().acquireResource(*twfs_shared_resource_);
+  }
+}
+
+void Crowd::takebackResources()
+{
+  if (walker_twfs_.size() > 0)
+  {
+    if (!is_shared_resource_lent)
+      throw std::runtime_error("Crowd::takebackResources resources was not lent out!");
+    is_shared_resource_lent = false;
+    walker_twfs_[0].get().releaseResource(*twfs_shared_resource_);
+  }
 }
 
 void Crowd::startBlock(int num_steps)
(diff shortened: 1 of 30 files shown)


src/QMCDrivers/DMC/WalkerControl.cpp
```diff
@@ -234,13 +235,20 @@ int WalkerControl::branch(int iter, MCPopulation& pop, bool do_not_branch)
...
+    if (!twfs_shared_resource_)
+      twfs_shared_resource_ = std::make_unique<ResourceCollection>(pop.get_golden_twf().getResource());
+
     const size_t num_walkers = walkers.size();
     // recomputed received and duplicated walkers, the first untouched_walkers walkers doesn't need to be updated.
     auto p_list =
         convertUPtrToRefVectorSubset(pop.get_elec_particle_sets(), untouched_walkers, num_walkers - untouched_walkers);
     auto wf_list = convertUPtrToRefVectorSubset(pop.get_twfs(), untouched_walkers, num_walkers - untouched_walkers);
+
+    ResourceCollectionLock<TrialWaveFunction> resource_lock(*twfs_shared_resource_, wf_list[0]);
```
PDoakORNL, line 251: I am now confused about whether twfs_shared_resource_ is expected to be persistent state or not.  Looks to me like it shouldn't be.  We are giving block scope access to the wavefunctions. If I want to clone that resource collection I also expect it to be in the gold twf but now there is a nullptr there.

src/QMCDrivers/MCPopulation.h
```diff
@@ -151,6 +151,7 @@ class MCPopulation
     auto walker_index = 0;
     for (int i = 0; i < walker_consumers.size(); ++i)
     {
+      walker_consumers[i]->initializeResources(trial_wf_->getResource());
```
PDoakORNL, line 154: Crowds only need to clone and acquire the shared resources 1 time when they are constructed.  This is called after every population adjustment.  It confuses analysis to have important setup procedures occur the first time something is called.  Also I'm am not sure that trial_wf_ always has a valid resource because above I saw WalkerControl move it to a member unique pointer.

PDoakORNL (changes requested):
Two things here are a big deal.  One is Resource Collection, its cursor, rewind and usage. I feel like if crowd owns the resource collection and is lending it to the head wf or ham in a crowd for a step almost all of this isn't needed.  It makes usage complicated and adds a bunch of state and non idiomatic behavior.

I don't think Crowd::InitializeResource is needed, the gold twf is available before a crowd is constructed so the ResourceCollection should be a constructor argument. The sequence of steps to reach a valid object should be as simple as possible.

A note on this PR, their needs to be an flex/mw api change.

The flex/mw_api should be a pure function API to make it easy to write specialized functions using it.

currently its broad out line is
```c++
mw_compute_something(walker_elements_1, walker_elements_2,...,walker_elements_n,results_1,results_2,results_n, uniform_var_1, ...)
```
because of the reliance on vtable dispatching we have to do this ugly thing
```c++
 walker_elements_1[0]->mw_compute_something(...)
```
But its not took logically dissonant though because we could just as easily use any of the walker_elements_1
I could write 

```c++
walker_elements_1.back()->mw_compute_something(...)
```

The storage of the crowd level wf or ham resource in walker 0 breaks this consistency in order to maintain the API signature. While this is fine for expedience it is a step backwards in clarity and easy of analysis. One thing we should use extra hands working on ECP for is to fix this API soon.

src/Utilities/ResourceCollection.h
```diff
@@ -0,0 +1,77 @@
...
+  ResourceCollection(const ResourceCollection&);
+  const std::string& getName() const { return name_; }
+  size_t size() const { return collection_.size(); }
+  void printResources() const;
+
+  size_t addResource(std::unique_ptr<Resource>&& res);
+  std::unique_ptr<Resource> lendResource();
+  void takebackResource(std::unique_ptr<Resource>&& res);
+
+  void rewind(size_t cursor = 0) { cursor_index_ = cursor; }
```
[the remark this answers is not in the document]
ye-luo: I tried the index based approach failed. The reason many classes under TWF doesn't has a proper copy constructor implemented and used constructor for making copies. The recorded index of resources got lost after making a clone. For this reason, I have to rely on an external cursor. This is fixable once copy constructor issues are addressed.

src/QMCDrivers/Crowd.cpp
```diff
@@ -50,6 +56,40 @@ void Crowd::setRNGForHamiltonian(RandomGenerator_t& rng)
     ham.setRandomGenerator(&rng);
 }
 
+void Crowd::initializeResources(const ResourceCollection& twf_resource)
```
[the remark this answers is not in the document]
ye-luo: So far, Crowd doesn't know the population at its constructor. It doesn't have access to the gold_twf. This this two step initialization is needed. Probably you can have a look the population and crowd construction logic and I support getting ride of this two step thing.

src/QMCDrivers/Crowd.cpp
```diff
@@ -50,6 +56,40 @@ void Crowd::setRNGForHamiltonian(RandomGenerator_t& rng)
...
+
+void Crowd::lendResources(size_t receiver)
+{
+  if (walker_twfs_.size() > 0)
+  {
+    if (twfs_shared_resource_->is_lent())
+      throw std::runtime_error("Crowd::lendResources resources are out already!");
+    if (receiver >= walker_twfs_.size())
+      throw std::runtime_error("Crowd::takebackResources receiver out of bound!");
+    twfs_shared_resource_->lock();
```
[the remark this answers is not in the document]
ye-luo: As I explained above, the cursor is a compromise so far. I want the index method initially but has to rollback for now.

src/QMCDrivers/DMC/WalkerControl.cpp
ye-luo: WalkerControl is outside any crowds and it need some recompute capability. For this reason, it owns a collection of resource copied from the golden_twf. The golden TWF should not be touched.

src/QMCDrivers/MCPopulation.h
ye-luo: I also expected creating resources once for the crowd. However, this is the only place I found access to both crowd and population. It ends up here.

I think this template distributeWalkers is a bit strange to me as well. Crowd and MCPopulation should know each other since a fat walker in a Crowd and MCPopulation should be identical.

ye-luo:
The crowd level resource is given to the first fat walker at a give scope and locks are implemented to secure such scopes.
In the #2881 for Tmove, I have to loop over single walkers and thus the shared resource is given to specific walker and takeback at the every iteration. So resources are managed in blocks.

I don't feel right having flex_/mw_ API fully static (stateless).
```
static TWF::flex_evaluate(TWF_list)
{
   TWF_list[0]->something // is bad.
}
```
I prefer 
```
TWF::flex_evaluate(TWF_list) const
{
   something (shared resource) can be taken from the caller.
}
```
We should also not assume caller TWF= TWF_list[0].

ye-luo:
Here is my summary,
1. ResourceCollection cursor should be removed as we further secure copy constructors.
2. Sort out a better handling of crowd resource initialization.
3. Keep refining the batched APIs.
Unfortunately there is not much I can do in this PR regarding these 3 points quickly.

PDoakORNL:
Yes it sounds like you were trying for something I would have liked more but we have more to do before we can get there. And merging this PR will make it possible for others to try approaches to 1,2,3 with an implementation that at least gets us some resource control.

PDoakORNL (approved):

PR #2882, opened 2021-02-08
Make hdf5 io interfaces respect size_t and reduce NVHPC compiler warnings

ye-luo:
## Proposed changes
Make io/hdf only accept size_t not int. Seems a nice thing to have.
PGI complains about "void static". Changed to "static void"
Make some suspicious virtual function overriding explicit.

Restart tests pass.

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):

PR #2891, opened 2021-02-10
Fix zion test with legacy CUDA

prckent:
## Proposed changes

Ion-ion coulomb test is always computed at full precision in the legacy CUDA code. Always use the full precision reference value.

(Also here: ignore the NiO 48 h5 file link created during testing in the git repo)

- Bugfix

- No

Nitrogen. CPU and CUDA builds, full and mixed precision.

## Checklist


.gitignore
```diff
@@ -10,3 +10,4 @@ qmcpack_manual.pdf
 /build_*/
 docs/_build
 nexus/tests/unit/*output
+tests/solids/NiO_a4_e48_pp/NiO-fcc-supertwist111-supershift000-S1.h5
```
ye-luo, line 13: I have this file locally as well.

ye-luo (approved):

.gitignore
prckent: All the tests in NiO_a4_e48_pp handle this file specially. The file ought to be linked into the cmake build+test directories. I think dirtying the git repo is bad practice, so this is just a fast workaround to hide it.

PR #2902, opened 2021-02-11
Refine input classes

ye-luo:
## Proposed changes
ParameterSet and AttributeSet uses OhmmsParameter which has been updated.
Review should focus on changes under `io/OhmmsData`
'unit' has been removed from OhmmsParameter. It seems not being used at all and clearly the use side got confused a lot.

Add ways to note and input tag
```
enum class TagStatus
{
  OPTIONAL, // there is a default value
  REQUIRED, // no default value, input required.
  DEPRECATED, // deprecated input tag. Input becomes optional
  DELETED // deleted input tag. Hard stop if detected.
};
```

Also allow adding candidate_values to check against.
```
  /*!\fn OhmmsParameter(T& a, const char* aname, const char* uname)
   *\param a the value to be referenced
   *\param aname the name of this object
   *\param candidate_values candidate values to be checked against, the first element is the default value. It can be left empty for backward compatibility or unbounded input.
   *\param status Tag status, See TagStatus enum.
   */
  OhmmsParameter(T& a, const char* aname, std::vector<T>&& candidate_values = {}, TagStatus status = TagStatus::OPTIONAL)
```

- New feature

- No

epyc-server

## Checklist


ye-luo:
intended to use this for finishing #2878

prckent (approved):
Massively cleaner. Once the optional/required/default etc. settings are initial used in your other PR we should add them strategically elsewhere.

PR #2904, opened 2021-02-11
Build ppconvert executable when configured

prckent:
## Proposed changes

When BUILD_PPCONVERT is enabled, build the executable by default. In case of problems, user can simply set BUILD_PPCONVERT=0. Enables ppconvert test to easily run.

- Build related changes

- No

Nitrogen. llvmdev.

## Checklist

- NA. This PR is up to date with current the current state of 'develop'


ye-luo (dismissed):

ye-luo (approved):

correaa:
@markdewing @prckent 

Please add `numdiff -a 0.0001 ../build/src/QMCTools/ppconvert/test/Oxygen/O.BFD.xml ../tests/solids/monoO_1x1x1_pp/O.BFD.xml` to the tests, (with dependency or as part of `ppconvert_does_not_crash`). I don't know how to do this.

Also consider making `-DBUILD_PPCONVERT=1` the default for cmake (if not already).

prckent:
@correaa copy your request to an issue please so that we don't lose it

We can't make BUILD_PPCONVERT the default until we have verified that it builds with all the usual compilers. A simple step, but it needs to be done. We can do that after your changes. It is an optional build in part because we can't risk breaking the main build and in part because a long while back it had no cmake configuration.

PR #2908, opened 2021-02-12
Make batched and non-batched delayed update implementation consistent

ye-luo:
## Proposed changes
1. Some signs are opposite. Now aligned. Some names are made less confusing.
2. Use scoped timers in single det classes.

- No

Epyc-server

## Checklist


src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
```diff
@@ -166,30 +164,32 @@ typename DiracDeterminant<DU_TYPE>::PsiValueType DiracDeterminant<DU_TYPE>::rati
...
-      simd::dot(invRow.data(), dspin_psiV.data(), invRow.size());
-  RatioTimer.stop();
+  {
+    ScopedTimer local_timer(&RatioTimer);
+    UpdateMode             = ORB_PBYP_PARTIAL;
+    const int WorkingIndex = iat - FirstIndex;
+    assert(WorkingIndex >= 0);
+    // This is an optimization.
+    // check invRow_id against WorkingIndex to see if getInvRow() has been called already
+    // Some code paths call evalGrad before calling ratioGrad.
```
prckent, line 180: Are there code paths that we can get rid of / simplify in that case? What is the use case for both calls?

prckent (approved):

src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
ye-luo: Calling evalGrad before ratioGrad is required by VMC driver and DMC. To remove this if-statement, we need to inject evalGrad where there is ratioGrad. So this is actually more of a safety mechanism.

ye-luo: I will update the notes in an upcoming PR.

PR #2915, opened 2021-02-15
Nexus: Add h-channel to list of possible local channels in pseudopotential.

mcbennet:
## Proposed changes

When writing pseudopotential to QMCPACK form, there is an issue if the local channel is _h_ -- namely, the _l_-value of the local channel (l_local) is set to -1 in this case. This PR resolves this. 

Note: The channels are now coded up to 'i'. Furthermore, I think the CASINO code also has the same issue. If so, I can fix this as well along with this PR.

- Bugfix

- No

Laptop

## Checklist


prckent (approved):

prckent:
OK to test

ye-luo (approved):

Issue #557, opened 2017-12-11
Expensive choices of non-local integration grids


prckent:
The printed "spherical grids" on the output are not consistent with the highest l channel, or (say) a very conservative lmax+1 default.

We appear to be using 12 points always, even for lithium. For anyone not familiar, see e.g. Lubos Mitas https://doi.org/10.1063/1.460849 and references within.

From NiO performance test: e.g.  https://cdash.qmcpack.org/CDash/testDetails.php?test=1240418&build=15523

O: 12 points
Ni: 12 points

From graphite: e.g. https://cdash.qmcpack.org/CDash/testDetails.php?test=1240439&build=15523

C: 12 points

LiH dimer pp: e.g. https://cdash.qmcpack.org/CDash/testDetails.php?test=1240520&build=15523

Li: 12 points
H: 0 points (Thankfully!)

While the current choice is not "wrong" it is so far from optimal that I am calling this a bug. (Separately we need an optional override for the integration rule for each atom in the ECP Hamiltonian definition)

jtkrogel:
Use of 12 points seems to be a long term choice.  Runs from 2010 involving Si, BN, and MnO all use 12 point grids.

jngkim:
Use nrule : Grep Nrule in QMCHamiltonians

<semilocal *units*="hartree" *format*="r*V" *npots-down*="2" *npots-up*="0"
 *l-local*="0">

ye-luo:
Found my old 32 atom TiO2 VMC calculation with 18-point Ti.
18-point
LocalEnergy           =        -1447.6878 +/-           0.0038
12-point
LocalEnergy           =        -1447.6819 +/-           0.0022
Within 2 sigma.

prckent:
@jngkim Thanks. I see all the plumbing appears to be there, and nrule is even briefly mentioned in the manual :-)

prckent:
@ye-luo 12 points formally does l=5 exactly.

prckent:
Rule 4, 12 points, is in the constructor , ECPComponentBuilder.cpp:31 (below). We could have a smarter default.

```
ECPComponentBuilder::ECPComponentBuilder(const std::string& aname, Communicate* c):
  MPIObjectBase(c),
  RcutMax(-1), NumNonLocal(0), Lmax(0), AtomicNumber(0), Zeff(0), Species(aname), Nrule(4),
  grid_global(0),pp_loc(0), pp_nonloc(0)
```

[...]

PR #2921, opened 2021-02-15
setting sensible default based on exact angular integration

camelto2:

## Proposed changes

here is a solution to the reasonable default choice of angular quadrature choice based on the value of lmax in the pseudo. The choices of Nrule are based on [J. Chem. Phys. 95 (3467) (1991)](https://aip.scitation.org/doi/10.1063/1.460849)
I'm setting this as a WIP until these defaults can be validated, but I wanted to present a simple solution. 

Assuming the defaults are reasonable in most cases, this should be fine. If the user wants to override this, adding "nrule=XXX" to the pseudopotential xml file is all that is needed since it is already parsed. I think this is the simplest solution and requires the least changes to the code.

- New feature

- No

My workstation

## Checklist


prckent:
See also https://github.com/QMCPACK/qmcpack/issues/557

src/QMCHamiltonians/ECPComponentBuilder.cpp
```diff
@@ -207,6 +207,43 @@ bool ECPComponentBuilder::put(xmlNodePtr cur)
   }
   if (pp_nonloc)
   {
+    if (Nrule == -1)
+    {
+      app_log() << "Quadrature rule not determined from pseudopotential xml file. Setting sensible default" << std::endl;
```
prckent, line 212: Move after the switch and print which nrule was chosen.

ye-luo: Please move the whole switch to a function.

prckent: @ye-luo I think this is fine as is. Function won't help much, just more indirection for the reader. A strong argument for a function would be needing this switch more than once. However this is init only and we don't.

camelto2: I moved the print statement. I can change it to a function if necessary

camelto2:
I think leaving the override for the Nrule in the pseudopotential xml file is simplest solution.

However, if you want the user override to be at the qmcpack input level, then it can be added in the input like
`
            <pseudo elementType="W" href="W.qmcpack.xml" nrule="X"/>
`
 and the nrule could be passed as a constructor argument to ECPComponentBuilder. Then if it doesn't see it as an argument, it would set the defaults as above. 
 
 That approach would potentially be problematic though. Nrule is already read from the pseudo input file. So to avoid having overrides in two different places that potentially could disagree with one another, we would need to remove the Nrule setting from the pseudo xml file. That would change the expected behavior of the code for users who have been setting Nrule already in the pseudopotential file.

src/QMCHamiltonians/ECPComponentBuilder.cpp
```diff
@@ -207,6 +207,43 @@ bool ECPComponentBuilder::put(xmlNodePtr cur)
...
+      //nonlocal channels.
+      //From J. Chem. Phys. 95 (3467) (1991)
+      switch (pp_nonloc->lmax)
+      {
+      case 0:
+        Nrule = 1;
+        break;
+      case 1:
+      case 2:
+        Nrule = 2;
```
jtkrogel, line 222: I recommend we keep a floor at nrule=4.  I'm pretty sure some d-local potentials have higher variance at these lower settings.

prckent:
Print a warning if nlrule is already set. Use case is someone testing the required integration rule. They should not have to modify the ECP file to do that. No need to over think this.

src/QMCHamiltonians/ECPComponentBuilder.cpp
jtkrogel: Also, we don't want to degrade the statistical performance of existing inputs.

prckent:
I am also OK with leaving the default minimum at nrule=4 as Jaron suggests. Changing this might surprise a few people. But we allow override.


( These options become more important the slower the basis function and single orbital evaluation is. Not very important for splines, but potentially important for other choices.)

src/QMCHamiltonians/ECPComponentBuilder.cpp
camelto2: changed the floor to nrule=4

camelto2:
I changed the nrule=4 as Jaron suggested. 
I also added the ability to modify nrule in the qmcpack input. 
Essentially, we only need to add 
<pseudo elementType="" href="" nrule=""/> to override the default settings. 

In the case that nrule is set in the pseudopotential input file, it prints a warning that it should be set/modified from the qmcpack input file, but proceeds using the value set in the pseudopotential file.

ye-luo:
> In the case that nrule is set in the pseudopotential input file, it prints a warning that it should be set/modified from the qmcpack input file, but proceeds using the value set in the pseudopotential file.

Not getting the logic here. I was thinking of if nrule
1. not in qmcpack input but in pp file. take pp file value. no warning
2. neither in qmcpck input nor in pp file, use qmcpack fallback default which needs to be mentioned as a warning.
3. in qmcpack but not in pp file. take qmcpack input. 
4. both qmcpack and pp file provide nrule. Print a warning that qmcpack input overrides the value.

always print warning that pp file doesn't have nrule.

camelto2:
> > In the case that nrule is set in the pseudopotential input file, it prints a warning that it should be set/modified from the qmcpack input file, but proceeds using the value set in the pseudopotential file.
> 
> Not getting the logic here. I was thinking of if nrule
> 
> 1. not in qmcpack input but in pp file. take pp file value. no warning
> 2. neither in qmcpck input nor in pp file, use qmcpack fallback default which needs to be mentioned as a warning.
> 3. in qmcpack but not in pp file. take qmcpack input.
> 4. both qmcpack and pp file provide nrule. Print a warning that qmcpack input overrides the value.
> 
> always print warning that pp file doesn't have nrule.

In almost all the pp files I've found, (pseudopotentiallibrary, pseudos in test directory, etc) the xml file does not have nrule by default. So it seems that almost every run will have a warning printed since almost no pseudo sets nrule. 

1, 2 & 3 are currently satisfied.  I can modify it to let qmcpack override instead of the pseudo input. The only change needed is to make the qmcpack input > pseudo input and decide where exactly we want warning vs. just printing the information.

camelto2:
Ok, warnings should be consistent with @ye-luo's previous comment. I also moved the default setting to the same location inside ECPComponentBuilder::buildSemiLocalAndLocal so that it is all in the same place

prckent:
Echoing my agreement with Ye's list of warnings priorities and overrides.

Indeed, many existing runs and potentials will result in warnings being printed, but I think this the appropriate.

src/QMCHamiltonians/ECPComponentBuilder.2.cpp
```diff
@@ -49,17 +49,66 @@ void ECPComponentBuilder::buildSemiLocalAndLocal(std::vector<xmlNodePtr>& semiPt
...
+  aAttrib.add(quadRule, "nrule");
   aAttrib.add(Srule, "srule");
   aAttrib.add(nso, "npots-so");
 
   xmlNodePtr cur_semilocal = semiPtr[0];
   aAttrib.put(cur_semilocal);
+
+  if (quadRule > -1 && Nrule > -1)
+  {
+    app_warning() << " Nrule setting found in both qmcpack input and pseudopotential file."
```
ye-luo, line 67: Please echo both values.

src/QMCHamiltonians/ECPComponentBuilder.2.cpp
```diff
@@ -49,17 +49,66 @@ void ECPComponentBuilder::buildSemiLocalAndLocal(std::vector<xmlNodePtr>& semiPt
...
+                  << " Using nrule setting in qmcpack input file\n";
+  }
+  else if (quadRule > -1 && Nrule == -1)
+  {
+    app_log() << " Nrule setting found in pseudopotential file.\n";
```
ye-luo, line 72: "Nrule setting found in pseudopotential file and used."
std::endl incluses a flush. it is preferred to "\n"

src/QMCHamiltonians/ECPComponentBuilder.2.cpp
```diff
@@ -49,17 +49,66 @@ void ECPComponentBuilder::buildSemiLocalAndLocal(std::vector<xmlNodePtr>& semiPt
...
+    Nrule = quadRule;
+  }
+  else if (quadRule == -1 && Nrule > -1)
+    app_log() << " Nrule setting found in qmcpack input file.\n";
```
ye-luo, line 76: as above.

src/QMCHamiltonians/ECPComponentBuilder.2.cpp
```diff
@@ -49,17 +49,66 @@ void ECPComponentBuilder::buildSemiLocalAndLocal(std::vector<xmlNodePtr>& semiPt
...
+  
+  if (Nrule == -1)
```
ye-luo, line 78: Prefer this to be an else connected tot he above to make it clear 4 cases.

ye-luo (commented):
Only some minor issue left. Thanks.

ye-luo:
Still WIP?

ye-luo:
> Echoing my agreement with Ye's list of warnings priorities and overrides.
> 
> Indeed, many existing runs and potentials will result in warnings being printed, but I think this the appropriate.

In LLVM, to differentiate from Warning, Remark is used.

camelto2:
> Still WIP?

Nope, the WIP was originally for if we were going to change to smaller nrules as the default for smaller angular momentum channel projector. But with the default set to nrule=4 for most pseudos, I think we are good to go.

ye-luo:
Okay to test

ye-luo (approved):

PR #2928, opened 2021-02-17
Introduce RefVectorWithLeader

ye-luo:
## Proposed changes
Intended to solve magic walker 0 in batched APIs.
RefVectorWithLeader<T> contains a collection of T plus a leader of type T as well.
The leader does the vtable dispatch and owns the shared resource.
The leader should not assumed to be the element 0 of the collection.

1. When there is no inheritance, like ParticleSet, TWF, QMCHamiltonian.
```
static void ParticleSet::flex_update(const RefVectorWithLeader<ParticleSet>& p_list, ...)
```
mw_/flex_ APIs are static functions, namely not a member of an object.

2. When inheritance is used, like SPOSet, WFC.
```
void SPOSet::mw_evalauteVGL(const RefVectorWithLeader<SPOSet>& spo_list, ...) const
```
All the mw_/flex_ APIs should be const. `assert(this == spo_list.getLeader())` is used to secure consistency vtable handling.

Prefer 1 to 2 but 2 is necessary to handle virtual function dispatch.

After doing this, the shared resources are more confined. Also error-prone type casting are wrapped in functions as well as `get()` from reference wrappers.

Crowds are not touched as the leader is defined in the driver not in the crowd.

Some comments from @PDoakORNL were an early discussion on the prototype.

- New feature

- No

Epyc-server

## Checklist


prckent (changes requested):
I don't have bandwidth to write a nuanced argument, but for me this fails the basic code smells of whether it is clear what is going on or whether the code is simple to understand. It is not, or at least I need a lot of convincing to hold my nose. Better a separate collection for the shared resource than the current mixed situation? BagOfThingsWithResource. What are other possible ways of solving the problem?

PDoakORNL (approved):
A step toward shared resource management. Which I think will be easier to continue to refactor once the consumption of the shared resources is sorted out.

prckent (approved):
After discussion I think that this is definitely a step in the right direction. In particular we need an official mechanism for handling shared resources, e.g. handles from libraries etc. Hopefully clearer naming will emerge as this become used and refined.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -337,11 +340,11 @@ void QMCCostFunctionBatched::checkConfigurations()
         // differently for each configuration.  Make sure the same initialization is
         // performed in correlatedSampling.
         *opt_data.get_rng_ptr_list()[ib] = opt_data.get_rng_save();
-        h_list[ib].get().setRandomGenerator(opt_data.get_rng_ptr_list()[ib].get());
+        h_list[ib].setRandomGenerator(opt_data.get_rng_ptr_list()[ib].get());
       }
 
       // Compute distance tables.
-      ParticleSet::flex_update(p_list);
+      ParticleSet::flex_update(p_list, true);
```
ye-luo, line 347: @markdewing this change I made can be suspicious.

markdewing: I'm testing it now.

markdewing: Yes, that is it.  The default value of skipSK is false.

PR #2930, opened 2021-02-17
Update Cades build script to use GCC 7 and parallel HDF5

tiihonej:
## Proposed changes

Update Cades build script based on new modules (mainly GCC 7 and parallel HDF5) created and suggested by Cades Support. Glory to @mcbennet for making the request!

- Build related changes

- No

- Passes real/complex deterministic tests on Cades Skylake

## Checklist


mcbennet (changes requested):
Thanks Juha! Perhaps modifying the comments in the file header is needed.

prckent (commented):
GCC 7.20 was released on 14 August 2017. Besides being well outside our 2 year support window, there are many more recent versions with useful bug fixes, performance increases etc. If a GCC compiler will be used v10.2.0 would be by far preferable.

Also, since it is an Intel system, the latest Intel compiler 2020/2021 would give the highest performance on this system.

Do CADES have any better compiler choices? Please point them at this PR if needed.

prckent:
I have checked the system. They have GCC 9.2.0. Please use that instead. Still GCC 10 would be preferable.

mcbennet:
> I have checked the system. They have GCC 9.2.0. Please use that instead. Still GCC 10 would be preferable.

Configuration step with this version of GCC produces the following output:

`#error too new libstdc++ from GNU for Intel 19, use GNU version <= 7`

tiihonej:
> Thanks Juha! Perhaps modifying the comments in the file header is needed.

You're right, thanks, I'll update this once we've resolved GCC.

> I have checked the system. They have GCC 9.2.0. Please use that instead. Still GCC 10 would be preferable.

I second what Chandler said: Cades doesn't seem to have Intel 2020/2021 available (yet), and for whatever reason Intel 19 won't work with GCC>7. I'd say we keep GCC 7 on Cades, which works fine (apart from potential performance loss) until they upgrade their Intel PE, which probably won't happen overnight but should be requested by us on these grounds.

prckent:
They are going to install GCC 10 and look into newer Intel compilers. We can revisit then.

prckent (approved):

tiihonej:
> They are going to install GCC 10 and look into newer Intel compilers. We can revisit then.

Great, thank you for finding out!

ye-luo (approved):

prckent:
FYI, GCC 10.2.0 was just installed. No need to hold up this PR.

ye-luo:
Okay to test

PR #2934, opened 2021-02-18
Fix determinisitic opt test failures in offload

ye-luo:
## Proposed changes

1. @markdewing suggested this fix. Removing the 1 electron special case which has been covered by the generic case.
2. Add batch="no" to deterministic tests which have corresponding batch="yes" tests already.

- Bugfix

- No

bora

## Checklist

### DIFF
--- a/src/QMCDrivers/DMC/DMCBatched.cpp
+++ b/src/QMCDrivers/DMC/DMCBatched.cpp
@@ -155,7 +155,6 @@ void DMCBatched::advanceWalkers(const StateForThread& sft,
           for (int i = 0; i < rr.size(); ++i)
             assert(std::isfinite(rr[i]));
 #endif
-          auto elecs = walker_elecs;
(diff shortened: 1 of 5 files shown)


prckent (commented):
Have you tried a 1 particle input in full QMCPACK after these changes? This seems like a gap in our test coverage.

Why are we disabling batched determinants in the short vmcbatch case? Is the idea to test this execution path or something else?

ye-luo:
> Have you tried a 1 particle input in full QMCPACK after these changes? This seems like a gap in our test coverage.
> 
We do have He which is one electron test under label example but that is not as determinstic test.

> Why are we disabling batched determinants in the short vmcbatch case? Is the idea to test this execution path or something else?
There are specific tests with sdbatch added in the names. The difference is batch=yes/no in slater determinant. Without this change, both tests were running batch=yes.

prckent (approved):

Issue #2698, opened 2020-09-10
Multideterminant code fails with only alpha determinants (no down electrons)


anbenali:
**Describe the bug**
When beta determinants are absent / only up electrons, the multideterminant code is broken. QMCPACK will request removing particle set for beta electrons with size 0, but multideterminant will require and sposet down. An sposet down with size 0 will force qmcpack to exit asking to remove particleset with 0 particle.    

**To Reproduce**
1. download example files from link: https://anl.box.com/s/p3xn8r64rpk8h0zj3nredxjyqs97e81o
2. run:
 >> qmcpack Li-quadruplet-MD.qmc.in-wfj.xml 

**Expected behavior**
Code will fail if particle set with down electron specified with size 0; 

>> Fatal Error. Aborting at ParticleSet::resetGroups() Failed. ParticleSet 'e' has group 'd' containing 0 particles. Remove this group from input!

When commented out from Li-quadruplet-MD.structure.xml 
>> Norm of ci vector (sum of ci^2): 1
Found 40 unique up determinants.
Found 1 unique down determinants.
Number of terms in pairs array: 39
Fatal Error. Aborting at Error in SlaterDetBuilder::createMSDFast, problems with ci configuration list. 
Abort(1) on node 0 (rank 0 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0

**System:**
all

camelto2:
Also relevant for molecular spin-orbit calculations, as there is only electron group by construction

camelto2:
For reference, I have a hacked version of qmcpack that works for this. Essentially, every variable that is hardcoded as X_up and X_dn I have generalized to be a vector called Xs and depends on the number of groups in the electron particleset. Since this part of the code is being redesigned in the future, in case anyone needs this working for a particular problem feel free to contact me for the hacked code

ye-luo:
VMC runs fine after removing beta
```
    <determinantset type="MolecularOrbital" name="LCAOBSet" source="ion0" transform="yes" cuspCorrection="yes" href="Li-quadruplet-MD.h5">
      <sposet basisset="LCAOBSet" name="spo-up" size="18" cuspInfo="spo-up.cuspInfo.xml">
        <occupation mode="ground"/>
        <coefficient size="18" spindataset="0"/>
      </sposet>
      <multideterminant optimize="no" spo_0="spo-up">
[...]

Issue #2735, opened 2020-10-02
Spin-orbit input logic simplification


ye-luo:
See discussion in https://github.com/QMCPACK/qmcpack/pull/2733#issuecomment-701734214. Expose minimal input needs to users.

PR #2935, opened 2021-02-18
MSD arbitrary species input

camelto2:
## Proposed changes

This PR generalizes the multi-slater build path. Previously, everything was hard wired to have alpha and beta electron species. The PR adds a new build path that can handle any number of fermion species to occupy the multi slater determinants. This partially addresses #2698 in that you could now use only "alpha" electrons (useful for spinors) or additional species for other applications. 

essentially, the input (if you want variable number of species) becomes
<multidetermiant fast="yes" optimize="no"  spo_0="name0" spo_1="name1" ... spo_(N-1)="nameN-1">
where the number of species is determined by the particle set. 
 
 and (in the xml example) a similarly modified det list
 
 <detlist size="5" ... nstates="8" nc0="0" ne0="5"  nc1="0" ne1="2" nc2="0" ne2="4" ...>
 
 and the CI specifcation
    
 <ci ... qchem_coeff="-0.8586" occ0="11111000" occ1="00101000" occ2="01101010"/>
          
The changes are similar for the hdf5 input style. 

This keeps all backwards compatibility, so all current unit tests/input will continue to work.

- New feature

- No. 
The new input style works and preserves the previous approach so all previous calculations/formats behave as expected. However, there does need to be an additional PR to properly generalize the MultiSlaterDeterminantFast evaluation routines to handle arbitrary sum_i c_i x (an arbitrary product of determinants). See #2947 which addresses the changes necessary for MultiSlaterDeterminantFast. 
My workstation
## Checklist


ye-luo (changes requested):
Hopefully my reaction is not totally a surprise.
Once  https://github.com/QMCPACK/qmcpack/issues/2735 is solved
I don't see why we need a new input tag ngroups
Also the old builder code-path should converge with the new code path, convince me if you really think it is needed to keep two versions.

camelto2:
> Hopefully my reaction is not totally a surprise.
> 
> 1. Once  #2735 is solved
>    I don't see why we need a new input tag ngroups. I mean ngroups can be figured out by query the electron particle set.
Sure that can be done. I was essentially just doing a first pass to get some comments. 

> 2. Also the old builder code-path should converge with the new code path, convince me if you really think it is needed to keep two versions. It seems the new code is a good move but we need to kill the old one. Also need input tests like test_spo_collection_input_LCAO_xml.cpp and test_spo_collection_input_MSD_LCAO_h5.cpp.
I don't think we need to support both paths once the evaluation and everything is cleaned up. I separated everything for now just to make sure I didn't break anything that currently worked.

ye-luo:
@camelto2 I think we don't need to wait for the fix on the evaluation side, as long as the generic input parsing can support the legacy parsing, then it is good enough for merging with some safeguard.
our input class do support multiple tags mapped to the same variable.
mparam.add(occ, "occ1") which is handled by the loop
mparam.add(occ, "occ_up") which can be added for backward compatibility.

camelto2:
> @camelto2 I think we don't need to wait for the fix on the evaluation side, as long as the generic input parsing can support the legacy parsing, then it is good enough for merging with some safeguard.
> our input class do support multiple tags mapped to the same variable.
> mparam.add(occ, "occ1") which is handled by the loop
> mparam.add(occ, "occ_up") which can be added for backward compatibility.

Ahh, I didn't think about that. I can update that accordingly to take care of this backward compatibility issue

ye-luo:
Also ParticleSet::groups() returns how many kinds of quantum particles are there.

camelto2:
I removed the need for separate build paths for the MSD fast algorithm. In cases where the number of species is two, I also allow for backwards compatibility. For example, in the hdf5 for two particle species, most hdf5 files will have "CI_Alpha" "CI_Beta". So if reading CI_0, CI_1 fails, it tries CI_Alpha and CI_Beta. Same in the xml path. 

I still have overloaded readDetList and readDetListH5, which is used for the legacy all_determinants algorithm, but that can be changed  as well  using the newer builder functions.  Updating that now

camelto2:
I added a unit test for the xml style MSD input, with an arbitrary number particle species to make sure everything in constructed correctly. I also removed the need for overloaded readDetList and readDetListH5 so that we can reduce the amount of code. Backwards compatibility is retained for both MultiSlaterDeterminant and MultiSlaterDeterminantFast.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -204,46 +204,48 @@ WaveFunctionComponent* SlaterDetBuilder::buildComponent(xmlNodePtr cur)
       if (multislaterdet_0 || multislaterdetfast_0)
         APP_ABORT("multideterminant is already instantiated.");
 
-      std::string spo_alpha_name;
-      std::string spo_beta_name;
+      int nGroups = targetPtcl.groups();
```
ye-luo, line 207: I found it a good habit to put a const in this case when a local short name variable is used for convenience. const protects unintended modification.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -204,46 +204,48 @@ WaveFunctionComponent* SlaterDetBuilder::buildComponent(xmlNodePtr cur)
...
+      std::vector<SPOSetPtr> spos(nGroups);
+      std::vector<std::unique_ptr<SPOSet>> spo_clones(nGroups);
 
-      if (spo_beta == nullptr)
+      for (int grp = 0; grp < nGroups; grp++)
       {
-        app_error() << "In SlaterDetBuilder: SPOSet \"" << spo_beta_name
-                    << "\" is not found. Expected for MultiSlaterDeterminant.\n";
-        abort();
+        spos[grp] = sposet_builder_factory_.getSPOSet(spoNames[grp]);
```
ye-luo, line 239: The pointer return from getSPOSet should be treated as a reference. It was kept pointer for checking nullptr. Please assign the value to a temporal pointer variable. In the current way, the SPOSet actually got destroyed when you reset the smart pointer.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -204,46 +204,48 @@ WaveFunctionComponent* SlaterDetBuilder::buildComponent(xmlNodePtr cur)
...
+        if (spos[grp] == nullptr)
+        {
+          app_error() << "In SlaterDetBuilder: SPOSet \"" << spoNames[grp]
+                      << "\" is not found. Expected for MultiSlaterDeterminant.\n";
+          abort();
```
ye-luo, line 244: Please use barrier_and_abort.

ye-luo (commented):
Great simplification. Seems only minor issues left.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
camelto2: I guess I'm confused as to what you mean. The smart pointer spo_clones[grp] calls reset, but the argument is spos[grp]->makeClone(). So the spo_clones[grp] should take ownership of the pointer being returned by makeClone(), not taking ownership of spos[grp] itself.  So I don't see how the original ptr returned by getSPOSet (spos[grp]) gets destroyed.

ye-luo: reset() calls delete on the previously stored pointer. This is not desired.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
camelto2: done

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
camelto2: done

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
camelto2: Sure, but spo_clones[grp] is initially a nullptr, so it doesn't actually call the delete. I guess reset implies that it  *was* storing something before, so it is bad practice. I changed it to be a bit more clear hopefully.

ye-luo: I was confused by spos and spo_clones. Could you remove spos and store the return of sposet_builder_factory_.getSPOSet in spo_tmp for nullptr check locally.
```
spo_clones.emplace_back(spo_tmp->makeClone());
```
should be OK.

ye-luo:
deterministic-unit_test_wavefunction_sposet fails now.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
camelto2: done

ye-luo (approved):
All look good to me.

ye-luo:
Okay to test

ye-luo:
https://cdash.qmcpack.org/CDash/viewTest.php?onlydelta&buildid=189182 failures from last night.

PR #2939, opened 2021-02-19
Change timer usage as references instead of pointers

ye-luo:
## Proposed changes
This PR reduces the use of pointers in timers. Timers returned by TimerManager::createTimer are actually a reference and will be changed to reference. It was left as a pointer because of the need for a vector of pointers. Now actually we can use reference_wrapper to hold a vector of reference, so it is better to reduce the use of pointers.
ScopeTimer constructor is also changed to take reference instead of pointers.

Time start/stop in real-space are changed by sed + clang-format with a bit manual tweaks.
```
1,$s/\([^/]*\)->start();/{ ScopedTimer local_timer(\1);/
1,$s/\([^/]*\)->stop();/}/
```
AFQMC still uses start/stop instead of ScopedTimer until pending problems are fixed.

- No

epyc-server

## Checklist

### DIFF
--- a/src/AFQMC/Estimators/BackPropagatedEstimator.hpp
+++ b/src/AFQMC/Estimators/BackPropagatedEstimator.hpp
@@ -181,7 +181,7 @@ public:
       return;
     }
 
-    AFQMCTimers[back_propagate_timer]->start();
+    AFQMCTimers[back_propagate_timer].get().start();
     int nrow(NMO * ((walker_type == NONCOLLINEAR) ? 2 : 1));
     int ncol(NAEA + ((walker_type == CLOSED) ? 0 : NAEB));
     int nx((walker_type == COLLINEAR) ? 2 : 1);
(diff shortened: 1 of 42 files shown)


src/AFQMC/Estimators/BackPropagatedEstimator.hpp
```diff
@@ -181,72 +181,74 @@ class BackPropagatedEstimator : public EstimatorBase
...
+            wgt[i] *= factors[hpos][i];
+        }
+      }
+      else if (!importanceSampling)
+      {
+        stdCVector phase(iextensions<1u>{wset.size()});
+        wset.getProperty(PHASE, phase);
         for (int i = 0; i < wgt.size(); i++)
-          wgt[i] *= factors[hpos][i];
+          wgt[i] *= phase[i];
```
prckent, line 236: Is  phase[i] guaranteed = factors[hpos][i] ?
(Not obvious from a browse of the diff on GitHub)

ye-luo: Lines are not matched in the two column view on github. I only changed timer.start/stop  to a scoped timer as an example. The added {} confused github. `git diff -b` should solve this issue.

prckent (approved):
Much cleaner.

The pointer usage around initial TimerManager timer creation is a little arcane, but there are not many instances.

PR #2942, opened 2021-02-20
revamp wavefunction restart logic in AFQMC

Paul-St-Young:
## Proposed changes

Concentrate wavefunction restart logic in `WavefunctionFactory::getHamOps`.
Allow user to restart without changing the input file as discussed in #2872

Once this PR is merged, issue #2851 should be fully addressed.

`WavefunctionFactory::fromASCII` is poised for deprecation, so I put an abort when that code path hits restart.

- Bugfix

- No

Fedora 32, E5-2620 v2, gcc 10.2.1, intel 19.1.2

## Checklist

### DIFF
--- a/src/AFQMC/Wavefunctions/WavefunctionFactory.cpp
+++ b/src/AFQMC/Wavefunctions/WavefunctionFactory.cpp
@@ -254,6 +254,8 @@ Wavefunction WavefunctionFactory::fromASCII(TaskGroup_& TGprop,
     hdf_archive dump(TGwfn.Global());
     if (restart_file != "")
     {
+      app_error() << " no restart fromASCII; use getHamOps not getHamiltonianOperations\n";
+      APP_ABORT("");
       if (TGwfn.Global().root())
       {
         if (!dump.create(restart_file, H5F_ACC_EXCL))
--- a/src/AFQMC/Wavefunctions/WavefunctionFactory.cpp
+++ b/src/AFQMC/Wavefunctions/WavefunctionFactory.cpp
@@ -248,46 +248,9 @@ Wavefunction WavefunctionFactory::fromASCII(TaskGroup_& TGprop,
       APP_ABORT("Error: Unknown wfn_type in WavefunctionFactory with MSD wavefunction.\n");
     }
     TGwfn.node_barrier();
-
-    // if requested, create restart file
-    // Will use phdf5 in the future, for now only head node writes
-    hdf_archive dump(TGwfn.Global());
-    if (restart_file != "")
-    {
-      app_error() << " no restart fromASCII; use getHamOps not getHamiltonianOperations\n";
-      APP_ABORT("");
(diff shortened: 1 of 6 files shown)


src/AFQMC/Wavefunctions/WavefunctionFactory.cpp
```diff
@@ -734,15 +736,14 @@ Wavefunction WavefunctionFactory::fromHDF5(TaskGroup_& TGprop,
...
-    {
+    app_log() << " Open " << restart_file << " for HamOps\n";
+    // first, try open existing restart file
+    if (!restart.open(restart_file, H5F_ACC_RDONLY))
+    { // make new restart file if cannot open existing
+      app_log() << " No restart_file create anew.\n";
       if (!restart.create(restart_file, H5F_ACC_EXCL))
       {
-        app_error() << " Error opening restart file in WavefunctionFactory. \n";
+        app_error() << " Error in WavefunctionFactory: Failed to create restart_file" << restart_file << "\n";
```
ye-luo, line 746: A separate issue from this PR.
If AFQMC error handling is the same as real-space part, then this error handling is probably wrong in the case of MPI runs unless this function is only called from rank 0. If all the MPI ranks hit here, app_error goes out on rank 0 only but other ranks hit abort first, then the error message is not visible to the user.

ye-luo (approved):
LGTM.
Saw a potential pathological error handling pattern in AFQMC but needs to be dealt in separate PRs.

src/AFQMC/Wavefunctions/WavefunctionFactory.cpp
prckent: Good point. Try to use the modern abort procedures consistently.

ye-luo:
Okay to test

ye-luo:
Please fix AFQMC unit tests.

Paul-St-Young:
Factory unit tests currently go through the `fromASCII`, so I enabled restart there.
The restart file is now opened and closed inside `WavefunctionFactory::getHamOps`, narrowing the scope of the `restart` hdf_archive.

These changes revealed three hidden problems:
1. Factory unit tests secretly test restart. A serial run first creates the restart file, then a parallel run (supposedly) reads that restart file.
2. fromASCII restart implementation was not working before this PR. The restart file is emptied by `hdf_archive::create` before `getHamiltonianOperations`, so every run is from scratch.
3. Once fromASCII restart is properly enabled, the factory unit test catches a bug in THC Hamiltonian restart. Fortunately, this is caught by `THCOpsIO::loadTHCOps` `if (dims[2] != NAEB)`, aborting execution.

In conclusion, after this PR, HOps restart should work for all but THC type.
However, intentional restart unit tests should be created, one for each H type.
For now, I disable the unintentional side-effect of factory unit tests, so they no longer tests restart.

ye-luo (approved):

PR #2944, opened 2021-02-22
Explicit two modes of distance table operations

ye-luo:
## Proposed changes
1. Clean up the accept/reject single particle move logic. Two modes regular and forward mode documented in PartcleSet.h
2. Full documentation for DynamicCoordinates class.

- Refactoring with api changes

- No

Epyc-server

## Checklist

### DIFF
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -2083,7 +2083,7 @@ Distance tables
 
 Distance tables store distances between particles. There are symmetric
 (AA) tables for distance between like particles (electron-electron or
-ion-ion) and asymmetric (BA) tables for distance between unlike
+ion-ion) and asymmetric (AB) tables for distance between unlike
 particles (electron-ion)
 
 The ``Distances`` and ``Displacements`` members contain the data. The
--- a/src/Particle/DistanceTableData.h
+++ b/src/Particle/DistanceTableData.h
@@ -204,9 +204,8 @@ public:
 
   /** fill the distance table by the pair relations for the old particle position. Used in forward mode when a move is reject
    * @param iat the particle with an accepted move
-   * @param partial_update If true, rows after iat will not be updated. If false, upon accept a move, the full table should be up-to-date
    */
-  virtual void updateForOldPos(IndexType jat) {};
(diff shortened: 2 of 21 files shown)


prckent (changes requested):
Mostly comment updates:

Please indicate how up to date the distance tables have to be/will be in "forward mode".

Exactly what parts of the distance tables are updated in forward mode? it would be good to add a comment stating explicitly which parts are invalid/out-of-date.

Can you add some logic (asserts?) to trip an error if the wrong kind of accept/reject call is made in forward mode or regular mode? This would made the intended logic clearer imo.

Please privatize acceptMove_impl and rejectMoveForwardMode. If you have time, privatizing more of what is currently protected would be good. We have hard evidence from historical QMCPACK development that classes should not be reaching inside other classes with state machines.

ye-luo:
> Mostly comment updates:
> 
> Please indicate how up to date the distance tables have to be/will be in "forward mode".
> 
> Exactly what parts of the distance tables are updated in forward mode? it would be good to add a comment stating explicitly which parts are invalid/out-of-date.
> 

This has been documented at ParticleSet::accept_rejectMove. Explaining the two modes and how to choose two modes and how up-to-date in forward mode is.

> Can you add some logic (asserts?) to trip an error if the wrong kind of accept/reject call is made in forward mode or regular mode? This would made the intended logic clearer imo.
> 

I have thought about this. For the moment, I don't have good solutions to write asserts. Expecting to retiring all acceptMove and rejectMove first and only use accept_rejectMove. Then I will have less entries to protect. Right now, the debug build assertion in drivers will be triggered if the wrong mode is chosen, and recomputed found different numbers. Which seems effective when I was working on this PR.

> Please privatize acceptMove_impl and rejectMoveForwardMode. If you have time, privatizing more of what is currently protected would be good. We have hard evidence from historical QMCPACK development that classes should not be reaching inside other classes with state machines.

Currently VirtualParticleSet is the only derived class of ParticleSet. So "protected" is "private" for most cases. Putting all the protected member private is probably OK but I cannot quickly draw the conclusion.

prckent (approved):
Important to keep an eye on these issues as the offload code matures.

PR #2949, opened 2021-02-23
Fix USE_NVTX_API option

markdewing:
The CMAKE_DEPENDENT_OPTION macro tokenizes the conditional on spaces, which is different than how the IF statement tokenizes. In order to handle parentheses correctly, there must be space between the parenthesis and adjacent text.

That argument joins list elements in an implicit-AND, so it could also be written as "ENABLE_TIMERS;QMC_CUDA OR ENABLE_CUDA" and avoid parentheses altogether.

For more discussion, see
https://discourse.cmake.org/t/cmake-dependent-option-with-complex-conditionals/2793

- Bugfix

- No

laptop

## Checklist


ye-luo (approved):
Good to learn this.

prckent:
Thanks Mark. Indeed a bit tricky.

prckent (approved):

PR #2952, opened 2021-02-24
Split flex_ APIs in TWF

ye-luo:
## Proposed changes
When called from the driver, the flex API is doing dispatch to single walker or multi walker on per-call basis. It is often enter corner cases when the number of walker in a crowd drops to 1. This PR change
```C++
TrialWaveFunction::flex_evaluate()
{
  if(nw >1)
  { batched operation }
  else if (nw ==1)
  { single walker }
}
```
into

```C++
TrialWaveFunction::mw_evaluate()
{
  { batched operation }
}
```
and
```C++
TWFdispatcher::flex_evaluate()
{
  if (use_batch)
    TrialWaveFunction::mw_evaluate
  else
    loop over walker
    { single walker }
}
```
use_batch is selected per driver. Input control will be added once the ParticleSet flex_ APIs are updated.

## What type(s) of changes does this code introduce?
- New feature

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
epyc-server

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'

PR #2953, opened 2021-02-24
Put spline SPO multi walker memory to shared resource and split QMCHamiltonian flex_ functions

ye-luo:
Review after #2952
## Proposed changes
1. Put spline SPO multi walker memory to shared resource
2. Put DiracDeterminantBatched multi walker memory to shared resource
3. Split QMCHamiltonian flex_ functions like #2952

1, 2 makes memory consumption mostly under control. Users should not see large memory (MB) increase as DMC progress. Still a few small buffers (KB) needs to be handled but later.

- New feature

- No

epyc-server

## Checklist


src/QMCWaveFunctions/BsplineFactory/OffloadMultiWalkerMem.h
```diff
@@ -0,0 +1,51 @@
...
+
+namespace qmcplusplus
+{
+
+template<typename ST, typename TT>
+struct OffloadMultiWalkerMem: public Resource
+{
+  template<typename DT>
+  using OffloadPinnedAllocator = OMPallocator<DT, PinnedAlignedAllocator<DT>>;
+
```
PDoakORNL, line 28: If I am a scope takes an OffloadMultiWalkerMem as an argument
what can I assume about state of the OMWM.

I'd assume mw_ratios_private, rg_private, mw_pos_copy might contain persistent data written to by some scopes maybe read by others. 

I'd assume mw_offload_scratch, mw_results_scratch, buffer_H2D I can just use for whatever I want to write transfer H2D read/write some more and then transfer D2H read.

det_ratios_buffer_H2D?

src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
```diff
@@ -392,28 +387,32 @@ void SplineC2ROMPTarget<ST>::mw_evaluateDetRatios(const RefVectorWithLeader<SPOS
                                                   std::vector<std::vector<ValueType>>& ratios_list) const
 {
   assert(this == &spo_list.getLeader());
-  auto& phi_leader      = spo_list.getCastedLeader<SplineC2ROMPTarget<ST>>();
-  const size_t nw       = spo_list.size();
-  const size_t orb_size = phi_leader.size();
+  auto& phi_leader            = spo_list.getCastedLeader<SplineC2ROMPTarget<ST>>();
+  auto& mw_mem                = *phi_leader.mw_mem_;
+  auto& det_ratios_buffer_H2D = mw_mem.det_ratios_buffer_H2D;
```
PDoakORNL, line 392: I'd rather save the vertical space than horizontal but that's your call.

src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
```diff
@@ -807,24 +801,37 @@ void SplineC2ROMPTarget<ST>::mw_evaluateVGL(const RefVectorWithLeader<SPOSet>& s
...
 {
   assert(this == &sa_list.getLeader());
-  auto& phi_leader   = sa_list.getCastedLeader<SplineC2ROMPTarget<ST>>();
-  const int nwalkers = sa_list.size();
-  phi_leader.multi_pos_copy.resize(nwalkers * 6);
+  auto& phi_leader = sa_list.getCastedLeader<SplineC2ROMPTarget<ST>>();
+  // make this class unit tests friendly without the need of setup resources.
+  if (!phi_leader.mw_mem_)
+  {
+    app_warning() << "SplineC2ROMPTarget : This message should not be seen in production (performance bug) runs but "
```
PDoakORNL, line 808: maybe an #ifndef or NDEBUG assert to avoid the conditional?

src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
```diff
@@ -376,7 +376,12 @@ class MatrixDelayedUpdateCUDA
   {
     // make this class unit tests friendly without the need of setup resources.
     if (!cuda_handles_)
+    {
```
PDoakORNL, line 379: Ok I see this is a testing convenience behavior.  Rather see this in a test helper of some sort. Figuring out how to do that easily is a separate task.

PDoakORNL (approved):
I think this can go in.

My Concerns about OffloadMultiWalkerMem and checking and allocating for testing resources in the class is something we should add to a punch list for a second pass design review after the dust has settled on shared resource refactoring. Where should we start collecting that list?

ye-luo:
Will deal with the review comments after merging to unblock beta testers.

src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
ye-luo: In this case, I prefer bring phi_leader.mw_mem_->det_ratios_buffer_H2D down to det_ratios_buffer_H2D.

src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
ye-luo: This is not a stopper but should allow continue running.

PR #2954, opened 2021-02-24
Remove Psi Ham reference in QMCDriverNew

ye-luo:
Review after #2953

## Proposed changes
The driver base class doesn't need to know the details of Psi and Hamiltonian which is about the details of MCPopulation. So gold copy access should be through MCPopulation.

- No

epyc-server

## Checklist


prckent (approved):
Fewer references -> good direction.

src/QMCDrivers/QMCDriverFactory.cpp
```diff
@@ -261,7 +261,7 @@ std::unique_ptr<QMCDriverInterface> QMCDriverFactory::createQMCDriver(xmlNodePtr
                                 MCPopulation(comm->size(), comm->rank(), qmc_system, &qmc_system, primaryPsi, primaryH)
 
                                     ,
```
ye-luo, line 263: Leave a note for myself. Delete those emoty lines in my next PR.

PR #2960, opened 2021-02-25
Fix CMake issue with XL and update Summit Clang build script

ye-luo:
## Proposed changes
Fix CMake issue with XL and update Summit Clang build script

- Bugfix
- Build related changes

- No

Summit and Epyc-server

## Checklist


prckent (approved):
Unfortunate there is no way to share outside of our INCITE project mat151 via permanent storage. https://docs.olcf.ornl.gov/accounts/olcf_policy_guide.html?highlight=world#data-storage-resources

markdewing (approved):

PR #2964, opened 2021-02-25
Split flex_ APIs in ParticleSet

ye-luo:
## Proposed changes

1, Split flex_ APIs in ParticleSet to PSdispatcher.h
2. Introduce "crowd_serialize_walkers" driver input tag to force flex_ API loop over single walker calls. default no. It covers all the flex_ calls to ParticleSet, TrialWaveFunction and QMCHamiltonian.

In batched driver
1. crowd_serialize_walkers=no default case
batched_driver -> dispatcher::flex_update -> ParticleSet::mw_update (mw_ API fails back to serial internally if no specialization exists)
2. crowd_serialize_walkers=yes. This is intended for CPU only runs to bypass any overhead or CPU unfriendly implementation in batched code paths.
batched_driver -> dispatcher::flex_update -> loop over ParticleSet::update

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/ParticleSet.cpp
+++ b/src/Particle/ParticleSet.cpp
@@ -399,25 +399,25 @@ void ParticleSet::update(bool skipSK)
 
 void ParticleSet::mw_update(const RefVectorWithLeader<ParticleSet>& p_list, bool skipSK)
 {
-    auto& p_leader = p_list.getLeader();
-    ScopedTimer update_scope(p_leader.myTimers[PS_update]);
+  auto& p_leader = p_list.getLeader();
+  ScopedTimer update_scope(p_leader.myTimers[PS_update]);
 
-    for (ParticleSet& pset : p_list)
-      pset.setCoordinates(pset.R);
+  for (ParticleSet& pset : p_list)
+    pset.setCoordinates(pset.R);
 
-    auto& dts = p_leader.DistTables;
-    for (int i = 0; i < dts.size(); ++i)
-    {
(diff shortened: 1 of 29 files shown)


src/Particle/ParticleSet.cpp
```diff
@@ -826,6 +806,15 @@ void ParticleSet::loadWalker(Walker_t& awalker, bool pbyp)
   activePtcl = -1;
 }
 
+void ParticleSet::mw_loadWalker(const RefVectorWithLeader<ParticleSet>& psets,
+                                const RefVector<Walker_t>& walkers,
+                                bool pbyp)
+{
+#pragma omp parallel for
```
PDoakORNL, line 813: The strange looking nested OMP is already called from a crowd concurrent section so I wonder why it needs more threads for what looks like a copy from one object to another. Then I reread load walker. 

I'm not sure the driver should know about the distance tables but an important cost of the algorithm is hidden from view here. The method documentation should perhaps emphasize that the this is actually a hot spot in some cases.

PDoakORNL (approved):
This looks good to me outside of prexisting documentation and design issues with Particle Set.

src/Particle/ParticleSet.cpp
ye-luo: loadWalker can be called outside the crowd scope now in WalkerControl. It can be a hot spot of quite a few walkers got exchanged and need to recompute full distance table.

PR #2969, opened 2021-02-28
Introduce debug build GL check in non-batched drivers and fix two tests

### LINKED ISSUE #2966: makecopy wrongly called in unit test wavefunction / legacy CUDA bug
**Describe the bug**

unit tests are failing for legacy CUDA build. Probably a simple call overlooked as part of API updates around test_TrialWaveFunction_He.cpp:207

Fails are consistent. e.g. https://cdash.qmcpack.org/CDash/testDetails.php?test=12136678&build=189652

**To Reproduce**

develop version, ctest -R  deterministic-unit_test_wavefunction_trialwf

**Expected behavior**

Test should pass

**System:**

nitrogen, sulfur, bora. Any legacy CUDA build.

**Additional context**

ye-luo:
## Proposed changes
1. Add debug build checkLogAndGL assertion in non-batched code. Fortunately, caught nothing.
2. Move distance table partial update logic out of DT::update() to DT::updatePartial() more straight forward logic.
3. Fix https://cdash.qmcpack.org/CDash/testDetails.php?test=9533311&build=190366
4. Fix #2966 by bypassing a unit test not suitable for legacy CUDA. 

- Bugfix

- No

Bora

## Checklist

### DIFF
--- a/src/Particle/ParticleSet.cpp
+++ b/src/Particle/ParticleSet.cpp
@@ -670,8 +670,11 @@ void ParticleSet::acceptMove_impl(Index_t iat, bool forward_mode)
   if (iat == activePtcl)
   {
     //Update position + distance-table
-    for (int i = 0, n = DistTables.size(); i < n; i++)
-      DistTables[i]->update(iat, forward_mode);
+    for (int i = 0; i < DistTables.size(); i++)
+      if (forward_mode)
+        DistTables[i]->updatePartial(iat, true);
+      else
+        DistTables[i]->update(iat);
 
     //Do not change SK: 2007-05-18
     if (SK && SK->DoUpdate)
@@ -723,8 +726,8 @@ void ParticleSet::rejectMoveForwardMode(Index_t iat)
 {
   assert(iat == activePtcl);
   //Update distance-table
-  for (int i = 0, n = DistTables.size(); i < n; i++)
-    DistTables[i]->updateForOldPosPartial(iat);
+  for (int i = 0; i < DistTables.size(); i++)
+    DistTables[i]->updatePartial(iat, false);
   activePtcl = -1;
 }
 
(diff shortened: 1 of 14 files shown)


src/Particle/ParticleSet.cpp
```diff
@@ -670,8 +670,11 @@ void ParticleSet::acceptMove_impl(Index_t iat, bool forward_mode)
   if (iat == activePtcl)
   {
     //Update position + distance-table
-    for (int i = 0, n = DistTables.size(); i < n; i++)
-      DistTables[i]->update(iat, forward_mode);
+    for (int i = 0; i < DistTables.size(); i++)
```
prckent, line 673: General question on our loops over particle set updates: are they currently required to be sequential starting from i=0? if not, can we write the loops more generally?

src/QMCDrivers/QMCUpdateBase.cpp
```diff
@@ -264,6 +264,46 @@ QMCUpdateBase::RealType QMCUpdateBase::getNodeCorrection(const ParticleSet::Part
...
+{
+  bool success         = true;
+  TrialWaveFunction::LogValueType log_value{twf.getLogPsi(), twf.getPhase()};
+  ParticleSet::ParticleGradient_t G_saved = twf.G;
+  ParticleSet::ParticleLaplacian_t L_saved = twf.L;
+
+  pset.update();
+  twf.evaluateLog(pset);
+
+  const RealType threshold = 100 * std::numeric_limits<float>::epsilon();
```
prckent, line 277: Odd indentation here. Missed clangformat?

prckent (approved):
LGTM, but please reformat and consider the general question.

OK to merge if the reformat only changes indentation.

src/Particle/ParticleSet.cpp
ye-luo: Order should not matter here and range based loop is better. It is probably good rainy day task reviewing all the loops in this file.

src/QMCDrivers/QMCUpdateBase.cpp
ye-luo: I think I forgot applying clang-format. Will get this done. when I update the other PR.

PR #2978, opened 2021-03-02
Move batched driver XML parsing printing

ye-luo:
## Proposed changes

1. Block stepsbetweensamples, samplesperthread, walkers input tag in batched drivers.
2. printout driver input parsing status.
3. Clean output of non local moves in DMCBatched.

- No

bora

## Checklist


prckent (approved):
I agree with the deprecated inputs, which aren't actually supported in any case/have easy alternatives.

There is clearly more for all of us to do in terms of cleanup + making the output more friendly.

PR #2981, opened 2021-03-03
Make QMC_SIMD_ALIGNMENT configured via CMake

ye-luo:
## Proposed changes
Renamed QMC_CLINE to QMC_SIMD_ALIGNMENT.
Set QMC_SIMD_ALIGNMENT (SIMD vector length in bytes, not necessarily cache line size) explicitly via CMake rather than depending on compiler defined macros.
The default value is determined by check  `__AVX512F__` macro. If found, set 64 otherwise 32.
Basically the compiler macro logic is moved to CMake and the value is configurable but checked in CMake.

Background:
The old code was fine when targeting only one architecture (CPU).
```C++
  #if defined(__AVX512F__)
    #define QMC_SIMD_ALIGNMENT 64
  #else
    #define QMC_SIMD_ALIGNMENT 32
  #endif
```

When OpenMP offload is used, the source code is compiled for two targets CPU and GPU.
```
int foo(Vector<float, aligned_allocator<float, QMC_SIMD_ALIGNMENT>>)
{
  #pragma omp target
  { offload kernel }
}
```
The compiler transformed offload kernel carries C++ mangled function name foo with QMC_CLINE being part of it. For this reason, the caller (CPU pass) and the callee(GPU pass) doesn't match and kernels failed being loaded on GPU. https://bugs.llvm.org/show_bug.cgi?id=45401

- Bugfix

- No

Bora, Summit, Tulip, Theta

## Checklist


CMakeLists.txt
```diff
@@ -427,6 +428,26 @@ IF (BUILD_AFQMC AND NOT APPLE)
   LINK_LIBRARIES("rt")
 ENDIF()
 
+#---------------------------------------------------------------------------
+# Check SIMD alignment for CPU only
+# This is intentionally placed before adding OpenMP offload compile options
+# to avoid contamination from device compilation pass
+#---------------------------------------------------------------------------
```
prckent, line 435: Please add an example of the potential problem, as detailed in your PR text.

prckent (changes requested):
1. Please change QMC_CLINE to a self-describing and longer name. e.g. QMC_SIMD_ALIGNMENT. What CLINE means is not obvious.

2. Rename CheckSIMDAlignement.cmake ->CheckSIMDAlignment.cmake

prckent (approved):
Thanks Ye!

ye-luo:
@prckent This should make the nightly test error on sulfur go away without need of passing "-march=skylake" as a workaround.

prckent:
Nightly scripts have been updated.

PR #2984, opened 2021-03-03
Nexus: update version ahead of release

jtkrogel:
## Proposed changes

This PR propagates the Nexus version to all executables in a guarded fashion (single source of truth) and updates the version number.


prckent:
OK to test

ye-luo (approved):

PR #2986, opened 2021-03-04
Request full ion-electron table ready at loadWalker for LCAO

ye-luo:
## Proposed changes
Request full ion-electron table ready at loadWalker for LCAO to fix spin-orbit case.

Without spin-orbit, evalGrad doesn't recompute SPO but read from stored values.
With spin-orbit, such values are recomputed. When LCAO is used, the old electron-ion pair distances are required.
For this reason, request full ion-electron table ready at loadWalker.

- Bugfix

- No

Ryzen-box

## Checklist


camelto2 (approved):
This fixes the issue. Thanks Ye

PR #2988, opened 2021-03-04
Improve QE converter test QMC robustness

prckent:
## Proposed changes

The QMC tests downstream from the QE converter tests are statistical in nature and fail too frequently. Double the error bars and also add a fixed seed.  Currently  a few builds occasionally result in wildly off energies, e.g. https://cdash.qmcpack.org/CDash/testDetails.php?test=12246670&build=191687 . Assumption is these result from particularly poor starting configurations, but if the problems persist with a fixed seed then there is a bug.

- Build related changes

- No

nitrogen, gcc10.

## Checklist


prckent:
Can we merge ahead of #2992 ? The statistical error bars need updating regardless.

ye-luo:
Will merge before #2992 but I will comment RNG seeds before merging.

ye-luo (approved):

PR #2990, opened 2021-03-04
Use samples per rank instead of node in output

markdewing:
When running multiple MPI processes per node, saying "per node" is ambiguous.  Use "per rank" instead.

Also add the global (total) number of samples to be collected.

Output before change
```
VMCBatched Driver collecting samples, samples_per_node = 25
```

Output after change
```
VMCBatched Driver collecting samples, samples per rank = 25
                                      total samples    = 100
```

- Other (please describe): Updating output

- No

laptop

## Checklist


prckent (approved):
Good catch and an improvement. Definitely confusing.

Can you please systematically eliminate these in the codebase, even if not printed for users? A quick grep on _per_node found a handful, and in nearly all cases it should be rank. We have a little bit of code to report the number of accelerators per node and to share GPU spline tables, but everything else should be per rank.

PR #2991, opened 2021-03-04
Improve XML error handling + Add success message

prckent:
## Proposed changes

Quality of life improvements: To ensure an actual error code is returned for bad XML, replace a few historical error calls with APP_ABORT.  Also print "QMCPACK execution completed successfully" just prior to application return (if no errors/exceptions etc.). Check for this in ctests to catch all early exits.

Motivation for making these changes included some bad XML inputs that I created that did not result in ctest failures.

- Bugfix
- New feature

- No

sulfur, nightly gcc10 configuration.

## Checklist


src/QMCApp/QMCAppBase.cpp
```diff
@@ -38,6 +38,7 @@ bool QMCAppBase::pushDocument(const std::string& infile)
   else
   {
     app_error() << "File " << infile << " is invalid" << std::endl;
+    APP_ABORT("QMCAppBase::pushDocument");
```
ye-luo, line 41: returning false is self-consistent. abort is handled outside. I could not agree with this change.

src/QMCApp/QMCMain.cpp
```diff
@@ -209,7 +209,7 @@ bool QMCMain::execute()
   bool success = validateXML();
   if (!success)
   {
-    ERRORMSG("Input document does not contain valid objects")
+    APP_ABORT("QMCMain::execute. Input document does not contain valid objects");
```
ye-luo, line 212: Use myComm->barrier_and_abort()

src/QMCApp/QMCMain.cpp
```diff
@@ -450,12 +450,14 @@ bool QMCMain::validateXML()
           popDocument();
         }
         else
-          myComm->abort();
+        {
+          APP_ABORT("QMCMain::validateXML");
```
ye-luo, line 454: Use myComm->barrier_and_abort().

src/QMCApp/QMCMain.cpp
```diff
@@ -450,12 +450,14 @@ bool QMCMain::validateXML()
...
+        }
       }
       else
       {
         app_error() << "tag \"include\" must include an \"href\" attribute." << std::endl;
-        myComm->abort();
+        APP_ABORT("QMCMain::validateXML")
```
ye-luo, line 460: Use myComm->barrier_and_abort() including the message sent to app_error.

src/QMCApp/QMCMain.cpp
```diff
@@ -485,17 +487,17 @@ bool QMCMain::validateXML()
...
   if (psiPool->empty())
   {
-    ERRORMSG("Illegal input. Missing wavefunction. ")
+    APP_ABORT("QMCMain::validateXML. Illegal input. Missing wavefunction.");  
     return false;
   }
   if (hamPool->empty())
   {
-    ERRORMSG("Illegal input. Missing hamiltonian. ")
+    APP_ABORT("QMCMain::validateXML. Illegal input. Missing Hamiltonian.");
```
ye-luo, line 500: Please use Use myComm->barrier_and_abort().

src/QMCApp/qmcapp.cpp
```diff
@@ -247,6 +247,7 @@ int main(int argc, char** argv)
   }
 
   OHMMS::Controller->finalize();
+  std::cout << "QMCPACK execution completed successfully" << std::endl;
```
ye-luo, line 250: We cannot print on very rank.
Already out to the MPI scope. Very hard to have a consistent logic to send clean output.
Considering we'd better not using this for ctest check. Removal is better for me.

ye-luo (changes requested):
Could you point to or provide a reproducer? Not clear where the issue is precisely.

CMake/macros.cmake
```diff
@@ -99,6 +99,7 @@ FUNCTION( RUN_QMC_APP_NO_COPY TESTNAME WORKDIR PROCS THREADS TEST_ADDED TEST_LAB
         ELSE()
             ADD_TEST( NAME ${TESTNAME} COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${PROCS} ${MPIEXEC_PREFLAGS} ${QMC_APP} ${ARGN} )
             SET_TESTS_PROPERTIES( ${TESTNAME} PROPERTIES FAIL_REGULAR_EXPRESSION "${TEST_FAIL_REGULAR_EXPRESSION}"
+                PASS_REGULAR_EXPRESSION "QMCPACK execution completed successfully"
```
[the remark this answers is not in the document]
prckent: This only tightens the pass criterion (??)

prckent:
Reproducer: Delete the / on a closing XML tag with existing code. e.g. An easy user error such as <random seed="666">

On test criteria, which are only tightened here: see https://cmake.org/cmake/help/latest/prop_test/PASS_REGULAR_EXPRESSION.html and (say ) https://cmake.org/cmake/help/latest/prop_test/SKIP_RETURN_CODE.html . All the pass conditions have to be met and all the fail conditions have to not be met.

Will update the calls and arrange for node 0 only output.

We are going to print something (said in PI voice)

CMake/macros.cmake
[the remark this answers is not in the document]
ye-luo: Tightening is good.

src/QMCApp/qmcapp.cpp
ye-luo: Since this helps tightening the pass criteria, we should add this. Please pay attention to ensemble runs.

prckent:
Maybe fixed. Not sure about ensemble runs, did check conventional MPI runs.

src/QMCApp/QMCMain.cpp
```diff
@@ -209,7 +209,7 @@ bool QMCMain::execute()
   bool success = validateXML();
   if (!success)
   {
-    ERRORMSG("Input document does not contain valid objects")
+    myComm->barrier_and_abort("QMCMain::execute. Input document does not contain valid objects");
     return false;
```
ye-luo, line 213: could you remove the return.

src/QMCApp/QMCMain.cpp
```diff
@@ -450,12 +450,13 @@ bool QMCMain::validateXML()
           popDocument();
         }
         else
-          myComm->abort();
+        {
```
ye-luo, line 453: could you remove the {} around a single line if possible.

src/QMCApp/qmcapp.cpp
```diff
@@ -218,8 +218,13 @@ int main(int argc, char** argv)
     else
       validInput = qmc->parse(inputs[0]);
     if (validInput)
+    { 
       qmc->execute();
```
ye-luo, line 222: This execute also returns true/false. Better to check.

ye-luo (dismissed):

ye-luo (dismissed):

ye-luo (approved):

PR #2993, opened 2021-03-05
Add test for multi slater determinant with NLPP

### LINKED ISSUE #2879: Need multi determinant determinstic tests
I'm surprised to see no deterministic test covering.
Multi slater determinant + NLPP
VMC + DMC (no Tmove)
16 walkers
Non batched driver.

Hyeondeok-Shin:
## Proposed changes

Close #2879 

 As some tests for multi slater determinant with NLPP is missing, This PR add statistical integration test for multi determinant + NLPP for solid for the real build and corresponding deterministic test for real/cplx/mixed precision build.

- No

## Checklist


ye-luo (dismissed):

prckent (commented):
How long are these expected to take to run?

Also, please remove the empty .h5.

ye-luo:
I ran the short ones on Bora. GNU build 10-30 second. The h5 file is not empty, probably a bug on github.
```
-rw-r--r-- 1 yeluo cels 168640 Mar  5 10:43 ../tests/solids/diamondC_2x1x1-Gaussian_pp_MSD/MultiDet.h5
```

prckent (changes requested):
Long test should be longer! e.g. 50 steps vs the 25 here. I don't think having a long test with only 5x better statistics than the short test, as now, is worth it (x2.23 on the error bar).

We aim for 10x the short https://qmcpack.readthedocs.io/en/develop/integration_tests.html#how-to-add-a-integration-test

Hyeondeok-Shin:
Addressed. Now long statistical tests have 10x more steps than short tests.

prckent:
I noticed that you updated the reference energies - these should come from the reference run. However in this case since there is no jastrow, the data could come from the numerical CI calculation. Is there any breakdown of the sCI energy available? The total energy should match, as you noted, but in principle the breakdown of the components is computable.

Hyeondeok-Shin:
We'll be able to use kinetic energy and local potential reference from QP2 output while Vee and Ven is not compatible as Vee is using density fitting and QP2 doesn't separate Ven into local and nonlocal component for now. (ion-ion is a constant with NLPP, so we keep it.) I can add reference for kinetic and local potential from QP2 and add comment to clarify where each reference come from.

prckent (approved):

prckent:
Test this please

PR #2995, opened 2021-03-05
Remove ratioGradAsync

ye-luo:
## Proposed changes
1. Use taskloop instead of task+taskwait. Keep the code simple
2. Write canonical loop in TWF. timer index is computed in-place instead of using the loop.

- No

Bora, Epyc-server

## Checklist


prckent (approved):
Fairly quick read to get in by the weekend, but LGTM. Subject to CI of course.

Q. How is the taskloop performing with llvmdev?

ye-luo:
> Fairly quick read to get in by the weekend, but LGTM. Subject to CI of course.
> 
> Q. How is the taskloop performing with llvmdev?

taskloop is a well supported CPU task feature. Performance wise it is doing its job. More async needs us to enable. It won't change much the performance by itself.
It looks a syntactic sugar but works very well with loop scopes. The previous issue of privatizing complex data structure in "omp task" can be perfectly bypassed.

PR #3000, opened 2021-03-08
Expand Intel-GNU C++ library version error messages

prckent:
## Proposed changes

More informative error messages for Intel-GNU C++ library version mismatches. Driven by recent user experience.

## What type(s) of changes does this code introduce?

- Build related changes
- Documentation changes

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?

None (no Intel compiler to hand. Requires CI).

## Checklist

- Yes. This PR is up to date with current the current state of 'develop'
- No. Code added or changed in the PR has been clang-formatted
- NA. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- NA. Documentation has been added (if appropriate)

PR #3001, opened 2021-03-08
Fix cmake warning by moving check code

ye-luo:
## Proposed changes
#3000 has a CMake warning about quote marks.
So just move the try_compile source code outside CMake.
This avoids writing a file at every cmake invocation and the source code is more readable as a regular cpp code.

- Bugfix
- Build related changes

- No

epyc-server

## Checklist


prckent (approved):
Sorry for the quotes. Works locally.

PR #3007, opened 2021-03-10
Guard determinisitc MSD tests in offload build

ye-luo:
## Proposed changes
Tests fail due to the default code path chosen differently in offload build.
1. introduce SUCCESS_STATUS_OFFLOAD
2. rename MP_SUCCESS to SUCCESS_STATUS_MP

- No

Ryzen-box

## Checklist


prckent:
I think we should have a broad discussion on how to handle this situation ahead of possible merge since it will reoccur.

ye-luo:
> I think we should have a broad discussion on how to handle this situation ahead of possible merge since it will reoccur.

It think It is more related to define what a test pass means. We probably discussed in the past but didn't settle it. This time, we should settle it and put it in the dev manual.

Regarding this PR. Since this makes nightly deterministic tests green, I would prefer merging it as it is and fix up things after unless you have other specif concerns.

prckent (approved):
In a subsequent PR please add a comment around lines 28 and 34 of tests/CMakeLists.txt defining SUCCESS_STATUS_MP and SUCCESS_STATUS_OFFLINE

(Not requiring now in case the CI isn't fast enough ahead of tonights tests)

PR #3012, opened 2021-03-12
Move recompute out of walker control and add TWF::mw_recompute

ye-luo:
## Proposed changes
In DMCBatch, a copied walker or transferred walker needs some recomputing. This was done previously in WalkerControl but it has too many side effects.
For this reason, those operations are moved to the beginning of a DMC p-by-p move section.

I went through the recompute() routine and added mw_recompute().
evaluateLog = recompute + evaluateGL.
This decomposition serves well with minimal duplicated code.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/ParticleSet.cpp
+++ b/src/Particle/ParticleSet.cpp
@@ -813,9 +813,15 @@ void ParticleSet::mw_loadWalker(const RefVectorWithLeader<ParticleSet>& psets,
                                 const RefVector<Walker_t>& walkers,
                                 bool pbyp)
 {
+  auto loadWalkerConfig = [](ParticleSet& pset, Walker_t& awalker)
+  {
+    pset.R     = awalker.R;
+    pset.spins = awalker.spins;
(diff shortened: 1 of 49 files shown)


PDoakORNL:
Why not just pass RefVectors with the walker elements to be recomputed?  That way the exceptional code stays here and the mw api stays clean.

ye-luo:
> Why not just pass RefVectors with the walker elements to be recomputed? That way the exceptional code stays here and the mw api stays clean.

`vector<bool> recompute` is a hint, the actually implementation may decide to recompute everything.
I'm considering to move certain per walker memory to shared buffer and thus recompute has to be done on all the walkers.

PDoakORNL:
Right as long as the underlying implementations are likely to be running on an accelerator with a transfer bottleneck we're not going to want to be changing the walker elements batched over within certain scopes. If we do we're going to have to provide a bunch of indexing information which will really complicate mw code.

PDoakORNL (approved):
We'll want to document this carefully once the dust settles.

PR #3013, opened 2021-03-12
Not print accelerator in QMC_CUDA build

### LINKED ISSUE #3008: Legacy CUDA builds incorrectly report accelerators=0
**Describe the bug**

Number of accelerators/GPUs is wrongly reported. Based on some spot checks of the results on cdash, this seems to be restricted to legacy CUDA builds. Offload runs report correctly.

e.g.

```
 Built without MPI. Running in serial or with OMP threads.

  Total number of MPI ranks = 1
  Number of MPI groups      = 1
  MPI group ID              = 0
  Number of ranks in group  = 1
  MPI ranks per node        = 1
  Accelerators per node     = 0
  OMP 1st level threads     = 16
  OMP nested threading disabled or only 1 thread on the 2nd level

  Base precision      = double
  Full precision      = double
  CUDA base precision = double
  CUDA full precision = double

  CUDA acceleration build option is enabled
  Timer build option is enabled. Current timer level is coarse
```

**To Reproduce**

Latest develop, legacy CUDA build.

**Expected behavior**

At least 1 accelerator should be reported.

**System:**

nitrogen, sulfur. Nightly test systems.

ye-luo:
## Proposed changes
In QMC_CUDA case, the GPU to MPI rank assignment is handled differently from all the rest cases. So Just don't print in this place. The same info has been printed before this line.

```
Rank =    0  Free Memory = 268400 MB
1 MPI ranks on node(s) sulfur.ornl.gov with 1 appropriate CUDA devices.
<- Rank 0 will use CUDA device #0 (Tesla V100-PCIE-16GB)
Rank 0: relative rank number = 0, number of devices = 1
Assigned device numbers: 0
Default MAX_GPU_SPLINE_SIZE_MB is 81920 MB.
  Input file(s): det_qmc_short_vmc_dmc.in.xml 

                    QMCPACK 3.10.9
```

Closes #3008

- Bugfix

- No

Bora

## Checklist

### DIFF
--- a/src/QMCApp/QMCMain.cpp
+++ b/src/QMCApp/QMCMain.cpp
@@ -86,7 +86,7 @@ QMCMain::QMCMain(Communicate* c)
       << "\n  MPI group ID              = " << myComm->getGroupID()
       << "\n  Number of ranks in group  = " << myComm->size()
       << "\n  MPI ranks per node        = " << NodeComm.size()
-#if defined(ENABLE_OFFLOAD) || defined(ENABLE_CUDA) || defined(QMC_CUDA) || defined(ENABLE_ROCM)
+#if defined(ENABLE_OFFLOAD) || defined(ENABLE_CUDA) || defined(ENABLE_ROCM)
       << "\n  Accelerators per node     = " << num_accelerators
 #endif
       << std::endl;


prckent (approved):
:-)

PR #3015, opened 2021-03-14
Change ParticleSet resource management API

ye-luo:
## Proposed changes
multi walker shared resources are assigned not to the leader of a crowd but a crowd with a leader. Memory views are set up and clean up when such assignment/release happens. This allows the distance table offload better optimized. 

- New feature

- No

epyc-server, summit

## Checklist

### DIFF
--- a/src/Particle/SoaDistanceTableABOMPTarget.h
+++ b/src/Particle/SoaDistanceTableABOMPTarget.h
@@ -225,9 +225,6 @@ public:
     }
   }
 
-  /** It has two implementation mw_evaluate_transfer_inplace and mw_evaluate_fuse_transfer with different D2H memory transfer schemes.
-   * Eventually, there will be only one version wihtout any transfer and solve the dilemma.
-   */
   inline void mw_evaluate(const RefVectorWithLeader<DistanceTableData>& dt_list,
                           const RefVectorWithLeader<ParticleSet>& p_list) const override
   {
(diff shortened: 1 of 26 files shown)


src/QMCHamiltonians/NonLocalECPComponent.h
```diff
@@ -190,6 +179,7 @@ class NonLocalECPComponent : public QMCTraits
                              const RefVectorWithLeader<TrialWaveFunction>& psi_list,
                              const RefVector<const NLPPJob<RealType>>& joblist,
                              std::vector<RealType>& pairpots,
+                                          ResourceCollection& collection,
```
PDoakORNL, line 182: clang-format

src/Utilities/ResourceCollection.h
```diff
@@ -76,5 +77,41 @@ class ResourceCollectionLock
   const bool active;
 };
 
+
```
PDoakORNL, line 80: Some documentation of why ResourceCollectionTeamLock is useful is needed.

At first glance it seems 
```c++
{
ResourceCollectionLock(res_coll, batched_things.get_leader())
callNeedingResource(batched_things);
}
```
Would do that same thing.

But it seems like what this really does is give the leader's acquire_resource call access to the whole refvector and not just the leader. 

I can't judge yet whether adjusting/checking the resource state just in time in a particular call or a mechanism like this is better, I'm guessing you added this because at the moment both are needed to refactor existing code. Perhaps we will want to prefer one for most resource consumers.

PDoakORNL (changes requested):
Of main concern is the possible bug with the distance table resource association cycle.

src/Particle/SoaDistanceTableABOMPTarget.h
```diff
@@ -111,23 +147,36 @@ class SoaDistanceTableABOMPTarget : public DTD_BConds<T, D, SC>, public Distance
...
 
-  void releaseResource(ResourceCollection& collection) override { collection.takebackResource(std::move(mw_mem_)); }
+  void releaseResource(ResourceCollection& collection,
+                       const RefVectorWithLeader<DistanceTableData>& dt_list) const override
+  {
+    collection.takebackResource(std::move(dt_list.getCastedLeader<SoaDistanceTableABOMPTarget>().mw_mem_));
+    for (size_t iw = 0; iw < dt_list.size(); iw++)
+    {
+      auto& dt = dt_list.getCastedElement<SoaDistanceTableABOMPTarget>(iw);
+      dt.distances_.clear();
```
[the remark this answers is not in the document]
ye-luo: distances_ is std::vector<OhmmsVector>. All the OhmmsVector are gone with clear of the vector. I was not calling clear on each OhmmsVector element.

PDoakORNL:
Since this passes the CI and the remaining issues are formatting and documentation I think we could move on to the next PR and address those later.  @ye-luo would speedier merges make a difference for you?

PDoakORNL (approved):

ye-luo:
I was about to say I will fix documentation and format in #3016

src/Utilities/ResourceCollection.h
ye-luo: The current code uses ResourceCollectionTeamLock for ParticleSet for my exploration but the reset has not been changed yet. It is a bit restrictive and needs acuqire/release APIs to be conformal. But maybe this restrictive way is better to discipline the implementation more.

PR #3016, opened 2021-03-14
Adjust recompute at the begining of a DMC step

ye-luo:
Review after #3015

## Proposed changes
1. Add recompute mask to loadWalker to reduce recompute impact.
2. Take advantage of mask in mw_recompute in DiracDeterminantBatch
3. turn on distance table offload by default.

- New feature

- No

epyc-server, Summit

## Checklist

### DIFF
--- a/src/Particle/DistanceTableData.h
+++ b/src/Particle/DistanceTableData.h
@@ -178,6 +178,21 @@ public:
       dt_list[iw].evaluate(p_list[iw]);
   }
 
+  /** recompute multi walker internal data, recompute
+   * @param dt_list the distance table batch
+   * @param p_list the target particle set batch
+   * @param recompute if true, must recompute. Otherwise, implementation dependent.
+   */
+  virtual void mw_recompute(const RefVectorWithLeader<DistanceTableData>& dt_list,
+                            const RefVectorWithLeader<ParticleSet>& p_list,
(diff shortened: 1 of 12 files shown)


prckent:
If you consider this version optimized enough, do you have any comments on when/where it makes sense to have DT offload? Are you seeing benefits at all system sizes?

ye-luo:
The default setting put gpu="yes" on all the particleset.
In the NiO benchmark running on Summit, I saw a32 ~20%, a64 ~30% and a128 100% speed-up at the throughput saturation area in the walker scan plots. So I consider this good to be on by default.

The optimal performance of the current code should be gpu="no" on the ion particleset and gpu="yes" on the electron particleset. but the impact is only 2-5% on all the problem sizes. So I'm not bothered by such fine tuning. As we move Jastrow factors to GPU, "yes" will be better for ion particleset as well.

ye-luo:
Test this please

prckent (approved):
LGTM ( I am biased here since the merge will be helpful for current runs. )

I suggest we consider expanding our CI a little since logic errors in this part of the code might not have good coverage in the CI. More of the deterministic integration tests would need to run.

Comments from #3015 have been addressed. If more updates are needed (@PDoakORNL ?) a subsequent PR can adjust.

PR #3022, opened 2021-03-18
Fix SlaterDetWithBackflow virtual function override mismatch

ye-luo:
## Proposed changes
Fix a breakage introduced in #3012
SlaterDetWithBackflow::evaluateLog has been corrected. All other changes are just adding override keyword to prevent future mistakes.

- Bugfix

- No

Bora

## Checklist


prckent (approved):
Presumably if we were override clean and had warnings/errors on, this problem could not have occurred...

PR #3023, opened 2021-03-19
AFQMC MULTI Update

correaa:
## Proposed changes

Updates multi with modifications to make it work with AFQMC

- No

## Checklist


ye-luo:
Okay to test

ye-luo:
@correaa could you fix the gnu warning as the warning-as-error build reports?

prckent:
@correaa We need the gnu warning-clean version to merge. Unless there is some deep reason it can't be avoided, we don't merge code that increases gnu errors. Over time the exception list is being decreased. Is it viable to get rid of the current gnu warnings?

correaa:
> @correaa We need the gnu warning-clean version to merge. Unless there is some deep reason it can't be avoided, we don't merge code that increases gnu errors. Over time the exception list is being decreased. Is it viable to get rid of the current gnu warnings?

yes, the last commits are for that. The gnu warnings were reasonable and fixable.

I didn't know you were suppressing some warning-as-errors already.
If that is the case you can now get rid of the warning-error:

```
qmcpack/src/AFQMC/Wavefunctions/WavefunctionFactory.cpp:1236:1: error: control reaches end of non-void function [-Werror=return-type]
```
which was there before I started this PR.

> Edit: I assume that once this PR is ready we can merge and then update the ppconvert one which will then also pass.

yes, that is what I wanted to mean in my last post.

ye-luo (approved):

PR #3028, opened 2021-03-19
Expand runtime control: global max_seconds and STOP file

### LINKED ISSUE #498: Stop the code on the fly
Sometimes, I feel the need to stop the code execution on the fly and exit clean. Is this useful?
Basically the code monitors a file at the end of each block.

ye-luo:
## Proposed changes
This PR enhances runtime controls.
1. `<project id="project-prefix" max_seconds="500">` to control time limit in batched drivers. default 360000 (100h) as legacy drivers which however still respect `max_seconds` in the per driver input section.
2. project-prefix.STOP can be used to stop a run sanely. With an ensemble run, all jobs ends. close #498
3. Fix this case. Once a driver section hits the time limit, all the subsequent driver sections won't run.

check frequency is per block.

explicitly tested ensemble job case. existing cpu_limit test results.
```
$ ctest -R cpu -j32 -E performance
Test project /home/yeluo/opt/qmcpack/build_clang_real_debug
    Start 893: cpu_limit-diamondC_1x1x1_pp-dmc-1-1
    Start 779: cpu_limit-bccH_1x1x1_ae-rmc-1-1
    Start 892: cpu_limit-diamondC_1x1x1_pp-vmc-4-4
    Start 891: cpu_limit-diamondC_1x1x1_pp-vmc-1-1
    Start   6: deterministic-unit_test_cpu
1/6 Test   #6: deterministic-unit_test_cpu ...........   Passed    0.38 sec
2/6 Test #891: cpu_limit-diamondC_1x1x1_pp-vmc-1-1 ...   Passed   49.85 sec
3/6 Test #892: cpu_limit-diamondC_1x1x1_pp-vmc-4-4 ...   Passed   50.08 sec
    Start 894: cpu_limit-diamondC_1x1x1_pp-dmc-4-4
4/6 Test #779: cpu_limit-bccH_1x1x1_ae-rmc-1-1 .......   Passed   50.33 sec
5/6 Test #893: cpu_limit-diamondC_1x1x1_pp-dmc-1-1 ...   Passed   90.87 sec
6/6 Test #894: cpu_limit-diamondC_1x1x1_pp-dmc-4-4 ...   Passed   47.34 sec
```

- New feature

- No

epyc-server

## Checklist


jtkrogel:
If `<project>` is going to expand into our "control" section, please can we use parameters rather than attributes?

I would really like to avoid cases like this if at all possible in the future:
```
<sposet_builder type="bspline" href="../nscf/pwscf_output/pwscf.pwscf.h5" tilematrix="1 1 -1 1 -1 1 -1 1 1" twistnum="0" source="ion0" version="0.10" meshfactor="1.0" precision="float" truncate="no">
```

ye-luo:
> If `<project>` is going to expand into our "control" section, please can we use parameters rather than attributes?
> 
> I would really like to avoid cases like this if at all possible in the future:
> 
> ```
> <sposet_builder type="bspline" href="../nscf/pwscf_output/pwscf.pwscf.h5" tilematrix="1 1 -1 1 -1 1 -1 1 1" twistnum="0" source="ion0" version="0.10" meshfactor="1.0" precision="float" truncate="no">
> ```

Agreed. I was expecting this request. Documentation is still missing in this PR. Putting out this PR for feedback.

jtkrogel:
Overall these look like good changes to me.

prckent:
I am against max_seconds="500" as an attribute the project section. It adds parsing complication as Jaron noted.
Ideally it should be at the same level as (say) the random number initialization for consistency. Maybe it has to be a project attribute, which then raises the question about whether the random initialization (seed) is really in the "correct" and consistent place.

Generally I am against the use of attributes when we could use parameters. This is much easier for users (including this one) to remember, read, and get right. Use of XML for input is unfortunate, but it is what we have currently.

We already have maxcpusecs that works at the driver level, and should recycle the same tag at a higher level.

It would also make sense to support the same parameter as a command line argument. Then when the use or robot writes a job submission file it can be written there without affecting the QMC input.

.STOP will be very useful!

I agree with the per block check, since part of our current block definition is that this is when we do I/O.

prckent:
@ye-luo Interested to know your current use case, or perhaps it was requested.

ye-luo:
I changed "max_seconds" from attribute to parameter but it remains under project.
"maxcpusecs" is an unfortunate choice. The name is better to be simple. I have an idea, since I have to tool now. I can mark "maxcpusecs" deprecated and add "max_seconds" in classic drivers.

Regarding command line arguments, I would like to move more command line arguments under project to improve reproducibility from a single input file. For example, we may consider moving verbosity.

I want to see the timer so the execution must stop clean. If I'm looking at the timer distribution and see which are heavier, just need to stop the code on the fly without changing input.

ye-luo:
@prckent let me know if you really want to stick to "maxcpusecs", otherwise, I will make it deprecated.

prckent:
max_seconds is good imo.

OK to mark the old tag as deprecated and abort - there can't be many people using it and it is an easy change to make. An edit might needed in NEXUS.

ye-luo:
"deprecated" will continue to run. Once 1 release is out, we change it to "deleted" then users get an abort.

ye-luo:
maxcpusecs has been superseded by max_seconds in classical drivers.
found no trace of maxcpusecs in nexus folder

src/Utilities/RunTimeManager.h
```diff
@@ -66,26 +65,47 @@ extern template class LoopTimer<FakeCPUClock>;
...
-    m_loop_margin            = 1.1;  // 10% margin on average loop time?
-  }
+    CONTINUE,           // keep running
+    MAX_SECONDS_PASSED, // all already passed max_seconds
+    NOT_ENOUGH_TIME,    // not enough time for next iteration
+    STOP_FILE,          // reqsuted stop from a file
+  } stop_status_;
 
+  bool enough_time_for_next_iteration(LoopTimer<CLOCK>& loop_timer);
+  bool stop_file_reqeusted();
```
markdewing, line 87: Spelling: stop_file_requested

src/Utilities/RunTimeManager.cpp
```diff
@@ -26,39 +27,127 @@ template class RunTimeManager<CPUClock>;
...
-  log << "  Iteration time per " << driverName << " block (seconds) = " << m_loop_time << std::endl;
-  log << "  Elapsed time (seconds)       = " << m_elapsed << std::endl;
-  log << "  Remaining time (seconds)      = " << m_remaining << std::endl;
+  log << "RunTimeControl in " << driverName << " kicks in." << std::endl;
+  if (stop_status_ == StopStatus::MAX_SECONDS_PASSED)
+    log << "Time limit reached. Stopping after block " << block << std::endl
+        << "Hard limit (seconds) " << MaxCPUSecs << ", elapsed (seconds) " << m_elapsed << std::endl;
+  else if (stop_status_ == StopStatus::NOT_ENOUGH_TIME)
+  {
+    log << "Insufficeint time for next block. Stopping after block " << block << std::endl;
```
markdewing, line 140: Spelling: Insufficient

src/Utilities/RunTimeManager.cpp
```diff
@@ -26,39 +27,127 @@ template class RunTimeManager<CPUClock>;
...
+
+template<class CLOCK>
+std::string RunTimeControl<CLOCK>::generateStopMessage(const std::string& driverName, int block) const
 {
   std::stringstream log;
-  log << "Time limit reached for " << driverName << ", stopping after block " << block - 1 << std::endl;
-  log << "  Iteration time per " << driverName << " block (seconds) = " << m_loop_time << std::endl;
-  log << "  Elapsed time (seconds)       = " << m_elapsed << std::endl;
-  log << "  Remaining time (seconds)      = " << m_remaining << std::endl;
+  log << "RunTimeControl in " << driverName << " kicks in." << std::endl;
```
markdewing, line 134: Message should be more formal than "kicks in".
Maybe "RunTimeControl activated in <drivername>"

src/Utilities/RunTimeManager.h
ye-luo: Fixed.

src/Utilities/RunTimeManager.cpp
ye-luo: Fixed

src/Utilities/RunTimeManager.cpp
ye-luo: Changed to `"RunTimeControl takes action in " << driverName << " driver."`

prckent (approved):
Verified by hand.

Added another mention in the documentation and increased the safety factor.

PR #3040, opened 2021-03-23
Fix index problem with 2-body Jastrow optimization

markdewing:
Addresses #2814 (at least partially)

Add a unit test that will fail in one case where the offsets are not computed correctly.

Fix the offset calculation in the case the first function does not have any optimizable parameters.

- Bugfix

- No

laptop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
+++ b/src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
@@ -64,6 +64,12 @@ public:
     delete_iter(lapLogPsi.begin(), lapLogPsi.end());
   }
 
+  // Accessors for unit testing
+  std::pair<int, int> getOffset(int index) { return OffSet.at(index); }
+
+  opt_variables_type& getVars() { return myVars; }
+
+
   void addFunc(int ia, int ib, FT* j)
   {
     // make all pair terms equal to uu initially
@@ -139,7 +145,18 @@ public:
         lapLogPsi[i]  = new ValueVectorType(NumPtcls);
       }
       OffSet.resize(F.size());
-      int varoffset = myVars.Index[0];
+
+      // Find first active variable for the starting offset
+      int varoffset = -1;
+      for (int i = 0; i < myVars.size(); i++)
+      {
+        varoffset = myVars.Index[i];
+        if (varoffset != -1)
+        {
+          break;
+        }
+      }
+
       for (int i = 0; i < F.size(); ++i)
       {
         if (F[i] && F[i]->myVars.Index.size())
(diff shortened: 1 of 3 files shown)


src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
```diff
@@ -139,7 +145,18 @@ class DiffTwoBodyJastrowOrbital : public DiffWaveFunctionComponent
...
+      int varoffset = -1;
+      for (int i = 0; i < myVars.size(); i++)
+      {
+        varoffset = myVars.Index[i];
+        if (varoffset != -1)
+        {
+          break;
+        }
+      }
+
```
ye-luo, line 159: If we have 3 distinct sets of coefficients (uu/ud/dd) and only the middle one is optimization. The code is still potentially wrong? Let me know your answer or your guess.

src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
```diff
@@ -139,7 +145,18 @@ class DiffTwoBodyJastrowOrbital : public DiffWaveFunctionComponent
...
       OffSet.resize(F.size());
-      int varoffset = myVars.Index[0];
+
+      // Find first active variable for the starting offset
+      int varoffset = -1;
+      for (int i = 0; i < myVars.size(); i++)
+      {
+        varoffset = myVars.Index[i];
+        if (varoffset != -1)
+        {
```
ye-luo, line 155: Please remove {} around break to save two lines.

src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
markdewing: I modified the unit test for this case, and it still passed.

src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
markdewing: fixed

src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
ye-luo: The unit test is only limited to check indices. So it is guaranteed up-to what is being checked. Not necessary a real J2 optimization will work. right?

markdewing: Yes, that's correct - this only fixes setting OffSet variable.  Shiv's PR had another change that I need to look at.

ye-luo: I see thanks.

prckent (approved):
LGTM

I changed the title since it wouldn't surprise me if this code pattern is repeated in other wavefunction components.

PR #3043, opened 2021-03-24
Fix indexing in diff Jastrow component

markdewing:
Second part of fix for #2814

The myVars in this object contains all the variational parameters aggregated from all the radial functions (in their myVars), including the non-active ones.

The OffSet variable maps the indices from the radial function myVars to myVars in this object.

The output for parameter derivatives should use a contiguous index (i.e. all the non-active parameters removed).  This is what the Index value in VariableSet maps (returned by myVars.where). In the code, index k covers all variables in myVars.  The index kk is the contiguous index for output.

Some of the arrays are indexed with k, but should be indexed with kk instead.

With this fix, the computation of gradients appears to be correct for a few examples (looking at the analytic vs. numerical output).

- Bugfix

- No

desktop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
+++ b/src/QMCWaveFunctions/Jastrow/DiffTwoBodyJastrowOrbital.h
@@ -187,7 +187,7 @@ public:
         continue;
       if (active.recompute(kk))
         recalculate = true;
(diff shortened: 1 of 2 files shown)


prckent:
@markdewing Are you able to reproduce this?

markdewing:
Closing because I have different approach to the fix.

Issue #3020, opened 2021-03-17
Adding GitHub Actions


williamfgc:
As brought up by @prckent . This is issue is regarding on exploring the addition of GitHub Actions to QMCPACK to current CI checks. The goal is to open discussion on the community on the scope of current and future quality checks (e.g. sanitizers, code coverage, etc.) as we go on.

williamfgc:
The docker container with dependencies is located in my local account [here](https://hub.docker.com/repository/docker/williamfgc/qmcpack-ci). After discussion, we can move it to a more project account on DockerHub.

prckent:
I think the key thing is to get the initial action up and running so that we can study the resource usage, learn more about github actions, and gradually expand add capabilities etc.

williamfgc:
@prckent that's correct, thanks!  The first workflow is ready to go we just need to allow external actions (as in one-line workflows like checking out code, perform clang-format checks, etc.), see [here](https://github.com/QMCPACK/qmcpack/actions/runs/666169074). In order to get external actions (either official actions or from marketplace) working we can do `Settings>Actions` and either: 
1. `Allow all actions`
2.`Allow selected actions` and add `actions/checkout@v1`.  

Essentially, we either allow all actions or cherry-pick the ones we need as we go. I usually go with the former to set and forget, thoughts?

williamfgc:
The first GitHub Actions workflow is up, see [here](https://github.com/QMCPACK/qmcpack/runs/2149217063?check_suite_focus=true). QMCPACK takes 25 minutes to build and deterministic tests took 3.5 min, here is the [Linux hardware](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). @prckent we can follow up on the next steps (more compiler combinations, tests, sanitizers, nightly, etc.), thanks.

prckent:
These timings seem about right for a build on one or few threads. Can the build be sped up? On a strong workstation  (make -j 32, flash filesystem etc.) the build time is a 2-3 minutes. Perhaps we only get one or a few threads, but ctest -j options might be adjustable. Our tests are generally built for up to 16 threads / cores total.

williamfgc:
I think we can close this issue.

PR #3044, opened 2021-03-24
GitHub Actions Sanitizers Checks: ubsan

williamfgc:

## Proposed changes

Enable sanitizer checks using clang compilers as a GitHub Actions matrix for: asan, ubsan, tsan and msan for deterministic tests. The goal is to catch more memory-related bugs and guide fixes.

- Testing changes (e.g. new unit/integration/performance tests)  sanitizers checks

- No

Local Ubuntu 20.04, the goal is to test it on GitHub Actions runners

## Checklist
- No. Code added or changed in the PR has been clang-formatted (NA)

Related to #3020


prckent:
William - I see that these are running so the answer to the following question will soon be clear, but I was wondering if you have been able to do a test run, and if so, what do the results look like in your opinion? i.e. Do you think that they are actionable as is, need to be redirected to another service, need some whitelisting to be useful/useable?

CMakeLists.txt
```diff
@@ -267,6 +267,23 @@ CMAKE_DEPENDENT_OPTION(USE_VTUNE_TASKS "USE VTune ittnotify task annotation" OFF
 # CMake note - complex conditionals in cmake_dependent_option must have spaces around parentheses
 CMAKE_DEPENDENT_OPTION(USE_NVTX_API "Enable/disable NVTX regions in CUDA code." OFF "ENABLE_TIMERS AND ( QMC_CUDA OR ENABLE_CUDA )" OFF)
 
+######################################################################
+# Sanitizer options
+######################################################################
+
+# Add optional sanitizers ASAN, UBSAN, MSAN
+SET(ENABLE_SANITIZER "NONE" CACHE STRING "NONE,ASAN,UBSAN,MSAN")
+SET_PROPERTY(CACHE ENABLE_SANITIZER PROPERTY STRINGS "NONE" "ASAN" "UBSAN" "MSAN")
```
ye-luo, line 276: Is it better to have "none" "asan" "ubsan" "msan" all in lower case?

CMakeLists.txt
```diff
@@ -267,6 +267,23 @@ CMAKE_DEPENDENT_OPTION(USE_VTUNE_TASKS "USE VTune ittnotify task annotation" OFF
...
+
+# Perform sanitizer option check, only works in debug mode
+IF( NOT "${ENABLE_SANITIZER}" STREQUAL "NONE")
+  MESSAGE( STATUS "Enable sanitizer ENABLE_SANITIZER=${ENABLE_SANITIZER}" )
+  IF( "${ENABLE_SANITIZER}" STREQUAL "ASAN" AND NOT "${CMAKE_BUILD_TYPE}" STREQUAL "DEBUG" )
+    # better to fail than override debug mode
+    MESSAGE( FATAL_ERROR "-DENABLE_SANITIZER=ASAN can only be enabled with -DCMAKE_BUILD_TYPE=Debug" )
```
ye-luo, line 283: To my understanding, this is not required. Release, Debug, RelWithDebugInfo all work but the level of info is different. RelWithDebugInfo is my recomendataion not Debug.

CMakeLists.txt
ye-luo: If I give a random string ("abc") not listed above, does cmake stop?

CMakeLists.txt
```diff
@@ -865,6 +882,30 @@ IF(ENABLE_GCOV)
...
+        CACHE STRING "UndefinedBehaviorSanitizer C++ compiler builds." FORCE)
+  ELSEIF( "${ENABLE_SANITIZER}" STREQUAL "MSAN" )
+    SET(CMAKE_CXX_FLAGS_SAN "-fsanitize=memory"
+        CACHE STRING "MemorySanitizer C++ compiler builds." FORCE)
+  ELSEIF( "${ENABLE_SANITIZER}" STREQUAL "TSAN" )
+    SET(CMAKE_CXX_FLAGS_SAN "-fsanitize=thread"
+        CACHE STRING "ThreadSanitizer C++ compiler builds." FORCE)
+  ENDIF()
+  
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_SAN}")
```
ye-luo, line 905: Please protect the case of none namely CMAKE_CXX_FLAGS_SAN not set to avoid adding a white space.

ye-luo (commented):
Need to clean up previous sanitizer options in CMake/ClangCompilers.cmake
and update manual entry.

williamfgc:
@prckent good points, I'd merge `ubsan` as-is (good job on that!) as it's low-overhead. `tsan` seems to be just hanging on GitHub Actions. `msan` is catching some stuff from Catch (pun not intended), so I need to set a filter of some sort.The most useful is `asan` as I see something that looks like a "true positive" from using `new` in `/src/Estimators/LocalEnergyEstimator.cpp:38`.

My preference would be to keep all the CMake configuration, and:
- Merge `ubsan`  as-is
- Deactivate `asan` (priority to resolve), `msan`, `tsan` on GitHub Actions until issues are resolved.

The above is to keep CI clean and short. Let me know what you think, then we can merge and start chasing leaks.

CMakeLists.txt
williamfgc: @ye-luo I can change it to lower case if that's the preference. Good catch on the second issue, I can make it error out if the entry is not valid. `SET_PROPERTY` is for `ccmake` or `cmake-gui`.

CMakeLists.txt
williamfgc: Will do.

prckent:
@williamfgc I agree with your recommendations. We put ubsan into production immediately , then work to fix the code for asan to pass and get that in production. asan actually looked quite close. msan is where the big work is imo, but I suspect there are many instances of the same error being repeated over and over.

After a few more PRs have run through I think we can set ubsan and the original linux build action as required to pass.

CMakeLists.txt
williamfgc: Good line, this line should be protected as it's executed inside a `if not equal none` block, but please double-check I'm not missing anything else. Thanks!

prckent:
OK to test

prckent (approved):
LGTM

Ye to merge or request further changes

ye-luo (approved):
LGTM

ye-luo:
@williamfgc could you relocate config/sanitizers/lsan.supp under tests/test_automation/github-actions/ci?

williamfgc:
@ye-luo just double-checking 'cause the file is not exclusive to github-actions/ci, is this still preferred?

prckent:
If my understanding is correct, I think it is not plumbed in to regularly sanitizer builds yet, but might be in future. In which case leave where it is. (We would need to configure ASAN_OPTIONS=suppression=${GITHUB_WORKSPACE}/config/sanitizers/lsan.supp ?)

williamfgc:
@prckent yes you're correct, it's not currently part of the GitHub Actions runners until leaks are resolved. It's needed locally when debugging leaks to avoid false positives. Yes, `LSAN_OPTIONS` has to be configured to this file (the same with `ASAN_OPTIONS` if we need one), see [here](https://github.com/QMCPACK/qmcpack/blob/develop/tests/test_automation/github-actions/ci/run_step.sh#L80). Thanks!

ye-luo:
As long as it is primarily used/maintained for github-actions/ci, it is better to keep this file with other files for the same purpose. I'm trying to keep `/config` folder for build recipes instead of any kind of configuration files. No need to move the file right now, just keep in mind and do it when other changes are needed for github actions.

PR #3048, opened 2021-03-24
[WIP] Batch m inverse

PDoakORNL:
## Proposed changes

batched full inversion and calculation of the log determinant.

There is a bit of a state machine mine field here and existing tests are not really unit tests but long tests of the complete classes that obscure the state sequence required to make a valid call to a particular method.

I've added some simpler unit tests focused on single methods to track down issues I've discovered or created at each layer I've climbed up toward the full application.

Currently use of this at the application level is blocked by batched grad evaluation producing zero grads.  This is reproduced in a minimal unit test in test_wavefunction_trialwf.
The exiting unit tests in test_wavefunction_trialwf do not reveal this defect.

- New feature

- Yes
- No

## Checklist


src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
```diff
@@ -374,67 +374,86 @@ class MatrixDelayedUpdateCUDA
...
-  inline OffloadPinnedValueMatrix_t& get_psiMinv() { return psiMinv; }
+  /** Why do you need to modify another classes data member?
+   */
+  inline const OffloadPinnedValueMatrix_t& get_psiMinv() const { return psiMinv; }
+
+  inline OffloadPinnedValueMatrix_t& get_nonconst_psiMinv() { return psiMinv; }
 
   inline T* getRow_psiMinv_offload(int row_id) { return psiMinv.device_data() + row_id * psiMinv.cols(); }
 
+  /** make this class unit tests friendly without the need of setup resources.
```
PDoakORNL, line 389: This is going to be deleted.  It did not turn out to be a good idea.

PDoakORNL (commented):
Self review part 1

ye-luo:
Could you break this PR in to several steps? A bit hard to follow. If many intermediate commits cannot build, better to fuse certain commits.

src/Platforms/CPU/SIMD/inner_product.hpp
```diff
@@ -211,6 +211,25 @@ inline void transpose(const T* restrict A, size_t m, size_t lda, TO* restrict B,
...
+     * @param A starting address, A(m,lda)
+     * @param m number of A rows
+     * @param lda stride of A's row
+     * @param B starting address B(m,ldb)
+     * @param n number of B rows
+     * @param ldb stride of B's row
+     *
+     * Blas-like interface
+     */
+template<typename T, typename TO>
```
PDoakORNL, line 224: I can't find a way around needing this when a Matrix that has the bogus extra columns added needs to appear as a Matrix where the cols and rows are assumed to be useful information.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
```diff
@@ -22,8 +23,8 @@ namespace qmcplusplus
  *@param spos the single-particle orbital set
  *@param first index of the first particle
  */
-template<typename DET_ENGINE_TYPE>
-DiracDeterminantBatched<DET_ENGINE_TYPE>::DiracDeterminantBatched(std::shared_ptr<SPOSet>&& spos, int first)
+template<typename DET_ENGINE>
```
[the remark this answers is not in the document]
ye-luo: Agreed but better to change this on develop directly instead of this PR

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
```diff
@@ -46,38 +47,129 @@ void DiracDeterminantBatched<DET_ENGINE_TYPE>::set(int first, int nel, int delay
...
+  for (int iw = 0; iw < nw; ++iw)
+  {
+    auto& det            = wfc_list.getCastedElement<DiracDeterminantBatched<DET_ENGINE>>(iw);
+    auto& engine_psiMinv = det.det_engine_.get_nonconst_psiMinv();
+    det.dummy_vmt.attachReference(engine_psiMinv.data(), engine_psiMinv.rows(), engine_psiMinv.cols());
+  }
+#endif
+}
+
+template<>
```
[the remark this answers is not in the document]
ye-luo: Not getting this point. Dropping the filter needs to be consist with the change how psiMinv is stored. So it is a decision inside DET_ENGINE, I don't think adding DiracDeterminantDetails is necessary.

src/QMCWaveFunctions/Fermion/DiracMatrixComputeOMPTarget.hpp
```diff
@@ -0,0 +1,310 @@
...
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2021 QMCPACK developers.
+//
+// File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Lab
+//
+// File created by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Lab
+//////////////////////////////////////////////////////////////////////////////////////
+
+#ifndef QMCPLUSPLUS_DIRAC_MATRIX_COMPUTE_OMPTARGET_H
```
ye-luo, line 12: This is largely a copy of DiracMatrix. Is this really needed? It was only a few lines of code in the MatrixUpdateOMPTarget.h

src/QMCWaveFunctions/WaveFunctionComponent.cpp
```diff
@@ -28,7 +28,7 @@ WaveFunctionComponent::WaveFunctionComponent(const std::string& class_name, cons
       Optimizable(true),
       is_fermionic(false),
       UpdateMode(ORB_WALKER),
-      LogValue(0.0),
+      LogValue(0.0),  // std::complex constructor defaults to 0.0 for real and imaginary so setting real to 0.0 is weird.
```
ye-luo, line 31: It is my oversight when changing LogValue to complex type

PDoakORNL:
I'll see what I can do.  The original code is certainly not easy to follow either. 
I think a series of PR's for the new code a build up from the CUDA to DiracDeterminantBatched is doable.

I would like to establish this passing CI so I can track breaking changes to the "working" implementation being introduced to develop.

src/QMCWaveFunctions/Fermion/DiracMatrixComputeOMPTarget.hpp
PDoakORNL: Needs to be determined, but I use DiracMatrixComputeXXX as a compute resource not as a "object" tat both computes and holds one walkers worth of data so that's a big issue.

ye-luo: DiracMatrix is a compute handle and doesn't hold data for one walker. Is it not the case?

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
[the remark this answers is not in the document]
PDoakORNL: I'd already gotten mw_recompute working with the unfiltered lists and wanted move to using only contiguous memory for the batched inversion when the filtered lists being handed to the DET_ENGINE from DiracBatchedDeterminant  was changed in your PR. I wanted to maintain API and walker ref list to memory resource relationship I had been working with.  I'll review this. If we eliminate the DiracDeterminantBase we won't need this either.

src/QMCWaveFunctions/Fermion/DiracMatrixComputeOMPTarget.hpp
PDoakORNL: The linear algebra section doesn't need to be repeated. Longer term delete DiracMatrix and the similarity of code isn't an issue.

src/QMCWaveFunctions/tests/CMakeLists.txt
```diff
@@ -51,8 +52,8 @@ IF (NOT QMC_CUDA)
   ENDIF()
 ENDIF()
 
-SET(TRIALWF_SRC test_wf.cpp test_TrialWaveFunction.cpp test_wavefunction_factory.cpp
-                test_TrialWaveFunction_diamondC_2x1x1.cpp test_TrialWaveFunction_He.cpp
+SET(TRIALWF_SRC test_wf.cpp FakeSPO.cpp test_TrialWaveFunction.cpp test_wavefunction_factory.cpp
```
PDoakORNL, line 55: probably will just merge test_TrialWaveFunction.cpp and test_TrialWaveFunction_diamondC_1x1x1.cpp back together since it turned out to be basically impossible to write decent unit tests without the xml and hdf5 mess.

src/QMCWaveFunctions/tests/SetupDiracDeterminantResources.hpp
```diff
@@ -0,0 +1,53 @@
+//////////////////////////////////////////////////////////////////////////////////////
```
PDoakORNL, line 1: I think this will be dropped as it was supposed to provide generic testing but made bad resource management harder to find.

src/QMCWaveFunctions/Fermion/DiracMatrixComputeOMPTarget.hpp
ye-luo: Since this doesn't bring any thing immediately helpful, I'm in favor of deleting this one and use the old way. Once we have the CUDA conterpart stable, we can come back to review this one and DiracMatrix. Maintaining duplicated code is not an optimal solution.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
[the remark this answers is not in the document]
ye-luo: It is a bit tricky because to use unfiltered list, the SPO evaluation needs to be fast enough, we are not there yet. So it is better to use the filtered list and only compute SPO and inversion for them in order not to hurt to much in performance.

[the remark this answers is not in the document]
prckent: Also we potentially need to support slow-to-evaluate SPO, so filtered list is good.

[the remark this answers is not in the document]
PDoakORNL: SPO evaluation should be very fast on the accelerator. The main issue with SPO evaluation is the data movement. I think on the accelerator leaving the layout of the walkers in device memory the same for an entire block is worth pursuing. If we hand the compute API partial walkers lists but want to use a shared and persistent memory layout this is a mess. 

It would make more sense to pass the execute masks all the way down through the API.  The filtered list is only convenient if you are terminating at single walker computes with per walker pointers to pointers. This only works nicely if you mix up the concept of computation method and data objects as the current DiracMatrix does. Doing this means that hoop jumping to allocated SoA style data will always be convoluted since the code always expects to find data in 1 walker compute objects instead of contiguous vectors for multiple walkers that have straight forward transfer semantics to and from the device and if we need to narrow or widen precision between those buffers on the the device.

It's always going to be much slower to make many calls to the async transfer engine for single walker data elements than for contiguous many walker element arrays. And I think this is frequently the important bottle neck.

I think we need some real benchmarks that reflect the impact of this.

src/QMCWaveFunctions/Fermion/DiracMatrixComputeOMPTarget.hpp
PDoakORNL: It would also mean completely changing the CUDA implementation and returning to DiracMatrix as a data object and not just a compute object that manages the scratch space needed for its batch size of computation.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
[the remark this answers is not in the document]
ye-luo: `DistanceTableData.h` mw_recompute() API carries a full list and masks.
SoaDistanceTableAB.h runs walker by walker. So can benefit from masks.
SoaDistanceTableABOMPTarget.h runs walker batched because a fused memory buffer and thus imply ignores the mask.

PR #3051, opened 2021-03-26
Fix two body Jastrow optimization with inactive variables

markdewing:
Second part of fix for #2814.  See recent comments on that issue for an improved description.

Construct a second mapping for component-level variational parameters to global-level variational parameters.

One potentially signficant change is that this maps the parameters as a block from the component level to the global level.  The previous code could be more fine-grained. If there are cases where the subcomponent variables could be a mix of
active and inactive, then this code will need updating.

- Bugfix

- No

Desktop

## Checklist

- No Documentation has been added (if appropriate)


markdewing:
Discussion on #2814 made me think of a much simpler fix.

PR #3053, opened 2021-03-26
Add manual for vmc_batch

ye-luo:
## Proposed changes
Adding manual for vmc_batch.
Using this PR to discuss styles.
I tried first having a separate section for `vmc_batch` parallel to `vmc`. It caused a lot of duplication and I frequently get lost which part I was reading/editing.
So I incorporate the additional tags of `vmc_batch` directly into the current `vmc` section.

- Documentation changes

- No

ryzen-box

## Checklist


prckent (commented):
Good question. I prefer a dedicated table for the batched drivers. While duplicative of many of the options it makes it even clearer what we are officially supporting and what we encourage use of. I would like to guard against us supporting historical options.

prckent (approved):

prckent (commented):
In the batched driver writeup, I have concerns about storeconfigs: its description is inconsistent with the old driver, and I am not sure of the status. Is it implemented and have we checked? Might be better to list as not ready or developmental.

ye-luo:
What did you mean "inconsistent with the old driver"? I actually don't know the status of this option. It is captured by the input class and shows some use in the batched drivers but I have no clue if it works or not.

prckent (approved):
@ye-luo No problem. I pushed some minor edits.

prckent:
And I broke the tables.

prckent (approved):
Tables are fixed.

prckent:
OK to test

PR #3049, opened 2021-03-26
moving RandomForTest so it can be used by tests outside Estimators/tests

PDoakORNL:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

Moving minimal Random provider for testing. In future PR this will be used for unit tests outside of Estimators.
RandomForTest also templated now.

## What type(s) of changes does this code introduce?

- Refactoring (no functional changes, no api changes)
- Build related changes

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
Leconte

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes This PR is up to date with current the current state of 'develop'
- Yes Code added or changed in the PR has been clang-formatted
- Yes This PR adds tests to cover any new code, or to catch a bug that is being fixed
- Yes Documentation has been added (if appropriate)

PR #3055, opened 2021-03-27
Add checkMatrix for testing

PDoakORNL:
This is intended to reduce repetition of matrix testing functions as well make it easy to do approximate comparison of complex and real values in the same build.  The current mechanism depends on the definition of QMC_COMPLEX. Personally I prefer not to do two builds to check code that must support complex and real arguments.

## Proposed changes

Adds a checkMatrix function to Utilities/for_testing.  This can actually compare both real and complex matrices without changing the QMC_COMPLEX definition.
Test program shows intended usage in tests.

- New feature

- No

## Checklist

- Yes  Code added or changed in the PR has been clang-formatted


PDoakORNL:
This should probably be rebased after #3049 is merged.

src/Utilities/for_testing/CMakeLists.txt
```diff
@@ -8,6 +8,11 @@
...
-  RandomForTest.cpp)
+  RandomForTest.cpp
+  checkMatrix.cpp)
 add_library(utilities_for_test ${test_utilities_src})
-TARGET_INCLUDE_DIRECTORIES(utilities_for_test PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
+target_include_directories(utilities_for_test PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
+target_include_directories(utilities_for_test PRIVATE "${PROJECT_SOURCE_DIR}/external_codes/catch")
+target_link_libraries(utilities_for_test PUBLIC containers)
+
+subdirs(tests)
```
ye-luo, line 18: prefer `Utilities/tests/for_testing` to  `Utilities/for_testing/tests`

src/Utilities/for_testing/checkMatrix.hpp
```diff
@@ -0,0 +1,93 @@
...
+        all_elements_match = false;
+        if (!check_all)
+          return {false, error_msg.str()};
+      }
+    }
+  return {all_elements_match, error_msg.str()};
+}
+
+extern template bool approxEquality<double>(double val_a, double val_b);
+extern template bool approxEquality<std::complex<double>>(std::complex<double> val_a, std::complex<double> val_b);
```
ye-luo, line 89: Need float and complex<float> as well.

src/Utilities/for_testing/CMakeLists.txt
PDoakORNL: That seems like it would double the work of modifying module directories layouts.
But I did notice the extra directory traversals in platform. 
Is the idea to have fewer test program targets?

ye-luo: I'm thinking of have one target for each folder under `src`.
`src/Utilities/test` only needs to depend on qmcutils. So we don't need to care about to many fine-grain CMake targets.
Currently qmcutils is in at "src/CMakeLists.txt" but it should be moved to "src/Utilities/CMakeLists.txt" but this is an on-going cmake refactoring out of the scope of this PR.

PDoakORNL: Ok I moved the tests.  I still prefer Utilities/for_testing as a separate target. They are not routines for use in the application code.

src/Utilities/for_testing/checkMatrix.hpp
PDoakORNL: added these

src/Utilities/for_testing/CMakeLists.txt
ye-luo: That is OK.

PDoakORNL:
Seems like there is a disk space problem.

ye-luo:
When the develop is updated by merging another PR during CI runs, this type of failure shows up. Just need another update once the other PR progressing to merge finishes.

ye-luo (approved):

PR #3057, opened 2021-03-28
batch Inversion and log det calculation with CUDA

PDoakORNL:
This is the lowest level code and interface to do batched inversion and log determinant calculations. In later PR's it would be called at the same level as the current DiracMatrix.

## Proposed changes
Doesn't change any currernt interfaces, does not bother with a non batched CUDA implementation.  The only place this may want to be considered is for the cublas::getrf and getri calls for very large matrices.  I still think they should have one API that takes vector style inputs.

- New feature

- No

## Checklist

Yes This PR is up to date with current the current state of 'develop'


src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -0,0 +1,212 @@
...
+  const int num_col_blocks = (n + COLBS - 1) / COLBS;
+  dim3 dimBlock(COLBS);
+  dim3 dimGrid(batch_size, num_col_blocks);
+  computeLogDet_kernel<COLBS>
+      <<<dimGrid, dimBlock, 0, hstream>>>(n, lda, CUDATYPECAST(LU_mat), pivots, CUDATYPECAST(logdets));
+  return cudaPeekAtLastError();
+}
+
+template<typename T>
+void computeLogDet_batched(cudaStream_t& hstream,
```
ye-luo, line 109: Is this function actually used in real code?
I thought computeInverseAndDetLog_batched is the way to go.

ye-luo (changes requested):
Review not yet done.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -0,0 +1,212 @@
...
+  const int iw                = blockIdx.x;
+  const int block_num         = blockIdx.y;
+  const T* __restrict__ lu_iw = mat_lus[iw];
+  __shared__ cuDoubleComplex logdet_vals[COLBS];
+  T lu_diag;
+  int n_index = threadIdx.x + block_num * COLBS;
+  if (n_index < n)
+  {
+    lu_diag                  = *(lu_iw + n_index * lda + n_index);
+    logdet_vals[threadIdx.x] = complexDetLog(lu_diag, n_index, pivots);
```
ye-luo, line 59: pivots has no offset for iw? looks strange.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -0,0 +1,212 @@
...
+namespace qmcplusplus
+{
+namespace cuBLAS_LU
+{
+
+__device__ cuDoubleComplex complexDetLog(const double lu_diag, const int n_index, const int* pivots)
+{
+  cuDoubleComplex log_value;
+  log_value.x = log(abs(lu_diag));
+  log_value.y = ((lu_diag < 0) != ((pivots[n_index] - 1) == n_index)) * M_PI;
```
ye-luo, line 25: Seems opposite to computeLogDet in DiracMatrix.h

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -0,0 +1,212 @@
...
+  return log_value;
+}
+
+__device__ cuDoubleComplex complexDetLog(const cuDoubleComplex lu_diag, const int n_index, const int* pivots)
+{
+  cuDoubleComplex diag;
+  double pivot_factor = 1 - 2 * ((pivots[n_index] - 1) == n_index);
```
ye-luo, line 32: Also seems the opposite sign.
why not straightforward.
```
pivots[n_index] - 1 == n_index ? 1 : -1
```

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -0,0 +1,212 @@
...
+  if (n_index < n)
+  {
+    lu_diag                  = *(lu_iw + n_index * lda + n_index);
+    logdet_vals[threadIdx.x] = complexDetLog(lu_diag, n_index, pivots);
+  }
+  if (threadIdx.x == 0 && blockIdx.y == 0)
+    logdets[iw] = {0.0, 0.0};
+  // insure that when we reduce logdet_vals all the threads in the block are done.
+  __syncthreads();
+  if (threadIdx.x == 0)
```
ye-luo, line 65: spawn one block per COLBS is not necessary, this kernel anyway doesn't have much computation.
I expect one walker per block is good enough. By doing that, we can avoid atomic operations since everything is within a block. The n iterations is still done by loop over COLBS segments and then do a binary tree reduction.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -0,0 +1,212 @@
...
+  __shared__ cuDoubleComplex logdet_vals[COLBS];
+  T lu_diag;
+  int n_index = threadIdx.x + block_num * COLBS;
+  if (n_index < n)
+  {
+    lu_diag                  = *(lu_iw + n_index * lda + n_index);
+    logdet_vals[threadIdx.x] = complexDetLog(lu_diag, n_index, pivots);
+  }
+  if (threadIdx.x == 0 && blockIdx.y == 0)
+    logdets[iw] = {0.0, 0.0};
```
ye-luo, line 62: To be safe, this needs to be atomic. I'm not suggesting to use atomic but change algorithms mentioned below.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -0,0 +1,212 @@
...
+  // insure that when we reduce logdet_vals all the threads in the block are done.
+  __syncthreads();
+  if (threadIdx.x == 0)
+  {
+    __shared__ cuDoubleComplex block_sum_log_det;
```
ye-luo, line 67: I don't think `__shared__` is necessary.

ye-luo (commented):
done with my review.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: no this is for testing only.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.hpp
```diff
@@ -0,0 +1,112 @@
...
+                                         OnTypesEqual<T, std::complex<double>**, cuDoubleComplex**>,
+                                         OnTypesEqual<T, std::complex<float>*, cuComplex*>,
+                                         OnTypesEqual<T, const std::complex<double>*, const cuDoubleComplex*>,
+                                         OnTypesEqual<T, const std::complex<float>*, const cuComplex*>,
+                                         OnTypesEqual<T, const std::complex<float>**, const cuComplex**>,
+                                         OnTypesEqual<T, const std::complex<double>**, const cuDoubleComplex**>,
+                                         default_type<void>>::type;
+
+// There must be a way to make this a template function. but it's so easy as a macro.
+#define CUDATYPECAST(var) reinterpret_cast<TypesMapper<decltype(var)>>(var)
```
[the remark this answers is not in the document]
PDoakORNL: will do this

src/QMCWaveFunctions/tests/test_cuBLAS_LU.cpp
```diff
@@ -0,0 +1,505 @@
...
+TEST_CASE("cuBLAS_LU::computeLogDet", "[wavefunction][CUDA]")
+{
+  auto cuda_handles = std::make_unique<testing::CUDAHandles>();
+  int n             = 4;
+  int lda           = 4;
+  auto& hstream     = cuda_handles->hstream;
+
+  void* vp_lu = nullptr;
+  cudaErrorCheck(cudaMallocHost(&vp_lu, sizeof(double) * 16), "cudaMallocHost failed");
+  double* lu = new (vp_lu) double[16]{7., 0.28571429, 0.71428571, 0.71428571,  5., 3.57142857, 0.12, -0.44,
```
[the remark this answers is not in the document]
PDoakORNL: I think no memory leak in this test, in place new of POD array does not require a call to a destructor so deallocation of the memory is all that is necessary, this is done with the cudaFree and cudaFreeHost.  I do not want to also be debugging the behavior of the allocators and containers here.  This is a unit test of cuBLAS_LU only.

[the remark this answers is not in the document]
PDoakORNL: I am changing the way the test and the direct_inversion target are built to make use of the llvm sanitizers possible.

PDoakORNL (commented):
Working on an update to this.  There are many dependency issues preventing clean test programs to fix.

PDoakORNL:
The build failure is due to no space on device.  I wish there was a way to flag these failures as it makes us look far sloppier than we are.

src/Platforms/CUDA/cuBLAS.hpp
```diff
@@ -228,6 +229,103 @@ inline cublasStatus_t gemm_batched(cublasHandle_t& handle,
...
+  return cublasZgetrfBatched(handle, n, reinterpret_cast<cuDoubleComplex* const *>(A), lda, PivotArray, infoArray, batchSize);
+}
+
+inline cublasStatus_t getri_batched(cublasHandle_t& handle,
+                                    int n,
+                                    float* A[],
+                                    int lda,
+                                    int* PivotArray,
+				    float* C[],
+				    int ldc,
```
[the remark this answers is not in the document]
PDoakORNL: This is fixed.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.hpp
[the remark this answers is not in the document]
PDoakORNL: done

prckent:
(We should try to get this in just soon after the v3.11 release, which hopefully won't be long now. )

src/QMCWaveFunctions/tests/test_cuBLAS_LU.cpp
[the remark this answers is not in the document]
PDoakORNL: no longer leaks memory., passes ASAN/LSAN and cuda-memcheck. 

This was accomplished by making a a new library target and separate test executable that do not include unnecessary dependencies. Both of those will have more code added in continuing PRs. Most of the code does not need to know about the more granular qmcwfs sub target as long as they link the qmcwfs target.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -0,0 +1,212 @@
...
+{
+  //LU is returned in Ms
+  cublasErrorCheck(cuBLAS::getrf_batched(h_cublas, n, Ms, lda, pivots, infos, batch_size),
+                   "cuBLAS::getrf_batched failed in computeInverseAndDetLog_batched");
+  // cudaErrorCheck(computeLUDiag_batched_impl(hstream, n, lda, Ms, LU_diags, batch_size),
+  //                "failed to extract LU diag values at cuomputeLUDiag_batched_impl");
+  cudaErrorCheck(computeLogDet_batched_impl(hstream, n, lda, Ms, pivots, log_dets, batch_size),
+                 "failed to calculate log determinant values in computeLogDet_batched_impl");
+  cublasErrorCheck(cuBLAS::getri_batched(h_cublas, n, Ms, lda, pivots, Cs, lda, infos, batch_size),
+                   "cuBLAS::getri_batched failed in computeInverseAndDetLog_batched");
```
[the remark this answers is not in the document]
PDoakORNL: There would ideally be no transfers at this level at all since this is the level equivalent to the cuBLAS calls.   Transfer and synchronization occurs at the DiracMatrix level.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: I thought you were wrong, but after more thought realized it is worse than that.  Worst case ordering I don't think atomic even works here.  Making a different change to make sure the logdet is properly zeroed I was trying to save a fill or a copy here and that was premature optimization. This will be the next push.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: I think you are right, this should be local and only visible in this thread scope.

PDoakORNL:
Rebase is incoming watch out for force push.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: this is fixed.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: fixed, next level up the call stack is the correct place to do this.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: This can be explored once I have profiling set up, for now I think this is the best way to implement this and there is no good reason to rewrite it while the rest of the code already developed for this feature diverges.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: Better to not have divergence.  Perhaps the compiler will produce something similar but likely not.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: remedied sign is fixed.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: Real bug, fixed.  Testing updated to include different pivot sequences so reading the previous determinants pivots fails test.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: This likely will be an outside API call in a future PR.

prckent:
I haven't been following this PR. After all the updates what is blocking a merge?

prckent:
Looking back, OK to abort if any part of the inverse calculation fails. This shouldn't happen for any reasonable set of wavefunctions and electron positions, but is conceivable. We can handle that if/when it occurs in real runs, but most likely it means something is unhealthy with the run.

CMake/unit_test.cmake
```diff
@@ -32,3 +32,23 @@ FUNCTION( ADD_UNIT_TEST TESTNAME PROCS THREADS TEST_BINARY )
     add_test_labels( ${TESTNAME} TEST_LABELS_TEMP )
     set_property(TEST ${TESTNAME} APPEND PROPERTY LABELS "unit")
 ENDFUNCTION()
+
+FUNCTION( ADD_NO_MPI_UNIT_TEST TESTNAME THREADS TEST_BINARY )
```
ye-luo, line 36: If the code is build without MPI, ADD_UNIT_TEST handles the pure no MPI case.
If MPI is used, unit tests still needs to be built and run with MPI although only with 1 MPI rank. This is intentional because on HPC requires launcher (runjob/mpirun) to dispatch the execution on the compute node even for sequential code. 
So I don't see the need of this special treatment.

src/Platforms/CUDA/cudaError.h
```diff
@@ -21,10 +21,19 @@
...
+// and it makes CUDA heavy code tedious to write and read
+#define cudaCheck(ans)                            \
+  {                                               \
+    cudaAssert((ans), __FILE__, __LINE__); \
+  }                                               \
+
+
 /// prints CUDA error messages. Always use cudaErrorCheck macro.
-inline void cudaAssert(cudaError_t code, const std::string& cause, const char* filename, int line, bool abort = true)
+inline void cudaAssert(cudaError_t code, const std::string& cause, int line, const char* filename = "")
```
ye-luo, line 36: function arguments are so strange compared to the caller side. There is no need to change this function.
```
cudaErrorCheck(xxx, "")
```
if you prefer not to give additional info via cause. write
```
cudaCheck(ans)
{
cudaAssert((ans),"",  __FILE__, __LINE__); 
}
```

PDoakORNL: You are right, with the macro there is no reason to do this by defaulting the arguement.

CMake/unit_test.cmake
PDoakORNL: If possible I don't want to run cmake again with different arguments to run a unit test. 

You don't need to link the message target, the mpi wrapper, the mpi libraries and all the problems with them and sanitizers to use jsrun to run a test. But you do need a different set of arguments.

But if you are not in an interactive job, which is the only way I run tests on summit. This then works fine.  Ctest is already controlling affinity and number of "jobs" It's also fine on workstations and the CI.

If we actually ran CI on summit via a submit script we would need to extend this ADD_NO_MPI_UNIT_TEST to include jsrun and a variable to pass the 
--smpiargs="-disable_gpu_hooks" which is needed to run non mpi test using CUDA with spectrum mpi but which in general I don't think we can pass to all our tests.
https://github.com/olcf/olcf-user-docs/pull/173/files

On most other HPC systems you could just use the mpirunner with 1 process and the test.

ye-luo: > If possible I don't want to run cmake again with different arguments to run a unit test.

You can multiple build folders one with and another one without MPI. I saw this better than adding more function variants in our cmake. You only have to cmake once more.

> You don't need to link the message target, the mpi wrapper, the mpi libraries and all the problems with them and sanitizers to use jsrun to run a test. But you do need a different set of arguments.
 
you can run an interactive job and then ssh to the allocated compute node to run what you need for the non-MPI build.

> But if you are not in an interactive job, which is the only way I run tests on summit. This then works fine. Ctest is already controlling affinity and number of "jobs" It's also fine on workstations and the CI.

This is not fine. you may run things on the MOM nodes and got warnings by admins.

> If we actually ran CI on summit via a submit script we would need to extend this ADD_NO_MPI_UNIT_TEST to include jsrun and a variable to pass the
> --smpiargs="-disable_gpu_hooks" which is needed to run non mpi test using CUDA with spectrum mpi but which in general I don't think we can pass to all our tests.
> https://github.com/olcf/olcf-user-docs/pull/173/files

We will not run complicated CI on Summit. More friendly x86 is probably sufficient.

> On most other HPC systems you could just use the mpirunner with 1 process and the test.

That is true. But we do have machines like Theta, login/MOM and compute nodes are completely different architectures.

Sorry, I don't see a critical case that we must use this added function.

PDoakORNL: I'll refine this today and PR, with summit and other systems with special mpi this is necessary.  Summit is the system that makes this necessary for me right now.

Your correct observation about the test code with regard to leaks and that I think all new code should pass sanitizers unambiguously in their unit tests now. This is needed. I of course know about multiple build folders but I don't see that as a good solution there is great devepoer hour cost to compilermany variants of the code inside of the edit, compile, debug, edit... loop.  I feel like (debug,  asan) * complex * precision is enough variants to compile at once. In my new code I am aiming at only requiring (debug, asan).

ye-luo: I just realized that ADD_NO_MPI_UNIT_TEST is only a runner, it has nothing to do with linking the message target, the mpi wrapper. I think if you fix the mpirunner in ADD_NO_MPI_UNIT_TEST, it will probably just the same as ADD_UNIT_TEST

PDoakORNL: That's what I just did :).  Looking in config though I'm not sure what will happen on theta.  That calls an aprunhelper script using the num procs argument. It might work, might not. Going to test on cades and summit.  Do you know of any other systems that have an issue like the cuda_hook library when a non mpi program is executed using jrsun, aprun, mpirun etc.?

ye-luo: Are you saying you find ADD_NO_MPI_UNIT_TEST and ADD_UNIT_TEST becomes identical and we don't need to keep ADD_NO_MPI_UNIT_TEST?

Regarding cuda_hook library, summit is the only one I'm aware of but I'm not surprised HPC center needs it to detect user GPU usage.

PDoakORNL: Accept on summit we don't need ADD_NO_MPI_UNIT_TEST.  On every system I've tried other than summit, the test executables that link `catch_main_no_mpi` instead of `catch_main` work fine with the same mpirunner setup as a 1 proc mpi executable.

ye-luo: Yes. Link `catch_main_no_mpi` should serve your needs. CI indicates you last change here breaks no-MPI case. So Please remove ADD_NO_MPI_UNIT_TEST and use ADD_UNIT_TEST which does everything you need already IMO.

PDoakORNL: Except on summit.

PDoakORNL: Works on summit QMC_MPI=1 with flags as in build_olcf_summit_(Clang|xl).sh, works with openmpi on workstation QMC_MPI=1,0 with no special cmake flags.

prckent (changes requested):
Can you please put a reference to tests/scripts/inversion_ref.py  in test_cuBLAS_LU.cpp ? (or perhaps I missed it).

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
ye-luo: I mean
`dimGrid(batch_size)` instead of `dimGrid(batch_size, num_col_blocks)`
So no more need of atomics. one full matrix per block.
Yes. We can look at this later.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -0,0 +1,253 @@
...
+                                       const int* pivots,
+                                       T* logdets,
+                                       const int batch_size)
+{
+  // Perhaps this should throw an exception. I can think of no good reason it should ever happen other than
+  // developer error.
+  if (batch_size == 0 || n == 0)
+    return cudaSuccess;
+
+  CUDAfill_n(logdets, batch_size, {0.0, 0.0});
```
ye-luo, line 119: I'm more convinced to do one full matrix per block. CUDAfill_n was never intended to be fast.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -0,0 +1,253 @@
...
+
+template void computeLogDet_batched<double>(cudaStream_t& hstream,
+                                            const int n,
+                                            const int lda,
+                                            double** LU_mat,
+                                            const int* pivots,
+                                            std::complex<double>* log_dets,
+                                            const int batch_size);
+
+template void computeLogDet_batched<float>(cudaStream_t& hstream,
```
ye-luo, line 244: Is float used? if not, please remove.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -0,0 +1,212 @@
...
+                                     int* infos,
+                                     std::complex<double>* log_dets,
+                                     const int batch_size)
+{
+  //LU is returned in Ms
+  cublasErrorCheck(cuBLAS::getrf_batched(h_cublas, n, Ms, lda, pivots, infos, batch_size),
+                   "cuBLAS::getrf_batched failed in computeInverseAndDetLog_batched");
+  // cudaErrorCheck(computeLUDiag_batched_impl(hstream, n, lda, Ms, LU_diags, batch_size),
+  //                "failed to extract LU diag values at cuomputeLUDiag_batched_impl");
+  cudaErrorCheck(computeLogDet_batched_impl(hstream, n, lda, Ms, pivots, log_dets, batch_size),
```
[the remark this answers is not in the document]
ye-luo: computeGetrf_batched already sync and check error

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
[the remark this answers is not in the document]
ye-luo: cuBLAS::getri_batched is at the cuBLAS call level. computeInverseAndDetLog_batched is already high level. We must transfer info and check error before exiting. Since anyway, transfer is needed, so prefer to transfer log_dets as well.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.hpp
```diff
@@ -0,0 +1,88 @@
...
+                                     T* Cs[],
+                                     T* LU_diags,
+                                     int* pivots,
+                                     int* host_infos,
+                                     int* infos,
+                                     std::complex<double>* log_dets,
+                                     const int batch_size);
+
+template<typename T>
+void computeGetrf_batched(cublasHandle_t& h_cublas,
```
ye-luo, line 58: All the three are probably for testing only. Need documentation.

src/QMCWaveFunctions/tests/test_cuBLAS_LU.cpp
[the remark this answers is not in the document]
ye-luo: We are throwing exceptions. So only RAII guarantee not missing calls to free. Use containers from the container target! Don't be afraid of having dependencies. Please remove all the pointers in the next PR.

src/Utilities/tests/for_testing/test_checkMatrix.cpp
```diff
@@ -11,58 +11,111 @@
...
+#include <string>
 
+/** \file
+ *  Tests checkMatrix and MatrixAccessor.
+ *  Separate unit tests are likely a waste of effort  since these are complementary
+ *   and in the same module.
+ */
 namespace qmcplusplus
 {
+// When linking containers does cause asan failure this should return
```
ye-luo, line 25: Is containers still the culprit? Not sure if disabling this test is a good move.

src/QMCWaveFunctions/tests/test_dirac_matrix.cpp
```diff
@@ -106,7 +94,256 @@ TEST_CASE("DiracMatrix_inverse", "[wavefunction][fermion]")
...
+{
+  DiracMatrix<std::complex<double>> dm;
+
+  Matrix<std::complex<double>> a, a_T, a_inv;
+  LogValueType LogValue;
+  a.resize(4, 4);
+  a_T.resize(4, 4);
+  a_inv.resize(4, 4);
+
+//     a(0,0) = { 2.0, 0.1 };
```
ye-luo, line 248: clean up?

src/QMCWaveFunctions/tests/CMakeLists.txt
```diff
@@ -63,14 +63,14 @@ SET(JASTROW_SRC test_bspline_jastrow.cpp test_counting_jastrow.cpp test_polynomi
 SET(DETERMINANT_SRC FakeSPO.cpp test_DiracDeterminantBatched.cpp test_dirac_det.cpp
                     test_multi_dirac_determinant.cpp test_dirac_matrix.cpp
                     test_multi_slater_determinant.cpp)
-
+
```
ye-luo, line 66: white spaces.

ye-luo (commented):
I don't really see any more blocking issues for this PR. So @PDoakORNL please go through the latest comment and add replies. For items you would like to fix in later PRs, please add a global comment with a list of bullet points. You may also just add changes to this PR if you prefer not to defer certain fixes. Remember this PR will be merged with all the commits being squashed.

src/Utilities/tests/for_testing/test_checkMatrix.cpp
PDoakORNL: I will try this again, but yes containers linked message

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.hpp
PDoakORNL: Let me get the next two PR's in and it will get nailed down.  I want to be clear about each layers responsibilities and dependencies.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: I'm sure I can write a `__device__` function that will be fast but I think that needs to be in another PR.

src/Platforms/CUDA/CUDATypeMapping.hpp
```diff
@@ -0,0 +1,73 @@
...
+#ifndef QMCPLUSPLUS_CUDA_TYPE_MAPPING_HPP
+#define QMCPLUSPLUS_CUBLAS_LU_HPP
+
+namespace qmcplusplus
+{
+// This namespace is to protect against the clash between this disjuntion and the std one
+// when c++17 is finally enabled for QMCPACK.  When 14 support is dropped confirm this can be dropped and move
+// the the std one.
+
+namespace c14disjunction
```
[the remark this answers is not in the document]
PDoakORNL: https://en.cppreference.com/w/cpp/types/disjunction

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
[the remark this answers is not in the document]
PDoakORNL: my understanding is if getrf has no infos getri won't either.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
PDoakORNL: Maybe in mixed precision, enough has changed here that I want to reserve that option for the next two PR's

src/QMCWaveFunctions/tests/CMakeLists.txt
PDoakORNL: Yuck, Will fix, normally these show up bright red in my emacs commit interface.

src/QMCWaveFunctions/tests/test_cuBLAS_LU.cpp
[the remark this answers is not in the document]
PDoakORNL: I think the testing of this bundle of functions shouldn't depend on them. For this level I just wanted to deal with memory at the same level the cudaCalls do.

src/QMCWaveFunctions/tests/test_dirac_matrix.cpp
PDoakORNL: yes

PDoakORNL (commented):
Working on these changes now.

src/Platforms/CUDA/CUDATypeMapping.hpp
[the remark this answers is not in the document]
ye-luo: I was asking asking for disjunction from 17 but c14disjunction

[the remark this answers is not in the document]
PDoakORNL: It's the possible implementation from cppreference.com

ye-luo (approved):
Soon to come PRs will address remaining issues.

prckent (approved):

ye-luo:
I wanted a squashed commit but I clearly forgot. I won't try to correct this.

Issue #3060, opened 2021-03-29
Memory leak due to observable_helper* allocations


williamfgc:
**Describe the bug**
Address and leak sanitizers included in #3044 exposes a leak for `observable_helper*` objects allocated with `new` in `test_local_energy_est`

```
==40251==ERROR: LeakSanitizer: detected memory leaks
12: 
12: Direct leak of 160 byte(s) in 1 object(s) allocated from:
12:     #0 0x53401d in operator new(unsigned long) (/home/wgodoy/workspace/qmcpack/build/tests/bin/test_estimators+0x53401d)
12:     #1 0x82ff7d in qmcplusplus::LocalEnergyEstimator::registerObservables(std::vector<qmcplusplus::observable_helper*, std::allocator<qmcplusplus::observable_helper*> >&, long) /home/wgodoy/workspace/qmcpack/build/../src/Estimators/LocalEnergyEstimator.cpp:36:20
12:     #2 0x5539cb in qmcplusplus::____C_A_T_C_H____T_E_S_T____4() /home/wgodoy/workspace/qmcpack/build/../src/Estimators/tests/test_local_energy_est.cpp:93:10
12:     #3 0x67220c in Catch::TestInvokerAsFunction::invoke() const /home/wgodoy/workspace/qmcpack/build/../external_codes/catch/catch.hpp:11009:9
12:     #4 0x6643d7 in Catch::TestCase::invoke() const /home/wgodoy/workspace/qmcpack/build/../external_codes/catch/catch.hpp:10910:15
12:     #5 0x664224 in Catch::RunContext::invokeActiveTestCase() /home/wgodoy/workspace/qmcpack/build/../external_codes/catch/catch.hpp:9774:27
12:     #6 0x65fdae in Catch::RunContext::runCurrentTest(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) /home/wgodoy/workspace/qmcpack/build/../external_codes/catch/catch.hpp:9748:17
12:     #7 0x65d8c5 in Catch::RunContext::runTest(Catch::TestCase const&) /home/wgodoy/workspace/qmcpack/build/../external_codes/catch/catch.hpp:9524:13
[...]

PR #3061, opened 2021-03-29
Refactor observable_helper pointer to object

williamfgc:
This is major refactoring affecting several files as we move toward RAII.
Avoid leaks if observable_helper* owns memory and it's not deallocated as exposed by asan.
Addresses #3060

## Proposed changes

Refactor all `observable_helper*` owning memory to objects for RAII

- Bugfix: address a leak by moving memory towards RAII
- Refactoring: this break internal object APIs 
- Documentation changes  (minor doc change)

- Yes: developers might have to rebase develop once this is merged

Ubuntu 20.04, needs to pass CI

## Checklist

- Yes. Documentation has been added (minor change)


williamfgc:
FYI, there is a dangling reference in one of the tests in `runQMC` > `VMC::run` > `EstimatorManager`. I'll try to debug to be able to merge this.

src/QMCHamiltonians/EnergyDensityEstimator.h
```diff
@@ -35,7 +35,7 @@ class EnergyDensityEstimator : public OperatorBase, public PtclOnLatticeTraits
   Return_t evaluate(ParticleSet& P);
   void addObservables(PropertySetType& plist) {}
   void addObservables(PropertySetType& plist, BufferType& olist);
-  void registerCollectables(std::vector<observable_helper*>& h5desc, hid_t gid) const;
+  void registerCollectables(std::vector<observable_helper>& h5desc, hid_t gid) const;
```
ye-luo, line 38: Since the API changed, it can be a chance to add override keyword on all the derived classes. It is your call to change directly in this PR or do it separately. I don't have a preference just want to point this out.

src/Estimators/EstimatorManagerBase.cpp
```diff
@@ -97,7 +97,7 @@ EstimatorManagerBase::~EstimatorManagerBase()
 {
   delete_iter(Estimators.begin(), Estimators.end());
   delete_iter(RemoteData.begin(), RemoteData.end());
-  delete_iter(h5desc.begin(), h5desc.end());
```
ye-luo, line 100: Could you say a bit how the leak happened? Is it because a pointe gets overwritten or it is a problem in the upper level classes? Since you looked at this so may have an answer.

ye-luo (commented):
Are our current tests sufficient to cover changed code paths?

src/Estimators/EstimatorManagerBase.cpp
williamfgc: It gets allocated with `new` and in some instance it doesn't get deallocated and asan with lsan raises the issue in one of the tests.

src/QMCHamiltonians/EnergyDensityEstimator.h
williamfgc: @ye-luo good point. My preference is to try to not use `override` as it breaks the `is-a` relationship. This case is simple and the derived classes should be compliant with the base signature. Thanks!

src/Estimators/tests/test_local_energy_est.cpp
```diff
@@ -87,7 +84,7 @@ TEST_CASE("LocalEnergy with hdf5", "[estimators]")
   (*W.begin())->Properties(WP::LOCALENERGY)    = 1.1;
   (*W.begin())->Properties(WP::LOCALPOTENTIAL) = 1.2;
 
-  std::vector<observable_helper*> h5desc;
```
williamfgc, line 90: @ye-luo I think the leak happens cause these are not deallocated. I could call delete and move on, but in general refactoring towards RAII is the goal to avoid more of these in the future. Hope it helps, thanks!

src/QMCHamiltonians/EnergyDensityEstimator.h
ye-luo: I don't get your point `it breaks the is-a relationship`. My understand of `override` is that it brings clarity and prevent developer errors.

src/Estimators/EstimatorManagerBase.cpp
ye-luo: I'm wondering why it doesn't get de-allocated. "leak" means the pointer is no more accessible which is more than "not deallocated". So need to understand why this happens.

src/Estimators/tests/test_local_energy_est.cpp
ye-luo: I'm not say one way or another. Having a clear understanding helps us writing better codes. "not deallcoated" can be forgetting to delete in the destructor or the object containing the pointer not gotten freed. So I'd like to understand what is the root cause. Since `delete_iter(h5desc.begin(), h5desc.end());` is supposed to free these pointers.

If you think delete can work, where do you think of to put it. It probably helps me to understand the full problem.

src/QMCHamiltonians/EnergyDensityEstimator.h
williamfgc: @ye-luo you are right on `override`. I totally mispoke I meant not using it. I'd still prefer using `final` instead where applicable, thoughts?

src/Estimators/tests/test_local_energy_est.cpp
williamfgc: Issue #3060 has steps to reproduce and the leak backtrace. You're right, delete could be called anywhere within object scope post usage. The goal of moving towards RAII is not having to worry about manual memory management (and fewer lines of code). Hope it helps.

src/QMCHamiltonians/EnergyDensityEstimator.h
ye-luo: Although my experience with `final` is limited, I feel it only useful to scan inheritance by the compiler.
```
class base
{
  virtual void foo() = 0;
}

class A: public Base
{
  virtual void foo() override { //1st level implementation } ;
}

class B: public A
{
  void foo() override { //2nd level implementation } ;
}
```
in this case, final cannot be added to `A::foo`.

final is stronger than override. The additional restriction from final cannot be easily justified by only knowing the base class abstraction.

williamfgc: Your example is correct `final` would only apply to B (or C if it exist) as a final leaf in the vtable.`B::foo() final` would kill the ambiguity of `B::foo() override` as we would mark it as "final function that can't we can't override", always executed if B is created. As for A (or B if C exists), `override` is the way to go as it is clear: "overrides `Base:foo()`, but it is not final. If C is to be final, we move `final``C::foo()` and redefine to `B:foo() override`. On a related note, some compilers (PGI if I recall correctly) issue warnings by default if not all virtual functions are overridden.

src/Estimators/tests/test_local_energy_est.cpp
prckent: Is this leaky pattern repeated in the main application or does it only occur in the test codes?

williamfgc: @prckent that's a good question. For this particular case, I have only see it in the test code. asan failures happen in unit test (real build so far). We can either just try to manage memory manually without modifying internal structures, or refactor towards RAII, so new tests won't have to worry about it. I can always reenable asan to take a look at the failures (mostly coming from `new`). Let me know your thoughts.

```
The following tests FAILED:
	 12 - deterministic-unit_test_estimators (Failed)
	 17 - deterministic-unit_test_einspline (Failed)
	 20 - deterministic-unit_test_wavefunction_trialwf (Failed)
	 21 - deterministic-unit_test_wavefunction_sposet (Failed)
	 22 - deterministic-unit_test_wavefunction_jastrow (Failed)
	 23 - deterministic-unit_test_wavefunction_determinant (Failed)
	 24 - deterministic-unit_test_hamiltonian (Failed)
	 26 - deterministic-unit_test_drivers (Failed)
	 27 - deterministic-unit_test_new_drivers (Failed)
	 29 - deterministic-unit_test_qmcapp (Failed)
```

williamfgc: It can be a bit convoluted. This line [`RandomNumberControl::make_seeds();]`(https://github.com/QMCPACK/qmcpack/blob/develop/src/QMCApp/tests/test_rng_control.cpp#L33) is exposed in the the leak, but looking at the API of `RandomNumberControl` I don't know if there is a way to deallocate directly.

```
Direct leak of 2600 byte(s) in 1 object(s) allocated from:
29:     #0 0x4d0e9d in operator new(unsigned long) (/home/wgodoy/workspace/qmcpack/build/tests/bin/test_qmcapp+0x4d0e9d)
29:     #1 0x656c8b in qmcplusplus::RandomNumberControl::make_children() /home/wgodoy/workspace/qmcpack/build/../src/OhmmsApp/RandomNumberControl.cpp:96:24
29:     #2 0x656832 in qmcplusplus::RandomNumberControl::make_seeds() /home/wgodoy/workspace/qmcpack/build/../src/OhmmsApp/RandomNumberControl.cpp:87:3
29:     #3 0x4dbe7a in qmcplusplus::____C_A_T_C_H____T_E_S_T____0() /home/wgodoy/workspace/qmcpack/build/../src/QMCApp/tests/test_rng_control.cpp:33:3
29:     #4 0x54b146 in Catch::TestInvokerAsFunction::invoke() const /home/wgodoy/workspace/qmcpack/build/../external_codes/catch/catch.hpp:11009:9
29:     #5 0x52bee2 in Catch::RunContext::runCurrentTest(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) /home/wgodoy/workspace/qmcpack/build/../external_codes/catch/catch.hpp:9748:17
29:     #6 0x528952 in Catch::RunContext::runTest(Catch::TestCase const&) /home/wgodoy/workspace/qmcpack/build/../external_codes/catch/catch.hpp:9524:13
29:     #7 0x53d3a9 in Catch::(anonymous namespace)::runTests(std::shared_ptr<Catch::Config> const&) /home/wgodoy/workspace/qmcpack/build/../external_codes/catch/catch.hpp:10073:39
29:     #8 0x53d3a9 in Catch::Session::runInternal() /home/wgodoy/workspace/qmcpack/build/../external_codes/catch/catch.hpp:10266:27
29:     #9 0x53ac1d in Catch::Session::run() /home/wgodoy/workspace/qmcpack/build/../external_codes/catch/catch.hpp:10223:24
29:     #10 0x7fb8d889c0b2 in __libc_start_main /build/glibc-eX1tMB/glibc-2.31/csu/../csu/libc-start.c:308:16
```

prckent:
asan failures have to be fixed outside of the CI. Our CI always has to pass otherwise it is simply not workable for the PR approvers or PR makers. Hence we can't enable asan in the CI until either all the problems are fixed or there is a giant exclusion list. Since I have not seen the current list of failures, what is the preferred route?

If the exclusion list is reasonably finite, stopping new asan problems in the deterministic test set would be an excellent step in my opinion. This would stop any further deterioration.

My preference is for quick iteration whatever the chosen path turns out to be.

src/QMCHamiltonians/EnergyDensityEstimator.h
prckent: I would like to see at least override used everywhere, final where appropriate. I think that we have had one error that would have been caught through a required override statement (error not warning). No need to add here though. We have been gradually adding override and could add a mention to the developer docs. I think it certainly brings clarity.

williamfgc:
@prckent Thanks! My bad, I meant to add `asan` to a PR we won't merge until issues are resolved. I agree, CI must be spotless. 

> If the exclusion list is reasonably finite, stopping new asan problems in the deterministic test set would be an excellent step in my opinion. This would stop any further deterioration.

Sounds good, let me try to add labels to the current failures and see if we can skip when running `asan` checks.

ye-luo:
If our github actions for sanitizers is just running ctest, we can use -R and give a list of tests we are sure to pass and then gradually increase as we fix up more tests.

prckent:
I like @ye-luo simple suggestion (== -E the_failing_ones)

williamfgc:
@prckent @ye-luo yup, that's the plan, see [here](https://gitlab.kitware.com/cmake/cmake/-/issues/18127). Thank you both for your input.

src/QMCHamiltonians/EnergyDensityEstimator.h
PDoakORNL: In practice I've found override to be quite useful at catching API issues during development. It hasn't made like the amount of virtual inheritance in the code more but its alerted me to quite a number of potential issues very early in coding rather than at runtime.

PDoakORNL:
I'd add that RAII or not the observable_helper is actually a handle with a vague lifetime to a particular part of an hdf5 file with hopefully an enclosing lifetime.  The overall design is quite tortured and not very abstracted from hdf5 at all (I'm not a huge fan of the file format). 

It really shouldn't be called observable_helper and it doesn't belong in QMCHamiltonian. Our data output design would benefit from a rethink.

ye-luo (changes requested):
The changes actually are positive in a good direction. Some minor change requests
1. observable_helper -> ObservableHelper to follow good naming conventions. Best to do in this PR since we are anyway changing all the relevant places.
2. Need a test for observable_helper copy constructor, I don't have good faith in it. Or mark the copy constructor delete and use emplace_back to construct in place and avoid push_back copy. I actually prefer the latter.
3. Some strange formatting, probably need to run the format before merging.
Let us get this moving.

ye-luo:
Test this please.

ye-luo:
```
yeluo@epyc-server:~/opt/qmcpack/build_gcc_real$ git diff
diff --git a/src/QMCHamiltonians/DensityEstimator.cpp b/src/QMCHamiltonians/DensityEstimator.cpp
index f596c5112..a9c23cd03 100644
--- a/src/QMCHamiltonians/DensityEstimator.cpp
+++ b/src/QMCHamiltonians/DensityEstimator.cpp
@@ -124,10 +124,10 @@ void DensityEstimator::registerCollectables(std::vector<observable_helper>& h5de
   std::vector<int> ng(OHMMS_DIM);
   for (int i = 0; i < OHMMS_DIM; ++i)
     ng[i] = NumGrids[i];
-  observable_helper h5o(myName);
+  h5desc.emplace_back(myName);
+  auto& h5o = h5desc.back();
   h5o.set_dimensions(ng, myIndex);
   h5o.open(gid);
-  h5desc.push_back(h5o);
 }
 
 void DensityEstimator::setObservables(PropertySetType& plist)
yeluo@epyc-server:~/opt/qmcpack/build_gcc_real$ ctest -R deterministic-diamondC_1x1x1_pp-vmc-estimator-density
Test project /home/yeluo/opt/qmcpack/build_gcc_real
    Start 959: deterministic-diamondC_1x1x1_pp-vmc-estimator-density-1-1
1/4 Test #959: deterministic-diamondC_1x1x1_pp-vmc-estimator-density-1-1 .........   Passed    1.00 sec
    Start 960: deterministic-diamondC_1x1x1_pp-vmc-estimator-density-1-1-check
2/4 Test #960: deterministic-diamondC_1x1x1_pp-vmc-estimator-density-1-1-check ...   Passed    0.83 sec
    Start 961: deterministic-diamondC_1x1x1_pp-vmc-estimator-density-4-4
3/4 Test #961: deterministic-diamondC_1x1x1_pp-vmc-estimator-density-4-4 .........   Passed   10.90 sec
    Start 962: deterministic-diamondC_1x1x1_pp-vmc-estimator-density-4-4-check
4/4 Test #962: deterministic-diamondC_1x1x1_pp-vmc-estimator-density-4-4-check ...   Passed    0.80 sec
```
copy constructor is bad.

williamfgc:
@ye-luo this is great, thanks a lot. I just put your fix to address the current issue and make sure CI passes. Once it passes I will make the other requested changes on formatting and removing the copy constructor (more likely tomorrow). Thanks for moving this forward.

ye-luo:
> @ye-luo this is great, thanks a lot. I just put your fix to address the current issue and make sure CI passes. Once it passes I will make the other requested changes on formatting and removing the copy constructor (more likely tomorrow). Thanks for moving this forward.

Only density is fixed, CI requires fixing other places as well.

williamfgc:
@ye-luo down to 4 after deleting the copy constructor:

```
        878 - deterministic-diamondC_1x1x1_pp-vmc-estimator-spindensity-1-1-check (Failed)
	880 - deterministic-diamondC_1x1x1_pp-vmc-estimator-spindensity-4-4-check (Failed)
	1048 - deterministic-diamondC_1x1x1_pp-vmc-dmc-trace_scalars-1-1-check (Failed)
	1050 - deterministic-diamondC_1x1x1_pp-vmc-dmc-trace_scalars_sel-1-1-check (Failed)
```
I need to see what's special about these two types, so far they follow similar patterns as the fixed ones.

williamfgc:
The error comes from `u` values in `det_qmc_spindens_short.s000.stat.h5` being all zeroes (which shouldn't be).  
- this PR: `<             (0,38): 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,`
- develop: `>             (0,38): 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0.333333, 0, 0, 0,`

I'll keep debugging.

ye-luo:
At the run, I saw some error printed out from the hdf5 library. I think it indicates something, If you are exhausted in debugging on this issue, I can help you having a look. So feel free to work on the other pending PR.

williamfgc:
@ye-luo that's correct, I got to that point. The `u` hdf5 handlers might be corrupted. I think mixing RAII with C APIs is adding some trickery as we're relying on the destructors for proper close of the hdf5 handlers. Right now, we call them explicitly with delete. Switching to objects the destructor is called also when reallocating `SpinDensity::h5desc`. I will implement a proper `close` function in `observable_helper`.

williamfgc:
@ye-luo  I will work on the naming changing request below now that this is passing GitHub Actions CI. Thanks.

> observable_helper -> ObservableHelper to follow good naming conventions. Best to do in this PR since we are anyway changing all the relevant places.

ye-luo:
@williamfgc could you elaborate how the move constructor got involved?

williamfgc:
@ye-luo sure, the default move constructor just "moves" the object members (probably making a copy as they are cheap). We need to also move resource (hdf5 handlers) "ownership" by somehow telling the input object (e.g. via a bool flag) that they can't close those resources anymore when their destructors are called. HDF5 handlers use global ids (hence always a shared resource), so anyone can call `H5X_close(global_id)` if `global_id` is valid.
 
```
// assume foo[0] = observable_u;
foo.emplace_back("d"); // here if foo is reallocated so ~foo[0] gets called when moving foo's underlying memory to another location. 

// foo belongs to bar
~bar::bar()
{
    //~foo[i] gets called again when foo goes out of scope...the HDF5 errors appeared here
}
```

The only thing guaranteed with `std::vector` is that memory is contiguous, but not necessarily persistent with reallocations (might be a consequence of using C's `realloc` under the hood or the equivalent OS call).  

What makes it really interesting is that this is only seen in a couple of tests, which I'm very grateful are there :). Thanks for your help, I'm glad we are moving forward.

ye-luo:
@williamfgc I see. The vector was resized as more pieces are added to it. So triggered move. Good catch. In this case, there is no chance we can know ahead the size and hence no way to reserve.

williamfgc:
@ye-luo thanks! This should be ready for testing now, I ended up refactoring the header only struct into a class with its own translation unit to avoid recompiling with every change.

ye-luo:
Okay to test

ye-luo (approved):

williamfgc:
Fixes #3060

PR #3076, opened 2021-04-05
Nexus: fix machines file and add tests

jtkrogel:
## Proposed changes

Fix new issues with machines.py (syntax, duplicate class) and add tests for Andes submission script generation.  Without these updates, Nexus fails to run.

- Bugfix

- No

Laptop

## Checklist


markdewing:
okay to test

prckent:
Hmm. Apologies Jaron.

Did ntest correctly break with this bad update?

We should use this breakage as motivation to add some of the nexus tests to the CI.

jtkrogel:
nxs-test does correctly produce failures here when run through ctest.  Essentially all tests failed.  

I think adding Nexus tests to the CI is a good idea.  Running all tests takes about 50s on my laptop.

ye-luo (approved):

PR #3081, opened 2021-04-07
Enable address sanitizer (asan) on GitHub Actions CI

williamfgc:
## Proposed changes
Enable address sanitizer for real build on GitHub Actions CI
Filter out non-passing tests
Related to #3020 

- Build related changes: add test labels to filter out failed asan tests
- Testing changes: new asan CI workflow on GitHub Actions

- No

Ubuntu 20.04
Must pass on GitHub Actions

## Checklist

- No. Code added or changed in the PR has been clang-formatted (NA)
- Yes. This PR adds tests to cover any new code, or to catch a bug that is being fixed : future asan failures won't pass CI


williamfgc:
For some reason the number of tests in the complex build goes down which takes down coverage.

prckent:
@williamfgc Are you happy with this now? The coverage looks to be restored after your last push.

williamfgc:
@prckent yes, it should be good to go. Thanks for checking.

prckent (approved):
Thanks.

Can you please make an issue with the list of excluded/failing tests? e.g. a checklist we can go through. I am happy and surprised that we are only having problems in the unit tests and the bigger integration tests are fine. The unit tests should be very fixable.

prckent:
OK to test

williamfgc:
@prckent sounds good, I opened #3083 to track and fix those. Should I add another job for complex asan checks?  Yes, QMCPACK itself has few issues, using C++ (e.g. STL) helps a lot in that regard.

prckent:
Please do.

ye-luo:
```
CC=$CCompiler CXX=$CXXCompiler 
 -DMPI_C_COMPILER=mpicc -DMPI_CXX_COMPILER=mpicxx
```
This is not the recommended way for QMCPACK users to indicate compilers because setting separate MPI compilers complicates setting and confuses users. Although CC, CXX environment variables are taken by cmake, we removed mentioning it in our docs and always use CMAKE_C_COMPILER and CMAKE_CXX_COMPILER which overrides any pre-existing environment variables.

Any real benefit setting separate MPI_Language_COMPILER?

I prefer asan with non-MPI builds to stay away from potential problems  MPI libraries. OpenMPI is one of them.
```
 -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx -DQMC_MPI=0
```

williamfgc:
@ye-luo thanks for bringing this up. Short answer: following cmake recommendation for HPC and Cray (nvcc, clang, gcc, mpicc all want to be CC). There is a cmake issue related [here](https://gitlab.kitware.com/cmake/cmake/-/issues/17538). It's a longer discussion for QMCPACK and @prckent as you know the community better, my recommendation is to just follow cmake.

> I prefer asan with non-MPI builds to stay away from potential problems from MPI libraries.

The lsan filter ignores those identified issues with C underlying libraries.

I think `codecov/project` has a floating point rounding error for coverage as nothing has changed since it passed.

ye-luo:
> @ye-luo thanks for bringing this up. Short answer: following cmake recommendation for HPC and Cray (nvcc, clang, gcc, mpicc all want to be CC). There is a cmake issue related [here](https://gitlab.kitware.com/cmake/cmake/-/issues/17538). It's a longer discussion for QMCPACK and @prckent as you know the community better, my recommendation is to just follow cmake.
> 
> > I prefer asan with non-MPI builds to stay away from potential problems from MPI libraries.
> 
> The lsan filter ignores those identified issues with C underlying libraries.
> 
> I think `codecov/project` has a floating point rounding error for coverage as nothing has changed since it passed.

Right now please stick to what QMCPACK recommends as written in README.md. We have sufficient experience with HPC all over the world. The reality is very HPC system is unique. Even Cray systems are not uniform. They all have them being taken care of. So please avoid introducing any divergence in the scripts, documentation, bug reports.

If what QMCPACK recommend doesn't work but your way works, open an issue specifically for that. We can start from there.

ye-luo (approved):

prckent:
Some background after discussion/reminders from Ye (thanks): Currently we should point users to only use CMAKE_C_COMPILER and CMAKE_CXX_COMPILER for both MPI and non-MPI builds, and not any of the other possible ways of configuring. Our scripts should ideally only use our recommended way, while we try to ensure the other routes also work. (e.g. autodetection of MPI is generally a disaster but does work if you have a clean system). The reason for being insistent is that the other routes introduce more ways for users to unintentionally use different and incompatible versions of compilers etc. and to generally get into trouble. This has happened in the past and we don't want to have to debug their misconfiguration or have them blame us. e.g. mismatched openmp runtimes, c++ libraries, mpi compilers.

williamfgc:
@ye-luo @prckent thanks for the background info. I understand, I'll follow the recommended way. For the sanitizers case, I'm using the clang compiler in combination with MPI, so they have to be set explicitly and using `CMAKE_C_COMPILER` makes them mutually exclusive. Please let me know if there is a preferred solution I should follow, I'll be happy to accommodate.

src/Estimators/tests/CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ TARGET_LINK_LIBRARIES(${UTEST_EXE} qmcutil qmcestimators_unit qmcham_unit qmcwfs
 ENDIF()
 
 ADD_UNIT_TEST(${UTEST_NAME} 1 1 "${QMCPACK_UNIT_TEST_DIR}/${UTEST_EXE}")
+IF( "${ENABLE_SANITIZER}" STREQUAL "asan" ) 
+  SET_TESTS_PROPERTIES(${UTEST_NAME} PROPERTIES LABELS noasan)
```
ye-luo, line 28: @williamfgc This one wipes out existing labels. Please append the noasan label

PR #3094, opened 2021-04-13
Exclude unit tests from codecov

prckent:
## Proposed changes

Exclude the unit test directories from codecov analysis

- No

None. Needs to run through CI and have codecov change verified.

## Checklist


ye-luo (approved):

prckent:
Seems OK based on the coverage report.

ye-luo:
Why the coverage % went down when less source code are checked against presumably the same amount of tests? Or I misunderstand the meaning of coverage?

prckent:
Presumably it is using the old reference counts to compute the percentage.

PR #3099, opened 2021-04-14
Nexus: Radial density of requested species only

mcbennet:
## Proposed changes

Allows users to calculate radial density of subset of species -- passed as an argument (`species`) to `radial_density`.
This can speed up workflows, for instance, when user is only interested in magnetic moment of one (or few) species.
Significant speed-up when user is resampling magnetic moment (or charge) of one (or few) species -- in this case `radial_density` will generally be called hundreds of times.

nxs-test yields 100% success rate.

- New feature

- No

Laptop

## Checklist


jtkrogel (approved):
This looks fine.

ye-luo:
Okay to test

PR #3110, opened 2021-04-16
Fix convertpw4qmc mpi and test

### LINKED ISSUE #3106: convertpw4qmc is failing with MPI and also showing as passing
**Describe the bug**

https://cdash.qmcpack.org/CDash/testDetails.php?test=12850469&build=201564

There is an MPI bug in convertpw4qmc causing it to abort. There is an additional bug in our handling of this error where the test is shown as passing.

```
Return code from conversion nonzero:  1
Stderr not empty 
b'xml file comes from quantum espresso\n*** The MPI_Type_free() function was called before MPI_INIT was invoked.\n*** This is disallowed by the MPI standard.\n*** Your MPI job will now abort.\n[sulfur.ornl.gov:2862854] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!\n'
   pass
```

The bug explains the absence of test coverage for convertpw4qmc in QMCTools

Based on the cdash entries only, the non-MPI builds might be OK.

**To Reproduce**

ctest -R converter_test_O2_pwscf 

**Expected behavior**

Converter runs to completion.

Test passes/fails as appropriate.

**System:**

nitrogen, sulfur, docker-ci...

**Additional context**

Absence of coverage at https://codecov.io/gh/QMCPACK/qmcpack/tree/develop/src/QMCTools

prckent:
## Proposed changes

Closes #3106

Add an MPI init and finalize. (This is workaround for some unintended MPI contamination and strictly shouldn't be needed.)
Improve logic in pwconvert_test.py
Increase test robustness via pass regex
Fix uninitialized data usage detected by ubsan now that the sanitizers are running
Don't check standard error in python test script if sanitizer output detected

- Bugfix

- No

nitrogen, gcc+openmpi

## Checklist


prckent:
(Added WIP: will see if the stderr handling can be improved)

prckent:
OK, this is working. Ready for review + merge.

ye-luo (approved):
LGTM

PR #3121, opened 2021-04-20
[CMake] Cleanup compiler detection and change version handling

ye-luo:
## Proposed changes
1. explicit list C and CXX in project language
2. identify compiler collection by CMAKE_CXX_COMPILER_ID only. The C one is ignored.
3. Error out if CMAKE_CXX_COMPILER_ID and CMAKE_C_COMPILER_ID don't match.
4. Use CMake project() to handle QMCPACK version.

- Build related changes

- No

epyc-server

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 ######################################################################
 # CMake version and policies
 ######################################################################
-CMAKE_MINIMUM_REQUIRED(VERSION 3.14.0)
+cmake_minimum_required(VERSION 3.14.0)
 
 # CMP0074: CMake find_package will use <PackageName>_ROOT CMake variable
 # and environment variable in search path.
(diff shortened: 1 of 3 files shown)


prckent (approved):
LGTM assuming the CI is OK

We'll have to update https://github.com/QMCPACK/qmcpack/wiki/Release-process , but pulling that text into the manual would be even better.

Did anything trigger this? e.g. an environment that makes it easy to use mismatched compilers.

ye-luo:
Mismatched compilers usually happens as a user mistake. It could happen when the environment put wrong environment variables like CC/CXX, it is unlikely on seriously maintained production machines but not always on random machines and can trick newbies. So better to add this guard.
I'm not aware of any C++ compiler which doesn't have a C compiler shipped with it.

PR #3122, opened 2021-04-20
No more malloc or free in test_cuBLAS_LU.cpp

PDoakORNL:
## Proposed changes

Removal inplace news and direct CUDA malloc and free from direct inversion testing.

- New feature

- No

leconte

## Checklist

- Yes  Documentation has been added (if appropriate)


src/Platforms/CUDA/CUDAallocator.hpp
```diff
@@ -113,6 +119,18 @@ struct CUDAAllocator
...
+
+  /** Don't do anything on construct, we can't touch this memory directly.
+   *
+   *  This is needed to use this allocator with std::vector in c++14
+   *  In c++20 this should parhaps be removed.  By then std::vector does something
+   *  better? It's somewhat unclear.
+   *
+   */
+  template< class U, class... Args >
+  void construct(U* p, Args&&... args)
```
ye-luo, line 131: Is this needed to prevent std::vector trying to initialize the memory? If yes, I think will not help eliding the initialization with C++17 because C++17 mark this deprecated which means not relying on this but still initialize the memory. I kind of not worried about adding containers as a dependency insteading of using std::vector. If you worked hard to make std::vector to work, I'm afraid is more dangerous than useful.

ye-luo:
Forgot to say, very happy to see the malloc/free going away.

src/Platforms/CUDA/CUDAallocator.hpp
PDoakORNL: For this testing I don't want the dependency on Ohmms.

If the `allocator_traits<CUDAAllocator<T>>` was actually a proper specialization of `std::allocator_traits` that's where for C++17 this construct method belongs.  Instead it is a misleading doppelganger of that template.
The correct thing to do would be to properly specialize std::allocator_traits. Currently this requires a complete and not partial specialization since qmcplusplus::allocator_traits changes the declaration of the template.

Through C++17 the unspecialized std::allocator_traits<Alloc> calls the Alloc.construct method. In C++20 that will be removed. I don't see the value to locking in OhmmsPETE below the level where that dependency is necessary.

ye-luo: `qmcplusplus::allocator_traits` should be renamed to avoid confusing.

My real concern is if the current code works or not with C++17? My guess is that since this function is deprecated in C++17, this function may or may not be used in the C++17 implementation of std::vector. So the added function makes C++14 to work but behaves as not existed when C++17 or C++20 is used. Could you try with C++17 to see if your test can be compiled?

PDoakORNL: That's a good idea.

src/Platforms/allocator_traits.hpp
```diff
@@ -20,7 +20,7 @@ namespace qmcplusplus
 /** template class defines whether the memory allocated by the allocator is host accessible
  */
 template<class Allocator>
-struct allocator_traits
+ struct qmc_allocator_traits
```
ye-luo, line 23: Formatting? Could you fix this when merging develop before we proceed to merge?

ye-luo (approved):
LGTM

Issue #3083, opened 2021-04-08
ASAN test failures 


williamfgc:
**Describe the bug**
Address sanitizers (ASAN) failures are currently reported from LSAN [as it is turned on by default on Linux](https://clang.llvm.org/docs/AddressSanitizer.html#memory-leak-detection). Tests are excluded from CI as added in PR #3081.
The following unit test have been identified as not currently passing ASAN and are currently labeled as `noasan` using cmake:

```
The following tests FAILED:
	 12 - deterministic-unit_test_estimators (Failed)
	 17 - deterministic-unit_test_einspline (Failed)
	 20 - deterministic-unit_test_wavefunction_trialwf (Failed)
	 21 - deterministic-unit_test_wavefunction_sposet (Failed)
	 22 - deterministic-unit_test_wavefunction_jastrow (Failed)
	 23 - deterministic-unit_test_wavefunction_determinant (Failed)
	 24 - deterministic-unit_test_hamiltonian (Failed)
	 26 - deterministic-unit_test_drivers (Failed)
	 27 - deterministic-unit_test_new_drivers (Failed)
	 29 - deterministic-unit_test_qmcapp (Failed)
```

**To Reproduce**
1. `develop` branch
2. configuration:
```
CCompiler=clang
CXXCompiler=clang++

CC=$CCompiler CXX=$CXXCompiler cmake -GNinja \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
  -DMPI_C_COMPILER=mpicc -DMPI_CXX_COMPILER=mpicxx \
  -DENABLE_SANITIZER=asan \
  -DQMC_COMPLEX=0 \
  ..
```
`export LSAN_OPTIONS=suppressions=${QMC_HOME}/config/sanitizers/lsan.supp`
`${QMC_HOME}` is your local QMCPACK source location.	
3. Build: `ninja`
4. `ctest -L noasan`

**Expected behavior**
Tests must pass.
After fixing those unit tests they must be incorporated into asan checks on CI

**System:**
 - Ubuntu 20.04
 - GitHub Actions
 - should be reproducible everywhere

**Additional context**
This is only for the real build. We can expand this ticket to complex build failures (if any)

ye-luo:
step 4 is it ninja or ctest?

williamfgc:
@ye-luo  good catch, thanks!

PDoakORNL:
[...]

PR #3126, opened 2021-04-21
[WIP] Refactor raw pointers in QMCDrivers into shared_ptr to address leak

williamfgc:

## Proposed changes

Refactor raw pointers in QMCDrivers into shared_ptr for RAII (automatic deallocation). 
Affects `OperatorBase`, `addOperator` and `makeclone`, `getHamiltonian` functions that retrieve/interact with `OperatorBase`
Raw pointers were creating a leak as part of #3083 identified tests
Tried `unique_ptr` first, but these pointers have shared ownership (some tests failed).
Must still use pointers due to runtime polymorphism.

- Bugfix: addressing leak related to using raw pointers
- Refactoring (no functional changes, no api changes): internal API changed for `OperatorBase`

- Yes: internal API changed for `OperatorBase`, no external change

Ubuntun 20.04, must pass CI

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCCostFunction.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunction.cpp
@@ -199,7 +199,7 @@ void QMCCostFunction::getConfigurations(const std::string& aroot)
       H_KE_Node[ip]->addOperator(hClones[ip]->getHamiltonian("Kinetic"), "Kinetic");
       if (includeNonlocalH != "no")
       {
-        OperatorBase* a = hClones[ip]->getHamiltonian(includeNonlocalH);
+        std::shared_ptr<OperatorBase> a = hClones[ip]->getHamiltonian(includeNonlocalH);
         if (a)
         {
           app_log() << " Found non-local Hamiltonian element named " << includeNonlocalH << std::endl;
@@ -268,8 +268,9 @@ void QMCCostFunction::checkConfigurations()
         HDerivRecords[ip]->resize(wRef.numSamples(), NumOptimizables);
       }
     }
-    OperatorBase* nlpp = (includeNonlocalH == "no") ? 0 : hClones[ip]->getHamiltonian(includeNonlocalH);
-    bool compute_nlpp  = useNLPPDeriv && nlpp;
+    std::shared_ptr<OperatorBase> nlpp =
+        (includeNonlocalH == "no") ? nullptr : hClones[ip]->getHamiltonian(includeNonlocalH);
+    bool compute_nlpp = useNLPPDeriv && nlpp;
     //set the optimization mode for the trial wavefunction
     psiClones[ip]->startOptimization();
     //    synchronize the random number generator with the node
@@ -394,8 +395,9 @@ void QMCCostFunction::engine_checkConfigurations(cqmc::engine::LMYEngine<Return_
         //HDerivRecords[ip]->resize(wRef.numSamples(),NumOptimizables);
       }
     }
-    OperatorBase* nlpp = (includeNonlocalH == "no") ? 0 : hClones[ip]->getHamiltonian(includeNonlocalH.c_str());
-    bool compute_nlpp  = useNLPPDeriv && nlpp;
+    std::shared_ptr<OperatorBase> nlpp =
+        (includeNonlocalH == "no") ? nullptr : hClones[ip]->getHamiltonian(includeNonlocalH.c_str());
+    bool compute_nlpp = useNLPPDeriv && nlpp;
     //set the optimization mode for the trial wavefunction
     psiClones[ip]->startOptimization();
     //    synchronize the random number generator with the node
--- a/src/QMCHamiltonians/OperatorBase.h
+++ b/src/QMCHamiltonians/OperatorBase.h
@@ -32,6 +32,7 @@
 #endif
 #include "QMCWaveFunctions/OrbitalSetTraits.h"
 #include <bitset>
+#include <memory> // std::unique_ptr
 
 namespace qmcplusplus
 {
@@ -336,7 +337,9 @@ struct OperatorBase : public QMCTraits
    */
   virtual void releaseResource(ResourceCollection& collection) {}
 
-  virtual OperatorBase* makeClone(ParticleSet& qp, TrialWaveFunction& psi) = 0;
+  virtual std::shared_ptr<OperatorBase> makeClone(ParticleSet& qp, TrialWaveFunction& psi) = 0;
+
+  //virtual std::unique_ptr<OperatorBase> makeClone(ParticleSet& qp, TrialWaveFunction& psi, QMCHamiltonian& H);
 
   virtual void setRandomGenerator(RandomGenerator_t* rng)
   {
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
@@ -184,7 +184,7 @@ std::unique_ptr<QMCHamiltonian> QMCCostFunctionBatched::extractFixedHamiltonianC
   KE_Ham->addOperator(H.getHamiltonian("Kinetic"), "Kinetic");
   if (includeNonlocalH != "no")
   {
-    OperatorBase* a = H.getHamiltonian(includeNonlocalH);
+    std::shared_ptr<OperatorBase> a = H.getHamiltonian(includeNonlocalH);
     if (a)
     {
       app_log() << " Found non-local Hamiltonian element named " << includeNonlocalH << std::endl;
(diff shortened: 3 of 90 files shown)


prckent:
OK to test

src/QMCDrivers/WFOpt/QMCCostFunction.cpp
```diff
@@ -268,8 +268,9 @@ void QMCCostFunction::checkConfigurations()
         HDerivRecords[ip]->resize(wRef.numSamples(), NumOptimizables);
       }
     }
-    OperatorBase* nlpp = (includeNonlocalH == "no") ? 0 : hClones[ip]->getHamiltonian(includeNonlocalH);
-    bool compute_nlpp  = useNLPPDeriv && nlpp;
+    std::shared_ptr<OperatorBase> nlpp =
```
ye-luo, line 271: nlpp is a reference and should not use shared_ptr.

src/QMCHamiltonians/OperatorBase.h
```diff
@@ -336,7 +337,9 @@ struct OperatorBase : public QMCTraits
    */
   virtual void releaseResource(ResourceCollection& collection) {}
 
-  virtual OperatorBase* makeClone(ParticleSet& qp, TrialWaveFunction& psi) = 0;
+  virtual std::shared_ptr<OperatorBase> makeClone(ParticleSet& qp, TrialWaveFunction& psi) = 0;
```
ye-luo, line 340: The change must be `unique_ptr` as the logic behind makeClone

ye-luo (changes requested):
Here is my concern. When we are updating all the raw pointer to smart pointers, one requirement is to clarify if a memory object has a unique or shared ownership or it should be just a reference or copy. The change in this PR currently defeats the purpose of using smart pointers and makes existing smart pointers confusing whether their ownership has been well defined. I understand some changes are definitely needed to get memory clean but putting shared_ptr blindly is not helping but making the source code worse.

I'm not blaming this PR but to remind people that moving to smart pointer does require understanding and necessary refactoring the code which are non-trival tasks.

I added two comments above, In OperatorBase, makeClone must return unique_ptr. Then the second case should be treated as a reference. So once we put unique_ptr in makeClone, the code should break a lot and then we should discuss how to fix them case by case.

williamfgc:
@ye-luo thanks for sharing your concerns, good points. I see two routes: 
1. Keep `shared_ptr` and resolve the current leak (I still have to fix CUDA related issues in this PR) merging this PR
2. Don't merge this PR and explore `unique_ptr` usage, but would hold asan/lsan fixes (as you rightly said, it's non-trivial)

I agree that `unique_ptr` is preferred, we can progressively move to `unique_ptr` on a case-by-case basis while defining ownership scope. In fact, I tried first using `unique_ptr` and some errors were introduced due to one of the pointers being managed by two separate threads, but I'm not sure if shared ownership is in fact intended. I still think `shared_ptr` can be seen as a patch providing some benefit over raw pointers on ownership information while avoiding leaks, but I'll let you and @prckent decide on the best route.

prckent:
I agree with Ye's points on needing to understand+update the relationships but am also very much in favor of multiple incremental updates. Being memory leak free would have major ongoing benefits since we could protect a large fraction of the codebase in CI going forward.

@williamfgc How leaky are the tests run in the CI after this update? If pretty close to being non-leaky, I suggest we make a new issue describing Ye's concerns and merge this PR, then focus on the few remaining memory issues and update to unique_ptr in sequence. I certainly expect we'll find some threading issues that will require more than a search & replace to fix.

PDoakORNL:
Myself I only want to see shared_ptr if it can be proven that the lifecycle and ownership scope of the object cannot be known. Then we need to consider whether that is a symptom of a design problem to be addressed or desirable.

ye-luo:
CI failure confirms the non-triviality of just going to shared_ptr when the code is not being understood. Based on the limited scope of change in this PR, I don't think get unique_ptr right is difficult. We should put the effort.
@williamfgc could you put up a PR/branch making the unique_ptr change in makeClone and make qmcham target built as much as possible. I think then the rest breakage can be sorted out quickly.

williamfgc:
@prckent there is still a few lingering leaks (I see libxml2 popping up). I'm going on a case-by-case basis to make sure CI is happy and to avoid feature creep on a large PR. @ye-luo yes, I'll be curious as well. Let me address the build issues in this PR, rebase from here and get the `unique_ptr` one on a separate PR/branch following your suggestion. That way we choose one or the other depending on complexity and to prevent getting hung up on addressing other leaks.  @PDoakORNL yes, the motivation is to address leaks, restore asan tests and then progressively move into more complex refactoring.

Thank you all, good discussion. I think it'd help for subsequent changes as there is more to do wrt manual memory management. @ye-luo is correct, it's not trivial.

williamfgc:
please disregard the `codecov/patch` check, it's essentially saying: "you touched this line, but no test is touching it". We'll get a lot of that when refactoring GPU code. `codecov/project` is more important for now.

williamfgc:
This is interesting, the current failing test shows a spurious seg fault on `libthread.so` depending on configuration. [2659](https://jenkins-ci.cels.anl.gov/job/QMCPACK-CI-GNU-MKL18/2659/) and [2658](https://jenkins-ci.cels.anl.gov/job/QMCPACK-CI-GNU-MKL18/2660/) shows different outcomes in 1, but the same in 2 configurations. 

I haven't caught any potential thread issue with `tsan` or `helgrind` on more recent platforms for this test.

@ye-luo is there a way I could have access either to the system or an environment where I can reproduce. I tried reproducing on a docker Ubuntu 16 image (after some of customization as we can't use default gcc, cmake, etc.), but I keep getting the issue reported in #526.

ye-luo:
@williamfgc on ubuntu 18.04, you can build glibc but keep only libmvec.so in the install folder and set the folder to LD_LIBRARY_PATH, then disable the FATAL_ERROR in cmake.

williamfgc:
@ye-luo thanks, yes I don't think if it's an issue with pthreas per-se, but something higher in the stack.
I have tried to reproduce with no success, I don't also don't see any leaks . BTW, I have a branch that attempts to implement `std::unique_ptr` but it's not as simple as it seems, see [here diffs](https://github.com/QMCPACK/qmcpack/compare/develop...williamfgc:asan-quick-fix-unique_ptr). It builds, but several crashes happen when testing. An unintended consequence is that the copy constructor is eliminated. One thing it's not clear is the ownership in [this function](https://github.com/QMCPACK/qmcpack/blob/develop/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp#L181). See my comment: `H` owns `OperatorBase` objects (as pointers), but `KE_Ham` will be owning them too.  But I don't know if shared ownership is intended. 

```
std::unique_ptr<QMCHamiltonian> QMCCostFunctionBatched::extractFixedHamiltonianComponents()
{
  auto KE_Ham = std::make_unique<QMCHamiltonian>();
  // H owns OperatorBase objects (as pointers), but KE_Ham will be owning them too. So not sure if shared_ptr or unique_ptr would be the correct choice.
  KE_Ham->addOperator(H.getHamiltonian("Kinetic"), "Kinetic");
  if (includeNonlocalH != "no")
  {
    OperatorBase* a = H.getHamiltonian(includeNonlocalH);
    if (a)
    {
      app_log() << " Found non-local Hamiltonian element named " << includeNonlocalH << std::endl;
      KE_Ham->addOperator(a, includeNonlocalH);
    }
    else
      app_log() << " Did not find non-local Hamiltonian element named " << includeNonlocalH << std::endl;
  }
  return KE_Ham;
}
```

In any case, the test is still showing some leaks. I'll keep investigating and make sure it's not a false positive or something in the underlying software stack. Thanks.

ye-luo:
@williamfgc I fixed some crash and PRed on your fork.
My preference is closing the operators instead of shared_ptr. @markdewing Please comment on the question on extractFixedHamiltonianComponents from @williamfgc.

markdewing:
Regarding `QMCCostFunctionBatched::extractFixedHamiltonianComponents()`

The `H_KE_Node` instance of QMCHamiltonian is a copy of a subset of the components in the main QMCHamiltonian instance. I think it would work to make copies (via `makeClone`) of the OperatorBase elements.  


There's a similar code pattern in QMCCostFunction, but it won't work as expected to make a copy.   There is a call to `H.setRandomGenerator` that also sets the RNG for the components of `H_KE_Node` as well.  If a copy is made, this could be fixed by adding a call to `H_KE_Node.setRandomGenerator`.

prckent:
For the newcomers here, it might be worth mentioning that the cost function is part of the wavefunction optimizer. This is a different set of use cases and requirements from the regular QMC drivers (variational monte carlo etc.)

williamfgc:
@markdewing @prckent @PDoakORNL @ye-luo thanks for all the input on this PR. @ye-luo was able to make `makeClone` return `unique_ptr` while addressing the current leak and test failures. I'm closing this PR.

williamfgc:
Superseeded by #3199

PR #3127, opened 2021-04-22
Update XL build recipe.

ye-luo:
## Proposed changes
Recent changes use more C++14 on the CUDA side and I have difficulty to teach the host compiler of nvcc using newer C++ so switch nvcc host compiler from XL to gcc when XL is used on Summit.

- Build related changes

- No

Sunmit

## Checklist


prckent (approved):
Reasonable change

LGTM

Issue #3134, opened 2021-04-25
Default DMC warmupsteps is 200 instead of 0 as documentation says


ye-luo:
**Describe the bug**
When warmupsteps is not given in the input.
source code set 200 but the doc says 0.
200 is bad as it is a hidden magic number.
0 is bad if user does not diligently set a desired good value.

The ultimate way will be having warmupsteps automatically determined but this is beyond the discussion of this issue.

The purpose of this issue is to have a discussion and pick either 0 or 200. Please comment.

prckent:
If we know that we are in the first DMC section, can we make this a required input?

Issue #3082, opened 2021-04-07
Rapid population explosion in batched GPU DMC on Summit


jtkrogel:
**Describe the bug**

In a select few cases, the batched DMC drivers crash hard on Summit.  The observed cases constitute 5 out of 2100 runs performed with VMC and DMC drivers.   The same cases complete successfully with the cpu version of the batched drivers on CADES (all other 2095 cases also complete successfully there).  All 5 cases run rapidly on a single Summit node (6 mpi tasks per node, 7 threads per task).  The failures are reproducible and all inputs use some version of T-moves.

The cases at hand are summarized by the following keys:
```
diamond_s14_tg231_sp1_comp_J3_ob_dmc_w66_tm0_batched
diamond_s14_tg111_sp1_real_J3_ob_dmc_w66_tm0_batched
diamond_s14_tg111_sp2_comp_J3_ob_dmc_w66_tm3_batched
diamond_s14_tg111_sp2_real_J3_ob_dmc_w66_tm0_dla_batched
graphene_s206_tg231_sp2_comp_J0_ob_dmc_w66_tm0_batched
```

The random seed for each run is determined by a hash of each respective key.

To decode the meaning of these inputs:
```
diamond_s14_tg111_sp2_real_J3_ob_dmc_w66_tm0_dla_batched

system           = diamond
size             = 14 atoms
twist_grid       = 1x1x1
nspin            = 2
qmcpack_build    = real
jastrow          = J3
lr_handler       = optimized breakup
qmc              = dmc
walkers_per_rank = 6
nonlocalmoves    = v0
dla              = yes
driver           = batched
```

**To Reproduce**

1. Copy the files from the following path on Summit: /gpfs/alpine/proj-shared/mat151/jtkrogel/ecp_roadtest/test_runs/bug_reproduction/set01
2. Remove the old qmcpack outputs from each directory
3. In each directory, replace the orbital href in each input file to point at the local copy of pwscf.pwscf.h5
4. In each directory, update the path to the qmcpack executable in dmc.bsub.in and submit

QMCPACK version: git hash dcf2dc9 (current on 31 Mar 2021).

**Expected behavior**
The runs should complete successfully w/o hard failures.

**System:**
 - system name: summit
 - modules loaded:
```
module load gcc/8.1.1
module load spectrum-mpi
module load cmake
module load git
module load cuda
module load essl
module load netlib-lapack
module load hdf5
module load python/3.6.6-anaconda3-5.3.0
[...]

PR #3135, opened 2021-04-25
Fix wrong trial energy in batched dmc when warmupsteps=0

ye-luo:
## Proposed changes
The first commit in this PR address the 0 Etrial reported at https://github.com/QMCPACK/qmcpack/issues/3082#issuecomment-826116420
In the main stage after warmup,
The branching cutoff window center is calculated as the running average of population energy.
In batched driver, this was done as
```
Eref = EnergyHist.mean();
// Eref is zero at step 0 when warmupsteps=0 as EnergyHist has do previous datapoint
// it ends up setting Etrial = 0 for step 1.
EnergyHist(Enow); // Enow is the population average of the current step.
```
Changed to
```
EnergyHist(Enow);
Eref = EnergyHist.mean();
```
The second commit hardens the logic of handling Trial energy in warmup stage and main stage in DMC batched driver.

To summarize, during warmup
Etrial is updated based on Enow as Enow is still exponentially decaying into a steady state. It is bad to collect and use EnergyHist. Using EnergyHist.mean() potentially causes huge delay in population control.

in main statge, Enow fluctuates and collecting EnergyHist and use EnergyHist.mean() for Eref makes sense.

This PR also splits the use of ToDoSteps into WarmUpToDoSteps and EtrialUpdateToDoSteps to avoid confusion.

Putting [WIP] as we need to 
 - [x] update deterministic tests reference values.
~~Sort out warmup step default value 200 but the documentation says 0. #3134~~

- Bugfix

- No

Bora

## Checklist


ye-luo:
This PR focuses on solving warmupsteps=0 case.
This is ready for review.

PDoakORNL (approved):
This is more clear behavior and documentation than the treatment of warmup steps and trial energy has ever enjoyed.  It will help with future refactoring of branching that will hopefully add some unit tests of the behavior paths as well.

prckent (approved):
`This is more clear behavior and documentation than the treatment of warmup steps and trial energy has ever enjoyed.`. Indeed.

PR #3136, opened 2021-04-26
Introduce ENABLE_OMP_TASKLOOP for enable/disable omp taskloop

ye-luo:
## Proposed changes
NVHPC 21.3 (released after 3.11 release) stops the compilation when it hits "omp taskloop" with an error saying this feature is not supported.
To make the compiler still being able to build and test the code. Introduce this flag.

- Build related changes

- No

epyc-server

## Checklist


CMake/PGICompilers.cmake
```diff
@@ -1,3 +1,7 @@
+if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 21.3 AND ENABLE_OMP_TASKLOOP)
```
prckent, line 1: Why CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 21.3 ? Does 21.2 not abort or somehow implement the feature?

A more cmake-like and robust solution would be to test for the feature availability and not test for any specific compiler version. i.e. Test compile a have_omp_taskloop.cpp code, set a configure variable. Then the code will naturally start working when 22.0 comes along, or whichever version supports it. Keep ENABLE_OMP_TASKLOOP for manual testing and workarounds.

prckent (changes requested):
See comment above.

I prefer to test for feature availability as well as have the flag.

CMake/PGICompilers.cmake
ye-luo: Good point on testing the feature. Will change.

ye-luo: Added a feature test. 21.2 doesn't abort. I don't think the feature is implemented but probably no-op in 21.2. Once there is a bit of tasking (probably due to target nowait) in the runtime, no-op is wrong for taskloop. My workaround is setting taskgroup.

ye-luo:
@prckent is this good now?

prckent (approved):

PR #3144, opened 2021-04-28
Update input for batch optimizer

markdewing:
Separate the XML parsing from the classes that do the work.

The individual optimizers will need to have their input changed to the same style.  Pass an xml node pointer for now.

This drops the "gpu" attribute, which wasn't connected anyway.

- No

laptop

## Checklist


src/QMCDrivers/WFOpt/QMCWFOptFactoryNew.cpp
```diff
@@ -24,9 +24,11 @@ QMCOptimizeBatched* QMCWFOptFactoryNew(xmlNodePtr cur,
   qmcdriver_input.readXML(cur);
   VMCDriverInput vmcdriver_input;
   vmcdriver_input.readXML(cur);
+  WFOptDriverInput wfoptdriver_input;
+  wfoptdriver_input.readXML(cur);
 
   QMCOptimizeBatched* opt = new QMCOptimizeBatched(project_data, w, std::move(qmcdriver_input),
-                                                   std::move(vmcdriver_input), std::move(pop), samples, comm);
+                                                   std::move(vmcdriver_input), std::move(wfoptdriver_input), std::move(pop), samples, comm);
```
ye-luo, line 31: Does it make sense of include the VMC input object as part of Opt driver input. I don't mean flatten VMC input object but it is a natural hierarchy.

markdewing: Maybe - I agree it does form a natural hierarchy.  However, the only real connection is the number of samples that get generated by the VMC part - so I can also see keeping the two inputs at the same level. The hierarchy isn't very deep - it seems like an easy change that could be made later.

ye-luo: Thanks for the feedback.

PDoakORNL: I mostly care that the XML is sensible and the hierarchical relation is clear there. There should be a clear ownership of xmlnodes and delegation of children of a node.  The legacy situation where both the VMCDriver and the Optimizer driver read from the same <qmc method=...> node and the loop is outside is terrible.
``` xml
 <qmc_optimize method="linear" loop max="6">
   <parameter name="MinMethod">    OneShiftOnly </parameter>
   <parameter name="minwalkers">           1e-4 </parameter>
    <!-- forget what was decided in the arguement about samples parameters -->
   <qmc method="vmcbatch" move="pbyp" gpu="yes">
     <parameter name="total_walkers">                1 </parameter>
     <parameter name="warmupSteps">            5 </parameter>
     <parameter name="blocks">                25 </parameter>
     <parameter name="timestep">             1.0 </parameter>
     <parameter name="usedrift">              no </parameter>
     <estimator name="LocalEnergy" hdf5="no"/>
   </qmc>
</qmc_optimize>
```

This PR need not address this, but I think eventually the readXML in the optimizer input should delegate to the VMCDriverInput when it finds a the vmc driver node inside.  It should then either
A. Have a composed VMCDriverInput that it calls readXML on.
B. Return a VMCDriverInput it constructed and has called readXML on.

A. seems simpler in the short term.

PDoakORNL (commented):
This is a useful first step

PDoakORNL (approved):
So I approve contingent on the mixed <qmc method="opt_batch"> xml node being eventually sorted into a a clear hierarchy where the outside node either reads every child itself or delegates child nodes in their entirety. We should not have an ongoing situation where the VMCDriverInput and WFOptDriverInput both paw over all the child nodes of <qmc> each reading some of them.

src/QMCDrivers/WFOpt/QMCWFOptFactoryNew.cpp
markdewing: I agree that the VMC and optimizer parameters should be separate and not mixed in the same XML section.   That will change the input format, which will require changes in Nexus to support, etc.   And so will need more design and review.

PR #3145, opened 2021-04-29
Add deterministic test for backflow optimization and VMC

### LINKED ISSUE #3103: Deterministic backflow test needed
Looking at the current CI coverage results, adding a deterministic backflow test based on the statistical one would catch a many files that are currently completely missed.

Hyeondeok-Shin:
## Proposed changes

 Close #3103 

 This PR is to add deterministic test for backflow optimization parameters and VMC runs.
 check_opt.py is updated to check backflow parameters in addition to Jastrows, and location is moved to /tests/scripts.

- No

 Intel Xeon, intel 19 compiler

## Checklist


prckent:
OK to test

Hyeondeok-Shin:
In failed CI, I noticed that optimization parameters are not deterministic now. I'll check it tomorrow.

prckent:
Is the vmc backflow determinsitic (no optimization)? If the runs are not deterministic it might be indicating a bug. Could neighbor lists not be correctly handled in backflow?

Hyeondeok-Shin:
Only VMC with backflow is fine. The issue is in only the optimization in Jastrow or backflow. Only Jastrow has the same issue but current deterministic test couldn't detect as it is using a single parameter. I'll make separated VMC test for the backflow only. I also talked with @ye-luo for this issue, and he will play with the optimization test with many number of parameters to see the problem when the test is merged with the small number of parameters. I will rework for them.

Hyeondeok-Shin:
I added VMC deterministic test for backflow, and fixed optimization test. I expect they will pass CI but need to see.

@ye-luo In the input, it seems there are 3 triggers making optimization test non-deterministic.
1. Increase size of backflow parameters to 3.
2. optimize backflow and Jastrow at the same time.
3. optimized Jastrow only but increase size to 3.

prckent (approved):
LGTM

prckent:
Notice that this increases our coverage by 1.16% :-)

PR #3153, opened 2021-05-04
Nexus: support grand-canonical twist averaging

Paul-St-Young:
## Proposed changes

Enable nexus to setup grand-canonical twist average grid using DFT Fermi energy and KS eigenvalues.
Activated when an occupation dependency is give to `generate_qmcpack`, e.g. `(p2q, 'gc_occupation')`

Tested using the attached BCC Li example:
[bcc-li-n16.zip](https://github.com/QMCPACK/qmcpack/files/6424040/bcc-li-n16.zip)

- New feature

- No

Fedora 32, E5-2620 v2, gcc 10.2.1, intel 19.1.2

## Checklist

### DIFF
--- a/nexus/lib/qmcpack.py
+++ b/nexus/lib/qmcpack.py
@@ -547,6 +547,38 @@ class Qmcpack(Simulation):
                         fobj.close()
                     #end if
                 #end for
+                grand_canonical_twist_average = 'nelecs_at_twist' in self
+                if grand_canonical_twist_average:
+                    for itwist, qi in enumerate(input.inputs):
+                        elecs = self.nelecs_at_twist[itwist]
+                        # step 1: resize particlesets
+                        nup = elecs[0]
+                        ndn = elecs[1]
+                        qi.get('u').set(size=nup)
+                        qi.get('d').set(size=ndn)
+                        # step 2: resize orbital sets
+                        sb = qi.get('sposet_builder')
+                        bb = sb.bspline  # hard-code for Bspline orbs
+                        assert itwist == bb.twistnum
+                        sposets = bb.sposets
+                        if len(sposets) == 1:  # RHF/ROHF
+                            norb = max(nup, ndn)
+                            sposets[list(sposets.keys())[0]].set(size=norb)
+                        else:  # UHF
+                            for spo in sposets:
+                                ispin = spo.get('spindataset')
+                                nelec = elecs[ispin]
+                                spo.set(size=nelec)
+                            #end for
+                        #end if
+                        # step 3: resize determinants
+                        dset = qi.get('determinantset')
+                        sdet = dset.slaterdeterminant  # hard-code single det
+                        for det, nelec in zip(sdet.determinants, elecs):
+                            det.set(size=nelec)
+                        #end for
+                    #end for
+                #end if
             #end if
         #end if
     #end def write_prep
--- a/nexus/lib/qmcpack.py
+++ b/nexus/lib/qmcpack.py
@@ -566,26 +566,24 @@ class Qmcpack(Simulation):
                         ndn = elecs[1]
                         qi.get('u').set(size=nup)
                         qi.get('d').set(size=ndn)
-                        # step 2: resize orbital sets
+                        # step 2: resize determinants
+                        dset = qi.get('determinantset')
+                        sdet = dset.slaterdeterminant  # hard-code single det
+                        spo_size_map = {}
+                        for det, nelec in zip(sdet.determinants, elecs):
+                            spo_name = det.get('sposet')
+                            spo_size_map[spo_name] = nelec
+                            det.set(size=nelec)
+                        #end for
+                        # step 3: resize orbital sets
                         sb = qi.get('sposet_builder')
                         bb = sb.bspline  # hard-code for Bspline orbs
                         assert itwist == bb.twistnum
                         sposets = bb.sposets
-                        if len(sposets) == 1:  # RHF/ROHF
-                            norb = max(nup, ndn)
-                            sposets[list(sposets.keys())[0]].set(size=norb)
-                        else:  # UHF
-                            for spo in sposets:
-                                ispin = spo.get('spindataset')
-                                nelec = elecs[ispin]
-                                spo.set(size=nelec)
-                            #end for
-                        #end if
-                        # step 3: resize determinants
-                        dset = qi.get('determinantset')
-                        sdet = dset.slaterdeterminant  # hard-code single det
-                        for det, nelec in zip(sdet.determinants, elecs):
-                            det.set(size=nelec)
+                        for spo in sposets:
+                            if spo.name in spo_size_map:
+                                spo.set(size=spo_size_map[spo.name])
+                            #end if
                         #end for
                     #end for
                 #end if
--- a/nexus/lib/qmcpack.py
+++ b/nexus/lib/qmcpack.py
@@ -384,6 +384,28 @@ class Qmcpack(Simulation):
             else:
                 self.error('incorporating wavefunction from '+sim.__class__.__name__+' has not been implemented')
             #end if
+        elif result_name=='gc_occupation':
+            from qmcpack_converters import gcta_occupation
+            if not isinstance(sim,Pw2qmcpack):
+                msg = 'grand-canonical occupation require Pw2qmcpack'
+                self.error(msg)
+            #endif
+            # step 1: extract Fermi energy for each spin from nscf
+            if len(sim.dependencies) != 1:
+                msg = 'need scf/nscf calculation for Fermi energy'
+                self.error(msg)
+            #end if
+            nscf = sim.dependencies[1].sim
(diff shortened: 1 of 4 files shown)


nexus/lib/qmcpack.py
```diff
@@ -525,6 +547,38 @@ def write_prep(self):
...
+                        nup = elecs[0]
+                        ndn = elecs[1]
+                        qi.get('u').set(size=nup)
+                        qi.get('d').set(size=ndn)
+                        # step 2: resize orbital sets
+                        sb = qi.get('sposet_builder')
+                        bb = sb.bspline  # hard-code for Bspline orbs
+                        assert itwist == bb.twistnum
+                        sposets = bb.sposets
+                        if len(sposets) == 1:  # RHF/ROHF
```
jtkrogel, line 564: This won't work if there are other sposets in the builder (e.g. density matrices).  You should iterate over the determinants, and then pull out and resize only the referenced sposets from the builder.

nexus/lib/qmcpack_converters.py
```diff
@@ -185,6 +185,91 @@ def generate_pw2qmcpack_input(prefix='pwscf',outdir='pwscf_output',write_psir=Fa
...
+#end def import_nexus_module
+
+
+def read_eshdf_eig_data(filename, Ef_list):
+    import numpy as np
+    from numpy import array,pi
+    from numpy.linalg import inv
+    from unit_converter import convert
+    try:
+        hdfreader = import_nexus_module('hdfreader')
```
jtkrogel, line 200: What case does this style of import help with?

nexus/lib/qmcpack_converters.py
```diff
@@ -185,6 +185,91 @@ def generate_pw2qmcpack_input(prefix='pwscf',outdir='pwscf_output',write_psir=Fa
...
+        return array(i,dtype=int)[0]
+    #end def h5int
+
+    h        = read_hdf(filename,view=True)
+    axes     = array(h.supercell.primitive_vectors)
+    kaxes    = 2*pi*inv(axes).T
+    nk       = h5int(h.electrons.number_of_kpoints)
+    ns       = h5int(h.electrons.number_of_spins)
+    if len(Ef_list) != ns:
+      msg = 'Ef "%s" must have same length as nspin=%d' % (str(Ef_list), ns)
```
jtkrogel, line 217: Why not just `error()` here instead of returning string?

nexus/lib/qmcpack.py
```diff
@@ -384,6 +384,28 @@ def process_jastrow(wf):
...
+            if len(sim.dependencies) != 1:
+                msg = 'need scf/nscf calculation for Fermi energy'
+                self.error(msg)
+            #end if
+            nscf = sim.dependencies[1].sim
+            na = nscf.load_analyzer_image()
+            Ef_list = na.fermi_energies
+            # step 2: analyze ESH5 file for states below Fermi energy
+            pa = sim.load_analyzer_image()
+            pa.analyze(Ef_list=Ef_list)
```
jtkrogel, line 403: This step might get repeated many times and get heavy (multiple calculations depending on Pw2qmcpack for gcta).  Please check first for presence of 'wfh5' in pa, then proceed with analyze and save if not present.

if 'wfh5' not in pa:
    pa.analyze(Ef_list=Ef_list)
    sim.save_analyzer_image(pa)
#end if

with `save_analyzer_image` added to class `Simulation` as

def save_analyzer_image(self,analyzer):
    analyzer.save(os.path.join(self.imresdir,self.analyzer_image))
#end def save_analyzer_image

jtkrogel (changes requested):

jtkrogel:
@Paul-St-Young thanks for adding this.

nexus/lib/qmcpack.py
```diff
@@ -384,6 +384,28 @@ def process_jastrow(wf):
...
+            if not isinstance(sim,Pw2qmcpack):
+                msg = 'grand-canonical occupation require Pw2qmcpack'
+                self.error(msg)
+            #endif
+            # step 1: extract Fermi energy for each spin from nscf
+            if len(sim.dependencies) != 1:
+                msg = 'need scf/nscf calculation for Fermi energy'
+                self.error(msg)
+            #end if
+            nscf = sim.dependencies[1].sim
```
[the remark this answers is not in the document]
Paul-St-Young: gotcha, will fix in next push

nexus/lib/qmcpack.py
Paul-St-Young: yes, resizing determinant first, then resize only SPOs used is the better way to go
will fix in next push

nexus/lib/qmcpack_converters.py
Paul-St-Young: idk. I simply copied from `eshdf`. Is it only needed there and not here?

nexus/lib/qmcpack_converters.py
Paul-St-Young: This stand-alone routine `gcta_occupation` is not part of any class, so does not have access to the `error` function.
Does it make sense to make it part of `Pw2qmcpackAnalyzer`?

nexus/lib/qmcpack.py
Paul-St-Young: yes, good point.
will cache `pa` in next push

nexus/lib/qmcpack_converters.py
Paul-St-Young: *`read_eshdf_eig_data` is stand-alone

nexus/lib/qmcpack_converters.py
jtkrogel: In that context is useful to ensure execution of binaries when someone has not taken the step to install Nexus.  I think a normal import is fine here.

nexus/lib/qmcpack_converters.py
jtkrogel: You can just import `error` from `developer`.  The class level error function wraps this one.

nexus/lib/qmcpack.py
```diff
@@ -525,6 +557,36 @@ def write_prep(self):
...
+                        # step 1: resize particlesets
+                        nup = elecs[0]
+                        ndn = elecs[1]
+                        qi.get('u').set(size=nup)
+                        qi.get('d').set(size=ndn)
+                        # step 2: resize determinants
+                        dset = qi.get('determinantset')
+                        sdet = dset.slaterdeterminant  # hard-code single det
+                        spo_size_map = {}
+                        for det, nelec in zip(sdet.determinants, elecs):
```
jtkrogel, line 573: `sdet.determinants` is dict-like and does not guarantee ordering.  The match for up/down needs to be done differently.

jtkrogel (changes requested):

nexus/lib/qmcpack_converters.py
Paul-St-Young: cool fix in next push

nexus/lib/qmcpack_converters.py
Paul-St-Young: gotcha fix in next push

nexus/lib/qmcpack.py
Paul-St-Young: Assuming each `<determinant>` has a `group` attribute, this matching could be done.
Please see next push.

jtkrogel: Looks OK now

jtkrogel:
Do all tests pass with `nxs-test`?

jtkrogel:
Is there a spin-polarized case you could try this on to check the functionality there, e.g. FM iron?  (just input generation, not the qmcpack run)

Paul-St-Young:
good call for a check!
I modified the Li example with some spin imbalance.
[s2-bcc-li-n16.zip](https://github.com/QMCPACK/qmcpack/files/6429872/s2-bcc-li-n16.zip)

There was a problem with Fermi energy reading, but everything else checks out.

jtkrogel:
OK, great.  Do all tests pass with nxs-test?

Paul-St-Young:
Yes, they all pass
```
 66/68 example_qmcpack_diamond.................   Passed  3.72 sec
 67/68 example_qmcpack_graphene................   Passed  2.28 sec
 68/68 example_qmcpack_oxygen_dimer............   Passed  2.10 sec

100% tests passed, 0 tests failed out of 68

Total test time = 76.22 sec
```

I should say that many fail if `mpirun` is not in my `PATH`.
Is this an intended requirement?
Perhaps `mpirun` can be circumvented by replacing all `machine='ws1'` with `machine='ws1',serial=True`.

jtkrogel:
Serial runs are rather the exception than the rule and the successful completion of tests (or not) is meant to convey whether your system is configured to drive executables via Nexus.

jtkrogel (approved):
LGTM now.

prckent:
OK to test

PR #3156, opened 2021-05-06
Format cmake using cmake-format

ye-luo:
## Proposed changes
Skipped formatting all commented lines

- Build related changes
- 
- No

Epyc-server

## Checklist


prckent:
Try https://github.com/cheshirekow/cmake_format ?

ye-luo:
> Try https://github.com/cheshirekow/cmake_format ?

The tool provides the feature that puts all commands in lower case. However, I could not find a way to turn off other features, a few places got messy, especially comments.

ye-luo:
Applied cmake-format without external codes.
```
find -name "CMakeLists.txt" -exec cmake-format -i {} \;
find -name "*.cmake" -exec cmake-format -i {} \;
```
It is not fully want I want but it is acceptable.

ye-luo:
This now formatted with cmake-format.

prckent:
I am happy with these formatting changes. The lower case keywords take some adjusting to but I see that the style is widely used elsewhere. I noticed that you set the linewidth consistent with our clang-format.

prckent (approved):
LGTM

Waiting to see if any comments are received via Slack discussion ahead of merge. If none significant received, will merge at/after 6pm EST today

PDoakORNL:
I think it would be good to document the cmake style in a future PR.  Otherwise I think this is well worth the conflicts its going to cause me.

prckent:
Agree, a future PR needs to add a mention of the formatter and style (nearly the default) in our docs. There is a long list of config options at https://cmake-format.readthedocs.io/en/latest/configuration.html and interesting case studies at https://cmake-format.readthedocs.io/en/latest/format-case_studies.html . Since it is automatic we can readily experiment as we find "bad" formatting.

ye-luo:
Thare two parts.
1. formatting. I didn't explore much the formatting options. .cmake-format should have all the rules set.
2. style. There is a cmake-tidy similar to clang-tidy. So we may just opt-in.
If we don't want to rely on human. We need to further use clang-tidy and cmake-tidy for good software practice.

PR #3157, opened 2021-05-07
Improve ppconvert

zenandrea:

## Proposed changes

Fix the conversion of casino format pseudos to xml format.
The behaviour was to take the cutoffs from the "summary.txt" file.
This file is no more provided on the new Trail and Needs pseudos (i.e., the CEPP and eCEPP).
Also, the cutoffs given there could be too short (for instance in the case of hydrogen).
In this change the behaviour is the following:
if the "summary.txt" file is not provided, the cutoffs are set to the point corresponding to an 
agreement of 1e-6 among the potentials in all the channels. 

- Bugfix
- New feature

- No

Conversion of first row atoms

## Checklist

- Not sure. This PR is up to date with current the current state of 'develop'


prckent:
OK to test

ye-luo:
Better to add a test to cover casino pp to xml

prckent (changes requested):
Just marking as changes requested consistent with the above discussion. Merge after addition of a test of the improved CASINO format conversion.

prckent:
I looked at quickly adding the casino test, but it will require atomic wavefunctions https://qmcpack.readthedocs.io/en/develop/additional_tools.html#listing-72  or more time than I have with the atomic wavefunction solver. I don't think it wise to use untested parameters since someone will copy them. @zenandrea if you have time to make a test please do so in a later PR. Merging this one as is to keep things moving.

prckent (approved):

PR #3160, opened 2021-05-10
Rename misleading test file

ye-luo:
## Proposed changes
Just to avoid confusion.

- No

Bora

## Checklist


shivupa:
Yup this naming makes more sense and after discussions with @markdewing and @ye-luo I don't think I should be touching this file anyways. Thanks!

PR #3164, opened 2021-05-11
Make open BC IonIon working in legacy_cuda.

ye-luo:
## Proposed changes
Enable spline SPO, open BC on legacy CUDA.
Only need to send the already pre-computed numbers out to scalar.dat
The pw2qmcpack workflow test has an atomO OBC case.
@zenandrea also confirmed that it works on his side.

- New feature

- No

Epyc-server

## Checklist

### DIFF
--- a/src/QMCHamiltonians/CoulombPotential.h
+++ b/src/QMCHamiltonians/CoulombPotential.h
@@ -16,15 +16,19 @@
 
 #ifndef QMCPLUSPLUS_COULOMBPOTENTIAL_H
 #define QMCPLUSPLUS_COULOMBPOTENTIAL_H
-#include "Particle/ParticleSet.h"
-#include "Particle/WalkerSetRef.h"
-#include "Particle/DistanceTableData.h"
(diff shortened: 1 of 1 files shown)


ye-luo:
> I have made QE available to the legacy CUDA nightly builds on sulfur & nitrogen, so the Oatom tests should be added on cdash sometime tomorrow.

I failed to find a CUDA variant. https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=qe-atomO-polarized-no-collect-np-16-0-16-nk-1-0-1-atomO_pp-vmc_noj-1-16-kinetic&date=2021-05-13

prckent:
Testset wildcard updated for legacy cuda just now. Hopefully tomorrow. Could run by hand now if you are concerned.

ye-luo:
> Testset wildcard updated for legacy cuda just now. Hopefully tomorrow. Could run by hand now if you are concerned.

I ran it before PR. It was good.

PR #3169, opened 2021-05-12
Remove CUDA contamination which breaks offload.

ye-luo:
## Proposed changes

fix offload non-cuda build. We will add CI to catch this.

- Bugfix

- No

epyc-server

## Checklist


PDoakORNL (approved):
This is my mistake.  This fix LGTM

Issue #3173, opened 2021-05-13
test_ion_derivs not working with offload


ye-luo:
**Describe the bug**
Offload turns on batched NLPP but MSD doesn't have corresponding implementations.
However even if I turn on batched NLPP, I still got crazy numbers, so need to have a look.

**To Reproduce**
1. add test_ion_derivs test for ENABLE_OFFLOAD.

**Expected behavior**
At least fallback should work.

PR #3174, opened 2021-05-13
Remove test_ion_derivs when offloading

ye-luo:
## Proposed changes
Disable the test when offload is on. See more details in #3173
fix OMP taskloop check in cmake

With this PR, we can run all the unit tests when offload to x86

- Yes

epyc-server

## Checklist


PDoakORNL (approved):
LGTM

PR #3177, opened 2021-05-13
Add MSD evaluateRatios and mark Fast tag as deleted

ye-luo:
## Proposed changes
1. Implement batched NLPP in MSD.
2. mark "Fast" tag as deleted
3. NLPP input tag changes. Input value "default" is removed. No input given actually means default. Add "non-batched" to explicitly call the non-batched code path.

- New feature
- Documentation changes

- Yes. but I don't think algorithm="default" is actually used anywhere but development only

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
+++ b/src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
@@ -637,7 +637,7 @@ void DiracDeterminantBatched<DET_ENGINE>::evaluateRatiosAlltoOne(ParticleSet& P,
     Phi->evaluateValue(P, -1, psiV_host_view);
   }
   for (int i = 0; i < psiMinv.rows(); i++)
-    ratios[FirstIndex+i] = simd::dot(psiMinv[i], psiV.data(), NumOrbitals);
+    ratios[FirstIndex + i] = simd::dot(psiMinv[i], psiV.data(), NumOrbitals);
 }
 
 
--- a/src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
+++ b/src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
@@ -636,8 +636,8 @@ void DiracDeterminantBatched<DET_ENGINE>::evaluateRatiosAlltoOne(ParticleSet& P,
     ScopedTimer local_timer(SPOVTimer);
     Phi->evaluateValue(P, -1, psiV_host_view);
   }
-  //FIXME due to padding in psiMinv
-  MatrixOperators::product(psiMinv, psiV.data(), &ratios[FirstIndex]);
+  for (int i = 0; i < psiMinv.rows(); i++)
+    ratios[FirstIndex+i] = simd::dot(psiMinv[i], psiV.data(), NumOrbitals);
 }
 
 
@@ -664,7 +664,7 @@ typename DiracDeterminantBatched<DET_ENGINE>::GradType DiracDeterminantBatched<D
   {
     resizeScratchObjectsForIonDerivs();
     Phi->evaluateGradSource(P, FirstIndex, LastIndex, source, iat, grad_source_psiM);
-    // psiMinv.cols() can be different from NumOrbitals
+    // psiMinv columns have padding but grad_source_psiM ones don't
     for (int i = 0; i < psiMinv.rows(); i++)
       g += simd::dot(psiMinv[i], grad_source_psiM[i], NumOrbitals);
   }
--- a/src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
+++ b/src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
@@ -664,8 +664,9 @@ typename DiracDeterminantBatched<DET_ENGINE>::GradType DiracDeterminantBatched<D
   {
     resizeScratchObjectsForIonDerivs();
     Phi->evaluateGradSource(P, FirstIndex, LastIndex, source, iat, grad_source_psiM);
-    // FIXME due padding of psiMinv
-    g = simd::dot(psiMinv.data(), grad_source_psiM.data(), psiMinv.size());
+    // psiMinv.cols() can be different from NumOrbitals
+    for (int i = 0; i < psiMinv.rows(); i++)
+      g += simd::dot(psiMinv[i], grad_source_psiM[i], NumOrbitals);
   }
 
   return g;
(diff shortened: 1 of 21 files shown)


src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
```diff
@@ -664,8 +664,9 @@ typename DiracDeterminantBatched<DET_ENGINE>::GradType DiracDeterminantBatched<D
   {
     resizeScratchObjectsForIonDerivs();
     Phi->evaluateGradSource(P, FirstIndex, LastIndex, source, iat, grad_source_psiM);
-    // FIXME due padding of psiMinv
-    g = simd::dot(psiMinv.data(), grad_source_psiM.data(), psiMinv.size());
+    // psiMinv.cols() can be different from NumOrbitals
```
PDoakORNL, line 667: This comment is a bit confusing.  Isn't the issue that psiMinv.cols() could be padded out because we want lda to be an aligned size?  My understanding is that this is what all the transposing in the cuBLAS and lapack is really about, we're taking a column major matrix with column padding but lapack interprets that as the transpose, i.e. that memory and padding is to it row major and the padding is on the "rows"

maybe
```
// psiMinv padding for SIMD row alignment is explicit i.e.
// psiMinv.cols() >= psiMinv.rows()  &&  psiMinv.cols() * sizeof(T) == is a multiple of QMC_SIMD_ALIGNMENT
// The actual Matrix with n = num_orbitals is stored in the square sublock psiMinv.rows() x psiMinv.rows().

File: src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp [hunk 2]
```diff
@@ -664,8 +664,9 @@ typename DiracDeterminantBatched<DET_ENGINE>::GradType DiracDeterminantBatched<D
...
+    for (int i = 0; i < psiMinv.rows(); i++)
+      g += simd::dot(psiMinv[i], grad_source_psiM[i], NumOrbitals);
```
PDoakORNL, line 669: The reason this is necessary is that grad_source_psiM does not have explicit padding and so doesn't have aligned rows, which seems like it could be a performance regression. SIMD::dot does not actually do the transpose of A but actually assumes A has been transposed.  If both A and B have the same row alignment then simple element wise product and accumulation is fine and likely optimal for vectorization. 

But if grad_source_psiM has unaligned rows by design and not omission then a unit test should capture this special behavior.  Because at face value taking an inner produce of a {P,N} matrix and a {N,N} matrix with P>N is nonsense.

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
```diff
@@ -189,14 +189,14 @@ void MultiDiracDeterminant::BuildDotProductsAndCalculateRatios(int ref,
...
-void MultiDiracDeterminant::evaluateDetsForPtclMove(ParticleSet& P, int iat)
+void MultiDiracDeterminant::evaluateDetsForPtclMove(const ParticleSet& P, int iat, int refPtcl)
 {
   UpdateMode = ORB_PBYP_RATIO;
   RatioTimer.start();
   evalOrbTimer.start();
   Phi->evaluateValue(P, iat, psiV);
   evalOrbTimer.stop();
-  WorkingIndex = iat - FirstIndex;
+  WorkingIndex = (refPtcl < 0 ? iat : refPtcl) - FirstIndex;
```
PDoakORNL, line 199: So what insures  
```
that
FirstIndex  is never > iat or refPtcl
```

I've seen this cause mayhem elsewhere.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
```diff
@@ -372,6 +371,42 @@ WaveFunctionComponent::PsiValueType MultiSlaterDeterminantFast::ratio(ParticleSe
   return curRatio;
 }
 
+void MultiSlaterDeterminantFast::evaluateRatios(const VirtualParticleSet& VP, std::vector<ValueType>& ratios)
```
PDoakORNL, line 374: So I guess there is no unit test?
New code should have unit tests.

PDoakORNL (changes requested):
Missing Tests and insufficient documentation of "interesting" behavior.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
ye-luo: The code is actually covered by deterministic tests diamondC_1x1x1-Gaussian_pp_MSD and diamondC_2x1x1-Gaussian_pp_MSD as I turned them on below for offload builds but codecov doesn't recognize it.

I agree that unit test will be helpful. I will expand test_multi_slater_determinant.

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
ye-luo: Good to have. Will add.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
ye-luo: Will add more verbose comment.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
ye-luo: psiMinv documentation added.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
ye-luo: psiMinv documentation added.

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
ye-luo: Added assert for all the similar places.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
ye-luo: Online a a few unit tests.

PDoakORNL: I wouldn't say it's "good" that codecov doesn't recognize the deterministic system tests as "coverage" but application level tests of features are not the same as unit test code. They do almost nothing to describe expected behavior of particular units of code. Thanks for adding these.

PDoakORNL (approved):

PR #3178, opened 2021-05-13
Add all-electron move spin-orbit VMC Driver

rcclay:

## Proposed changes
Add an all-electron SOVMC driver (without drift).  Having all-electron drivers on hand is nice for debugging new wave functions.  

- New feature

- No

## Checklist


ye-luo (approved):
LGTM

prckent:
1. Just in conversation, @rcclay , what benefit do you think we get from having separate all electron move drivers? I haven't edited/updated these codes directly, so there might be some things that are convenient as they are. However it does seem that unifying the drivers in future would be result in some simplification. Having the spin-move drivers separate from the non-spin orbit case makes much more sense, but they are also an avenue for bugs to crawl in.
2. Is there a reason that you did not implement drift at this point?

ye-luo:
Echo @prckent , we should use template tricks to incorporate spin moves in batched driver. No copy will be accepted.

ye-luo:
Also would like to see #2735 resolved soon.

rcclay:
@prckent Sorry for the long response time.
1.) I wouldn't use an all electron driver in production, but it's really nice to have around for development/debugging purposes.  Recent problem:  multi-determinants for spinor wave functions gave a wonky e-e energy.  WFTester has single configuration wave function and ratio tests to check for obvious bugs/inconsistencies in the wave function, but having an actual all-electron driver helps check if anything crazy is happening to the walker distributions. PbyP and all-electron should yield identical distributions via completely different code paths, but all-electron is much simpler code and relies on easier tested code.
2.) The reason that drift wasn't implemented is because the incorporation of spin degrees of freedom is clunky as is, and having to add the data structures and interfaces to deal with spin gradients wasn't worth the effort and clutter.  No drift already accomplishes the debugging purpose mentioned in point 1.

camelto2 (approved):
LGTM

prckent:
@rcclay Thanks for the answers. No question on the utility of all electron move drivers, just how to go about implementing them. Given the current situation in the "legacy CPU drivers" it seems not unreasonable to add this one, particularly since we need spin orbit debugger. However as also agreed by @ye-luo , when it comes to making the batched versions -- which we'll need to work on soon after spin-orbit is "working" -- we need a strategy to expand the existing batched drivers vs make wholly new ones. All electron moves are easier than spin-orbit and should be done first. I think even a naive switch{} would be better than a new driver, and we can likely do better than that with some templating.

prckent (approved):

PR #3181, opened 2021-05-15
Fix parameter derivatives in CI coeff

### LINKED ISSUE #3172: Parameter derivatives incorrect for CI coefficients
Parameter derivatives are incorrect for CI coefficients in a multideterminant wavefunction. The "test" optimizer output shows a large discrepancy between the analytic and numerical gradient values.  (This certainly also affects the linear method, as the matrices are built from the same ingredients as the parameter derivatives)

The regression was introduced in https://github.com/QMCPACK/qmcpack/pull/2947
specifically, commit https://github.com/QMCPACK/qmcpack/commit/769807541057d636b75e7faa0b917db2d5dd33f9

(Thanks to "git bisect" for making it straightforward to isolate the commit)

(The real problem is we didn't have a test to catch a regression like this)

It can be reproduced with the C2_pp test case.  In the wavefunction, set the "optimize" attribute to "yes" in the  "multideterminant" tag.  (Can also turn off optimization for the Jastrow factors to make the output clearer).  Use the "test" optimize method, as
```
    <qmc method="opt" move="pbyp" checkpoint="-1">
      <optimize method="test" output_param_file="yes"/>
     ...
```

First few lines of the analytic-numerical comparison
```
Param_Name  Value    Numeric            Analytic       Percent
CIcoeff_1 3.3152e-01  1.1184e-01  -2.6137e-02  1.2337e+02
CIcoeff_2 8.4180e-02  1.3111e-01  6.6619e-02  4.9189e+01
CIcoeff_3 8.4180e-02  -6.7154e-01  2.2867e-02  1.0341e+02
```

markdewing:
Fixes #3172 

The temporary variables q0 and v need to be defined in the right scope to get reset to zero.

Add unit testing that would catch the issue.

Change the signature of evaluateDerivatives in MultiSlaterDeterminant (not Fast) to also call that routine in complex builds.(Otherwise the generic version in WaveFunctionComponent gets called, and the derivatives are zero).

- Bugfix

- No

laptop

## Checklist


src/QMCWaveFunctions/tests/test_multi_slater_determinant.cpp
```diff
@@ -118,6 +121,27 @@ void test_LiH_msd(const std::string& spo_xml_string,
...
+  int nparam = active.size_of_active();
+  REQUIRE(nparam == 1486);
+
+  using ValueType = QMCTraits::ValueType;
+  std::vector<ValueType> dlogpsi(nparam);
+  std::vector<ValueType> dhpsioverpsi(nparam);
+  twf.evaluateDerivatives(elec_, active, dlogpsi, dhpsioverpsi);
+
+  // Numbers not validated
+  REQUIRE(dlogpsi[0] == ValueApprox(0.006449058893092842));
```
ye-luo, line 137: https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md#natural-expressions
FYI, I learnt that CHECK doesn't stop execution. I found it better suited for value checks.
REQUIRE seems better suited for critical sanity checks.
The above REQUIRE on nparam remains a good choice. If it goes wrong, the execution doesn't need to continue.

prckent: Useful - most all of our tests could be CHECKs

prckent (approved):
LGTM

PR #3183, opened 2021-05-17
Add script to build, test and install real and complex versions of QMCPACK at the same time

### LINKED ISSUE #3151: Build, test and install real and complex versions of QMCPACK at the same time

Currently there isn't an option to build, test and install real and complex versions of QMCPACK at the same time.

I created an example using CMake's externalproject_add to build with both -DQMC_COMPLEX=0 and -DQMC_COMPLEX=1. Ctest runs both unit tests, though passing arguments is proving to be challenging and may require a pull request upstream. Copying executables to the correct location appears straightforward.
https://github.com/quantumsteve/qmcpack_superbuild

Something similar may be possible using the superbuild infrastructure. 
https://gitlab.kitware.com/paraview/common-superbuild

Initially, I tried modifying the CMakelists.txt within QMCPACK. Building additional targets for real and complex would add additional complexity and the compile definitions means files still need to be compiled twice.

I also tried using CMAKE FetchContent but ran into configuration errors due to duplicate target names. Variables like `EXECUTABLE_OUTPUT_PATH` and  `LIBRARY_OUTPUT_PATH` must be the same in real and complex builds.

**Additional context**

quantumsteve:

## Proposed changes

This fixes #3151

This uses CMake's [externalproject_add](https://cmake.org/cmake/help/latest/module/ExternalProject.html) to build, test and install qmcpack with both -DQMC_COMPLEX=0 and -DQMC_COMPLEX=1. Any arguments passed on the command line are passed to both builds. To build, change `cmake [options] <path-to-source>` to `cmake [options] <path-to-source>/superbuild`.

Ctest is configured with two tests, one for the real build and another for the complex build. Passing arguments is proving to be challenging. By default unit tests are run. To change arguments set the environment variable `TEST_ARGUMENTS`

TESTS_ARGUMENTS="-R short -LE unstable" ctest --verbose

The install target copies all executables from the real build along with `qmcpack_complex` and `qmcpack_complex.settings` from the complex build.

- New feature
- Build related changes

- No

Ubuntu 20.04

## Checklist

- n/a. Code added or changed in the PR has been clang-formatted


quantumsteve:
I originally wrote this as a separate repository that fetched the qmcpack repo.
https://github.com/quantumsteve/qmcpack_superbuild

I think having everything in the same repo will make it easier to keep everything working and up-to-date.

CMake/RunTestsSuperbuild.cmake
```diff
@@ -0,0 +1,14 @@
+function(run_tests)
+  execute_process(COMMAND ${TEST_EXECUTABLE} ${ARGN} RESULT_VARIABLE result ECHO_OUTPUT_VARIABLE COMMAND_ECHO STDOUT)
+  if(NOT "${result}" STREQUAL "0")
+    message(FATAL_ERROR "Test failed with return value '${result}'")
+  endif()              
+endfunction()
+
+if(NOT DEFINED ENV{TESTS_ARGUMENTS})
+  set(TESTS_ARGUMENTS "-R unit")
```
ye-luo, line 9: Prefer matching label instead of name. So "-L unit"

prckent (commented):
I'll review this in a bit.

Q. Can we think of a more user friendly/accessible/clear name than superbuild?

prckent (changes requested):
(Marking as changes required but perhaps just a little polishing needed)

This worked for me. I initially specified QMC_COMPLEX=0 by habit/mistake and the configure correctly aborted since it is not meaningful for a unified build.

Builds were successful and ctest resulted in the unit tests being run. (Should change to the deterministic tests once this matures).

My main concern is about the install experience: by default this does not result in some installdir/bin/qmcpack and installdir/bin/qmcpack_complex etc. , and we don't give any guidance about specifying CMAKE_INSTALL_DIR. What is practical here? I like the idea of setting the install dir to builddir/qmcpack by default and then installing/copying the executables and eventually other support files (e.g. nexus) there. Also worth putting a reminder that the default install directory is being used if it is not specified.

The builds can run concurrently leading to confusing configuration output - can we set one of the builds as dependent on the other completing to serialize then?

Final question might be how we are going to test this in nightlies and configure test reports to go to cdash. All the while the total test output is summarized as a unique yes/no for the two builds, that will work since it will be different from all the other tests.

quantumsteve:
@prckent I tried to address your first two comments.

With respect to cdash, I think it would current do two submissions [accumulating the results?](https://cmake.org/pipermail/cmake/2011-May/044419.html). I could copy the CDASH information into unifiedbuild/CMakeLists.txt so the two upstream (real and complex build) tests also report, but doubt that's useful information.

prckent:
(I am in report writing mode currently, but poke me if you want/need this looked at. I don't think that there is much risk of a merge conflict arising.)

quantumsteve:
Closing in favor of #4099

PR #3185, opened 2021-05-18
Allow Coeff_0 for ground state CI coefficients

markdewing:
Some converters use the "Coeff" tag for the ground state in the HDF5 file, and some use "Coeff_0".  Try "Coeff_0" first, and then try "Coeff" if reading the ground state.

- Bugfix

- No

laptop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.h
+++ b/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.h
@@ -151,7 +151,8 @@ private:
 
     extVar=extVar+"_imag";
     if(!hin.readEntry(CIcoeff_imag, extVar))
-       app_log() << "Coeff_imag not found in h5. Set to zero." << std::endl;
+      if (ext_level != 0 || !hin.readEntry(CIcoeff_imag, "Coeff_imag"))
+        app_log() << "Coeff_imag not found in h5. Set to zero." << std::endl;
  
     for (size_t i = 0; i < n_dets; i++)
       ci_coeff[i] = VT(CIcoeff_real[i], CIcoeff_imag[i]);
--- a/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.h
+++ b/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.h
@@ -117,19 +117,15 @@ private:
                             (std::is_floating_point<VT>::value), int> = 0>
   void readCoeffs(hdf_archive& hin, std::vector<VT>& ci_coeff, size_t n_dets,int ext_level)
   {
-    ///Determinant coeffs are stored in Coeff for the ground state and Coeff_N 
-    ///for the Nth excited state. 
-    ///The Ground State is always stored in Coeff 
-    ///Backward compatibility is insured
+    ///Some converters store determinant coeffs in Coeff for the ground state
+    ///and some store it in Coeff_0.  The Nth excited state in is in Coeff_N.
     std::string extVar;
-    if (ext_level==0)
-       extVar="Coeff";
-    else
-       extVar="Coeff_"+std::to_string(ext_level);
+    extVar="Coeff_"+std::to_string(ext_level);
 
     if (!hin.readEntry(ci_coeff,extVar))
-       APP_ABORT("Could not read CI coefficients from HDF5");
-      
+      if (ext_level != 0 || !hin.readEntry(ci_coeff,"Coeff"))
+          APP_ABORT("Could not read CI coefficients from HDF5");
+
   }
   template<typename VT,
            std::enable_if_t<(std::is_same<VT, ValueType>::value) &&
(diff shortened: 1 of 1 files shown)


prckent:
Do all our current converters use a uniform tag? ( Hoping this is just for legacy support... )

anbenali:
Well, This is a bit ambarrasing as it is affecting nly newly generated HDF5 from QP as we were not sure what nomenclature to use and decided to revert to the Coeff for the ground state. 
In the meantime, I uploaded the H5edit.py tool that fixes the rare legacy files that were not caught by me but mainly, not generated by me. 
Not knowing who was using the multidet code, I assumed nobody was since I have not heard of issues (until now) and have been using an internal fix for generating H5 files with the correct naming. 
Since your PR is ready, I am fine with it even if in principle it should never need the fall back to Coeff_0... Is there a way you want to proceed?

markdewing:
It comes from the current QP2 HDF output.  The loop for generating the coefficient labels can be kept more regular if it uses "Coeff_0" instead of "Coeff".  
Accepting both variants seems like it would be okay, unless we want to be strict for the sake of consistency.

prckent:
What is the source of the "old" QP2 HDF output with coeff_0? Do we have some archived files? Do we have any of these files in the repo for testing?

@anbenali Just to confirm, no current converter writes coeff_0? We consistently use coeff everywhere?

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.h
```diff
@@ -117,19 +117,15 @@ class SlaterDetBuilder : public WaveFunctionComponentBuilder
...
-       extVar="Coeff";
-    else
-       extVar="Coeff_"+std::to_string(ext_level);
+    extVar="Coeff_"+std::to_string(ext_level);
 
     if (!hin.readEntry(ci_coeff,extVar))
-       APP_ABORT("Could not read CI coefficients from HDF5");
-      
+      if (ext_level != 0 || !hin.readEntry(ci_coeff,"Coeff"))
+          APP_ABORT("Could not read CI coefficients from HDF5");
```
ye-luo, line 127: myComm->barrier_and_abort is better.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.h
```diff
@@ -142,24 +138,22 @@ class SlaterDetBuilder : public WaveFunctionComponentBuilder
...
-    else
-      extVar="Coeff_"+std::to_string(ext_level);
-    
+    extVar="Coeff_"+std::to_string(ext_level);
+
 
     if(!hin.readEntry(CIcoeff_real, extVar))
-       APP_ABORT("Could not read CI coefficients from HDF5")
+      if (ext_level != 0 || !hin.readEntry(CIcoeff_real, "Coeff"))
+        APP_ABORT("Could not read CI coefficients from HDF5")
```
ye-luo, line 150: myComm->barrier_and_abort is better.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.h
```diff
@@ -142,24 +138,22 @@ class SlaterDetBuilder : public WaveFunctionComponentBuilder
...
 
     extVar=extVar+"_imag";
     if(!hin.readEntry(CIcoeff_imag, extVar))
-       app_log() << "Coeff_imag not found in h5. Set to zero." << std::endl;
-         
+      if (ext_level != 0 || !hin.readEntry(CIcoeff_imag, "Coeff_imag"))
+        app_log() << "Coeff_imag not found in h5. Set to zero." << std::endl;
```
ye-luo, line 155: This message is not accurate when `ext_level != 0`

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.h
```diff
@@ -142,24 +138,22 @@ class SlaterDetBuilder : public WaveFunctionComponentBuilder
...
+    ///Determinant coeffs are stored in Coeff_N where N is Nth excited state.
+    ///The Ground State is stored in Coeff or Coeff_0.
 
     std::string ext_var;
-    if (ext_level==0)
-      extVar="Coeff";
-    else
-      extVar="Coeff_"+std::to_string(ext_level);
-    
+    extVar="Coeff_"+std::to_string(ext_level);
```
ye-luo, line 145: It will helpful to run clang-format

ye-luo (changes requested):

anbenali:
@prckent Yes nothing in archive or anywhere is affected. This "issue" was introduced when we activated the excited states and needed to identify how to access the coefficients. At that point @ye-luo had requested a change in the labeling to avoid confusion and coeff_0 was eliminated in a subsequent pull request. all changes did not trickle down properly between one code to the other. The main reason was that all our archived data used Coeff and not Coeff_0 which would have forced the fall back. Only newly generated data was affected by my lack of follow up.

markdewing:
I'm going to close this PR.   The naming issue should get fixed from the QP2 side.

PR #3186, opened 2021-05-18
Fix ubuntu installation in documentation

aannabe:
## Proposed changes

Fix ubuntu installation in docs. Add missing `sudo` and `install` keywords.

- Documentation changes

- No

Ubuntu 20.04, personal laptop

## Checklist

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -681,8 +681,8 @@ production, a platform-optimized BLAS should be used.
 
 ::
 
-  apt-get cmake g++ openmpi-bin libopenmpi-dev libboost-dev
-  apt-get libatlas-base-dev liblapack-dev libhdf5-dev libxml2-dev fftw3-dev
+  sudo apt-get install cmake g++ openmpi-bin libopenmpi-dev libboost-dev
+  sudo apt-get install libatlas-base-dev liblapack-dev libhdf5-dev libxml2-dev fftw3-dev
   export CXX=mpiCC
   cd build
   cmake ..


ye-luo:
Okay to test

ye-luo (approved):
Thanks for catching this.

prckent:
@aannabe Can you confirm that at least all the deterministic tests are working for you? (ctest -L deterministic)

I tried this scheme in a fresh Virtualbox+Ubuntu 20.04 and had a couple of failures. Investigating.

aannabe:
2/36 failed in unit tests.
93/775 failed in deterministic tests.

I am not sure why there are so many failed tests. Both are attached.

[det.txt](https://github.com/QMCPACK/qmcpack/files/6504335/det.txt)
[unit.txt](https://github.com/QMCPACK/qmcpack/files/6504336/unit.txt)

prckent:
Interesting. I was able to get them all working after rerunning. Still investigating.

prckent:
Looks like you have a bad MPI. My guess is that any of 4-4 or 16-1 tests will also fail for you. Rerun with -VV or directly.

ye-luo:
`ctest -L deterministic --output-on-failure`

prckent:
Crash in MPC interaction setup, or the next thing after it. I'll explore more - clearly this is something users could hit . Perhaps there is some incompatible library, if not a bug.
```
 18/668 Test  #858: deterministic-diamondC_1x1x1_pp-dmc_sdj_excited-1-1 ...........................................***Failed  Required regular expression\
 not found. Regex=[QMCPACK execution completed successfully
c] 10.31 sec
Rank =    0  Free Memory =  1421 MB
  Input file(s): det_qmc_short_vmc_dmc_excited.in.xml

                    QMCPACK 3.11.9

       (c) Copyright 2003-  QMCPACK developers

                    Please cite:
 J. Kim et al. J. Phys. Cond. Mat. 30 195901 (2018)
      https://doi.org/10.1088/1361-648X/aab9c3

  Git branch: develop
  Last git commit: 589b0ccd2efa1b7356b0ce6712882df7b7bfbf04-dirty
  Last git commit date: Tue May 18 13:31:00 2021 -0500
  Last git commit subject: Merge pull request #3186 from aannabe/ubuntu_install
  Global options

  Total number of MPI ranks = 1
  Number of MPI groups      = 1
  MPI group ID              = 0
  Number of ranks in group  = 1
  MPI ranks per node        = 1
  OMP 1st level threads     = 1
  OMP nested threading disabled or only 1 thread on the 2nd level

  Base precision      = double
  Full precision      = double

  CPU only build
  Timer build option is enabled. Current timer level is coarse

  Input XML = det_qmc_short_excited.in.xml

  Project = det_qmc_short_excited
  date    = 2021-05-18 17:53:40 EDT
  host    = pkent-VirtualBox

...
 Using NonLocalECP potential
    Maximum grid on a sphere for NonLocalECPotential: 12
  QMCHamiltonian::addOperator NonLocalECP to H, physical Hamiltonian
QMCHamiltonian::addOperatorType added type pseudo named PseudoPot

  === Initializing MPC interaction ===
  Using 609 G-vectors for MPC interaction.
   Using real-space box of size [20,20,20] for MPC spline.
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpiexec noticed that process rank 0 with PID 0 on node pkent-VirtualBox exited on signal 9 (Killed).
--------------------------------------------------------------------------
```

aannabe:
I found that there are essentially two reasons for failure in my tests:
1. This one is trivial. I had `~/anaconda3/bin` in my `PATH` since I use it pretty extensively. There are MPI libraries in that path that are shipped with Anaconda. This can cause issues. After commenting out Anaconda from my path, the majority of tests passed.  Since Anaconda is very popularly used, a comment about this might be helpful in the documentation.
2.  There were still some errors in my tests (see attached openmpi_*). I then purged `openmpi` and tried `mpich` for MPI. This now resolved all test failures. I think another comment could be added which states that an alternative MPI library could be tried in case of test failures: `sudo apt-get install mpich mpich-doc`. 
See attached test outputs for details.

[openmpi_unit.txt](https://github.com/QMCPACK/qmcpack/files/6505037/openmpi_unit.txt)
[openmpi_det.txt](https://github.com/QMCPACK/qmcpack/files/6505038/openmpi_det.txt)
[mpich_unit.txt](https://github.com/QMCPACK/qmcpack/files/6505039/mpich_unit.txt)
[mpich_det.txt](https://github.com/QMCPACK/qmcpack/files/6505040/mpich_det.txt)

prckent:
@aannabe Yes we can strengthen the cautions and update the recommendations, good points.  Did you configure qmcpack with something closer to 'cmake ../qmcpack' or to 'cmake -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx' ? The latter should be safer since the wrappers should result in the correct libraries being used.

aannabe:
I used what is exactly shown in the documentation for testing purposes.  Agreed, wrappers are safer.

ye-luo:
@prckent that helps. However, unfortunately it is more complicated than that. Even if mpicc get the MPI libraries right at linking. anaconda pollutes LD_LIBRARY_PATH and an incompatible version of library may still get picked up from anaconda. I end up using miniconda on some systems or not loading anaconda all the time.

prckent:
Ultimately users have to setup their environment properly. We can give some pointers, reminders so that we are less likely to be blamed though.

PR #2926, opened 2021-02-16
[WIP] Multi-Slater Determinant  - Batched with Offload to accelerator

anbenali:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

The MultiSlater Determinant code is now fully batched with some kernels offloaded to accelerators using openmp offload. 
Some functions part of the table method were still too messy to offload completely and a change in algortihm to simplify is needed to get efficiency. Therefore this is considered as a first attempt and a preparation for the change in algorithm that we believe will be required to achieve performance.  

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Batched
- Accelerator offload
- New functionality

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
- Argonne JLSE - V100
## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- Yes This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #3194, opened 2021-05-21
Using Shared_ptr in MultiDiracDeterminant

ye-luo:
## Proposed changes
Shared_ptr is the best solution for shared dataset between WF object clones and prevents leaks.
More private and more const in MultiDiracDeterminant class.
Add a unit test to help me understand ci_configuration2.
Fix leaks in MultiSlaterDeterminant

- Bugfix

- No

epyc-server

## Checklist


prckent (commented):
Q. Can we get rid of  MultiDiracDeterminant.2.cpp now? Merge with the regular .cpp? It think these legacy multi-suffix files should either be merged with their parent.cpp or the structural-problem-in-plain-sight talked about, made an issue/fixed.

Q about additional cleanup: What is the reason why detValues, new_detValues etc. can not be moved to private? Which class is the (ab)user?

ye-luo:
> Generally LGTM
> 
> Q. Can we get rid of MultiDiracDeterminant.2.cpp now? Merge with the regular .cpp? It think these legacy multi-suffix files should either be merged with their parent.cpp or the structural-problem-in-plain-sight talked about, made an issue/fixed.
> 

I think it can be merged but not right now. I need to take care of the pending PR #2926 first. Actually it is on my list this week. I would like to tackle a few other things in the MSD class and pending PR is a pain.

> Q about additional cleanup: What is the reason why detValues, new_detValues etc. can not be moved to private? Which class is the (ab)user?

There are more fundamental issue with detValues which scales `exp(Nelec)`. Completely different variables will replace their existence as the algorithm changes.

prckent (approved):
Good point on that long pending PR

PR #3196, opened 2021-05-21
Spin logic simplification

camelto2:
## Proposed changes

This mostly address #2735 

In this PR, I have introduced a new boolean flag to ParticleSet called IsSpinor, which is used throughout the code to set appropriate spinor info. This simplifies the input burden on the user, and the only requirement is that if a user wants to use spinors (aside from providing a correctly converted hdf5 file to read), they only have to provide a `spinors="yes"` flag to particleset. 

The user still can change things in the QMCDrivers like the spin mass for spin sampling, but this is only needed in some circumstances and otherwise the input requirements are limited to the spinors=yes flag. 

Internally, when the SPOSetBuilders are being determined it will check this flag and call the appropriate spinor set (EinsplineSpinorSetBuilder or LCAOSpinorBuilder) as opposed to users having to specify "spinorbspline" or "molecularspinor". Additionally, when the QMCDrivers are being built, it checks the targetPtcl.IsSpinor flag and appropriately chooses the spin movers (SOVMCUpdatePbyP or SODMCUpdatePbyPFast) instead of the user having to specify `spinMoves = yes` in the QMC input. 

The documentation and various unit tests have been appropriately changed.

I will note that particle set still always stores the spin variables. This can be changed in another PR where we only allocate and store spins when required (e.g. unique_ptr<ParticleScalar_t> for spins that is allocated when IsSpinor = true). I wanted to keep this PR as small as possible so that can be handled at a later date. This at least accomplishes the task of simplifying the input. 

Once merged, this flag will be used to simplify some of the changes proposed in #3188. 

- No

my workstation, gnu 9.2.0

## Checklist


ye-luo:
Test this please

ye-luo (approved):

PR #3203, opened 2021-05-27
Update docker images for CI

williamfgc:
## Proposed changes

Add vim and serial hdf5 dev in docker images (already published in DockerHub) used for GitHub Actions CI.

- Other (please describe): Docker containers

- No

Ubuntu 20.04, CI will pull these containers

## Checklist

### DIFF
--- a/config/docker/dependencies/ubuntu/openmpi/Dockerfile
+++ b/config/docker/dependencies/ubuntu/openmpi/Dockerfile
@@ -18,6 +18,7 @@ RUN export DEBIAN_FRONTEND=noninteractive &&\
                     git \
                     libopenmpi-dev \
                     libhdf5-openmpi-dev \
+                    libhdf5-serial-dev \
                     hdf5-tools \
                     libfftw3-dev \
                     libopenblas-openmp-dev \
@@ -27,6 +28,7 @@ RUN export DEBIAN_FRONTEND=noninteractive &&\
                     rsync \
                     wget \
                     software-properties-common \
+                    vim \
                     -y
 
 # Python packages for tests
--- a/config/docker/dependencies/ubuntu/clang-latest/Dockerfile
+++ b/config/docker/dependencies/ubuntu/clang-latest/Dockerfile
@@ -23,6 +23,7 @@ RUN export DEBIAN_FRONTEND=noninteractive &&\
                     rsync \
                     wget \
                     software-properties-common \
+                    vim \
                     -y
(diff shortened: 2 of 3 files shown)


williamfgc:
@ye-luo this is only updating Dockefiles, the containers are updated in DockerHub. Also, the `sudo` password for user is `user` if you need to install additional packages using `apt` other than vim. Users can always run as root, but it's a better compromise to run as non-root with sudo.

config/docker/dependencies/ubuntu/openmpi/Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN export DEBIAN_FRONTEND=noninteractive &&\
                     git \
                     libopenmpi-dev \
                     libhdf5-openmpi-dev \
+                    libhdf5-serial-dev \
```
ye-luo, line 21: The serial one is libhdf5-dev

ye-luo:
> @ye-luo this is only updating Dockefiles, the containers are updated in DockerHub. Also, the `sudo` password for user is `user` if you need to install additional packages using `apt` other than vim. Users can always run as root, but it's a better compromise to run as non-root with sudo.

Great. I was fighting for the password for user yesterday... Could you add line in the docker documentation about this?

I assume you will update the docker image and the CI setting. right?

config/docker/dependencies/ubuntu/openmpi/Dockerfile
prckent: I like the clarity of the serial nature explicitly in the name and think it should stay. Any reason not to use this one?

ye-luo: I was aware of the existence of libhdf5-serial-dev as I look at packages at synaptic.
If I use apt, it exists.
```
Note, selecting 'libhdf5-dev' instead of 'libhdf5-serial-dev'
```
So let us keep 'serial'

ye-luo (approved):

williamfgc:
> I assume you will update the docker image and the CI setting. right?

@ye-luo good point. Yes, I usually test locally and upload to DockerHub the updated containers prior to doing a PR here to further tests them on GitHub Actions.

williamfgc:
@ye-luo I just added a line in the docs

ye-luo:
Test this please

ye-luo (approved):

PR #3206, opened 2021-05-28
Add script to convert gamess ECP to QMCPACK format 

mcbennet:
## Proposed changes

Converts gamess gaussian PPs to .xml format.

- New feature

- No

Laptop. Linux.

## Checklist


jtkrogel:
I recommend dropping the ".py" suffix on the file.  We are making an executable and it doesn't matter the language it is implemented in. 

A simpler input syntax that supports batching would also be nice `convert-gamess-ecp-to-qmcpack ./some/path/to/*.gamess`.  This also simplifies the script:
```
for filepath in sys.argv[1:]:
    path,filename = os.path.split(filepath)
    filename = filename.replace('.gms','').replace('.gamess','')+'.xml'
    pp = GaussianPP(filepath,format='gamess')
    pp.write(os.path.join(path,filename))
#end for
```

utils/convert-gamess-ecp-to-qmcpack
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
+
+import sys,os
+sys.path.append(os.path.join(os.path.dirname(__file__), '../nexus/lib'))
```
jtkrogel, line 4: In case someone moves the exe and tries to use it, perhaps add this guard:

```
nexus_lib = os.path.join(os.path.dirname(__file__), '../nexus/lib')
sys.path.append(nexus_lib)
try:
    from pseudopotential import GaussianPP
except:
    print('Nexus must be installed to run this script.\nExpected Nexus library at {}'.format(nexus_lib))
    sys.exit(1)
#end try
```
The script will run without error when either 1) the script has not been moved, 2) Nexus has been installed.  This at least tells them what to do if they have made this mistake instead of just having an `ImportError` printed.

mcbennet: Thank you, Jaron. This looks good.

mcbennet: Commit pushed. All 3 cases work as expected:
1. unmoved exe works successfully
2. moved exe with Nexus installed works successfully
3. mved exe without Nexus installed fails with verbose message "Nexus must be installed ..."

jtkrogel:
Do you have any other additions planned?  Otherwise this LGTM.

utils/convert-gamess-ecp-to-qmcpack
```diff
@@ -0,0 +1,26 @@
...
+    sys.exit(1)
+#end try
+
+executable_name = sys.argv[0]
+if len(sys.argv)<2:
+    print('Usage:\n{} <gamess-input-pp>'.format(executable_name))
+    sys.exit(2)
+#end if
+
+from pseudopotential import GaussianPP
```
jtkrogel, line 19: This line is redundant now.

mcbennet:
> Do you have any other additions planned? Otherwise this LGTM.

I have some concern about the output filename, actually. For instance, though I understand this is en edge case, perhaps the user prefers verbose filenames and has an informative string telling them about the contents of the file (one such example being 'O.bfd-gms.gms').

What about:
1. replacing .gms or .gamess with .xml if it appears at the end but leave other instances alone.  
2. Allow custom filename by user?
3. Just append .xml to whatever the input filename is

utils/convert-gamess-ecp-to-qmcpack
mcbennet: Fixed

jtkrogel:
`O.bfd-gms.gms` would be written out as `O.bfd-gms.xml` by the current script.

An extra cautious approach would be something like:
```
if '.' in filename:
    fn,ext = filename.rsplit('.',1)
    if ext in ('gms','gamess'):
        filename = fn
    #end for
#end if
filename += '.xml'
```

mcbennet:
WIP removed

jtkrogel (approved):
LGTM

prckent:
Test this please

ye-luo (approved):

PR #3209, opened 2021-05-31
Resolve MSD table method numerical instability

ye-luo:
## Proposed changes
In the MSD implementations
MSD(R) = \sum_i C_i * det^up_i(R) * det^dn_i(R)
This value grows as exp(Ne) as individual determinant grows the same way.
The old implementation uses MSD(R) in a few places and cased NaN due to over or underflow of float point numbers.
The new implementation fixes by using
MSD'(R)= C_ref + \sum_(i!=ref) C_i * det^up_i(R) * det^dn_i(R) / (det^up_ref(R) * det^dn_ref(R))
which is not vulnerable to over/underflow.

This should fixes https://groups.google.com/g/qmcpack/c/Yb3mVKBzH0Q 

- Bugfix

- No

epyc-server

## Checklist


prckent (changes requested):
I couldn't find sposet_collection in our docs. Please explain the what/where/why etc.

ye-luo:
> I couldn't find sposet_collection in our docs. Please explain the what/where/why etc.

it is interchangeable with `sposet_builder` right now which can be used to build SPOs outside determinant tag. Calling `builder` is just to techinical. When using the following style
```
<sposet_collection ....>
  <sposet name="a" ....>
  <sposet name="b" ....>
</sposet_collection>
```
a set of sposets of the same type are built. Thus, I prefer calling it `sposet_collection`

prckent:
Thanks for the explanation. We'll need to update the docs as `sposet_collection` is solidified.

prckent (approved):

prckent:
LGTM. Good improvement. I'll note we have a deterministic test on MSD wavefunctions run in the CI, so this code is tested. We should keep an eye on optimization tests in the nightlies.

ye-luo:
I did all the development work based on unit tests and then everything works out-of-box. I enjoyed the dev experience this time. I did test optimization in `short`.

PR #3212, opened 2021-06-01
Fix batch optimizer issue with accepting moves

### LINKED ISSUE #3211: Batched optimizer rejects parameter moves when run in parallel
When the batched version of the linear optimizer runs in parallel (MPI ranks > 1), it rejects more parameter moves than the old implementation.  This results in very slow convergence.

The linear method computes a new set of variational parameters. Correlated sampling is used to check the cost function at the new set of parameters.  If the cost function becomes NaN, or if the effective number of samples drops below a threshold ("minwalkers" in the input), the parameter move is rejected.  Rejected moves then start over collecting new samples at the old parameters.  This leads to slow convergence.

To check the output, the key phrases are
"The new set of parameters is not valid. Revert to the old set!"
or
"The new set of parameters is valid. Updating the trial wave function!"

It is especially noticeable in offload mode on Summit and for larger systems.  For NiO 192 electrons, it did not happen much.  But the 384 electron system (and higher), the effect was clear.

markdewing:
Fixes #3211

In QMCCostFunctionBatched::correlatedSampling, there is code to trim extreme weights.  The weight calculation uses "inv_n_samples" during accumulation.  That should use the global number of samples (the number of samples across all MPI ranks), not the local number of samples.

Add the global number of samples to the SampleStack class.

API design commentary:
Adding the number of ranks to setMaxSamples is intended to keep the object consistent between the number of samples and the global number of samples. Inconsistent state is possible in the case of a separate function to set the ranks or global number of samples.

However, setMaxSamples also allocates storage, and the number of ranks has nothing to do with that.  From that perspective, setting the number of ranks/global number of samples should be a separate function.

- Bugfix

- No

Summit and Andes

## Checklist

- Partially This PR adds tests to cover any new code, or to catch a bug that is being fixed


ye-luo:
I found similar pattern `inv_n_samples` in the code and we should move away from it.
It is better to accumulate the number of samples in-place and in the same way to accumulate other statistics which will be averaged. I'm wondering if you can explore this direction.

prckent:
I think we would be wise to add a few deterministic multiple-rank optimizer tests to catch/track/stop-reappearing these  normalization issues. The 'old ways' have proven error prone.

prckent (approved):

ye-luo:
@markdewing if you prefer merging this first and then add test + explore the safer route, please open an issue to track this.

markdewing:
@ye-luo I can look into computing the global number of samples.   It would require an extra allreduce call for a scalar.  But that shouldn't take much time.  Plus it handles the case where the number of samples may not be the same on every node.

ye-luo:
> @ye-luo I can look into accumulating the global number of samples. It would require an extra allreduce call for a scalar. But that shouldn't take much time. Plus it handles the case where the number of samples may not be the same on every node.

I don't think one more reduction really matters but if it is your concern, just use a local vector and do the allreduction.

PDoakORNL:
If it at all possible pass it through mpi in a typesafe way without obfuscating the variable name, there are mpi optimizations in the code which produce potentially unsafe vector accesses and type conversions between integral and floating point types to avoid defining an MPI structure and/or making an additional async transfer.

ye-luo:
> If it at all possible pass it through mpi in a typesafe way without obfuscating the variable name, there are mpi optimizations in the code which produce potentially unsafe vector accesses and type conversions between integral and floating point types to avoid defining an MPI structure and/or making an additional async transfer.

Could you put up an issue with a pointer to the unsafe source code? Not quite get which one you pointed.

PDoakORNL:
Yes this should be an issue, will do.

prckent:
So we don't lose track, please @markdewing , @PDoakORNL  create issues mentioned above before we merge.

PR #3217, opened 2021-06-03
Address leak related to functor in JastrowBuilder

williamfgc:
## Proposed changes

Replace raw pointer with `std::unique_ptr` in `J3Unique` container as it produced a leak caught by `asan/lsan`.
Make `eeI_JastrowBuilder` a class and move `putkids` to a private scope.
This is a minor PR only affecting `deterministic-unit_test_wavefunction_jastrow`. It doesn't reinstate the test as there a more direct leaks.

- Bugfix: leak fix

- No

Ubuntu 20.04 , must pass CI

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
+++ b/src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
@@ -21,6 +21,7 @@
 #include "CPU/SIMD/algorithm.hpp"
 #include <map>
 #include <numeric>
+#include <memory>
 
 namespace qmcplusplus
 {
@@ -69,7 +70,7 @@ class JeeIOrbitalSoA : public WaveFunctionComponent
   ///container for the Jastrow functions
   Array<FT*, 3> F;
 
-  std::map<std::string, FT*> J3Unique;
+  std::map<std::string, std::unique_ptr<FT>> J3Unique;
   //YYYY
   std::map<FT*, int> J3UniqueIndex;
 
@@ -137,9 +138,9 @@ public:
           typename std::map<const FT*, FT*>::iterator fit = fcmap.find(F(iG, eG1, eG2));
           if (fit == fcmap.end())
           {
-            FT* fc = new FT(*F(iG, eG1, eG2));
+            std::unique_ptr<FT> fc = std::make_unique<FT>(*F(iG, eG1, eG2));
+            fcmap[F(iG, eG1, eG2)] = fc.get();
             eeIcopy->addFunc(iG, eG1, eG2, fc);
-            fcmap[F(iG, eG1, eG2)] = fc;
           }
         }
     // Ye: I don't like the following memory allocated by default.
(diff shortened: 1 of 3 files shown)


src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
```diff
@@ -137,9 +138,9 @@ class JeeIOrbitalSoA : public WaveFunctionComponent
           typename std::map<const FT*, FT*>::iterator fit = fcmap.find(F(iG, eG1, eG2));
           if (fit == fcmap.end())
           {
-            FT* fc = new FT(*F(iG, eG1, eG2));
+            std::unique_ptr<FT> fc = std::make_unique<FT>(*F(iG, eG1, eG2));
```
ye-luo, line 141: auto?

src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
```diff
@@ -196,16 +197,16 @@ class JeeIOrbitalSoA : public WaveFunctionComponent
 
   void initUnique()
   {
-    typename std::map<std::string, FT*>::iterator it(J3Unique.begin()), it_end(J3Unique.end());
+    typename std::map<std::string, std::unique_ptr<FT>>::iterator it(J3Unique.begin()), it_end(J3Unique.end());
```
ye-luo, line 200: auto? Two separate lines.

src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
```diff
@@ -214,7 +215,7 @@ class JeeIOrbitalSoA : public WaveFunctionComponent
     }
   }
 
-  void addFunc(int iSpecies, int eSpecies1, int eSpecies2, FT* j)
+  void addFunc(int iSpecies, int eSpecies1, int eSpecies2, std::unique_ptr<FT>& j)
```
ye-luo, line 218: rvalue

src/QMCWaveFunctions/Jastrow/eeI_JastrowBuilder.cpp
```diff
@@ -45,13 +52,13 @@ bool eeI_JastrowBuilder::putkids(xmlNodePtr kids, J3type& J3)
...
       rAttrib.put(kids);
       typedef typename J3type::FuncType FT;
-      FT* functor        = new FT(ee_cusp, eI_cusp);
-      functor->iSpecies  = iSpecies;
-      functor->eSpecies1 = eSpecies1;
-      functor->eSpecies2 = eSpecies2;
-      int iNum           = iSet.findSpecies(iSpecies);
-      int eNum1          = eSet.findSpecies(eSpecies1);
-      int eNum2          = eSet.findSpecies(eSpecies2);
+      std::unique_ptr<FT> functor = std::make_unique<FT>(ee_cusp, eI_cusp);
```
ye-luo, line 55: auto?

ye-luo (changes requested):

src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
ye-luo: After a bit study, I think value is better than rvalue to express the clear ownership.

williamfgc: @ye-luo the way it's consumed right now requires an `lvalue` as we need to pass the underlying pointer to a temporary map (fcmap).

```
    auto fc = std::make_unique<FT>(*F(iG, eG1, eG2));
    fcmap[F(iG, eG1, eG2)] = fc.get();
    eeIcopy->addFunc(iG, eG1, eG2, fc);
```

Since `rvalue` is not an option we can do by value + `std::move`. What do you think?

src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
williamfgc: I'll refactor to use a for each loop

src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
ye-luo: I don't see why rvalue is not an option. fcmap takes the value of the raw pointer and has nothing to do with smart pointers.
Yes, value + move is the preferred way.

williamfgc: So fcmap will be populated from the raw pointer directly? If that's the case, rvalue is an option.

ye-luo:
Test this please

ye-luo (approved):

PR #3219, opened 2021-06-03
Fix NVHPC OpenMP offload compilation and suppress a few warnings

ye-luo:
## Proposed changes
1. NVHPC doesn't support user defined reduction(UDR). In this PR, CMake first test UDR. If it doesn't work try implicit reduction of complex, if it still doesn't work, stop.
2. Suppress a few PGI warnings.

NVHPC 21.5 now compiles fine but still has issues at run time.

The warnings are not PGI specific, we suppressed them in other compilers. @williamfgc good targets for software improvements.

- Build related changes

- No

epyc-server

## Checklist

### DIFF
--- a/src/Platforms/OMPTarget/CMakeLists.txt
+++ b/src/Platforms/OMPTarget/CMakeLists.txt
@@ -2,7 +2,7 @@
 #// This file is distributed under the University of Illinois/NCSA Open Source License.
 #// See LICENSE file in top directory for details.
 #//
-#// Copyright (c) 2020 QMCPACK developers.
+#// Copyright (c) 2021 QMCPACK developers.
 #//
 #// File developed by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
 #//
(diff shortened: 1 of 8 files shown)


src/Platforms/OMPTarget/CMakeLists.txt
```diff
@@ -22,5 +22,35 @@ ELSE()
...
+else()
+  # Test if C++ compiler supports OpenMP user defined reduction on complex type
+  try_compile(OMP_UDR_COMPLEX_OKAY ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/test_user_defined_reduction_complex.cpp
+              CMAKE_FLAGS "${CMAKE_CXX_FLAGS}"
+              OUTPUT_VARIABLE COMPILE_OUTPUT)
+
+  if (NOT OMP_UDR_COMPLEX_OKAY)
+    set(COMPILE_FAIL_OUTPUT omp_UDR_complex_fail.txt)
+    file(WRITE "${CMAKE_BINARY_DIR}/${COMPILE_FAIL_OUTPUT}" "${COMPILE_OUTPUT}")
+    message(STATUS "OpenMP user-defined reduction functionality compilation check failed!"
```
prckent, line 35: WARNING and not STATUS?

prckent (changes requested):
I think the compilation check failures should be reported as WARNINGs. Otherwise LGTM.

src/Platforms/OMPTarget/CMakeLists.txt
ye-luo: Because the next step makes an attempt to try the implicit way. So it is a recoverable failure and always print when using NVHPC. So I prefer not to print a scary warning.

prckent (approved):

prckent:
Test this please

PR #3234, opened 2021-06-09
Fewer catch_main

ye-luo:
## Proposed changes
Compiling catch_main is slow. So better to keep fewer catch_main variants.
catch_benchmark_main is removed as catch_main turns on benchmark by default.

Overall compilation time change is noticeable but it becomes more friendly to partial build.

- Build related changes

- No

epyc-server

## Checklist


ye-luo:
I did measure on unit tests other than the one actually using benchmark. The compilation time increase is in the noise.

PR #3236, opened 2021-06-10
Cleanup tests run with QMC_CUDA + vla fix

prckent:
## Proposed changes

Cleanup which tests are run with QMC_CUDA enabled. Tests are either skipped or set to expect fail where we have appropriate error handling. Use message_verbose to limit cmake output consistently.

Also fix a vla warning in non-local eval. Moved vector outside the potential loop due to noticing an initial few percent slowdown in small tests.

- Bugfix

- No

spock using cuda2hip version.

## Checklist


ye-luo (commented):
Apart from my in-source review comment.  I put up an clean up task in #3237 which can be picked up by free hands.

src/QMCHamiltonians/NonLocalECPotential_CUDA.cpp
```diff
@@ -278,8 +280,10 @@ void NonLocalECPotential_CUDA::addEnergy(MCWalkerConfiguration& W, std::vector<R
       }
 #endif
       RatioList.resize(QuadPosList.size());
-      RealType vrad[pp.nchannel];
-      RealType lpol[pp.lmax + 1];
+      if (vrad.size()<pp.nchannel)
```
ye-luo, line 283: This `if` is not necessary, resize performance the same amount of allocation/deallocation and resizing to smaller size cost nothing.

prckent: Updated.

ye-luo (approved):

PR #3235, opened 2021-06-10
Fix memory leaks in deterministic-unit_test_new_drivers

quantumsteve:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

Describe what this PR changes and why.  If it closes an issue, link to it here
with [a supported keyword](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

This fixes all direct and indirect leaks found in deterministic-unit_test_new_drivers by clang address sanitizer.

This is part of #3083.

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Bugfix
- Refactoring (no functional changes, no api changes)

### Does this introduce a breaking change?

- No

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- n/a. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- n/a. Documentation has been added (if appropriate)

PR #3238, opened 2021-06-10
Mark most LRHandler evaluate functions const and clean up its usage in Coulomb

ye-luo:
Review after merging #3235

## Proposed changes
Before this PR, shared_ptr are used for long range handler and short range pair functions in Coulomb PBC AA class. In fact when particle sets are active, the shared_ptr is used as unique_ptr. The AB class use unique_ptr.

In this PR, all the evaluation functions of  long range handler and short range pair functions are defined const (thread-safe).
So a shared copy of long range handler and short range pair function can be used among all the walkers safely. Both makeClone functions become calling compiler generated copy constructor.

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/LongRange/EwaldHandler.cpp
+++ b/src/Particle/LongRange/EwaldHandler.cpp
@@ -106,7 +106,7 @@ void EwaldHandler::fillFk(KContainer& KList)
   app_log().flush();
 }
 
-EwaldHandler::mRealType EwaldHandler::evaluate_vlr_k(mRealType k)
+EwaldHandler::mRealType EwaldHandler::evaluate_vlr_k(mRealType k) const
 {
   mRealType uk = 0.0;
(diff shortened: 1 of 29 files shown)


prckent (approved):
LGTM

I think Ray is tied up until next week. Wait or merge as convenient?

ye-luo:
@prckent I added a few more override. Need another approval. All the changes are relatively low risk. We can merge as convenient. I subscribed @rcclay is more of sending him a notification.

prckent:
An override too many / A virtual function too few according to the CI

ye-luo:
> An override too many / A virtual function too few according to the CI

fixed now. Looks like a bug was surfaced but it doesn't affect anything as the function is disabled anyway.

prckent (approved):

PR #3240, opened 2021-06-11
Add DTModes which represents switches used in distance tables.

ye-luo:
## Proposed changes
DTModes records operations switches required by distance table consumers.
Use strongly typed enum.
Intended to avoid unnecessary D2H transfer when offloading Jastrow factors.
The avail 'modes' are not final. Will adjust as the implementation goes.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/DTModes.h
+++ b/src/Particle/DTModes.h
@@ -1,0 +1,49 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2021 QMCPACK developers.
+//
+// File developed by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//
+// File created by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+#ifndef QMCPLUSPLUS_DTMODES_H
+#define QMCPLUSPLUS_DTMODES_H
+
+#include <cstdint>
+
+namespace qmcplusplus
+{
+enum class DTModes : uint_fast8_t
+{
+  ALL_OFF = 0x0,
+  /** whether full table needs to be ready at anytime or not
+   * Optimization can be implemented during forward PbyP move when the full table is not needed all the time.
+   * DT consumers should know if full table is needed or not and request via addTable.
+   */
+  NEED_FULL_TABLE_ANYTIME   = 0x1,
+  NO_NEED_TEMP_DATA_ON_HOST = 0x2
+};
+
+constexpr bool operator&(DTModes x, DTModes y)
+{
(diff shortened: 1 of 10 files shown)


src/Particle/DistanceTableData.h
```diff
@@ -17,12 +17,13 @@
 #define QMCPLUSPLUS_DISTANCETABLEDATAIMPL_H
 
 #include "Particle/ParticleSet.h"
+#include <limits>
+#include <bitset>
```
PDoakORNL, line 21: I think since you went the enum route bitset isn't needed.

src/Particle/DTModes.h
```diff
@@ -0,0 +1,49 @@
...
+{
+enum class DTModes : uint_fast8_t
+{
+  ALL_OFF = 0x0,
+  /** whether full table needs to be ready at anytime or not
+   * Optimization can be implemented during forward PbyP move when the full table is not needed all the time.
+   * DT consumers should know if full table is needed or not and request via addTable.
+   */
+  NEED_FULL_TABLE_ANYTIME   = 0x1,
+  NO_NEED_TEMP_DATA_ON_HOST = 0x2
```
prckent, line 28: NO_NEED_TEMP_DATA_ON_HOST needs a proper definition (comment) here when you start using it.

ye-luo: it is currently not well defined. It is added just to show the design of this enum. Will probably rename it with actual definition added.

src/Particle/DTModes.h
```diff
@@ -0,0 +1,49 @@
...
+};
+
```
PDoakORNL, line 30: Since this is idiom and not straight forward some documentation is needed.
I don't think its very obvious that:
`DTModes dtmodes = static_cast<DTModes>(0x1 | 0x2)`
is ok.  I'm not sure if here is the place for the documentation or we want to have a section in the developer guide where we setout how we want to implement bitmasked options in the code going forward.

PDoakORNL (approved):
Basically fine but lets decide how to handle documentation. If not in this PR make an issue.

src/Particle/DTModes.h
PDoakORNL: Or add a unit test that shows the ways the enum class can be used. This would protect against someone replacing DTModes with another implementation that didn't work the same.

ye-luo: No one should write ` DTModes dtmodes = static_cast<DTModes>(0x1 | 0x2)`
`DTModes dtmodes = DTModes::NEED_FULL_TABLE_ANYTIME | DTModes::NO_NEED_TEMP_DATA_ON_HOST` is the right way.

src/Particle/DistanceTableData.h
ye-luo: removed

src/Particle/DTModes.h
ye-luo: Unit test added.

ye-luo:
Test this please

prckent (approved):
Approving to make progress, but more docs are still needed.

Issue #2699, opened 2020-09-10
Require C++ 17


prckent:
**Is your feature request related to a problem? Please describe.**

Enable+require use of C++17 after the v3.10.0 release and ahead of the v4 release. By the "two year" support rule this is in scope and sufficiently supported by a broad range of compilers.

**Describe the solution you'd like**

Modify cmakelists to require this language

Drop XL compiler if it still only supports C++14. Use Clang offload support on Summit.

**Describe alternatives you've considered**

**Additional context**

PR #3243, opened 2021-06-12
Merge DiffOneBodyJastrowOrbital.h into J1OrbitalSoA.h 

shivupa:
## Proposed changes

This PR addresses some maintenance work suggested by @ye-luo for One body jastrows. This merges the two files mentioned in the title, and uses `unique_ptr`. I used #3217 as a guide.

- No

Laptop i7-7700HQ

## Checklist


shivupa:
I noticed some things that need changed on this PR. I'll update here when I finish

shivupa:
Fixed

prckent:
Thanks. Will wait for @ye-luo comments.

ye-luo:
heads up. It takes a bit longer than expected to review this. I found a quite a few issues not necessarily directly caused by this PR but surfaced.

shivupa:
No worries no rush, just let me know if there is anything I should do to help.

src/QMCWaveFunctions/Jastrow/J1OrbitalSoA.h
```diff
@@ -59,11 +71,19 @@ struct J1OrbitalSoA : public WaveFunctionComponent
   aligned_vector<int> DistIndice;
   Vector<posT> Grad;
   Vector<valT> Lap;
-  ///Container for \f$F[ig*NumGroups+jg]\f$
+  ///Container for \f$F[ig*NIons+jg]\f$
   std::vector<FT*> F;
+  ///container for the unique Jastrow functions
+  std::vector<std::unique_ptr<FT>> J1Unique;
```
ye-luo, line 77: J1Unique replace the old role of F. However, F was redefined. For this reason, all the F in the the derived class OneBodyJastrowOrbitalBspline.cpp should be updated.

ye-luo:
@shivupa I fixed the legacy CUDA build but the actual code needs to be updated. Pull before you continue.

ye-luo:
Test this please

shivupa:
Rebased on develop and ready for review! Thanks @ye-luo

src/QMCWaveFunctions/Jastrow/J1OrbitalSoA.h
```diff
@@ -140,8 +172,8 @@ struct J1OrbitalSoA : public WaveFunctionComponent
       const auto& displ = d_ie.getDisplRow(iel);
       for (int iat = 0; iat < Nions; iat++)
       {
-        int gid    = Ions.GroupID[iat];
-        auto* func = F[gid];
+        int gid    = Ions.getGroupID(iat);
+        auto* func = J1Unique[gid].get();
         if (func != nullptr)
```
PDoakORNL, line 177: How can a I get a func that is a null pointer?  Only if I didn't addfunc for that group so I get no it calls get on an out of bounds J1Unique so no. It could happen if J1Unique[gid] was added as a nullptr which seems like nonsense as well. If for some reason an ion type had no jastrow function a noOpFunction or Optional seems a safer way to handle this.

PDoakORNL (changes requested):
the nullptr early bails are quite a code smell.  I assume somewhere you don't have to add input to define a functor for each pair or triple and this is a lazy and potentially error prone way to deal with that. sdt::Optional or a no-op functor see a better way to deal with this. It doesn't look right the the J1 jastrow at all.

PDoakORNL:
The test does not check the very likely case that for multiple ions and "electron" species some of the Jastrows are not defined in input.  From my reading of the code this really needs to be part of the testing of these classes.

prckent:
^^^ Good point on testing for the missing in input case, and not just for the code in this PR. Most times this would be an error but it could also be intentional. The code should handle all cases.

shivupa:
Thanks for the review I'll start working on these!

ye-luo:
Test this please

ye-luo:
The current GPU CI failure is actually caused by develop branch. The legacy CUDA build with ASAN enabled already fails. This is the reason why I put up https://github.com/QMCPACK/qmcpack/issues/3280 yesterday.

src/QMCWaveFunctions/Jastrow/J1OrbitalSoA.h
ye-luo: To me, if an ion specie doesn't have a Jastrow, the functor should be nullptr. Adding noOpFunction is just adding source of trouble.

src/QMCWaveFunctions/tests/test_J1OrbitalSoA.cpp
```diff
@@ -62,6 +61,108 @@ TEST_CASE("J1 evaluate derivatives Jastrow", "[wavefunction]")
...
+  elec_.addTable(elec_);
+  elec_.addTable(ions_);
+  elec_.update();
+
+  ions_.get(app_log());
+  elec_.get(app_log());
+
   const char* jasxml = "<wavefunction name=\"psi0\" target=\"e\"> \
   <jastrow name=\"J1\" type=\"One-Body\" function=\"Bspline\" print=\"yes\" source=\"ion0\"> \
     <correlation elementType=\"H\" cusp=\"0.0\" size=\"2\" rcut=\"5.0\"> \
```
ye-luo, line 168: @shivupa Missing ion group in J1 seems not fully supported. The added test was causing problems, so I add J1O parameter to make it passing for now. We definitely need to fix the case of missing group but not in this PR and also not adding two dummy parameters as the added test indicates.

ye-luo:
Test this please

src/QMCWaveFunctions/tests/test_J1OrbitalSoA.cpp
shivupa: Ok I can take a look in a following PR

src/QMCWaveFunctions/Jastrow/J1OrbitalSoA.h
shivupa: I agree to move to `std::optional` in the future, but also that would have to wait until #2699 is required.

ye-luo:
@PDoakORNL This PR so far looks good for me now. OK to merge? Trying to eliminate long standing PRs.

PDoakORNL (approved):

PR #3244, opened 2021-06-14
Replace deprecated OpenMP get nested call

prckent:
## Proposed changes

omp_get_nested is deprecated in OpenMP >=5.0. Implementations now create warnings. Use the recommended replacement call which has been present since at least 3.0.

- New feature

- No

nitrogen, llvm12.

## Checklist

### DIFF
--- a/src/QMCHamiltonians/ECPotentialBuilder.cpp
+++ b/src/QMCHamiltonians/ECPotentialBuilder.cpp
@@ -82,7 +82,7 @@ bool ECPotentialBuilder::put(xmlNodePtr cur)
 #ifdef ENABLE_OFFLOAD
     NLPP_algo = "batched";
 #else
-    NLPP_algo = (omp_get_max_active_levels() > 1) ? "batched" : "non-batched";
+    NLPP_algo = "non-batched";
 #endif
 
   bool doForces = (forces == "yes") || (forces == "true");
--- a/src/QMCHamiltonians/ECPotentialBuilder.cpp
+++ b/src/QMCHamiltonians/ECPotentialBuilder.cpp
@@ -82,7 +82,7 @@ bool ECPotentialBuilder::put(xmlNodePtr cur)
 #ifdef ENABLE_OFFLOAD
     NLPP_algo = "batched";
 #else
-    NLPP_algo = omp_get_nested() ? "batched" : "non-batched";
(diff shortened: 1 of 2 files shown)


src/QMCHamiltonians/ECPotentialBuilder.cpp
```diff
@@ -82,7 +82,7 @@ bool ECPotentialBuilder::put(xmlNodePtr cur)
 #ifdef ENABLE_OFFLOAD
     NLPP_algo = "batched";
 #else
-    NLPP_algo = omp_get_nested() ? "batched" : "non-batched";
+    NLPP_algo = (omp_get_max_active_levels() > 1) ? "batched" : "non-batched";
```
ye-luo, line 85: I didn't make this change because I remember gcc is not giving 1 when OMP_NESTED is false.

only gcc 11 prints 1, all gcc before 11 prints 2147483647.

So we may need an intermediate function
```
int qmcplusplus_get_max_active_levels()
{
#if defined(SAFE_OMP_GET_MAX_ACTIVE_LEVELS)
  omp_get_max_active_levels();
#else
  return 1; 
#endif
}

SAFE_OMP_GET_MAX_ACTIVE_LEVELS can be set based on GCC version.
I thought about testing it in CMake but this can be difficult for cross-compiling.

ye-luo: We probably need a unit test for omp_get_max_active_levels making sure we can capture non-1 cases for a new compiler.

ye-luo: nvc++ print 1.
icpc prints 2147483647
We must be careful here.

ye-luo: The other options is drop omp_get_max_active_levels and default NLPP_algo to "non-batched"; instead of being conditional.
At this point, I would prefer just delete omp_get_nested/omp_get_max_active_levels

ye-luo (changes requested):

src/QMCHamiltonians/ECPotentialBuilder.cpp
prckent: Lets delete. Simpler.

prckent: Addressed

ye-luo (approved):

PR #3248, opened 2021-06-16
Remove building particleset from hdf5

ye-luo:
## Proposed changes
1. Remove ESHDFParticleParser which creates particleset based on hdf5 from pw2qmcpack. This code path is not widely used and have its limitations. It also uses the outdated HDFAttribIO implementation. So better remove now.
2. Remove the communicator stored in the hdf_archive. It was serving GaussianBasisSet.h and SlaterBasisSet.h but now we have a clean way by just passing the communicator via function arguments.
3. Remove outdated restart input format version HDFWalkerInput_0_0. HDFWalkerInput_0_4 is the one in use.

This is the part of replacing HDFAttribIO to hdf_archive work. Stage what I have now, not gonna continue for a while.

- Other (please describe): remove outdated codes

- Yes
Feature removal

epyc-server

## Checklist


src/QMCWaveFunctions/BsplineFactory/SplineSetReader.h
```diff
@@ -336,44 +336,8 @@ struct SplineSetReader : public BsplineReaderBase
 
   void initialize_spline_psi_r(int spin, const BandInfoGroup& bandgroup)
   {
-    //not used by may be enabled later
+    // old implemenation burried in the history
```
PDoakORNL, line 339: spelling, buried

src/QMCWaveFunctions/WaveFunctionPool.cpp
```diff
@@ -54,29 +54,8 @@ bool WaveFunctionPool::put(xmlNodePtr cur)
...
-      }
-      tcur = tcur->next;
-    }
-  }
-
-  if (qp == 0)
-  {
-    APP_ABORT("WaveFunctionPool::put Target ParticleSet is not found.");
-  }
+  if(qp == nullptr)
```
PDoakORNL, line 57: This seems like a UniformCommunicateError, why not use that std::exception subtype?  I guess then we need to figure out where is going to be caught but it looks like the ValidateXML method in QMCMain should have a `try{} catch` since this is an exception path but of a frequent sort.

Another good reason is this code will eventually be shifted to an input class and shouldn't have a dependency on Message which directly using a myComm method in causes.

PDoakORNL (changes requested):
Overall a good simplification I think.  Just one significant point for discussion.

src/QMCWaveFunctions/BsplineFactory/SplineSetReader.h
prckent: spelling, implementation

src/QMCWaveFunctions/WaveFunctionPool.cpp
ye-luo: If there is a communicator available to the builder, The stop should be handled locally instead of relying on a global try somewhere.

Builder class can benefit from carrying a communicator to allow parallel construction of things if the construction needs a lot of computation which can be benefited from MPI.

src/QMCWaveFunctions/BsplineFactory/SplineSetReader.h
ye-luo: Fixed.

src/QMCWaveFunctions/WaveFunctionPool.cpp
PDoakORNL: On the Second point I agree, but I don't think building and input parsing have the same necessary dependencies.  But its too big a design issue to resolve in this PR if there isn't concensus.

PDoakORNL (approved):

PR #3250, opened 2021-06-17
Reduce clang warning -Wunused-but-set-variable

ye-luo:
## Proposed changes
Clang develop adds more warning which is so annoying.

- No

epyc-server

## Checklist


prckent (approved):
Happy to see the improvements in compiler defaults pushing us to make a more readable and interpretable source code.

PR #3255, opened 2021-06-21
Clean up ham unit tests

ye-luo:
## Proposed changes
1. remove ParticleSetPool dependency in most Hamiltonian unit tests
2. Use SFINAE to remove read functions when containers are const. The error message is much more readable than tons of template compilation failure.

- No

epyc-server

## Checklist


prckent (approved):
I think `typename = std::enable_if_t<!std::is_const<T>::value>` is still easy to follow, so no concerns from me. We do need to be careful around readability and interpretability of our code if we add more of these.

PR #3259, opened 2021-06-25
Redefine substeps

### LINKED ISSUE #3247: substeps definition is not logical and should be revised
Revising the definition of substeps has been discussed in the past but I could not find an issue, so I am creating this one ahead of a PR.

Substeps are used for Monte Carlo moves during VMC where only the electrons are moved and no energy or other observable quantities are computed. The main use is to more cheaply decorrelate the runs than measuring at every Monte Carlo step.

The historic definition of substeps is howevever potentially confusing: if substeps is one, no "no energy measurement" steps are taken. If substeps is two, each step in the Monte Carlo where energies are measured is preceded by one "no energy measurement" step. i.e. The count is off by one from what would be logically guessed, with substeps=0 implying the energy is measured at every Monte Carlo step and substeps=1 implying alternating energy/no energy measurement Monte Carlo updates etc.

In the current scheme the total simulation time is conveniently `dt*nblocks*nsteps*nsubsteps`, while in the more logical scheme it would be `dt*nblocks*nsteps*(1+nsubsteps)`.

While this is strictly a not backwards compatible change, it is a safe to make change since any old inputs would do an extra substep and produce more decorrelated output.

I propose to update the drivers and all the inputs to address this. By my current count it involves changes to at least 534 files which is why this change likely has not been made previously.

prckent:
## Proposed changes

This change makes the definition of substeps more logical in my mind and adopts a definition that I am happier to teach. 

Very simply, substeps is now the number of Monte Carlo steps in between energy and other observable evaluations. e.g. With a timestep of 0.3 a.u. and substeps=0 the energy is measured at times 0, 0.3, 0.6, 0.9... while with substeps=1, the energy is only measured at times 0, 0.6, 1.2...

I am aware that any behavior can be explained by appropriately choosing the definitions. Since I interpret substeps as an extra flavor of MC step, this new definition is consistent.

Input files using the old definition can be changed to be consistent with the new definition by subtracting 1 from the old substep count. Since the main purpose of substeps is decorrelation, old inputs run through the new code will simply be slightly more decorellated due to taking an extra substep, i.e. the change harmless.

The actual code changes in this PR are very small. I have also updated the documentation and hope to have changed all the input files in the repository. The latter was done by a few scripts. The input update is of course essential for the deterministic tests. I have not changed any of the old output files - this would require rerunning all the calculations. The revised input files will generate the same statistics if rerun. Nexus examples are untouched.

Closes #3247

- New feature
- Documentation changes

- Yes

nitrogen, llvm.

## Checklist


prckent:
@jtkrogel I didn't find anything that seemed to require updating in Nexus. Please comment if you agree or if there is something that you think I should also update.

ye-luo (changes requested):
Let us finalize the discussion in #3247 first.

prckent:
Closing due to privately voiced objections. I am not going to push this since we have far more important user facing problems (restart, better ways of specifying the work & target accuracy in the inputs) to deal with, plus all the GPU business. I still think that the current choice is not what new users expect, so may revisit in future.

PR #3260, opened 2021-06-29
[WIP] Dummy change to trigger CI

williamfgc:
## Proposed changes

Just a dummy change to trigger and test CI on self-hosted runners implemented in #3258 

## What type(s) of changes does this code introduce?

- New feature
- Testing changes (e.g. test new CI checks for GPU builds and tests)

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
Sulfur, must run on GitHub Actions CI

## Checklist

- Yes. This PR is up to date with current the current state of 'develop'
- No. Code added or changed in the PR has been clang-formatted
- Yes. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #3266, opened 2021-06-30
CI changes

williamfgc:
## Proposed changes

Address CI changes from feedback in #3260 targeting self-hosted runners.
Move general checks early to report `issue_comment` event checks as skipped
Always issue check status in case of failure
Add `set -x` to CI script for output trace

- Testing changes (e.g. new unit/integration/performance tests) CI self-hosted

- No

Must pass and be tested on CI

## Checklist

### DIFF
--- a/.github/workflows/ci-github-actions-self-hosted.yaml
+++ b/.github/workflows/ci-github-actions-self-hosted.yaml
@@ -7,7 +7,12 @@ on:
 
 jobs:
   gpu-cuda:
-    if: github.repository_owner == 'QMCPACK'
+    if: |
+        github.repository_owner == 'QMCPACK' &&
+        github.event.issue.pull_request &&
+        ( startsWith(github.event.comment.body, 'Test this please') || 
+          startsWith(github.event.comment.body, 'Start testing in-house') )
+          
     runs-on: [self-hosted,Linux,X64,gpu,cuda]
     
     env:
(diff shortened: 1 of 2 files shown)


williamfgc:
@ye-luo @prckent please review and feel free to merge since the actors check need to be tested against non-authorized users.

prckent (commented):
LGTM

Once this is in we should simply test it with comments from an unauthorized user.

.github/workflows/ci-github-actions-self-hosted.yaml
```diff
@@ -7,7 +7,12 @@ on:
 
 jobs:
   gpu-cuda:
-    if: github.repository_owner == 'QMCPACK'
+    if: |
+        github.repository_owner == 'QMCPACK' &&
+        github.event.issue.pull_request &&
+        ( startsWith(github.event.comment.body, 'Test this please') || 
+          startsWith(github.event.comment.body, 'Start testing in-house') )
```
ye-luo, line 14: Does "startsWith" mean a comment starts with the given phrase
or testing starts from the last commit and applies to all subsequent commits?

It looks like the former. github action is new to me, so I got confused.

If I understand correctly, With this PR, 'Start testing in-house' will behave exactly the same as 'Test this please' on the github action side CI testing and we don't have something similar to "Okay to test" on Jenkins side as I described in https://github.com/QMCPACK/qmcpack/pull/3260#issuecomment-870886859.

ye-luo (approved):
LGTM

ye-luo:
Okay to test

ye-luo:
Test this please

.github/workflows/ci-github-actions-self-hosted.yaml
williamfgc: Yes, it's the former. Just to make sure the string starts with it as that the intention, since a comment can contain those phrases in a different context. 

> If I understand correctly, With this PR, 'Start testing in-house' will behave exactly the same as 'Test this please' on the github action side CI testing and we don't have something similar to "Okay to test" on Jenkins side as I described in #3260 (comment).

Yes, for now it's behaving that way. Let me see if there is a way to replicate "Okay to test" with the GitHub Actions API (check if status exists?).

PR #3271, opened 2021-07-01
Working dualspace attached containers

PDoakORNL:
## Proposed changes

When I construct or attach an `OhmmsMatrix` or `OhmmsVector` with a dual space allocator and a pointer to the memory owned by another object using a dual space allocator I expect to end up with a valid device_ptr.

There are problems with the whole attach to a bare pointer to another objects memory design in the OhmmsContainers, but it is very widely used in the code. At the moment this becomes a mess if both Containers have dual space allocators, it just doesn't work as it should. This PR "fixes" this issue while leaving the sort of c++03 container and allocator semantics intact.  

It is needed because the psiM_vgl VecotorSoAContainer is such an important entity for DiracDeterminantBatched and I believe other parts of the WFC ecosystem.  Without this working sanely MatrixDelayUpdateCUDA and DiracMatrixComputeCUDA would end either making many copies of psiM or have a hacky API.

These changes also leave the way is left open to bring the allocators and containers up to c++11 compliance. At which point we should be able to have single functions to attach and construct from another containers dual space "pointer" into another containers data. I think we will be able to use the allocator propagation traits that are part of the >C++11 Allocator documentation to make this allocators much more stl compatible as well.

There are now some useful unit tests of this integration in 
src/tests/integration
where I propose test programs that deal with integration tests like this should go.

These should provide coverage but unit tests for the allocators could be easily written.

- Bugfix
- Documentation changes

- No

Leconte

## Checklist

- No This PR is up to date with current the current state of 'develop'


prckent:
General comment: We seem to have lots of containers, more than we genuinely need. And Ohmms is a now a vestigial tail. It would be healthy to do some merging/deletion/(re)design. Or at least renaming.

src/Platforms/OMPTarget/OMPallocator.hpp
```diff
@@ -34,6 +35,22 @@ T* getOffloadDevicePtr(T* host_ptr)
   return device_ptr;
 }
 
+/** Return the device ptr for memory inside of the dual memory allocation.
+ *  Since we have containers of containers memory and dual memory spaces we need this.
+ */
+template<typename T>
+T* getOffloadDevicePtr(T* allocator_ptr, T* host_ptr)
```
ye-luo, line 42: https://github.com/QMCPACK/miniqmc/blob/OMP_offload/src/Platforms/tests/OMPTarget/test_omp_allocator.cpp
I don't think we need function

src/Platforms/OMPTarget/OMPallocator.hpp
```diff
@@ -71,12 +109,20 @@ struct OMPallocator : public HostAllocator
...
 
 private:
   // pointee is on device.
-  T* device_ptr = nullptr;
+  T* device_ptr_                  = nullptr;
+  T* allocator_host_ptr_ = nullptr;
+
+public:
+  T* get_device_ptr() const { return device_ptr_; }
+  T* get_allocator_host_ptr() const { return allocator_host_ptr_; }
```
ye-luo, line 125: The reason we have device_ptr_ is because it is not stored anywhere else.
The allocated host has been stored in the container. So really prefer not to store a second copy.

I believe there is some design issue but without some example code in a test, I don't see the full picture of the problem you are trying to solve.

src/Platforms/OMPTarget/OMPallocator.hpp
PDoakORNL: We defintely need this because I want the API to include that actual data dependence of the computation.  I don't want to use VectorSoAContainer object as an argument but a Matrix but a matrix made from a part of a psiM_vgl as psiV_temp is has a broken allocator that can not provide a device pointer. I should not need to know about the allocator returned pointer that the owned object has only the pointer it was constructed or attached with.

PDoakORNL:
Unfortunately everything in QMCWaveFunction makes extensive use of the Ohmms containers.  So they're not actually very vestigial at all.

src/Platforms/OMPTarget/OMPallocator.hpp
ye-luo: I probably don't understand what your real need is.

src/Platforms/OMPTarget/OMPallocator.hpp
```diff
@@ -34,6 +35,22 @@ T* getOffloadDevicePtr(T* host_ptr)
...
+template<typename T>
+T* getOffloadDevicePtr(T* allocator_ptr, T* host_ptr)
+{
+  int offset = host_ptr - allocator_ptr;
+  if (offset < 0)
+  {
+    throw std::logic_error("host_ptr is on wrong side of allocator_ptr so must be illegal");
+  }
+  T* device_ptr;
+  PRAGMA_OFFLOAD("omp target data use_device_ptr(allocator_ptr)") { device_ptr = host_ptr; }
```
ye-luo, line 50: I have difficulty to understand what this part does.

prckent:
Then sed 's/Ohmms//g' ? We need more synonyms for historical/legacy etc....

src/Platforms/OMPTarget/OMPallocator.hpp
PDoakORNL: I have a integration test between OhmmsMatrix, VectorSoAContainer, OMPAllocator just about done that will hopefully clarify this.

PDoakORNL:
Still need to test QMC_CUDA=1 and just CPU.

src/Platforms/OMPTarget/OMPallocator.hpp
PDoakORNL: Correct you are @ye-luo.
nonsense function removed.

src/Platforms/OMPTarget/OMPallocator.hpp
PDoakORNL: I think we need to store a copy of it to properly abstract the container level from the details of the stateful relationship between the host and device ptrs supplied by the allocator.

ye-luo:
> These changes make no change to the "current" behaviour that loses the device memory when copying, attaching, or constructing a OhmmsContainer<T, DualAllocator> from another OhmmsContainer<T, DualAllocator>::data(). 

I don't think we loose memory.

I think the issue was the host-device relation is broken in a container with a dual space allocator
 when it is attached to another container with a dual space allocator.  Is this the major issue this PR tries to solve?

src/tests/integration/test_dual_allocators_ohmms_containers.cpp
```diff
@@ -0,0 +1,105 @@
+//////////////////////////////////////////////////////////////////////////////////////
```
ye-luo, line 1: could yo move this to src/Containers/tests?

PDoakORNL:
Yes, to my knowledge we do well in not leaking memory, i.e. I haven't found a problem with OhmmsPETE  or OhmmsSoA containers use of allocate deallocate for dual space allocators. 

> I think the issue was the host-device relation is broken in a container with a dual space allocator
> when it is attached to another container with a dual space allocator. Is this the major issue this PR tries to solve?
> 
Yes

src/tests/integration/test_dual_allocators_ohmms_containers.cpp
PDoakORNL: I'll try, I do think it is an integration test, but its debatable since support for dual space is part of the Ohmms containers API now.

ye-luo (changes requested):
I'm doing some experiment. Blocking merging temporarily.

src/Platforms/CUDA/CUDAallocator.hpp
```diff
@@ -152,6 +153,53 @@ struct CUDAAllocator
...
+    {
+      cudaErrorCheck(cudaMemcpy(to_ptr, from_ptr, sizeof(T) * n, cudaMemcpyDeviceToDevice),
+                     "cudaMemcpy failed in copyDeviceToDevice");
+    }
+  }
+  
+  void set_stream(cudaStream_t stream) { hstream_ = stream; }
+
+private:
+  cudaStream_t hstream_ = 0;
```
ye-luo, line 202: All these copy APIs should be used for testing or performance irrelevant locations. Adding "hstream_" is not a sane design. Prefer remove.

ye-luo (commented):
All the rest looks good to me.

src/Containers/OhmmsPETE/OhmmsMatrix.h
```diff
@@ -108,6 +118,18 @@ class Matrix
     X.attachReference(ref, TotSize);
   }
 
+  /** Attach to pre-allocated memory and propagate the allocator of the owning container.
+   *  Required for sane access to dual space memory
+   */
+  template<typename CONTAINER>
+  inline void attachReference(const CONTAINER& other, T* ref, size_type n, size_type m)
```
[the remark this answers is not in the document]
PDoakORNL: You mean in a comment?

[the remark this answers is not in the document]
ye-luo: Sorry for the confusion. I really just mean leaving this note on github. But if you feel the need of adding a line of comment in the source code, go ahead.

src/Platforms/CUDA/CUDAallocator.hpp
PDoakORNL: I think its much more important to be able to write good test of everything. The API's need to be consistent but openmp target is still 1st class and native c++ second class. Stream is secret state in the openmp runtime so it can't be in the API for transfers. But if any comparison is going to be made between openmp target code and c++ with minimal native vendor code cuda transfers need to be able to set stream.  Perhaps the transfers should be handled by a separate functor, but the allocator is where the device_ptr state is and device_ptr's will definitely have a state relation with streams in future cuda code (there are now stream local allocator calls). And the the openmp runtime should take advantage of this.

ye-luo: The transfer calls updateTo/From in OMPallocator is synchronous. OpenMP has nothing to do with streams. In CUDA code if stream is used, i would see it being passed at use not via CUDAallocator.

src/Platforms/OMPTarget/OMPallocator.hpp
PDoakORNL: Just before closing this, we don't need this for OMPAllocator because openmp maintains mapping ranges from host to device ptrs.  We do need it in the case of the general DualAllocator where there is no runtime looking out for these mappings.

src/Platforms/CUDA/CUDAallocator.hpp
PDoakORNL: how do we get an asynchronous transfer from openmp target?

src/Platforms/OMPTarget/OMPallocator.hpp
ye-luo: I still think it is not needed in DualAllocator except for adding assertions. Anyway, the impact has been minimized.

ye-luo:
Test this please

ye-luo (approved):
All my concerns are addressed.

PR #3275, opened 2021-07-06
workaround findMPI issue

ye-luo:
## Proposed changes
FindMPI has some issue https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6304
On QMCPACK side, I added a workaround by directly caching MPI_CXX_SKIP_MPICXX

In addition, this PR updates the build recipe on tulip.

- Build related changes

- No

epyc-server, tulip

## Checklist


prckent (changes requested):
"If true, the MPI-2 C++ bindings are disabled using definitions"

Please change this to include guidance for a less technical user. Is the message a good thing or a bad thing? What are the consequences? Also, why print "If true"?

ye-luo:
> "If true, the MPI-2 C++ bindings are disabled using definitions"
> 
> Please change this to include guidance for a less technical user. Is the message a good thing or a bad thing? What are the consequences? Also, why print "If true"?

Updated. The previous message was copied from findMPI line. The updated message is more meaningful for QMCAPCK.

ye-luo:
Start testing in-house

prckent (approved):

PR #3284, opened 2021-07-09
Spline bugfix for spinors

camelto2:
## Proposed changes

This PR addresses the bug for splined spinors that has prevented the SOC applied to solids. 

The issue was specific to the splines, in that an orbital rotation was being applied to balance the real and imaginary components of the splines. In the spinor case, since the spinor has the form phi_up e^{is} + phi_dn e^{-is}, if the splined phi_up and phi_dn are modified, the spin functions further alter the phase from the spin functions...and ultimately results in a bias in the spinors and therefore the entire wavefunction. This PR adds a bypass to the orbital rotation in the EinsplineSpinorSetBuilder.

I also had to modify the einspline spinor unit test to reflect this rotation not being applied, and instead used an external python tool to directly evaluate the plane wave orbitals/spinors and ues those as reference values.

This should (hopefully) be the last fix, and we can finally move on to trying out spin-orbit in solids. 

- Bugfix

- No

My workstation, gnu 9.2.0

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/BsplineReaderBase.h
+++ b/src/QMCWaveFunctions/BsplineFactory/BsplineReaderBase.h
@@ -181,7 +181,7 @@ struct BsplineReaderBase
   /** Set the checkNorm variable */
   inline void setCheckNorm(bool new_checknorm) { checkNorm = new_checknorm; };
 
-  /** Set the orbital rotation flagi */
+  /** Set the orbital rotation flag. Rotations are applied to balance the real/imaginary components. */
   inline void setRotate(bool new_rotate) { rotate = new_rotate; };
 
   void initialize_spo2band(int spin,
--- a/src/QMCWaveFunctions/BsplineFactory/BsplineReaderBase.h
+++ b/src/QMCWaveFunctions/BsplineFactory/BsplineReaderBase.h
@@ -44,6 +44,8 @@ struct BsplineReaderBase
   bool checkNorm;
   ///save spline coefficients to storage
   bool saveSplineCoefs;
+  ///apply orbital rotations
+  bool rotate;
   ///map from spo index to band index
   std::vector<std::vector<int>> spo2band;
 
@@ -178,6 +180,9 @@ struct BsplineReaderBase
 
   /** Set the checkNorm variable */
   inline void setCheckNorm(bool new_checknorm) { checkNorm = new_checknorm; };
+
+  /** Set the orbital rotation flagi */
+  inline void setRotate(bool new_rotate) { rotate = new_rotate; };
 
   void initialize_spo2band(int spin,
                            const std::vector<BandInfo>& bigspace,
--- a/src/QMCWaveFunctions/einspline_helper.hpp
+++ b/src/QMCWaveFunctions/einspline_helper.hpp
@@ -243,20 +243,14 @@ inline void fix_phase_rotate_c2c(const Array<std::complex<T>, 3>& in,
     }
 }
 
-template<typename T, typename T1, typename T2>
-inline void fix_phase_norotate_c2c(const Array<std::complex<T>, 3>& in,
+template<typename T, typename T1>
+inline void split_real_components_c2c(const Array<std::complex<T>, 3>& in,
                                  Array<T1, 3>& out_r,
-                                 Array<T1, 3>& out_i,
-                                 const TinyVector<T2, 3>& twist,
-                                 T& phase_r,
-                                 T& phase_i)
+                                 Array<T1, 3>& out_i)
 {
   const int nx = in.size(0);
   const int ny = in.size(1);
   const int nz = in.size(2);
-
-  phase_r = T(1.0);
-  phase_i = T(0.0);
 
 #pragma omp parallel for
   for (size_t ix = 0; ix < nx; ++ix)
@@ -268,8 +262,8 @@ inline void fix_phase_norotate_c2c(const Array<std::complex<T>, 3>& in,
       T1* restrict i_ptr                     = out_i.data() + offset;
       for (size_t iz = 0; iz < nz; ++iz)
       {
-        r_ptr[iz] = static_cast<T1>(phase_r * in_ptr[iz].real() - phase_i * in_ptr[iz].imag());
-        i_ptr[iz] = static_cast<T1>(phase_i * in_ptr[iz].real() + phase_r * in_ptr[iz].imag());
+        r_ptr[iz] = static_cast<T1>(in_ptr[iz].real());
+        i_ptr[iz] = static_cast<T1>(in_ptr[iz].imag());
       }
     }
 }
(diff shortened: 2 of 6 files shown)


rcclay:
Test this please.

src/QMCWaveFunctions/einspline_helper.hpp
```diff
@@ -243,6 +243,37 @@ inline void fix_phase_rotate_c2c(const Array<std::complex<T>, 3>& in,
     }
 }
 
+template<typename T, typename T1, typename T2>
+inline void fix_phase_norotate_c2c(const Array<std::complex<T>, 3>& in,
```
ye-luo, line 247: This function should be simplified to only copy from `in` to `out_r` and `out_i`.
Setting phase_r and phase_i can be set at the caller side directly.

src/QMCWaveFunctions/tests/test_einset_spinor.cpp
```diff
@@ -97,7 +97,7 @@ TEST_CASE("Einspline SpinorSet from HDF", "[wavefunction]")
 
 
   const char* particles = "<tmp> \
-   <sposet_builder name=\"A\" type=\"einspline\" href=\"o2_45deg_spins.pwscf.h5\" tilematrix=\"1 0 0 0 1 0 0 0 1\" twistnum=\"0\" source=\"ion\" size=\"3\" precision=\"float\"> \
```
ye-luo, line 100: How  much does the memory increase? mesh 4 will be 64 times larger.

src/QMCWaveFunctions/tests/test_einset_spinor.cpp
```diff
@@ -291,26 +293,26 @@ TEST_CASE("Einspline SpinorSet from HDF", "[wavefunction]")
 
   for (unsigned int iat = 0; iat < 3; iat++)
   {
-    REQUIRE(psiM[iat][0] == ComplexApprox(psiM_ref[iat][0]));
-    REQUIRE(psiM[iat][1] == ComplexApprox(psiM_ref[iat][1]));
-    REQUIRE(psiM[iat][2] == ComplexApprox(psiM_ref[iat][2]));
+    REQUIRE(psiM[iat][0] == ComplexApprox(psiM_ref[iat][0]).epsilon(h));
```
ye-luo, line 296: Let us use CHECK instead of REQUIRES.

src/QMCWaveFunctions/tests/test_einset_spinor.cpp
rcclay: Looks like the whole h5 file is 1M.

src/QMCWaveFunctions/einspline_helper.hpp
rcclay: I'm OK having an auxiliary function here, like "real_component_split_c2c" that 1.) performs a direct copy for the real and imaginary parts as Ye suggests, and 2.) sets the phase phi_r=1 and phi_i=0.  I'd rather have the phase determined by the choice of operation than leave this to the user.  Not necessary though.

jtkrogel:
Is this only an issue for the real code, or does it show up in complex also?

camelto2:
> Is this only an issue for the real code, or does it show up in complex also?

This only impacts spinors from what I can tell, and we only ever run the spinors with complex. Normal splined (real/complex) orbitals are fine to have these rotations, the problem only arises for spinors when you add the up and down components and apply the spin function to each component. . So we just have to disable it in this use case.

src/QMCWaveFunctions/tests/test_einset_spinor.cpp
camelto2: When I was first making the test, I had to crank up the mesh factor to get the splines to exactly agree with the raw PW values and I just left it there. As far as memory is concerned, when running the test the up and down components each print `MEMORY 43 MB allocated...`. So if this is too high for a test, I can reduce the meshfactor and will therefore change the tolerances for the checks.

src/QMCWaveFunctions/einspline_helper.hpp
camelto2: done

src/QMCWaveFunctions/tests/test_einset_spinor.cpp
camelto2: done

rcclay (approved):
After changes, LGTM.

rcclay (approved):

src/QMCWaveFunctions/BsplineFactory/BsplineReaderBase.h
```diff
@@ -179,6 +181,9 @@ struct BsplineReaderBase
   /** Set the checkNorm variable */
   inline void setCheckNorm(bool new_checknorm) { checkNorm = new_checknorm; };
 
+  /** Set the orbital rotation flag. Rotations are applied to balance the real/imaginary components. */
+  inline void setRotate(bool new_rotate) { rotate = new_rotate; };
```
ye-luo, line 185: keep a note for my self. We will need to kill all the setXX function and pass via constructor.

ye-luo (approved):

src/QMCWaveFunctions/tests/test_einset_spinor.cpp
ye-luo: 43MB is not a concern

ye-luo:
Test this please

src/QMCWaveFunctions/tests/test_einset_spinor.cpp
prckent: The discussion here is about increasing the size of the git repo permanently.

prckent: It is good that the large file was spotted. If there is a chance to make it smaller we should take advantage. however I think 1MB is fair provided there are not plans to keep updating it.

Newly developed or improved effective core potentials have been made available via the pseudopotentiallibrary.org website.

@camelto2 Do you think that this is a good file that will be stable for a long time? I assume so but if there is another system that you think might be more appropriate we should update.

camelto2: I don't see a reason why this file would need to change at all in the future. It is just the output from a QE noncollinear calculation, so it should be stable.

With that said, since I now have an independent tool to directly get the reference values from plane waves, I can run whatever system with a smaller PW cutoff and get a smaller hdf5 file if that is desired.

We just chose this system because we used it to benchmark the noncollinear wave functions against some collinear calculations in the past.

ye-luo: @prckent I got confused by your message. the h5 in the repo is 924K. I saw an change of mesh factor in the test and thus I was wondering if memory needs for testing increases or not. 43M memory in a unit test is not an issue.

prckent: I thought this was a new file. In any case, good to have the discussion.

prckent:
Thanks @camelto2

PR #3287, opened 2021-07-13
Split compute and transfer in distance table AA offload

ye-luo:
## Proposed changes
The data transfer becomes on-demand when a consumer needs the data on the host.
Clean up the accept_rejectMove API in forward and regular mode. Much straightforward call logic and adds unit tests to cover all the batched and non-batched APIs.

- New feature

- No

epyc-server

## Checklist


src/Particle/SoaDistanceTableAAOMPTarget.h
```diff
@@ -352,6 +359,24 @@ struct SoaDistanceTableAAOMPTarget : public DTD_BConds<T, D, SC>, public Distanc
...
+  {
+    // if temp data on host is not updated during p-by-p moves, there is no need to update distance table
+    if (!(modes_ & DTModes::NEED_TEMP_DATA_ON_HOST)) return;
+
+#pragma omp parallel for
+    for (int iw = 0; iw < dt_list.size(); iw++)
+      dt_list[iw].updatePartial(jat, from_temp[iw]);
+  }
+
+  void mw_finalizePbyP(const RefVectorWithLeader<DistanceTableData>& dt_list,
```
PDoakORNL, line 372: I guess that means a side effect of this function is a total transfer of the distance table device to host?  Hasn't that already been done on the device? So mw_evaluate seems to got to evaluate which the base class doesn't mention doing transfer between memory spaces and neither does the implementation here. Needs docs.

src/Particle/ParticleSet.cpp
```diff
@@ -517,6 +517,7 @@ void ParticleSet::mw_computeNewPosDistTablesAndSK(const RefVectorWithLeader<Part
       const auto dt_list(extractDTRefList(p_list, i));
       p_leader.DistTables[i]->mw_move(dt_list, p_list, new_positions, iat, maybe_accept);
     }
+    PRAGMA_OFFLOAD("omp taskwait")
```
PDoakORNL, line 520: I think document what is being waited on, `omp taskwait` is completely generic.  I'm pretty sure it just the transfer to host (for some combinations of DT_options) of distance table values calculated on the device. It's also possible the the distance tables won't be needed for some time and if so there is no need to wait here,  that optimization will be easy to see if this is documented.  Does the SK->makeMove call depend on these tasks?  It doesn't seem to, can the pragma be moved after it?

PDoakORNL (changes requested):
Some of the changes needed are technical debt from other offload + distance table work going in without much in the way of documentation. But I think there is important transfer and synchronization logic here that can't be easily understood by reading the source right now. No one is going to understand the interaction of the openmp target pragmas, its runtime state machine, and the distance table state machine in the future.

Somewhere I expect to see a test of the behavior of the distance table data movement or lack of it with respect to the different DTMode options.  You might get coverage from 'long distance_pbc_z batched APIs ee NEED_TEMP_DATA_ON_HOST' but it doesn't explain much about the behavior. Have many calls now update or don't the host side data?

src/Particle/SoaDistanceTableAAOMPTarget.h
```diff
@@ -237,7 +237,7 @@ struct SoaDistanceTableAAOMPTarget : public DTD_BConds<T, D, SC>, public Distanc
       ScopedTimer offload(offload_timer_);
       PRAGMA_OFFLOAD("omp target teams distribute collapse(2) num_teams(nw * num_teams) \
                         map(always, to: rsoa_dev_list_ptr[:rsoa_dev_list.size()]) \
-                        map(always, from: r_dr_ptr[:nw_new_old_dist_displ.size()])")
+                        nowait depend(out: r_dr_ptr[:nw_new_old_dist_displ.size()])")
```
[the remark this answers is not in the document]
ye-luo: Add documentation in DistanceTableData::mw_move about the asynchronicity and transfer behavior.

src/Particle/SoaDistanceTableAAOMPTarget.h
```diff
@@ -333,6 +339,7 @@ struct SoaDistanceTableAAOMPTarget : public DTD_BConds<T, D, SC>, public Distanc
   void updatePartial(IndexType jat, bool from_temp) override
```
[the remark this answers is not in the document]
ye-luo: updatePartial is used by accept_rejectMove in forward mode. Forward mode has been documented at ParticleSet.h

src/Particle/SoaDistanceTableAAOMPTarget.h
ye-luo: AA table is Ne^2/2 which is too much for GPU. So the full table is never stored fully on GPU. Since the "temp" data are discarded during p-by-p move, a recompute is needed to update the table on the host.

src/Particle/ParticleSet.cpp
ye-luo: I added in-place documentation about the wait and also added explanation about asynchronous behavior of DistanceTableData::mw_move. The `SK->makeMove` question, probably better to put taskwait after but we don't really care, I'm not ware any production feature activates that. So I will just leave it there.

src/Particle/tests/test_distance_table.cpp
```diff
@@ -490,35 +496,159 @@ TEST_CASE("distance_pbc_z", "[distance_table][xml]")
...
+
+  electrons.makeMove(1, disp);
+  electrons.accept_rejectMove(1, false, true);
+  CHECK(ee_dtable.getDistRow(1)[0] == Approx(2.7239676944));
+  CHECK(ee_dtable.getDisplRow(1)[0][0] == Approx(-2.7));
+  CHECK(ee_dtable.getDisplRow(1)[0][1] == Approx(0.3));
+  CHECK(ee_dtable.getDisplRow(1)[0][2] == Approx(0.2));
 } // TEST_CASE distance_pbc_z
 
+#if defined(ENABLE_OFFLOAD)
```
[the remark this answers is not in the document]
ye-luo: expanded.

ye-luo:
Start testing in-house

ye-luo:
Start testing in-house

PDoakORNL (approved):
I'm happy with the documentation now.  I hope for more "unit" testing in the future.

PR #3288, opened 2021-07-13
Keep only lower triangle in AA distances

ye-luo:
review after #3287

## Proposed changes
Keep the table up-to-date and symmetry is costly in human effort, memory and computational cost. So I changed all the
lower triangle only.
The displacements has been changed to lower triangle when SoA was developed. Now just make one more step.

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/DistanceTableData.h
+++ b/src/Particle/DistanceTableData.h
@@ -55,9 +55,11 @@ protected:
   /*@{*/
   /** distances_[i][j] , [N_targets][N_sources]
    *  Note: Derived classes decide if it is a memory view or the actual storage
-   *        For derived AA, only the lower triangle (j<i) is defined and up-to-date after pbyp move.
-   *          The upper triangle is symmetric to the lower one only when the full table is evaluated from scratch.
-   *          Avoid using the upper triangle because we may change the code to only allocate the lower triangle part.
+   *        For derived AA, only the lower triangle (j<i) data can be accessed safely.
+   *            There is no bound check to protect j>=i terms as the nature of operator[].
+   *            When the storage of the table is allocated as a single memory segment,
+   *            out-of-bound access is still within the segment and
+   *            thus doesn't trigger an alarm by the address sanitizer.
    *        For derived AB, the full table is up-to-date after pbyp move
    */
   std::vector<DistRow> distances_;
@@ -65,7 +67,7 @@ protected:
   /** displacements_[N_targets]x[3][N_sources]
    *  Note: Derived classes decide if it is a memory view or the actual storage
    *        displacements_[i][j] = r_A2[j] - r_A1[i], the opposite sign of AoS dr
-   *        For derived AA, A1=A2=A, only the lower triangle (j<i) is defined.
+   *        For derived AA, A1=A2=A, only the lower triangle (j<i) is defined. See the note of distances_
    *        For derived AB, A1=A, A2=B, the full table is allocated.
    */
   std::vector<DisplRow> displacements_;
--- a/src/Particle/SoaDistanceTableAA.h
+++ b/src/Particle/SoaDistanceTableAA.h
@@ -93,7 +93,7 @@ struct SoaDistanceTableAA : public DTD_BConds<T, D, SC>, public DistanceTableDat
     for (int iat = 0; iat < N_targets; ++iat)
     {
       DTD_BConds<T, D, SC>::computeDistances(P.R[iat], P.getCoordinates().getAllParticlePos(), distances_[iat].data(),
-                                             displacements_[iat], 0, N_targets, iat);
+                                             displacements_[iat], 0, iat, iat);
       distances_[iat][iat] = BigR; //assign big distance
     }
   }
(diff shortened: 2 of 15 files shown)


src/Particle/SoaDistanceTableAA.h
```diff
@@ -25,8 +25,8 @@ struct SoaDistanceTableAA : public DTD_BConds<T, D, SC>, public DistanceTableDat
   ///number of targets with padding
   int Ntargets_padded;
 
-  ///actual memory for displacements_
-  aligned_vector<RealType> memory_pool_displs_;
+  ///actual memory for dist and displacements_
+  aligned_vector<RealType> memory_pool_;
```
PDoakORNL, line 29: Are RealType and T the same? somtimes/alaways  I gave up trying to figure that out.

src/Particle/SoaDistanceTableAA.h
```diff
@@ -66,13 +66,13 @@ struct SoaDistanceTableAA : public DTD_BConds<T, D, SC>, public DistanceTableDat
...
     const size_t total_size = compute_size(N_targets);
-    memory_pool_displs_.resize(total_size * D);
+    memory_pool_.resize(total_size * (1 + D));
     distances_.resize(N_targets);
     displacements_.resize(N_targets);
     for (int i = 0; i < N_targets; ++i)
     {
-      distances_[i].resize(Ntargets_padded);
-      displacements_[i].attachReference(i, total_size, memory_pool_displs_.data() + compute_size(i));
+      distances_[i].attachReference(memory_pool_.data() + compute_size(i), i);
```
PDoakORNL, line 74: Its hard to understand whats going on here. I tried defining similar functions to compute_size in the python shell.

lets say sizeof(T) is 8 and align is 32
computesize() mapped over (0, 1, 2, 3,4, 5) -> (0,4,8,12,16,24)
So I'm attaching references every 4 T's (actually RealTypes because thats what memory_pool_ is) but the size of distances_ grows larger than 4 T's. clearly starts to be bad.  Hopefully I just don't understand compute_size.  

Yet I just don't trust this code without tests.  It seems like the exact details of this memory layout are important but they're cryptic.

src/Particle/SoaDistanceTableAA.h
```diff
@@ -128,19 +125,28 @@ struct SoaDistanceTableAA : public DTD_BConds<T, D, SC>, public DistanceTableDat
           min_dist = temp_r_[jat];
           index    = jat;
         }
-      if (index >= 0)
-        dr = temp_dr_[index];
+      assert(index >= 0);
```
PDoakORNL, line 128: This sort of makes up for the lack of unit test, but I'd rather there was one and it was run in debug. At the moment I can't see how anything but temp_r_'s of all NaN or infty could trip this.  What insures the calling code never produces such input?

src/Particle/SoaDistanceTableAAOMPTarget.h
```diff
@@ -98,13 +98,13 @@ struct SoaDistanceTableAAOMPTarget : public DTD_BConds<T, D, SC>, public Distanc
     // initialize memory containers and views
     Ntargets_padded         = getAlignedSize<T>(N_targets);
     const size_t total_size = compute_size(N_targets);
```
PDoakORNL, line 100: See above.

PDoakORNL (changes requested):
I'm hoping lack of documentation has just confused me, or there may be a problem with the memory pool and compute_size

src/Particle/SoaDistanceTableAA.h
```diff
@@ -90,12 +90,9 @@ struct SoaDistanceTableAA : public DTD_BConds<T, D, SC>, public DistanceTableDat
   {
     ScopedTimer local_timer(evaluate_timer_);
     constexpr T BigR = std::numeric_limits<T>::max();
-    for (int iat = 0; iat < N_targets; ++iat)
-    {
+    for (int iat = 1; iat < N_targets; ++iat)
       DTD_BConds<T, D, SC>::computeDistances(P.R[iat], P.getCoordinates().getAllParticlePos(), distances_[iat].data(),
-                                             displacements_[iat], 0, N_targets, iat);
-      distances_[iat][iat] = BigR; //assign big distance
```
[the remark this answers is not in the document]
ye-luo: We were allocating it but no more after this change.
In reality, it may get allocated due to padding but this is invisible to users.

src/Particle/SoaDistanceTableAA.h
ye-luo: I will figure that out.

src/Particle/SoaDistanceTableAA.h
ye-luo: row 0, size 0
row 1, size 4
row 2, size 4,
row 3, size 4
row 4, size 8
row 5, size 8

(0, 1, 2, 3, 4, 5 ) -> (0, 4, 8, 12, 20, 28)

src/Particle/SoaDistanceTableAA.h
ye-luo: I don't think we can protect NaN anywhere in the code. I never tound this causing an issue. So I just leave the assert.

src/Particle/SoaDistanceTableAA.h
ye-luo: Probably some cleaning will be need both in the base also in this class.

src/Particle/SoaDistanceTableAA.h
ye-luo: Unit test added.

src/Particle/DistanceTableData.h
```diff
@@ -55,17 +55,15 @@ class DistanceTableData
...
+   *        For derived AA, only the lower triangle (j<i) is defined. The diagonal and upper triangle are not accessible.
    *        For derived AB, the full table is up-to-date after pbyp move
    */
   std::vector<DistRow> distances_;
 
   /** displacements_[N_targets]x[3][N_sources]
    *  Note: Derived classes decide if it is a memory view or the actual storage
    *        displacements_[i][j] = r_A2[j] - r_A1[i], the opposite sign of AoS dr
-   *        For derived AA, A1=A2=A, only the lower triangle (j<i) is defined.
+   *        For derived AA, A1=A2=A, only the lower triangle (j<i) is defined. The diagonal and upper triangle are not accessible.
```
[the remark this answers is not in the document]
ye-luo: Strong words added.

src/Particle/SoaDistanceTableAA.h
PDoakORNL: Perhaps an issue makes sense?

ye-luo:
Start testing in-house

src/Particle/SoaDistanceTableAA.h
PDoakORNL: Ok I get it now, while it seems i did port it correctly I did fail to do some basic arithmetic in my head correctly 24-16 != 4.  The sizes make sense if the diagonal isn't included.

src/Particle/SoaDistanceTableAA.h
PDoakORNL: I think given the limited use of this function we can table it, at least there is an assert.

PDoakORNL:
Unit test provides good clarity on the alignment requirements.

PDoakORNL (approved):

PR #3293, opened 2021-07-15
Nexus: Allow bandstructure plotting with custom k-path

mcbennet:
## Proposed changes

This PR adds to the function `plot_bandstructure` in `./nexus/lib/pwscf_analyzer.py`. It allows the user to provide a custom set of labels for the high symmetry k-points (in the form k_labels=['GAMMA','','','','M','','',...]) where the length of the label array should match the number of k-points in the analyzer object. When custom labels are give, spglib is no longer used since it is assume that the k-points in the analyzer object are different from the ones that would be generated by spglib.

- New feature

- No

Laptop. Linux

## Checklist

### DIFF
--- a/nexus/lib/pwscf_analyzer.py
+++ b/nexus/lib/pwscf_analyzer.py
@@ -843,8 +843,9 @@ class PwscfAnalyzer(SimulationAnalyzer):
     #end def make_movie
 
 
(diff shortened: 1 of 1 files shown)


prckent:
We should probably add this to the PR template: are any python package dependencies changed?

mcbennet:
> We should probably add this to the PR template: are any python package dependencies changed?

None of the dependencies are altered.

ye-luo:
Test this please

ye-luo (approved):

Issue #2994, opened 2021-03-05
energy density matrix not correctly printed in stat.h5


stefdis:
I am using QMCPACK 3.10.0 and I found that using energy_matrix = yes in the dm1b module produce vanishing 1b reduced energy density matrix (energy_matrix in *.stat.h5) and 1b reduced density matrix (number_matrix).
When I use energy_matrix = no the 1bRDM is normally printed and elements are not all zeros.

For the dm1d estimator I am using

         <estimator type="dm1b" name="DensityMatrices">
            <parameter name="energy_matrix" > yes         </parameter>
            <parameter name="basis"         >  dm_basis    </parameter>
            <parameter name="evaluator"     >  matrix      </parameter>
            <parameter name="integrator"    >  uniform     </parameter>
            <parameter name="samples"       >  64          </parameter>
            <parameter name="scale"         >  1.0         </parameter>
            <parameter name="center"        >  0 0 0       </parameter>
         </estimator>

jtkrogel:
Please try again with `<traces array="yes" write="no"/>` inserted between the `<qmcsystem>` and `<qmc>` elements in your input, similar to the energy density example at https://github.com/QMCPACK/qmcpack/blob/develop/tests/solids/diamondC_1x1x1_pp/qmc_edens_cell_dmc_short.in.xml.

stefdis:
With this extra line I get a segmentation fault

[compute-3-6:08564] *** Process received signal ***
[compute-3-6:08564] Signal: Segmentation fault (11)
[compute-3-6:08564] Signal code: Address not mapped (1)
[compute-3-6:08564] Failing at address: (nil)
[compute-3-6:08564] [ 0] /lib64/libpthread.so.0(+0xf630)[0x7f491e9f9630]
[compute-3-6:08564] [ 1] qmcpack[0x88ec36]
...

stefdis:
Ia using the diamondC_1x1x1_pp/qmc_1rdm_noJ_short.in.xml example to test the RDEM routine.
The estimator block

```
         <estimator type="dm1b" name="DensityMatrices">
            <parameter name="energy_matrix"       >    yes                    </parameter>
            <parameter name="integrator"          >    uniform_grid          </parameter>
            <parameter name="points"              >    6                     </parameter>
            <parameter name="scale"               >    1.0                   </parameter>
            <parameter name="basis"               >    spo_ud spo_dm         </parameter>
            <parameter name="evaluator"           >    matrix                </parameter>
[...]

PR #3294, opened 2021-07-15
Possible fix for energy density

markdewing:
Addresses two issues that cause #2994

One is the loop bounds in `QMCHamiltonian::initialize_traces` starts at 1 rather than 0. This would skip the kinetic energy, an operator in the quantum domain would not be in the loop, and a trace of Vq would not requested.

The second is ensuring the per-particle S(k) state is initialized for the traces. Otherwise there is a segfault from eikr_r/i being
zero-length vectors.

- Bugfix

- No

laptop

## Checklist


src/QMCHamiltonians/CoulombPBCAA.cpp
```diff
@@ -216,6 +216,10 @@ CoulombPBCAA::Return_t CoulombPBCAA::evaluate_sp(ParticleSet& P)
     }
     else
     {
+      if (!P.getPerParticleSKState())
+      {
+        P.turnOnPerParticleSK();
```
ye-luo, line 221: I doubt this is the right place to flip the switch which doesn't make the data ready which seems needed by the evalluate below.

ye-luo:
> One is the loop bounds in QMCHamiltonian::initialize_traces starts at 1 rather than 0. This would skip the kinetic energy, an operator in the quantum domain would not be in the loop, and a trace of Vq would not requested.

@jtkrogel Was this intentional

src/QMCHamiltonians/CoulombPBCAA.cpp
markdewing: It looks like `StructFact::turnOnStorePerParticle` calls `FillRhok`, which initializes the `eikr_r` array.  I'm open to suggestions for a better way or place to enable it.

PDoakORNL: I feel like this belongs in the constructor, its called on the ref ParticleSet passed there if ComputeForces == true. This seems a bit wrong as well but we can unravel one level of tangle at a time. I would assume that the evaluate_sp ParticleSet& should be getting called on the ref passed to the constructor or a clone of it. No one would expect to have to manipulate the state of the ParticleSet here to make it valid.

markdewing: I'm not sure it's easily possible to move this to the constructor - this is not know at construction time.  I can try move it up the call stack so it's in a more general place.

markdewing:
If the loop is only about potentials, then there is no one-body electron potential, so Vq never gets added from a potential.

jtkrogel:
That's true for most Hamiltonians we deal with.  It is present for generality.

markdewing:
I'm still confused about how the energy density code could have worked.   Even from the oldest commit of DensityMatrices1B.cpp, computing the energy matrix requires a Vq trace.  And the loop in QMCHamiltonian::initialize_traces that would add it still started at 1, so the kinetic energy would have been skipped.

One possibility is at the time of the initial commit, deaca41a22cf0cbd4f76d1fd467a7b873ca9c0df , the APP_ABORT in TraceSample::get_combined_trace was not present.  It would have returned NULL to Vq_trace in DensityMatrix1B.  The place where Vq_trace is used, in accum_sample, has a guard for NULL.  So the code would not have seg. faulted.  It would have ignored contributions from Vq.

The APP_ABORT in get_combined_trace was added shortly afterwards in commit e63019fdeed6485fe59680f0d0ff6201c669af83

markdewing:
I think I get it now - if there is no external potential, there is no contribution from Vq.  So for a fix, there needs to be a way for DensityMatrices1B to ignore a trace that is not present.

markdewing:
The fix is updated
1. To handle Vq being optional, the null test and APP_ABORT were removed from get_combined_trace.  If this is not desired in general, one solution would be to add a flag to get_combined_trace that indicates whether the value is required or optional.  And then the null test and abort would only occur for required values.
2. The per-particle S(k) check and activate was moved to checkout_particle_quantities.  Another option (that I haven't tried) would be to move this code to DensityMatrices1B (either the constructor or some other method of that class).

src/QMCHamiltonians/CoulombPBCAA.cpp
```diff
@@ -141,6 +141,10 @@ void CoulombPBCAA::checkout_particle_quantities(TraceManager& tm)
   streaming_particles = request.streaming_array(myName);
   if (streaming_particles)
   {
+    if (!Ps.getPerParticleSKState())
```
PDoakORNL, line 144: If I'm following here, this is checked because certain state needs to be initialized by Ps in SK if SK being run per particle, I don't see a way to set this in SK except by calling turnOnPerParticleSK(), which makes this whole exercise seem redundant as getPerParticleSKState will only be true if turnOnPerParticleSK has already been called on SK. 

This only happens at these locations
```
src/QMCHamiltonians/CoulombPBCAA.cpp:42:    ref.turnOnPerParticleSK();
src/QMCHamiltonians/CoulombPBCAB.cpp:38:    PtclA.turnOnPerParticleSK();
src/QMCHamiltonians/EnergyDensityEstimator.cpp:69:    Pdynamic->turnOnPerParticleSK();
src/QMCHamiltonians/EnergyDensityEstimator.cpp:81:      Pstatic->turnOnPerParticleSK();
src/QMCHamiltonians/ForceChiesaPBCAA.cpp:41:  ions.turnOnPerParticleSK();
```
What is the code path that results in StructFact.StorePerParticle==true without the call to `ParticleSet::turnOnPerParticleSK`?

This strange state sequence should be explained in documentation

PDoakORNL (changes requested):

src/QMCHamiltonians/CoulombPBCAA.cpp
markdewing: I guess the check can be eliminated, and just call turnOnPerParticleSK.

PDoakORNL: So for the other Hamiltonian components that turnOnPerParticleSK when forces are computed, do they also need to turn it on if streaming_particles == true. This would be CoulombPBCAB and ForceChiesaPBCAA

markdewing: If we had an test case with an all-electron calculation, CoulombPBCAB would probably need to be fixed.    I don't know what conditions would be needed for ForceChiesaPBCAA.    I would prefer there be a test case present before changing the code.

markdewing: I tested an all-election case, and what happened was the code in CoulombPBCAA called StorePerParticle for the electrons and also for the ions.  So when ColoumbPBCAB was called, the per-particle stuff was were already activated.
Though I could construct some unlikely cases that would make it segfault (like leaving out the ion-ion potential).

PDoakORNL: Thanks @markdewing, can you make an issue for those segfaults.

PDoakORNL: The remaining question is whether we drop the ParticleSet::getPerParticleSKState method.  I'd just like to drop it. Someone reading it in a year is going to wonder why its there.

ye-luo: Storing SK data per particle instead of per specie needs a lot of memory. In a simulation of solide, is not a problem. In a simulation of 2D material, I have seen 3GB memory just taken by SK temporal storage which can be reduced by reduction over species. That is why the consumer is responsible to turn on the per particle setting.

ye-luo (commented):
1. add an assertion at evaluate_sp original place to check the status of SK storage per particle.
2. document your understanding of the loop starting from 1 above the loop.

All the rest looks good to me.

src/QMCHamiltonians/CoulombPBCAA.cpp
ye-luo: I pushed more documentation in StructFact.h

ye-luo:
ping @markdewing

markdewing:
working on it

jtkrogel:
The energy density estimator is currently working:
https://cdash.qmcpack.org/CDash/testDetails.php?test=14660932&build=229279
https://cdash.qmcpack.org/CDash/testDetails.php?test=14379785&build=229242

That estimator uses traces of single particle quantities, just like the energy density matrix in DensityMatrices1B, so I'm a little confused about what is unique here.

markdewing:
1. The EnergyDensityEstimator has a call to turnOnPerParticleSK in the `put` method.  
2. The required traces are weight, Kinetic, and LocalPotential, which I think will all exist by default.

ye-luo:
Start testing in-house

ye-luo (approved):

PDoakORNL (approved):

PR #3299, opened 2021-07-19
Separate oneAPI compiler from Clang settings.

ye-luo:
## Proposed changes
Separate oneAPI compiler from Clang settings as more precise option controls are needed.
Since IntelLLVM compiler ID is only available since cmake 3.20, for older CMake we detect the compiler by invoking `$CMAKE_CXX_COMPILER --version` and checking output.

Use `-ffp-model=precise` in mixed precision to make CPU mixed precision tests green.

- Build related changes

- No

dell-laptop

## Checklist


CMake/IntelCompilers.cmake
```diff
@@ -1,79 +1,132 @@
...
 
-# Set prefetch flag
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -qopt-prefetch")
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fstrict-aliasing")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fstrict-aliasing -D__forceinline=inline")
 
-if(MIXED_PRECISION)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -prec-sqrt")
+  # Set extra optimization specific flags
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fomit-frame-pointer")
```
PDoakORNL, line 46: as mentioned in #3300 a debug build with -fomit-frame-pointer is no desirable.

PDoakORNL (changes requested):
This looks good except for imposing -fomit-frame-pointer on Debug builds.

CMake/IntelCompilers.cmake
ye-luo: Changed. So only debug build gets `-fno-omit-frame-pointer` added. In other cases, stick to the compiler default.

PDoakORNL (approved):

PR #3300, opened 2021-07-20
Debug build use consistent options.

ye-luo:
## Proposed changes

A Debug build is intended to debug QMCPACK source code rather than a compiler option. For this reason, Debug build should carry compiler options as similar to Release build.
GNU/Clang -ffast-math and NVHPC -fast are applied to Debug build as well. So apply them direct to CMAKE_CXX_FLAGS

- Build related changes

- No

epyc-server

## Checklist

### DIFF
--- a/CMake/ClangCompilers.cmake
+++ b/CMake/ClangCompilers.cmake
@@ -55,10 +55,8 @@ if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 10.0)
 endif()
 
 # Set extra optimization specific flags
-set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -fomit-frame-pointer -ffast-math")
-set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fomit-frame-pointer -ffast-math")
-set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} -fomit-frame-pointer -ffast-math")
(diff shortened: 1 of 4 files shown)


CMake/ClangCompilers.cmake
```diff
@@ -66,14 +66,12 @@ if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 10.0)
 endif()
 
 # Set extra optimization specific flags
-set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -fomit-frame-pointer -ffast-math")
-set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fomit-frame-pointer -ffast-math")
-set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} -fomit-frame-pointer -ffast-math")
-set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -fomit-frame-pointer -ffast-math")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fomit-frame-pointer -ffast-math")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fomit-frame-pointer -ffast-math")
```
PDoakORNL, line 70: -fomit-frame-pointer is no good for debugging.
hence the split.

PDoakORNL (changes requested):
Default Debug build should have -fno-omit-frame-pointer

prckent:
Concurring with Peter on wanting (actually needing) the frame-pointer in the debug build.

PDoakORNL (approved):

PDoakORNL:
It's unclear to me that we have the best gcc and pgi debug flags but I think the exploration of that belongs in another PR.

ye-luo:
Start testing in-house

PR #3308, opened 2021-07-23
Make function for previous default linear optimizer

markdewing:
Prior to the introduction of the different "shift" optimizers, the code for the linear method was in "run".  

This PR moves that code to a separate function (putting it on the same organizational footing as the other linear method variants).

I wasn't sure what to call the previous algorithm while naming the function, so I called it "previous default".

- No

Desktop

## Checklist


src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.h
```diff
@@ -87,6 +87,10 @@ class QMCFixedSampleLinearOptimizeBatched : public QMCLinearOptimizeBatched,
   // perform optimization using a gradient descent algorithm
   bool descent_run();
 
+  // Previous default linear optimizer
```
ye-luo, line 90: I first got confused what "previous" means. Then I realized that it refers to pre "shift" implementations. So please add a note listing the names of "previous"  optimizer.

ye-luo: The change is a good move.

ye-luo: "rescale" is also contained. So I don't get "default".

markdewing: It was the code that ran prior to the introduction of the "shift" implementations.  Both "rescale" and "quartic" options are included - but they seem more like sub-options than different algorithm implementations.   I said in the description I didn't know what to collectively call this code :)
After the "shift" implementations it seems like this code is run if MinMethod is not in one of those if-tests, so it feels like this code is the fall-through or "default" case (I realize it's not really because MinMethod is initialized to OneShiftOnly)

markdewing: Maybe the function should be named "previous_linear_method_run"?

ye-luo: I'm OK with "previous_linear_methods_run" (I changed to prural) and in the documentation, explain that "previous" means.
Or may be change all the function names from XXX_run to run_XXX?

markdewing: Besides "rescale" and "quartic", it appears the code does something else if neither of these is specified.  It calls "lineoptimization2" instead of "lineoptimization3" (what "quartic" uses).   I think calling it "previous_linear_methods_run" is good enough, and I mention "rescale" and "quartic" in the comment.

ye-luo (commented):
LGTM. Need non-ANL approval

prckent (approved):
The .cpp part really ought to be a switch - I don't like how the condition on current_optimizer_type_ to reach the previous linear methods run is not clear.

I saw this code a while back and consider it emblematic of the historical lack of clarity over which optimizers are supported and how new ones are added. We could have a discussion about deleting some of the old optimizers and then ensuring that we have good deterministic test + CI coverage on those that remain...

PR #3324, opened 2021-07-29
Adjust multi walker resource acquire/release API in TWF and Ham

ye-luo:
## Proposed changes
Adopt the same multiwalker resource management API as  #3015 for ParticleSet.
The benefit of added arguments is allowing managing views when resources are acquired and released.
It makes certain optimization possible.
Remove the resource redistribution in DMC T-moves. More static behavior should reduce potential bugs.

If something goes wrong, I should see segfault but both deterministic tests and performance tests (intended for T-moves) runs happily.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCDrivers/DMC/DMCBatched.cpp
+++ b/src/QMCDrivers/DMC/DMCBatched.cpp
@@ -78,95 +78,90 @@ void DMCBatched::advanceWalkers(const StateForThread& sft,
                                                                 crowd.get_walker_hamiltonians());
 
   ResourceCollectionTeamLock<ParticleSet> pset_res_lock(crowd.getSharedResource().pset_res, walker_elecs);
+  ResourceCollectionTeamLock<TrialWaveFunction> twfs_res_lock(crowd.getSharedResource().twf_res, walker_twfs);
+  ResourceCollectionTeamLock<QMCHamiltonian> hams_res_lock(crowd.getSharedResource().ham_res, walker_hamiltonians);
 
   {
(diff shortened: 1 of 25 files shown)


ye-luo:
start testing in-house

PDoakORNL (commented):
I still think this is basically a good idea as I thought back with #3015 but since 
then I have many tests (which are still stranded in my WIP PR) which check the integrity of the single walker API's in the batch context i.e. called through DiracBatchedDeterminant.

This needs code like this

```c++
ResourceCollection res_col("test_determinant");
psi.createResource(res_col);
psi.acquireResource(res_col);
...
double logpsi = psi.evaluateLog(elec_);
...
  std::unique_ptr<TrialWaveFunction> psi_clone(psi.makeClone(elec_clone));

  elec_clone.update();

  res_col.rewind();
  psi.releaseResource(res_col);
  res_col.rewind();
  psi_clone->acquireResource(res_col);
  double logpsi_clone = psi_clone->evaluateLog(elec_clone);
...
  res_col.rewind();
  psi_clone->releaseResource(res_col);
  res_col.rewind();
  psi.acquireResource(res_col);
...
  RefVectorWithLeader<TrialWaveFunction> wf_ref_list(psi, {psi, *psi_clone});
...
TrialWaveFunction::mw_evaluateLog(wf_ref_list, p_ref_list);
...
```
Now I need to make a RefVectorWithLeader with the psi I want to use with the single walker API even though I'm not going to use it. Still if one can then avoid all the rewind's that would be nice.

I still think passing the resource to the functions is looking better compared these semantics.

PDoakORNL (approved):
After discussion this looks good to me.

ye-luo:
start testing in-house

PR #3333, opened 2021-08-02
Multiple fixes to multi walker shared resource use and disable ParticleSet::mw_accept_rejectMove not in forward  mode

ye-luo:
## Proposed changes
1. Some unit tests and the batched optimization driver relies on ad-hoc creation of multi-walker resource.
Such behavior is discouraged. Fixed in this PR.
2. When ParticleSet::mw_accept_rejectMove is not used in forward mode, there is potential bugs when data movements behavior changes for example, the upcoming offload J2. This can be fixed but penalizes certain code paths. Because this is only used in unit test, I just disabled this mode for now and will revisit when J2 landed.

- Bugfix

- No

epyc-server

## Checklist


src/Particle/SoaDistanceTableAAOMPTarget.h
```diff
@@ -120,6 +120,13 @@ struct SoaDistanceTableAAOMPTarget : public DTD_BConds<T, D, SC>, public Distanc
   const DistRow& getOldDists() const override { return old_r_; }
   const DisplRow& getOldDispls() const override { return old_dr_; }
 
+  const RealType* getMultiWalkerTempDataPtr() const override
```
PDoakORNL, line 123: What's the reason for this raw pointer?  There are no docs here and the documentation in the base class just says its to the "full distance table data"  But here it is to mw_mem->nw_new_old_dist_displ.data().  But I don't find it used.

I couldn't get an idea of what is going on from just reading DistanceTable code so there is likely some documentation on its mw resources needed as well.

PDoakORNL (changes requested):
The strange raw access to distance table data what is that about?

src/Particle/SoaDistanceTableAAOMPTarget.h
ye-luo: My bad. I copied the code but didn't update the documentation for getMultiWalkerTempDataPtr.
Currently I need two functions to access multi walker data
```
getMultiWalkerTempDataPtr
getMultiWalkerDataPtr
```
Optimally it should be replaced with proper containers. Thank you for the question. I have some ideas now but will probably try to clean them up after I finish the J2 offload work.
Documentation will be updated in the next PR.

PDoakORNL: Ok

PDoakORNL (approved):

ye-luo:
start testing in-house

PR #3334, opened 2021-08-02
Nexus: expand eshdf features

Paul-St-Young:
## Proposed changes

No change to default behavior. Options added:

1. --fnk3d outputs the 3D n(k) to the specified hdf file
2. --ispin=0,1 targets spin channel
3. --ncore and --nval define ``active'' orbitals that contribute to n(k)

- New feature

- No

workstation, Xeon 6234, CentOS 7

## Checklist

### DIFF
--- a/nexus/bin/eshdf
+++ b/nexus/bin/eshdf
@@ -74,7 +74,7 @@ def user_error(msg):
 
 
 
-def read_eshdf_nofk_data(filename,Ef):
+def read_eshdf_nofk_data(filename,Ef,ncore=0,nval=-1):
     from numpy import array,pi,dot,sqrt,abs,zeros
     from numpy.linalg import inv,det
     from unit_converter import convert
@@ -119,7 +119,10 @@ def read_eshdf_nofk_data(filename,Ef):
             spin = h.get_path(path)
             eig = convert(array(spin.eigenvalues),'Ha','eV')
             nst = h5int(spin.number_of_states)
-            for st in range(nst):
+            nlast = nst
+            if nval > 0:
+              nlast = ncore + nval
+            for st in range(ncore, nlast):
                 e = eig[st]
                 if e<E_fermi:
                     stpath = path+'/state_{0}/psi_g'.format(st)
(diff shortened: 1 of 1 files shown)


jtkrogel:
The functionality here looks good.  The intended usage of `eshdf` is `eshdf [mode] [options]`.  Please can you move the new functionality to a new function at the same level as `kinetic` (e.g. `write_nk`) that will process only what is needed for n(k) according to the new options you introduce?

nexus/bin/eshdf
```diff
@@ -183,6 +186,19 @@ def kinetic():
                       action='store_true',default=False,
                       help='Print per orbital kinetic energies (default=%default).'
                       )
+    parser.add_option('--ispin',dest='ispin',default=0,
```
jtkrogel, line 189: Pull parser options into new function e.g. `write_nk`

nexus/bin/eshdf
```diff
@@ -297,6 +315,28 @@ def kinetic():
             #end for
         #end for
     #end if
+
+    if opt.fnk3d is not None:
+        ispin0 = int(opt.ispin)
```
jtkrogel, line 320: Move contents to new function (e.g. `write_nk`).  The `--fnk3d` option could simply be something like `-o` and `--outfile` (not a big deal though).

jtkrogel (changes requested):

Paul-St-Young:
@jtkrogel  I split out the `write_nk` subcommand like you suggested.
I also pulled out two option parsing subroutines from `kinetic()` to reuse in `write_nk()`
1. `first_eshdf_file`
2. `interpret_nk_options`

Please see if this PR is now satisfactory.

jtkrogel:
@Paul-St-Young This looks OK now.  Please can you just quickly confirm that running `eshdf kinetic ...` produces the same output as before?

Paul-St-Young:
@jtkrogel yes, I can confirm `eshdf kinetic` works the same as before on my example wf h5 file.

PR #3344, opened 2021-08-05
Direct inversion only for ENABLE_OFFLOAD=1 ENABLE_CUDA=1

PDoakORNL:
## Proposed changes
Backport of batched direct inversion using CUDA, returns DiracMatrix per walker for single walker API.  Removes ability to pass resource in single walker API because openmp target doesn't need it and its awkward in the extreme to use the ResourceLock over refvectorwithleader and resource to put the resource into the leader state. 

Still some remenants of refactoring I would prefer to leave if there is actually interest in getting ENABLE_CUDA=1 ENABLE_OFFLOAD=0 working.  Although I now realize any acceleration not using openmp target is controversial and we should probably have a full discussion of whether abstractions or changes in code structure that make both that and offload possible are desirable.

Testing has been reduced since it was being frequently broken. There is a enough testing. Offload J2 turned off by default since it segfaults on my compiler / cuda version combination.

- New feature
- Build related changes
- Documentation changes

- No

Leconte

## Checklist


PDoakORNL:
I'll come back to this when I have time to deal with ifdef garbage in SlaterDetBuilder

PR #3346, opened 2021-08-06
add find filesystem and its usage

PDoakORNL:

## Proposed changes
Deal with the fact that filesystem wasn't treated as part of the c++ stdlib until gcc9.

I don't deal with multi using an obsolete path to the include.

I think something like this is needed
```c++
#if __has_include(<filesystem>)
#include <filesystem>
namespace filesystem = std::filesystem;
#else
#include <experimental/filesystem>
namespace filesystem = std::experimental::filesystem;
#endif
```

- Build related changes

- No

leconte
## Checklist

### DIFF
--- a/CMake/FindFilesystem.cmake
+++ b/CMake/FindFilesystem.cmake
@@ -1,0 +1,258 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
+# file Copyright.txt or https://cmake.org/licensing for details.
+
+#[=======================================================================[.rst:
+
+FindFilesystem
+##############
+
+This module supports the C++17 standard library's filesystem utilities. Use the
+:imp-target:`std::filesystem` imported target to
+
+Options
+*******
+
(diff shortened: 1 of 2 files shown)


PDoakORNL:
I took this slack message on general to mean it was needed.

```
Wiliam F Godoy
On the C++ and Python conversation, keep in mind that things that have been in Python for a while like std::filesystem, std::optional are only available since C++17. But I don't know how much are needed for the current discussion. std::filesystem is another library. IIRC, @Steven Hahn ran into this before: https://gitlab.kitware.com/cmake/cmake/-/issues/17834

GitLabGitLab
Portable linking for C++17 std::filesystem (#17834) · Issues · CMake / CMake
If using CMAKE_CXX_STANDARD=17 and std::filesystem, GCC requires linking of an extra library: stdc++fs. I think I read a while back the intention might be to fold this back...
Author
Roger Leigh
```

I now see I might have been in error.

ye-luo:
We can revisit when a need shows up.

Issue #3312, opened 2021-07-26
ASAN/LSAN leaks in deterministic tests


williamfgc:
**Describe the bug**
Complement fixes reported in #3083 for unit tests to extend ASAN/LSAN CI checks to deterministic tests
Create reference issue for future PR fixes

**To Reproduce**
1. `develop` branch
2. cmake command:
   ```
   CCompiler=clang
   CXXCompiler=clang++

  CC=$CCompiler CXX=$CXXCompiler cmake -GNinja \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
  -DENABLE_SANITIZER=asan \
  -DQMC_COMPLEX=0 \
  -DQMC_MPI=0 \
   ${QMC_SRC_DIR}
```
3. Build: `ninja`
4. Tests: `ctest -L deterministic`

**Expected behavior**
Tests shouldn't fail due to memory leaks. 
Currently: 
```
The following tests FAILED:
	127 - deterministic-He_ae-opt-1-1 (Failed)
	129 - deterministic-He_ae-opt_vmc-1-1 (Failed)
	131 - deterministic-He_ae-opt-batch-1-1 (Failed)
	133 - deterministic-He_ae-opt_vmc-batch-1-1 (Failed)
	182 - deterministic-LiH_solid_1x1x1_pp-gamma-vmc_hf_noj-1-1 (Failed)
	397 - deterministic-diamondC_1x1x1_pp-vmc-estimator-density-1-1 (Failed)
	399 - deterministic-diamondC_1x1x1_pp-dmc-estimator-density-1-1 (Failed)
	401 - deterministic-diamondC_1x1x1_pp-vmc-estimator-spindensity-1-1 (Failed)
	403 - deterministic-diamondC_1x1x1_pp-dmc-estimator-spindensity-1-1 (Failed)
	405 - deterministic-diamondC_1x1x1_pp-vmc-noJ-estimator-1rdm-1-1 (Failed)
	407 - deterministic-diamondC_1x1x1_pp-vmc-J2-estimator-1rdm-1-1 (Failed)
	409 - deterministic-diamondC_1x1x1_pp-dmc-estimator-1rdm-1-1 (Failed)
	411 - deterministic-diamondC_1x1x1_pp-vmc-estimator-momentum-1-1 (Failed)
	413 - deterministic-diamondC_1x1x1_pp-dmc-estimator-momentum-1-1 (Failed)
[...]

PR #3350, opened 2021-08-09
Fix converter leaks and undefined behavior

williamfgc:
## Proposed changes

Address remaining leaks and undefined behavior in converter tests. 
Move local container to a static scope as sanitizer finds out of scope use. 
Cache/thread-friendly, can't use `constexpr static std::array` due to lack of `inline constexpr` in C++14 see [here](https://stackoverflow.com/a/28846608/470441).
Add bounds check in valence calculation which creates a Heinsenbug due to undefined behavior in `converter_test_Bi_dirac`. 
converter tests should pass ASAN/LSAN checks after this PR.
Related to #3312

- Bugfix

- Yes

Ubuntu 20.04, must pass CI

## Checklist


src/QMCTools/QMCGaussianParserBase.cpp
```diff
@@ -475,8 +479,11 @@ xmlNodePtr QMCGaussianParserBase::createIonSet()
       hout.push(SpecieID.str().c_str(), true);
       int z          = static_cast<int>(ionSpecies(AtomicNumberIndex, i));
       double valence = ionSpecies(IonChargeIndex, i);
-      if (valence > CoreTable[z] && FixValence != true)
-        valence -= CoreTable[z];
+      if (z < gCoreTable.size())
+      {
+        if (valence > gCoreTable[z] && !FixValence)
+          valence -= gCoreTable[z];
+      }
```
ye-luo, line 486: Please add else and throw error.

williamfgc: @ye-luo a couple of things as it seems like a larger effort due to unintended consequences:
1. `converter_test_Bi_dirac` would trigger this error as `z` is going out of bounds (tests pass as gcc/clang are permissive and zero initialize memory for the most part). 
2. Throwing an exception generates leaks due to the skipped `xmlFree*` lines (mixing C++ and C)

prckent: The original code is simply wrong if someone runs an element heavier than Xe.  We should error abort for z too large and print a meaningful error message.

prckent (changes requested):

prckent:
I think that the original code is simply wrong if someone runs a too heavy element (>Xe). We run out of lookup table and should guard against it. e.g. Error abort and print a meaningful message.
(Deleted original comment by accident)

src/QMCTools/QMCGaussianParserBase.cpp
camelto2: I think adding `FixValence = true;` to the DiracParser constructor would avoid this issue for the converter test. The valence info is read and set correctly in the ascii parsing of the dirac output. This is similar to the QPParser and LCAOHDFParser constructors

williamfgc: I'll try @camelto2 suggestion to have at least a leaky exception (but no test doing it). Thank you all.

williamfgc:
This should be ready, the last commit prevents a leak for the original failing test and `FixValence=true` fixes it. I don't see converter leaks with this PR. Thanks!

camelto2 (approved):
LGTM

ye-luo (changes requested):
Please add documentation for FixValence. Not obvious what it means. What actions are taken or not taken if set true/false.

williamfgc:
@ye-luo good point, `QMCGaussianParserBase::FixValence` predates this PR. I tried finding it in doxygen [docs](https://docs.qmcpack.org/doxygen/doxy/d8/d2c/a05128.html#aa31023af5fb1ca1ca8fd9b676cefed0e) without luck. It seems like a `bool` flag only used in the lines of code this PR touches. That's another reason for requesting @camelto2 review on this as the `DiracParser` was recently added, I also see that `LCAOHDFParser` sets it to `true`. I'll defer to you and @prckent for the original motivation as I'll probably get it wrong :)

ye-luo:
@camelto2 It will be super helpful if you can put the closest definition here and then @williamfgc can finalize this PR.

camelto2:
> @camelto2 It will be super helpful if you can put the closest definition here and then @williamfgc can finalize this PR.

If I'm being honest, I'm not entirely sure why we need it at all. it is written to the xml as the 'valence' tag, but what is actually used internally by the code is 'charge'

What I think it it trying to do is add some definition of a distinction between the total number of electrons  vs. the  valence electrons in the outermost shell. In the example of say a BFD pseudopotential for O where we remove the He core, you have a total of 6 electrons with 2 removed. the 'charge' is read from the GAMESS output as 6, and the atomic number is 8. Then for GAMESS FixValence is false since it doesn't set it in the constructor, so it looks up in the table and sees that gCoreTable[z=8] = 2. So then it takes the actual number of electrons which is 6 and subtracts 2 to give 4. So nominally, the charge remains as 6,  but there are only 4 "valence" p electrons in this case while the 2s electrons would be considered semi-core. So that is what gets written to the structure.xml file. But to me it really isn't a meaningful distinction here, the only thing the code cares about the 'charge' element, which is correctly set by reading the GAMESS file. At least that is what I gather by quickly looking at code.

I think we could just drop it altogether and just leave 'valence' and 'charge' equal to one another for the ion species. That is what happens when FixValence is true. "valence" isn't even used by ParticleSet, so I don't even know why it is written to any of the xml files in the first place

ye-luo:
Fixing/removing valence and charge input tags needs to be explored separately.

prckent (approved):
I agree with Ye's above comment. Explore removing/updating the supported tags in a separate issue/PR.

ye-luo:
Test this please

PR #3352, opened 2021-08-10
[WIP]Add HDF file for inputs to linear method matrices

markdewing:
Add the parameter "output_matrix_inputs_hdf" to write out the per-sample values for the wavefunction values, derivatives, Hamiltonian derivatives, etc.   These values are used in `fillOverlapHamiltonianMatrices`.  Along with the file from
"output_linear_matrices_hdf", these files can be used in the `fillFromH5nunit` test.

One set of files is included in this check-in as a regression test. It should be straightforward to create additional sets of files for testing locally, for additional test coverage when making changes to `fillOverlapHamiltonianMatrices`.
(We don't want to increase the size of the repository unnecessarily)

- New feature

- No

desktop

## Checklist


ye-luo:
(paused ALCF CI for troubleshooting)

PDoakORNL:
While the binary files for this are politely small I think we should consider a real solution for the repo and binary files. #3354

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -724,6 +725,29 @@ QMCCostFunctionBatched::Return_rt QMCCostFunctionBatched::fillOverlapHamiltonian
 {
   ScopedTimer tmp_timer(fill_timer_);
 
+  hdf_archive hout;
+  if (output_matrix_inputs_hdf_)
```
PDoakORNL, line 729: It seems like when this is set true if I have multiple ranks its going to be a problem.  Am I missing something.

PDoakORNL (changes requested):
Question about feature in multi rank case.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
markdewing: Yes, it's only for one rank.  I'm not sure how to handle multiple ranks, or if that's a case that's worth handling.

PDoakORNL: Just realized if this only used with test programs it should removed from the input and the output argument set to true from the test program. Then some documentation in the code would suffice.

But if its is in the user facing input there must be a fatal error if used on multiple ranks or it needs to do something like put the rank number into the file name and be documented as writing for each rank.

markdewing:
I'm going to mark this as WIP since it probably needs some work on the binary file storage.   The test it enables is targeted at one function.  I can probably move forward using this code/testing locally.

markdewing:
Closing this as some of the testing parts are replaced by #3372.
The commits in this PR will need to be redone to remove the binary files.

PR #3357, opened 2021-08-13
Adjust scalar/vector math function handling

ye-luo:
## Proposed changes
Part A:
sincos function is used both inside and outside openmp offload region.
When used inside target region, the generic call signature defined in glibc is expected.
We need to be sure they can be compiled and linked on both host and device.

On the other hand, vendor may provide optimized functions requires call redirection like amd_sincos.
If we use inline function, this pollutes the compilation of target region. For this reason, This PR splits math functions in two namespaces.
1. qmcplusplus for CPU only.
2. omptarget for being used inside target region.

Part B:
When doing the above work, I reviewed all the scalar vector math handling. To reduce entanglement, `QMC_MATH_VENDOR` is introduced with providers GENERIC, INTEL_VML, AMD_LIBM, IBM_MASS. `ENABLE_MASS` has been removed.

When MKL is detected in BLAS/LAPACK, INTEL_VML will be picked as in the past. But if GENERIC is requested, the use of VML can be skipped. In general, no change for users but we have more precise control.

Another note, I explored the AMD LibM but it is slower than Clang inlined functions. The status of vector functions are also not clear, https://github.com/amd/aocl-libm-ose/issues/8 So it is added only for experiments and should be avoided for production use.

Part C:
Added a unit test for StructFact. it is very useful to quickly check sincos performance.

- Build related changes
- Testing changes
- Documentation changes

- No

epyc-server, dell-laptop, summit

## Checklist


PDoakORNL (approved):
This looks good to me.  The cmake improvements are welcome.

ye-luo:
Start testing in-house

PR #3359, opened 2021-08-16
Clean up ParticleSet/SK

ye-luo:
## Proposed changes
This is mostly non-functional change before I enable structure factor evaluation with offload.
1. Only allow ParticleSet::createSK being called if SK is nullptr.
2. change KContainer to shared_ptr in StructFact as the data is identical across walkers.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/LongRange/EwaldHandler3D.cpp
+++ b/src/Particle/LongRange/EwaldHandler3D.cpp
@@ -40,9 +40,9 @@ void EwaldHandler3D::initBreakup(ParticleSet& ref)
   app_log() << "  Sigma=" << Sigma << std::endl;
   Volume     = ref.Lattice.Volume;
   PreFactors = 0.0;
-  fillFk(ref.SK->KLists);
-  fillYkgstrain(ref.SK->KLists);
-  filldFk_dk(ref.SK->KLists);
+  fillFk(ref.SK->getKLists());
+  fillYkgstrain(ref.SK->getKLists());
+  filldFk_dk(ref.SK->getKLists());
 }
 
 EwaldHandler3D::EwaldHandler3D(const EwaldHandler3D& aLR, ParticleSet& ref)
@@ -51,7 +51,7 @@ EwaldHandler3D::EwaldHandler3D(const EwaldHandler3D& aLR, ParticleSet& ref)
   SuperCellEnum = aLR.SuperCellEnum;
 }
 
-void EwaldHandler3D::fillFk(KContainer& KList)
+void EwaldHandler3D::fillFk(const KContainer& KList)
 {
   Fk.resize(KList.kpts_cart.size());
   Fkg.resize(KList.kpts_cart.size());
--- a/src/Particle/LongRange/StructFact.cpp
+++ b/src/Particle/LongRange/StructFact.cpp
@@ -24,7 +24,7 @@ namespace qmcplusplus
 {
 //Constructor - pass arguments to KLists' constructor
 StructFact::StructFact(ParticleSet& P, RealType kc)
-    : DoUpdate(false), SuperCellEnum(SUPERCELL_BULK), StorePerParticle(false)
+    : DoUpdate(false), SuperCellEnum(SUPERCELL_BULK), KLists(std::make_shared<KContainer>()), StorePerParticle(false)
 {
   if (qmc_common.use_ewald && P.LRBox.SuperCellEnum == SUPERCELL_SLAB)
   {
@@ -41,9 +41,9 @@ StructFact::~StructFact() {}
 void StructFact::UpdateNewCell(ParticleSet& P, RealType kc)
 {
   //Generate the lists of k-vectors
-  KLists.UpdateKLists(P.LRBox, kc);
+  KLists->UpdateKLists(P.LRBox, kc);
   //resize any array
-  resize(P.getSpeciesSet().size(), P.getTotalNum(), KLists.numk);
+  resize(P.getSpeciesSet().size(), P.getTotalNum(), KLists->numk);
   //Compute the entire Rhok
   FillRhok(P);
 }
@@ -81,7 +81,7 @@ void StructFact::FillRhok(ParticleSet& P)
   rhok_r = 0.0;
   rhok_i = 0.0;
   //algorithmA
-  const int nk = KLists.numk;
+  const int nk = KLists->numk;
   if (StorePerParticle)
   {
     // save per particle and species value
@@ -95,7 +95,7 @@ void StructFact::FillRhok(ParticleSet& P)
 #pragma omp simd
       for (int ki = 0; ki < nk; ki++)
       {
-        qmcplusplus::sincos(dot(KLists.kpts_cart[ki], pos), &eikr_i_ptr[ki], &eikr_r_ptr[ki]);
+        qmcplusplus::sincos(dot(KLists->kpts_cart[ki], pos), &eikr_i_ptr[ki], &eikr_r_ptr[ki]);
         rhok_r_ptr[ki] += eikr_r_ptr[ki];
         rhok_i_ptr[ki] += eikr_i_ptr[ki];
       }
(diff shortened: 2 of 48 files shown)


src/Particle/LongRange/StructFact.cpp
```diff
@@ -23,8 +23,12 @@
 namespace qmcplusplus
 {
 //Constructor - pass arguments to KLists' constructor
-StructFact::StructFact(ParticleSet& P, RealType kc)
-    : DoUpdate(false), SuperCellEnum(SUPERCELL_BULK), StorePerParticle(false)
+StructFact::StructFact(const ParticleSet& P, RealType kc)
```
PDoakORNL, line 26: Actual constructor arguments should be `SpeciesSet&, kc, ParticleLayout&`

src/Particle/LongRange/StructFact.cpp
```diff
@@ -69,19 +73,32 @@ void StructFact::resize(int ns, int nptcl, int nkpts)
...
+  ScopedTimer local(sk_leader.update_all_timer_);
+  for (int iw = 0; iw < sk_list.size(); iw++)
+    sk_list[iw].FillRhok(p_list[iw]);
+}
 
 
 /** evaluate rok per species, eikr  per particle
  */
-void StructFact::FillRhok(ParticleSet& P)
+void StructFact::FillRhok(const ParticleSet& P)
```
PDoakORNL, line 94: Bad method name I'd assume this filled with zeros or something.

What is the actual data dependence? The entire function has to be read in detail to figure it out.  Looks like it could just be a species set.

src/Particle/ParticleSet.cpp
```diff
@@ -811,11 +811,18 @@ void ParticleSet::mw_donePbyP(const RefVectorWithLeader<ParticleSet>& p_list)
   for (ParticleSet& pset : p_list)
   {
     pset.coordinates_->donePbyP();
-    if (pset.SK && !pset.SK->DoUpdate)
-      pset.SK->UpdateAllPart(pset);
     pset.activePtcl = -1;
   }
 
+  if (p_leader.SK && !p_leader.SK->DoUpdate)
```
PDoakORNL, line 817: Rename DoUpdate this reads as nonsense. I assume its a boolean.  SK would also really be better named structure_factor_, its very long lived and doesn't help the code be readable.

I do not like behavior control based on nullptr's but that's a bigger issue.

src/Particle/LongRange/EwaldHandler3D.cpp
```diff
@@ -40,9 +40,9 @@ void EwaldHandler3D::initBreakup(ParticleSet& ref)
   app_log() << "  Sigma=" << Sigma << std::endl;
   Volume     = ref.Lattice.Volume;
   PreFactors = 0.0;
-  fillFk(ref.SK->KLists);
-  fillYkgstrain(ref.SK->KLists);
-  filldFk_dk(ref.SK->KLists);
+  fillFk(ref.SK->getKLists());
```
PDoakORNL, line 43: I think practically everywhere KLists is has been touched.  Rename it.  KLists looks like a type that is a list of KList, `k_lists_`

KContainer clearly wants to be named KLists but couldn't because someone loved a member variable named KLists, great time to fix this as well.

src/Particle/LongRange/StructFact.cpp
```diff
@@ -69,19 +73,32 @@ void StructFact::resize(int ns, int nptcl, int nkpts)
 }
 
 
-void StructFact::UpdateAllPart(ParticleSet& P) { FillRhok(P); }
+void StructFact::UpdateAllPart(const ParticleSet& P)
```
PDoakORNL, line 76: Since we have this otherwise pointless wrapper for FillRhok this would be a good place to find out what that function actually takes as input.

src/Particle/LongRange/StructFact.cpp
```diff
@@ -69,19 +73,32 @@ void StructFact::resize(int ns, int nptcl, int nkpts)
...
+{
+  ScopedTimer local(update_all_timer_);
+  FillRhok(P);
+}
+
+void StructFact::mw_UpdateAllPart(const RefVectorWithLeader<StructFact>& sk_list,
```
PDoakORNL, line 82: Why isn't this mw_FillRhok or whatever name would actually give a clue up the stack what was happening.

PDoakORNL (changes requested):
The naming changes should really happen now, most of the code involving KLists and others is being touched in this PR. We must make some improvements in this regard as we go.

The shared_ptr I don't like, it's clear that the scope the KLists is the "system" i.e., which should also own the SpeciesSet and a number of other things. That needs to be a clear thing. The disaster of the input parsing and construction of a ParticleSet makes this difficult currently but the tech debt shouldn't always get pushed.

But @ye-luo I suspect that KLists storage for offload is driving this. If so that's an ok reason.

src/Particle/LongRange/EwaldHandler3D.cpp
ye-luo: I have changed KLists to k_lists_ inside StructFact.
I'm following the current naming rule for functions. 
https://qmcpack.readthedocs.io/en/develop/developing.html#member-function-names

src/Particle/ParticleSet.cpp
ye-luo: I have a quick assess of DoUpdate. It is used to control SK update during p-by-p but the only wavefunction piece which need SK is k-space backflow which doesn't seem to have any p-by-p implementation. However, those functions may still needed for NLPP. I'd like to defer DoUpdate related changes as more reading will be needed.  Renaming SK also need to move SK out of public and add accessor. Needs to be done separately.

src/Particle/LongRange/StructFact.cpp
ye-luo: Actually, only needs the number of particles, the number of species and the lattice. Changed.

src/Particle/LongRange/StructFact.cpp
ye-luo: The is function is temporarily kept. It can be deleted, once I implement mw_updateAllParts without calling it.

src/Particle/LongRange/StructFact.cpp
ye-luo: The data dependency is on all the particle coordinates. So I will change the argument to DynamicCoordinates to pull ParticleSet out of this class fully. It has to be done in a separate PR.

src/Particle/LongRange/StructFact.cpp
ye-luo: The plan is to keep updateAllPart and mw_updateAllPart and remove FillRhok.

ye-luo:
> The shared_ptr I don't like, it's clear that the scope the KLists is the "system" i.e., which should also own the SpeciesSet and a number of other things. That needs to be a clear thing. The disaster of the input parsing and construction of a ParticleSet makes this difficult currently but the tech debt shouldn't always get pushed.
> 

I think ParticleSet is largely our "system" right now. It contains

1. Lattice: lattice, the boundary conditions. There is also LRbox, long range box with Z scaled with more vacuum.
2. R, coordinates_: the position of particles
3. SK: structure factor which depends on both 1 and 2.

We may consider moving Lattice and KLists to the ParticleSetPool and use only reference inside ParticleSet as a refactor move. Then the reference from ion and electron ParticleSets should help reducing the amount of copy/state-tracking.

shared_ptr is only intended to shared between ParticleSet/SK clones to shared the same data on GPU.

PDoakORNL:
Some those things are one to many walkers, some are many to many, that's what the problem is. They don't belong in the same object. This doesn't fix that in fact it helps bake it in because now some of the runtime cost is turned into more tech debt. I don't think the shared_ptr is the right solution, leave this alone I think its an optimization for code I haven't even seen work.

PDoakORNL:
I'm going to make a show of faith that this will not be the beginning of abuse of shared_ptr to dodge lifetime and ownership issues. They should be sorted out.

PDoakORNL:
But I would like are response on why when the type gets changed we might not just as well change the name at the same time.  Anything accessing the KLists should be getting updated anyway so I don't see why splitting that across two PR's for the purposes of reverting is necessary.

ye-luo:
> But I would like are response on why when the type gets changed we might not just as well change the name at the same time. Anything accessing the KLists should be getting updated anyway so I don't see why splitting that across two PR's for the purposes of reverting is necessary.

I probably missed what you asked for. In StructFact class, KLIsts has been renamed as k_lists_ with a public accessor getKLists(), the return type is const. Is this what you are looking for?

PDoakORNL:
Yes.  I think renaming KContainer to KLists would make sense since it refers to KLists in methods that seem quite specialized to KLists. But perhaps I'm missing something.

PDoakORNL (approved):

ye-luo:
KContainer right now doesn't have ambiguity. I actually prefer KContainer over KLists as the class name. It is a structure holding various vectors of K. One vector in index, one vector in real-space.

ye-luo:
Start testing in-house

ye-luo:
Start testing in-house

PDoakORNL (approved):
LGTM

PR #3363, opened 2021-08-16
Introduce DeviceManager.

ye-luo:
## Proposed changes

DeviceManager is introduced to capture device information as an object although the object remains global.
Replace single function assignAccelerators which is full of ifdefs.

- No

jlse multi-GPU nodes.

## Checklist


prckent (changes requested):
For starters, please add comments/documentation detailing the purpose of this. I can potentially infer, but what do you envisage it to be? How will multiple GPUs be handled? e.g. We need a design suited for eventual spline buffer sharing / remote kernel execution as we have in the legacy CUDA.

ye-luo:
Added documentation in `DeviceManager.h`

A multi-GPU algorithm needs to know the the default device.
So it can query DeviceManager::global->getDefaultDeviceNum() for generic info.
If platform specific info is needed, query DeviceManager::global->getCUDADM().

Having a subset of GPUs forming a group should be based on MPI communicators. It should not assume all the visible GPUs participate in the group. We may have dual socket machines each with a few GPUs, we may only place a a group with in a socket and avoid cross-socket traffic.

PDoakORNL:
this is not an idiomatic way to define a singleton:

For c++17 and later
DeviceManager& get()
{
    static DeviceManager instance;
    return instance;
}

The first get() constructs the DeviceManager.
Subsequent access is through the get(), I've never seen the static pointer thing before.

The only advantage I can see for it is I'm sure when the DeviceManage constructor gets called, where as with the standard idiom I'm not sure.

ye-luo:
> this is not an idiomatic way to define a singleton:
> 
> For c++17 and later
> DeviceManager& get()
> {
> static DeviceManager instance;
> return instance;
> }
> 
> The first get() constructs the DeviceManager.
> Subsequent access is through the get(), I've never seen the static pointer thing before.
> 
> The only advantage I can see for it is I'm sure when the DeviceManage destructor gets called, where as with the standard idiom I'm not sure.

I guess this is singleton class vs singleton function. I need to initialize the static storage explicitly after MPI initialization. So I don't think what you pointed out can be applied here.

src/Platforms/DeviceManager.h
```diff
@@ -0,0 +1,89 @@
...
+ * MPI initialization in main().
+ */
+class DeviceManager
+{
+public:
+  /* constructor
+   * @param local_rank the rank id of the node local MPI communicator
+   * @param local_size the size of the node local MPI communicator
+   */
+  DeviceManager(int local_rank, int local_size);
```
PDoakORNL, line 54: I didn't completely get how this class worked when we talked yesterday, I assumed there was limited access to the constructor. In the example I gave yesterday the get() function goes into the singleton class and you make the constructor private. The reason you put the one static copy of your class in a local function is that "Initialization of function-local statics is guaranteed to occur only once even when called from multiple threads, and may be more efficient than the equivalent code using std::call_once."

The pattern is attributed to Scott Meyers. The singleton is constructed on first access. Since singletons are by convention only accessed through their acessor or the reference it returns there is no "trouble." There can be trouble in certain linkage scenarios the one I know if is if the calls to the singleton are in header files which are included from different compilation units.

PDoakORNL (approved):
Despite the pedantic discussion of singletons I couldn't resist, I just think reference to singletons should be taken out of the documentation. That will remove the confusion. This is a little better than the usual global because we do know when its is constructed. 

But I think the instance should not be held by a static pointer at global scope. Other QMCMain scope objects like this have been moved into QMCMainState and are explicitly passed as needed down the function chain as in the new drivers. 

This makes the dependence on QMCMain scope objects explicit and one can then reason about their accessibility to concurrent sections or parts of the code that should not be entangled with device code. I think it's likely in the future the device manager will be capable of doing more than just storing a pair of numbers and this will matter.  

I don't think it needs to be done in this PR.  Likely dependent PR's depend on the global access. But its a design defect that has been added to the code when we have been working toward removing global scope objects and dependencies. Individual Crowd contexts probably shouldn't have access to DeviceManager and the logic dealing with that information should be at the driver level.

src/Platforms/DeviceManager.h
ye-luo: std::call_once seems solving a different problem I'm looking at.
Now I changed the static DeviceManager::global to private.
initializeGlobalDeviceManager is for initialization
If being called multiple times, the second and later calls will throw error.
getGlobal() is for access the global instance.
It checks if global is nullptr.

ye-luo:
Start testing in-house

src/Platforms/DeviceManager.h
PDoakORNL: I'm satisfied with this. It's not threadsafe but it is good enough for us. The general rule in QMCPACK is that thread safety is the caller's responsibility. I will catch bad single thread application or test program logic which I think is the most likely issue.

There is nothing wrong with std::call_once (although I doubt it is compatible with openmp) for situations where you have a single resource that you are going to initialize lazily and you know that will be from a potentially concurrent context it works fine. You could easily build an abstraction that either used the stdlib object or implemented the functionality in openmp. I'm sure there are other uses as well.

PDoakORNL (approved):

ye-luo:
> There is nothing wrong with std::call_once (although I doubt it is compatible with openmp) 

it is a wrapper over mutex. OpenMP should not break base language. So you should never worry.

ye-luo:
@prckent let me know if the documentation still needs improvement.

PDoakORNL:
I agree open mp should not break the base language or the stdlib but It was my impression that OpenMP declared use of Concurrency features added in c++11 (basically the whole thread support library) including std::thread, std::mutex, futures, promises, etc as UB if openmp is used. Additionally I thought they also specified use of std::atomic as UB.

This impression comes from https://www.openmp.org/spec-html/5.0/openmpse7.html#x28-270001.7.

ye-luo:
> I agree open mp should not break the base language or the stdlib but It was my impression that OpenMP declared use of Concurrency features added in c++11 (basically the whole thread support library) including std::thread, std::mutex, futures, promises, etc as UB if openmp is used. Additionally I thought they also specified use of std::atomic as UB.
> 
> This impression comes from https://www.openmp.org/spec-html/5.0/openmpse7.html#x28-270001.7.

I think your concern is real. OpenMP implementation is usually done as one for C/C++/Fortran and there is no guarantee working with C++ concurrency. In reality on Linux, both OpenMP and libstdc++ are built on top of pthreads. I guess this is the reason why we don't get trouble with std:mutex our OpenMP codes.

ye-luo:
Start testing in-house

prckent (changes requested):
It looks to be OK, but can you/did you verify that an ensemble run is still OK? We are notably missing any tests of this functionality.

ye-luo:
> It looks to be OK, but can you/did you verify that an ensemble run is still OK? We are notably missing any tests of this functionality.

Default GPU assignment is based on the WORLD communicator instead of sub communicator in an ensemble run. So it is agnostic to ensemble runs.

Only when doing GPU sharing, the sub-communicator needs to be relied on to create GPU groups. Right now GPU sharing + ensemble job has never been supported even in the legacy CUDA.

ye-luo:
Start testing in-house

ye-luo:
Manually ran 4 MPI ranks with 2 jobs in the ensemble. All the four jobs are populated properly.

prckent:
I also confirmed that the ensemble runs are still "working". 

Topic for another day: Sadly deterministic tests give different results within the ensemble and as compared to non ensemble runs.

prckent (approved):

ye-luo:
Brainstorming a test. We create an ensemble test one writes to 001 and the other writes to 002. then the existing check scheme can be used.

prckent:
Test can recycle  a combination of different deterministic tests and their numerical check data. The headache is needing to write new custom cmake to setup the files, run the job, and to process the results.

PR #3372, opened 2021-08-18
Add larger test for fillOverlapHamiltonianMatrices

markdewing:
The data for this test is from 10 samples and 12 variational parameters.
The data is stored as text ("diamond_fill_data.h") to mesh better with the git repository.  It was converted from HDF to text with the convert_hdf_to_cpp.py script.

This PR replaces the test additions in #3352. 

## What type(s) of changes does this code introduce?
- Testing changes (e.g. new unit/integration/performance tests)

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
desktop

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- Yes. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #3373, opened 2021-08-19
Parallelize the filling of linear matrices

markdewing:
Parallelize the batch code version of fillOverlapHamiltonianMatrices function using ParallelExecutor.

The version in the non-batch code was parallelized with an omp parallel for over the `pm` loop.  There are no reduction variables or other sync needed because iterations are filling in different matrix entries.

Can be tested with #3372

- New feature
- No

desktop

## Checklist


PDoakORNL (dismissed):
fillOverlapHamiltonianMatrices needs some documentation both at the base class level and briefly for this implementation.

I think a threaded test of this would be worthwhile.  I seems like it would be fairly easy to produce with code that could be shared with the existing test for the matrix filling.

Otherwise this looks good to me, it doesn't appear to be inside another threaded context.

markdewing:
The test in #3372 checks the 1 and 2 thread cases.  The loop bounds can be extended for checking cases with more threads.

prckent:
Looks to me like the comments have all been addressed. Propose to approve after #3372 is merged.

prckent (approved):
Approving since the comments appear to have been addressed (thanks Mark), and further revision wouldn't be difficult in a later PR or affect other ongoing work.

prckent:
Start testing in-house

PR #3375, opened 2021-08-19
fix for get_first_neighbor if there is only one particle

camelto2:

## Proposed changes

This PR fixes an issue that was introduced in #3288 which causes the code to crash in the case where there is only one atom in the simulation cell. 

The distance tables function get_first_neighbor should only be called if there are actually neighbors, so an assertion is added in each of these functions. 

Also, in EinsplineSetBuilderESHDF.fft.cpp, now the get_first_neighbor is only called if there are multiple atoms in the simulation cell. 

- Bugfix

- No

my workstation, gnu 9.2.0

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
+++ b/src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
@@ -185,8 +185,9 @@ bool EinsplineSetBuilder::ReadOrbitalInfo_ESHDF(bool skipChecks)
     {
       AtomicCentersInfo.non_overlapping_radius[i] = std::numeric_limits<RealType>::max();
       //should only call get_first_neighbor to set non_overlapping_radius if there are more than one atom  in the cell
-      for (int j = 0; j < Super2Prim.size() && (Super2Prim.size() > 1); j++)
-      {
+      if (Super2Prim.size() == 1)
+        continue;
+      for (int j = 0; j < Super2Prim.size(); j++)
         if (Super2Prim[j] == i)
         {
           // set GroupID for each ion in primitive cell
@@ -209,7 +210,6 @@ bool EinsplineSetBuilder::ReadOrbitalInfo_ESHDF(bool skipChecks)
           AtomicCentersInfo.non_overlapping_radius[i] = 0.5 * r;
           break;
         }
-      }
     }
 
     // load cutoff_radius, spline_radius, spline_npoints, lmax if exists.
--- a/src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
+++ b/src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
@@ -182,7 +182,11 @@ bool EinsplineSetBuilder::ReadOrbitalInfo_ESHDF(bool skipChecks)
     const auto& ii_table = SourcePtcl->getDistTable(table_id);
     SourcePtcl->update(true);
     for (int i = 0; i < IonPos.size(); i++)
-      for (int j = 0; j < Super2Prim.size(); j++)
+    {
+      AtomicCentersInfo.non_overlapping_radius[i] = std::numeric_limits<RealType>::max();
+      //should only call get_first_neighbor to set non_overlapping_radius if there are more than one atom  in the cell
+      for (int j = 0; j < Super2Prim.size() && (Super2Prim.size() > 1); j++)
+      {
         if (Super2Prim[j] == i)
         {
           // set GroupID for each ion in primitive cell
@@ -205,6 +209,8 @@ bool EinsplineSetBuilder::ReadOrbitalInfo_ESHDF(bool skipChecks)
           AtomicCentersInfo.non_overlapping_radius[i] = 0.5 * r;
           break;
         }
+      }
+    }
 
     // load cutoff_radius, spline_radius, spline_npoints, lmax if exists.
     const int inner_cutoff_ind   = SourcePtcl->mySpecies.findAttribute("inner_cutoff");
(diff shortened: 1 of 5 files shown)


src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
```diff
@@ -182,7 +182,11 @@ bool EinsplineSetBuilder::ReadOrbitalInfo_ESHDF(bool skipChecks)
     const auto& ii_table = SourcePtcl->getDistTable(table_id);
     SourcePtcl->update(true);
     for (int i = 0; i < IonPos.size(); i++)
-      for (int j = 0; j < Super2Prim.size(); j++)
+    {
+      AtomicCentersInfo.non_overlapping_radius[i] = std::numeric_limits<RealType>::max();
+      //should only call get_first_neighbor to set non_overlapping_radius if there are more than one atom  in the cell
+      for (int j = 0; j < Super2Prim.size() && (Super2Prim.size() > 1); j++)
```
ye-luo, line 188: better to have `Super2Prim.size() > 1` condition standalone. and make it more explicitly visible.

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
```diff
@@ -182,7 +182,11 @@ bool EinsplineSetBuilder::ReadOrbitalInfo_ESHDF(bool skipChecks)
...
+      {
```
ye-luo, line 189: No need of this pair {}

prckent:
Thanks for spotting and fixing this Cody.

I am disappointed that this bug got in. Exactly what would be needed to hit this in a test? I think we need some 1 particle tests, e.g. 1 atom 1 electron case. @ye-luo Any thoughts on an existing test we could clone to avoid adding more wavefunction files?

PDoakORNL:
Refactor until you can write a unit test. There appears to be no testing of this input parsing at all. You can generate the hdf5 file in the setup to the unit test it doesn't need to contain a real wavefunction.

lshulen:
Strongly disagree.  The breakage should be fixed.  Then a unit test should be a high priority.  We should not leave the code in a broken state while we await perfection.

prckent:
This is easy - we merge this fix. If Cody wants+is able to address Ye's minor comments, fine. But we can't leave the code broken, and the length of the ball of string/depth of can of worms could be massive here.

@camelto2 's choice on path forward.

ye-luo:
We were relying an undefined behavior before #3288 and #3288 indeed adds assertion and this broken corner case got caught. it is unreasonable to ask very PR to fix previously owned debts. We just need to be realistically maximize cost effectiveness.

@prckent we have atomO test in pw2qmcpack workflow, It is not hitting this issue because of "n n n". We chose to put it in pw2qmcpack workflow in order to avoid storing the h5. I also found most of our solid tests has two atoms, we need to diversify them.  For testing purpose, we actually can add a cheap one atom system with low cutoff and test deterministically only. If we decide to add one case, @Hyeondeok-Shin can help @camelto2 on this. @prckent what do you think?

PDoakORNL:
I just commented, I didn't request changes, I don't disagree with @lshulen and @prckent on the fix.

test_einset.cpp seems to encompass what we have for this so far although it is an integration test of SPOSet, SPOSetBuilderFactory, and the consequent subset of EinsplineSetBuilder. Whereas the test here should be straightforward and of EinsplineSetBuilder::ReadOrbitalInfo_ESHDF. When faced with a 1 particle ESHDF behave in the manner expected.  
Obviously there is a missing test for the no bugged case of >1 patricle ESHDF.

It's a bit of a challenge since the member function only mutates the internal state of ESB but that's part of the pain of having a testable code.

PDoakORNL:
I'll disagree with @ye-luo a bit here. I think when you have a bug, especially one that will potentially recur and it can be caught in a unit test, there should be a unit test added. It deserves at least as much credit as fixing the bug after all if it had been written already there wouldn't have been one.

prckent:
For a later PR: We can add a 1 atom case with low cutoff. There are already Li and H pseudopotentials in the repo that could be used for as a basis for this. We require both the statistical and deterministic versions. If the test added has one electron, the variance will be tiny, so a "short" statistical test will be accurate. It would be best to know + use the exact numerical result for this case. e.g. run HF in QE.

ye-luo:
> For a later PR: We can add a 1 atom case with low cutoff. There are already Li and H pseudopotentials in the repo that could be used for as a basis for this. We require both the statistical and deterministic versions. If the test added has one electron, the variance will be tiny, so a "short" statistical test will be accurate. It would be best to know + use the exact numerical result for this case. e.g. run HF in QE.

I think both Li and H pseudo need high PW cutoff to have reasonable variance. Better to avoid them.

ye-luo:
> I'll disagree with @ye-luo a bit here. I think when you have a bug, especially one that will potentially recur and it can be caught in a unit test, there should be a unit test added. It deserves at least as much credit as fixing the bug after all if it had been written already there wouldn't have been one.

I think once we got the simulation test with h5 provided, test_einset.cpp like unit test will be easy to add.

camelto2:
> Since there has been some discussion, just to summarize the situation: @camelto2 is welcome to update as suggested by @ye-luo above, but we can merge this and work on expanding the tests or making other improvements later.

sorry for taking so long on this, I was out on Friday so missed all of this discussion. I'll go ahead and fix the minor issues from Ye and think about an effective test for a future PR

src/QMCWaveFunctions/EinsplineSetBuilderESHDF.fft.cpp
camelto2: done

Hyeondeok-Shin:
I can add deterministic test for a single atom case discussed in here. single Li atom will be good to make the test with low cut-off. I'll work on it.

ye-luo:
Test this please

ye-luo (approved):

prckent (approved):

PR #3382, opened 2021-08-24
Downgrade Summit toolchain to cuda 10.1.243

ye-luo:
## Proposed changes
The default CUDA 11.0.3 after OS upgrade on Summit, caused frequent hanging with the offload code in production runs.
Stick to CUDA 10.1.243 for now.

PS: a small change on CUDA_NVCC_FLAGS is needed to allow passing additional arguments via command line.

- Build related changes

- No

Summit

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -773,7 +773,7 @@ if(QMC_CUDA OR ENABLE_CUDA)
       message(STATUS "Setting CUDA FLAGS=${CUDA_NVCC_FLAGS}")
     else(CUDA_NVCC_FLAGS MATCHES "arch")
       # Automatically set the default NVCC flags
-      set(CUDA_NVCC_FLAGS "-Drestrict=__restrict__;-DNO_CUDA_MAIN;-std=c++14")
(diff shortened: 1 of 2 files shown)


prckent:
Please put a note about "CUDA 11.0.3 caused frequent hanging with the offload code in production runs." in the file by the CUDA module, including which offload compiler version.

ye-luo:
@prckent comments added.

ye-luo:
Start testing in-house

prckent (approved):
Just to note here, which compiler is /usr/bin/gcc currently? 
What is the problem with `which gcc`? Why not (un)load specific gcc?

ye-luo:
> Just to note here, which compiler is /usr/bin/gcc currently?
> What is the problem with `which gcc`? Why not (un)load specific gcc?

CUDA 10.1 works only with `gcc <= 8` the system GCC is 8.3. This gcc is only used by nvcc internally. The Clang part still uses gcc/9.3.0. The nvcc option which supports using unsupported compiler is only introduced since CUDA 11.

PR #3392, opened 2021-08-26
Top level  CMakeLists.txt cleanup and remove QMC_VERBOSE_CONFIGURATION option

### LINKED ISSUE #3237: Drop message_verbose
Our customized message_verbose(XXX) can be replace with `message(VERBOSE XXX)` from cmake.
This feature starts in CMake 3.15 release on 2019-07-17
Although I don't know the next release date, it is unlikely before 2021-07-17. So I think it is safe to demand 3.15.
If 3.15 is really a concern, we can stay 3.14 but 3.14 user will always got the message printed as "VERBOSEXXX" but nothing really problematic.

ye-luo:
## Proposed changes
Top level CMakeLists.txt has been heavily edited to reorder its contents.
The new order is
1. CMake project related info
2. build/install options
3. compiler handling
4. library handling
5. final adjustment before parsing source tree
6. Test related settings

Also take care of message_verbose("abc") to message(VERBOSE "abc"). closes #3237
If QMC_VERBOSE_CONFIGURATION is still set, a CMake warning will be printed.

- Build related changes

- No

epyc-server

## Checklist


prckent:
Before merging this I would like to verify that the cdash drop is working correctly.

ye-luo:
> Before merging this I would like to verify that the cdash drop is working correctly.

I'm wondering if we should actually delete all the lines related to cdash drop because we use ctest_script and it has https://github.com/QMCPACK/qmcpack/blob/9818a47032c2c3822bc8a783943afd221283a11f/CMake/ctest_script.cmake#L260

ye-luo:
Start testing in-house

prckent:
Updating cdash support would be a change too far for this PR. Best tackled in another one.

prckent (approved):
Much cleaner!

ye-luo:
Start testing in-house

PR #3395, opened 2021-08-29
Fix ifdef for ENABLE_CUDA=ON ENABLE_OFFLOAD=OFF case

ye-luo:
## Proposed changes
`#elif ENABLE_CUDA` was causing problems. I should actually be `#elif defined(ENABLE_CUDA)`.
Anyway just drop all and use "DualAllocatorAliases.hpp"

- Bugfix
- Build related changes

- No

epyc-server

## Checklist


ye-luo:
Start testing in-house

PDoakORNL (approved):
LGTM

Issue #3383, opened 2021-08-24
Remove usage of shared_ptr<BackflowTransformation>


quantumsteve:
**Is your feature request related to a problem? Please describe.**

PR #3347 attempts to fix memory leaks of BackflowTransformation by using std::shared_ptr<BackflowTransformation> and relying on the reference counter to ensure each object is deleted.

**Describe the solution you'd like**

Pass either a `std::unique_ptr<BackflowTransformation>` or `BackflowTransformation&` in the constructor and remove member functions `::setBF(std::shared_ptr<BackflowTransformation>)` and `add(add(std::unique_ptr<Determinant_t> det, int ispin)` 

@ye-luo recommended MultiSlaterDeterminantWithBackflow and SlaterDetWithBackflow owns unique_ptr, while DiracDeterminantWithBackflow owns a reference.

**Describe alternatives you've considered**

I tried making these changes in #3347, but am struggling to implement it. Building functions weren't designed to have all `DiracDeterminantBase` objects constructed before the object that owns them. The inheritance hierarchy make something like `::setBF(std::unique_ptr<BackflowTransformation>)` and ::setBF(BackflowTransformation&)` error-prone.

**Additional context**

The last two commits here might be useful https://github.com/QMCPACK/qmcpack/compare/develop...quantumsteve:remove_setBF

PR #3397, opened 2021-08-30
Some cleanup in single det classes

ye-luo:
## Proposed changes
Remove add() which was used to add individual deteriminant to SingleDet class.
Remove set() which has too many logic holes.
All the necessary info will be passed via constructor.

This results in SlaterDet simplification. This is a warm up PR of a fix to #3383

- No

epyc-server

## Checklist


PDoakORNL:
This mostly looks good to me, I will get a review in this afternoon.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -116,18 +116,12 @@ std::unique_ptr<WaveFunctionComponent> SlaterDetBuilder::buildComponent(xmlNodeP
...
-      }
-      UseBackflow = true;
-      // creating later due to problems with ParticleSets
-      //BFTrans = new BackflowTransformation(targetPtcl,ptclPool);
-      BFTrans = nullptr;
-      BFnode  = cur;
-      // read xml later, in case some ParticleSets are read from hdf5 file.
-      //BFTrans->put(cur);
+      if (BFTrans)
+        myComm->barrier_and_abort("Only a single backflow block is allowed in the xml. "
```
PDoakORNL, line 120: Isn't this just a UniformCommunicateError? It appears myComm is only used to call this error mechanism I'd like to see far fewer objects dependent on Communicate.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -281,65 +268,36 @@ std::unique_ptr<WaveFunctionComponent> SlaterDetBuilder::buildComponent(xmlNodeP
...
-    return nullptr;
-  }
-  if (multiDet && (!multislaterdet_0 && !multislaterdetfast_0))
+  if (built_singledet_or_multidets)
+    return built_singledet_or_multidets;
+  else
   {
     //fatal
-    PRE.error("Failed to create a MultiSlaterDeterminant.", true);
+    PRE.error("Failed to create a Single Slater determinant or Multi Slater determinants object.", true);
```
PDoakORNL, line 298: Then there is this alternative PRE.error mechanism for fatal errors which then returns a nullptr.  As far as I can tell that will eventually cause an exception to be thrown in TrialWaveFunction::addComponent when a operator-> on the unique pointer returned results in rubbish.  The PRE.error message is lost. It would be much better to have an exception that could be trapped in the debugger at the site things went wrong. This would require more time to figure out.  For a fatal error just throw an appropriate exception where it occurs. If its a problem all ranks will encounter throw a UniformCommunicateError.

src/QMCWaveFunctions/Fermion/SlaterDetWithBackflow.h
```diff
@@ -31,17 +31,20 @@ class SlaterDetWithBackflow : public SlaterDet
...
 
   ///set BF pointers
-  void setBF(std::shared_ptr<BackflowTransformation> bf) override
+  void setBF(std::shared_ptr<BackflowTransformation> bf)
   {
     BFTrans = bf;
     for (int i = 0; i < Dets.size(); i++)
       Dets[i]->setBF(BFTrans);
+
+    if (BFTrans->isOptimizable())
```
PDoakORNL, line 46: This seems different from the semantics of Optimizable in the SlaterDet base class. What if none of the dets are optimizable, this still makes sense?

src/QMCWaveFunctions/PlaneWave/PWOrbitalBuilder.cpp
```diff
@@ -113,38 +112,33 @@ std::unique_ptr<WaveFunctionComponent> PWOrbitalBuilder::putSlaterDet(xmlNodePtr
...
     }
     cur = cur->next;
   }
 
   if (spin_group)
-    return sdet;
+    return std::make_unique<SlaterDet>(targetPtcl, std::move(dets));;
 
-  APP_ABORT(" Failed to create a SlaterDet at PWOrbitalBuilder::putSlaterDet ");
+  myComm->barrier_and_abort(" Failed to create a SlaterDet at PWOrbitalBuilder::putSlaterDet ");
```
PDoakORNL, line 141: See above re errors, myComm etc.

PDoakORNL (approved):
Well this doesn't make things worse.

I think it can be merged in its current state.  But a few small fixes would be nice.

I think error handling in large unform sections i.e. every rank does the same thing, should be handled by having a catch around the uniform section and not every object called needs to have its own handle to the communicator to have a fatal error. I think many builders are used in such sections.  I think everytime we use this myComm->barrier_and_abort we bake in an unecessary and illogical dependence Communicate and MPI.
There are already issues where this is argued about without resolution.

We need a new issue about this return nullptr and PRE.error stuff.  How wide spread is this.

ye-luo:
Start testing in-house

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
ye-luo: This is a sophisticated builder and should carry myComm. So call barrier_and_abort directory. I agree that classes which have myComm must be limited.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
ye-luo: ProgressReportEngine should go away as #3457

src/QMCWaveFunctions/Fermion/SlaterDet.cpp
```diff
@@ -23,35 +23,24 @@ namespace qmcplusplus
...
 
-SlaterDet::SlaterDet(ParticleSet& targetPtcl, const std::string& class_name) : WaveFunctionComponent(class_name)
+SlaterDet::SlaterDet(ParticleSet& targetPtcl, std::vector<std::unique_ptr<Determinant_t>> dets, const std::string& class_name) : WaveFunctionComponent(class_name), Dets(std::move(dets))
 {
-  Optimizable  = false;
+  assert(Dets.size() == targetPtcl.groups());
+
   is_fermionic = true;
 
+  Optimizable = false;
```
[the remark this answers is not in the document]
ye-luo: Indeed, it is possible that we optimize the orbitals of spin-up det but freeze the spin-down one.

Historically, I found WFC::Optimizable not clearly/consistently defined.
Does this flag mean at this class level or this level + all the underlying members?
But this is a separate issue.

src/QMCWaveFunctions/Fermion/SlaterDet.cpp
```diff
@@ -256,13 +245,11 @@ void SlaterDet::copyFromBuffer(ParticleSet& P, WFBufferType& buf)
...
 std::unique_ptr<WaveFunctionComponent> SlaterDet::makeClone(ParticleSet& tqp) const
 {
-  auto myclone         = std::make_unique<SlaterDet>(tqp);
-  myclone->Optimizable = Optimizable;
-  for (int i = 0; i < Dets.size(); ++i)
-  {
-    auto newD = Dets[i]->makeCopy(Dets[i]->getPhi()->makeClone());
-    myclone->add(std::unique_ptr<DiracDeterminantBase>(newD), i);
-  }
+  std::vector<std::unique_ptr<Determinant_t>> dets;
```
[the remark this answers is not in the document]
ye-luo: I'm not getting this. This object is local not a member of a class.

src/QMCWaveFunctions/PlaneWave/PWOrbitalBuilder.cpp
ye-luo: This is a builder so just replace APP_ABORT with `myComm->barrier_and_abort`. Overall we need to reduce myComm pollution.

PR #3356, opened 2021-08-13
Use unique_ptr for QMCHamiltonian, TrialWaveFunction and MCWalkerConfiguration

quantumsteve:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

Describe what this PR changes and why.  If it closes an issue, link to it here
with [a supported keyword](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

This fixes memory leaks allocated by QMCHamiltonian::makeClone and TrialWaveFunction::makeClone. Following this pattern, I did the same for MCWalkerConfiguration.

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Bugfix
- Refactoring (no functional changes, no api changes)

### Does this introduce a breaking change?

- No

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted

PR #3413, opened 2021-09-02
Clear clones before exiting QMCMain

ye-luo:
## Proposed changes
Since #3356,
Crashes shown as https://cdash.qmcpack.org/CDash/testDetails.php?test=15105035&build=239150
are caused by freeing objects owned by CloneManager statically after CUDA runtime got destroyed (not controlled by QMCPACK).
This PR destroy those objects before exiting QMCMain.

- Bugfix
- No

epyc-server

## Checklist


ye-luo:
Start testing in-house

PDoakORNL (approved):
Clearly this is better than a crash.
But clone_manager is hot garbage and this just another layer of leaky plastic.

PR #3415, opened 2021-09-02
[WIP] Working Direct Inversion with ENABLE_CUDA=1 and ENABLE_OFFLOAD=1

PDoakORNL:
## Proposed changes
As it turns out some omp target "code" didn't support the complex state model in twf and friends required by qmchamiltonian and friends.

From #3344 which was the previous attempt at this

Backport of batched direct inversion using CUDA, returns DiracMatrix per walker for single walker API. Removes ability to pass resource in single walker API because openmp target doesn't need it and its awkward in the extreme to use the ResourceLock over refvectorwithleader and resource to put the resource into the leader state.

Still some remenants of refactoring I would prefer to leave if there is actually interest in getting ENABLE_CUDA=1 ENABLE_OFFLOAD=0 working. Although I now realize any acceleration not using openmp target is controversial and we should probably have a full discussion of whether abstractions or changes in code structure that make both that and offload possible are desirable.

Testing has been reduced since it was being frequently broken. test_TrialWaveFunction_diamondC_2x1x1.cpp is broken but I'm not sure the reference values there are any good.  An assert in DMCBatched needed to be commented out, I'm not sure it does anything more that test the tortured state machine of the legacy single walker API.  I would prefer to debug that without OMP target muddying the waters.  Short tests run seem sensible but I need to run more in release, debug is appalling slow, much worse than I remember. This is not causes by the direct inversion code.

An actual full implementation of DiracBatchedDeterminant. Some architecture improvements.  Still much cleanup should be done, but first proper abstraction needs to be added to remove hard dependency on OMP target.

- New feature
- Build related changes
- Documentation changes

- Not sure yet

Leconte

## Checklist


PDoakORNL:
Requested offline by @ye-luo to split this PR again.

ye-luo:
Yes. Requested to PR separately
1. DiracMatrixComputeCUDA and DiracMatrixComputeOMPTarget (and their tests) and releated CUDA change. Since they are not connected in the develop. So I assume we can still PR them separately.
2. Everything remaining except DiracDetermijnantBatched (and its tests). This mostly look like cleanings here and there and mostly non functional changes. You are free to break up changes further.

PR #3430, opened 2021-09-03
WIP:  Fast Force Evaluation for Determinant-Only Wavefunctions

rcclay:

## Proposed changes
This implements the fast force scheme of Filippi et. al [https://aip.scitation.org/doi/full/10.1063/1.4948778](url) for single determinant (but multispecies) wave functions.  Including Jastrows is straightforward, but will require quite a bit more code.  

Current status:  Unit tests correctness-check the algorithm and its subcomponents.  Performance tests have been done on real systems.  This is working code.

This is being PR'd now to start a discussion, get specific actionable feedback, and hopefully get this into QMCPACK.  

- New feature
- Other (please describe):  Included a new wave function wrapper class, currently called "TWFPrototype", around which this algorithm is designed.  

- No

## Checklist


ye-luo:
My quick thoughts on this. Break up this PR into SPOSet level changes, WFC level changes, TWF level changes, and Hamiltontian level changes (Maybe split KE and NLPP). Start with easy ones and use a bottom up approach.

markdewing:
Could you provide an informal description of what makes this method different from the way computations are currently done in the code?

prckent:
@markdewing This uses the Filiippi, Assarf, Moroni scheme https://doi.org/10.1063/1.4948778 which is maybe N^3 cheaper than the current scheme. Similar to iterative updates/table method, it avoid recomputing common matrix elements that do not need to be recomputed & uses an elegant factorization of the terms that are needed.

markdewing:
I'm looking for a bridge between the code and the paper.   @prckent's comment is a very helpful high-level summary.  Is there another level of description between that and needing to understand the full paper?

I think it would be helpful to have more comments connecting the code to the paper.  For example
"This function evaluates matrix A from eq (4) in <reference to paper>"  or "This matrix is the specialization of the B matrix in eq (15) to spatial derivatives for the local energy"

If I understand correctly, the key factorization in the paper is equation 4, which computes a trace over a matrix product.   Where is the trace computed in the code?

One question - How does this evaluation scheme interact with the wavefunction  API (API discussion in #3459)?  Does it need to do anything different?

PDoakORNL:
Is this still going to be run against CI, if not lets close it.  I think issue #3668 is a better way to manage the fast force merge.

PR #3432, opened 2021-09-08
Introduce InputSection

jtkrogel:

## Proposed changes

This PR contains a proposal for separating (validated) XML parsing and object instantiation.  Initial use case is in batched estimators.

- New feature

- No

Laptop

## Checklist


src/Estimators/InputSection.h
```diff
@@ -0,0 +1,116 @@
...
+public:
+  // Query functions
+  bool is_attribute(std::string name){return attributes.find(name)!=attributes.end();};
+  bool is_parameter(std::string name){return parameters.find(name)!=parameters.end();};
+  bool is_required(std::string name){return required.find(name)!=required.end();};
+
+  bool is_string(std::string name){return strings.find(name)!=strings.end();};
+  bool is_bool(std::string name){return bools.find(name)!=bools.end();};
+  bool is_integer(std::string name){return integers.find(name)!=integers.end();};
+  bool is_real(std::string name){return reals.find(name)!=reals.end();};
```
ye-luo, line 68: Should all these is_* functions protected or even private?

src/Estimators/MomentumDistributionInput.h
```diff
@@ -0,0 +1,42 @@
...
+namespace qmcplusplus
+{
+
+
+/** Native representation for Momentum Distribution Estimators inputs
+ */
+class MomentumDistributionInput : public InputSection
+{
+public:
+  MomentumDistributionInput()
```
ye-luo, line 25: Please write a test based on MomentumDistributionInput. Which would help understanding how to use the input class.

ye-luo (changes requested):

src/Estimators/InputSection.h
jtkrogel: They are not strictly needed to be public, but also no harm (read only information).  I can redesignate private, if you prefer.

ye-luo: Usually minimal exposure of public APIs is preferred. This reduces entanglements.

ye-luo: If rediscover is needed from outside the class, we may consider a negative point on the design.

jtkrogel: This information is not strictly needed beyond the class.  I will move these functions to private.

src/Estimators/InputSection.h
```diff
@@ -0,0 +1,116 @@
...
+
+  bool has_default(std::string name){return default_values.find(name)!=default_values.end();};
+  bool has(std::string name){return values.find(name)!=values.end();};
```
ye-luo, line 71: please mark all the above is_* has_* functions const.

src/Estimators/InputSection.h
```diff
@@ -0,0 +1,116 @@
...
+
+
+  // Enable read-only access to variable values.
+  //   Needs updating to allow copy-less return.
+  template<typename T>
+  T get(std::string name){return std::any_cast<T>(values[name]);};
```
ye-luo, line 77: This one can be const function as well.

src/Estimators/InputSection.h
```diff
@@ -0,0 +1,116 @@
...
+  //};
+
+
+  // Read variable values (initialize) from XML input.
+  //   Later, this should call a correctness checking function and enforce immutability.
+  //   (required values are all provided, provided values fall in allowed ranges)
+  void readXML(xmlNodePtr cur);
+
+  //  Simple write of contents.  Can be replaced/removed in any final implemenation.
+  void report();
```
ye-luo, line 102: I expect this always be const function.

src/Estimators/InputSection.h
```diff
@@ -0,0 +1,116 @@
...
+
+private:
+  // Perform typed read and assignment of input variables
+  void set_from_stream(std::string name, std::istringstream& svalue);
```
ye-luo, line 106: Usualy for name.
`const std::string& name` can be better than `std::string name`

prckent:
Standalone testing required imo ("unit test"). If the design is right this should  be straightforward.

jtkrogel:
Agreed.  I will work on unit tests after I have addressed Ye's requests above.

src/Estimators/InputSection.h
jtkrogel: Done

src/Estimators/InputSection.h
jtkrogel: Done

src/Estimators/InputSection.h
jtkrogel: Done

src/Estimators/InputSection.h
jtkrogel: Done

src/Estimators/InputSection.h
quantumsteve: [std::string_view](https://en.cppreference.com/w/cpp/string/basic_string_view)?

ye-luo: string_view is a C++17 feature. I think it makes sense in some use cases. In QMCPACK string operation is never the major computational cost, so leave contributors to decide whether to opt-in this feature.

src/Estimators/MomentumDistributionInput.h
jtkrogel: Tests added.

jtkrogel:
I've added tests for key functionality.  The tests also demo InputSection class usage.

jtkrogel:
Ready to move forward now.

src/Estimators/tests/test_InputSection.cpp
```diff
@@ -0,0 +1,345 @@
...
+  
+  // invalid type access results in thrown exception
+  try
+  {
+    auto v = ti.get<int>("name");
+  }
+  catch(std::bad_cast& exception)
+  {
+  }
+  try
```
ye-luo, line 292: Add a line before try for reading easily.

src/Estimators/tests/test_InputSection.cpp
```diff
@@ -0,0 +1,345 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2020 QMCPACK developers.
+//
+// File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Lab
+//
+// File created by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Lab
```
ye-luo, line 9: update license header, the year and owner.

ye-luo (commented):
Need formatting but let me know if you don't have clang-format around.

src/Estimators/tests/test_InputSection.cpp
jtkrogel: Done.  I don't have clang-format setup.

src/Estimators/tests/test_InputSection.cpp
jtkrogel: Done.

ye-luo:
Test this please

ye-luo (dismissed):

ye-luo:
@jtkrogel mixed precision build fails in 
```
	 13 - deterministic-unit_test_estimators (Failed)
```

src/Estimators/InputSection.cpp
```diff
@@ -0,0 +1,176 @@
...
+  }
+  else if (is_real(name))
+  {
+    Real value;
+    svalue >> value;
+    values[name] = value;
+  }
+  else
+  {
+    std::stringstream error;
```
PDoakORNL, line 148: catch2 has a mechanism to check that an exception is thrown this should have test coverage since this is real functionality of the class.
```c++
CHECK_THROWS_AS(..., UniformCommunicateError);
```
see src/QMCDrivers/tests/QMCDriverNewTestWrapper.h

src/Estimators/InputSection.cpp
```diff
@@ -0,0 +1,176 @@
...
+  set_defaults();
+
+  // check input validity
+  check_valid();
+
+  //report();
+};
+
+
+void InputSection::report() const
```
PDoakORNL, line 99: Needs testing someone may mess with the is_xxx in the future this has a bunch that can break.

If you imagine yourself or anyone parsing this output you should test and therefor specify which parts are "guaranteed" to be conserved.

src/Estimators/tests/test_InputSection.cpp
```diff
@@ -0,0 +1,335 @@
...
+    CHECK(ti.get<bool>("full") == false);
+    CHECK(ti.get<int>("count") == 15);
+    CHECK(ti.get<std::string>("name") == "demo");
+    CHECK(ti.get<int>("samples") == 20);
+    CHECK(ti.get<RealType>("width") == Approx(1.0));
+    CHECK(ti.get<bool>("rational") == false);
+  }
+
+
+  { // test init: complete attributes and parameters
```
PDoakORNL, line 226: catch2 has test sections
```c++
SECTION("complete attributes and parameters") {
...
}
```

src/Estimators/tests/test_InputSection.cpp
```diff
@@ -0,0 +1,335 @@
...
+      {"kmax", 3.0},
+      {"full", false},
+      {"label", std::string("relative")},
+      {"count", 15},
+      {"width", 2.5},
+      {"rational", true},
+  });
+
+  // invalid type access results in thrown exception
+  try
```
PDoakORNL, line 277: use CHECK_THROWS or CHECK_THROWS_AS

PDoakORNL: As is I'm not sure what this actually tests since I'm pretty sure it can't fail.

PDoakORNL (changes requested):
We need better documentation of tests and how to write them. Although most of this is in the catch2 docs they are not very good.

src/Estimators/InputSection.cpp
jtkrogel: This is intended primarily as a developer/debugging function and it called nowhere in the main code.  I will mark it as such with additional comments and move it to private so as not to confuse it with the user API.

src/Estimators/tests/test_InputSection.cpp
jtkrogel: No, it can't based on the current underlying implementation.  It should be tested in case that implementation changes for some reason down the road.

markdewing:
The design uses inheritance from a base class for functionality.  This pattern raises some concerns as we're trying to move away from some uses of inheritance in the code.    This design may be okay, but I'd like to make sure we think carefully about it.

My intuition is that new designs use trait classes and templates to separate out specifications from implementation.  The class would be used something like:
```
InputSection<TestInput> t1;
```
and `TestInput` would define `attributes`,`parameters`, etc.
One downside is that the specific input classes would have to declare the types for `attributes`, etc.  Annoying to put that in every class.
A fix for this would be a base class that has just these variables (and nothing else).   Inheritance still exists but the scope is smaller.
```
struct InputSectionBase {
    std::unordered_set<std::string> attributes;
     ...
};

struct TestInput : public InputSectionBase {
   TestInput() {
         attributes = {"type", "name"};
   )
};

InputSection<TestInput> t1;
```

This achieves some separation, but adds extra classes and plumbing.

src/Estimators/InputSection.cpp
jtkrogel: Updated.

src/Estimators/tests/test_InputSection.cpp
jtkrogel: Updated.

jtkrogel:
@markdewing This implementation is meant as a proposal for more discussion down the road, not as something we are committing to spreading throughout the code right now.  InputSection will not be used for more than couple of estimators unless it is later agreed on as a route forward (and then subsequently upgraded).

The idea was to take a bit of an "eat your own cooking" approach to see if the idea can work or not.  I suggest we move the conceptual discussion to an issue, since we have not worked out a final agreement on what the eventual input classes will look like.

jtkrogel:
@PDoakORNL I've addressed your review comments.

jtkrogel:
@ye-luo mixed precision build also works now.

ye-luo (approved):

PDoakORNL (approved):

ye-luo:
Test this please

PR #3436, opened 2021-09-10
Update test_ResourceCollection

### LINKED ISSUE #3424: Unit tests for ResourceCollection don't cover ResourceCollectionTeamLock
**Describe the bug**
actual usage pattern is also not reflected in any test case. i.e. all access is supposed to be through ResourceCollectionTeamLock.
I should have stopped this in #3324 

**To Reproduce**
Read code

**Expected behavior**

**System:**
 - system name [e.g. fusiont5, summit]
 - modules loaded [e.g. output of `module list`]
 - other systems where this is reproducible [e.g. "my laptop", "none"]

**Additional context**

ye-luo:
## Proposed changes
Remove unused implementation and update tests. Closes #3424

- No

epyc-server

## Checklist

### DIFF
--- a/src/Utilities/CMakeLists.txt
+++ b/src/Utilities/CMakeLists.txt
@@ -13,3 +13,7 @@
 #\todo move qmcutils here
 
 subdirs(for_testing)
+
+if(BUILD_UNIT_TESTS)
+  add_subdirectory(tests)
+endif()
--- a/src/Utilities/tests/test_ResourceCollection.cpp
+++ b/src/Utilities/tests/test_ResourceCollection.cpp
@@ -52,7 +52,7 @@ public:
     collection.addResource(std::move(external_memory_handle));
   }
 
-  void acquireResource(ResourceCollection& collection)
+  void acquireResource(ResourceCollection& collection, const RefVectorWithLeader<WFCResourceConsumer>& wfcrc_list)
   {
     auto res_ptr = dynamic_cast<MemoryResource*>(collection.lendResource().release());
     if (!res_ptr)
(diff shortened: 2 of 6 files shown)


src/Utilities/CMakeLists.txt
```diff
@@ -13,3 +13,7 @@
 #\todo move qmcutils here
 
 subdirs(for_testing)
+
+if(BUILD_UNIT_TESTS)
```
PDoakORNL, line 17: This always makes me imagine a developer testing their new code using the application. If you are going through enough recompiles that this is a good option, write a test.

Is there another use case?

ye-luo: BUILD_UNIT_TESTS was introduced to allow disabling building unit test to speed-up compilation. I never turned it off.
Since it was there, I just keep it around as I move bits in CMake.

ye-luo:
Start testing in-house

src/Utilities/CMakeLists.txt
PDoakORNL: Ok we might want to revisit it in a all hands meeting someday.

src/Utilities/tests/test_ResourceCollection.cpp
```diff
@@ -52,15 +52,15 @@ class WFCResourceConsumer
     collection.addResource(std::move(external_memory_handle));
   }
 
-  void acquireResource(ResourceCollection& collection)
```
PDoakORNL, line 55: So this now tests the "public" API as we expect unit tests to do.

PDoakORNL (approved):
LGTM unit tests updated to reflect usage.

Issue #2554, opened 2020-06-22
Change Kinetic Energy to depdend on TWF


ye-luo:
**Is your feature request related to a problem? Please describe.**
Current KE are computed from P.G and P.L. This was for convenience but unreasonable.

**Describe the solution you'd like**
 A more clean way is to compute KE from TWF.G and TWF.L.

PR #3439, opened 2021-09-13
Hamiltonian batched APIs depend on TWF

ye-luo:
## Proposed changes
Hamiltonian depends on ParticleSet and TrialWaveFunction. The former is passed in via the arguments but the latter is handled based on the actual Hamiltonian component. It is better to avoid holding any references inside an object and always pass in via argument.
So all the batched APIs have TrialWaveFunction added in this PR.
This is part of #2554

- New feature

- No

epyc-server

## Checklist


PDoakORNL (approved):
This looks fine to me, but we may end up adding an overload that doesn't include the wavefunctions back in. Not sure if this did make sense somewhere or the API should have been this from the beginning.

ye-luo:
I think for operators, we prefer it not being a state-machine. Thus internally keeping references to "dynamic" pieces seems to be an hassle and it is better to hand in those pieces explicitly via arguments. So I would say having TWF as an argument should have been from the beginning.

ye-luo:
Start testing in-house

PR #3442, opened 2021-09-14
Move SpeciesSet from Utilities to Particle

ye-luo:
## Proposed changes
triggered by the move, add more cleaning to SpinDensityNew.

- No

epyc-sever

## Checklist

### DIFF
--- a/src/Estimators/tests/test_SpinDensityNew.cpp
+++ b/src/Estimators/tests/test_SpinDensityNew.cpp
@@ -16,7 +16,7 @@
 #include "ValidSpinDensityInput.h"
 #include "SpinDensityNew.h"
 #include "RandomForTest.h"
-#include "Utilities/SpeciesSet.h"
+//#include "Utilities/SpeciesSet.h"
 #include "ParticleSet.h"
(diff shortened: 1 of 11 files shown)


src/Estimators/tests/test_SpinDensityNew.cpp
```diff
@@ -16,7 +16,7 @@
 #include "ValidSpinDensityInput.h"
 #include "SpinDensityNew.h"
 #include "RandomForTest.h"
-#include "Utilities/SpeciesSet.h"
+//#include "Utilities/SpeciesSet.h"
```
PDoakORNL, line 19: should be able to just delete now.

PDoakORNL (approved):
I think this makes more sense. And thanks for all the formatting.

src/Estimators/tests/test_SpinDensityNew.cpp
ye-luo: Will handle it when I deal with merge conflicts.

ye-luo:
Start testing in-house

PR #3445, opened 2021-09-14
Move QMCLinearOptimizeBatched to QMCFixedSampleLinearOptimizeBatched

markdewing:
Move remaining functionality from QMCLinearOptimizeBatched to QMCFixedSampleLinearOptimizeBatched and remove QMCLinearOptimizeBatched.
 
Also removed commented-out code from generalized eigenvalue solvers

Future: The "FixedSample" part of the name is no longer necessary. A future PR will rename the class to QMCLinearOptimizeBatched. It was easier step-wise to move everything to the more specific QMCFixedSampleLinearOptimizeBatched class first.

- No

desktop

## Checklist


ye-luo:
Changes LGTM.

About the future naming, removing FixedSample makes batched class doesn't look like the one wihtout batch. I mean QMCFixedSampleLinearOptimize correspond to QMCLinearOptimizeBatched It is a bit confusing.
How easy it is to apply the same class fusion to QMCFixedSampleLinearOptimize?

markdewing:
I would like to avoid doing too much with the non-batched classes.  I assume they will go away eventually.   Adding some comments about the class correspondence should be sufficient to reduce confusion.   I can also put off renaming QMCFIxedSampleLinearOptimizeBatched since it's not critical.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
```diff
@@ -1493,4 +1600,154 @@ bool QMCFixedSampleLinearOptimizeBatched::hybrid_run()
 }
 #endif
 
+QMCFixedSampleLinearOptimizeBatched::RealType QMCFixedSampleLinearOptimizeBatched::getLowestEigenvector(
```
PDoakORNL, line 1603: Unit test?  Doesn't belong in this PR but it should be in one.

PDoakORNL (approved):
Other than maybe revealing a missing unit test this refactor looks good to me and removes unecessary inheritance.

ye-luo:
> I would like to avoid doing too much with the non-batched classes. I assume they will go away eventually. Adding some comments about the class correspondence should be sufficient to reduce confusion. I can also put off renaming QMCFIxedSampleLinearOptimizeBatched since it's not critical.

I vote for just not doing the renaming. The only downside is long class name but it is just minor issue.

ye-luo:
Start testing in-house

PR #3446, opened 2021-09-15
Add restrict keyword in sincos wrappers.

ye-luo:
## Proposed changes
Some oversight.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/Platforms/CPU/math.hpp
+++ b/src/Platforms/CPU/math.hpp
@@ -29,36 +29,36 @@ namespace qmcplusplus
 /// sincos function wrapper
 #if defined(__APPLE__)
 
-inline void sincos(double a, double* s, double* c)
+inline void sincos(double a, double* restrict s, double* restrict c)
 {
   ::__sincos(a,s,c);
 }
 
-inline void sincos(float a, float* s, float* c)
+inline void sincos(float a, float* restrict s, float* restrict c)
 {
   ::__sincosf(a,s,c);
 }
 
 #elif defined(HAVE_AMD_LIBM)
 
-inline void sincos(double a, double* s, double* c)
+inline void sincos(double a, double* restrict s, double* restrict c)
 {
   ::amd_sincos(a,s,c);
 }
 
-inline void sincos(float a, float* s, float* c)
+inline void sincos(float a, float* restrict s, float* restrict c)
 {
   ::amd_sincosf(a,s,c);
 }
 
 #elif defined(HAVE_SINCOS)
 
-inline void sincos(double a, double* s, double* c)
+inline void sincos(double a, double* restrict s, double* restrict c)
 {
   ::sincos(a,s,c);
 }
 
-inline void sincos(float a, float* s, float* c)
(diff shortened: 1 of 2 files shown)


ye-luo:
Start testing in-house

ye-luo:
Start testing in-house

PDoakORNL:
I feel like `restrict` in c++ is often a cargo cult sort of change.  I.e. made in hope that sprinkling it everywhere will result in optimization.  Does this actually cause vectorization passes to work on some compilers?  Do we know it doesn't break any?

src/Platforms/CPU/math.hpp
```diff
@@ -29,36 +29,36 @@ namespace qmcplusplus
...
-inline void sincos(double a, double* s, double* c)
+inline void sincos(double a, double* restrict s, double* restrict c)
 {
   ::sincos(a,s,c);
 }
 
-inline void sincos(float a, float* s, float* c)
+inline void sincos(float a, float* restrict s, float* restrict c)
 {
 #if defined(HAVE_MASS)
```
PDoakORNL, line 63: another thing that is strange is there is a vsincos in massv.h

ye-luo: What did you mean?

PDoakORNL: API in mass contains no restrict keyword.

PDoakORNL: I was confused by this comment and the rest of the code being cutoff

ye-luo:
> I feel like `restrict` in c++ is often a cargo cult sort of change. I.e. made in hope that sprinkling it everywhere will result in optimization. Does this actually cause vectorization passes to work on some compilers? Do we know it doesn't break any?

This is a wrapper over C library. sincos always assume no pointer aliasing. So adding restrict is safe.
```
Does this actually cause vectorization passes to work on some compilers?
```
Not exploring this in depth.
```
Do we know it doesn't break any?
```
No breakage observed so far.

src/Platforms/CPU/math.hpp
ye-luo: restrict has no impact on the function declaration. It only affects function definition when doing code compilation.

PDoakORNL:
So my only real comment is mass doesn't include restrict in its API. Can't say I really understand it in the context where one is calling an external function with the same pointers but no restrict keywords.

ye-luo:
Even if sincos declaration in math libraries doesn't have restrict keyword, there is nothing wrong for us to suggest restrict. Depending on the compiler, it may inline math functions with fast-math option and thus, Having restrict convey the message of no pointer aliasing.

src/Platforms/CPU/math.hpp
PDoakORNL: Alright I think this will cause us no harm. Interested to hear if it helps anywhere.

PDoakORNL (approved):
LGTM

PR #3453, opened 2021-09-16
Fix NLPP race in batched implementation.

ye-luo:
## Proposed changes
There is a data race in NLPP where the joblist is prepared. The race is caused by sharing the same RNG withing a crowd.
The issue was not exposed because this level of threading is only active when crowd level threading is turned off which is a scenario when I debug codes.

- Bugfix

- No

epyc-server, bora

## Checklist

### DIFF
--- a/src/QMCHamiltonians/NonLocalECPotential.cpp
+++ b/src/QMCHamiltonians/NonLocalECPotential.cpp
@@ -248,17 +248,16 @@ void NonLocalECPotential::mw_evaluateImpl(const RefVectorWithLeader<OperatorBase
                                           bool Tmove)
 {
   auto& O_leader           = o_list.getCastedLeader<NonLocalECPotential>();
-  ParticleSet& pset_leader = p_list[0];
-  const size_t ngroups     = pset_leader.groups();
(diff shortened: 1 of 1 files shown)


markdewing:
Could you explain more about how this fixes the race.
Is the removal of the OpenMP parallel region?  (No data races if no parallel code??)
Is it moving/changing the computation of max_num_jobs?

ye-luo:
In this case no "omp parallel" no race.

Actually both.
RNG is shared. Calling it under "omp parallel" creates a race.
max_num_jobs was also part of the race but I prefer to move it to a more local scope anyway.

ye-luo:
Start testing in-house

PDoakORNL (approved):
LGTM
Better caught in debugging than production.
We need a better way to handle nested concurrency contexts,  even if in most production cases on HPC systems weak in  CPU and strong in GPU it is not used much.

prckent (approved):
LGTM

Do you have any suggestions for how we could have caught this?

ye-luo:
Indeed, we don't have NLPP batched API unit tests #3454

PR #3455, opened 2021-09-16
Use full precision in MSD reductions to improve accuracy of mixed precision computation

ye-luo:
## Proposed changes
MSD are sensitive to numerics and cased MSD deterministic tests to fail in mixed precision builds.
So change reduction related places to use full precision. It seems help both NVHPC and OneAPI compilers.

- Bugfix

- No

epyc-server

## Checklist


prckent (changes requested):
Please put a comment around the new/changed variable definitions.
"Use full precision values due to numerical sensitivity" or similar.

Operations on value and laplacian should also use full precision consistently.

ye-luo:
@prckent comments added.

ye-luo:
Test this please

prckent (approved):
OK.

Lets see what this does to the stability.

Issue #3411, opened 2021-09-01
General refactoring


williamfgc:
This is just a general issue to track PRs related to General Refactoring

**Is your feature request related to a problem? Please describe.**
This is part of the overall request to clean the current C++ codebase and adopt modern practices and standards

**Describe the solution you'd like**
- Adopt modern C++ features practices and uses classes, templates, naming conventions, testing and docs appropriately
- Use exceptions to clean memory appropriately (even before having to call `MPI_Abort`)
- Bump code coverage by deterministic tests
- Work towards compact code that is more legible, cache friendly, and accessible to the general community (code is "data")

**Describe alternatives you've considered**
Depends on community feedback

jtkrogel:
Outdated

PR #3458, opened 2021-09-17
Refactoring OperatorBase public members naming

williamfgc:
## Proposed changes

Refactoring public members in `OperatorBase` class to accommodate to QMCPACK developer guide
Replace `typedef` with `using`
Rename current public member variables across the hierarchy
Remove unused `IonForce` and `depName`
Component of original PR #3412 
Related to #3411 

- No

Ubuntu 20.04, must pass CI

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/HamiltonianRef.cpp
+++ b/src/QMCDrivers/WFOpt/HamiltonianRef.cpp
@@ -45,7 +45,7 @@ FullPrecRealType HamiltonianRef::evaluate(ParticleSet& P)
   {
     const auto LocalEnergyComponent = Hrefs_[i].get().evaluate(P);
     if (std::isnan(LocalEnergyComponent))
-      APP_ABORT("QMCHamiltonian::evaluate component " + Hrefs_[i].get().myName + " returns NaN\n");
+      APP_ABORT("QMCHamiltonian::evaluate component " + Hrefs_[i].get().getMyName() + " returns NaN\n");
     LocalEnergy += LocalEnergyComponent;
   }
   return LocalEnergy;
--- a/src/QMCDrivers/WFOpt/HamiltonianRef.cpp
+++ b/src/QMCDrivers/WFOpt/HamiltonianRef.cpp
@@ -45,7 +45,7 @@ FullPrecRealType HamiltonianRef::evaluate(ParticleSet& P)
   {
     const auto LocalEnergyComponent = Hrefs_[i].get().evaluate(P);
     if (std::isnan(LocalEnergyComponent))
-      APP_ABORT("QMCHamiltonian::evaluate component " + Hrefs_[i].get().getMyName() + " returns NaN\n");
+      APP_ABORT("QMCHamiltonian::evaluate component " + Hrefs_[i].get().getName() + " returns NaN\n");
     LocalEnergy += LocalEnergyComponent;
   }
   return LocalEnergy;
--- a/src/QMCHamiltonians/OperatorBase.h
+++ b/src/QMCHamiltonians/OperatorBase.h
@@ -110,37 +110,44 @@ public:
     NONLOCAL    = 5,
   };
 
-  ///set the current update mode
-  std::bitset<8> UpdateMode;
-
-  ///number of dependents: to be removed
-  int Dependants;
-  ///current value
-  Return_t Value;
-
-  /// This is used to store the value for force on the source
-  /// ParticleSet.  It is accumulated if setComputeForces(true).
-  ParticleSet::ParticlePos_t IonForce;
-
-  //Walker<Return_t, ParticleSet::ParticleGradient_t>* tWalker;
-  ///name of this object
-  std::string myName;
-  ///name of dependent object: to be removed
-  std::string depName;
-
-#if !defined(REMOVE_TRACEMANAGER)
-  ///whether traces are being collected
-  TraceRequest request;
-
-  std::vector<RealType> ValueVector;
-
-#endif
-
   ///constructor
   OperatorBase();
 
   ///virtual destructor
   virtual ~OperatorBase() {}
+
+  // getter for update_mode member
+  /**
+   * @brief get update_mode_ 
+   * @return std::bitset<8>& reference of get_update_mode_
+   */
+  std::bitset<8>& getUpdateMode() noexcept;
+
+  /**
+   * @brief get a copy of value_
+   * @return Return_t copy of value_
+   */
+  Return_t getValue() const noexcept;
+
+  /**
+   * @brief getter a copy of my_name_, rvalue small string optimization
+   * @return std::string copy of my_name_ member
+   */
+  std::string getMyName() const noexcept;
+
+  /**
+   * @brief Set my_name member, uses small string optimization (pass by value)
+   * @param name input
+   */
+  void setMyName(const std::string name) noexcept;
+
+#if !defined(REMOVE_TRACEMANAGER)
+  /**
+   * @brief Get request_ member
+   * @return TraceRequest& reference to request_
+   */
+  TraceRequest& getRequest() noexcept;
+#endif
 
   inline bool is_classical() { return quantum_domain == CLASSICAL; }
   inline bool is_quantum() { return quantum_domain == QUANTUM; }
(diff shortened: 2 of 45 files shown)


src/QMCHamiltonians/OperatorBase.h
```diff
@@ -108,28 +111,21 @@ class OperatorBase : public QMCTraits
   };
 
   ///set the current update mode
-  std::bitset<8> UpdateMode;
+  std::bitset<8> update_mode;
```
ye-luo, line 114: since you are touching variable names. Please add underscore to all member variables which has been touched.

williamfgc: @ye-luo quick question, in the docs it says only protected and private, see [here](https://qmcpack.readthedocs.io/en/develop/developing.html#class-data-members). Should this also apply to public members? I'm only touching those 'cause it's already touching many files.

ye-luo: I went through the manual and found
```
Class private/protected data members names should follow the convention of variable names with a trailing underscore (_).
```
@PDoakORNL do you remember why there is no mention of public members? Is it because we do not expect public members? Since there are public members from existing code, should we just request public members to carry underscore as well?

prckent: Probably the comment about _ on public members was missed since there really shouldn't be any?

ye-luo: It seems that renaming anyway changes the consumer side, I think we should put them as protected and then add const accessor for the consumer.

williamfgc: How do we want to handle POD struct public members?

ye-luo: > How do we want to handle POD struct public members?

I think POD struct doesn't need underscore.
But if a struct has member functions, we should use class instead of struct and also minimize public members.

prckent: Echoing Ye's comment: nearly all the structs should be classes. And the public access to their data should be removed and go via accessors. There are a lot of these so it is something for us all to chip away at as we come across them.

williamfgc: Sound good. Would the exception "global" read-only such as `enum class` and `const/constexpr static`? See those proposed in PR #3412 for [enum class](https://github.com/williamfgc/qmcpack/blob/refactoring-OperatorBase/src/QMCHamiltonians/OperatorBase.h#L90) and [static constexpr](https://github.com/williamfgc/qmcpack/blob/refactoring-OperatorBase/src/QMCHamiltonians/OperatorBase.h#L90).

PDoakORNL: If the spirit of an object is still that of a data structure then public access to data members is fine. But member functions should not do "computations" and it should be strongly considered whether a pure function might be more appropriate.

PDoakORNL: To be clear, I don't think there should be any public data members in a class excepting consts which I don't think should have an underscore. (I have no good reason for that). I don't count type aliases in this, they are just named as types whether they are public or private in a class. 

Currently I put _ onto existing data members if I keep them as I move them to protected or private.

ye-luo: @williamfgc So in summary,
when a class data member shows up as public, it is already a red flag. We'd like to have no pubic class member except compile time constants or types.
Then there will be no question about whether to add underscore or not on public data member.

POD struct has its own use. it is meant to be all public. So we don't need to add underscore.

The purpose of adding underscore to easily identify whether a variable is internal or local in a member function.
```
foo::boo()
var_name // must be local or argument
var_name_ // internal
```

ye-luo: Could add a line in the doc explaining why public is not listed?
"Rethink why using a public data member at the first place." something similar.

So in this PR, I think it will be good to use protected.

ye-luo: We should also think about how to take advantage of clang-tidy.

williamfgc: Sounds like a plan, thanks. Also, do we want to standardize accessor functions with a `getcX` and `getX` pattern? Something similar to what it's outlined in the [docs](https://qmcpack.readthedocs.io/en/develop/developing.html#class-format) and also how iterators distinguish between `const` and `non-const`.

```
const X& getcX() const; // read-only reference
X& getX(); // expect to modify member
```
Using templates is another option `get<T>()`, but would have the same effect (`getX ~ get<X>` if X is registered).
It would make things "findable" since when are trying to make code [FAIR](https://www.nature.com/articles/sdata201618) data.  Getting philosophical, but it's good to address this.

ye-luo: If a member data needs to be accessed outside the class, then we provide const reference accessor. Only if a non-const reference accessor is really really necessary, we provide it and its name should be the same as the const one. For example "std::vector::data()" has both const and non-const version.

I don't get the second part of your comment about template. X is a data member instead of data type.

PDoakORNL:
Current standard is `get_variable_name()` if its a trivial accessor. This is a purposeful exception to the function naming rules.
In cases where other classes need to break encapsulation and get a r/w refrences to a data member I think we've settled on 
`get_ref_variable_name()`.

If there really isn't a better naming choice for an accessor with computation or other costs then using `getVariableName()` is allowable.

ye-luo:
I don't see the value of distinguishing get_ref_variable_name and get_variable_name for accessors, The compiler should handle choosing const and non-const version.

src/QMCHamiltonians/OperatorBase.h
williamfgc: @ye-luo OK, we don't foresee much use of non-const access, that's good. Agree with `std::vector::data()` example, iterators certainly took a separate semantics route.

> I don't get the second part of your comment about template. X is a data member instead of data type.

with C++ `auto template` (since C++17) we can pass enum entries as types (the member would be `X_`, while `X` is a enum that lists publicly accessed variables), I'm doing something like that [here](https://github.com/ORNL/ncio/blob/main/bindings/CXX17/ncio/cxx17/cxx17DataDescriptor.tcc#L113), the type is captured with `decltype` for completeness. Probably, the amount of boilerplate doesn't justify for this case as getX and get<X> are equally findable in any IDE or editor (getX is simpler) which is one of the goals.

williamfgc: @ye-luo nevermind, C++17 `T& get <auto, T>` T recognition doesn't play well with return functions (the example shown is with `void`), not to mention consumer `lvalues` can be `auto` making even more confusing to the compiler. Nothing to win, but I learned something new today. Thanks.

src/QMCDrivers/WFOpt/HamiltonianRef.cpp
```diff
@@ -45,7 +45,7 @@ FullPrecRealType HamiltonianRef::evaluate(ParticleSet& P)
   {
     const auto LocalEnergyComponent = Hrefs_[i].get().evaluate(P);
     if (std::isnan(LocalEnergyComponent))
-      APP_ABORT("QMCHamiltonian::evaluate component " + Hrefs_[i].get().myName + " returns NaN\n");
+      APP_ABORT("QMCHamiltonian::evaluate component " + Hrefs_[i].get().getMyName() + " returns NaN\n");
```
ye-luo, line 48: Could you update `getMyName` to  `getName` to align with other name inquiry functions?

ye-luo:
Test this please

williamfgc:
Start testing in-house

src/QMCHamiltonians/QMCHamiltonian.cpp
```diff
@@ -1095,7 +1095,7 @@ void QMCHamiltonian::evaluate(MCWalkerConfiguration& W,
     ScopedTimer h_timer(my_timers_[i]);
     H[i]->addEnergy(W, LocalEnergyVector, Txy);
   }
-  KineticEnergyVector = H[0]->ValueVector;
```
williamfgc, line 1098: @PDoakORNL can you confirm if removing `OperatorBase::ValueVector` is reasonable? GPU CI is passing now and it's commented out in line 1057.

ye-luo: Do it as separate commit please

ye-luo:
Test this please

ye-luo (approved):
LGTM

PR #2913, opened 2021-02-13
Distance table AA type offload

ye-luo:
## Proposed changes
Initial version. Correctness guaranteed. More performance tuning and adjust API calls will be needed.

## What type(s) of changes does this code introduce?
- New feature
- Refactoring (no functional changes, no api changes)

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
epyc-server

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'

PR #3462, opened 2021-09-21
Fix distance table AA out of bound

ye-luo:
## Proposed changes
#2913 introduced a bug which copies more elements than old_r_ has due to padding.
The issue is similar to #1334 but I'm consider a different solution by removing the padding requirement.
Alignment is always satisfied. Adding padding only removes the remainder loop.
I consider the claimed performance benefit is negligible.

I tried asan but it doesn't report invalid access. So I think somewhere in the memory management adds padding and prevents crash.

- Bugfix

- No

Epyc-server

## Checklist

### DIFF
--- a/src/Particle/SoaDistanceTableAA.h
+++ b/src/Particle/SoaDistanceTableAA.h
@@ -80,9 +80,7 @@ struct SoaDistanceTableAA : public DTD_BConds<T, D, SC>, public DistanceTableDat
 
     old_r_.resize(N_targets);
     old_dr_.resize(N_targets);
-    // The padding of temp_r_ and temp_dr_ is necessary for the memory copy in the update function
-    // temp_r_ is padded explicitly while temp_dr_ is padded internally
-    temp_r_.resize(Ntargets_padded);
+    temp_r_.resize(N_targets);
     temp_dr_.resize(N_targets);
   }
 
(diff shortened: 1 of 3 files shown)


ye-luo:
Start testing in-house

PDoakORNL (approved):
Looks better to me.
This sort of padding issue has potential to pop up elsewhere as well. If I understand properly in this case you're just removing this optimization which is not proven to actually have any significant performance boost?

I guess the original assumption is the copy_n is so dumb that you'll get all non simd copy if its not a multiple of the alignment?

ye-luo:
The original intention was to remove the remainder loop. When N is large, the remainder is anyway negligible.

prckent:
Good discussion above.

I have always been worried about the padding implementation. It makes it very easy for bugs like this to exist due to a missed or incorrect resize. Would setting the alignment to 1 have caught this? It would be a good to if we could configure one of our debug builds in a way that asan would have caught this problem.

Clearly the whole implementation could be rethought, even its very existence (why do we resize anything when we know the max sizes?... and, Peter's question, is there much/any benefit at all with modern hardware/compilers?), but for now we should be concerned about correctness.

ye-luo:
I adjusted one of the unit test and in debug build the issue can be easily caught. We do have debug build in CI.

Alignment is critical for SIMD which we don't lose here. Some internal padding is still necessary. However we should not compute/copy data on the padding.

prckent:
Why didn't we catch this before though? I am thinking far more generally than just the distance tables. Of course, we need alignment for SIMD, GPUs etc. in general, I am not questioning that. But it could perhaps be done differently. e.g. A cleaner mechanism to do the resizing or simply avoiding resizing.

ye-luo:
I don't think resize is relevant to this issue. Here resize is only done once when the object is constructed. The issue comes up where the consumer side assumes padding has been applied and operates on padded size while the container didn't use alignment aware allocator. That was an oversight.

We didn't caught that issue because our tests relies to much on C which gives size of 4, 8 or 16, too good to detect alignment bugs. In the original unit test, the system has 4 up and 4 down electrons and thus I changed to 2 up 1 down, in total 3.

Apart of writing more non-regular tests, we should always keep in mind that code needs to be written as hard-ware agnostic as possible. Unless absolutely needed, avoid leaking such as possible.

PR #3464, opened 2021-09-21
update distance table internal API

ye-luo:
## Proposed changes
All distance table internal change.
Update the boundary condition computeDistancesOffload API.
There will be use cases where the stride size of position array is different from the stride size of r/dr.
So it is better to keep them separate.

Also added a minor fix, I don't remember why I asked Cody to add assert(size>1) to AB table in  #3375. it should not.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Start testing in-house

PR #3467, opened 2021-09-21
Move unit tests

### LINKED ISSUE #3428: restructure unit test installation needed
Two issues
1. unit tests are not added following cmake target hierarchy.
In src/CMakeLists.txt 
```
  if(BUILD_UNIT_TESTS) #{
    #Unit test directories
    add_subdirectory(Numerics/tests)
    add_subdirectory(Utilities/tests)
    add_subdirectory(einspline/tests)
    add_subdirectory(spline2/tests)
    add_subdirectory(Concurrency/tests)
    add_subdirectory(QMCWaveFunctions/tests)
    add_subdirectory(QMCHamiltonians/tests)
    add_subdirectory(type_traits/tests)
    add_subdirectory(QMCDrivers/tests)
    add_subdirectory(QMCApp/tests)
    add_subdirectory(Message/tests)
    add_subdirectory(Optimize/tests)
    add_subdirectory(QMCTools/tests)
  endif() #}
```
This caused an issue that if I type make in `<build>/src/QMCWavefunctions`. Unit tests from `<source>/src/QMCWavefunctions/tests` is not built.
So better to move `add_subdirectory(XXX/tests)` to  `<source>/src/XXX/CMakeLists.txt `

2. Currently all the unit tests are installed at `<build>/tests/bin` and need to run at `<build>/tests/XXX`. Prefer to install binaries and run them at the same location which is `<build>/src/XXX/tests`.

ye-luo:
## Proposed changes
Closes #3428 part 2.
With this PR, unit test work directories will be the same as build directories.
For example, when working on tests from `src/QMCWaveFunctions/tests`
Before this PR, make and ctest in `<build>/src/QMCWaveFunctions/tests` can already run the relevant pieces only. However, to run tests manually, you would have to jump to `<build>/tests/wavefunction` and the test binary was at `<build>/tests/bin`
After this PR, the binary will be in ~~`<build>/src/QMCWaveFunctions/tests/bin`~~ `<build>/src/QMCWaveFunctions/tests` and you can run tests manually in `<build>/src/QMCWaveFunctions/tests` without jumping around.

- Build related changes

- No

epyc-server

## Checklist


ye-luo:
Start testing in-house

PDoakORNL:
I guess this restores the expected behavior with `make` and in tree builds but with empty src directories.  In do expect to see executables not copied to a project wide bin directory just in their src directory, are lots of little bin directories common?  

For me having the tests not end up in tests/bin is not great, unit tests without external file dependencies conveniently end up in one place which is great. Ninja is always run from the build root, if you want to limit it to specific targets you just specify them, `ninja test_Concurrency` for instance. So the idea of cd into the fake src directories in the build directory is already strange. 

I'd like to see the tests just find their file dependencies from project root, no switching directories at all then. i.e. from the build directory 
`tests/bin/test_wavefunction_trialwf` would just work.

But overall its slightly less convenient for me since I still had to cd to tests/wavefunction and then back to compile and so forth. If its less surprising to make users I guess it is good.  Someone else should review.

ye-luo:
I found all the test binary `<build>/tests/bin` is quite messy.
I'm trying to achieve `<project_root>/src/XXX/tests` place all the binaries in `<build>/src/XXX/tests/bin` and use `<build>/src/XXX/tests` as the working directory
.
So if you are at `<build>/src/XXX/tests`, `make` builds the tree under it and this way is more convenient than `ninja test_Concurrency` which I need to remember the name of a test. ctest runs all the tests under it. If you want to inspect output files, you are already there. If running tests manually is needed, test binary is at `./bin` which is way more convenient than looking for `<build>/tests/bin`.

Test build and run are fully contained inside `<build>/src/XXX/tests`
There will be many `tests/bin` instead of a global `tests/bin`

ye-luo:
Actually, if we don't set "<build>/src/XXX/tests/bin" cmake default output to "<build>/src/XXX/tests" I'm also OK with this option.

PDoakORNL:
So I'm up to for this even if it's slightly less convenient for Ninja.  I assume you want to get #3468 and not this one?

ye-luo:
> So I'm up to for this even if it's slightly less convenient for Ninja. I assume you want to get #3468 and not this one?

yes I need #3468 in first

PDoakORNL (approved):
This like #3468 results in less surprising make behavior.

ye-luo:
Start testing in-house

PDoakORNL (approved):
LGTM

PR #3473, opened 2021-09-23
Make InputSection::get() const

ye-luo:
## Proposed changes
For map, operator[] is "access and insert". So it can never be const.
Instead for the use case of get(), at() is more appropriate.

Other changes are just remove excessive semi-colon at the end of function definition.

- No

epyc-server

## Checklist


jtkrogel (approved):
LGTM

ye-luo:
Start testing in-house

Issue #3496, opened 2021-09-30
"VMC" vs "VMCBatched" GPU version energies differ.


aannabe:
**Describe the bug**
Total energies differ when using the GPU version of the code with `vmc` vs `vmc_batch` drivers.
The inputs are identical except for some changes in the walker counts.
Using `vmc` driver:
```
  LocalEnergy           =        -2233.5973 +/-           0.0020    1.0 
  Variance              =            19.174 +/-            0.032    1.3 
  Kinetic               =           837.337 +/-            0.076    2.5 
  LocalPotential        =         -3070.934 +/-            0.076    2.5 
  AcceptRatio           =          0.505145 +/-         0.000020    2.9 
```
Using `vmc_batch` driver:
```
  LocalEnergy           =        -2233.6471 +/-           0.0073    2.1 
  Variance              =            19.600 +/-            0.041    1.3 
  Kinetic               =           837.204 +/-            0.045    1.0 
  LocalPotential        =         -3070.851 +/-            0.046    1.0 
  AcceptRatio           =          0.505153 +/-         0.000012    1.4 
```
I have previously run tests on the same system for CPU version vs non-batched GPU version and they agreed.
So I think that the non-batched energy is the correct value.

**To Reproduce**
Both runs use the below version of the code:
```
  Git branch: update-summit-recipes
  Last git commit: 4cdbfa380f4e17e5923631c765696631ac2d7c70-dirty
  Last git commit date: Mon Aug 23 19:27:10 2021 -0400
  Last git commit subject: Update Summit build recipes.
```
Inputs and outputs can be found at:
[issue_batch.tar.gz](https://github.com/QMCPACK/qmcpack/files/7261311/issue_batch.tar.gz)

**Expected behavior**
VMC energies should agree.

**Machine:**
 - OLCF/Summit
 - Loaded modules and job details are in the attached tar file.

**System/Material**
- Charged (anion) solid RuCl3 in 2x1x2 conventional cell (593 electrons). Orbitals are from Espresso.

prckent:
Please can you update the issue to say what the system that you are simulating is?

aannabe:
Edited to give more info

ye-luo:
I'm testing a fix to non-deterministic behaviors in offload J2. Will test your case.

prckent:
Once this is understood we will need to add a small reproducer test to avoid anything similar.

prckent:
[...]

PR #3498, opened 2021-10-01
Fix and optimize offload J2

ye-luo:
## Proposed changes
517dea7 adds a fix to non-deterministic behavior of offload kernel in mw_updateVGL. The commit log includes full reproducer.
The second commit adds an optimization which reduces re-computation to only on a walker which has been migrated.
5% savings in benchmarks.

- Bugfix
- Other (please describe): Optimization

- No

Epyc-sever, Summit

## Checklist

### DIFF
--- a/src/Containers/OhmmsSoA/VectorSoaContainer.h
+++ b/src/Containers/OhmmsSoA/VectorSoaContainer.h
@@ -62,7 +62,8 @@ struct VectorSoaContainer
   {
     if (myData != in.myData)
     {
-      resize(in.nLocal);
+      if (nLocal != in.nLocal)
+        resize(in.nLocal);
       std::copy_n(in.myData, nGhosts * D, myData);
     }
     return *this;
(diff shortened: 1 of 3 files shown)


prckent:
Do you have a test that reproduces this problem? Why did our nightlies not fail?

ye-luo:
It behaves like a race, so I can only manually reproduce it as described in the commit log. The error is so subtle and often within our tolerance.

prckent:
Does this fix https://github.com/QMCPACK/qmcpack/issues/3496 ?

ye-luo:
> Does this fix #3496 ?

I handed executables to @aannabe. He is doing testing. On the other hand, the issue in this PR is understood and we don't need wait for the testing results. That is treated as a separate issue.

prckent:
Agreed, just wondering if we could put a Close tag here.

prckent (approved):

ye-luo:
Start testing in-house

PR #3512, opened 2021-10-05
Collapsed style in StlPrettyPrint.

ye-luo:
## Proposed changes
Long vectors are lengthy. walkers_per_crowd has num_crowd elements. walkers_per_rank has num_rank elements.
num_crowd can be hundreds in CPU runs and num_ranks can be tens of thousands.
Usually these vectors has very limited distinct values and are better printed in collapsed style.
[3, 3, 2, 2] is printed as [3(x2), 2(x2)]

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Utilities/StlPrettyPrint.hpp
+++ b/src/Utilities/StlPrettyPrint.hpp
@@ -18,20 +18,32 @@
 namespace qmcplusplus
 {
 
+/** collapsed vector printout
+ * [3, 3, 2, 2] is printed as [3(x2), 2(x2)]
+ */
 template<typename T>
 std::ostream& operator<<(std::ostream& out, const std::vector<T>& rhs)
 {
   out << "[";
-  if (!rhs.empty())
+  auto cursor = rhs.begin();
+  while (cursor != rhs.end())
   {
-    auto last = rhs.end();
-    last--;
-    copy(rhs.begin(), last, std::ostream_iterator<T>(out, ", "));
-    out << *last;
+    // each iteration handles one unique value
+    const T ref_value = *cursor;
+    size_t count      = 1;
+    while (++cursor != rhs.end() && *cursor == ref_value)
+      count++;
+    out << ref_value;
+    // identical elements are collapsed
+    if (count > 1)
+      out << "(x" << count << ")";
+    // if not the last element, add a separator
+    if (cursor != rhs.end())
+      out << ", ";
   }
   out << "]";
   return out;
 }
 
-}
+} // namespace qmcplusplus
 #endif
--- a/src/QMCDrivers/QMCDriverNew.cpp
+++ b/src/QMCDrivers/QMCDriverNew.cpp
@@ -121,12 +121,13 @@ void QMCDriverNew::checkNumCrowdsLTNumThreads(const int num_crowds)
  * - initialize Estimators
  * - initialize Walkers
  */
-void QMCDriverNew::startup(xmlNodePtr cur, QMCDriverNew::AdjustedWalkerCounts awc)
+void QMCDriverNew::startup(xmlNodePtr cur, const QMCDriverNew::AdjustedWalkerCounts& awc)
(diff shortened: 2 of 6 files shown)


src/Utilities/StlPrettyPrint.hpp
```diff
@@ -18,20 +18,32 @@
...
+      count++;
+    out << ref_value;
+    // identical elements are collapsed
+    if (count > 1)
+      out << "(x" << count << ")";
+    // if not the last element, add a separator
+    if (cursor != rhs.end())
+      out << ", ";
   }
   out << "]";
```
markdewing, line 44: The internals of this operator would be better extracted to a function.  More precisely, two functions - one for the original code (print all the values), and one for the run-length encoded output.  Then the << operator can call the appropriate function.   It seems likely that there would be cases where one would want to specify the type of output directly (regular or compressed).

ye-luo: All the use of this StlPrettyPrint.hpp needs to change to the new style. Keep two functions is only necessary when both use cases appear.

markdewing: Are you certain that the compressed output format is desirable every where an STL vector is printed? It seems unlikely to me (though that's just a prior belief before actually analyzing all such places).    A point of program variation has been identified, and it seems wise to express it in the code.   We may want separation between policy (what "<<" outputs for an STL vector) and implementation (how detail of a particular format work).

I imagine the structure to be something like:
```
template <typename T>
std::ostream& prettPrintSTLVector(std::ostream& out, const std::vector<T>& rhs)
{
   ... old formatting code ...
}

template <typename T>
std::ostream& prettPrintSTLVectorRLE(std::ostream& out, const std::vector<T>& rhs)
{
   ... new formatting code ...
}

template<typename T>
std::ostream& operator<<(std::ostream& out, const std::vector<T>& rhs)
{
    return prettyPrintSTLVectorRLE(out, rhs);
}
```
Then the unit test code tests the implementations (`prettyPrintSTLVector...`) and not the policy ("<<").

ye-luo: Actually even with two functions. I feel it is cumbersome to make a selection via `operator <<`. by including different header files?

ye-luo: Right now QMCDriverNew::startup is the only consumer. In general, there is no reason to print in qmcpack output in uncompressed format. As the print is always in one line and we never want long lines.
If printing is needed for debugging purpose, the develop can do whatever they need assume.
So I don't really feel the need to keep both.

PDoakORNL (approved):
I initially used this for debugging, and one of the purposes I used this for was dump out data in a format python liked. Then I used to to write output from QMCDriverNew and that appears to be the only place in the codebase it's still used. The compressed version is much nicer for general users and for the places its actually used.

I would like there still to be an operator<< available for vectors because its annoying not to have one in application. 

iostream does have some hooks that allow code to put temporary state into a stream which I think is what I would do if I wanted two operator<< implementations for std::vector I could switch between. That's definitely not needed at the moment and the c++ golf implementation will still be in the repo to recover if needed.

ye-luo:
Start testing in-house

src/Utilities/StlPrettyPrint.hpp
ye-luo: @markdewing what does RLE stands for?
If we implement two versions, there is definitely need to separate them and test against individual ones. What @PDoakORNL mentioned about state manipulation is interesting but it looks like for rainy day fun. Can we defer all these attempts?

markdewing: RLE = Run-length encoding.

markdewing: Yes, we can defer this.

PR #3514, opened 2021-10-06
Fix omp offload without cuda

ye-luo:
## Proposed changes
Intel engineers caught #3452 breaks ENABLE_OFFLOAD=ON and ENABLE_CUDA=OFF case.
#3452 was intended to change data types only but some unintended change got in.
#3452 connected MatrixUpdateOMPTarget to DiracMatrixComputeOMPTarget and caused the breakage.
So this PR 3191d09 comments out the problematic portion and restored the old code. So we can troubleshooting separately.
I expect DiracMatrixComputeOMPTarget carries issues similar to what was surfaced in #3470.

Also in 4b2ef72 I add const to RefVector type function arguments to prevent any modification.

- Bugfix

- No

epyc-server, dell-laptop

## Checklist


ye-luo:
Start testing in-house

PR #3515, opened 2021-10-06
Harmful to be smart for size 1 det.

ye-luo:
## Proposed changes
When QMCPACK is built with ENABLE_CUDA=ON, deterministic-He_ae-opt tests fails on size 1 code path in DiracDeterminant. So remove size 1 special case following #2934

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Start testing in-house

PR #3519, opened 2021-10-11
Remove newnline when closing dmcStream 

ye-luo:
## Proposed changes
dmcStream adds an additional empty line when closing the dmc.dat file.
energy.pl complains "NiO-fcc-S32-dmc.s003.dmc.dat is not rectangular"
This additional line will also be a trouble if we enable dmc.dat continuation.
So remove it.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Start testing in-house

PR #3523, opened 2021-10-12
Remove unused attributes from createSPOSetBuilder

mdewing:
These are attributes on the `sposet_collection` tag but not used at this point in the code.

- No

laptop

## Checklist


markdewing:
The attributes get handled in `LCAOrbitalBuilder`.   I think what happened is older versions of the code would use the `cuspCorrection` attribute and pass it as an argument to `LCAOrbitalBuilder`.  The attribute handling was moved, but the old location that handled them was never updated.
It gets confusing when looking through the code to see where various input features are handled and some of them are dead-ends.

PR #3535, opened 2021-10-15
Control driver debug checks from input

ye-luo:
## Proposed changes
checkLogAndGL calls were exclusive to CMAKE_BUILD_TYPE=Debug under assert(). This comes with two issues.
1. Debug runs are very slow due to heavy checkLogAndGL there is no way to opt-out
2. Non-debug runs cannot opt-in checkLogAndGL.
Overall checkLogAndGL is more suitable for being controlled from input.

`<parameter name="debug_mode">    checkGL_after_moves  </parameter>` turns on checks.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Estimators/OneBodyDensityMatrices.cpp
+++ b/src/Estimators/OneBodyDensityMatrices.cpp
@@ -27,13 +27,6 @@ using MatrixOperators::diag_product;
 using MatrixOperators::product;
 using MatrixOperators::product_AtB;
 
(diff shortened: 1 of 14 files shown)


markdewing:
On more thought, I wonder if naming the tag something like `debug_checks` would be better than `debug_mode`?  The operation is additional checks for consistency at discrete points in the code.  The checks are not as pervasive as `mode` seems to imply (at least to me).

markdewing:
I'm good with the changes. The points I brought up have all been addressed.

ye-luo:
@PDoakORNL I also fixed checkLogAndGL to make sure the offload mixed precision can pass.

Issue #3529, opened 2021-10-13
Retire QMCOptimize and QMCOptimizeBatched


ye-luo:
testDerivOptimization.h is the only relevant consumer and it can probably be connected at `QMCFixedSampleLinearOptimizeBatched::run` directly.
This helps minimizing code paths.

With QMCOptimize and QMCOptimizeBatched removed, a lot of unused codes under `src/Optimize` can be removed as well.

@markdewing thoughts?

PR #3540, opened 2021-10-19
Move parameter gradient testing code

mdewing:
This is the first part to addressing #3529.  A future PR can deal with removing the unused code.

The still-used functionality accessed through the QMCOptimize and QMCOptimizeBatched drivers is for testing the parameter gradients.  The code compares the parameter gradient (numerical vs. analytical) and also produces an output file for the test system.
The code was moved from `src/Optimize/testDerivOptimization.h` to `src/QMCDrivers/WFOpt/GradientTest.{h,cpp}`

The functionality is now accessed through the batched linear optimizer.  Two consequences:
1.  The parameter gradient code will not run in mixed precision, because the linear optimizer does not work there. 
2. The input file format changed.  I think it's safe to assume the only users of this code are in the test system (Generating this input is not in Nexus)  Though until the old code is removed, the previous file format will still work.

Implementation note:  The previous code was templated on the precision used.  This code uses a type alias for `QMCTraits::RealType`.   In general, we are moving away from QMCTraits and towards more templates (on precision). However, the new code is not accessible in mixed precision, there doesn't seem to be a strong need to easily access it (other than it should compile.)   Also,  the new code depends on the definition of QMCCostFunctionBase, which uses `QMCTraits::RealType` (as a template parameter to `CostFunctionBase`, from which it inherits).  That would need to be changed as a prerequisite.

- Moving code.  More disruptive than refactoring.

- No.

desktop

## Checklist

### DIFF
--- a/src/Estimators/OneBodyDensityMatrices.cpp
+++ b/src/Estimators/OneBodyDensityMatrices.cpp
@@ -27,13 +27,6 @@ using MatrixOperators::diag_product;
 using MatrixOperators::product;
 using MatrixOperators::product_AtB;
 
-UPtr<OneBodyDensityMatrices> createOneBodyDensityMatrices(OneBodyDensityMatricesInput&& obdmi,
-                                                          const PtclOnLatticeTraits::ParticleLayout_t& lattice,
-                                                          const SpeciesSet& species)
-{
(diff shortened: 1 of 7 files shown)


ye-luo:
Start testing in-house

PR #3571, opened 2021-10-31
Harden ROCM in CMake.

ye-luo:
## Proposed changes
Confirm needed packages at CMake instead of during compilation.

- Bugfix
- No

epyc-server

## Checklist


ye-luo:
Start testing in-house

prckent (approved):
LGTM

I assume this works with both ROCm 4.5 and 4.4 (current spock).

PR #3572, opened 2021-11-01
Introduce QMC_OFFLOAD_MEM_ASSOCIATED cmake option.

ye-luo:
## Proposed changes
Some OpenMP runtime library implementations may not have good interoperability with vendor runtime systems.
The outstanding issue is that a device pointer allocated via OpenMP runtime library cannot be de-referenced in a kernel or vendor library created with the vendor native programming model. https://github.com/ROCm-Developer-Tools/aomp/issues/253
If the reverse way works, namely a device pointer allocated by the vendor native programming model can be de-referenced in OpenMP offload region. We can always allocate memory from the vendor native programming model and associate it to OpenMP runtime.

So QMC_OFFLOAD_MEM_ASSOCIATED=ON makes ENABLE_CUDA=ON+QMC_CUDA2HIP+ENABLE_OFFLOAD=ON largely working when ROCM shipped LLVM, AOMP, upstream LLVM offload compilers are used.

- Bugfix
- Build related changes

- No

epyc-server

## Checklist


prckent (changes requested):
Please make the setting for ENABLE_CUDA=ON+QMC_CUDA2HIP+ENABLE_OFFLOAD=ON  automatic. We can adjust in future it needed.

The CMake way to support this would be to add a test for the workaround requirement and set the flag automatically, with some provision for user override. Is it possible to add a compile time test or does it require running on the GPU, as I suspect?

ye-luo:
This option is not needed when Cray compilers are used. Cray is also identified as LLVM by cmake. We may detect cray environment but it may cause issue if there is a cray wrapper for rocm compilers. So baking  such logic into our CMake is error prone. Second, testing the feature requires running the code on a node with a GPU. The test cannot be made reliable against various settings. So prefer to keep it simple and hopefully things improves in the following months and we never need to expose this option to users.

prckent:
How about printing a informative message when ENABLE_CUDA=ON+QMC_CUDA2HIP+ENABLE_OFFLOAD=ON is set? ("Detected xyz. As of 11/2021, this is not needed for Cray-provided compilers but is needed for AMD and mainline LLVM."

Please list the current compiler-option combinations in CMakeLists.txt so we have some documentation to follow.

Unfortunately I am confident that I will forget to do set this at least once and bug you if there is no message.

Hopefully this can all be deleted in future.

ye-luo:
Added a warning printing. I'm 100% sure I will forget as well. If I find a better way, I will update the CMake.

prckent (approved):

ye-luo:
Test this please

PR #3573, opened 2021-11-01
Add deterministic test of the excited state for batched driver

Hyeondeok-Shin:
## Proposed changes

There are a lot of deterministic tests for batched driver now but missing tests of the observables for batched driver needs to be added.
This PR is to add deterministic test of the excited state which is still missing for batched driver tests. 
The other missing tests will be added in the order of priority.
  

- Other (please describe): Add tests

- No

 Workstation

## Checklist


prckent:
Start testing in-house

prckent:
Test this please

ye-luo:
Test this please

prckent (commented):
Is the following documented anywhere?
```
                  <occupation mode="excited" spindataset="0">
                   -8 9
                  </occupation>
```

prckent (commented):
I think we might be missing a short/long test of this specific excitation. How do we know that the batched implementation is correct &/or consistent with the unbatched case? This test is a good check to catch future changes, but if we don't already have a statistical batched excitation test or other way to know that the excitation code is correctly invoked, then the test is needed.

Hyeondeok-Shin:
For the wavefunction example, only (ti, bi) notation is included in the manual, but nexus example shows both (ti, bi) and orbital index notation. (https://qmcpack.readthedocs.io/en/develop/lab_excited.html?highlight=excited#basics-and-excited-state-experiments)  We may add example for the orbital index notation to the manual. 

 Right. short/long statistical test for the excited state is needed for both batched and unbatched code to see if it is working well. I will add those after this PR.

prckent:
@Hyeondeok-Shin When you make the statistical cases for the batched code, please can you add a short description in the main part of the manual? The lab should not be the first place this is mentioned. Somewhere in the "Trial wavefunction specification" section looks appropriate.

prckent (approved):
LGTM.
Additional tests+docs for later PR as discussed.

PR #3574, opened 2021-11-01
Change LogDet reduction pattern.

ye-luo:
## Proposed changes
The code before change was fine. But would like to avoid CUDAfill_n and atomic.
CUDAfill_n calls cudaMemset which uses the default stream and blocks other streams.
CUDAfill_n was implemented to prevent getting NaN when containers allocates memory. So better to avoid using it outside containers.
In addition, complexDetLog is a very light compute, it is better to have multiple iterations instead of employing more thread blocks.

- No

epyc-server

## Checklist


src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -66,35 +65,38 @@ __global__ void computeLogDet_kernel(const int n,
...
-  logdet_vals[threadIdx.x].x = 0.0;
-  logdet_vals[threadIdx.x].y = 0.0;
-  T lu_diag;
-  int n_index = threadIdx.x + block_num * COLBS;
-  if (n_index < n)
+  const int tid      = threadIdx.x;
+  logdet_vals[tid].x = 0.0;
+  logdet_vals[tid].y = 0.0;
+
+  const int num_col_blocks = (n + COLBS - 1) / COLBS;
```
PDoakORNL, line 76: This needs something like
```c++
static_assert(COLBS % 2 == 0);
```
And a warning for the unwary that this reduction stratedgy only works for powers of 2.  Also do you have profiling information of how much this is worth?  While I too believe it should be faster that's just speculation.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
```diff
@@ -66,35 +65,38 @@ __global__ void computeLogDet_kernel(const int n,
...
+    __syncthreads();
+    if (tid < iend)
     {
-      block_sum_log_det.x += logdet_vals[iv].x;
-      block_sum_log_det.y += logdet_vals[iv].y;
+      logdet_vals[tid].x += logdet_vals[tid + iend].x;
+      logdet_vals[tid].y += logdet_vals[tid + iend].y;
     }
-    // No atomicAdd in cuda 10 for cuComplex/cuDoubleComplex
-    atomicAdd((double*)&(logdets[iw].x), block_sum_log_det.x);
```
PDoakORNL, line 95: I agree it best to remove the atomic add.

src/Platforms/CUDA/CUDAfill.hpp
```diff
@@ -19,6 +19,14 @@
 
 namespace qmcplusplus
 {
+/** fill device memory with a given value.
```
PDoakORNL, line 22: I also dislike cudafill it should be async.

src/QMCWaveFunctions/detail/CUDA/cuBLAS_LU.cu
ye-luo: I added a static assert to check exactly power of 2. I have not not looked at the performance but it should be anyway cheaper than getrf.

PDoakORNL (approved):
I did the benchmarking myself.  Seems to make no difference at all but I think dropping the defensive CUDAfill_n and possibly freeing up some compute on the GPU is worth it.

ye-luo:
Start testing in-house

PR #2947, opened 2021-02-23
MSD arbitrary species evaluation (1/N)

shivupa:
## Proposed changes

This PR generalizes the functions in `MultiSlaterDeterminantFast` to work with an arbitrary number of particle types. This lifts the previous restriction of alpha/beta electrons. This is related to #2935.

## What type(s) of changes does this code introduce?
- New feature
- Refactoring (no functional changes, no api changes)
### Does this introduce a breaking change?
- No, this is backwards compatible.
## What systems has this change been tested on?
Workstation Ryzen 5900x/ GCC 10.2 Real build

## Checklist

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- No. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate). This is not needed I think.

In future PR(s), further generalization is necessary. This involves most of what needs to be touched in `MultiSlaterDeterminantFast.*` so I am opening this for review.

PR #3577, opened 2021-11-01
fix MSDFast evaluateDerivatives for arbitrary species

camelto2:

## Proposed changes

This PR fixes an issue for MultiSlaterDeterminantFast when trying to do optimizations with spinor wavefunctions which have only one electron group (or any MSD WF with more than 2 electron groups (e.g. up, down, positron). Jastrow and/or CI coefficient optimizations would fail in the evaluateDerivatives and evaluateDerivativesWF calls because of a runtime error if there weren't exactly 2 determinant groups. 

The runtime error was added when the MultiSlaterDeterminantFast interface was updated (see #2947) because the MultiDiracDeterminant::evaluateDerivatives (which gets the orbital parameters to optimize) call still uses an old design requiring up and down determinants always. However, this runtime error should only be triggered if we are actually trying to optmize the orbitals inside the MultiDiracDeterminant. We should still be able to optmize jastrows and CI coefficients even if we don't optimize the orbitals. This fix should address that. 

I honestly thought this was already merged in a previous PR. I guess I overlooked it. 

- Bugfix
- Refactor

- No

my workstation, gnu9.2.0

## Checklist


rcclay (approved):

rcclay:
Test this please.

ye-luo (approved):

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
```diff
@@ -962,17 +962,27 @@ void MultiSlaterDeterminantFast::evaluateDerivatives(ParticleSet& P,
...
-  {
-    throw std::runtime_error(
-        "MultiSlaterDeterminantFast::evaluateDerivatives only compatible with two quantum particle types.");
-  }
-  else
+  //Currently, the MultiDiracDeterminant::evaluateDerivatives works with a legacy design, essentially requiring only up and down determinants.
+  //e.g. for spinor cases, we only have one determinant so this interface doesn't work.
+  //Here we throw an error only if the optimization is turned on for MultiDiracDeterminants until the code is updated
+  bool all_Optimizable = true;
+  for (size_t id = 0; id < Dets.size() && all_Optimizable; id++)
```
PDoakORNL, line 969: this is a little baroque or perhaps even medieval
```c++
for(auto det : Dets)
  if(!det->Optimizable)
   return;
nameOfDerivativeOnlyCalculatedForSpinSystems();
//or
if(std::all_of(Dets.begin(), Dets.end(), [](auto det){ return det->Optimizable; }))
  nameOfDerivativeOnlyCalculatedForSpinSystems();
```

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
```diff
@@ -962,17 +962,27 @@ void MultiSlaterDeterminantFast::evaluateDerivatives(ParticleSet& P,
...
+    all_Optimizable = Dets[id]->Optimizable;
+
+  if (all_Optimizable)
```
PDoakORNL, line 972: This entire section could/should be refactored into a private method, its signature and name might even give offer up some clue as to what if any of the previous computation it has data dependence on and what it is.  I guess its the partial derivative with respect to the other spin multideterminant.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
```diff
@@ -1046,20 +1056,30 @@ void MultiSlaterDeterminantFast::evaluateDerivativesWF(ParticleSet& P,
...
-  {
-    throw std::runtime_error(
-        "MultiSlaterDeterminantFast::evaluateDerivativesWF only compatible with two quantum particle types.");
-  }
-  else
+  //Currently, the MultiDiracDeterminant::evaluateDerivativesWF works with a legacy design, essentially requiring only up and down determinants.
+  //e.g. for spinor cases, we only have one determinant so this interface doesn't work.
+  //Here we throw an error only if the optimization is turned on for MultiDiracDeterminants until the code is updated
+  bool all_Optimizable = true;
+  for (size_t id = 0; id < Dets.size() && all_Optimizable; id++)
```
PDoakORNL, line 1063: as above

PDoakORNL (commented):
I really think that's some convoluted code to check all elements have true for a property.

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
camelto2: updated.
Had to use (auto cost& det : Dets) since Dets is a vector of unique_ptr. (auto det : Dets) tries to do a copy which fails since they are unique_ptr

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
camelto2: added new private function evaluateMultiDiracDeterminantDerivatives to handle this

src/QMCWaveFunctions/Fermion/MultiSlaterDeterminantFast.cpp
camelto2: added new private function evaluateMultiDiracDeterminantDerivativesWF to handle this

ye-luo (approved):

ye-luo:
Test this please

PR #3580, opened 2021-11-02
Change indexing for multispecies pair correlation estimator

shivupa:
## Proposed changes
The indexing for the pair correlation estimator was incorrect if more than two particle types were used.

- Bugfix

- No

Workstation AMD Ryzen
## Checklist


shivupa:
I am not sure if this needs a test. If no then I can convert this from a draft PR otherwise I can set up a test.

markdewing:
It looks to me like the `pair_id` computed here should match the index value in `set_norm_factor` (?)

I wrote a little python script to verify the pair_id formula, and it doesn't seem to match for ngroups > 2.

Script:
```
# Test formulas for pair_id in QMCHamiltonians/PairCorrEstimator.cpp
# The formula in the "evaluate" function should match the index computed in "set_norm_factor"
# There is an offset of +1 to indx in `set_norm_factor`, but the code uses `pair_id+1` in accessing it, 
#    so the indices here should be consistent

ngroup = 4

print("ig jg idx old  new")
idx = 0
for ig in range(ngroup):
    for jg in range(ig,ngroup):
        old_formula = ig*(ig+1)//2 + jg
        new_formula = (ngroup*(ngroup-1))//2 - ((ngroup - jg)*(ngroup - jg - 1))//2 + ig
        print(ig, " " ,jg, " ",idx, " ",old_formula," ", new_formula)
        idx += 1
```

Columns in the output:
`ig`,`jg` are the species (group) indices
`idx` is `indx-1` from `set_norm_factor`
`old` is the old computation of `pair_id`
`new` is the new computation of `pair_id`

Output:
```
ig jg idx old  new
0   0   0   0   0
0   1   1   1   3
0   2   2   2   5
0   3   3   3   6
1   1   4   2   4
1   2   5   3   6
1   3   6   4   7
2   2   7   5   7
2   3   8   6   8
3   3   9   9   9
```

shivupa:
Sorry i was working only with 3 particles of only 3 groups which ended up constraining `j>i` but this wasn't general. The general case is:
```
ngroup = 4

print("ig jg idx old  new")
idx = 0
for ig in range(ngroup):
    for jg in range(ig,ngroup):
        old_formula = ig*(ig+1)//2 + jg
        new_formula = 0
        if jg < ig:
            new_formula = (ngroup*(ngroup-1))//2 - ((ngroup - jg)*(ngroup - jg - 1))//2 + ig
        else:
            new_formula = (ngroup*(ngroup-1))//2 - ((ngroup - ig)*(ngroup - ig - 1))//2 + jg
        print(ig, " " ,jg, " ",idx, " ",old_formula," ", new_formula)
        idx += 1
        
```
output
```
ig jg idx old  new
0   0   0   0   0
0   1   1   1   1
0   2   2   2   2
0   3   3   3   3
1   1   4   2   4
1   2   5   3   5
1   3   6   4   6
2   2   7   5   7
2   3   8   6   8
3   3   9   9   9
```

shivupa:
Adding logging to the `set_norm_factor` and `evaluate` 
Set norm
```
i0 j0       1
i0 j1       2
i0 j2       3
i1 j1       4
i1 j2       5
i2 j2       6
```

evaluate (which should be index -1)
```
i1 j0       1
i2 j0       2
i2 j1       4
```

Thanks for taking a look at this!

PDoakORNL:
This requires a unit test.  That much discussion makes it clear.

markdewing:
The calculation of the pair_id could be moved to a function, and then that function can be unit-tested.

shivupa:
Updated!

markdewing:
A comment on the design of `gen_pair_id`:   The function depends on `ig` and `jg` as input parameters and `num_species` as a class member.   If `num_species` were made an argument, the function would be independent of the class.  There is a shift towards this functional-style programming in software engineering.
At a technical level, this function could then be made a static class member.  Then it could be called without constructing the class (which is, as you've seen, complicated) and `gen_pair_id` could be tested in isolation.

This doesn't need to be changed for this PR - the code is good enough as it is. (It's an improvement on the previous code. And all the work to construct the PairCorrEstimator is necessary if further tests of the estimator are added.)

markdewing (dismissed):

src/QMCHamiltonians/PairCorrEstimator.cpp
```diff
@@ -102,6 +102,19 @@ PairCorrEstimator::PairCorrEstimator(ParticleSet& elns, std::string& sources)
 
 void PairCorrEstimator::resetTargetParticleSet(ParticleSet& P) {}
 
+// The value should match the index to norm_factor in set_norm_factor
+const int PairCorrEstimator::gen_pair_id(const int& ig, const int& jg)
+{
+  if (jg < ig)
+  {
+    return ((num_species * (num_species - 1)) / 2) - (((num_species - jg) * (num_species - jg - 1)) / 2) + ig;
```
ye-luo, line 110: Please remove {} around one line and reduce () as well. Do the same for else as well.

src/QMCHamiltonians/PairCorrEstimator.h
```diff
@@ -40,6 +40,9 @@ class PairCorrEstimator : public OperatorBase
   /* evaluate the pair correlation functions */
   Return_t evaluate(ParticleSet& P) override;
 
+  /// generate the unique pair id from the group ids of particle i and j
+  const int gen_pair_id(const int& ig, const int& jg);
```
ye-luo, line 44: POD type can just pass value. No need of reference. The function needs to be marked as cost member. Return "const" is not needed and has no effect.

src/QMCHamiltonians/tests/test_PairCorrEstimator.cpp
```diff
@@ -203,4 +226,110 @@ TEST_CASE("Pair Correlation", "[hamiltonian]")
...
+
+  REQUIRE(elec->SameMass);
+  REQUIRE(elec->getName() == "e");
+
+  // Move the particles manually onto B1 lattice
+  // NB: Spins are grouped contiguously
+  // Up spins
+  elec->R[0][0] = 0.0;
+  elec->R[0][1] = 0.0;
+  elec->R[0][2] = 0.0;
```
ye-luo, line 274: `elec->R[0] = {0.0, 0.0, 0.0}` is more compact.

ye-luo (changes requested):

shivupa:
Ok Thanks that was helpful. I updated with a static function and this did clean up the test a lot!

ye-luo (approved):

ye-luo:
Test this please

PR #3583, opened 2021-11-03
Test compile main after compile option setting.

ye-luo:
## Proposed changes
When incorrect compile options are passed through. FindLAPACK fails simply because it is the first library which tries to invoke the compiler. So it is better to have a check before any library searching.

- New feature

- No

epyc-server

## Checklist


ye-luo:
Start testing in-house

prckent (approved):
I think this is a good defensive measure to add.

Issue #3120, opened 2021-04-20
ntest_simulation_module failing without MPI


prckent:
**Describe the bug**
https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=ntest_nexus_simulation_module&date=2021-04-20

The most recent updates #3108 fixed the mpi situation but not the no mpi situation.

Edit: Interestingly the Intel2019.1 build passed while gcc, llvm, and pgi builds did not. But perhaps enough of Intel MPI was picked up even though not enabled(???) 

```
Test output

Test name     : simulation_module
Test sublabel : test_execute
Test exception: "AssertionError: "
Test backtrace:
  File "/scratch/pk7/QMCPACK_CI_BUILDS_DO_NOT_REMOVE/qmcpack/nexus/bin/nxs-test", line 478, in run
    self.operation()
  File "/scratch/pk7/QMCPACK_CI_BUILDS_DO_NOT_REMOVE/qmcpack/nexus/bin/nxs-test", line 1064, in simulation
    nunit('execute')
  File "/scratch/pk7/QMCPACK_CI_BUILDS_DO_NOT_REMOVE/qmcpack/nexus/bin/nxs-test", line 349, in nunit
    run_external_unit_test(test_name,unit_test)
  File "/scratch/pk7/QMCPACK_CI_BUILDS_DO_NOT_REMOVE/qmcpack/nexus/bin/nxs-test", line 388, in run_external_unit_test
    unit_test()
  File "/scratch/pk7/QMCPACK_CI_BUILDS_DO_NOT_REMOVE/qmcpack/nexus/tests/unit/test_simulation_module.py", line 2386, in test_execute
    assert(open(outfile,'r').read().strip()=='run')

Test status: fail

```

**To Reproduce**

Latest develop, software versions as nitrogen/sulfur. (ornl_versions.sh)

**Expected behavior**

Test passes

**System:**
 - system name [e.g. fusiont5, summit]
 - modules loaded [e.g. output of `module list`]
 - other systems where this is reproducible [e.g. "my laptop", "none"]

**Additional context**

jtkrogel:
Nexus assumes you are going to run MPI applications with it.  MPI availability is basically a requirement then for a correct Nexus install, which this failing test exposes.

For the purpose of running tests with QMCPACK "no MPI", perhaps a flag should be passed to `nxs-test` (e.g. `--no_mpi`) by ctest for these builds?  The flag could modify the Nexus tests to not require MPI.

[...]

PR #3584, opened 2021-11-03
Nexus: no longer require mpirun for test passage

jtkrogel:
## Proposed changes

Addresses #3120.  Nexus tests no longer strictly require mpirun.  If mpirun is detected in the path, it is executed in the tests ("mpirun -np 1 echo run").  Otherwise a serial test is performed ("echo run").

- Bugfix

- No

Laptop

## Checklist

### DIFF
--- a/nexus/tests/unit/test_simulation_module.py
+++ b/nexus/tests/unit/test_simulation_module.py
@@ -2361,8 +2361,11 @@ def test_execute():
 
     tpath = testing.setup_unit_test_output_directory('simulation','test_execute',divert=True)
 
+    import shutil
+    serial = shutil.which('mpirun') is None
+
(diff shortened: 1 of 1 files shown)


prckent (approved):
Thanks Jaron. I think this should be a more than good enough fix for the problems we have been seeing. It is simpler and more elegant than what I was envisaging (command line option) so I think we should go with your solution.

ye-luo:
Test this please

ye-luo (approved):

williamfgc:
Fixed unstable [spock nightly test](https://cdash.qmcpack.org/CDash/viewTest.php?onlydelta&buildid=259587), nice!

PR #3585, opened 2021-11-03
Nexus: Resource lock nexus examples

prckent:
## Proposed changes

Speculative update to serialize nexus example tests and avoid potential race condition. Result should be no nexus examples sporadically failing on cdash. Only the example tests are serialized so the effect on overall test runtime is minimal.

- Bugfix

- No

## Checklist


ye-luo:
Test this please

ye-luo (approved):
Hopefully this will be all what we need.

Issue #3475, opened 2021-09-23
Need unit test for MomentumDistribution


ye-luo:
**Describe the bug**
Momentum Distribution was allowed into the codebase without unit tests. It needs some.  See test_SpinDensityNew.cpp, test_OneBodyDensityMatrices.cpp, etc. For guidance.

**To Reproduce**
no `src/Estimators/tests/test_MomentumDistribution.cpp`

**Expected behavior**
All new code must have unit testing.

**Is your feature request related to a problem? Please describe.**
Work left from #3447

jtkrogel:
I'm starting this now, but there is a strong interaction between tests that follow a similar pattern of use of built/accessed ParticleSet/TrialWavefunction components.

For example, this test case does nothing, but introducing it breaks existing OneBodyDensityMatrices tests:

```
TEST_CASE("MomentumDistribution::MomentumDistribution", "[estimators]")
{
  auto lattice     = testing::makeTestLattice();

  Communicate* comm;
  comm = OHMMS::Controller;
  outputManager.pause();

  MinimalParticlePool mpp;
  ParticleSetPool particle_pool = mpp(comm);
  MinimalWaveFunctionPool wfp;
  WaveFunctionPool wavefunction_pool = wfp(comm, particle_pool);
  auto& pset                         = *(particle_pool.getParticleSet("e"));
  auto& wf_factory                   = *(wavefunction_pool.getWaveFunctionFactory("wavefunction"));
}
```

Resulting output from DM tests:
```
31: -------------------------------------------------------------------------------
31: OneBodyDensityMatrices::evaluateMatrix
31: -------------------------------------------------------------------------------
31: /home/j1k/projects/ecp/16_batched_nofk/qmcpack_batched_nofk/src/Estimators/tests/test_OneBodyDensityMatrices.cpp:482
31: ...............................................................................
31: 
31: /home/j1k/projects/ecp/16_batched_nofk/qmcpack_batched_nofk/src/Estimators/tests/test_OneBodyDensityMatrices.cpp:307: FAILED:
31:   CHECK( returned_data[id] == Approx(data[id]) )
31: with expansion:
31:   -1.5547128175 == Approx( -0.4029298437 )
```

jtkrogel:
After discussion w/ @PDoakORNL, it looks like this is likely to be an issue w/ the global RNG interacting w/ ParticleSet instantiation.

[...]

PR #3588, opened 2021-11-04
Support initializer list for ParticleAttrib for tests with PSet

PDoakORNL:
## Proposed changes

Add initializer list support for ParticleAttrib so ParticleSet R can be easily initialized using the native C++ initializer list format. This requires small changes in OhmmsVector and TinyVector. I have added unit tests to cover the changes in OhmmsVector.  I have not added unit tests to TinyVector since I just added the default move constructor and move assignment to the class.

As found by @jtkrogel in #3475 There is a surprising issue with writing unit tests.

We have a global Random object to support legacy code that wants to just grab a nondeterministic random number this is bad practice for many reasons but in tests that rely on boot strapping objects through the legacy put methods i.e. with MinimalWhateverPools or larger mocking of the application environment it means you will get ParticleSets whose initialized R vary based runtime ordering.
i.e. We can't reason about the state of the global Random in tests. 

Also due to use of FakeRandom in unit tests and the fact that we bundle unit tests into module level executables with many many tests cases we break the ODR in a way that is not caught at compile or link time. The type and accessibility global Random is unpredictable for a particular test. Whether it will be the same as the legacy object using global random is unpredictable. So we can't just initialize it to a known state in each test case.

- Bugfix
- New feature
- Documentation changes

- No

Leconte

## Checklist

### DIFF
--- a/src/Containers/OhmmsPETE/OhmmsVector.h
+++ b/src/Containers/OhmmsPETE/OhmmsVector.h
@@ -43,7 +43,7 @@ public:
   typedef Vector<T, Alloc> This_t;
 
   /** constructor with size n*/
-  explicit inline Vector(size_t n = 0, Type_t val = Type_t()) : nLocal(n), nAllocated(0), X(nullptr)
+  explicit inline Vector(size_t n = 0, Type_t val = Type_t()) : nLocal(n)
   {
     if (n)
     {
@@ -53,10 +53,10 @@ public:
   }
 
   /** constructor with an initialized ref */
-  explicit inline Vector(T* ref, size_t n) : nLocal(n), nAllocated(0), X(ref) {}
+  explicit inline Vector(T* ref, size_t n) : nLocal(n), X(ref) {}
 
   /** copy constructor */
-  Vector(const Vector& rhs) : nLocal(rhs.nLocal), nAllocated(0), X(nullptr)
+  Vector(const Vector& rhs) : nLocal(rhs.nLocal)
   {
     if (nLocal)
     {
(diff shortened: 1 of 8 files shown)


src/Containers/OhmmsPETE/OhmmsVector.h
```diff
@@ -79,10 +79,29 @@ class Vector
...
+   *  and nontrivial nested elements with move assignment operators.
+   */
   Vector(std::initializer_list<T> ts)
   {
     if (qmc_allocator_traits<Alloc>::is_host_accessible)
-      std::copy_n(ts.begin(), ts.size(), X);
+      if constexpr (std::is_trivial<T>::value)
+      {
+        std::size_t num_elements = ts.size();
+        resize_impl(num_elements);
```
ye-luo, line 91: resize_impl requires num_elements protected not zero.

src/Containers/OhmmsPETE/OhmmsVector.h
```diff
@@ -79,10 +79,29 @@ class Vector
...
+        std::copy_n(ts.begin(), ts.size(), X);
+      }
+      else
+      {
+        std::size_t num_elements = ts.size();
+        resize_impl(num_elements);
```
ye-luo, line 97: Similar here.

src/Containers/OhmmsPETE/OhmmsVector.h
```diff
@@ -79,10 +79,29 @@ class Vector
...
+        auto ts_it = ts.begin();
+        for (int i = 0; i < num_elements; ++i, ++ts_it)
+        {
+          auto& this_element = (*this)[i];
+          this_element       = std::move(*ts_it);
```
ye-luo, line 102: use line 101 and 102 in one?

src/Estimators/tests/test_OneBodyDensityMatrices.cpp
```diff
@@ -506,6 +506,18 @@ TEST_CASE("OneBodyDensityMatrices::evaluateMatrix", "[estimators]")
...
   OneBodyDensityMatrices obdm(std::move(obdmi), pset_target.Lattice, species_set, wf_factory, pset_target);
   auto& trial_wavefunction = *(wavefunction_pool.getPrimary());
+
+  // We can't reason about the state of the global Random in tests. A User can run only some tests,
+  // new tests will get added, other tests modified so global Random is called more times or fewer.
+  // Also due to use of FakeRandom in unit tests in other tests of this executable its difficult
+  // to know which global Random this test will have have access to. So trying to initialize it to
+  // a known state is not maintainable.
+  // So we must initialize particle positions to known values.
+  pset_target.R = ParticleSet::ParticlePos_t{{1.751870349, 4.381521229, 2.865202269}, {3.244515371, 4.382273176, 4.21105285},
```
ye-luo, line 516: if you remove ParticleSet::ParticlePos_t will the compiler figure it out?

ye-luo (changes requested):
Good change for a desired feature. Only minor change requested.

ye-luo (approved):

ye-luo:
Start testing in-house

PR #3591, opened 2021-11-05
Updated GitHub actions rst file afqmc

walshmm:
## Proposed changes

This PR updates documentation around Github-Actions and the tests that it runs.
It provides some reference information allowing contributors to more easily understand and reproduce the tests ran on their PR's.

- Documentation changes

- No

* Ubuntu 20.04

## Checklist


docs/github_actions.rst
```diff
@@ -256,11 +260,175 @@ gpu-cuda (gcc-complex-gpu-cuda-full)
...
+
+
+gpu-cuda (clang14_dev-real-gpu-enable-cuda-afqmc-offload-full)
+""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
++---------------+-------------------------------------------------------------------------------------------------------------------------------------+
+| Compiler      | Clang 14                                                                                                                            |
++---------------+-------------------------------------------------------------------------------------------------------------------------------------+
+| Build Command | `cmake -GNinja -DCMAKE_C_COMPILER=clang-14 \                                                                                        |
+|               | -DCMAKE_CXX_COMPILER=clang++-14 \                                                                                                   |
+|               | -DMPI_C_COMPILER=/usr/lib64/openmpi/bin/mpicc \                                                                                     |
```
ye-luo, line 402: No `MPI_C_COMPILER, MPI_CXX_COMPILER` please. If you need to change the compiler underneath MPI wrapper. Use OMPI_C and OMPI_CXX. Since this needs to be aligned with run_step.sh, there is no need to change in this PR.

ye-luo (approved):

docs/github_actions.rst
prckent: This needs to be updated. I think it was updated in the script so perhaps this doc is all that needs updating.
As background, we treat MPI as part of the compiler. The multiple specifications route is more error prone, more complex for users, and also happens to be untested and unsupported :-)

ye-luo:
Test this please

docs/github_actions.rst
williamfgc: @walshmm to add some background. CMake goes after `mpicc` which wasn't defined on this system, so it needed to be added explicitly. We could add a note up front telling users to follow the documentation for building QMCPACK, not this page as this is for very specific systems. Another option is to export `mpicc`, `mpicxx` and `mpirun` on run_step.sh, but I try not to use env variables (that's probably more invasive if users are following this), thoughts?

ye-luo: CMake look for mpicc/mpicxx only because CC=mpicc CXX=mpicxx environment variables are set.
That is harmful. I think we should remove them and explicitly set CMAKE_C_COMPILER=mpicc and CMAKE_CXX_COMPILER=mpicxx directly if needed.

Here is what I do.
In case of MPI
```
cmake -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx ..
```

In case of using OpenMPI wrapper for GCC but I need Clang.
```
export OMPI_CC=clang
export OMPI_CXX=clang++
cmake -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx ..
```

incase of no-MPI
```
cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DQMC_MPI=OFF ..
```

ye-luo: @williamfgc @walshmm will you update `run_step.sh` accordingly in this PR or a separate one?

williamfgc: @ye-luo I'm working on refactoring the current CI, so probably a few separate ones to follow.

ye-luo:
Test this please

ye-luo (approved):

PR #3592, opened 2021-11-05
Add unit tests for MomentumDistribution

jtkrogel:
## Proposed changes

Addresses #3475.  Verified absence of interaction with the new 1RDM tests (order independence, etc).

- No

Laptop

## Checklist


PDoakORNL (approved):
This looks good to me. 
The logic here is that its passes statistical testing so we assume that it is a correct implementation and so with known input the result of a call to accumulate should not change.

ye-luo:
Test this please

ye-luo (approved):

PR #3603, opened 2021-11-11
Make it possible to get a working ParticleSet for CPU fallback

PDoakORNL:
## Proposed changes
ParticleSet gets used to actually pass coordinates and via manipulation of its state machine access to distance tables to for those coordinates, this leads to access violations and/or synchronization errors if the recipient of the ParticleSet argument expects a different DistanceTable DTMode than the table actually has which is for instance true of any CPU only SPOSet implementation when ENABLE_OFFLOAD=1. Some/Most Estimators of ParticleSet arguments do not add their required distance tables they just assume they are there in the walker ParticleSet and of a usable implementation.

This allows us to get a a for sure working ParticleSet to use as the coordinate argument for an estimator. We need with as little of the unecessary other data in ParticleSet. This is also nice since we don't manipulate the state machine of an object that should really be const with respect to estimator accumulation.

I expect to need this further in the near/mid term so I did this "properly" with tag evaluation.

In the long term ParticleSet needs to be refactored/redesigned and functions that consume need to be fed arguments that reflect their actual data dependencies.
I'd also point out that makeMove does distance table updating whether or not they are used for the evaluation taking the ParticleSet argument which is bad.

- New "feature"
- Documentation changes

- No

Leconte clang14
## Checklist

### DIFF
--- a/src/Particle/ParticleSet.h
+++ b/src/Particle/ParticleSet.h
@@ -37,7 +37,6 @@ namespace qmcplusplus
 class DistanceTableData;
 class ResourceCollection;
 class StructFact;
-class CompositeSPOSet;
 
 /** Specialized paritlce class for atomistic simulations
  *
@@ -76,21 +75,6 @@ public:
   /// ParticleSet that will work with a CPU only SPOSet like CompositeSPOSet and who knows what else as we port Estimators.
   struct MinimalCPU
   {};
-
-  /** Compilers not compliant with c++17 17.8.3.2 (Explicit Specialization section)  may claim these can 
-   *  not be specialized in class scope.  They are wrong.
-   */
-  template<class T>
-  struct minimal_trait
-  {
-    using tag = MinimalDefault;
-  };
-  template<>
-  struct minimal_trait<CompositeSPOSet>
-  {
-    using tag = MinimalCPU;
-  };
-
 
   ///quantum_domain of the particles, default = classical
   quantum_domains quantum_domain;
@@ -765,6 +749,12 @@ protected:
    * @param iat the electron whose proposed move gets rejected.
    */
   void rejectMoveForwardMode(Index_t iat);
+};
+
+template<class T>
+struct ParticleSet_minimal_trait
+{
+  using tag = ParticleSet::MinimalDefault;
 };
 
 } // namespace qmcplusplus
--- a/src/Particle/ParticleSet.h
+++ b/src/Particle/ParticleSet.h
@@ -37,6 +37,7 @@ namespace qmcplusplus
 class DistanceTableData;
 class ResourceCollection;
 class StructFact;
+class CompositeSPOSet;
 
 /** Specialized paritlce class for atomistic simulations
  *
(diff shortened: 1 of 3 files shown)


ye-luo:
"Some/Most Estimators of ParticleSet arguments do not add their required distance tables they just assume they are there in the walker ParticleSet and of a usable implementation." Why don't they call addTable and obtain a valid DT?

src/Particle/tests/test_particle.cpp
```diff
@@ -27,6 +27,44 @@ using std::string;
...
+  outputManager.pause();
+
+  MinimalParticlePool mpp;
+  ParticleSetPool particle_pool = mpp(comm);
+  auto& pset_target = *(particle_pool.getParticleSet("e"));
+  auto& pset_source = *(particle_pool.getParticleSet("ion"));
+  int table_id = pset_target.addTable(pset_source, DTModes::ALL_OFF);
+  ParticleSet pset_minimal_default(pset_target, ParticleSet::MinimalDefault{});
+  // In the cass of ENABLE_OFFLOAD the whole DistanceTable creation ignores DTModes::ALL_OFF which
+  // isn't really a flag.
```
ye-luo, line 52: It is not ignoring DTModes::ALL_OFF but means optional behaviors are not requested from this add.

src/Particle/ParticleSet.h
```diff
@@ -37,6 +37,7 @@ namespace qmcplusplus
 class DistanceTableData;
 class ResourceCollection;
 class StructFact;
+class CompositeSPOSet;
```
ye-luo, line 40: Unnecessary?

ye-luo (changes requested):
Can I see the code which triggers the need of this change?

src/Particle/ParticleSet.h
PDoakORNL: Was needed before change to placate gcc

PDoakORNL:
It's a good question.
I have two good reasons for not doing that: 

1. You would then update that table(s) for every single move. They are only needed when the estimator evaluates. Specifically by calling the SPOSet which has a different implementation from the "global" one. 

2. the QMCDriverNew owns the EstimatorMangerNew which is created in its constructor.  Right now the EstimatorMangerNew also serves as EstimatorManagerInput and constructs the Estimators. The QMCDriver should not be changing the golden particle set. It's construction should be complete by the time the driver is constructed. Is it later going to undo those changes? Is the global ParticleSet going to drag around whatever cruft each QMCSection stuffs into it?

3. putting off the redesign of the WFC and SPOSet evaluation API to express their actual data dependence in the API's and the removal of strong coupling between ParticleSet and QMCWaveFunction.

src/Particle/tests/test_particle.cpp
PDoakORNL: I'll remove my whining, I think I added enough docs elsewhere that it's no longer a surprise.

PDoakORNL:
As it turns out this is a problem of implicit resource aquisition and can be solved by other means.

PR #3604, opened 2021-11-11
Fix seg fault in batched optimizer

markdewing:
When 'nonlocalpp' is specified in the optimization section of the input file, but the Hamiltonian doesn't use pseudopotentials, `nlpp` will be null and the access will cause a segfault.

Ignoring the value in the input file (by using a null pointer test) is the same behavior as the non-batched optimizer.

The input files produced by Nexus enable `nonlocalpp` by default.

Also enable output earlier so messages about the presence or absence of the non-local Hamiltonian elements get printed.

- Bugfix

- No

desktop

## Checklist


ye-luo:
Test this please

ye-luo (commented):
LGTM. need non-ANL approval.

prckent (approved):
Good catch.

PR #3605, opened 2021-11-11
If ENABLE_OFFLOAD=1 particle set move requires mw resources

PDoakORNL:
## Proposed changes
Since ParticleSet doesn't need to support the legacy t-move code single walker API calls can rely on multiwalker resources. That means estimators must be in a scope where the crowd walker element reference vectors leaders hold resources.

- Bugfix

- No

Leconte
## Checklist

### DIFF
--- a/src/QMCDrivers/VMC/VMCBatched.cpp
+++ b/src/QMCDrivers/VMC/VMCBatched.cpp
@@ -39,7 +39,8 @@ void VMCBatched::advanceWalkers(const StateForThread& sft,
                                 Crowd& crowd,
                                 QMCDriverNew::DriverTimers& timers,
                                 ContextForSteps& step_context,
-                                bool recompute)
+                                bool recompute,
+                                bool accumulate_this_step)
 {
   if (crowd.size() == 0)
     return;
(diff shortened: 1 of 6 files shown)


src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -221,13 +224,20 @@ void VMCBatched::runVMCStep(int crowd_id,
   crowd.setRNGForHamiltonian(rng);
 
   int max_steps = sft.qmcdrv_input.get_max_steps();
-  // \todo delete
-  RealType cnorm = 1.0 / static_cast<RealType>(crowd.size());
   IndexType step = sft.step;
   // Are we entering the the last step of a block to recompute at?
   bool recompute_this_step = (sft.is_recomputing_block && (step + 1) == max_steps);
   advanceWalkers(sft, crowd, timers, *context_for_steps[crowd_id], recompute_this_step);
```
ye-luo, line 230: I saw accumulate_this_step a very useful flag in DMC advanced walkers. It can be controlled not only for warm up but also for frequency. I mean accumulate every N steps. So could you align VMC driver with DMC driver and also move "crowd.accumulate(rng);" into the end of collectables_timer scope of VMCBatched::advanceWalkers. Also make sure warmupsteps doesn't collect data.

ye-luo (changes requested):
Much more coherent in resource scopes.

src/QMCDrivers/VMC/VMCBatched.cpp
PDoakORNL: Done

PDoakORNL:
We don't get to escape scalar.dat that easily. The fix is obviously necessary and the PR still does what it has to.

ye-luo (dismissed):

ye-luo (approved):

ye-luo:
Start testing in-house

PR #3613, opened 2021-11-16
Split DistanceTableData class

ye-luo:
## Proposed changes
DistanceTableData is spited in this PR.
All the control APIs are moved into DistanceTable for being managed by ParticleSet.
DistanceTableAA and DistanceTableAB are derived from DistanceTable.
Data and access APIs are implemented in DistanceTableAA/AB. Access control can be AA/AB type specific.
All the DistanceTable consumers know the actual AA/AB type. 

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/DistanceTable.h
+++ b/src/Particle/DistanceTable.h
@@ -229,13 +229,11 @@ public:
   virtual void createResource(ResourceCollection& collection) const {}
 
   /// acquire a shared resource from a collection
-  virtual void acquireResource(ResourceCollection& collection,
-                               const RefVectorWithLeader<DistanceTable>& dt_list) const
+  virtual void acquireResource(ResourceCollection& collection, const RefVectorWithLeader<DistanceTable>& dt_list) const
   {}
 
   /// return a shared resource to a collection
-  virtual void releaseResource(ResourceCollection& collection,
-                               const RefVectorWithLeader<DistanceTable>& dt_list) const
+  virtual void releaseResource(ResourceCollection& collection, const RefVectorWithLeader<DistanceTable>& dt_list) const
   {}
 };
 
@@ -272,9 +270,7 @@ protected:
 
 public:
   ///constructor using source and target ParticleSet
-  DistanceTableAA(const ParticleSet& target, DTModes modes)
-      : DistanceTable(target, target, modes)
-  {}
+  DistanceTableAA(const ParticleSet& target, DTModes modes) : DistanceTable(target, target, modes) {}
 
   /** return full table distances
    */
--- a/src/Particle/DistanceTable.h
+++ b/src/Particle/DistanceTable.h
@@ -31,10 +31,10 @@ class ResourceCollection;
 /** @ingroup nnlist
  * @brief Abstract class to manage pair data between two ParticleSets.
  *
- * Each DistanceTableData object is fined by Source and Target of ParticleSet types.
+ * Each DistanceTable object is fined by Source and Target of ParticleSet types.
  *
  */
-class DistanceTableData
+class DistanceTable
 {
 public:
   static constexpr unsigned DIM = OHMMS_DIM;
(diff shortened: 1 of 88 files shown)


src/Particle/DistanceTable.h
```diff
@@ -31,10 +31,10 @@ class ResourceCollection;
 /** @ingroup nnlist
  * @brief Abstract class to manage pair data between two ParticleSets.
  *
- * Each DistanceTableData object is fined by Source and Target of ParticleSet types.
+ * Each DistanceTable object is fined by Source and Target of ParticleSet types.
  *
  */
-class DistanceTableData
+class DistanceTable
```
PDoakORNL, line 37: I think it would be better if this was just an interface i.e. a pure virtual base class. Push the data members into the derived classes as well as all the accessors, which I believe are always called with knowledge of the type.

src/Particle/createDistanceTableAB.cpp
```diff
@@ -25,14 +25,14 @@ namespace qmcplusplus
  *\param s source/target particle set
  *\return index of the distance table with the name
  */
-DistanceTableData* createDistanceTableAB(const ParticleSet& s, ParticleSet& t, std::ostream& description)
+DistanceTable* createDistanceTableAB(const ParticleSet& s, ParticleSet& t, std::ostream& description)
```
PDoakORNL, line 28: why not a smart pointer?

These need to throw an exception if they fail, when it happens it will frequently be in another classes a constructor.

PDoakORNL (changes requested):
Overall this a good step toward better type safety and clarity for distance tables.  For this PR I'd like to see basic handling of failure to create DistanceTables.  Most of my other comments are discussion of future design.

src/Particle/DistanceTable.h
ye-luo: Isn't this exactly what I did in this PR? I added more comments above to make it clear ;)

src/Particle/ParticleSet.cpp
```diff
@@ -388,6 +388,15 @@ int ParticleSet::addTable(const ParticleSet& psrc, DTModes modes)
   return tid;
 }
 
+const DistanceTableAA& ParticleSet::getDistTableAA(int table_ID) const
+{
+  return dynamic_cast<DistanceTableAA&>(*DistTables[table_ID]);
```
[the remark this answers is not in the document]
ye-luo: Storage has been moved from DistanceTable to DistanceTableAA/AB. The type erased base class is used for ParticleSet to dispatch computation. The data access can only be done via AA/AB derived classes.

Regarding table_ID, I still like it for being safe with clone and also easy to table look up. Maybe I didn't fully you mean regarding table_ID.

src/Particle/createDistanceTableAB.cpp
ye-luo: I also caught this but consider this a follow-up topic as more places needs to be touched.

src/Particle/DistanceTable.h
```diff
@@ -31,10 +31,10 @@ class ResourceCollection;
 /** @ingroup nnlist
  * @brief Abstract class to manage pair data between two ParticleSets.
  *
- * Each DistanceTableData object is fined by Source and Target of ParticleSet types.
+ * Each DistanceTable object is fined by Source and Target of ParticleSet types.
```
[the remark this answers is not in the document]
ye-luo: fixed

src/Particle/createDistanceTableAB.cpp
ye-luo: Added #3616 to remember it.

src/Particle/ParticleSet.cpp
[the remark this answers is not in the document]
PDoakORNL: I think the table_ID use is outside the scope here.  Thanks for the storage move.

src/Particle/DistanceTable.h
PDoakORNL: That's what I thought, I just wanted to push it the rest of the way.

PDoakORNL (approved):

PR #3615, opened 2021-11-16
Refactor memory locality handling out of estimators copy constructors

PDoakORNL:
## Proposed changes

In preparation for integration of OneBodyDensityMatrices this PR cleans up complication of OperatorEstBase derived Estimators copy constructors.  All the non copying logic is moved to the clearly named spawnCrowdClone method. The estimator data is no longer a unique_ptr to vector created on the heap but just a vector. Minor renaming is included.
 
- Documentation changes

- No

Leconte

## Checklist

### DIFF
--- a/src/Estimators/MomentumDistribution.h
+++ b/src/Estimators/MomentumDistribution.h
@@ -77,13 +77,15 @@ public:
 
   //MomentumDistribution(const MomentumDistribution& md);
 
+  MomentumDistribution(const MomentumDistribution& md, DataLocality dl);
+  
   /** This allows us to allocate the necessary data for the DataLocality::queue 
    */
   void startBlock(int steps) override;
 
   /** standard interface
    */
-  std::unique_ptr<OperatorEstBase> clone() const override;
+  std::unique_ptr<OperatorEstBase> spawnCrowdClone() const override;
 
   /** accumulate 1 or more walkers of MomentumDistribution samples
    */
--- a/src/Estimators/MomentumDistribution.h
+++ b/src/Estimators/MomentumDistribution.h
@@ -5,6 +5,7 @@
 // Copyright (c) 2021 QMCPACK developers.
 //
 // File developed by: Jaron T. Krogel, krogeljt@ornl.gov, Oak Ridge National Laboratory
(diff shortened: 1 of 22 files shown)


src/Estimators/MomentumDistribution.h
```diff
@@ -77,13 +77,15 @@ class MomentumDistribution : public OperatorEstBase
 
   //MomentumDistribution(const MomentumDistribution& md);
```
ye-luo, line 78: is it better if put this one = default and under private to prevent unintended use from outside?

src/Estimators/OperatorEstBase.h
```diff
@@ -131,7 +136,7 @@ class OperatorEstBase
    *  And it make's datalocality fairly easy but
    *  more descriptive and safe data structures would be better
    */
-  static Data createLocalData(size_t size, DataLocality data_locality);
+  Data createLocalData(size_t size);
```
ye-luo, line 139: is this function still needed? Didn't see a call.

src/Estimators/SpinDensityNew.h
```diff
@@ -54,15 +54,18 @@ class SpinDensityNew : public OperatorEstBase
                  const Lattice&,
                  const SpeciesSet& species,
                  const DataLocality dl = DataLocality::crowd);
-  SpinDensityNew(const SpinDensityNew& sdn);
 
+  SpinDensityNew(const SpinDensityNew& sdn) = default;
```
ye-luo, line 58: Same argument as above. Should we put it under private?

ye-luo (changes requested):
I like the improved clarity. Only a few minor comments to be addressed.

src/Estimators/MomentumDistribution.h
PDoakORNL: Probably

src/Estimators/OperatorEstBase.h
PDoakORNL: No it should be deleted this is an error in branch slicing.

ye-luo (approved):

ye-luo:
Test this please

Issue #3490, opened 2021-09-29
GPAW to QMCPACK converter


tiihonej:
I'm not really requesting as much as offering to write a converter from GPAW orbitals to QMCPACK. I'm anticipating some usability benefits (e.g. native and versatile python bindings) and potential new users from this capability. However, it would really help me out to get some answers from the experts instead of digging them up from the source codes.

Let me start with the following topics:

1. GPAW capabilities

GPAW currently runs with three alternative representations of the wavefunction: real-space grid, plane-waves and LCAO. As far as I know, each of these can be converted for QMCPACK, albeit quite differently. The priority will be the plane-wave representation, and the rest can be left for later.

GPAW also implements at least the following Hamiltonians: all-electron, projector-augmented waves (PAW), and a limited support for norm-conserving KB pseudopotentials in the UPF format. The priority is AE, because it can be readily implemented and tested. The PAW method won't do. The UPF route is important, but a more general support for correlation-consistent ECPs will need consideration by the GPAW developers. 

The GPAW release notes indicate support for simulation of (non-)collinear spin polarizations and spin-orbit coupling. I'll save these for later and start with spin-unpolarized calculations. 

Does this sound correct and feasible?

2. The converter

GPAW is written in Python, so it seems natural to write the converter also in Python (in contrast to the present Fortran or C++ converters). The converter could take a GPAW restart file (.gpw) as an input, and if everything checks out, produce an output ES-HDF using h5py and a template XML input. All the modes outlined above could be considered by the same script.

A feature like that could ship with the GPAW package, if they are willing, and be invoked with a single command. Alternatively, it could be a part of Nexus or a standalone executable Python script.

3. About ES-HDF

I've been reverse engineering to understand the soul of an appropriate ES-HDF file for the plane-wave representation. Feel free to address the following propositions and questions:

- There is no supported representation for a real-space mesh, but the appropriate way (e.g. from RMG) is conversion to PWs, according to ES-HDF, which QMCPACK will convert to bsplines in the runtime.
- The plane-wave bases are expressed as integer Miller indices, because they take the least storage space.
- The are several complex arrays associated with individual PW bases, which represent the electronic densities and a few lowest wave functions for each spin species. Are there implicit or explicit expectations of normalization within each individual quantity, or consistency between them? Namely, is the electronic density basically the ground state wavefunction squared, or is it more complicated than that? It somewhat puzzles me that (at least from QE) the densities and wavefunctions have differently sized PW bases.
[...]

PR #3617, opened 2021-11-18
ESHDF python class and GPAW converter

tiihonej:
## Proposed changes

I wrote and added a generic python class to handle ESHDF files for QMCPACK. The class is presently used by a new GPAW converter, which implements tools to either read a GPAW restart file (.gpw) or convert live calculator objects to ESHDF.

The present layout of the code allows the same Python class framework to be used in any representation of the wavefunction and bind to any mean-field code, which currently implements conversion to QMCPACK. This way the maintenance of the ESHDF specification, and also (in principle) most of the existing converters could be centralized on a user-friendly python platform. I'm not proposing to converting each of the present converters to python, but this remains a possible long-term strategy.

The present implementation is work in progress. (Actually, some of the intended usage of GPAW relies on updates that are still underway.) But generally speaking, the wavefunction conversion from GPAW works fine (see #3490) and the remaining big questions consider refactoring, writeup and testing of the code. I'll write some units tests soon, and I appreciate any ideas on how to do it. Also, I need your opinions on the general layout and some reassurance that the project is facing the right direction. What do you think? @prckent ?

Another related but isolated discussion topic will be GPAW integration in Nexus. I'll work on that with @jtkrogel, but generally with the converter framework in place this should not differ much from how PySCF is presently used.

- New feature

- No

-RHEL 8.4 laptop

## Checklist

- Not yet. This PR adds tests to cover any new code, or to catch a bug that is being fixed


PDoakORNL (changes requested):
Copyright needs to be correct.
Unit tests would help show the input expected by match_gv and desymmetrize_z as well as being good form.

src/QMCTools/Eshdf.py
```diff
@@ -0,0 +1,248 @@
+######################################################################################
+## This file is distributed under the University of Illinois/NCSA Open Source License.
```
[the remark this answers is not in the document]
tiihonej: I'm not 100% what the convention is. I'll change in my own name and 2021. Is the phrase 'This file is distributed ...' okay or not?

src/QMCTools/gpaw4qmcpack
```diff
@@ -0,0 +1,67 @@
...
+
+  The GPAW restart file (INFILE) must be created with mode='all' to include 
+  the wave function information.
+  
+''')
+    exit()
+#end def
+
+# parse command lines arguments
+def parse_args(args):
```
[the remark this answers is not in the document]
tiihonej: Good idea. I'll migrate to argparse

src/QMCTools/Eshdf.py
```diff
@@ -0,0 +1,248 @@
...
+        #TODO: various tests
+        return True
+    #end def
+
+    # standard write: outfile
+    #   default: write to HDF5
+    def write(self, outfile = None):
+        if not self.loaded:
+            print('Must load data first')
+            return
```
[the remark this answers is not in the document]
tiihonej: yep, let's make it a raise

tiihonej:
Thank you @PDoakORNL . I'll make the changes. I assume the best way to make unit tests is to write, for example, tests/converter/gpaw4qmcpack_test.py, which tests the core features and methods, and refer to in in the local CMakeLists.txt. Sounds good?

src/QMCTools/Eshdf.py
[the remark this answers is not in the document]
PDoakORNL: That seems to be how we do it.  The lawyers have seen this wording.

PDoakORNL (dismissed):

PDoakORNL:
Sorry for the slow turn around on approving the change.  I notice this is still [WIP] other than bringing it up to date with develop are there remaining issues to sort out?

tiihonej:
No worries. The WIP status was intentional as long as the testing side is missing. The latest commits add a test, where I built upon the earlier converter tests and which I tried to make as comprehensive but modestly sized as possible. I also added conditions to find GPAW and ASE before the executable or the tests are added.

There's room for testing other functionality and implementing features for the converter, but I think this one should catch most of the breaking changes and is now ready to be added to main repository as an alpha version.

src/QMCTools/Eshdf.py
```diff
@@ -0,0 +1,246 @@
+######################################################################################
+## This file is distributed under the University of Illinois/NCSA Open Source License.
+## See LICENSE file in top directory for details.
+##
+## Copyright (c) 2021 Juha Tiihonen and QMCPACK developers.
```
PDoakORNL, line 5: I was a bit unclear here, don't add your name to the copyright.  

[license in header](https://qmcpack.readthedocs.io/en/develop/developing.html#files)

I believe this will force us to hunt you down specifically for any future license change. Our contribution guidelines should explicitly require you to assign copyright to the project.

PDoakORNL (changes requested):
This looks solid.  The copyright statement should be updated.

src/QMCTools/Eshdf.py
tiihonej: Okay, seems I misunderstood you at first. I don't personally care about these so much, so let me remove the name.

PDoakORNL (approved):

src/QMCTools/CMakeLists.txt
```diff
@@ -51,6 +51,10 @@ if(USE_OBJECT_TARGET)
   target_link_libraries(qmcfinitesize qmcparticle qmcutil)
 endif()
 
+if (HAVE_GPAW)
+  configure_file(gpaw4qmcpack ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/gpaw4qmcpack COPYONLY)
```
ye-luo, line 55: This seems to be an action at install stage.

src/QMCTools/gpaw4qmcpack
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
```
ye-luo, line 1: Add ".py" surffix.

Please add a README.md file at  `src/QMCTools` with documentation about the relation of python files added in this folder by the current PR.

src/QMCTools/Eshdf_gpaw.py
```diff
@@ -0,0 +1,281 @@
...
+## See LICENSE file in top directory for details.
+##
+## Copyright (c) 2021 QMCPACK developers.
+##
+## File developed by: Juha Tiihonen, tiihonen@iki.fi, University of Jyvaskyla
+##
+## File created by: Juha Tiihonen, tiihonen@iki.fi, University of Jyvaskyla 
+#######################################################################################
+
+from numpy import array,pi,dot,real,imag,concatenate,unique,zeros
```
ye-luo, line 12: both Eshdf_gpaw and gpaw4qmcpack need a brief documentation at the beginning about what they do. Probably just need one line.

tests/gpaw/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+# add tests for GPAW4QMCPACK converter
```
ye-luo, line 1: is `tests/gpaw` intended to test the converter or workflow?
if converter, move under `tests/converters/gpaw`
if workfload, rename as `gpaw4qmcpack`

tests/gpaw/test_Si_diamond/run_si.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
```
ye-luo, line 1: What is the above `restart.gpw` file archived for?
If it is not needed for the test, better remove.

tests/gpaw/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+function(ADD_GPAW4QMCPACK_TEST test_name)
+      set(EXE_NAME ${qmcpack_BINARY_DIR}/bin/gpaw4qmcpack)
+  execute_process(COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/gpaw4qmcpack_test.py"
```
ye-luo, line 4: Why does this need to happen at every ADD_GPAW4QMCPACK_TEST?

tests/gpaw/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+                          "${CMAKE_CURRENT_BINARY_DIR}")
+  execute_process(COMMAND ${CMAKE_COMMAND} -E copy_directory "${CMAKE_CURRENT_SOURCE_DIR}/${test_name}"
```
ye-luo, line 6: Better to create the destination folder softlink necessary files instead of copy.

ye-luo (changes requested):
I added many comments but I may misunderstand certain places.

src/QMCTools/CMakeLists.txt
tiihonej: It is. Is there a more suitable way to copy the file over with an intent to install it along with the compiled binaries? The present way does it prematurely, but I'm not an expert on CMake. Also, I think it's debatable whether the python scripts like this should be installed to bin/ or not. I find it convenient to only have one path to add, but for instance PyscfToQmcpack.py doesn't get copied over.

tests/gpaw/test_Si_diamond/run_si.py
tiihonej: `restart.gpw` is the most important file for the test, as it contains the orbitals to be converted in the GPAW format. The reference script `run_si.py` is unimportant (but also small). Should I remove it? I notice there's an inconsistency with the file name. The script has been pointing to `Si.gpw`, which has been renamed to `restart.gpw`.

tests/gpaw/CMakeLists.txt
tiihonej: You're right, it probably doesn't. It's an artifact from a simpler test case for the test. I'll move it outside of the function.

tests/gpaw/CMakeLists.txt
tiihonej: You're probably right. I just copied this from the existing converter tests in `tests/converter/.` Let me look into it.

tiihonej: So let me make sure: first create an actual destination directory, then soft link the necessary contents? If I soft link the whole directory it will clutter the source code upon testing, right?

tests/gpaw/CMakeLists.txt
tiihonej: The present test is only for the converter, not workflow, and no running with GPAW. I wasn't sure where to put it. Your proposition sounds good, I'll make the change.

tiihonej:
Thanks for the comments, I think they're all on point. I'll make the necessary changes early next year.

src/QMCTools/gpaw4qmcpack
tiihonej: Good. I'll write something. Maybe the Pyscf methods should also be covered in the same README, but I don't know too much about them. I find it a little bit confusing that there are .cpp and .py under the same directory with very different installation mechanisms and documentation, but that's a broader question and probably not so important until we have many more python tools.

src/QMCTools/CMakeLists.txt
ye-luo: I removed the `if`. Making softlink or installing the file doesn't need gpaw, users may install gpaw later.

I changed your `configure_file` to `file(CREATE_LINK)` you can use this an example for other requested changes.

`PyscfToQmcpack.py` doesn't have a main, it is more of a library instead of a standalone tool. We may need to think about how to properly install python files.

src/QMCTools/gpaw4qmcpack
ye-luo: You are only responsible for what you added.

tests/CMakeLists.txt
```diff
@@ -68,6 +68,17 @@ if(HAVE_PYSCF)
     endif()
   endif()
 endif()
+
+if(HAVE_GPAW)
+    include("${qmcpack_SOURCE_DIR}/CMake/python.cmake")
```
ye-luo, line 73: If the converter doesn't invoke anything from GPAW. All the HAVE_GPAW should be removed.
Please move the python checks into `gpaw/CMakeLists.txt`

tests/gpaw/CMakeLists.txt
ye-luo: Unfortuantely, we can not softlink the folder. Multiple build may share the same source folder.

src/QMCTools/CMakeLists.txt
ye-luo: I need a bit more thinking about the necessity of adding files to `build/bin` leave the problem now. Please make sure the converter test doesn't refer to `gpaw4qmcpack` under `build/bin` but directly to the original file.

tests/gpaw/CMakeLists.txt
tiihonej: I now create a new folder and soft link the files needed for testing

src/QMCTools/gpaw4qmcpack
tiihonej: I wrote a bit of README documentation for the latest additions.

src/QMCTools/CMakeLists.txt
tiihonej: The converter test now refers to the original file under `src/QMCTools`. It seems that the symbolic link is still copied to `build/bin` before installation stage. Is this an issue? We can also choose to not install the script to `build/bin`, if that's cleaner

tests/CMakeLists.txt
tiihonej: This is still open. The converter needs GPAW, and the tests shouldn't be added without it. I've moved the clause down to `tests/converter`, where it controls whether or not to add subdirectory `gpaw`. Also since we're calling the converter in the original directory, there is no need to require `Eshdf` and `Eshdf_gpaw` classes. Is this okay?

CMakeLists.txt
```diff
@@ -1024,6 +1024,18 @@ if(NOT QMC_NO_SLOW_CUSTOM_TESTING_COMMANDS)
   endif()
 endif()
 
+#-------------------------------------------------------------------
+# Check if GPAW is available
+#-------------------------------------------------------------------
+if(NOT QMC_NO_SLOW_CUSTOM_TESTING_COMMANDS)
+  test_python_module(gpaw HAVE_GPAW)
```
ye-luo, line 1031: Please update this to check_python_reqs as pyscf.

ye-luo (commented):
@tiihonej this is my last request. Once you update this, I will approve the PR.

src/QMCTools/CMakeLists.txt
```diff
@@ -53,11 +53,10 @@ if(USE_OBJECT_TARGET)
   target_link_libraries(qmcfinitesize qmcparticle qmcutil)
 endif()
 
-file(CREATE_LINK ${CMAKE_CURRENT_SOURCE_DIR}/gpaw4qmcpack.py ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/gpaw4qmcpack.py SYMBOLIC)
-install(FILES gpaw4qmcpack.py DESTINATION bin)
```
ye-luo:
```
file()
install()
```
These two lines should stay

tiihonej: Okay, wasn't sure and took a wrong guess. It's amended now.

ye-luo (approved):
LGTM (I mostly looked at the CMake part)

tiihonej:
The testing seems stale. I'm closing this PR and making a new one.

PR #3620, opened 2021-11-18
Split BackflowTransformation into h/cpp

ye-luo:
## Proposed changes
Reduce header inclusion entanglement.

- No

epyc-server

## Checklist


prckent (approved):
LGTM. I think we still have a few more of these heavy .h around.

Issue #3619, opened 2021-11-18
Wavefunction state serialization capability needed


markdewing:
A number of wavefunction parameters do not get saved during optimization (CI coeff, orbital rotations)

The current scheme updates XML nodes in `QMCCostFunctionBase::updateXmlNodes`.  That function needs to get updated for every new parameter.

See discussions in #444 and #2068

We want a general solution to saving variational parameters.

**Describe the solution you'd like**

The general solution is the ability to dump the wavefunction, including the parameters and other structure.
Desirable features:
* complete state and structure so the wavefunction can be re-used easily
  * As mentioned in issue 444, keeping the CI coefficients and Jastrow factors together is desirable.  The easy and common path should be to reuse all the variational parameters as a set.
* Both XML and HDF formats
  * XML is easy to read and modify for small cases
  * HDF is necessary for scaling to large numbers of parameters
* Implementation should be distributed to classes rather than centralized as it is now

**Describe alternatives you've considered**
***Generic object serialization***
In principle, object serialization is what we want.  Something like boost::serialization would do this.  However, the format is opaque.  I think we want to re-use the input file format for uniformity.

***Saving the list of variational parameters***
One easy approach would be to store just the variational parameters in a file.   This would require two files to reload the state: the input file with the wavefunction structural information plus the file with the updated variational parameters. For the common use case, this shouldn't be too confusing.  It would get a little trickier when modifying a wavefunction - the WF input file and VP file would need updated in sync.

This approach would be the fastest to implement.    The more I think about it, the more appealing this solution looks (at least as an interim solution).  

The code changes would be limited in scope. The list of variational parameters is centralized to the `OptVariables` list in the optimizer.  This would effectively be serializing this structure to/from disk.

***Saving the entire wavefunction***
The original design had the ideal of each class reading from XML (the `put` methods).  While appealing in theory, it has led to problems.  XML handling gets mixed into every class.  And the `put` method becomes the de-facto constructor which makes unit testing difficult.
The new approach is to separate the XML input into a separate class (the various `Input` classes). 

In order to make the input and output more agnostic towards XML and HDF, we need a more independent (declarative) description of the input data.  The `Input` classes seem like a natural place to add this data model, and to perform output.  For example, the `OhmmsAttbuteSet` and `ParameterSet` classes would be fairly straightforward to extend to writing XML, and to handling HDF.  

This approach requires defining the new model and implementing in all the needed classes before it can be used, so it will be the slowest to implement.

[...]

PR #3625, opened 2021-11-22
Save and restore variational parameters to XML

markdewing:
The first step towards a solution to save and restore all variational parameters (VP) as discussed in #3619

Reading from a VP file is not intended to create the list of variational parameters.  The usage model is that the variational parameters are set up as usual in the code, and then new values for those parameters are read in from the file.

This enables a workflow where some subset of parameters is optimized and then the wavefunction is enlarged (or the optimization of more parameters enabled) and further optimization is performed. Values that are not present in the VP file are not changed. Reading the VP file acts like an overlay of its values on the existing values read in from the wavefunction section.

The the list of values is contained under a "name_value_pairs" tag.  In the future one might want to store values differently, and that could be done under a different tag. For example, coefficients are often stored as Coeff_0, Coeff_1, etc. for a large number of values. It would take less space to store the prefix once and store all the values in an array.

XML is being used to get this capability started.  Eventually it will use HDF.

The XML file looks like this (from the unit test):
```
<?xml version="1.0"?>
<variational_parameters>
  <name_value_pairs>
    <vp>
      <name>s</name>
      <value>11234.6</value>
    </vp>
    <vp>
      <name>second</name>
      <value>0.000256789</value>
    </vp>
    <vp>
      <name>really_long_name</name>
      <value>-1.2</value>
    </vp>
  </name_value_pairs>
</variational_parameters>
```

- New feature

- No

laptop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/VariableSet.cpp
+++ b/src/QMCWaveFunctions/VariableSet.cpp
@@ -18,8 +18,11 @@
 #include <iomanip>
 #include <ios>
 #include <algorithm>
+#include "io/OhmmsData/Libxml2Doc.h"
 
 using std::setw;
+
+using qmcplusplus::app_log;
 
 namespace optimize
 {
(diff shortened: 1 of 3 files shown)


prckent:
Maybe worth noting that other codes are already doing 10K parameter optimization, so the solution needs to be scalable. i.e. HDF5 as the preferred mode. But XML is certainly more human readable and possibly easier to debug.

ye-luo (changes requested):
I think it will be better if we can first sort out how we would like the XML structure to look like. Could you arrange a discussion?

markdewing:
Can we have the discussion here?  Or do want it to have a separate issue?

ye-luo:
I was thinking of arranging a call. I feel there are more than this PR or #3619 to discuss before actually implementing anything. More generally, one key task is to define what VariableSet is. How to handle individual parameters, how to handle a set of related parameters. "How to handle 1k parameters I/O?" question starts from how VariableSet handles parameters. `coeff_0, coeff1, ...., coeff1k` is already a problem.

prckent:
If there needs to be much discussion, perhaps start with HDF5? Minimum viable product etc. This should be super simple - just dump a single array. I don't think we even need to be concerned with sets of parameters at this point, and I don't see many people using it initially. It will be immediately useful to have this capability, so nice to haves shouldn't delay us.

ye-luo:
> If there needs to be much discussion, perhaps start with HDF5? Minimum viable product etc. This should be super simple - just dump a single array. I don't think we even need to be concerned with sets of parameters at this point, and I don't see many people using parameter sets initially in any case. It will be immediately useful to have this capability, so nice to haves shouldn't delay us.

That looks like simple to implement and immediately useful. Only need dataset "all_parameters".

markdewing:
Regarding VariableSet - are you concerned about its scalability in memory?  It does store the string name for every parameter and uses a std::map to store the string name -> value mapping.  
Or do you have other concerns about the current VariableSet implementation?

ye-luo:
> Regarding VariableSet - are you concerned about its scalability in memory? It does store the string name for every parameter and uses a std::map to store the string name -> value mapping. Or do you have other concerns about the current VariableSet implementation?

`std::vector<pair_type> NameAndValue;` each pair contains a string and a value. Memory issue is real. In addition, there is a local VariableSet and global VariableSet, then 2x memory again. Third, this way of storing parameter, doing any computation needs to copy in and out to a array for numerical operation.

markdewing:
Will probably use only HDF for simplicity.

PR #3627, opened 2021-11-23
Split and rename scalar_traits.h

### LINKED ISSUE #3378: Clean _traits classes
There are many `_traits` classes in QMCPACK and while they attempt to make getting dependent type information and dependent relationships of type from template parameters or the application level ifdef'd type aliases they've become stale.

Go through these classes 1 by 1, replace with new code capable of taking over the necessary work.

First target:
**scalar_traits.h should be removed.**
scalar_traits all of its functionality is redundant and most clashes with normal c++ idiom.
Familiar
`std::complex<T>::value_type = T;`
surprising
```c++
scalar_traits<std::complex<T>::value_type == std::complex<T>
scalar_traits<std::complex<T>::real_type == T
```

Type conversion via return arguments with many specialized versions of 
```
template<typename T1, typename T2>
void convert(const T1& in, T2& out);
```
These are inconsistent in their results.
- They do static_casts and assignment
- For complex and floating point types it calls .real() on the complex argument and assigns to the fp out.
- They provides a redundant version of the type templated TinyVector operator= 
- They provide a redundant version of OhmmsPete containers operator=
- They provide a dubious assignment between OhmmsPETE objects of unequal size with no protection against
overruns and semantics for Matrices and Tensors that are unexpected i.e. linearly into the raw data instead of into some corner of the object.

the scalar_traits themselves mostly serve to provide support to container_proxy which provides a seemingly unecessary way of wrapping Ohmms and STL objects containing a complex<T> type as a blob of T of twice the size.  

They are also used as a way to get Real type underlying a Value type.
```c++
typename scalar_traits<Value>::real_type my_real
```

Replace use with `complex_help.hpp` used in testing and batched code.
```c++
RealAlias<Value> my_real
```
`complex_help.hpp` also provides a IsComplex<T> and IsReal<T> appropriate for `if constexpr` and `std::enable_if_t` use.

Using these classes in new code instead of cleaner alternatives.

**Additional context**

ye-luo:
## Proposed changes
Close #3378
1. Move scalar_traits class to container_proxy.h
2. Rename all the convert() to convert2real to reduce confusion. We have other functions called convert and do other things.
3. Thus rename file scalar_traits.h as convert2real.h

- No

epyc-server

## Checklist

### DIFF
--- a/src/AFQMC/HamiltonianOperations/KPTHCOps.hpp
+++ b/src/AFQMC/HamiltonianOperations/KPTHCOps.hpp
@@ -26,7 +26,7 @@
 #include "multi/array.hpp"
 #include "multi/array_ref.hpp"
 #include "AFQMC/Numerics/ma_operations.hpp"
-#include "type_traits/scalar_traits.h"
+#include "type_traits/complex_help.hpp"
 #include "AFQMC/Wavefunctions/Excitations.hpp"
 #include "AFQMC/Wavefunctions/phmsd_helpers.hpp"
 #include "AFQMC/Utilities/myTimer.h"
--- a/src/AFQMC/HamiltonianOperations/THCOps.hpp
+++ b/src/AFQMC/HamiltonianOperations/THCOps.hpp
@@ -24,7 +24,7 @@
 #include "Utilities/FairDivide.h"
 #include "AFQMC/Utilities/taskgroup.h"
 #include "mpi3/shared_communicator.hpp"
-#include "type_traits/scalar_traits.h"
+#include "type_traits/complex_help.hpp"
 #include "AFQMC/Wavefunctions/Excitations.hpp"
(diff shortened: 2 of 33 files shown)


ye-luo:
Test this please

PDoakORNL (changes requested):
This is some welcome cleanup and removal of legacy functionality since replaced. But t think we can use an extra character and have the function name conform to the coding style i.e.
`convert2real` -> `convertToReal`

ye-luo:
> This is some welcome cleanup and removal of legacy functionality since replaced. But t think we can use an extra character and have the function name conform to the coding style i.e. `convert2real` -> `convertToReal`

Good suggestion. updated.

ye-luo:
Start testing in-house

ye-luo:
Start testing in-house

ye-luo:
Should be all good now.

PDoakORNL (approved):

PR #3634, opened 2021-11-30
[WIP]Variational parameter checkpoint prototype

markdewing:
**This PR is not intended for merging!**
This PR is to give an overview and gather feedback on design issues.  Ultimately, individual parts will be submitted as separate PR's (and the code cleaned up there).

Also, this PR doesn't work correctly with multiple threads.  I suspect the VP's are not set to the new values everywhere.

Contents:
1. Save/load XML - adds version and timestamp to #3625.   The version is for versioning the file format.  The timestamp records when the file was written, just to provide more information in case it's needed.
2. Save/load HDF.  HDF naturally stores data as arrays, so this stores two arrays - one for the parameters, and one for the names. (This is different from the XML, which stores an array of name/value pairs)
3. Save the VP file as part of the optimization cycle in `QMCCostFunctionBase::reportParameters`.  The format is currently hardcoded to HDF.  Eventually there would be an input tag to switch between XML and HDF formats (with the default being HDF)
4. Restore VP parameters in the optimizer using "variational_parameter_file_to_load" tag. The file format is detected from the file name suffix.  This should work for both legacy and batched optimizers.   Since the optimizer already sets up the collection of optimization variables, there is a natural place to restore the values from the file.
5. Restore VP parameters in batched VMC.  (Does not work in legacy VMC driver).  This needs to set up the collection of optimization variables, since it is not done already.   Right now it is done in `VMCBatched::process`.  I don't know if there's a better place to put this functionality?

- New feature

- No

### DIFF
--- a/src/QMCDrivers/MCPopulation.cpp
+++ b/src/QMCDrivers/MCPopulation.cpp
@@ -283,9 +283,6 @@ void MCPopulation::set_variational_parameters(const opt_variables_type& active)
 {
   for (auto it_twfs = walker_trial_wavefunctions_.begin(); it_twfs != walker_trial_wavefunctions_.end(); ++it_twfs)
   {
-    opt_variables_type tmpVar;
-    (*it_twfs).get()->checkInVariables(tmpVar);
-    (*it_twfs).get()->checkOutVariables(active);
     (*it_twfs).get()->resetParameters(active);
   }
 }
--- a/src/QMCDrivers/MCPopulation.cpp
+++ b/src/QMCDrivers/MCPopulation.cpp
@@ -283,6 +283,9 @@ void MCPopulation::set_variational_parameters(const opt_variables_type& active)
 {
   for (auto it_twfs = walker_trial_wavefunctions_.begin(); it_twfs != walker_trial_wavefunctions_.end(); ++it_twfs)
   {
+    opt_variables_type tmpVar;
+    (*it_twfs).get()->checkInVariables(tmpVar);
+    (*it_twfs).get()->checkOutVariables(active);
     (*it_twfs).get()->resetParameters(active);
   }
 }
--- a/src/QMCDrivers/QMCDriverInput.h
+++ b/src/QMCDrivers/QMCDriverInput.h
@@ -93,6 +93,8 @@ protected:
   std::string drift_modifier_{"UNR"};
   RealType drift_modifier_unr_a_ = 1.0;
 
+  std::string variational_parameter_file_;
+
   /** @}
    */
 
@@ -129,6 +131,8 @@ public:
 
   const std::string get_drift_modifier() const { return drift_modifier_; }
   RealType get_drift_modifier_unr_a() const { return drift_modifier_unr_a_; }
+
+  const std::string get_variational_parameter_file() const { return variational_parameter_file_; }
 };
 
 // These will cause a compiler error if the implicit move constructor has been broken
--- a/src/QMCDrivers/QMCDriverInput.h
+++ b/src/QMCDrivers/QMCDriverInput.h
@@ -93,8 +93,6 @@ protected:
   std::string drift_modifier_{"UNR"};
   RealType drift_modifier_unr_a_ = 1.0;
 
-  std::string variational_parameter_file_;
-
   /** @}
    */
 
@@ -131,8 +129,6 @@ public:
 
   const std::string get_drift_modifier() const { return drift_modifier_; }
   RealType get_drift_modifier_unr_a() const { return drift_modifier_unr_a_; }
-
-  const std::string get_variational_parameter_file() const { return variational_parameter_file_; }
 };
 
 // These will cause a compiler error if the implicit move constructor has been broken
--- a/src/QMCDrivers/VMC/VMCBatched.cpp
+++ b/src/QMCDrivers/VMC/VMCBatched.cpp
@@ -247,6 +247,37 @@ void VMCBatched::process(xmlNodePtr node)
                                 qmcdriver_input_.get_walkers_per_rank(), 1.0, qmcdriver_input_.get_num_crowds());
 
     Base::startup(node, awc);
+
+
+    if (!qmcdriver_input_.get_variational_parameter_file().empty())
+    {
+      TrialWaveFunction& Psi = population_.get_golden_twf();
+      optimize::VariableSet tmpVar;
+      Psi.checkInVariables(tmpVar);
+      tmpVar.resetIndex();
+      Psi.checkOutVariables(tmpVar);
+
+      std::string vp_file = qmcdriver_input_.get_variational_parameter_file();
+      bool is_xml         = false;
+
+      if (vp_file.find(".xml") != std::string::npos)
+      {
+        is_xml = true;
+      }
+
+      if (is_xml)
+      {
+        tmpVar.readFromXML(vp_file);
+      }
+      else
+      {
+        tmpVar.readFromHDF(vp_file);
+      }
+      Psi.resetParameters(tmpVar);
+      population_.set_variational_parameters(tmpVar);
+      app_log() << "Variational parameters loaded from " << vp_file << std::endl;
+      tmpVar.print(app_log());
+    }
   }
   catch (const UniformCommunicateError& ue)
   {
--- a/src/QMCDrivers/VMC/VMCBatched.cpp
+++ b/src/QMCDrivers/VMC/VMCBatched.cpp
@@ -258,21 +258,8 @@ void VMCBatched::process(xmlNodePtr node)
       Psi.checkOutVariables(tmpVar);
 
       std::string vp_file = qmcdriver_input_.get_variational_parameter_file();
-      bool is_xml         = false;
+      tmpVar.readFromHDF(vp_file);
 
-      if (vp_file.find(".xml") != std::string::npos)
-      {
-        is_xml = true;
-      }
-
-      if (is_xml)
-      {
-        tmpVar.readFromXML(vp_file);
-      }
-      else
-      {
-        tmpVar.readFromHDF(vp_file);
-      }
       Psi.resetParameters(tmpVar);
       population_.set_variational_parameters(tmpVar);
       app_log() << "Variational parameters loaded from " << vp_file << std::endl;
--- a/src/QMCDrivers/VMC/VMCBatched.cpp
+++ b/src/QMCDrivers/VMC/VMCBatched.cpp
@@ -247,7 +247,6 @@ void VMCBatched::process(xmlNodePtr node)
                                 qmcdriver_input_.get_walkers_per_rank(), 1.0, qmcdriver_input_.get_num_crowds());
 
     Base::startup(node, awc);
-
   }
   catch (const UniformCommunicateError& ue)
   {
--- a/src/QMCDrivers/VMC/VMCBatched.cpp
+++ b/src/QMCDrivers/VMC/VMCBatched.cpp
@@ -248,23 +248,6 @@ void VMCBatched::process(xmlNodePtr node)
 
     Base::startup(node, awc);
 
-
-    if (!qmcdriver_input_.get_variational_parameter_file().empty())
-    {
-      TrialWaveFunction& Psi = population_.get_golden_twf();
-      optimize::VariableSet tmpVar;
-      Psi.checkInVariables(tmpVar);
-      tmpVar.resetIndex();
-      Psi.checkOutVariables(tmpVar);
-
-      std::string vp_file = qmcdriver_input_.get_variational_parameter_file();
-      tmpVar.readFromHDF(vp_file);
-
-      Psi.resetParameters(tmpVar);
-      population_.set_variational_parameters(tmpVar);
-      app_log() << "Variational parameters loaded from " << vp_file << std::endl;
-      tmpVar.print(app_log());
-    }
   }
   catch (const UniformCommunicateError& ue)
   {
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
@@ -242,18 +242,8 @@ void QMCCostFunctionBase::reportParameters()
 
 
     std::ostringstream vp_filename;
-    bool vp_xml_format = false;
-
-    if (vp_xml_format)
-    {
-      vp_filename << RootName << ".vp.xml";
-      OptVariables.saveAsXML(vp_filename.str());
-    }
-    else
-    {
-      vp_filename << RootName << ".vp.h5";
-      OptVariables.saveAsHDF(vp_filename.str());
-    }
+    vp_filename << RootName << ".vp.h5";
+    OptVariables.saveAsHDF(vp_filename.str());
   }
 }
 /** This function stores optimized CI coefficients in HDF5 
@@ -496,26 +486,9 @@ bool QMCCostFunctionBase::put(xmlNodePtr q)
   if (!variational_parameter_file.empty())
   {
     app_log() << "Loading variational parameters from " << variational_parameter_file << std::endl;
-    bool is_xml = false;
 
-    if (variational_parameter_file.find(".xml") != std::string::npos)
-    {
-      is_xml = true;
-    }
+    OptVariables.readFromHDF(variational_parameter_file);
 
-    bool okay = true;
-    if (is_xml)
-    {
-      okay = OptVariables.readFromXML(variational_parameter_file);
-    }
-    else
-    {
-      OptVariables.readFromHDF(variational_parameter_file);
-    }
-    if (!okay)
-    {
-      APP_ABORT("Failure reading variational parameter file");
-    }
   }
   //     app_log() << "<active-optimizables> " << std::endl;
   //     OptVariables.print(app_log());
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
@@ -330,7 +330,6 @@ bool QMCCostFunctionBase::put(xmlNodePtr q)
 {
   std::string writeXmlPerStep("no");
   std::string computeNLPPderiv("no");
-  std::string variational_parameter_file;
   ParameterSet m_param;
   m_param.add(writeXmlPerStep, "dumpXML");
   m_param.add(MinNumWalkers, "minwalkers");
@@ -341,7 +340,6 @@ bool QMCCostFunctionBase::put(xmlNodePtr q)
   m_param.add(GEVType, "GEVMethod");
   m_param.add(targetExcitedStr, "targetExcited");
   m_param.add(omega_shift, "omega");
-  m_param.add(variational_parameter_file, "variational_parameter_file_to_load");
   m_param.put(q);
 
   tolower(targetExcitedStr);
@@ -481,14 +479,6 @@ bool QMCCostFunctionBase::put(xmlNodePtr q)
   if (NumOptimizables == 0)
   {
     APP_ABORT("QMCCostFunctionBase::put No valid optimizable variables are found.");
-  }
-
-  if (!variational_parameter_file.empty())
-  {
-    app_log() << "Loading variational parameters from " << variational_parameter_file << std::endl;
-
-    OptVariables.readFromHDF(variational_parameter_file);
-
   }
   //     app_log() << "<active-optimizables> " << std::endl;
   //     OptVariables.print(app_log());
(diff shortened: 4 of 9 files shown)


jtkrogel:
These look like nice features.  What about just "parameter_file" for the tag?

markdewing:
There some discussion of the tag name in #3619.  I think the word "parameter" by itself is too vague and could be ambiguous,  so usually I like to qualify it with "variational".

PDoakORNL:
re: parameter, especially in the context of our input files it's quite vague.  I will get a complete review done today.

prckent:
Perhaps mention _wavefunction_ ?

PDoakORNL (changes requested):
Some superficial refactoring seems in order.

ye-luo:
I would favor variational parameters over wavefunction parameters.

ye-luo:
Do you store one string for each parameter or one string for each set of parameters in hdf5? Ideally, I'd like the hdf5 to contain
```
h5/
    J1/
       parameter_array
    J2/
       parameter_array
```
since we will probably start with a single fused array.
```
h5/
    name_array_by_set
    first_last_by_set
    all_parameter_array
```
What is the current layout you picked?

markdewing:
The format for the HDF file is one string per parameter.   Currently it is a flat array that doesn't have any of the wavefunction structure.  Retaining that information will require changes to the way parameters are stored in VariableSet. 

The name array and the value arrays are stored in the "name_value_lists" group.  This should allow some backwards compatibility if and when we change VariableSet to have a more compact representation.  The new representation can be stored into a new group.

An example dump (there's only the e-e Jastrow in this wavefunction)
```
HDF5 "he_opt3.s009.vp.h5" {
GROUP "/" {
   GROUP "name_value_lists" {
      DATASET "names" {
         DATATYPE  H5T_STRING {
            STRSIZE H5T_VARIABLE;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SIMPLE { ( 4 ) / ( 4 ) }
         DATA {
         (0): "jud_0", "jud_1", "jud_2", "jud_3"
         }
      }
      DATASET "values" {
         DATATYPE  H5T_IEEE_F64LE
         DATASPACE  SIMPLE { ( 4 ) / ( 4 ) }
         DATA {
         (0): 0.72053, 0.153233, -0.626879, -0.242434
         }
      }
   }
   DATASET "timestamp" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
      DATA {
      (0): "2021-12-02 10:48:05 CST"
      }
   }
   DATASET "version" {
      DATATYPE  H5T_STD_I64LE
      DATASPACE  SIMPLE { ( 3 ) / ( 3 ) }
      DATA {
      (0): 1, 0, 0
      }
   }
}
}
```

ye-luo:
What are the stored name used for? Do you have a use case?

markdewing:
The parameter names are used for setting the values.  The loader in VariableSet matches on the names to set the new values.

ye-luo:
> The parameter names are used for setting the values. The loader in VariableSet matches on the names to set the new values.

Does the all-in-one VariableSet needs to take each parameter by looking up a map? It cannot just dump the whole array in?

markdewing:
The code to insert the parameters is this:
```
         if (find(vp_name) != end())
            (*this)[vp_name] = vp_value;
```
Note the `find` call on VariableSet is a linear lookup.

This isn't a complete serialization of VariableSet - ParameterType and Recompute information is not saved.
It seems more robust to match up by names.  That way we don't have to worry if the ordering changes.

ye-luo:
At the point of loading or dumping parameter. All the parameters should have been checked in the all-in-one VariableSet. All the look-up should succeed and there is no insert but overwriting only. Is this not the case?

markdewing:
Yes, on load, it should only overwrite the value of an existing parameter.
The reason for the `find` is that the `[]` operator will insert the value if it doesn't exist, and we don't want that behavior.
We might want to print a warning if the value isn't found (it may indicate a mismatch between the VP file and the wavefunction)

prckent:
It think we should support "rogue" use input parameters that don't exactly match but print plenty of warnings if so. From the discussion above this seems to be the case. There are some circumstances where this might be useful. e.g. Changing one part of the jastrow, adding a piece etc.

By default it would be good to determine and print if all the loaded parameters correspond exactly to the current wavefunction parameters (no more, no fewer). Not needed in first implementation!

ye-luo:
I don't think current logic of partial updating parameters is welll defined and requires updating VariableSet to store parameter names by sets. As a first implementation, I prefer not storing parameter names on file and only support the number parameters exactly matched.

prckent:
An example of why partial matching is useful :  adding a three body jastrow  when 1+2 body are available and saved from elsewhere.

I think this is for Mark to decide on the preferred route. Getting things working on 100 parameters covers nearly all the immediate use cases. Issues for orbital optimization can be figured out later but won't change the user-facing aspect.
The file format will not be guaranteed between versions of the code, so we have total freedom to change it.

ye-luo:
If the vp file contains J1 and J2 but I only need J2 part, some additional input tag will be needed rather than taking all parameters from the file. To facilitate paramters selection, names of parameter sets and index ranges will be needed. My strong concern is to store names for each parameter in memory/files. I feel the need to get it correct now if we decide to store names for lookup or we don't store names and limit use case.

jtkrogel:
To improve the scaling of the name lookup, is it possible to use a hash table?  If so, the O(1) lookup time would be a benefit.

prckent:
I strongly think we need to support partial sets of parameters as best we are able. Users will expect this to work, and it is a nice productivity boost for anyone building the complexity of their wavefunctions.

jtkrogel:
Doesn't matter to me if version 1 has full names.  In later versions it would seem sufficient to name bunches of parameters that are then stored in arrays.  If we really only want to differentiate e.g. jastrow components, then named bunches would be sufficient for the matching and also be low storage.

markdewing:
For the linear-time parameter name lookup, I think it's something we might need to keep an eye on to make sure it doesn't take too much time.    I suspect it will not be a problem, at least not until we run into the other issues with VariableSet that require its modification.

ye-luo:
"bunch names" + index for parameter names will go away. So prefer to keep it as implementation details not exposed to users. Files are better with only "bunch names". If we expose the old way to files, keeping backward compatibility of the parameter file will be an issue. So I'd like to minimize such impact.

ye-luo:
When we run optimizer, the inactive variational parameters are stripped. While for vp file, do we want to keep all parameters or just the one touched by the optimizer?

jtkrogel:
If we only keep the ones modified, then we will have to specify the "untouched" parameters some other way, like by merging the full original files with the "delta" ones.

I would prefer writing all parameters so the wavefunction is "single file complete" as this is better for provenance and a simpler usage overall.  If the "delta" version is needed due to large parameter sets in the future, then I would add this as a later feature.

markdewing:
I think it depends on how the parameters are inactivated.  Using `optimize="no"` on the individual elements, the parameters never get added to the OptVariables list, and so never get written to the VP file.   There is another method for inactivating parameters by specifying the name.  This is not currently supported.  Given that it's not documented, I don't think it's used.

markdewing:
My guess is that support for individually named parameters will always be present.  Using a base name + index will be extend it, not replace it entirely.  Some wavefunction components don't use the base name + index format - such as the Pade Jastrows.  Now those aren't used much, so maybe it would be okay to force them to change their parameter names (?)

ye-luo:
> My guess is that support for individually named parameters will always be present. Using a base name + index will extend it, not replace it entirely. Some wavefunction components don't use the base name + index format - such as the Pade Jastrows. Now those aren't used much, so maybe it would be okay to force them to change their parameter names (?)

individual named parameter can be treated as individual groups of size 1. I don't expect only one group for each WFC. For example, J2 u-u and u-d are two arrays. We just need to chose one conversion.

markdewing:
So it sounds like we need to redesign VariableSet to handle parameter groups first (and change the file format to match).

ye-luo:
> So it sounds like we need to redesign VariableSet to handle parameter groups first (and change the file format to match).

I think we have to change that as we are moving forward. But right now, loading and dumping can be made to work without parameter names to have a working solution.

markdewing:
Okay, so I'll update the initial format to have just the list of parameter values to save/restore.

markdewing:
The prototype code has been simplified and updated.
 The XML format has been dropped in favor of HDF only.
The HDF file format stores only the parameters and not the names.   There is no name lookup on restore - the layout of VariableSet is expected to be exactly the same.

src/QMCDrivers/MCPopulation.cpp
```diff
@@ -283,6 +283,9 @@ void MCPopulation::set_variational_parameters(const opt_variables_type& active)
 {
   for (auto it_twfs = walker_trial_wavefunctions_.begin(); it_twfs != walker_trial_wavefunctions_.end(); ++it_twfs)
   {
+    opt_variables_type tmpVar;
+    (*it_twfs).get()->checkInVariables(tmpVar);
+    (*it_twfs).get()->checkOutVariables(active);
```
ye-luo, line 288: Some state was not maintained correctly by clone? I expect only resetParameters is necessary.
Please also use a range based loop.

src/QMCDrivers/QMCDriverInput.h
```diff
@@ -129,6 +131,8 @@ class QMCDriverInput
 
   const std::string get_drift_modifier() const { return drift_modifier_; }
   RealType get_drift_modifier_unr_a() const { return drift_modifier_unr_a_; }
+
+  const std::string get_variational_parameter_file() const { return variational_parameter_file_; }
```
ye-luo, line 135: `const std::string&` can be better.

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -247,6 +247,24 @@ void VMCBatched::process(xmlNodePtr node)
...
+
+    if (!qmcdriver_input_.get_variational_parameter_file().empty())
+    {
+      TrialWaveFunction& Psi = population_.get_golden_twf();
+      optimize::VariableSet tmpVar;
+      Psi.checkInVariables(tmpVar);
+      tmpVar.resetIndex();
+      Psi.checkOutVariables(tmpVar);
+
+      std::string vp_file = qmcdriver_input_.get_variational_parameter_file();
```
ye-luo, line 260: "const std::string&" or "auto&"

src/QMCDrivers/QMCDriverInput.h
```diff
@@ -93,6 +93,8 @@ class QMCDriverInput
   std::string drift_modifier_{"UNR"};
   RealType drift_modifier_unr_a_ = 1.0;
 
+  std::string variational_parameter_file_;
```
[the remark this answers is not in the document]
ye-luo: We need this for DMC as well. So good in QMCDriverInput.h.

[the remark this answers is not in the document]
ye-luo: @markdewing is this is prefix or full file name?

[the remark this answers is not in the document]
markdewing: It's the full file name.

src/QMCDrivers/MCPopulation.cpp
PDoakORNL: This seems like nonsense since checkInVariables documentation reads like it should have no significant side effects and its purpose is to populate the aggregate optimizable parameters list. If this must be called with the dummy variableSet because some component with optimizable parameters requires a side effect, file an issue if possible assign it against the author and record what requires this checkInVariables in a comment here.

It seems like this happens in QMCCostFunctionBase::put using the golden twf.
So I'm thinking of this cost function as the Rank scope cost function. And it seems to be the only one.

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -247,6 +247,24 @@ void VMCBatched::process(xmlNodePtr node)
                                 qmcdriver_input_.get_walkers_per_rank(), 1.0, qmcdriver_input_.get_num_crowds());
 
     Base::startup(node, awc);
+
+
+    if (!qmcdriver_input_.get_variational_parameter_file().empty())
```
PDoakORNL, line 252: This is not behavior that is meaningful to a VMCBatched driver not running inside of an optimizer, so this shouldn't be here. Move this code into the optimizer driver this will in turn make it easy to put the parameter into the right input class. 

It looks to me that the time to do it would be before constructing the vmcEngine so it is constructed with a "populated" golden wavefunction.
If you don't want that to be the one from the Wavefunction Pool, then make a clone of it.

I think this will also remove the need for set_variational_parameters to be called from this code block since your're not reading the parameters after you have cloned the gold wavefunction for the population.

PDoakORNL: I don't think behavior should be controlled by the presense or absense of this file input. And the default behavior of a program should not be controlled by implicitly observed state of the working directory without a explicit input parameter or cli argument.

PDoakORNL (changes requested):
Have classes do their own work.

src/QMCDrivers/VMC/VMCBatched.cpp
ye-luo: So move the input into `<wavefunction/>` where the gold is constructed?

ye-luo: > I don't think behavior should be controlled by the presense or absense of this file input. And the default behavior of a program should not be controlled by implicitly observed state of the working directory without a explicit input parameter or cli argument.

I have difficulty to follow this argument. You see an input tag and then act on it. It doesn't seem bothering "default".

PDoakORNL: The input tag and reading of it should be handled by WfOptDriverInput.

This code block should be in QMCFixedSampleLinearOptimizeBatched::processOptXML not in VMCBatched.

PDoakORNL: the variable name does not specify this is a file to load, just that it is a file name for the variational parameters.
The tag does say its to load but I had to drill into the implementation of the input reading to see that :( If I'm not coding the parsing of the input file I don't ever want to look at the text tags for input.

Where do the new variation parameters get saved and if I want to just use the same input.xml over and over 
again to keep optimizing (say in a premptable queue) can that be done?

ye-luo:
As loading variational parameters is an action against the gold TWF. Loading from file should be removed from drivers and implemented in WaveFunctionFactory which parses <wavefunction/> tag.

The Opt drivers only need to be responsible for writing vp files not reading them.

Let us call the input tag
<wavefunction>
... // usual conents
  <load_variational_parameters href="xxx.vp.h5">
      J2_uu rotatation J1 // this can be added in the future to select specific components to override
  </load_variational_parameters>
</wavefunction>

Then it will be driver agnostic.

PDoakORNL:
@ye-luo  I see what you mean now and yes that is cleaner and clearer.

markdewing:
Updated to read the file from the wavefunction section as Ye suggested.  It turns out the checkInVariables sequence is called right after that section is read in, and that makes an easy place to put the read call.

src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
```diff
@@ -239,6 +239,11 @@ void QMCCostFunctionBase::reportParameters()
     *msg_stream << "  </optVariables>" << std::endl;
     updateXmlNodes();
     xmlSaveFormatFile(newxml, m_doc_out, 1);
+
+
```
ye-luo, line 243: One empty can be sufficient.

src/QMCWaveFunctions/VariableSet.cpp
```diff
@@ -285,4 +291,55 @@ void VariableSet::print(std::ostream& os, int leftPadSpaces, bool printHeader) c
   }
 }
 
+
+
```
ye-luo, line 295: Having three empty lines is an overkill.

src/QMCWaveFunctions/VariableSet.cpp
```diff
@@ -285,4 +291,55 @@ void VariableSet::print(std::ostream& os, int leftPadSpaces, bool printHeader) c
...
+  std::vector<qmcplusplus::QMCTraits::ValueType> param_list;
+  for (auto& pair_it : NameAndValue)
+  {
+    param_list.push_back(pair_it.second);
+  }
+
+  hout.write(param_list, "parameter_values");
+
+
```
ye-luo, line 313: two empty lines.

src/QMCWaveFunctions/VariableSet.h
```diff
@@ -357,6 +357,13 @@ struct VariableSet
   void setDefaults(bool optimize_all);
 
   void print(std::ostream& os, int leftPadSpaces = 0, bool printHeader = false) const;
+
+
```
ye-luo, line 361: Keep one empty line and add documentation.

src/QMCWaveFunctions/VariableSet.h
```diff
@@ -357,6 +357,13 @@ struct VariableSet
...
+  void saveAsHDF(const std::string& filename) const;
+
+  /// Read variational parameters from an HDF file.
+  /// This assumes the VPs are already set up.  The wavefunction must match exactly.
```
ye-luo, line 365: VP can be "virtual particleset" as well. better to put the full name.

src/QMCWaveFunctions/WaveFunctionFactory.cpp
```diff
@@ -172,6 +180,12 @@ bool WaveFunctionFactory::build(xmlNodePtr cur, bool buildtree)
   targetPsi->checkInVariables(dummy);
   dummy.resetIndex();
   targetPsi->checkOutVariables(dummy);
+
+  if (!vp_file_to_load.empty()) {
+    app_log() << "  Reading variational parameters from " << vp_file_to_load << std::endl;;
```
ye-luo, line 185: apply clang format to adjust the { and also remove one ';'

src/QMCWaveFunctions/VariableSet.cpp
```diff
@@ -285,4 +291,55 @@ void VariableSet::print(std::ostream& os, int leftPadSpaces, bool printHeader) c
...
+  std::vector<int> vp_file_version{1, 0, 0};
+  hout.write(vp_file_version, "version");
+  std::string timestamp(getDateAndTime("%Y-%m-%d %H:%M:%S %Z"));
+  hout.write(timestamp, "timestamp");
+
+  std::vector<qmcplusplus::QMCTraits::ValueType> param_list;
+  for (auto& pair_it : NameAndValue)
+  {
+    param_list.push_back(pair_it.second);
+  }
```
ye-luo, line 309: remove {} around one line.

ye-luo (commented):
Could you add the parameter loading xml block to opt.xml written by the driver?

ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
Test this please

markdewing:
I prefer putting the parameter names in the save file, and loading by parameter name (these are two separate considerations).

Reasons:
1. WIth the names present, the file is still comprehensible on its own. Putting the names in the file gives some indication as to what parameters are in the wavefunction.
2. Changing to store the base name + range requires changes to VariableSet.  These changes are likely to be pervasive, and will probably take a while to implement.
3. Loading by name enables the common use-case of incremental optimization out of the box.  (No need to wait on item 2 to be finished)
4. If one wants to take subsets of parameters from different files, it would be straightfoward to write a script to do this (or use a script that converts to a simple text format, edit that format, and then use a script to convert back to HDF)

Responses to some objections:
- The parameter names are printed in the QMCPACK output, so they are exposed to the user already.
- Backwards compatibility: The parameter names are pretty regular, so it seems like it would not be hard to automatically convert (either in the code or as a separate script).  The risk here is the difficulty in writing and using the converter to update from version 1.0.0.
- Inserting by name is N^2.  I ran some timing tests.  Loading took about 7 seconds with 50K parameters and 30 seconds with 100K parameters.  Note that setting up VariableSet proceeds by a series of insert calls, so it takes as much time to set up the class as to load by name (The given times are just the load step).    There are so many scaling challenges that the added cost of loading by name doesn't seem to be large drawback.

Some of these issues will be addressed by a new implementation of VariableSet, but that may take a while (2, above).  Having incremental optimization and subsets available immediately seems worth the scalability drawbacks and backwards compatibility risks.

ye-luo:
To 2. name + range only makes sense if we store all the parameters in a single array. Once we refactor VariableSet, probably there won't be a single array but in groups.
3. good point.
4. We shouldn't mention this to users but do the improvement on our side. Power users may do whatevery they want.

Respond to `Responses to some objections`:
- Current parameter names are printed only and users don't have ways to control it and influence the code behavior. So I consider they are "not exposed"
- I don't think they can be converted easily. How many digits in the index? Tricky.
- OK. we can leave the N^2 problem for now but once VariableSet is fixed, this issue should go away by design.

PDoakORNL (dismissed):

markdewing:
This functionality has been implemented and checked in.

PR #3637, opened 2021-12-02
Remove CrayPE specialization for OpenMP offload.

ye-luo:
## Proposed changes
CrayPE has architecture modules. The GPU ones like craype-accel-nvidia80 introduces a lot of undesirable side effects. Since such modules are mostly useful for GPU aware MPI which we don't need, we can just avoid using them and rely on our usually ways of setting GPU flags.

- Build related changes

- No

Polaris. Tested with PrgEnv-gnu,  PrgEnv-nvhpc,  PrgEnv-cray. I also tested Cray compiler offload to AMD gpus, the compilation went as expected.

## Checklist


PDoakORNL:
Does this require a manual update as well?

ye-luo:
> Does this require a manual update as well?

Pushed an update

prckent:
Please can you confirm that this is OK on spock.

PDoakORNL (approved):
I agree with @ye-luo with how to handle Cray

prckent:
After discussing with Ye, I am also OK with this change.

prckent (approved):

prckent:
Start testing in-house

ye-luo:
I removed clang14dev as required. I think we should use stable clang13 instead for offload builds in CI.

prckent:
@ye-luo Do you have evidence that clang13 is OK? Using a release version is an improvement, I admit.

@williamfgc  Please can you update the CI based on the eventual responses.

williamfgc:
@ye-luo @prckent clang13 has that bug that increased memory allocation when offloading that @ye-luo reported and is fixed [in the development branch](https://reviews.llvm.org/rG01d59c0de822099c62f12f275c41338f6df9f5ac). I'll upgrade to clang14 when it comes out. Let me know how to proceed.

prckent:
@ye-luo Newer clang dev OK?

ye-luo:
Current clang dev main has regression since it moved to a new DeviceRTL. A few bugs still needs to be eliminated before being recommended for use. Now I remember that the clang14dev used in CI is likely a version I recommended. I think for the moment, we should stick to this version at least until I have clang 14 commit to recommend.

PR #3640, opened 2021-12-06
Save and restore variational parameters

markdewing:
1. Save the variational parameters during optimization to '*.vp.h5' files
2. Add an 'override_variational_parameters' tag to the 'wavefunction' section to load a set of variational parameters.
3. Output the override tag to intermediate wavefunction structures during optimization

Item 3 has the largest potential to change the behavior of the code during optimization.  A flag was added to make it off by default.  Use `<parameter name="output_vp_override">yes</parameter>` in the optimizer section to enable it.   This will be enabled by default eventually, but it needs more testing.

This is more implementation of #3634.

A test is forthcoming (makes this PR simpler to not have the test here)

- New feature

- No

desktop

## Checklist


prckent:
"Output the override tag to intermediate wavefunction structures during optimization". If this refers to the current parameters outputs (XML) during optimization, we have a bunch of scripts that currently expect these outputs. e.g. NEXUS knows how to parse the different interaction and is smarter than QMCPACK in selecting the "best" wavefunction. Both should be kept for now / wider discussion held. Updates will be needed. Fairly easy but they need doing. We can debate the importance of being able to recreate the XML inputs with new parameters.

ye-luo:
When 002.vp.h5 is written. I expect 002.opt.xml to carry the override tag. Otherwise, it can easily go wrong if some parameters are only written to vp.h5 without in xml.

markdewing:
"Output the override tag to intermediate wavefunction structures during optimization" - yes, this does refer to the .opt.xml files produced during optimization.  These are the files that are *supposed* to contain all the updated parameters at each step, but do not for some types of parameters.
If the code adds the override tag, then all the updated parameters should be accessible by referencing the .opt.xml file.

ye-luo (approved):
OK. We can introduce `output_vp_override` option for development purpose but we need to remove it once the workflow integration details are sorted out #3653.

ye-luo:
Start testing in-house

ye-luo:
Need non-ANL approval

prckent (approved):
@ye-luo Thanks for making the tracking issue for the big release.

PR #3660, opened 2021-12-12
CMake fix for NVHPC

ye-luo:
## Proposed changes
1. Catch problems early in CMake when CMake failed to handle NVHPC internal CUDA toolkit.
2. Check CMAKE_CXX_FLAGS_RELWITHDEBINFO and CMAKE_CXX_FLAGS_RELEASE existence before trying to REPLACE.

- Build related changes

- No

epyc-server

## Checklist


prckent:
Historically, removing NDEBUG from ppconvert caused it to not converge with certain compilers. We may have fixed enough of the historical problem that this is no longer the case, but it should be checked carefully.

ye-luo:
> Historically, removing NDEBUG from ppconvert caused it to not converge with certain compilers. We may have fixed enough of the historical problem that this is no longer the case, but it should be checked carefully.

We need an issue to track this. Right now for some compilers `CMAKE_CXX_FLAGS_RELEASE` and `CMAKE_CXX_FLAGS_RELWITHDEBINFO` are not explicitly defined. For this reason, CMake stops and complains about mismatched arguments when REPLACE. So add some protection here.

prckent:
The outcome should be no special cmake / compiler handling needed for the ppconvert executable. The fastmath dependency is presumably related. I am unsure if the problem still exists or if the sanitizer cleanup fixed things.

CMakeLists.txt
```diff
@@ -730,6 +730,11 @@ if(QMC_CUDA OR ENABLE_CUDA)
     set(CMAKE_CUDA_EXTENSIONS OFF)
     enable_language(CUDA)
     find_package(CUDAToolkit REQUIRED)
+    if(NOT TARGET CUDA::cublas)
+      message(FATAL_ERROR "Found an incomplete CUDA toolkit installation. "
+                          "This often happens when CMake failed in recognizing the NVHPC internal CUDA toolkit. "
+                          "Set CMAKE_CUDA_COMPILER to the full path of nvcc from a complete CUDA toolkit installation.")
+    endif()
```
quantumsteve, line 737: Does NVHPC require CMake 3.22?

ye-luo: No. NVHPC needs to work as a CPU compiler. So current default CMake version minimum is good.

quantumsteve:
Start testing in-house

quantumsteve (approved):
Looks good to me. Tracking problems requiring these workarounds in a separate issue.

ye-luo:
Start testing in-house

PR #3665, opened 2021-12-14
Fix restrict type qualifier misuse.

ye-luo:
## Proposed changes
NVHPC compiler engineers identified this issue.

We define `const RealType* restrict Ylm_v = Ylm[0];` but then directly modified it via `Ylm.evaluateV(grid[k][0], grid[k][1], grid[k][2]);`
Now we just define `Ylm_v` after `Ylm.evaluateV` and removed restrict which doesn't really have benefit in this case.

- Bugfix
- No

epyc-server

## Checklist


prckent (approved):
LGTM

Thanks to the NVHPC team for catching this.

ye-luo:
Start testing in-house

ye-luo:
Start testing in-house

PR #3671, opened 2021-12-16
Refactor ACForce into a class

williamfgc:
## Proposed changes

Refactored ACForce into a class
Separate declaration (.h) from definition (.cpp)
Initial attempt to add some unit test to `ACForce`, currently at 0% in codecov
Related to #2238 and #3411

- No

Ubuntu 20.04, must pass CI

## Checklist

### DIFF
--- a/src/QMCHamiltonians/ACForce.h
+++ b/src/QMCHamiltonians/ACForce.h
@@ -23,69 +23,71 @@
 
 namespace qmcplusplus
 {
-struct ACForce : public OperatorBase
+class ACForce : public OperatorBase
 {
-  typedef ParticleSet::ParticlePos_t Force_t;
+public:
+  using Force_t = ParticleSet::ParticlePos_t;
+
   /** Constructor **/
   ACForce(ParticleSet& source, ParticleSet& target, TrialWaveFunction& psi, QMCHamiltonian& H);
+
   /** Destructor **/
-  ~ACForce() override{};
-  /** Copy constructor **/
-  //ACForce(const ACForce& ac)  {};
+  ~ACForce() final = default;
 
   /** I/O Routines */
-  bool put(xmlNodePtr cur) override;
-  bool get(std::ostream& os) const override { return true; };
+  bool put(xmlNodePtr cur) final;
+
+  bool get(std::ostream& os) const final;
 
   /** Cloning **/
   //We don't actually use this makeClone method.  We just put an APP_ABORT here
   std::unique_ptr<OperatorBase> makeClone(ParticleSet& qp, TrialWaveFunction& psi) final;
+
   //Not derived from base class.  But we need it to properly set the Hamiltonian reference.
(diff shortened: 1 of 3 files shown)


rcclay (approved):
LGTM.  I'll need to beef up the unit test in the future.

ye-luo:
Test this please

src/QMCHamiltonians/ACForce.h
```diff
@@ -23,69 +23,71 @@
 
 namespace qmcplusplus
 {
-struct ACForce : public OperatorBase
+class ACForce : public OperatorBase
 {
-  typedef ParticleSet::ParticlePos_t Force_t;
+public:
+  using Force_t = ParticleSet::ParticlePos_t;
```
PDoakORNL, line 29: Don't see any reason for the _t  Force is obviously a type.

src/QMCHamiltonians/tests/test_force.cpp
```diff
@@ -425,4 +425,84 @@ TEST_CASE("Ion-ion Force", "[hamiltonian]")
...
+  ions.update();
+
+  elec.addTable(ions);
+  elec.update();
+
+  // defaults
+  TrialWaveFunction psi;
+  QMCHamiltonian qmcHamiltonian;
+
+  ACForce force(ions, elec, psi, qmcHamiltonian);
```
PDoakORNL, line 480: naming is going to become confusing once you start checking results i.e. the `Force` type shows up in tests.

PDoakORNL (changes requested):
only minor changes needed

src/QMCHamiltonians/ACForce.h
PDoakORNL: Additionally its actually the vector of the per particle forces.  so `Forces` would provide a much better clue for what is going on.

src/QMCHamiltonians/tests/test_force.cpp
williamfgc: @PDoakORNL since we are switching `Force_t` to `Forces`, do you think the naming can stay consistent with existing tests coming from other force types? At least for now, naming refactoring would have a larger scope in a separate PR.

williamfgc:
Start testing in-house

src/QMCHamiltonians/tests/test_force.cpp
PDoakORNL: I think that would be nice ultimately but I don't like to see "new" code going in with old names, my experience is then there is no pressure to update at all. If this just makes the naming inconsistent fine,  however if breaks all those tests and will bring in many more files into this PR maybe it should get split into a separate PR up to you.

In the case of other force tests I'm not even sure if those will be kept once @rcclay 's new implementation is in so I wouldn't worry about them unless it breaks the build.

PDoakORNL (approved):

PR #3677, opened 2021-12-17
Update qmcfinitesize to use UBsplines instead of NUBsplines

camelto2:

## Proposed changes

This PR removes the dependence of qmcfinitesize on the non-uniform bsplines in favor of uniform bsplines. 
I recently noticed a bug on my mac with gnu compilers for the qmcfinitesize tool where I was getting unreasonably large error bars. I ended up removing the random resampling techniques and worked with just fixed S(k) data (which should always give the same FS correction since the input data is fixed and there is no resampling), and I would see that every few executions of the code I would get erroneous finite size corrections. I ultimately tracked this down specifically to the use of NUBspline_1d_d for the integral part of the finite size correction. Even though all of the underlying input data was the same to generate the spline, either the construction or evaluation of the NUBspline would sometimes break and I'm entirely sure why.  

The use of NUBspline use was a remnant carried over from the original tool, and we no longer actually define the integrand on a nonuniform grid anyway, so I switched everything to using UBsplines, and every evaluation is now consistent and gives the expected answer. I have only ever noticed this with gnu compilers on my mac...when I used intel compilers on one our compute nodes I never saw this behavior iirc. 

I'm not even sure if NUBspline is well tested, whereas UBsplines are used more commonly throughout the code...so making the switch would probably be a good thing even if there wasn't this strange bug. 

- Bugfix

- No

my macbook, gnu-11

## Checklist

### DIFF
--- a/src/QMCTools/QMCFiniteSize/QMCFiniteSize.cpp
+++ b/src/QMCTools/QMCFiniteSize/QMCFiniteSize.cpp
@@ -5,9 +5,6 @@
 #include <cmath>
 #include "Configuration.h"
 #include "einspline/bspline_eval_d.h"
-#include "einspline/nubspline_eval_d.h"
-#include "einspline/nugrid.h"
-#include "einspline/nubspline_create.h"
 #include "QMCTools/QMCFiniteSize/FSUtilities.h"
 #include "Utilities/RandomGenerator.h"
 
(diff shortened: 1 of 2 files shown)


rcclay (approved):

prckent (approved):
Very happy to see this.

I agree with the questions around the nub-spline code.

prckent:
Test this please

prckent:
Start testing in-house

PR #3678, opened 2021-12-18
Remove MuffinTin and NUBspline 

ye-luo:
## Proposed changes
Remove MuffinTin which was integrated in the obsolete EinsplineSet class.
And thus managed to remove all traces of NUBspline in the entire repo.
~20k lines removed including unused test code from einspline.

- No

epyc-server

## Checklist


ye-luo:
Start testing in-house

prckent:
Thanks Ye! I think we have both been itching to remove this code. I think we should take a careful look at the coverage  in  the remaining files - some of the test codes you are  removing could potentially be used to get better coverage. They also could work as unit tests. (Written before the CI coverage has run)

prckent:
@PDoakORNL Is there anything you would  like to keep?

ye-luo:
> Thanks Ye! I think we have both been itching to remove this code. I think we should take a careful look at the coverage in the remaining files - some of the test codes you are removing could potentially be used to get better coverage. They also could work as unit tests. (Written before the CI coverage has run)

We should not use any evaluation functions from einspline. So a lot of test cores are irreverent. Right now we only keep the data structure and the value to spline coefficients construction routine, they should have been covered by `spline2/tests` or `einspline/tests`. So safe to remove.

ye-luo:
The hole project metric increased.
`codecov/project Successful in 1s — 43.12% (+1.13%) compared to 59e65c8`

prckent:
Stesting testing in-house

prckent:
Start testing in-house

prckent (approved):

PR #3681, opened 2021-12-20
c++17 API for lowerCase, removal of tolower and XMLAttrString refactor of getNodeString

PDoakORNL:
## Proposed changes

* add a std::string_view based lowerCase to replace OhmmsElementBase.h tolower function.
* add sub target for qmcutil, qmcutil_basic to allow linking for this function without pulling in message and qmcio.

* remove XMLAttrString and replace with std::string and getXMLAttributeValue function. Nothing actually makes use of the extra state and methods of the derived class and it complicates string processing.

* refactor getNodeString to actually return a string rather than write to output argument.  This combined with a large amount of sloppy scoping / premature optimization. Allows it to be chained with lowerCase as well.

* add castCharToXMLChar and castXMLCharToChar casting functions to replace c casts. As we overhaul input we should use these to eliminate c cast from the application xml parser boundary code.

* Add implementation file ParameterSet.cpp and explict instantiations of all used member template functions.

This will be used by future PR's improving input handling in EstimatorManagerNew.

- New feature
- Documentation changes

- No

Leconte

## Checklist

### DIFF
--- a/src/Particle/ParticleIO/ParticleLayoutIO.cpp
+++ b/src/Particle/ParticleIO/ParticleLayoutIO.cpp
@@ -24,7 +24,7 @@
 #include "OhmmsData/AttributeSet.h"
 #include "QMCWaveFunctions/ElectronGas/HEGGrid.h"
 #include "LongRange/LRCoulombSingleton.h"
-#include "Utilities/ModernStringUtils.hpp"
+#include "ModernStringUtils.hpp"
 
 namespace qmcplusplus
 {
--- a/src/Particle/ParticleIO/ParticleLayoutIO.cpp
+++ b/src/Particle/ParticleIO/ParticleLayoutIO.cpp
@@ -44,8 +44,6 @@ bool LatticeParser::put(xmlNodePtr cur)
   bool bconds_defined  = false;
   int boxsum           = 0;
 
-  std::string handler_type("opt_breakup");
-
   app_summary() << std::endl;
   app_summary() << " Lattice" << std::endl;
   app_summary() << " -------" << std::endl;
@@ -111,7 +109,7 @@ bool LatticeParser::put(xmlNodePtr cur)
       }
       else if (aname == "LR_handler")
       {
-        std::string handler_type;
+        std::string handler_type("opt_breakup");
         //This chops whitespace so the simple str == comparisons work
         putContent(handler_type, cur);
         handler_type = lowerCase(handler_type);
--- a/src/Particle/ParticleIO/ParticleLayoutIO.cpp
+++ b/src/Particle/ParticleIO/ParticleLayoutIO.cpp
@@ -24,6 +24,7 @@
 #include "OhmmsData/AttributeSet.h"
 #include "QMCWaveFunctions/ElectronGas/HEGGrid.h"
 #include "LongRange/LRCoulombSingleton.h"
+#include "Utilities/ModernStringUtils.hpp"
 
 namespace qmcplusplus
 {
@@ -55,24 +56,24 @@ bool LatticeParser::put(xmlNodePtr cur)
     if (cname == "parameter")
     {
       const XMLAttrString aname(cur, "name");
-      if (aname == "scale")
+      if (aname.getValue() == "scale")
       {
         putContent(a0, cur);
       }
-      else if (aname == "lattice")
+      else if (aname.getValue() == "lattice")
       {
         const XMLAttrString units_prop(cur, "units");
-        if (!units_prop.empty() && units_prop != "bohr")
+        if (units_prop.hasValue() && units_prop.getValue() != "bohr")
         {
           APP_ABORT("LatticeParser::put. Only atomic units (bohr) supported for lattice units. Input file uses: "
-                    << std::string(units_prop));
+                    << std::string(units_prop.getValue()));
         }
 
         putContent(lattice_in, cur);
         lattice_defined = true;
         //putContent(ref_.R,cur);
       }
-      else if (aname == "bconds")
+      else if (aname.getValue() == "bconds")
       {
         putContent(bconds, cur);
         bconds_defined = true;
@@ -100,18 +101,20 @@ bool LatticeParser::put(xmlNodePtr cur)
                 "LatticeParser::put. In \"bconds\", non periodic directions must be placed after the periodic ones.");
         }
       }
-      else if (aname == "vacuum")
+      else if (aname.getValue() == "vacuum")
       {
         putContent(ref_.VacuumScale, cur);
       }
-      else if (aname == "LR_dim_cutoff")
+      else if (aname.getValue() == "LR_dim_cutoff")
       {
         putContent(ref_.LR_dim_cutoff, cur);
       }
-      else if (aname == "LR_handler")
+      else if (aname.getValue() == "LR_handler")
       {
+        std::string handler_type;
+        //This chops whitespace so the simple str == comparisons work
         putContent(handler_type, cur);
-        tolower(handler_type);
+        handler_type = lowerCase(handler_type);
         if (handler_type == "ewald")
           LRCoulombSingleton::this_lr_type = LRCoulombSingleton::EWALD;
         else if (handler_type == "opt_breakup")
@@ -121,11 +124,11 @@ bool LatticeParser::put(xmlNodePtr cur)
         else
           APP_ABORT("\n  Long range breakup handler not recognized.\n");
       }
-      else if (aname == "LR_tol")
+      else if (aname.getValue() == "LR_tol")
       {
         putContent(ref_.LR_tol, cur);
       }
-      else if (aname == "rs")
+      else if (aname.getValue() == "rs")
       {
         lattice_defined = true;
         OhmmsAttributeSet rAttrib;
@@ -135,7 +138,7 @@ bool LatticeParser::put(xmlNodePtr cur)
         rAttrib.put(cur);
         putContent(rs, cur);
       }
-      else if (aname == "nparticles")
+      else if (aname.getValue() == "nparticles")
       {
         putContent(nptcl, cur);
       }
--- a/src/Particle/ParticleIO/ParticleLayoutIO.cpp
+++ b/src/Particle/ParticleIO/ParticleLayoutIO.cpp
@@ -55,25 +55,25 @@ bool LatticeParser::put(xmlNodePtr cur)
     std::string cname((const char*)cur->name);
     if (cname == "parameter")
     {
-      const XMLAttrString aname(cur, "name");
-      if (aname.getValue() == "scale")
+      const std::string aname(getXMLAttributeValue(cur, "name"));
+      if (aname == "scale")
       {
         putContent(a0, cur);
       }
-      else if (aname.getValue() == "lattice")
+      else if (aname == "lattice")
       {
-        const XMLAttrString units_prop(cur, "units");
-        if (units_prop.hasValue() && units_prop.getValue() != "bohr")
+        const std::string units_prop(getXMLAttributeValue(cur, "units"));
+        if (!units_prop.empty() && units_prop != "bohr")
         {
           APP_ABORT("LatticeParser::put. Only atomic units (bohr) supported for lattice units. Input file uses: "
-                    << std::string(units_prop.getValue()));
+                    << units_prop);
         }
 
         putContent(lattice_in, cur);
         lattice_defined = true;
         //putContent(ref_.R,cur);
       }
-      else if (aname.getValue() == "bconds")
+      else if (aname == "bconds")
       {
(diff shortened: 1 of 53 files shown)


ye-luo:
Could you retire tolower now? It can test if strToLower fully cover all the use case of tolower.
Prefer avoiding both functions kept in the repo.

function name better start with verbs. How about lowerCase(string)?

src/io/OhmmsData/XMLParsingString.h
```diff
@@ -50,41 +89,60 @@ class XMLNodeString : public std::string
   }
 };
 
-/** convert an xmlNode attribute into a std::string
+/** convert an xmlNode attribute into a name value pair.
+ * used to inherit from std::string but to no advantage
+ * and repeated confusion about whether comparisons to an attribute string were between
```
ye-luo, line 94: This is part of the commit log or PR description but not class documentation. It is only understandable when doing diff.

src/io/OhmmsData/XMLParsingString.h
```diff
@@ -50,41 +89,60 @@ class XMLNodeString : public std::string
...
+    xmlChar* attr_char = xmlGetProp(cur, castCharToXMLChar(name));
     if(attr_char)
     {
-      assign((const char*)attr_char);
+      attribute_.assign(castXMLCharToChar(attr_char));
       xmlFree(attr_char);
     }
   }
-
+
```
ye-luo, line 115: Looks like need some formatting.

src/io/OhmmsData/XMLParsingString.h
```diff
@@ -50,41 +89,60 @@ class XMLNodeString : public std::string
...
   {
-    xmlSetProp(cur, (const xmlChar*)attribute_name_.c_str(), (const xmlChar*)this->c_str());
+    xmlSetProp(cur, castCharToXMLChar(attribute_name_.c_str()), castCharToXMLChar(attribute_.c_str()));
   }
+
+  /** retain surprising behavior where this inherently name value object 
+   *  can also operate as just a string of its value
+   *  \todo remove this its a footgun and makes for less readable code
+   */
+  operator const std::string&() { return attribute_; }
```
ye-luo, line 141: Now. I'm confused. You turned the base class to a member because of confusion. Then you provide this convenience conversion which retain the confusion.

I kind of still like the old way. To reduce the confusion. We can call it XMLAttrValueString. Having String in the type name is to say it is derived from std::string and can be used as an std::string object.

With the current modification, the class need to be removed probably as XMLStringAttr indicating it is an XML attribute for string data.

src/io/OhmmsData/XMLParsingString.h
```diff
@@ -19,10 +19,49 @@
...
+
+/** xmlChar* to char* cast
+ *  certainly not typesafe but at this interface between libxml and c++
+ *  well it beats c style casts and might be more correct than a reinterpret.
+ *
+ *  This is fine with UTF-8 bytes going into a std::string.
+ */
+inline char* castXMLCharToChar(xmlChar* c)
+{
+  return static_cast<char*>(static_cast<void *>(c));
```
ye-luo, line 37: What is the reason for `void *`?

src/io/OhmmsData/XMLParsingString.h
PDoakORNL: I'll just make the extra changes and delete the conversion. 

It's a pair of string buffers. a name value pair. It's methods should make this clear. It could just be XMLAttr. I didn't like the way it interacted with lowerCase so I think its time to change it. Also the fact that it has an attribute_name is often called aname and is most frequently used for attributes with the tag "name" makes me think less implicit here is going to be much more clear.

Subclassing standard stl containers and giving them weird semantics is not good practice. It results in surprising behavior. There is no good reason to subclass this except to get a bunch of cute operators that only effect the value half of the object. Or apply mutating functions for std::string that perhaps surprisingly only effect the value part of the pair.

ye-luo: I thought about XMLAttr but it doesn't say the value type is a string or int or float. For this reason, XMLStringAttr comes out.

PDoakORNL: It seems easy to just remove the type altogether I could find anywhere the state is actually used.

ye-luo: > It seems easy to just remove the type altogether I could find anywhere the state is actually used.

Confused. Which type? Which state?

src/io/OhmmsData/XMLParsingString.h
PDoakORNL: Apparently guarantees that the address doesn't change. But it seems a bit controversial, some claim since c++11 reinterpret_cast means the same thing.  I think maybe since all the types involved here are of the same size on most platforms it never mattered. 
I think in c++17 it can be replaced with a reinterpret_cast at no loss but that looks like an emergency.

src/io/OhmmsData/XMLParsingString.h
PDoakORNL: XMLAttrString which differs from a normal string in that it remembers the name of the attribute it was constructed from.

PDoakORNL:
Start testing in-house

src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
```diff
@@ -164,8 +164,8 @@ std::unique_ptr<WaveFunctionComponent> RadialJastrowBuilder::createJ2(xmlNodePtr
   using J2Type     = typename JastrowTypeHelper<RadFuncType, Implementation>::J2Type;
   using DiffJ2Type = typename JastrowTypeHelper<RadFuncType, Implementation>::DiffJ2Type;
 
-  XMLAttrString input_name(cur, "name");
-  std::string j2name = input_name.empty() ? "J2_" + Jastfunction : input_name;
+  std::string input_name(getXMLAttributeValue(cur, "name"));
+  std::string j2name = !input_name.empty() ? input_name : "J2_" + Jastfunction;
```
ye-luo, line 168: Prefer changing `!empty()` back to `empty()`.

src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
```diff
@@ -348,8 +348,8 @@ std::unique_ptr<WaveFunctionComponent> RadialJastrowBuilder::createJ1(xmlNodePtr
   using TH     = JastrowTypeHelper<RadFuncType, Implementation>;
   using J1Type = typename std::conditional<SPIN, typename TH::J1SpinType, typename TH::J1Type>::type;
 
-  XMLAttrString input_name(cur, "name");
-  std::string jname = input_name.empty() ? Jastfunction : input_name;
+  std::string input_name(getXMLAttributeValue(cur, "name"));
+  std::string jname = !input_name.empty() ? input_name : Jastfunction;
```
ye-luo, line 352: Prefer changing `!empty()` back to `empty()`.

src/QMCWaveFunctions/Jastrow/eeI_JastrowBuilder.cpp
```diff
@@ -129,9 +129,8 @@ std::unique_ptr<WaveFunctionComponent> eeI_JastrowBuilder::buildComponent(xmlNod
     tAttrib.add(ftype, "function");
     tAttrib.put(cur);
 
-    XMLAttrString input_name(cur, "name");
-    std::string jname = input_name.empty() ? "JeeI_" + ftype : input_name;
-
+    std::string input_name(getXMLAttributeValue(cur, "name"));
+    std::string jname = !input_name.empty() ? input_name : "JeeI_" + ftype;
```
ye-luo, line 133: Prefer changing `!empty()` back to `empty()`.

src/QMCWaveFunctions/LCAO/LCAOrbitalBuilder.cpp
```diff
@@ -132,8 +132,8 @@ LCAOrbitalBuilder::LCAOrbitalBuilder(ParticleSet& els, ParticleSet& ions, Commun
   processChildren(cur, [&](const std::string& cname, const xmlNodePtr element) {
     if (cname == "basisset")
     {
-      XMLAttrString basisset_name_input(element, "name");
-      std::string basisset_name(basisset_name_input.empty() ? "LCAOBSet" : basisset_name_input.c_str());
+      std::string basisset_name_input(getXMLAttributeValue(element, "name"));
+      std::string basisset_name(!basisset_name_input.empty() ? basisset_name_input : "LCAOBSet" );
```
ye-luo, line 136: Prefer changing `!empty()` back to `empty()`.

src/Utilities/ModernStringUtils.hpp
```diff
@@ -0,0 +1,51 @@
...
+ *  on the string buffer which is undesirable and should be explicitly evident in the code.
+ *  i.e.
+ *  my_string = strToLower(my_string);
+ *
+ *  According to en.cppreference.com std::tolower is only defined for usigned char and EOF. the std::tolower conversion
+ *  is based on the _c_ locale which to makes it unclear on what might happen to char > 127. 
+ *  *  For tags, and keywords where we define explicitly define they are ASCII encoded and this is fine. 
+ *  *  For other XML derived text this should never be used since we should assume that to be UTF-8 encoded.
+ */
+inline std::string lowerCase(const std::string_view s)
```
ye-luo, line 39: Separating hpp and cpp is still beneficial. Let us do that. Could you also make the documentation up to date by removing mention of `tolower`

ye-luo (commented):
Only minor things left. I like the change from XMLAttrString to getXMLAttributeValue

src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
PDoakORNL: These are returned to there original configuration.

src/Particle/ParticleIO/ParticleLayoutIO.cpp
```diff
@@ -110,8 +111,10 @@ bool LatticeParser::put(xmlNodePtr cur)
       }
       else if (aname == "LR_handler")
       {
+        std::string handler_type;
+        //This chops whitespace so the simple str == comparisons work
         putContent(handler_type, cur);
-        tolower(handler_type);
+        handler_type = lowerCase(handler_type);
```
ye-luo, line 117: There was a default value.

ye-luo (commented):
I need a bit more thinking of the CMake change.

src/Utilities/CMakeLists.txt
```diff
@@ -21,15 +26,19 @@ set(UTILITIES
     ResourceCollection.cpp
     ProjectData.cpp
     RandomNumberControl.cpp)
-add_library(qmcutil ${UTILITIES})
+add_library(qmcutil_integrated ${UTILITIES})
```
ye-luo, line 29: I dislike `qmcutil_basic` and `qmcutil_integrated` it doesn't help making more organized targets. Just like `_new`, `_basic` only adds confusion. We need something that improves the clarity.  Probably rename `qmcutil_basic` to `cxx_helper` and restrict it only to enhance c++ standard library and the dependency is restricted to cxx language and standard library.

ye-luo: I'm now wondering if we should just avoid significant CMake changes in this PR and just use qmcutil. breaking up qmcutil is better on a separate PR.

PDoakORNL: I'll give it a shot.  At one point I was having linking errors I thought due to circular dependency between qmcio and qmcutil and that's what induced this.  I found another cause for that as well.

PDoakORNL: if I restore that target split in the future I'll go with the cxx_helper name.

ye-luo: I see what you mean "circular dependency between qmcio and qmcutil". If you add the new cpp to qmcuitl, It could happen, then just add cxx_helper.

src/Particle/ParticleIO/ParticleLayoutIO.cpp
PDoakORNL: I got rid of the other definition that was shadowed.  I copied the default initialization but it took me a while to decide the istream corner case that doesn't clear the string iff isn't UB. that aspect of putContent didn't actually catch my attention before.

src/Utilities/CMakeLists.txt
PDoakORNL: for now we are getting away with it so I think its low priority.

ye-luo (approved):

ye-luo:
Test this please

PR #3682, opened 2021-12-21
Write HDF strings compatible with h5py V2 and V3

### LINKED ISSUE #3679: Faile to execute convert_vp_format.py
**Describe the bug**
CMake succeeds but with python error.

```
  File "/home/yeluo/opt/cleanup/qmcpack/tests/molecules/He_param/convert_vp_format.py", line 211, in <module>
    convert_from_text_to_hdf(fname_in, fname_out, args.complex)
  File "/home/yeluo/opt/cleanup/qmcpack/tests/molecules/He_param/convert_vp_format.py", line 179, in convert_from_text_to_hdf
    write_to_hdf(vp, fname_out, output_complex)
  File "/home/yeluo/opt/cleanup/qmcpack/tests/molecules/He_param/convert_vp_format.py", line 163, in write_to_hdf
    f.create_dataset("timestamp",data=[vp.timestamp])
  File "/usr/lib/python3/dist-packages/h5py/_hl/group.py", line 136, in create_dataset
    dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
  File "/usr/lib/python3/dist-packages/h5py/_hl/dataset.py", line 118, in make_new_dset
    tid = h5t.py_create(dtype, logical=1)
  File "h5py/h5t.pyx", line 1634, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1656, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1717, in h5py.h5t.py_create
TypeError: No conversion path for dtype: dtype('<U23')
```

**To Reproduce**
It is on my workstation. No idea what happens.

**Expected behavior**
Pass.

**System:**
 - Epyc server.

**Additional context**
`execute_process(COMMAND` needs to check return code and takes action.

markdewing:
Fixes #3679

The issue seems to be different handling of strings in h5py v2 vs. v3.  In v3, passing an array of strings to `create_dataset` seems to work as expected.  In v2, the handling of strings appears to be the same as converting the array of strings to a numpy array via `np.array`. This conversion of a string to a numpy array makes fixed-length strings (eg. dtype is 'U23'), which h5py does not know how to handle.  The solution seems to be create the array as an array of objects (`dtype=object`), and then use the `dtype` parameter on `create_dataset` to specify that it is a string type.

https://docs.h5py.org/en/latest/strings.html

- Bugfix

- No

laptop

## Checklist


ye-luo (approved):
Verified. Works on my workstation now.

ye-luo:
Test this please

tests/molecules/He_param/convert_vp_format.py
```diff
@@ -160,10 +160,10 @@ def write_to_hdf(vp, fname_out, output_complex):
     size = len(vp.name_value_pairs)
 
     f = h5py.File(fname_out,"w")
-    f.create_dataset("timestamp",data=[vp.timestamp])
+    f.create_dataset("timestamp",data=np.array([vp.timestamp],dtype=object),dtype=h5py.string_dtype('ascii'))
```
PDoakORNL, line 163: why not utf-8?

markdewing: That would require handling UTF-8 in the C++ code that writes and reads it.  Or at least testing it that way (maybe it would work correctly already?)

PDoakORNL: earlier in this file timestamp is decoded as utf-8.  Personally I think we should use and assume use of utf-8 encoding. We can still process our internal set strings as ascii but user supplied strings should be allowed to contain utf-8 codepoints.  In the QMCPACK code we don't seem to ever explicitly set encoding we just use datatype H5T_C_S1 which can be either ASCII or UTF-8.

markdewing: The *.vp.h5 files from QMCPACK are written with the strings as "CST H5T_CST_ASCII".  This script tries to reproduce the format of the file exactly.
If I change the script to produce 'utf8', the loader test will segfault.

PDoakORNL: What about line 132 in this file?

Dealing with the loader segfaulting on utf-8 is beyond the scope of this PR.

PDoakORNL (changes requested):

markdewing:
About line 132:
I would prefer to leave it as decoding utf-8 because that means there would be fewer places to change in the future.  I assume any future fixes would move to save the HDF file strings as utf-8, once the C++ code was changed.

Background/context:
The strings read by h5py (`f["timestamp"][0]`) are of type `byte` (regardless of the CSET list in the HDF file). They must be decoded to be converted to a python string.  As long as all the characters are ASCII, decoding as utf-8 should be compatible.

PDoakORNL (approved):

PR #3687, opened 2021-12-23
Change check_python_reqs argument type.

ye-luo:
## Proposed changes
Found a bad call from https://github.com/QMCPACK/qmcpack/blob/9b2bef4605ec870003069048f66de0c70e9c8d74/tests/molecules/He_param/CMakeLists.txt#L30
This was caught on my laptop without h5py. 

check_python_reqs was error-prone for its counter-intuitive argument.
Previously it needs the variable name of the list of modules to be tested.
With this PR, it takes the list directly.

- Bugfix

- No

laptop

## Checklist


CMake/python.cmake
```diff
@@ -15,15 +15,18 @@ function(TEST_PYTHON_MODULE MODULE_NAME MODULE_PRESENT)
...
 
 # Test python module prerequisites for a particular test script
-#   module_list - input - list of module names
+#   module_list - input - list of module names, for example "numpy;h5py"
 #   test_name - input - name of test (used for missing module message)
 #                     - use empty string to silence output
 #   add_test - output - true if all modules are present, false otherwise
 function(CHECK_PYTHON_REQS module_list test_name add_test)
+  if(NOT ${ARGC} EQUAL 3)
+    message ("Mismathced number of arguments! Wrongly called with ${ARGC} arguments.")
```
markdewing, line 24: Message should say it's expecting 3 arguments. Maybe something like:
`Expecting 3 arguments in CHECK_PYTHON_REQS.  Was called with ${ARGC} arguments.`

ye-luo: Fixed

markdewing:
If someone still wants to use a variable to hold the list, it needs to be placed in quotes to work as expected.
For example:
```
set(afqmctools_python_reqs numpy;h5py;mpi4py;scipy;pyscf)
check_python_reqs("${afqmctools_python_reqs}" afqmc_workflow add_tests)
```

If the quotes are not present, the list gets expanded into the list of arguments
```
check_python_reqs(${afqmctools_python_reqs} afqmc_workflow add_tests)
```
is the same as
```
check_python_reqs (numpy h5py mpi4py scipy pyscf afqmc_workflow add_tests)
```
This incorrect use case is what the ARGC check will catch.  Or the case where the list doesn't have quotes. 
This should maybe be added to the error message.

CMake/python.cmake
```diff
@@ -15,15 +15,18 @@ function(TEST_PYTHON_MODULE MODULE_NAME MODULE_PRESENT)
 endfunction()
 
 # Test python module prerequisites for a particular test script
-#   module_list - input - list of module names
+#   module_list - input - list of module names, for example "numpy;h5py"
```
markdewing, line 18: Should state that the quotation marks are necessary, since it's easy to read them as setting apart the example.
```
module_list - input - list of modules names, for example "numpy;h5py" (including the quotation marks)
```

ye-luo: Added. Also corrected the "FATAL_ERROR" which was originally intended.

CMake/python.cmake
```diff
@@ -15,15 +15,18 @@ function(TEST_PYTHON_MODULE MODULE_NAME MODULE_PRESENT)
...
 
 # Test python module prerequisites for a particular test script
-#   module_list - input - list of module names
+#   module_list - input - list of module names, for example "numpy;h5py" (including the quotation marks)
 #   test_name - input - name of test (used for missing module message)
 #                     - use empty string to silence output
 #   add_test - output - true if all modules are present, false otherwise
 function(CHECK_PYTHON_REQS module_list test_name add_test)
+  if(NOT ${ARGC} EQUAL 3)
+    message(FATAL_ERROR "Expecting 3 arguments in CHECK_PYTHON_REQS. Was called with ${ARGC} arguments.")
```
markdewing, line 24:
```
Expecting 3 arguments in CHECK_PYTHON_REQS. Was called with ${ARGC} arguments.  Most likely quotation marks are missing from the first argument.
```

ye-luo: Updated.

ye-luo:
Test this please

markdewing (approved):
Thanks for making the message changes.  This feels like a sharp edge to CMake that is likely to trip up people.

prckent (approved):

ye-luo:
Test this please

PR #3694, opened 2021-12-29
Make hdf_archive support writting data from const object.

ye-luo:
## Proposed changes
In the hdf_archive abstraction, there was an issue that writing data from a const object caused compilation failure.
```
hdf_archive h5("abc.h5");
const int a = 0;
h5.write(a, "data"); // compilation failure.
```

It was causing by the design of h5data_proxy class which keeps a reference of the data object.
By removing this reference and passing the object via arguments, const data object can be written properly.
This fix is restrict to the internal detail of hdf_archive and there is no change to the library interface.

- New feature

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):

ye-luo:
Test this please

PR #3697, opened 2021-12-31
Refine RNG compilation treatment. Boost and Fake RNGs in a single library target.

ye-luo:
## Proposed changes
I'm experiencing some compiler bugs which affect RNG from boost. Since we incorporate it in BoostRandom.hpp which lands in many cpp files in QMCPACK, we have many compiled copies in object files. Changing compilation options of a cpp file which seems to unrelated to RNG ends up affecting the linked executable.
This PR uses extern template to restrict compilation to the BoostRandom.cpp and RandomGenerator.cpp files.

For the moment, USE_FAKE_RNG is still needed to select regular or fake RNG for `RandomGenerator_t` and `Random`. Taking it out will need additional work.

CMake target qmcrng ships both regular and fake RNG in a single library. This is a necessary step to  eventually remove USE_FAKE_RNG with twice compilation and divergent linking lines.

In addition.
Remove `Particle/WalkerSetRef.h` which is not used anywhere.
Take functions using global RNG out of `RandomSeqGenerator.h` into `RandomSeqGeneratorGlobal.h` to catch such code path to be refactored eventually.

- Build related changes. Internal

- No

epyc-server

## Checklist


PDoakORNL:
Overall this looks like an improvement to me.  I think dropping boost random should be strongly considered.  I believe all its functionality we use has been adopted into the c++ standard via <random> boost random appears to be unmaintained since 2010.

src/Utilities/RandomGenerator.h
```diff
@@ -30,118 +30,46 @@
...
 };
 
-extern RNGThreadSafe Random;
+extern template class RNGThreadSafe<FakeRandom>;
+extern template class RNGThreadSafe<BoostRandom<float>>;
+extern template class RNGThreadSafe<BoostRandom<double>>;
 
-} // namespace qmcplusplus
+#ifdef USE_FAKE_RNG
+using RandomGenerator_t = FakeRandom;
```
PDoakORNL, line 65: I'd really like to see this behavior by typedef dropped.  But that is Outside of this scope.

PDoakORNL (approved):
a step.

ye-luo:
Test this please

src/Utilities/RandomGenerator.h
ye-luo: Which line has typedef?

ye-luo:
> 

When I was caught by the compiler issue, unit tests passed but system tests failed because we are not testing the RNGs. If we moved to the standard library, correctness can be more safely assumed.

src/Utilities/RandomGenerator.h
PDoakORNL: 65, Well type alias, but basically it uses the c preprocessor + linking to a different definition of the type to fake a template class. In much new code I've just made the RandomGenerator a function argument and its type a function template parameter (I honestly prefer this). Then no USE_FAKE_RNG defines are needed.  The RNG shouldn't be state for anything other than the crowd/thread context.

I don't think anything here makes that harder, and we have many existing tests that depend on this type alias swap.

PDoakORNL (approved):
Reapproving after formatting

PR #3699, opened 2022-01-03
Add QMC_RNG_BOOST option to use RNG from C++ RNG from boost or C++ standard library.

ye-luo:
## Proposed changes
Make the connection to use RNG from C++ standard library.
It is a compile time selection QMC_RNG_BOOST.
OFF default value. uses C++ RNG from C++ standard library.
ON still uses boost.

To match boost behavior, I added a customized uniform distribution layer. See https://github.com/QMCPACK/qmcpack/pull/3699#issuecomment-1004370804

Note: There is no compatibility of random.h5 restart file between QMCPACK builds with different libraries(boost, libc++, libstdc++)

- New feature

- No

bora

## Checklist


prckent:
What problem does this solve?

Merging won't be possible until all the tests are working again imo. This is the challenge of updating the rng for any reason, so we should ensure that any updates strongly move us towards the long-term/ultimate desired rng situation.

ye-luo:
I hit deterministic test failure with a new Clang dev. When compared with an old Clang. The difference starts from the RNG which is from boost and we fully trust that it has been fully validated. But what if it has a bug? I tend to trust C++ standard library more than boost.

This PR can be merged as the default remains boost but allows now investigating standard library.

ye-luo:
After in-depth investigation, I found that std::uniform_real_distribution doesn't behave the same as boost::random::uniform_real_distribution.

boost::random::uniform_real_distribution always consume one integer random number when generating one return value regardless of the return float point type. std::uniform_real_distribution consider that as an issue. The double type offers higher resolution than 32 bit integer, as a result, it consumes two random numbers to create one return double. In the case float type return, only one integer random number is needed. See https://en.cppreference.com/w/cpp/numeric/random/generate_canonical.

I think choice made by C++ standard makes sense by itself. However, we would prefer matching the boost behavior to avoid updating all the tests and also avoid double/float discrepancy, I added uniform_real_distribution_as_boost to restore previous behavior.

@PDoakORNL I think this is the reason why you where forced to create two distinct set of reference data when StdRandom is used. With this PR, the complexity of reference data can be simplified. I also need  your help to make test_OneBodyDensityMatrices.cpp pass. I tried to update it but got a bit lost.

In principle, we now can switch on StdRandom as default now. My suggestion for the process will be.
1. @PDoakORNL update the reference data for that only unit test.
2. we merge this PR and make sure no problem show up in the nightly tests.
3. Swtich default and make sure no issue shows up in the night tests.
4. Delete BoostRandom.

prckent:
Faster proposal: make the c++ standard library route the default in this PR and remove the new cmake option. I think we have more than enough coverage in CI to thoroughly test this. I + anyone else interested can then also run the short+long tests. Then we merge and delete boostrandom in a subsequent PR.

ye-luo:
I have set std random as the default. The CMake option needs to be deleted together with the BoostRandom class.

prckent:
As expected, deterministic-unit_test_estimators needs a fix (@PDoakORNL ?). I assume this has its own random number path.

markdewing:
The type rename of `RandomGenerator_t` to `RandomGenerator` makes it difficult to find the substantive changes.  Changes that are trivial and pervasive should be in their own PR.

ye-luo:
Unfortunately, I have too many WIP branches at hand and I need to do the rename to fix a few compilation failures in Sandbox.
Please focus on files under `src/Utilities`

ye-luo:
I will make a separate PR to take the renaming out of the review.

ye-luo:
Renaming changes has been merged in develop and thus no more show diff here.

PDoakORNL:
@prckent the estimators tests have been updated and pass regardless of whether QMC_RNG_BOOST is true or false.

prckent:
@ye-luo What is your preferred route?

We can merge and watch the nightlies, but it is hard to spot changes.

I could run this PR or the merged version through the nightlies with/without the BOOST option.

If things work I would like to remove the QMC_RNG_BOOST option in the next PR.

ye-luo:
@prckent Merge it as it is now. If nothing bad happens, we remove QMC_RNG_BOOST and BoostRandom in the next PR.

ye-luo:
Test this please

ye-luo:
Test this please

prckent (approved):
LGTM

I am doing some dedicated runs with ~all the tests so that we can remove the boost option shortly.

ye-luo:
Test this please

PR #3708, opened 2022-01-05
Add debug disable branching in classic DMC driver

ye-luo:
## Proposed changes
Using `reconfiguration = runwhileincorrect ` can be risky when not running strictly 1 node 16 walkers. It reminds me #1817
debug_disable_branching is supported in the batched DMC driver. I think it is good to add the same option for the classic DMC driver and make performance tests not vulnerable to reconfiguration issues.

This option was added in the documentation section of both classic and batched DMC drivers.

- New feature

- No

epyc-server

## Checklist


PDoakORNL:
I think there is now some documentation/error messages in DMC.cpp that doesn't make sense.

ye-luo:
> I think there is now some documentation/error messages in DMC.cpp and DMCDriverInput.cpp that doesn't make sense.

I'm not getting this `reconfiguration = runwhileincorrect` option is still available.

PDoakORNL:
If I understand the point of this correctly.

```c++
  if (reconfig_str == "yes")
    throw std::runtime_error(
        "Reconfiguration is currently broken and gives incorrect results. Set reconfiguration=\"no\" or remove the "
        "reconfiguration option from the DMC input section. To run performance tests, please set reconfiguration to "
        "\"runwhileincorrect\" instead of \"yes\" to restore consistent behaviour.");
  reconfiguration_ = (reconfig_str == "yes");
```

This is not what we suggest to run performance tests.  Instead we suggest `debug_disable_branching=yes`

ye-luo:
> If I understand the point of this correctly.
> 
> ```c++
>   if (reconfig_str == "yes")
>     throw std::runtime_error(
>         "Reconfiguration is currently broken and gives incorrect results. Set reconfiguration=\"no\" or remove the "
>         "reconfiguration option from the DMC input section. To run performance tests, please set reconfiguration to "
>         "\"runwhileincorrect\" instead of \"yes\" to restore consistent behaviour.");
>   reconfiguration_ = (reconfig_str == "yes");
> ```
> 
> This is not what we suggest to run performance tests. Instead we suggest `debug_disable_branching=yes`

Thanks. Updated all the similar occurrence.

PDoakORNL (approved):
LGTM

prckent (approved):
Good idea. The intent in the inputs is clearer now and this is a definite improvement over runwhileincorrect.

prckent:
https://qmcpack.readthedocs.io/en/develop/methods.html#dmc-driver needs an update. Can wait if other updates will be soon.

ye-luo:
Test this please

Issue #3668, opened 2021-12-16
Commit plan for fast force derivatives.  


rcclay:
**Is your feature request related to a problem? Please describe.**
This is my commit plan for the fast force derivative feature.  This is largely targeting my own use cases (fast forces), but if others see straightforward opportunities to generalize this to other derivatives, let me know.  

**Describe the solution you'd like**
We're going to implement and review fast force derivatives in stages.  This is for single determinant no jastrow.  Jastrow capability will be added in a future PR.  

- [x] Commit the initial integration test.  
- [x] Implement new trial wave function wrapper, up to and including its initialization.  Enable relevant checks in unit test.
- [x] Implement B matrix (plus derivatives) computation into QMCHamiltonian and OperatorBase.  
- [x] Implement B matrix specializations for Kinetic energy and NLPP.  Enable asserts in unit test for correctness check.  
- [ ] Implement operator and derivative matrix math in QMCHamiltonian.  
- [ ] Swap current force operator evaluation pathway to the fast derivative path.  

**Describe alternatives you've considered**

**Additional context**

rcclay:
Here's a link to the paper that is being implemented: https://aip.scitation.org/doi/full/10.1063/1.4948778

PDoakORNL:
This plan see a good intial breakup.
A few things I think will make this more successful.

- Make clear distinctions between unit and integration tests as the plan progresses. Keep test cases focused and from getting to rambling.
- Actual unit tests should go in files whose name corresponds directly to the tested unit.
- Integration tests shouldn't be named with the unit pattern, `test_ion_derivs.cpp` is a good integration test file name.

rcclay:
Forgot to close this.

PR #3713, opened 2022-01-06
[Fast Derivatives] Stage 2:  Introduce TrialWavefunction Wrapper

rcclay:
## Proposed changes
In line with commit plan in Issue #3668, this implements a trial wave function wrapper to perform fast force evaluations.  For now, this pulls SPOSet data out of the determinants in TrialWavefunction and handles the determinant math and auxiliary matrix construction itself.  Jastrows will be incorporated after the full fast force capability works.  This is the most controversial part of the fast force plan, so feedback now is greatly appreciated.  

- New feature

- No

RHEL7 Linux server with Intel Xeons.

## Checklist


src/QMCWaveFunctions/WaveFunctionComponent.h
```diff
@@ -170,6 +170,11 @@ class WaveFunctionComponent : public QMCTraits
   /** print the state, e.g., optimizables */
   virtual void reportStatus(std::ostream& os) = 0;
 
+  /** Register the component with the TWFPrototype wrapper.  Pure virtual, so will throw
+   * an error at compile time
```
markdewing, line 174: Comment is incorrect. This function is not pure virtual, and will throw an error at runtime.

src/QMCWaveFunctions/TWFPrototype.h
```diff
@@ -0,0 +1,223 @@
...
+   * 
+   *  @param A.  Vector of matrices A.  
+   *  @param B.  Vector of matrices B.
+   *  @return Value of Sum_i Tr(A_i*B_i).
+   */
+  ValueType trAB(const std::vector<ValueMatrix_t>& A, const std::vector<ValueMatrix_t>& B);
+
+
+private:
+  std::vector<IndexType> num_ptcls;
```
PDoakORNL, line 210: Most member functions deduce most of these values from thier arguments.  Yet there is often a state copy as well, none seem "necessary" `spos_`, get rid of the rest.

src/QMCWaveFunctions/TWFPrototype.cpp
```diff
@@ -0,0 +1,316 @@
...
+#include "QMCWaveFunctions/TWFPrototype.h"
+#include "Numerics/DeterminantOperators.h"
+#include <iostream>
+namespace qmcplusplus
+{
+TWFPrototype::TWFPrototype() : initialized(false) { std::cout << "TWFPrototype Constructed\n"; }
+
+TWFPrototype::IndexType TWFPrototype::get_group_index(const IndexType gid)
+{
+  IndexType ndets = groups.size();
```
PDoakORNL, line 21: what is the source of truth as far an ndets go?  This code is all over the place on that.

src/QMCWaveFunctions/WaveFunctionComponent.h
rcclay: Good catch.  I experimented with something that ended up not working.

src/QMCWaveFunctions/TWFPrototype.cpp
```diff
@@ -0,0 +1,316 @@
...
+  IndexType nptcls = 0;
+  IndexType gid    = 0;
+  IndexType first  = 0;
+  IndexType last   = 0;
+  for (IndexType i = 0; i < ndets; i++)
+  {
+    gid     = groups[i];
+    first   = P.first(i);
+    last    = P.last(i);
+    nptcls  = last - first;
```
PDoakORNL, line 69: Also available as object state. although I agree you shouldn't use the state.

src/QMCWaveFunctions/TWFPrototype.h
PDoakORNL, line 210: This is new code it must follow the naming conventions for member variables.  This whole PR is harder to read because of this.

src/QMCWaveFunctions/TWFPrototype.h
```diff
@@ -0,0 +1,223 @@
...
+ *  SPOSet objects.  This is so that observables can be recast in matrix form and their derivatives taken efficiently. 
+ *  Currently this is hard coded for ground state slater jastrow wave functions, but generalization to
+ *  arbitrary occupations and multideterminants are straightforward and will come online as the code is tested and validated. 
+ *
+ *  Please see : J. Chem. Phys. 144, 194105 (2016) https://doi.org/10.1063/1.4948778 for implementation details and formalism.
+ */
+class TWFPrototype
+{
+public:
+  using ValueMatrix_t = SPOSet::ValueMatrix_t;
```
PDoakORNL, line 36: _t is redundant.

src/QMCWaveFunctions/TWFPrototype.h
```diff
@@ -0,0 +1,223 @@
...
+  using GradMatrix_t  = SPOSet::GradMatrix_t;
+  using HessMatrix_t  = SPOSet::HessMatrix_t;
+  using IndexType     = QMCTraits::IndexType;
+  using RealType      = QMCTraits::RealType;
```
PDoakORNL, line 40: xxxType is redundant.

src/QMCWaveFunctions/TWFPrototype.h
```diff
@@ -0,0 +1,223 @@
...
+  using ValueType     = QMCTraits::ValueType;
+
+  using ValueVector_t = SPOSet::ValueVector_t;
+  using GradVector_t  = SPOSet::GradVector_t;
+
+  TWFPrototype();
+  void add_determinant(const ParticleSet& P, const IndexType groupid, SPOSet* spo);
+  inline void add_jastrow(WaveFunctionComponent* j) { J.push_back(j); };
```
PDoakORNL, line 48: This is new code it must follow the coding conventions for naming methods.

src/QMCWaveFunctions/TWFPrototype.h
```diff
@@ -0,0 +1,223 @@
...
+
+
+private:
+  std::vector<IndexType> num_ptcls;
+  std::vector<IndexType> num_orbs;
+  std::vector<SPOSet*> spos;
+  std::vector<IndexType> groups;
+  std::vector<ValueMatrix_t> psiM;
+  std::vector<ValueMatrix_t> psiMinv;
+  std::vector<WaveFunctionComponent*> J;
```
PDoakORNL, line 216: No more 1 letter variable names.  maybe these are going to be `jastrows`?

src/QMCWaveFunctions/TWFPrototype.cpp
```diff
@@ -0,0 +1,316 @@
...
+#include <iostream>
+namespace qmcplusplus
+{
+TWFPrototype::TWFPrototype() : initialized(false) { std::cout << "TWFPrototype Constructed\n"; }
+
+TWFPrototype::IndexType TWFPrototype::get_group_index(const IndexType gid)
+{
+  IndexType ndets = groups.size();
+  for (IndexType i = 0; i < ndets; i++)
+    if (gid == groups[i])
```
PDoakORNL, line 23: So since you wrote no docs it took awhile to figure out that since addDetermimant might be called out of group order you need this TWFP index to groups id.  Seems like it would be better to just construct the object with a list of determinants sorted in groupid order and reduce internal state.

src/QMCWaveFunctions/TWFPrototype.cpp
```diff
@@ -0,0 +1,316 @@
...
+  IndexType first  = 0;
+  IndexType last   = 0;
+  for (IndexType i = 0; i < ndets; i++)
+  {
+    gid     = groups[i];
+    first   = P.first(i);
+    last    = P.last(i);
+    nptcls  = last - first;
+    norbs   = spos[i]->getOrbitalSetSize();
+    mvec[i] = 0;
```
PDoakORNL, line 98: I thought these got overwritten, why zero them?

PDoakORNL (changes requested):
On top of the specific comments I have, I would repeat please write documentation and follow the naming conventions.

rcclay:
@PDoakORNL Thank you for the feedback.  I dropped the ball on the coding conventions, but thanks for enforcing them.  I'll also pay a lot more attention to minimizing state information and tightening up some of the design decisions.  I'll start seriously addressing these on Monday.

src/QMCWaveFunctions/TWFPrototype.cpp
rcclay: What's current best practices for documentation?  Are you talking about more ample source code comments?  A write up in the manual?  Something else?

lshulen: I thought to point to the manual (https://qmcpack.readthedocs.io/en/develop/developing.html#comments) on comments here since a response was not forthcoming.  Upon looking at the code it seems like there is extensive documentation in the header file for TWFPrototype for both the class and the functions.  The only thing that seemed missing to me was labeling parameters [in,out], but Peter seems to be suggesting something broader.  @PDoakORNL , what are you looking for?

PDoakORNL (commented):
some clarification with regards to docs.

But I should say to me its more important that:

For this PR
- [ ] Code is consistent about internal state vs. function arguments as the source of truth. (I obviously favor function arguments and less internal state)
- [ ] All new code follows current coding conventions unless overriding legacy virtual functions.

For future PRs perhaps not just yours.
- [ ] All the indexes for species, group, and det are problematic, SpeciesSet isn't great but it at least provides some clarity on this. Species id, and group id are rather confused in ParticleSet as well and groupid there certainly doesn't have a unique mapping across particle sets with regards to species id.

src/QMCHamiltonians/tests/test_ion_derivs.cpp
```diff
@@ -769,9 +770,9 @@ TEST_CASE("Eloc_Derivatives:proto_sd_noj", "[hamiltonian]")
   REQUIRE(psi != nullptr);
   //end incantation
 
-//  TWFPrototype twf;
+  TWFPrototype twf;
 
-//  psi->initialize_TWF_Prototype(elec, twf);
+  psi->initialize_TWF_Prototype(elec, twf);
```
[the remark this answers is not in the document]
rcclay: I do like the idea of a factory, but I chose to do it this way for two reasons.  1.) This object is designed to be completely dependent on a properly constructed TrialWaveFunction object.  TrialWaveFunction always exists, whereas this is a wrapper that is constructed if force derivatives need to be computed.  2.) If I don't want to rebuild the whole wave function from scratch, I need some private TrialWaveFunction tags to determine if things are Jastrows or Determinants.  Having psi know how to construct it's own fancy wrapper seemed to be the simpler solution for me at the time.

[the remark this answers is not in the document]
ye-luo: As @rcclay has mentioned, this is a wrapper intended to be used only for the force derivatives codes he is pushing upstream. So this will be contained instead of being part of the WaveFunctionFactory if it is the factory you mean. We will explore routes without the need of wrapper but this what we have now.

src/QMCWaveFunctions/WaveFunctionComponent.h
```diff
@@ -170,6 +170,11 @@ class WaveFunctionComponent : public QMCTraits
...
+   */
+  virtual void register_TWF_Prototype(ParticleSet& P, TWFPrototype& twf);
```
ye-luo, line 176: If register_TWF_Prototype is to query existing data form ParticleSet and TWF. It is better to have P const and this function const. Maybe const is not eligible on the function if register_TWF_Prototype does initialize internal data structures. I meant use const whenever feasible.

src/QMCWaveFunctions/TrialWaveFunction.h
```diff
@@ -517,6 +521,8 @@ class TrialWaveFunction
   ///a list of WaveFunctionComponents constituting many-body wave functions
   std::vector<std::unique_ptr<WaveFunctionComponent>> Z;
 
+  /// For now, TrialWaveFunction will own the wrapper.
+  TWFPrototype twf_prototype;
```
ye-luo, line 525: Your test case shows it is owned outside TWF. Do we still need this?

src/QMCWaveFunctions/TWFPrototype.h
```diff
@@ -0,0 +1,223 @@
...
+  //This takes a particle set group index, and returns the "determinant" this refers to.
+  IndexType get_group_index(const IndexType gid);
+  SPOSet* get_sposet(const IndexType sid) { return spos[sid]; };
+
+  /** @brief Returns log(Psi).  Should be consistent with QMCPACK unwrapped TrialWavefunction.
+   *
+   *  @param P. Particle set.
+   *  @return log(Psi).
+   */
+  RealType evaluateLog(ParticleSet& P);
```
ye-luo, line 66: If possible, it is better to skip this level of indirection.

src/QMCWaveFunctions/WaveFunctionComponent.h
rcclay: Done.

src/QMCWaveFunctions/TWFPrototype.h
```diff
@@ -0,0 +1,223 @@
...
+  using GradMatrix_t  = SPOSet::GradMatrix_t;
+  using HessMatrix_t  = SPOSet::HessMatrix_t;
+  using IndexType     = QMCTraits::IndexType;
+  using RealType      = QMCTraits::RealType;
+  using ValueType     = QMCTraits::ValueType;
+
+  using ValueVector_t = SPOSet::ValueVector_t;
+  using GradVector_t  = SPOSet::GradVector_t;
+
+  TWFPrototype();
```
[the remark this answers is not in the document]
rcclay: Dropped the initialize flag, and now rely on default constructor.

src/QMCWaveFunctions/TWFPrototype.cpp
```diff
@@ -0,0 +1,316 @@
...
+  {
+    groups.push_back(gid);
+    spos.push_back(spo);
+    IndexType first = P.first(gid);
+    IndexType last  = P.last(gid);
+    IndexType norbs = spo->getOrbitalSetSize();
+    num_orbs.push_back(norbs);
+    num_ptcls.push_back(last - first);
+  }
+  initialized = true;
```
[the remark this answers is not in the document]
rcclay: Fair point.  The logic was bad, the flag was bad.  It has been removed, and should be sane.

src/QMCWaveFunctions/TWFPrototype.h
rcclay: Got rid of the num_orbs and num_ptcls lists.  The first will be inferred from the SPOSet, the latter from either the ParticleSet or from the input matrix dimensions for matrix math convenience functions.  Get functions referencing these have been removed.

src/QMCWaveFunctions/TWFPrototype.h
rcclay: Done.

src/QMCWaveFunctions/TWFPrototype.h
rcclay: Now it's jastrow_list_.

src/QMCWaveFunctions/TWFPrototype.h
```diff
@@ -0,0 +1,223 @@
...
+//
+// Copyright (c) 2021 QMCPACK developers.
+//
+// File developed by:   Raymond Clay III, rclay@sandia.gov, Sandia National Laboratories
+//
+// File created by:   Raymond Clay III, rclay@sandia.gov, Sandia National Laboratories
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+/**@file TWFPrototype.h
```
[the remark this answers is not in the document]
rcclay: Beefed up the documentation massively and am now using standard Doxygen group blocks.

src/QMCWaveFunctions/TWFPrototype.h
rcclay: While I agree that it is redundant, unless you feel strongly about this, I would prefer to keep it the way it is out of consistency, since these typedefs are used widely in the various numerics parts of the code.

src/QMCWaveFunctions/TWFPrototype.cpp
rcclay: I did a lot of cleaning and renaming here.  "Determinants" is sloppy, and has been renamed to "groups".  Groups are collections of particles which for this TWF object are antisymmetric under particle exchange, hence why the naming got sloppy.  Hopefully this is fixed and is more consistent now.

src/QMCWaveFunctions/TWFPrototype.h
rcclay: I agree.  This will be removed.

src/QMCWaveFunctions/TrialWaveFunction.h
rcclay: I had always envisioned this as being owned by TrialWaveFunction, but based on this example...  what do you think of the following:  What if the force observable owns the wrapper for now?  That would make initialization unambiguous...  it would be initialized by the object that needs it.

ye-luo: I forgot if I have mentioned this in any comments. TWFPrototype is still not very informative.
TWFStructureMetaData can be better. We can discuss the name here but any renaming can happen after all the pieces land.

ye-luo: > I had always envisioned this as being owned by TrialWaveFunction, but based on this example... what do you think of the following: What if the force observable owns the wrapper for now? That would make initialization unambiguous... it would be initialized by the object that needs it.

If calls are only from force estimator, the force estimator can own it and call the initialization in its constructor.

rcclay: You had not, but I agree...  TWFPrototype was a placeholder name.  TWFStructureMetaData can be fine, TWFDerivWrapper might be descriptive.  I'm totally open to suggestions.  Should just be a sed command once we decide.  I'd probably kick the can down the road until we have a complete working code, but I don't feel particularly strongly one way or the other.

rcclay: > 

We'll do it that way for now then.  If it looks like this can/will be generalized to other derivatives, then we can change the ownership/scope then.

src/QMCWaveFunctions/TWFPrototype.h
rcclay: Done.

src/QMCWaveFunctions/WaveFunctionComponent.h
rcclay: Done.

rcclay:
OK.  I made a good faith effort to address all the points raised.  @PDoakORNL @ye-luo , how does it look now?

ye-luo:
Test this please

ye-luo (approved):

rcclay:
I realize I can increase the coverage a bit if I uncomment some TWFPrototype calls in test_ion_derivatives.  I'll go ahead and do that.  Otherwise from a code coverage perspective, the complicated mixed derivative SPO calls will be hit in a future PR.

src/QMCWaveFunctions/TWFPrototype.cpp
ye-luo: I think Peter means moving the definition inside the for loop. They are all scalars. So performance wise, it is better inside the loop. The code also gets more concise.

rcclay: Done.

prckent (dismissed):

ye-luo (dismissed):

src/QMCWaveFunctions/TWFPrototype.h
```diff
@@ -0,0 +1,229 @@
...
+{
+/**
+ *  TWFPrototype is a wrapper class for TrialWavefunction that provides separate and low level access to the Jastrow and 
+ *  SPOSet objects.  This is so that observables can be recast in matrix form and their derivatives taken efficiently. 
+ *  Currently this is hard coded for ground state slater jastrow wave functions, but generalization to
+ *  arbitrary occupations and multideterminants are straightforward and will come online as the code is tested and validated. 
+ *
+ *  Please see : J. Chem. Phys. 144, 194105 (2016) https://doi.org/10.1063/1.4948778 for implementation details and formalism.
+ */
+class TWFPrototype
```
PDoakORNL, line 30: WaveFunctionForceWrapper?
TWFPrototype especially when you name an instance of it twf (used elsewhere in the code for trial wave function) isn't helping.
If we are just refering to the stage of development ForcesPrototype would be better.

PDoakORNL (changes requested):
Its getting closer.  I think its worth getting some of the naming fixed sooner than later.

src/QMCHamiltonians/tests/test_ion_derivs.cpp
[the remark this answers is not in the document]
ye-luo: `getOrbitals()` should not be widely used except for testing purpose. Don't assume the internal structure of TWF. For this reason, my choice will be the current code.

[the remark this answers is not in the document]
PDoakORNL: Having the more basic type TrialWaveFunction have knowledge of extending type  is not scalable. TrialWaveFunction is a compute interface not an initializer for other objects. This register call is properly part of how a TWFPrototype becomes valid, it should be with the implementation of TWFPrototype.

It should preferably be called from the constructor of TWFPrototype.
I can see there is a similar issue pending for the register_TWF_prototype call on the WFC but I'm waiting to see what that looks like.  

The important news here is that the actual data dependence that TWFPrototype has is on the fermionic wavefunction components it shares with TWF it makes sense for that to be expressed in the code for TWFPrototype.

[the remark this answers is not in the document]
ye-luo: Since WFC are virtual, its iteration needs to be implemented on the WFC side and hence `register_TWF_Prototype()` calls.

> It should preferably be called from the constructor of TWFPrototype.

Agree
```
TWFPrototype twf_proto(twf);
```
is nicer than directly calling initialize_TWF_Prototype.

PDoakORNL:
Convince me that dropping the cyclic dependency between TWFPrototype, TWF, and WFC and having a constructor in the usual way for TWFPrototype isn't better.

Meaning
```c++
class TWFPrototype
{
  TWFPrototype(ParticleSet& pset, RefVector<WaveFunctionComponent>);
};
```
If `is_jastrow` and `is_determinant` flags were added to WFC interface and How to deal with the nested SlaterDet was resolved (worth doing).  The parts of the TWFPrototype construction that is hidden in the WFCs could be moved into the actual TWFPrototype constructor as well

```c++
// Now when we read the constructor we can see what is required 
// and what must be done with it to create a valid TWFProtoType
TWFProtoType::TWFPrototype(ParticleSet& pset, RefVector<WaveFunctionComponent> wfcs)
{
  for( WaveFunctionComponent& wfc : wfcs)
  {
     if (wfc.is_jastrow)
        addJastrow(wfc);
     else if (wfc.is_determinant)
        addGroup(pset.something, pset.somethingelse, wfc.getPhi());
        //yes getPhi() isn't part of the WFC interface but adding it seems less problematic
        //Than a type that is actually dependent on WFC having interfaces dependent on it added
  };
}
```

The alternative is just to override my review.  I'm not part of a consensus that this is a good way to proceed.

PDoakORNL (approved):

ye-luo (approved):

ye-luo:
Test this please

Issue #3696, opened 2021-12-30
Building with -DQMC_OMP=OFF fails.


NAThompson:
**Describe the bug**

Disabling OMP does not work.

**To Reproduce**

On top of git hash 174d200a04bd2564564e1cc5becff:

```
~/q/build ❯❯❯ cmake -DQMC_OMP=OFF ../ -G Ninja
-- The C compiler identification is AppleClang 12.0.5.12050022
-- The CXX compiler identification is AppleClang 12.0.5.12050022
~/q/build ❯❯❯ ninja
[16/713] Building CXX object src/QMCTools/CMakeFiles/qmc-check-affinity.dir/check-affinity.cpp.o
FAILED: src/QMCTools/CMakeFiles/qmc-check-affinity.dir/check-affinity.cpp.o
~/qmcpack/src/QMCTools/check-affinity.cpp:15:10: fatal error: 'omp.h' file not found
#include <omp.h>
         ^~~~~~~
1 error generated.
```

**System:**
 - M1 Mac.

Patch for the fix is attached.

[patch.txt](https://github.com/QMCPACK/qmcpack/files/7794841/patch.txt)

NAThompson:
Sadly, now I'm here:

```
qmcpack/external_codes/boost_multi/multi/./memory/allocator.hpp:64:3: error: static_assert failed due to requirement 'alignof(long double) >= 16' "for cuda"
                static_assert( alignof(std::max_align_t) >= 16 , "for cuda");
                ^              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
ninja: build stopped: subcommand failed.
```

prckent:
Hi NAThompson -

Which compilers and tool versions are you using to build QMCPACK? And which version are you compiling? I verified that QMCPACK develop built on macOS a few weeks ago using either a very recent gcc or clang (on travel, can't check). Generally people use a set of software from macports or brew - there are instructions in the manual for this, e.g. https://qmcpack.readthedocs.io/en/develop/installation.html#installing-on-mac-os-x-using-macports . Our CMakeLists will pick up Apple's veclib for blas. We haven't checked recently whether QMCPACK builds with OpenMP disabled - this could be fixed if needed for strong enough reasons. At most I would expect that there is a missed #ifdef or two needed to protect omp includes as you found.

NAThompson:
[...]

PR #3723, opened 2022-01-11
Enable compilation and tests with QMC_OMP=0 and remove ENABLE_OPENMP definition

williamfgc:
## Proposed changes

This PR solves build conflicts encountered when disabling OpenMP `ENABLE_OPENMP=0`
~~Skip~~ Enable `check-affinity` test
Resolve mpi wrapper conflicts between `formic/utils/openmp.h` and `Message/OpenMP.h` (~~TODO: unify them~~ unified under `Concurrency/OpenMP.h`)
There are still tests that don't pass locally (will open an issue). 
Current CI doesn't cover no-OpenMP configurations
Related #3696 

- Bugfix
- Build related changes

- No

Ubuntu 20.04. must pass CI

## Checklist


ye-luo:
I think it is better to have "Message/OpenMP.h" to handle switching between with and without OpenMP. All other places include it. In such way, we keep only 1 entry of "omp.h"

src/Concurrency/Info.hpp
```diff
@@ -51,10 +58,13 @@ unsigned int getWorkerId();
 template<>
 inline unsigned int getWorkerId<Executor::OPENMP>()
 {
+#ifdef ENABLE_OPENMP
   return omp_get_thread_num();
+#else
+  return 1;
```
markdewing, line 64: I think this should return zero, since in OMP the threads are numbered from 0 to omp_get_num_threads()-1.

prckent: Agreed.

prckent (commented):
I agree with the previous comments.

Generally LGTM. I was expecting that updated cmake run test macros would be needed but perhaps the code was already future proof. With OpenMP disabled but MPI enabled, only the "16-1" and "1-1" tests are appropriate, which is not very many. And if both are disabled the testset will be rather small.

src/Concurrency/Info.hpp
williamfgc: Did I mix it with `omp_get_num_threads()`? I'll fix.

williamfgc:
> With OpenMP disabled but MPI enabled, only the "16-1" and "1-1" tests are appropriate, which is not very many.

Thanks @prckent , I might as well update the cmake test configuration 'cause other tests are failing.

prckent:
Incompatible tests should automagically not run. Likely some make of our cmake needs to know more about the max available threads (=1 with no omp). When MPI is not available, with the current configuration, the MPI dependent tests are correctly not run.

PDoakORNL:
The way the cmake is currently set up it, I think QMC_OMP=0 is the correct way to turn off omp.  Also we have/had a working parallel executor for std thread which should be what should get used in that case, when we initially made that abstraction crowd level parallelism worked find with std::thread in tests.

src/Message/OpenMP.h
```diff
@@ -21,12 +21,12 @@
 #if defined(ENABLE_OPENMP)
 #include <omp.h>
 #else
+#include "formic/utils/openmp.h"
+
```
PDoakORNL, line 25: This file belongs in concurrency and looks to add only open mp functions none of them should be defined if QMC_OMP=0 since none of them should be called.

prckent:
It looks like a documentation update is needed since neither QMC_OMP or ENABLE_OPENMP are mentioned. I can't think of a good reason to have two different options. Adding an ERROR in cmake if the "wrong" one is specified could be done immediately.

ye-luo:
We should remove `ENABLE_OPENMP` and only keep `QMC_OMP`.

williamfgc:
Just getting back to this. Glad to see it needs more discussion.

1. `QMC_OMP` is the CMake option, but `ENABLE_OPEMP` is defined in the generated config.h file. It is not a bad practice to separate those two as the latter propagates to source code, but if there is an existing pattern I will follow.
2. Agree with @ye-luo , only have a single OpenMP wrapper to unify calls to OpenMP and avoid propagating #ifdef through the codebase.
3. Larger topic, are OpenMP and C++ Std threads meant to be mutually exclusive in QMCPACK? Threads are part of the standard library, it shouldn't be treated as an external dependency.

prckent:
Lets not worry about std threads here :-) 

Once the new drivers are more the default and there are more options to do a fixed amount of work, we can revisit.

ye-luo:
I guess from the very beginning. QMC_OMP is cmake request. If checks pass, ENABLE_OMP is set and passed via the config.h.
However, I found this not necessary. If QMC_OMP is requested but OpenMP but failed some basic checks, we should just stop instead of setting ENABLE_OPENMP=0 and continue. Then there is no need of ENABLE_OPENMP being different from QMC_OMP.

PDoakORNL:
As far as I know openmp considers itself to be a superior alternative to std::thread and guarantees no interoperability with it at all, so they must be mutually exclusive.

PDoakORNL (approved):

PDoakORNL:
I will just go back over the new driver tests at a later time to enable std::thread based ParallelExecutors.

williamfgc:
@ye-luo so switching `#cmakedefine ENABLE_OPENMP @ENABLE_OPENMP@` to `#cmakedefine QMC_OMP @QMC_OMP@` in `config.h.in` is what we're looking for? and fix includes where appropriate.

@PDoakORNL  sounds good, thanks.

ye-luo:
clean up and replace ENABLE_OPENMP with QMC_OMP in CMake source code.
Consolidate `Message/OpenMP.h` and use it as the entry for other places need OpenMP runtime library calls.

williamfgc:
@ye-luo please take a look, I changed the PR description. @prckent I excluded the `diamondC_2x1x1` tests that are failing locally. We can always add CI for this scenario, thoughts?

ye-luo:
1. Now I remember, once compiler option `-fopenmp` is on. `_OPENMP` is defined. This macro is portable across all OpenMP compilers. So in QMCPACK source doe, there is no need to define `QMC_OMP` for the source code. Use `_OPENMP` and `-fopenmp` will be the only control.
2. I think it is time to move OpenMP.h under src/Concurrency and introduce a CMake target "concurrency". Since it is header only, so it will be an interface library.
3. The ParallelExecutor  src/formic/utils/openmp.h and  should just include OpenMP.h.
4. 
Let me know if you need help on the CMake side.

prckent:
Better to have a max test threads setting if implementation is sensible, and filter on that? Should be doable in cmake. We really ought to have tests with "strange" numbers of threads and mpi to stress the logic more. Doesn't need to be done here though.

williamfgc:
@ye-luo @prckent please see the latest changes.

1. Yes, `_OPENMP` or anything standard should be offloaded to backends. Good one!
2. `OpenMP.h` is now under `src/Concurrency`, the CMake interface can be part of a separate PR? Right now it's a simple header include, it seems like using an interface involves more discussion as there are other existing files.
3. Done
4. Please review the proposed filtering for threads > 1. Not sure if multithreaded cases should be run when disabling OpenMP. I run things locally and all tests pass (CI is still pending, but I don't know if it's worth it).

src/Concurrency/Info.hpp
```diff
@@ -17,7 +17,10 @@
  *  @brief Abstraction of information on executor environments
  */
 
+#ifdef _OPENMP
 #include <omp.h>
```
ye-luo, line 21: Use
```
#include "OpenMP.h"
```
since  the file is in the same folder now.

src/QMCTools/CMakeLists.txt
```diff
@@ -31,9 +31,11 @@ target_link_libraries(qmc-extract-eshdf-kvectors PUBLIC qmcio)
 
 add_executable(qmc-get-supercell getSupercell.cpp)
 
-add_executable(qmc-check-affinity check-affinity.cpp)
-if(HAVE_MPI)
-  target_link_libraries(qmc-check-affinity MPI::MPI_CXX)
+if(QMC_OMP)
+  add_executable(qmc-check-affinity check-affinity.cpp)
```
ye-luo, line 35: Can we fix check-affinity.cpp?

src/QMCTools/tests/CMakeLists.txt
```diff
@@ -31,26 +31,28 @@ add_unit_test(${UTEST_NAME} 1 1 $<TARGET_FILE:${UTEST_EXE}>)
 set_tests_properties(${UTEST_NAME} PROPERTIES WORKING_DIRECTORY ${UTEST_DIR})
 
 # Minimal test for qmc-check-affinity. Check for "OpenMP" in output, no errorcodes.
-if(HAVE_MPI)
-  add_test(NAME check-affinity_runs COMMAND ${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} 1 ${MPIEXEC_PREFLAGS}
-                                            $<TARGET_FILE:qmc-check-affinity>)
-else()
-  add_test(NAME check-affinity_runs COMMAND $<TARGET_FILE:qmc-check-affinity>)
+if(QMC_OMP)
+  if(HAVE_MPI)
```
ye-luo, line 35: And restore the test.

ye-luo (commented):
CMake target. I agree we can leave it out.
Tests Nthreads>1. Some default value (number of walkers) depends on the number of threads. For this reason, just disable them for now. Batched drivers are not affected by this but running the workload of 16 threads on 1 core is anyway painful. So fine to disable them for now. We can revisit after.

src/Concurrency/Info.hpp
williamfgc: Good one, I cleaned some redundant lines in that file as well.

src/QMCTools/tests/CMakeLists.txt
williamfgc: Done, it's running locally but let's see what CI says.

src/QMCTools/CMakeLists.txt
```diff
@@ -51,9 +53,15 @@ if(USE_OBJECT_TARGET)
   target_link_libraries(qmcfinitesize qmcparticle qmcutil)
 endif()
 
-install(TARGETS convert4qmc qmc-extract-eshdf-kvectors qmc-get-supercell qmc-check-affinity convertpw4qmc qmcfinitesize
+install(TARGETS convert4qmc qmc-extract-eshdf-kvectors qmc-get-supercell convertpw4qmc qmcfinitesize
```
ye-luo, line 56: restore here.

williamfgc: My bad, target install should not depend on `QMC_OMP`. Thanks!

ye-luo (approved):
LGTM.

ye-luo:
Test this please

PR #3731, opened 2022-01-14
Update documentation around wavefunction

ye-luo:
## Proposed changes
I revisited the current documentation around wavefunction for derivatives for optimization.
I have two commits.
The first one updates the actual documentation content.
The second one recorder contests. backflow and estimator docs are moved after wavefunction. It causes massive diff unfortunately.
So please review on the first commit.

In addition, I caught a bug that the mass is not applied properly when taking kinetic energy derivatives and particles don't have equal mass.

- Documentation changes

- No

ryzen-box

## Checklist


docs/developing.rst
```diff
@@ -2149,12 +2149,21 @@ An example of the second approach is
...
-``WaveFunctionComponent``. The code contains an example of a
+A full ``TrialWaveFunction`` is formulated as a product
+of all the components. Each component derives from ``WaveFunctionComponent``.
+
+.. math::
+     \psi = \prod_c {\tilde \psi_c}
+
+QMCPACK doesn't directly use the product form but mostly use the log of the wavefunction which
+fits the formulation of QMC algorithms very well. It also offers a numerical advantage on computers.
+The log value grows linearly instead of exponentioally with respect to the electron counts
```
markdewing: The advantage of using the log vs. the value should be that the log doesn't exceed the range of double precision, where the value potentially could (??)  I'm not sure scaling wrt electron count matters, except that larger systems are more likely to suffer overflow/underflow on the value.

ye-luo: by scaling I meant overflow/underflow on the value. Let me update.

ye-luo: Fixed.

ye-luo:
Test this please

prckent:
From a casual reading this LGTM. Thanks @markdewing for the spelling corrections and typo fixes. Due to the understandable size of the diff I propose that we leave this open until late today to allow others to comment, then merge.

prckent (approved):

prckent:
Start testing in-house

PR #3737, opened 2022-01-19
Move the ownership of lattice to ParticleSetPool

ye-luo:
## Proposed changes
Before this change, the are many copies lattice objects (cell vectors and other meta data) around the code.
In a simulation ParticleSetPool owns the initial copy by parsing the input XML, then all the ParticleSet makes a copy.
There are other additional copies around.

This PR introduces SimulationCell class which contains a few cell related objects which were previously directly defined in ParticleSet. The SimulationCell owned by ParticleSetPool is now the single source of truth. All the ParticleSets own a const reference of SimulationCell which is checked in at the constructor of ParticleSet.

There are many benefits of having one SimulationCell and all the others are references. 
1. No need to track if the data is up-to-date. More friendly to future structural optimization
2. Lattice accessed via ParticleSet becomes const (read-only).

- New feature

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):
Q. Did you verify that everything in sandbox is OK? We don't have complete CI coverage there.

Very happy to see this cleanup in the design. To be consistent with my comments on other PRs, even with these changes that affect so many files, we should have sufficient coverage in the CI.

prckent:
Start testing in-house

ye-luo:
> Q. Did you verify that everything in sandbox is OK? We don't have complete CI coverage there.

Sanbox is fixed accordingly. I still would like to have one build on CI. So put up https://github.com/QMCPACK/qmcpack/issues/3743

Issue #3712, opened 2022-01-06
Commit plan for batched spin-orbit code


camelto2:
**Is your feature request related to a problem? Please describe.**
Below is a checklist to help me keep track of what is done for the SO capability in the batched code. This helps close #3508 

**Describe the solution you'd like**

- [x] add spin moves in multiwalker APIs (#3710, #3728)
- [x] add spin gradients (e.g. evalGradWithSpin APIs) to TWF and TWFdispatcher for mw evaluation (#3714, #3887)
- [x] add spin gradients (e.g. evalGradWithSpin APIs) to  DiracDeterminant for mw evaluation (#3887, #4182, #4208, #4269 )
- [x] add mw evaluation to SpinorSet (#3726, #3730, #4208, #4226, #4231)
- [x] update VMC/DMC batched drivers to include spin moves (#3774, #3762, #3867)
- [x] update SOECPComponent and SOECPotential to include mw evaluations(#4375, #4504)

**Describe alternatives you've considered**

**Additional context**

prckent:
Please can you update the status? What remains?

camelto2:
In terms of running calculations, everything is technically there for single determinants and users should be able to run. The only caveat is that for SOECPotential, they fall back to single walker APIs from the definition in operator base. I'm working on the actual mw_ APIs for SOECPotential now, which should be finished shortly. After that, the only missing piece is MSD

camelto2:
closing this and moving the mw_ APIs needed for MSD to its own issue. Batched Spin-Orbit is finished for our standard workflow with single slater jastrow wave functions

PR #3742, opened 2022-01-20
[WIP] Batched drivers with spin moves

camelto2:

## Proposed changes
This PR adds the functionality needed to perform spin moves in the batched drivers, needed for SOC. Essentially, I added a new advanceWalkersWithSpin which is a copy of advanceWalkers and adds the required spin moves and WF calls. The selection of which advanceWalkers call is used is determined at the runQMCstep level after checking if the particleset leader is a spinor or not. Adding this new function was much cleaner than adding a bunch of if statements everywhere, and until we have a cleaner templated solution this seemed like the easiest path forward. Some documentation is added to the batched drivers since the spin_mass variable is added to QMCDriverInput

Part of #3712 

- New feature

- No

macOSX

## Checklist


src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -215,6 +215,216 @@ void VMCBatched::advanceWalkers(const StateForThread& sft,
...
+
+  timers.movepbyp_timer.start();
+  const int num_walkers = crowd.size();
+  // Note std::vector<bool> is not like the rest of stl.
+  std::vector<bool> moved(num_walkers, false);
+  constexpr RealType mhalf(-0.5);
+  const bool use_drift = sft.vmcdrv_input.get_use_drift();
+  std::vector<TrialWaveFunction::GradType> grads_now(num_walkers);
+  std::vector<TrialWaveFunction::GradType> grads_new(num_walkers);
+  std::vector<TrialWaveFunction::ComplexType> spingrads_now(num_walkers);
```
ye-luo, line 251: Divergent

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -215,6 +215,216 @@ void VMCBatched::advanceWalkers(const StateForThread& sft,
...
+  // local list to handle accept/reject
+  std::vector<bool> isAccepted;
+  std::vector<std::reference_wrapper<TrialWaveFunction>> twf_accept_list, twf_reject_list;
+  isAccepted.reserve(num_walkers);
+
+  for (int sub_step = 0; sub_step < sft.qmcdrv_input.get_sub_steps(); sub_step++)
+  {
+    //This generates an entire steps worth of deltas.
+    step_context.nextDeltaRs(num_walkers * sft.population.get_num_particles());
+    step_context.nextDeltaSpins(num_walkers * sft.population.get_num_particles());
```
ye-luo, line 270: Divergent

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -215,6 +215,216 @@ void VMCBatched::advanceWalkers(const StateForThread& sft,
...
+        auto delta_r_start    = step_context.deltaRsBegin() + iat * num_walkers;
+        auto delta_r_end      = delta_r_start + num_walkers;
+        auto delta_spin_start = step_context.deltaSpinsBegin() + iat * num_walkers;
+        auto delta_spin_end   = delta_spin_start + num_walkers;
+
+        if (use_drift)
+        {
+          twf_dispatcher.flex_evalGradWithSpin(walker_twfs, walker_elecs, iat, grads_now, spingrads_now);
+          sft.drift_modifier.getDrifts(tauovermass, grads_now, drifts);
+          sft.drift_modifier.getDrifts(spintauovermass, spingrads_now, spindrifts);
```
ye-luo, line 299: Divergent

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -215,6 +215,216 @@ void VMCBatched::advanceWalkers(const StateForThread& sft,
...
+                         [sqrtspintau](const ParticleSet::Scalar_t& delta_spin) { return sqrtspintau * delta_spin; });
+        }
+
+        ps_dispatcher.flex_makeMoveWithSpin(walker_elecs, iat, drifts, spindrifts);
+
+        // This is inelegant
+        if (use_drift)
+        {
+          twf_dispatcher.flex_calcRatioGrad(walker_twfs, walker_elecs, iat, ratios, grads_new);
+          twf_dispatcher.flex_calcRatioGradWithSpin(walker_twfs, walker_elecs, iat, ratios, grads_new, spingrads_new);
```
ye-luo, line 325: Divergent

ye-luo: need both 324 and 325 lines?

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -215,6 +215,216 @@ void VMCBatched::advanceWalkers(const StateForThread& sft,
...
+                         [oneover2spintau](const ParticleSet::Scalar_t& spindrift, const RealType& lgb) {
+                           return lgb - oneover2spintau * spindrift * spindrift;
+                         });
+        }
+        else
+        {
+          twf_dispatcher.flex_calcRatio(walker_twfs, walker_elecs, iat, ratios);
+        }
+
+        std::transform(ratios.begin(), ratios.end(), prob.begin(), [](auto ratio) { return std::norm(ratio); });
```
ye-luo, line 357: From this line and below, the code is spin agnostic. We need to think harder about an abstract to wrap the divergence between spin and non-spin cases.

ye-luo (changes requested):

src/QMCDrivers/VMC/VMCBatched.cpp
camelto2: oops. good catch

src/QMCDrivers/VMC/VMCBatched.cpp
camelto2: same is true in the DMC batched.

src/QMCDrivers/VMC/VMCBatched.cpp
camelto2: fixed

prckent:
(General comment, not intended to slow this down in any way) We have discussed this in the past, but to leave a record here - while the batched spin move drivers are worked on and merged, we should all think about how it might be possible to write a unified driver in the future that handles the non-spin move/spinor case as well. This is a matter of more than "simple" abstractions since the Monte Carlo algorithms are different, but they are not that different. One of the problems with the old CPU code is the proliferation of drivers, which adds risk, testing, & maintenance burden due to the proliferation of code paths. Would be good to avoid in the new code if practical.

camelto2:
made this WIP while I work on an abstraction layer after talking with Ye and in line with Paul's comments above. Might be better to go ahead and get that sorted out sooner rather than later

PR #3744, opened 2022-01-20
Update timer and Summit recipe.

ye-luo:
## Proposed changes

The source code change is to make timer names more brief.

Update Summit recipe with up-to-date Clang and CUDA 11. Previously older versions of Clang may hang offload code at execution when CUDA 11 was used. This issue has been addressed in upstream and next clang 14 release (Feb) will be problem free.

Remove CMake workaround as the default installation 3.21.3 on Summit doesn't need it.

- Build related changes

- No

Summit

## Checklist


prckent (approved):
Thank you.

ye-luo:
Test this please

PR #3746, opened 2022-01-20
Turn on NLPP batched algorithm by default.

ye-luo:
## Proposed changes
Using the quadrature point batched algorithm in NLPP by default.
Only MSD-traditional and k-space don't carry an implementation.

- Other (please describe):

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (commented):
We seem to be missing a default implementation in the wfcomponent base definitions? Just a loop over the serial version. We have to avoid features that don't work everywhere. This was a huge pain with the SoA version for KNL. Make an issue or fix here?

I did do some casual timings, and even on cpu with Intel compiler on sulfur, the batched algorithm default is sensible for the performance runs:

```
build_intel/ctest_perf.out:17/94 Test #134: performance-NiO-a96-e1152-batched_driver-w16-DU32-1-4 .........   Passed  171.65 sec
build_intel_nlpp/ctest_perf_batched.out:17/94 Test #134: performance-NiO-a96-e1152-batched_driver-w16-DU32-1-4 .........   Passed  163.72 sec
build_intel/ctest_perf.out:19/94 Test #142: performance-NiO-a128-e1536-batched_driver-w16-DU32-1-4 ........   Passed  322.53 sec
build_intel_nlpp/ctest_perf_batched.out:19/94 Test #142: performance-NiO-a128-e1536-batched_driver-w16-DU32-1-4 ........   Passed  317.84 sec
```

ye-luo:
The are two kind of virtual moves.
1. evaluateRatiosAlltoOne: Make a move from all the electrons to a temporal locations. This one can be implemented as a loop of serial ratio.
2. evaluateRatios: This is what I'm putting as default. Make move a single electron to a set of particles recorded in VirtualParticleSet. So there is no temporal proposed moves and we cannot loop over the serial version because the temporal location is not available. This is on purpose, if the temporal location is used, it needs to be updated and thus violates const ParticleSet restriction in TWF APIs. This const restriction prevents bugs from dangerous use cases.

For this reason, we don't provide a fallback. Implementing evaluateRatios is not difficult. The difference from serial version is minimal. We will need to implement it in kSpace Jastrow. Recarding MSD traditional, I strongly suggest we delete the implementation. There is really zero value to keep it. We are not sure it is fully correct as nightly tests fails intermittently, no-one uses frequently and high maintenance is needed. I still remember how painful it is too work with when I would like to consolidate pointer ownership and I have to open a backdoor to make it work via shared_ptr.
Currently multi-determinant backflow depends on it but I don't have confident in that implementation either.

prckent (approved):
I misinterpreted your comment. This is not about fallback.

prckent:
Test this please

PR #3755, opened 2022-01-24
Spin off matrix inversion from DelayedUpdateCUDA.h

### LINKED ISSUE #3739: Refactor cuSolver integration in DelayedUpdateCUDA.h
Currently, cusolver is directly called in DelayedUpdateCUDA.h to execute matrix inversion. It is necessary to build an abstraction and move the cusolver implementation under `Platforms/CUDA`.
With this refactoring, we will able able to add  `Platforms/ROCm` to have an rocsolver implementation. Note that cuSolver API is different from rocSolver APIs when handling scratch spaces and we can not just map functions via macro define.
By doing this, we will achieve sharing DelayedUpdateCUDA.h between CUDA and HIP while specialized to call cusolver or rocsolver.

ye-luo:
## Proposed changes
Closes #3739
The cusolver portion has been moved to standalone class. In this way, DelayedUpdateCUDA can be used with ROCm although the inversion currently runs on the host. We will need a rocsolver implementation.

Now in the DiracDeterminant implementation, input option selecting host/accel is respected similarly to DiracDeterminantBatched. Expanded DiracDeterminant unit tests to test both cases.

- No

epyc-server

## Checklist


ye-luo:
Test this please.

prckent (approved):
Confirmed inverter tests are working locally (gcc + MKL)

PR #3758, opened 2022-01-24
Remove MSD traditional plus Backflow

### LINKED ISSUE #3757: Multideterminant (traditional) plus backflow can be removed

The current implementation does not appear to be used, has some implementation issues, and is also the last user of the classic multislater code. This requires upkeep and maintenance we don't want to do given current pressures and bandwidth. The current code will also not be performant for large systems.

Based on a Slack discussion, since the code has no champions or identified users, we can remove it.  If at a future date a clear use is found, it should be cleanly reimplemented.

Find a champion who can modernize the code.

ye-luo:
## Proposed changes
Closes  #3757

Remove MSD traditional and the backflow build on top of it.
Make SPO unique_ptr in DiracDeterminantBase.
Rename MultiSlaterDeterminantFast to MultiSlaterDetTableMethod.
Update documentation.

- Documentation changes

- Yes
algorithm="all_determinants" is no more accepted.

epyc-server

## Checklist


ye-luo:
Test this please

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.h
```diff
@@ -124,7 +119,7 @@ class SlaterDetBuilder : public WaveFunctionComponentBuilder
       extVar = "Coeff_" + std::to_string(ext_level);
 
     if (!hin.readEntry(ci_coeff, extVar))
-      APP_ABORT("Could not read CI coefficients from HDF5");
+      throw std::runtime_error("Could not read CI coefficients from HDF5");
```
PDoakORNL, line 122: Unless I'm missing something this appears to get called on all ranks.  Since I'd expect them to all fail this should be a `UniformCommunicateError` this should be our way to signal an error is expected on all ranks so we should wait for all of them to fail and report the error on the head rank.

PDoakORNL (changes requested):
Lots of exception that will frequently result in lost error output for the user.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.h
ye-luo: I'm surprised to see this is called from all ranks. A fix will be on the caller side.
Reading redundant 1M coefs from all the ranks is not good.

prckent: Perhaps a topic for an issue / subsequent PR? To me it looks like the SlaterDetBuilder hasn't been through cleanup with a view of potentially having lots of MPI ranks. Also, there is functionality in the .h that has no business not being in a .cpp.

PDoakORNL: Alright this needs to be revisited on a couple of fronts and isn't something that can be solved in this PR.  I'm going to approve.

PDoakORNL (approved):

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.h
prckent: (If updates are going to wait, the issue should be created ahead of merge)

PDoakORNL: I've made issues for the UCE related stuff.  The reading a million coefficients on every rank issue is separate and not added.

ye-luo:
Test this please

PR #3762, opened 2022-01-25
Batched driver move abstraction

camelto2:

## Proposed changes

This PR adds an abstraction layer to the batched drivers for the particle moves. It is intended to keep the VMC/DMC advanceWalkers agnostic about the actual moves that are being done (just particle positions, positions and spins, etc). Right now advanceWalkers is templated on a move type, and uses the templated class MoveAbstraction to handle the specifics of the moves. I am happy to change any names for classes/functions/etc here to make things more clear, I was more concerned with getting something working.

note: Eventually this should be able to handle the work needed for the L2potentials as well

part of #3712, and would allow me to delete #3742

- New feature

- No

macOSX

## Checklist


camelto2:
Need to follow up and add comments to everything

camelto2:
> Need to follow up and add comments to everything

I added documentation detailing the intention of each method in the MoveAbstraction class and describing each member variable. Also updated the docs to include the spin mass in the batched driver sections

ye-luo:
Here is my thoughts
1. MoveAbstraction constructor should capture ps_dispatcher and walker_elecs. Leave twf_dispatcher out.
2. Pass twf_dispatcher explicitly in APIs need it.
3. There should no trace of Crowd object in MoveAbstraction
4. acceptMove should be wrapped symmetrically to makeMove.

Replace mw_makeMoveWithSpin with mw_makeSpinMove which only deal with spin part. Call flex_makeMove and mw_makeSpinMove in the abstraction.
Add mw_accept_rejectSpinMove. Move the touch of spins from mw_accept_rejectMove to this new function and call it from the wrapper.

still thinking about a better name for MoveAbstraction. MoveManipulator? The object name should be a noun, not "move". Will can do the change once we settle the class name.

PDoakORNL (dismissed):
This PR sets up a new state machine and mixed data and operations class.
Due to the large upfront capture of variables and state members it hides the actual data dependence of many steps of the qmc section which I consider to be undesirable.
The methods are almost all state transformations of the object and not clean functions.

While I feel that breaking up advance walkers into pure functions would be worth while I think templating or giving them if constexpr sections based the Monte Carlo Coordinate type is a much cleaner way to add this functionality.

I would propose working from a different start such as in #3767

ye-luo:
@PDoakORNL can we first complete this one and then work on refactoring? #3767 is a prototype and only partially address what is abstracted in this PR. So it takes a bit more time to work out a fully satisfactory design.

src/QMCDrivers/MoveAbstraction.h
```diff
@@ -0,0 +1,444 @@
...
+//
+// File developed by: Cody A. Melton, cmelton@sandia.gov, Sandia National Laboratories
+//
+// File created by: Cody A. Melton, cmelton@sandia.gov, Sandia National Laboratories
+//////////////////////////////////////////////////////////////////////////////////////
+
+#ifndef QMCPLUSPLUS_MOVEABSTRACTION_H
+#define QMCPLUSPLUS_MOVEABSTRACTION_H
+
+#include "QMCDriverNew.h"
```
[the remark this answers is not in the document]
camelto2: done

camelto2:
> Here is my thoughts
> 
>     1. MoveAbstraction constructor should capture ps_dispatcher and walker_elecs. Leave twf_dispatcher out.
> 
>     2. Pass twf_dispatcher explicitly in APIs need it.
> 
>     3. There should no trace of Crowd object in MoveAbstraction

This is taken care of with the recent push. Looking at the changes to particleset now.

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -233,7 +213,13 @@ void VMCBatched::runVMCStep(int crowd_id,
   const bool recompute_this_step = (sft.is_recomputing_block && (step + 1) == max_steps);
   // For VMC we don't call this method for warmup steps.
   const bool accumulate_this_step = true;
-  advanceWalkers(sft, crowd, timers, *context_for_steps[crowd_id], recompute_this_step, accumulate_this_step);
+  const bool spin_move            = crowd.get_walker_elecs()[0].get().isSpinor();
```
ye-luo, line 216: This line can be dangerous if there is no element.
Pass the state via the sft object.
Also use `population_.get_golden_electrons()`

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -80,68 +78,37 @@ void VMCBatched::advanceWalkers(const StateForThread& sft,
   std::vector<std::reference_wrapper<TrialWaveFunction>> twf_accept_list, twf_reject_list;
   isAccepted.reserve(num_walkers);
 
+  MoveAbstraction<COORDS> move(ps_dispatcher, walker_elecs, step_context.get_random_gen(), sft.drift_modifier,
```
ye-luo, line 81: Let us change this to 'mover' a noun.

camelto2: done

src/QMCDrivers/VMC/VMCBatched.cpp
camelto2: done

ye-luo (dismissed):
LGTM. I'm inclined to merge this for fast turnaround and avoid working on forks although I consider MoveAbstraction should go away as we get #3767 ready.

camelto2:
Just incorporated the newer MCCoords stuff into this PR. @ye-luo also wants me also to template the ParticleSet mw_makeMove APIs on the CoordsType, so I will follow up with that with additional commits

camelto2:
Made a significant change to this PR. This should be more in line with what @PDoakORNL was thinking.

I completely removed the MoveAbstraction class which this PR introduced.
Instead, I have introduced a new struct TWFGrads, similar to how MCCoords was introduced in #3774, which gets templated on the CoordsType. This allows me to carry around TWFGrads<CT> grads_now, grads_new in the advanceWalkers which will carry spatial gradients and spin gradients depending on the CoordsType. This also allowed me to template the TWFdispatcher on CoordsType, so that I can use a templated  flex_ API to decide if it is including spin or not for the gradients.
I also templated  DriftModifierBase which removes a lot of the divergent code for normal vs. spin calculations in advanceWalkers. 

With a little more work, the advanceWalkers will be further simplified so it doesn't have to have any "if constexpr (CT==CoordsType::POS_SPIN)" code

PDoakORNL:
Will start rereview this afternoon.

src/QMCDrivers/TauParams.hpp
```diff
@@ -0,0 +1,65 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2022 QMCPACK developers.
+//
+// File developed by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
```
PDoakORNL, line 7: @ye-luo didn't write this.
Put my name, yours and we have been writing
`// Refactored from MCCoords.hpp`
instead of File created by.

I think that is more meaningful when you're tearing up or porting a well known file.
The name is better.

PDoakORNL (approved):
This looks like a good step.  If there is time fix the attribution for TauParams.

src/QMCDrivers/TauParams.hpp
ye-luo: I did the refactoring. So I will take care of this.

src/QMCWaveFunctions/tests/test_MO_spinor.cpp
```diff
@@ -271,8 +275,11 @@ void test_lcao_spinor()
   for (int iat = 0; iat < 1; iat++)
   {
     std::vector<ParticleSet::SingleParticlePos> displs = {dR[iat], dR[iat]};
+    std::vector<ParticleSet::Scalar_t> sdispls = {dS[iat], dS[iat]};
     elec_.mw_makeMove(p_list, iat, displs);
+    elec_.mw_makeSpinMove(p_list, iat, sdispls);
```
ye-luo, line 280: Could update this test to call the new APIs.

camelto2: yea, I was just working on this

camelto2: fixed

camelto2:
ok, added a unit test for the TWFGrads class I added. With the current changes, there are only a few spots left in VMCBatched and DMCBatched that need to be abstracted, but those will be handled by a subsequent PR. This one is large enough as it is

ye-luo:
Test this please

ye-luo (dismissed):

ye-luo (approved):

ye-luo:
Test this please

PR #3763, opened 2022-01-25
Use minimum image of particles in EwaldRef

### LINKED ISSUE #2105: Incorrect ion-ion energy in some periodic systems

ye-luo:
## Proposed changes
The reference calculation went wrong this time.
When particles don't start within the unit cell, the grid summation of EwardRef  doesn't work well. See first commit.
The QMCPACK value is OK. https://github.com/QMCPACK/qmcpack/issues/2105#issuecomment-1019520086 passes with this fix.

Closes #2105

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

jtkrogel:
It's not clear where the key change (putting particles in the box) is made in the code given the refactoring.  Please point out the location of the actual fix.

jtkrogel:
Is the underlying issue a premature truncation of the real space sum?  

If r_i-r_j is outside the cell, then the first term in the sum (centered at zero lattice vector shift) could be rather small, perhaps smaller than the tolerance for the sum overall.

ye-luo:
> Is the underlying issue a premature truncation of the real space sum?

Yes

> If r_i-r_j is outside the cell, then the first term in the sum (centered at zero lattice vector shift) could be rather small, perhaps smaller than the tolerance for the sum overall.

The tolerance check is on the sum of the absolute values contributed by grid points on the surface shell.
This first term (all 0 offset) is always excluded from the tolerance check. Instead some later shell went under tolerance before reaching the shells which contributes significant values when the initial coordinate delta are very far from the center.

prckent:
I also thought about testing for several shells in a row being below tolerance, but I think once the particles are minimum imaged that is no longer necessary. Also, if a particle ever wandered "far enough" there could still be an issue with not testing enough shells.

Since this fixes the reported problem, adds a test for it, and doesn't require any changes to the values used in other tests, can we consider this safe to merge?

ye-luo:
Test this please

jtkrogel (approved):
Yes, let's merge this.

ye-luo:
Test this please

PR #3767, opened 2022-01-25
[Proposal] Abstract the Monte Carlo coordinates

PDoakORNL:
## Proposed changes

Make an abstraction for the Monte Carlo Coords.

- New feature

- Yes
- No

## Checklist

### DIFF
--- a/src/QMCDrivers/GreenFunctionModifiers/DriftModifierUNR.h
+++ b/src/QMCDrivers/GreenFunctionModifiers/DriftModifierUNR.h
@@ -14,6 +14,7 @@
 #define QMCPLUSPLUS_DRIFTMODIFIER_UNR_H
 
 #include "QMCDrivers/GreenFunctionModifiers/DriftModifierBase.h"
+#include "QMCDrivers/MCCoords.hpp"
 
 namespace qmcplusplus
 {
@@ -23,8 +24,9 @@ public:
   using RealType = QMCTraits::RealType;
   using PosType  = QMCTraits::PosType;
 
-  void getDrifts(RealType tau, const std::vector<GradType>& qf, std::vector<PosType>&) const final;
-
+  void getDrifts(RealType tau, const std::vector<GradType>& qf, MCCoords<MCCoordsTypes::RSSPINS>&) const final;
+  void getDrifts(RealType tau, const std::vector<GradType>& qf, MCCoords<MCCoordsTypes::RS>&) const final;
+  
   void getDrift(RealType tau, const GradType& qf, PosType& drift) const final;
 
   void getDrift(RealType tau, const ComplexType& qf, ParticleSet::Scalar_t& drift) const final;
@@ -38,6 +40,7 @@ private:
   RealType a_;
 };
 
+  
 } // namespace qmcplusplus
 
 #endif
--- a/src/QMCDrivers/MCCoords.hpp
+++ b/src/QMCDrivers/MCCoords.hpp
@@ -1,0 +1,118 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2022 developers.
+//
+// File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Laboratory
+//
+// File created by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+#ifndef QMCPLUSPLUS_MCCOORDS_HPP
+#define QMCPLUSPLUS_MCCOORDS_HPP
+
+#include "Configuration.h"
+#include "type_traits/complex_help.hpp"
+#include "ParticleBase/RandomSeqGenerator.h"
+
+#include <vector>
+
+namespace qmcplusplus
+{
+
+enum class MCCoordsTypes
+{
+  RS,
+  RSSPINS
+};
+
+template<MCCoordsTypes CT = MCCoordsTypes::RS>
+struct MCCoords;
+
+template<MCCoordsTypes MCT>
+struct MCCoords
+{
+  static constexpr MCCoordsTypes mct = MCT;
+  // This cleans up some other code.
+  void resize(const std::size_t size);
+  std::vector<QMCTraits::PosType> rs;
+};
+
+template<>
+struct MCCoords<MCCoordsTypes::RSSPINS>
+{
+  static constexpr MCCoordsTypes mct = MCCoordsTypes::RSSPINS;
+  // This cleans up some other code.
+  void resize(const std::size_t size)
+  {
+    rs.resize(size);
+    spins.resize(size);
+  }
+  std::vector<QMCTraits::PosType> rs;
+  std::vector<std::complex<QMCTraits::RealType>> spins;
+};
+
+template<MCCoordsTypes CT = MCCoordsTypes::RS>
+struct MCCIt
+{
+  std::vector<QMCTraits::PosType>::iterator irs;
+};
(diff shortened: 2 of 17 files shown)


ye-luo:
I think you are right. The data object needs to be templated on coordinates.
Since you are touching ContextForSteps, it is time to think hard what is the actual role of ContextForSteps.
It ships Deltas and RNG. It seems serving two tasks.
Deltas. I feel it doesn't t need to be persistent and thus not needed to be passed via argument.
RNG. That needs to be passed in.

So If we can can split the role of ContextForSteps, we may not need to make templates from so high up levels like run().

camelto2:
> I think you are right. The data object needs to be templated on coordinates. Since you are touching ContextForSteps, it is time to think hard what is the actual role of ContextForSteps. It ships Deltas and RNG. It seems serving two tasks. Deltas. I feel it doesn't t need to be persistent and thus not needed to be passed via argument. RNG. That needs to be passed in.

in the approach used in #3762, all of the deltas are handled in the MoveAbstraction class and could be removed from ContextForSteps

PDoakORNL:
The mistake in ContextForSteps is that the deltas are state.  They should have just been returned by pure function not in ContextForSteps.

```c++
template<MCCoordTypes MCT, class RNG>
MCCoords<MCT> generateDeltas(RNG rng, size_t num_rs)
{
   MCCoords<MCT> mc_coords;
   mc_coords.rs.resize(num_rs);
   makeGaussRandomWithEngine(mc_coords.rs, rng);
   if constexpr (std::is_same<decltype(walker_deltas_), MCCoords<MCCoordsTypes::RSSPINS>>::value)
  {
      mc_coords.spins.resize(num_rs);
      makeGaussRandomWithEngine(mc_coords.spins, rng); 
  }
  return mc_coords;
}
```

PDoakORNL:
@ye-luo It wasn't too hard reverse the bad delta_r in ContextForSteps decisions and that indeed makes it easy to push down the templating.

It becomes clear here that a number of templated pure functions would remove duplicated code from DMC/VMC Batched.  Some of these are the same code moved to the Move abstraction with uncessary state in #3762.  Using `if contexpr` considerable repetition can be removed from the pure functions and the actual data dependence of computations can remain visible from the driver code.

src/QMCDrivers/GreenFunctionModifiers/DriftModifierUNR.h
```diff
@@ -23,8 +24,9 @@ class DriftModifierUNR : public DriftModifierBase
   using RealType = QMCTraits::RealType;
   using PosType  = QMCTraits::PosType;
 
-  void getDrifts(RealType tau, const std::vector<GradType>& qf, std::vector<PosType>&) const final;
-
+  void getDrifts(RealType tau, const std::vector<GradType>& qf, MCCoords<MCCoordsTypes::RSSPINS>&) const final;
+  void getDrifts(RealType tau, const std::vector<GradType>& qf, MCCoords<MCCoordsTypes::RS>&) const final;
```
ye-luo, line 28: I like this pattern much more than `conexpr if`. Namely, API becomes reactive. Cody has refactored the ParticleSet APIs
mw_makeMoveWithSpin has been replaced with mw_makeSpinMove.
The old mw_makeMoveWithSpin can be achieved with mw_makeMove and mw_makeSpinMove.
Then I think we can introduce
```
makeMove(MCCoords<MCCoordsTypes::RSSPINS>&)
makeMove(MCCoords<MCCoordsTypes::RS>&)
```
on top of existing APIs.

PDoakORNL: I'd much prefer a templated function but because of the unecessary virtual inheritance in the DriftModifiers we are stuck with the overloading.

ye-luo: What you want is a templates function with specialization without the class being templated. We have to go for CRTP I guess. Then do we really want the whole class being templated? Since we anyway need to specialize the function. Just putting distinct functions is totally fine.

PDoakORNL: I think in many cases we don't want the whole class templated. If we do maybe CRTP is useful maybe we actually really gain something from the virtual inheritance.

But drift modifier would be better without inheritance. I think this a case that clearly would be better written with `if constrexpr` and no vtable.  The overloads are going to cost several vtable calls without repeating code.

PDoakORNL: do you mean a `ParticleSet::makeMove` overload set?
That I think makes sense.
Hiding the QMC algorithm in another state machine I'm not in favor of.

src/QMCDrivers/MCCoords.hpp
```diff
@@ -0,0 +1,118 @@
...
+  RSSPINS
+};
+
+template<MCCoordsTypes CT = MCCoordsTypes::RS>
+struct MCCoords;
+
+template<MCCoordsTypes MCT>
+struct MCCoords
+{
+  static constexpr MCCoordsTypes mct = MCT;
```
ye-luo, line 36: Hopefully this indirection can be removed

src/QMCDrivers/MCCoords.hpp
```diff
@@ -0,0 +1,118 @@
...
+{
+  RS,
+  RSSPINS
+};
+
+template<MCCoordsTypes CT = MCCoordsTypes::RS>
+struct MCCoords;
+
+template<MCCoordsTypes MCT>
+struct MCCoords
```
ye-luo, line 34: Yes. We need this class.

src/QMCDrivers/MCCoords.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+template<>
+struct MCCIt<MCCoordsTypes::RSSPINS>
+{
+  std::vector<QMCTraits::PosType>::iterator irs;
+  std::vector<std::complex<QMCTraits::RealType>>::iterator spins;
+};
+
+template<typename Real, MCCoordsTypes CT = MCCoordsTypes::RS>
+struct Taus
```
ye-luo, line 70: We need this class.

src/QMCDrivers/MCCoords.hpp
```diff
@@ -0,0 +1,118 @@
...
+  {
+    Real spin_tauovermass = Base::tauovermass / spin_mass;
+    Real spin_oneover2tau = 0.5 / (spin_tauovermass);
+    Real spin_sqrttau     = std::sqrt(spin_tauovermass);
+  }
+};
+
+
+template<class MCC, class RNG>
+MCC generateDeltas(RNG rng, size_t num_rs)
```
ye-luo, line 100: Let us move this class to `src/Particle/ParticleBase/RandomSeqGenerator.h`
Also move this file to `src/Particle`

src/QMCDrivers/MCCoords.hpp
```diff
@@ -0,0 +1,118 @@
...
+#include "Configuration.h"
+#include "type_traits/complex_help.hpp"
+#include "ParticleBase/RandomSeqGenerator.h"
+
+#include <vector>
+
+namespace qmcplusplus
+{
+
+enum class MCCoordsTypes
```
ye-luo, line 24: Yes. we need this enum. `CoordsTypes` no need of MC.
Prefer longer names. So use `POS` and `POS_SPIN`?

ye-luo (commented):
I like the direction. It is the right thing we have to explore.
My suggestions is we merge #3762, then gradually the COORD abstraction in ParticleSet and dissolve MoveAbstraction.

Whether to start from a fresh branch or rebase. It will be your choice.

PDoakORNL:
This has served it purpose and a useful subset submitted as #3774

PR #3769, opened 2022-01-26
NLPP house keeping

ye-luo:
## Proposed changes
Change vrad type back to Real. It is only used as real numbers.
Expand unit tests.

## What type(s) of changes does this code introduce?
- Refactoring (no functional changes, no api changes)
- Testing changes (e.g. new unit/integration/performance tests)

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
epyc-server

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'

PR #3770, opened 2022-01-26
Remove DiffWaveFunctionComponent

### LINKED ISSUE #958: DiffWaveFunctionComponent needs retirement
DiffWaveFunctionComponent(previously DiffOrbitalBase) and its derived classes wrere introduced to handle multiple implementation of WF derivatives. It is no more needed and the evaluateDerivatives needs to be placed directly into the corresponding derived classes.

Jastrow/DiffTwoBodyJastrowOrbital.h holds both the AoS and the SoA implementations. and they should be disentangled.

```
DiffWaveFunctionComponent.h
DiffWaveFunctionComponent.cpp
Jastrow/DiffOneBodySpinJastrowOrbital.h
Jastrow/DiffTwoBodyJastrowOrbital.h
Jastrow/DiffOneBodyJastrowOrbital.h
```
should go away

ye-luo:
Review after #3769

## Proposed changes
Closes  #958
evaluateDerivatives becomes mandatory. WFC don't support optimization must supply an empty implementation.
evaluateDerivativesWF is still problematic in the sense that not many WFC implements it as it was added to make things barely work. It will be clear if we should eventually get rid of this API when the NLPP batch API takes over.

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):
Will be interesting to see how many requests we receive for evaluateDerivativeWF implementation  and for which of the less popular wavefunction component types.

Issue #1386, opened 2019-02-15
Inconsistent twist directions between electron gas and einspline wavefunctions


Paul-St-Young:
**Sympton**: electron gas momentum distribution n(k) is wrong at fractional twist (i.e. not 0 or 0.5)
**Direct cause**:  `MomentumEstimator::putSpecial` subtracts the twist vector from kpoint grid, whereas `HEGGrid::create_kpoints` adds the twist vector to the kpoint grid.
**Root cause**: einspline wavefunction puts negative sign in front of kgrid?

I found this bug when trying to calculate the momentum distribution n(k) of the homogeneous electron gas (HEG) at a fractional twist e.g. (0.3, 0.0, 0.0). I realized that the `MomentumEstimator` *subtracts* the twist vector from the Gamma-point kgrid
`       ikpt[0]=i-twist[0];
        ikpt[1]=j-twist[1];
        ikpt[2]=k-twist[2];`
whereas the electron gas orbital builder *adds* the twist to the Gamma-point kgrid
`PosType k(i0+tw[0],i1+tw[1],i2+tw[2]);`

If I add the twist to kgrid in `MomentumEstimator`, then n(k) becomes correct. (see figure below).
I have also attached a tiny example (2 electrons)  [short.zip](https://github.com/QMCPACK/qmcpack/files/2871113/short.zip), which could be added as a short test for the bug in question.
![heg-nk](https://user-images.githubusercontent.com/8060441/52888035-c20f6400-313f-11e9-9082-438b12bc938a.png)

Of course, the current `MomentumEstimator` is correct when used with einspline wavefunction, so I think the bigger question is: why does einspline wavefunction negate the twist direction?

Unfortunately, I am not sure how exactly the einspline wavefunction sets up the kgrid and the twist vectors, because the wavefunction construction is quite complicated. File potentially involved:
- EinsplineSetBuilderCommon.cpp
- EinsplineSetBuilder_createSPOs.cpp
- EinsplineSetBuilderESHDF.fft.cpp
- EinsplineSetBuilder.h
- EinsplineSetBuilderOld.cpp
- EinsplineSetBuilderReadBands_ESHDF.cpp
- EinsplineSet.cpp
- EinsplineSet.h
- Bspline*?

In my opinion, adding twist to a positive kgrid is the intuitive option, so I am tempted to change `MomentumEstimator` and the einspline wavefunction. On the other hand, changing `HEGGrid` is a lot easier. How should I proceed?

[...]

PR #3792, opened 2022-01-31
Remove twistnum in bspline SPO XML attribute

ye-luo:
## Proposed changes
The existing documentation clearly says avoid using twistnum and I just made one more step by removing it.
Use twist="x.x y.y z.z" input to indicate the exact twist. From now on, all twistnum input will be ignored.
Massive changes are due to updating input files.
I explicit tested complex build.

- Other (please describe): Simply input tags.

- No

epyc-server

## Checklist

### DIFF
--- a/docs/hamiltonianobservable.rst
+++ b/docs/hamiltonianobservable.rst
@@ -1682,7 +1682,7 @@ Additional information:
   :caption: Example ``sposet`` initialization for density matrix use.  Occupied and virtual orbital sets are created separately, then joined (``basis="spo_u spo_uv"``).
   :name: Listing 39
 
-  <sposet_builder type="bspline" href="../dft/pwscf_output/pwscf.pwscf.h5" tilematrix="1 0 0 0 1 0 0 0 1" twistnum="0" meshfactor="1.0" gpu="no" precision="single">
+  <sposet_builder type="bspline" href="../dft/pwscf_output/pwscf.pwscf.h5" tilematrix="1 0 0 0 1 0 0 0 1" meshfactor="1.0" gpu="no" precision="single">
     <sposet type="bspline" name="spo_u"  group="0" size="4"/>
     <sposet type="bspline" name="spo_d"  group="0" size="2"/>
     <sposet type="bspline" name="spo_uv" group="0" index_min="4" index_max="10"/>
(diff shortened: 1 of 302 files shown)


prckent:
Lets imagine an old input file with twistnum but no twist specified in the XML. Do we abort?

ye-luo:
First if stopper is desired, I can add it but that requires updating nexus. Users must ensure matching nexus and QMCPACK. In the current way, mostly old input file works as intended with new QMCPACK.

In the massive updates of tests, I only saw two cases where I need to modify twist.
1. graphite performance test, the input was created probably a decade ago. twistnum=2 and I need to add twist="0 0 0.5".
2. All LiH-arb tests, I think you added the tests and the input file was manually created. Thus there was twistnum=0 without a twist specified. The code complains about not being able find the twist. In this case, I added twsit="-0.1 -0.2 -0.3".

In all the input files with a twist input, twistnum="-1" are already specified.

In brief, the only case that breaks is having twistnum input without twist input. The intended twist is not gamma but gamma does exist in the h5.

prckent:
Thanks. I was just asking the question. Safe option would be abort with a useful error if twistnum is specified and !=-1. A few minutes work on our part could avoid a bad surprise in future.

NEXUS is an easy update. Yes people should sync, no issues there. (@jtkrogel ?)

jtkrogel:
OK, first of all, I don't think `twistnum` should be removed.  I thought @ye-luo was going to open a separate issue on this for discussion.  There are two reasons for this:

1) Backwards compatibility.  Overall, I think we don't prioritize this enough and are overly willing to put users off just because "we don't like something".

2) Nexus is not a minor update and I have been postponing making this update because of the inverted sign issue, as documented in #1386.  

Instead, we should first make a PR to default use `twist` by default (and not require `twistnum="-1"` to be provided), and then if it is not provided, use `twistnum`.  Next, #1386 should be fixed in a PR, and we should be sure that HEG, Bspline, and LCAO wavefunctions are using the same convention.  Finally, Nexus should be updated in a PR to use the now correct and consistent definition of twists.  At this point, `twistnum` should remain an option for backwards compatibility reasons, but could be removed from the documentation with `twist` being promoted there instead.

jtkrogel (changes requested):
#1386 and #3795 should be resolved first.

ye-luo:
My second thought on this. I think changing input precedence and keep old files running is better.
We need to remember twistnum has defect "twistnum is not robust against h5 file changes."

ye-luo:
If neither twistnum nor twist exist, is default to `twist=0 0 0` acceptable? The alterrnative is default twistnum=0 which I dislike.

prckent:
One possibility might be to print a warning ("This is potentially ambiguous. Specifying twist=xyz is preferred.") 

We could at least stop creating new files with twistnum.

lshulen:
I prefer requiring either twist or twistnum exist to having a default behavior.

ye-luo:
See  #3799

PR #3796, opened 2022-02-01
Nexus: Add --user $USER to squeue command

### LINKED ISSUE #2380: How to contine the running using NEXUS
Dear all,

I have finished the Jastrow optimization using the input file generated by NEXUS.

The state of workflow is shown as follow,
  cascade status 
    setup, sent_files, submitted, finished, got_output, analyzed, failed 
    111111  0  25846     scf     ./graphene_p-1L-tw/scf  
    111111  0  25977     nscf    ./graphene_p-1L-tw/nscf  
    111111  0  26108     p2q     ./graphene_p-1L-tw/nscf  
    **111000**  0  26152     opt     ./graphene_p-1L-tw/opt  
    000000  0  ------    dmc     ./graphene_p-1L-tw/dmc  
    setup, sent_files, submitted, finished, got_output, analyzed, failed 

Now I want to use the results of Jastrow optimization to generate the input file for DMC.
I find NEXUS can only continue the running with all the the cascade status is 1. In this case (111000), it does not work.

I have a tricky way that copy a **sim_opt** folder whose status is 111111 to ./opt. But I don't know whether or not it will effect the generated input of DMC.
 
So, how to continue the running (now it should start from the **got_output** and **analyzed** in opt ) in this case?

Best,
Justin

mcbennet:
## Proposed changes

Adding flag `--me` to squeue command in `machines.py`. This prevents users from being forced to provide their usernames to avoid strange errors when other users format their job names in weird ways. This also resolves #2380 

- Other (please describe): Add stability

- No. All `nxs-test` tests pass

Laptop. Fedora Linux

## Checklist


jtkrogel:
What Slurm versions support the `--me` flag?

mcbennet:
I found this running list of updates (https://github.com/SchedMD/slurm/blob/master/NEWSupdates). It does not mention `--me` being added to `squeue` all the way back to v2.1.0 -- so I assume it was supported all the back to this point(?).

CADES uses v20.02.4 and supports the `--me` flag
Cori uses v20.11.08 and supports the `--me` flag

Perhaps someone at Sandia could let us know if `squeue --me` is supported on those machines? @camelto2

prckent:
It looks to have been added to the longoptions of slurm in https://github.com/SchedMD/slurm/commit/848378051f9dd1fd96b952fb00cc074177d5d35d

mcbennet:
Thanks Paul, but I am correct that the commit there relates to "auto completion" functionality? And it looks to be added in v21, however, Cori and CADES support the flag and use v20

mcbennet:
It was added to the man page 5 months ago it appears

Line 1362 of
https://github.com/SchedMD/slurm/blame/master/doc/man/man1/squeue.1

mcbennet:
Correction to my previous comment, it appears to show up in man page for the first time on Feb 2020

https://github.com/SchedMD/slurm/blob/cc6a1be5ee2cf0e1ac5a7c14a6df97023701e8e8/doc/man/man1/squeue.1

@ye-luo This might be okay, but may let Jaron weigh in

prckent:
Wearing my sometimes sysadmin hat, updating slurm is one of the last things i would want to do on a running system. So we can't rely on this --me feature being available or slurm being highly up to date => this can only be a nice to have and not overly important feature. I think Jaron is the decider here. Catching any errors would be wise.

mcbennet:
Just throwing out one other alternative that is worth chewing on. We could consider formatting squeue's output such that it only gives what nexus needs (jobid and status only?). Something like

For instance,
```
squeue -o "%.18i %.2t" 
```
Or something similar. The parsing done by Nexus will need a minor change as well

jtkrogel:
I like the simplicity offered here, but I have to agree with Paul that the chance of breaking based on the local environment is too high.

For the feature to be introduced, I think the current behavior with `user` should be preserved and a new flag used instead to invoke this querying mode.

jtkrogel:
A possible confounding issue is that `squeue` is not the Slurm tool at all sites, and is sometimes replaced by a custom in-house script.  I would prefer providing the option when it is needed.  

If it is known to be safe at certain centers, then perhaps this information should be documented as machine specific in Nexus, with plain `squeue` being invoked most places and `squeue --user $USER` being invoked where it is known to be safe (i.e. the `--user` option actually exists).  This information could be stored in the class-level variables in each Machine class and used to control the default flags.

jtkrogel:
Perhaps `squeue --user $USER` can be the default (test it at a few sites please).  There needs to be a way to turn this behavior off, perhaps by passing `user=False` to `settings`.

mcbennet:
The following command has been run on multiple machines at multiple sites

```python3
execute('squeue --user=$USER')
```

The tested machines and their results are:
* ALCF/Theta (SUCCESS)
* OLCF/Andes (SUCCESS)
* NERSC/Cori (SUCCESS)
* Sandia mchanines (SUCCESS)
* ORNL/CADES (SUCCESS)

The SLURM machines that Nexus supports:

Edison (RETIRED)
Cori 
Chama
Uno
Eclipse
Attaway
Skybridge
Solo
Stampede2
CadesSlurm
Rhea
Andes

jtkrogel:
Good.  I think we can proceed with a change to using `--user=$USER`.

Perhaps the logic should be:
```
if isinstance(self.user,bool) and self.user==False:
    extra = ''
elif self.user is not None:
    extra = ' -u {}'.format(self.user)
else:
    extra = ' --user=$USER'
#end if
```

mcbennet:
Thank you @jtkrogel . pushed

jtkrogel:
LGTM.  Tests pass?

mcbennet:
Yes. 100% of `nxs-test` tests pass

jtkrogel (approved):
Great

ye-luo:
Test this please

PR #3797, opened 2022-02-01
Flip the sign of einspline twist input

### LINKED ISSUE #1386: Inconsistent twist directions between electron gas and einspline wavefunctions

ye-luo:
## Proposed changes
h5 from DFT has [0 0.5 0.25]
Before this PR, we input `twist="0.0   -0.5  -0.25"`
```
Super twist #0:  [   0.00000   0.00000   0.00000 ]
Super twist #1:  [  -0.00000   0.50000  -0.25000 ]
  Using supercell twist 1:  [  -0.00000   0.50000  -0.25000]
```
Now we should input  `twist="0.0 0.5  0.25"` as DFT.
```
Super twist #0:  [   0.00000   0.00000   0.00000 ]
Super twist #1:  [   0.00000   0.50000   0.25000 ]
  Using supercell twist 1:  [   0.00000   0.50000   0.25000]
```
This change breaks backward compatibility. Tools generating input for QMCPACK should be updated.

The fix in splines is in the builders, updating all the spline implementation is to much more involving. Thus kept the old sign.
We may change that after delete EinsplineSet but that doesn't affect users.

I also changed the convention in MomentumEstimator.
closes #1386.

I also deleted code that was commented out.

- Bugfix

- No

epyc-server

## Checklist


jtkrogel:
I will review/comment here, but likely not until tomorrow.

ye-luo:
Test this please

jtkrogel:
The sign issue is only partly an input problem.  Does this just change the sign on the input, but retain the inverted sign internally during orbital construction or evaluation?

If the inverted sign is retained internally, this is a problem because a Bloch state with inverted wavevector exists in the wavefunction.  The actual wavevector of the Bloch state is probed with the momentum distribution, and this can result in no signal in the case of inconsistency.

ye-luo:
The issue of momentum distribution is a pure input problem.
Because orbital saves the twist to ParticleSet and then momentum distribution reads from it.
However ElectronGasBuilder and EinsplineSetBuilder doesn't store the twist in the same convention. By flipping the sign in EinsplineSetBuilder, momentum distribution always take twists in the usual DFT convention.
Note that EinsplineSetBuilder is an object different from EinsplineSet. Momentum distribution doesn't and shouldn't know what underlying SPOSet is.

About EinsplineSet/BsplineSet internal. kPoints are stored as -twist with some additional lattice shift. When evaluating an orbital. A bloch state is `e^(i twist r)u(r)`. u(r) is the spline orbital. Because kPoints carries opposite sign. The evaluation function computes `e^(- i k r)`. See https://github.com/QMCPACK/qmcpack/blob/d69bce8c77060d69d35f78861e501a6408c3cfee/src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp#L185

How is u(r) is constructed. It is first fft from planewave coefficient. Then it is rotated. Rotating a phase doesn't change the physics of orbital itself and there is neither phase imposed from QE. The rotation angle determination is empirical and it needs to evaluate orbitals and there seems to be a sign flip see `einspline_helper.hpp`. but as long as rotation doesn't change physics, it should not hurt. Change it requires updating unit test references.

We need to anatomize the overall issue and take incremental steps.

jtkrogel:
Probably best to have a quick chat/call on this.

ye-luo:
Test this please

jtkrogel (approved):
LGTM.  It seems that Einspline* and Momentum* were correcting for the inverted sign independently, so this changes here should not affect the functioning of n(k) estimators.

The changes here should fix #1386.  Someone should rerun the inputs provided in that issue and check after this PR goes in.

PR #3799, opened 2022-02-02
Make twist attribute taking precedence over twistnum.

ye-luo:
## Proposed changes
Once twist is detected, twistnum is ignored with a warning if found.
If there is only twistnum, take it and print a warning about its ambiguity.
If nothing found, set Gamma.

before the change from this PR, when both twistnum and twist are not provided, the default is twistnum=0.
This is a risky default as twistnum is risky. I made the choice to use Gamma point as the default to reduce input burden. @lshulen suggested just put an error requiring either twist or twistnum input. Let me know your choice.

Note I'm not changing tests in this PR to avoid complicating the review. They will be updated in a separate PR and it does require #3797.

- New feature
- Documentation changes

- Yes

epyc-server

## Checklist


jtkrogel:
Having `twist` take precedence is good.  In the event of no twist or twistnum, I'm OK with either twistnum=0, twist=Gamma, or abort as the default.

Giving `twist` precedence in the input does not need to be dependent on the resolution of the Bloch wavevector sign issue.

src/QMCWaveFunctions/EinsplineSetBuilderCommon.cpp
```diff
@@ -398,101 +399,139 @@ bool EinsplineSetBuilder::TwistPair(PosType a, PosType b)
...
+        char buf[1000];
+        snprintf(buf, 1000,
+                 "AnalyzeTwists2. Input twist [ %9.5f %9.5f %9.5f] not found in the list of super twists above.\n",
+                 twist[0], twist[1], twist[2]);
+        throw UniformCommunicateError(buf);
+      }
+      return twist_num;
+    };
+
+    if (twist_inp[0] > -10 || twist_inp[1] > -10 || twist_inp[2] > -10)
```
jtkrogel, line 492: Why was `-10` chosen here?

prckent: Good catch. I am not for arbitrary constants. We should use clearer logic.

(I also think we should avoid the arbitrary sized buffers elsewhere in this code. It is not a good code pattern and one we should get away from.)

ye-luo: Just added comments in source and commit log.

ye-luo:
Test this please

src/QMCWaveFunctions/EinsplineSetBuilderCommon.cpp
ye-luo: I have replaced -10 with constexpr variables.

ye-luo:
Test this please

ye-luo:
I have replace the magic numbers in the source code with constants. @jtkrogel @prckent OK to merge?

ye-luo:
Test this please

prckent (approved):
Thanks for the updates

PR #3809, opened 2022-02-06
Particles are always grouped internally.

### LINKED ISSUE #541: What does 'Particles are grouped' mean?
In the Particle/ParticleSet.cpp, the code outputs a message about groups. Either 'Particles are grouped. Safe to use groups.'   or a message about needing to use IndirectID.

What does this mean?

In particular, does this have any effect on the user - does it have an effect on how the input file should be constructed?  Or does it affect how the code should be written (which is how the non-grouped message sounds.)

ye-luo:
## Proposed changes
We were allowing creating N particles or {Na, Nb, ...} in ParticleSet.
ParticleSet created in the former way needs to rediscover grouping and it may not be possible if the input order was (O, Ni, O, Ni) instead of (O, O, Ni, Ni). For this reason, we have to keep code path for both grouped and non-grouped case.
This PR removes the former case. Closes #541

This PR makes sure all the particles stored in ParticleSet is grouped and stored in the same order as the order of 'group' XML nodes in the partcleset XML node. For example,
```xml
<particleset name="ion" size="4">
  <group name="O" />
  <group name="Ni" />
  <attrib name="ionid" >
    Ni O Ni O
  </attrib>
```
The final order in the storage is `O O Ni Ni` and input positions are reordered accordingly as well. This feature is unit tested.

After doing this change, code paths relies on `isGrouped() == false` are moved.

Simplify `XMLParticleParser::put/putSpecial` and rename it to  `XMLParticleParser::readXML`

- New feature

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (commented):
I think this is a good simplification, as discussed a long time ago. Given our experience with edge cases and multiple code paths hiding bugs since then, I think the arguments for having a single code path are even stronger.

Will leave for others to comment, but plan to merge this and the follow-up PRs unless there are any objections.

PDoakORNL (approved):
This is a bit frightening since the use of species membersize attribute was pretty widespread but it was always a bad case of two of truth.

If the strange offset style was dropped for just using a vector of sizes and starts for each species that would seem more robust. The assumption that the first of each group is the end of the next in the common storage bothers me. Why not just record the actual size of each species rather than inferring it.  Why not have the calls that this API seems to be for the convenience of take the much safer looking pos, size rather than starting_index plus one past the final index of the group?

ye-luo:
"a vector of sizes and starts for each species" did you mean two vectors? Then I will vote for a single array of starts and sizes are derived which is the way we have now. But this is anyway very subtle.

ye-luo:
@prckent No objections received so far. Go?

PDoakORNL:
I'm all right with this but the estimators are going to be broken until I get around to fixing them. Likely on Wednesday.

ye-luo:
> I'm all right with this but the estimators are going to be broken until I get around to fixing them. Likely on Wednesday.

I updated my comment above. Nothing is broken.

prckent (changes requested):
There are concerns that estimators might be broken. Getting more info, in particular if we have enough tests run in CI there. I would hope we do, but perhaps we don't.

I'll note that all our tests are very simple in terms of # species, groupings etc.

PDoakORNL:
Reading this again I believe that the new estimators are not broken.  I can port them and remove the last of the membersize later this week.

prckent (approved):

prckent:
Approved based on comments above and discussions. Some more complex test cases would give confidence that e.g. normalizations are not changed by using systems with more variable numbers of species.

PR #3808, opened 2022-02-06
emphasize update() in developer manual

Paul-St-Young:
## Proposed changes

Emphasize the need to call `ParticleSet::update()` in the developer manual.
The coexistence of `R` and `RSoA` did not sink in for me until now.
I spent many hours trying to figure out why my unit test distance tables are garbage...
Just want to add some emphasis in the manual to help the next unfortunate soul.

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Documentation changes

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?

workstation

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- No. Code added or changed in the PR has been clang-formatted
- No. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- Yes. Documentation has been added (if appropriate)

PR #3813, opened 2022-02-07
Remove ParticleSet::setCoordinates(R)

ye-luo:
## Proposed changes
This function is neither needed nor safe-to-use (breaks encapsulation).
Thus removal is the best option to avoid confusion like #3808

- Other (please describe): Cleaning

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):

PR #3814, opened 2022-02-07
swap args to H5Sselect_hyperslab for better efficiency

khou2020:

## Proposed changes

Update HDF5 module to call H5Sselect_hyperslab in the intended way.
According to the HDF5 manual (https://support.hdfgroup.org/HDF5/doc/RM/H5S/H5Sselect_hyperslab.htm), the fifth argument (count) means the number of hyper-slabs to select while the sixth argument (block) is the size of each hyper-slab.
In this case, qmcpack is writing a single hyper-slab in a dataset, so the intended way to call H5Sselect_hyperslab is H5Sselect_hyperslab(space_id, H5S_SELECT_SET, offsets, NULL, one, blocks)

- Bugfix

- No

Ubuntu

## Checklist


build.sh
```diff
@@ -1,6 +1,44 @@
 #!/bin/bash
-echo Configuring and building froom inside the build directory.
-echo Check the results of the CMake configuration to ensure that the preferred
-echo compilers and libraries have been selected. See README and documentation 
-echo for guidance.
-cd build; cmake ..; make; cd ..
+
+#BUILD_MODULES=config/load_olcf_summit_modules.sh
```
ye-luo, line 3: Do not touch this file.

src/QMCHamiltonians/ObservableHelper.cpp
```diff
@@ -147,7 +147,10 @@ void ObservableHelper::write(const value_type* first_v, const value_type* /*firs
   if (rank)
   {
     H5Sset_extent_simple(space1_id, rank, &curdims[0], &maxdims[0]);
-    H5Sselect_hyperslab(space1_id, H5S_SELECT_SET, &offsets[0], NULL, &mydims[0], NULL);
+    hsize_t *one = (hsize_t*)malloc(sizeof(hsize_t) * rank);
+    for (int i = 0; i < rank; i++)  one[i] = 1;
```
ye-luo, line 151: const std::vector<hsize_t> one(rank, 1);

ye-luo (changes requested):

src/QMCHamiltonians/ObservableHelper.cpp
khou2020: fixed

build.sh
khou2020: fixed.

src/QMCHamiltonians/ObservableHelper.cpp
```diff
@@ -147,7 +147,8 @@ void ObservableHelper::write(const value_type* first_v, const value_type* /*firs
   if (rank)
   {
     H5Sset_extent_simple(space1_id, rank, &curdims[0], &maxdims[0]);
-    H5Sselect_hyperslab(space1_id, H5S_SELECT_SET, &offsets[0], NULL, &mydims[0], NULL);
+    const std::vector<hsize_t> one(rank, 1);
```
PDoakORNL, line 150: Document this, I just see a vector called one that does seem to have some size and is all ones. It's basically a magic number.  It's sort of explained in your PR but it should be documented here.

PDoakORNL (changes requested):
I'm not following how this isn't a breaking change for anything that reads what is being written here.  Perhaps the HDF5 API is really complex enough that these two calls result in data that can be read for the hd5 file with the same call?

src/QMCHamiltonians/ObservableHelper.cpp
khou2020: > I'm not following how this isn't a breaking change for anything that reads what is being written here. Perhaps the HDF5 API is really complex enough that these two calls result in data that can be read for the hd5 file with the same call?

The net effect is the same. In the origin version, it is telling HDF5 that you are selecting mydims hyper-slabs along each dimensions with each hyper-slab of unit size 1x1x... . In this PR, it is selecting a single hyper-slab of size mydims.

In summary:
H5Sselect_hyperslab (*,*,*,*, count, NULL);
is equivalent (in terms final result) to
H5Sselect_hyperslab (*,*,*,*, one, count);
but they have different high-level meaning, and the later one is closer to what qmcpack intended to do.

khou2020: > Document this, I just see a vector called one that does seem to have some size and is all ones. It's basically a magic number. It's sort of explained in your PR but it should be documented here.

Should I put the explanation as comments before every H5Sselect_hyperslab?
I think it is intuitive given the spec of H5Sselect_hyperslab.
The vector of one means we want to write one hyper-slab (block) along each of the dimensions and each hyper-slab is of size count.
The original version write count hyper-slabs along each dimension and each hyper-slab is of size one.
They lead to the same result but literally means different objective.

prckent: Thanks for these welcome improvements. Perhaps just comment the first usage in each file. Can also put this PR URL.

prckent:
Test this please

prckent (commented):
LGTM but I'll wait for Peter or Ye to approve. CI is running.

PDoakORNL (approved):

wkliao:
Any plan to incorporate this PR?
It improves performance quite a bit.

ye-luo:
> Any plan to incorporate this PR? It improves performance quite a bit.

I was asking @khou2020 performance data at scale and got no rely. If you have alternative data to show, that is also OK.

prckent (approved):
While I would prefer to see some performance data at scale, I also recognize that it can be difficult to come by.

@ye-luo Since this would be easy to unwind in case any problems show up (say, some weirdness on GPFS), how about we merge?

ye-luo:
Since @wkliao mentioned "It improves performance quite a bit." He should have some numbers to tell us. The code seems OK but our CI doesn't check parallel HDF5. I will need to run restart tests manually once.

wkliao:
@khou2020 and I are developing a new HDF5 [log-based VOL](https://github.com/DataLib-ECP/vol-log-based)
and using QMCPACK to evaluate it. The performance difference between
with this PR and without is significant. One read case shows 65 seconds
vs. 1 second.

Just to clarify, this PR is more of correcting QMCPACK's mis-use of 
`H5Sselect_hyperslab`. If you checked the [HDF5 reference manual](https://support.hdfgroup.org/HDF5/doc/RM/H5S/H5Sselect_hyperslab.htm),
it describes the difference of arguments count and block. Also see the example
provide there which helps understand it. When argument stride is NULL, the
argument count used in QMCPACK should be all 1s and argument block
should be the argument count.

`H5Sselect_hyperslab` is not only used in parallel I/O.
It affects serial I/O as well.

ye-luo (approved):
Restart tests passed.

ye-luo:
Test this please

prckent:
Test this please

ye-luo:
The CI failure is unrelated. The Clang14 offload runs do have sporadic failures. I will trigger failed test again.

PR #3816, opened 2022-02-08
Update GCC OpenMP offload settings.

ye-luo:
## Proposed changes
Although GCC still doesn't fully work, some compiler options need update for GCC 12.
GCC 12 is not released, so it is not a hard requirement. Only WARNING is issued.
Also add a warning that GCC OpenMP offload is not ready for users.

- Build related changes

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):

PR #3817, opened 2022-02-08
Disable offload explicitly in GCC

ye-luo:
## Proposed changes

When GCC is built with OpenMP offload support. GCC turns on offload by default. This is undesired.
So turn off offload explicitly. Tested that `-foffload=disable` exists in GCC 7.0 (minimal requirement), 8,9,10,11.

- Build related changes

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):
Wise

PR #3831, opened 2022-02-11
Simplify StructFact resize logic and fix a bug

ye-luo:
## Proposed changes
Bug: the constructor takes first the number of particles and second number of species. But #3359 makes calls with these two parameters swapped. Since the number of particles is only used by StorePerParticle is turned on and turnOnStorePerParticle  actually ignores the internal num_ptlcs. Runs are just fine fine. When StorePerParticle is false, only num_species is used, since the number of particles is always larger than species, it only increases memory usage without any run failure.

When I I worked on moving StructFact::resize, one of the unit test fails and this bug got surfaced. Since I moved the resize anyway, I just get rid of StructFact internal num_ptlcs and num_species fully. Since the resize is moved into the evaluation function, the number K is up-to-date and thus the vulnerability of StructFact to cell change is resolved.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/LongRange/StructFact.cpp
+++ b/src/Particle/LongRange/StructFact.cpp
@@ -45,7 +45,6 @@ StructFact::~StructFact() = default;
 
 void StructFact::resize(int nkpts)
 {
-  phiV.resize(nkpts);
   rhok_r.resize(num_species, nkpts);
   rhok_i.resize(num_species, nkpts);
   if (StorePerParticle)
@@ -53,8 +52,6 @@ void StructFact::resize(int nkpts)
     eikr_r.resize(num_ptcls, nkpts);
     eikr_i.resize(num_ptcls, nkpts);
   }
-  eikr_r_temp.resize(nkpts);
-  eikr_i_temp.resize(nkpts);
 }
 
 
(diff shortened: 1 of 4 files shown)


ye-luo:
Test this please

src/Particle/LongRange/StructFact.cpp
```diff
@@ -120,13 +115,23 @@ void StructFact::computeRhok(const ParticleSet& P)
         rhok_i_ptr[ki] += s;
       }
 #else
-      for (int ki = 0; ki < nk; ki++)
-        phiV[ki] = dot(k_lists_.kpts_cart[ki], pos);
-      eval_e2iphi(nk, phiV.data(), eikr_r_temp.data(), eikr_i_temp.data());
-      for (int ki = 0; ki < nk; ki++)
+      // make the compute over nk by blocks
+      constexpr size_t kblock_size = 512;
```
ye-luo, line 119: I made this change to make phiV, eikr_r_temp, eikr_i_temp fixed size at compile time and lives on the stack.
On CPU this is usually called cache blocking and I will port to GPU similarly later.

prckent: Iff this is profiled to still be a large kernel after offloading, I do wonder if the code inside the loop will need a bit more simplification to gain max benefit.

prckent (approved):

src/Particle/LongRange/StructFact.cpp
ye-luo: It goes to GPU if offload is on. On CPU, I did see performance gain due to cache blocking. The biggest gain to me is getting rid of phiV, eikr_r_temp, eikr_i_temp as class members.

PR #3835, opened 2022-02-13
Replace std::min inside target region

ye-luo:
## Proposed changes
At a point NVHPC was having issue with std::min being used inside offload region. Let's just use a simple omptarget::min and move on.

- No

epyc-server

## Checklist


prckent (approved):

ye-luo:
Test this please

PR #3836, opened 2022-02-13
Early return to avoid undefined behavior.

ye-luo:
## Proposed changes
This was discovered when offloading to AMD GPUs with LLVM/AOMP. libomptarget tries to make a data transfer of size 0. https://github.com/llvm/llvm-project/issues/53803

- Workaround

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/RealSpacePositionsOMPTarget.h
+++ b/src/Particle/RealSpacePositionsOMPTarget.h
@@ -139,6 +139,10 @@ public:
         num_accepted++;
       }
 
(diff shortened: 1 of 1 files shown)


ye-luo:
Test this please

prckent (approved):
Are you confident that there are no other potential zero-size transfers that we need to protect against, or is this just the first one we have clearly hit?

Edit: CI broke - LiH is wrong. OK with the idea of this workaround though.

PR #3838, opened 2022-02-15
Jastrows skip recompute in evaluateGL.

ye-luo:
## Proposed changes
Unlike determinants, Jastrow factors don't suffer from numerical error accumulation and thus recomputing is not needed.
This is consistent with the old choice of evaluateGL being called from updateBuffer by classic drivers.
evaluateLog = recompute() + computeGL;
evaluateGL = computeGL in Jastrow factors. recompute is always skipped.

- Other (please describe): 

- No

epyc-server

## Checklist


ye-luo:
Test this please

PR #3846, opened 2022-02-16
Recompute DTAA at donePbyP on demand only.

ye-luo:
Need #3838 and #3842

## Proposed changes
Introduce DTModes::NEED_FULL_TABLE_ON_HOST_AFTER_DONEPBYP to handle DTAA recompute at donePbyP
on demand.

This is the last bits of putting CoulombPBCAA on the GPU.

- New feature

- No

Summit, Crusher

## Checklist


prckent (commented):
Will approve after #3842 addressed.

prckent (approved):
It would be very good to find a C++ architecture way to automatically handle these optional and state machine like options. This would surface the dependencies at the (say) the per class level. The same could be done for the numerically sensitive/need recompute wavefunction components.

prckent:
Test this please

PR #3851, opened 2022-02-17
Remove uses of size and stride with runtime argument in AFQMC

correaa:
## Proposed changes

Use simplified interface for Multi in AFQMC

- No

Ubuntu 21.10

## Checklist

### DIFF
--- a/src/AFQMC/Estimators/BackPropagatedEstimator.hpp
+++ b/src/AFQMC/Estimators/BackPropagatedEstimator.hpp
@@ -188,14 +188,14 @@ public:
       int nx((walker_type == COLLINEAR) ? 2 : 1);
 
       // 1. check structures
-      if (Refs.size(0) != wset.size() || Refs.size(1) != nrefs || Refs.size(2) != nrow * ncol)
+      if (std::get<0>(Refs.sizes()) != wset.size() || std::get<1>(Refs.sizes()) != nrefs || std::get<2>(Refs.sizes()) != nrow * ncol)
         Refs = mpi3CTensor({wset.size(), nrefs, nrow * ncol}, Refs.get_allocator());
       DeviceBufferManager buffer_manager;
       StaticMatrix detR({wset.size(), nrefs * nx},
                         buffer_manager.get_generator().template get_allocator<ComplexType>());
 
       int n0, n1;
-      std::tie(n0, n1) = FairDivideBoundary(TG.getLocalTGRank(), int(Refs.size(2)), TG.getNCoresPerTG());
+      std::tie(n0, n1) = FairDivideBoundary(TG.getLocalTGRank(), int(std::get<2>(Refs.sizes())), TG.getNCoresPerTG());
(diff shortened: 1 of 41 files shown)


correaa:
Please test and/or merge

correaa:
Please test and/or merge

src/AFQMC/Estimators/BackPropagatedEstimator.hpp
```diff
@@ -188,14 +188,14 @@ class BackPropagatedEstimator : public EstimatorBase
       int nx((walker_type == COLLINEAR) ? 2 : 1);
 
       // 1. check structures
-      if (Refs.size(0) != wset.size() || Refs.size(1) != nrefs || Refs.size(2) != nrow * ncol)
+      if (std::get<0>(Refs.sizes()) != wset.size() || std::get<1>(Refs.sizes()) != nrefs || std::get<2>(Refs.sizes()) != nrow * ncol)
```
ye-luo, line 191: I just feel the code is less readable. What is the gain of switching to compile time?

correaa: Runtime access to dimension index is an was anomaly in Multi that is hard to maintain. Compile-time access also detects errors when compiling; and in fact detected an bug already in AFQMC. The changes look ugly because they were done mechanically (but surely).

ye-luo: Possible to have
```
template<unsigned DIM>
size() { return std::get<DIM>(sizes()); }
```
so we can write `Refs.size<0>()`?

correaa: Yes, it is quite awkward currently in C++ to have static index access. Supposedly in future versions of C++ there will be a `get` member (`m.sizes().get<D>()`). I am debating what to do. The problem with `m.size<D>()` is that `m.size()` is already a function. `D=0` could be the default but it is generate inconsistencies with other functions where the nice default is `D=1` (like `m.rotated()`). The other option is to have `m.sizes_at<D>()` but independently of that it generates a lot of duplicated code, like `m.extensions_at`, `m.strides_at`, `m.offsets_at`.

prckent:
I would like to get this in, or at least the warning reductions part of it. It is worth to keep polishing AFQMC C++ code (@mmorale3).

ye-luo (approved):
We should merge.

ye-luo:
Test this please.

ye-luo:
@correaa some AFQMC tests failed.

correaa:
In short, the idea of "static" access to array dimensions is to allow mixing compile-time and dynamic sizes in the same array. 
For example, In principle `std::get<1>(sizes)` can be a compile-time constant and `std::get<0>(sizes)` can be a dynamic variable.

I am splitting this PR into smaller ones that also do more cleaning:

https://github.com/QMCPACK/qmcpack/pull/3936
https://github.com/QMCPACK/qmcpack/pull/3937

PR #3852, opened 2022-02-17
Resurrect Ewald2D

Paul-St-Young:

## Proposed changes

Resurrect true 2D Ewald sum without using `OHMMS_DIM`.

- New feature

- No

Workstation

## Checklist


prckent:
Paul -

Thanks for looking at this. It would be good to have a proper 2D evaluation... if this is a proper implementation. How thoroughly have you checked this code and its outputs? I think it is safe to say that shortcuts were made in the historical Coulomb code, so before we can advertise 2D Coulomb as working and add examples etc. I think we need to be very thorough with testing, convergence checks, validation etc.

Paul-St-Young:
@prckent  I looked at Madelung energy of square and triangular lattices using the unit cell and select supercells.
The potential energy of the perfect crystal checks out to 8 digits.
I also see reasonable numbers in VMC and DMC runs (which will require a few more PRs to enable).

That said, I'm not comfortable with this becoming a production-level feature right now.
I do want to get to that point in the near future (after more integration tests).

For now, this feature can be enable only via an undocumented flag `<parameter name="ndim"> 2 </parameter>` in `<simulationcell>`. Otherwise, the usual 3D code path is taken, so current users should be unaffected.

prckent:
`I looked at Madelung energy of square and triangular lattices using the unit cell and select supercells.
The potential energy of the perfect crystal checks out to 8 digits.` That is a relief to hear. Can you add some of these to test2D? It is good to capture these checks and establish tests that anyone fiddling with the code in future will have to keep working. An anisotropic case would be good as well.

Paul-St-Young:
I'll need to figure out how to add these to the unit test.
Attached are the inputs and outputs in case I lose them ...

[tb11f2_vmad.zip](https://github.com/QMCPACK/qmcpack/files/8092453/tb11f2_vmad.zip)

jtkrogel:
I would like to discuss a little about the use of `ndim` in the input.  As I understand, the physical systems we can simulate right now are fully 3D, while Ewald2D is meant to apply to systems that have two periodic dimensions.  Is this correct?

If so, I don't think we should use `ndim` in the `<simulationcell/>` input.  Rather, the use of Ewald2D should be triggered by the boundary conditions, i.e. bconds==ppp => Ewald3D while bconds==ppn/pnp/npp =>Ewald2D, and bconds==pnn/npn/nnp => abort.

As a follow-on, does the current code support ppn/pnp/npp or just ppn?

prckent:
`Rather, the use of Ewald2D should be triggered by the boundary conditions, i.e. bconds==ppp => Ewald3D while bconds==ppn/pnp/npp =>Ewald2D, and bconds==pnn/npn/nnp => abort.` Agree. Note that there is no need to expose an option here - if someone wants to reproduce a 3D ion-ion energy they can use ppp. And we should definitely abort if someone tries pnn. This would be consistent with the general plan of having sensible default options for users and avoid/reduce the number of input parameters. Note that some ion-ion Coulomb checks based on the input wavefunctions might need bypassing for 2D, e.g. QE inputs. I think PySCF has proper low dimensional Coulomb handling now.

Paul-St-Young:
@jtkrogel Yes, Ewald2D is meant for a system that's periodic only in 2 directions.
I hard-coded for `ppn`, so this code is not correct for `pnp` and `npp` conditions.

I don't think we can have `ppn` directly trigger Ewald2D, because one could be simulating a quasi-2D slab.
In that case, the EwaldHandlerQuasi2D should be used instead.

jtkrogel:
Could you explain more?  I don't understand the distinction (2D/quasi-2D) if all systems in question exist in a 3D space.

Paul-St-Young:
By true 2D, I mean that all particles stay exactly on the same plane.
By quasi 2D, I mean that particles are much more confined in one of the spatial direction than the others, but can still move around a little bit along that direction. This out-of-plane movement is completely forbidden in the true 2D case.

src/Particle/LongRange/KContainer.cpp
```diff
@@ -111,9 +111,11 @@ void KContainer::BuildKLists(const ParticleLayout& lattice, bool useSphere)
       kvec[0] = i;
       for (int j = -mmax[1]; j <= mmax[1]; j++)
       {
+        if ((ndim <= 1) and (j != 0)) continue;
         kvec[1] = j;
         for (int k = -mmax[2]; k <= mmax[2]; k++)
         {
+          if ((ndim <= 2) and (k != 0)) continue;
```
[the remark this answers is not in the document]
Paul-St-Young: great idea!

Paul-St-Young:
Yes, I want to re-enable strict 2D calculations.

Paul-St-Young:
Basically, `OHMMS_DIM == 2` without having to recompile.

jtkrogel:
That may generate separate discussion.  On a different note, do you know if EwaldHandlerQuasi2D is correct/operational?

Paul-St-Young:
I got `EwaldHandlerQuasi2D` to run, but it produced wrong numbers in my true 2D tests.
That's why I made a separate `Ewald2D` class, which is much simpler.
I could be using the quasi-2D version wrong though

Paul-St-Young:
@prckent  what I'm working towards is exactly what you mentioned in your second point:
 support 2D by restricting z in a non-hackish way
I am happy to make a fresh issue to track progress towards this goal.
I can implement the necessary changes in the batch driver directly.

Regarding your first point, I'm not an expert on the quasi-2D version. My understanding is that there are extra subtleties compared to the true 2D case, but it should reduce to the true 2D case when z=0.
The code in this PR is very straightforward, so I think it's a good check to have as the quasi-2D version is being worked on.

Paul-St-Young:
I created #3868 to track progress towards strict 2D capability.
Is there anything else that need to be addressed before this PR could be merged?

prckent:
Paul - hopefully you can see the test failures, but there are SEGVs in a couple of the tests. I'll be happy to retrigger our self-hosted CI if you ask &/or when the github actions ones are passing.

```
The following tests FAILED:
	 28 - deterministic-unit_test_hamiltonian_coulomb (SEGFAULT)
	 29 - deterministic-unit_test_hamiltonian_force (SEGFAULT)
```

Paul-St-Young:
I should also say that the quasi-2D Ewald sum is harder to optimize than the strict-2D version, so I think we should keep both the strict-2D and the quasi-2D  versions.
For quasi-2D, each term in the k-space sum depends on the particle separation perpendicular to the slab, i.e. `vlr_k` is a function of `z=|z_i-z_j|`, so \sum_{i,j} vlr_k e^{i k (r_i-r_j)} does NOT reduce to vlr_k \rho_k \rho_{-k}.

Whether this can speed up current slab calculations remains to be seen.

prckent:
Test this please

prckent:
> On second thought, `ewald_quasi2d` itself should be sufficient. I'd like to reserve the `ndim=2` input for strict-2D.

Why not `ewald_true2d` or similar? i.e. Call this mode of runing "true 2D". This makes specifying ndim unnecessary. It should force ndim=2. It is also a bit clearer for the user.

As this develops I can imagine wanting to specify "true 2D" earlier in the input, since finding out at the point the long range interactions are specified is quite late.

> In any case, I think the old way of using `qmc_common.use_ewald` from an input option should be deprecated.

Yes, this should go. Please just delete it. AFAIK there are zero known historic 2D inputs to be concerned about, and we would have no reasons to trust them in any case.

src/QMCHamiltonians/CoulombPBCAA.h
```diff
@@ -75,6 +75,8 @@ struct CoulombPBCAA : public OperatorBase, public ForceBase
   Vector<ComplexType> del_eikr;
   /// Flag for whether to compute forces or not
   bool ComputeForces;
+  /// Flag for whether to use quasi-2D Ewald
+  bool Quasi2D;
```
PDoakORNL, line 79: This looks like a type give it a variable format name.

src/Particle/LongRange/KContainer.h
```diff
@@ -72,14 +72,14 @@ class KContainer : public QMCTraits
...
-  void updateKLists(const ParticleLayout& lattice, RealType kc, bool useSphere = true);
+  void updateKLists(const ParticleLayout& lattice, RealType kc, unsigned ndim, bool useSphere = true);
 
   const auto& get_kpts_cart_soa() const { return kpts_cart_soa_; }
 private:
   /** compute approximate parallelpiped that surrounds kc
    * @param lattice supercell
    */
-  void FindApproxMMax(const ParticleLayout& lattice);
+  void FindApproxMMax(const ParticleLayout& lattice, unsigned ndim);
```
PDoakORNL, line 82: if you have the time fix the formatting of this method name since you are changing it.

PDoakORNL (changes requested):
Documentation, use proper formating for new variable and method names.

Paul-St-Young:
@prckent 
> Why not `ewald_true2d` or similar? i.e. Call this mode of running "true 2D". This makes specifying ndim unnecessary. It should force ndim=2. It is also a bit clearer for the user.

Yes, I think `ewald_strict2d` is a good name.
Good idea to automatically set `ndim=2` in this case!

> Yes, this should go. Please just delete it. AFAIK there are zero known historic 2D inputs to be concerned about, and we would have no reasons to trust them in any case.
I will delete the option and add more unit tests to make cover more 2D cases.

src/QMCHamiltonians/CoulombPBCAA.cpp
```diff
@@ -436,11 +440,41 @@ CoulombPBCAA::Return_t CoulombPBCAA::evalConsts(bool report)
...
+        throw std::runtime_error("Quasi2D assumes same charge");
+    if (report)
+    {
+      app_log() << "    vlr(r->0) = " << vl_r0 << std::endl;
+      app_log() << "   1/V vsr_k0 = " << vs_k0 << std::endl;
+    }
+    // make sure we can ignore the short-range Madelung sum
+    mRealType Rws = Ps.getLattice().WignerSeitzRadius;
+    mRealType rvsr_at_image = Rws*AA->evaluate(Rws, 1.0/Rws);
+    if (rvsr_at_image > 1e-6)
```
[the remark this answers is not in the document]
Paul-St-Young: I don't have an example where the default Ewald `alpha = sqrt(kc/(2*rc))` fails this condition.
This is just a defensive check in case someone runs into the weird case in the future.

Paul-St-Young:
@prckent I removed the `use_ewald` option and added enough 2D tests to be confident of both the `strict2d` and `quasi2d` handlers.
I think this PR is ready to be merged.
Could you trigger the remaining CIs please?

@PDoakORNL I think I addressed all your comments. Please see if this satisfactory now.

markdewing:
Test this please

src/Particle/Lattice/LRBreakupParameters.h
```diff
@@ -31,9 +31,10 @@ class LRBreakupParameters<T, 3>
   T LR_rc;
   T LR_kc;
   T LR_tol;
+  unsigned ndim;
```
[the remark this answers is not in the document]
ye-luo: three slash instead of two. Also say a bit more. ''3D, quasi 2D, ndim =3. Strict 2D, ndim=2"

src/Particle/LongRange/EwaldHandlerQuasi2D.h
```diff
@@ -2,133 +2,74 @@
 // This file is distributed under the University of Illinois/NCSA Open Source License.
 // See LICENSE file in top directory for details.
 //
-// Copyright (c) 2020 QMCPACK developers.
+// Copyright (c) 2022 QMCPACK developers.
 //
-// File developed by: Jeongnim Kim, jeongnim.kim@gmail.com, University of Illinois at Urbana-Champaign
-//                    Jeremy McMinnis, jmcminis@gmail.com, University of Illinois at Urbana-Champaign
-//                    Mark A. Berrill, berrillma@ornl.gov, Oak Ridge National Laboratory
+// File developed by: Yubo "Paul" Yang, yubo.paul.yang@gmail.com, CCQ @ Flatiron
```
ye-luo, line 7: Did you create every line from scratch. If not, it is better to keep the old author list.

src/QMCHamiltonians/tests/test_ewaldq2d.cpp
```diff
@@ -0,0 +1,319 @@
+#include "catch.hpp"
```
ye-luo, line 1: test_ewaldquasi2d.cpp is more readable.

ye-luo (commented):
Do you have data to show 3D expanding converge to quasi2D value?

src/QMCHamiltonians/CoulombPBCAA.h
```diff
@@ -75,6 +75,8 @@ struct CoulombPBCAA : public OperatorBase, public ForceBase
   Vector<ComplexType> del_eikr;
   /// Flag for whether to compute forces or not
   bool ComputeForces;
+  /// Flag for whether to use quasi-2D Ewald
+  bool quasi2d;
```
ye-luo, line 79: This should be const with its value being initialized by the constructor member initializer lists.

Paul-St-Young:
> Do you have data to show 3D expanding converge to quasi2D value?

No, I do not have a 3D input with expanding z direction.

src/Particle/LongRange/EwaldHandlerQuasi2D.h
Paul-St-Young: I did rewrite the whole thing.
Although, I didn't intend to erase the historic contributions.
They have been restored.

Paul-St-Young:
On second thought, 3D ppn with large z will not match the quasi 2D value because they have different neutralizing backgrounds.
The 3D positive background charge density for an N-electron simulation cell is +e*N/V, i.e. \sum_i +e/V uniformly spread out over the entire cell volume V, whereas the quasi 2D background charge density is \sum_i +e/A * \delta(z-z_i), i.e. one plane of 2D positive charge at the z height of each electron. A is the 2D cell area.

PDoakORNL (changes requested):
It looks like my previous review was addressed where possible for this PR except for this.

src/QMCHamiltonians/CoulombPBCAA.h
ye-luo: I guess you changed the initialization but forgot to add const.

src/Particle/LongRange/EwaldHandlerQuasi2D.h
ye-luo: Thanks. When I read the code, "EwaldHandlerQuasi2D::evaluate_slab" before and after look very similar. So I felt better to keep the historic contributions.

ye-luo:
Could you also document the new option value in the manual?

Paul-St-Young:
@PDoakORNL  I clarified the exception message and added a test with CHECK_THROWS.
Turns out I just had to set a very small `LR_dim_cutoff` to get a violation of my condition.
The short range part being < 1e-6 at the WSR should apply to any LRHandler.
Although the message about alpha is specific to the Ewald handler.

Paul-St-Young:
@ye-luo I added documentation for `LR_handler` in the manual.

src/Particle/LongRange/KContainer.cpp
```diff
@@ -79,17 +79,23 @@ void KContainer::FindApproxMMax(const ParticleLayout& lattice)
...
+
+  //overwrite the non-periodic directon to be zero
+  if (lattice.SuperCellEnum == SUPERCELL_SLAB)
+  {
+    app_log() << "  No kspace sum perpendicular to slab " << std::endl;
+    mmax[2] = 0;
+  }
+  if (ndim < 3)
+  {
+    app_log() << "  No kspace sum along z in strict 2D " << std::endl;
```
ye-luo, line 95: The printout doesn't seem accurate ndim=1 also hit here.

src/QMCHamiltonians/tests/test_ewald_quasi2d.cpp
```diff
@@ -0,0 +1,351 @@
...
+  int upIdx                  = tspecies.addSpecies("u");
+  int chargeIdx              = tspecies.addAttribute("charge");
+  int massIdx                = tspecies.addAttribute("mass");
+  tspecies(chargeIdx, upIdx) = -1;
+  tspecies(massIdx, upIdx)   = 1.0;
+
+  elec.createSK();
+  elec.addTable(elec); // !!!! crucial to compute distance table
+
+  CoulombPBCAA caa = CoulombPBCAA(elec, true, false, false);
```
ye-luo, line 331: `CoulombPBCAA caa (elec, true, false, false);` more concise.

ye-luo (commented):
Only minor things left.

Not directly related to this PR. I noticed quasi2D doesn't seem to be connected in PBCAB.

ye-luo (approved):
LGTM

ye-luo:
Test this please

ye-luo:
@PDoakORNL OK to merge?

prckent (approved):
It think that this meets easily meet the merge criteria now, and it can clearly be refined further as/if needed. Thanks for enduring the review Paul. I think that the code is better for it.

PDoakORNL:
@Paul-St-Young satsified my important concerns.

PR #3858, opened 2022-02-19
Make the inter std::map of pools access from outside as const

ye-luo:
## Proposed changes
Make the inter std::map of PaticleSetPool, WaveFuncttionPool const when being accessed from outside.
Also in WaveFunctionPool remove setPrimary. This is handled via addFactory function.

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):

PR #3859, opened 2022-02-20
Replace vector of SPOSetBuilder with map of SPOSet

ye-luo:
## Proposed changes
SPOSetBuilderFactory keeps an array of SPOSetBuilder pointers and each SPOSetBuilder contains a map of SPOSet. To make all the SPOSets available for lookup later, SPOSetBuilderFactory is persist and thus all the builders are persistent thought out the the whole run. This PR makes SPOSetBuilder short-lived. Once all the SPOSets from a collection are processed by a SPOSetBuilder, there is no need to keep the builder. Thus the map of SPOSets are directly kept in SPOSetBuilderFactory class.

- No

epyc-server

## Checklist


ye-luo:
Test this please

ye-luo:
Test this please

prckent (approved):

Issue #3756, opened 2022-01-24
Integrate rocsolver for LU and inversion


ye-luo:
**Is your feature request related to a problem? Please describe.**
After #3755 in DelayedUpdateCUDA.h, the LU and inversion solver is ifdef to run on the host on AMD GPU. Due to the fact that rocsolver API differs from cusolver, we will need rocSolverInverter equivalent to cuSolverInverter.

**Describe the solution you'd like**
Note that rocSolver buffer handling API doesn't look the same as cuSolver.
Also need a unit test to cover both cuSolverInverter and rocSolverInverter.

markdewing:
Trying to get straight all the different inversion methods, options, and code paths.

They get set up in SlaterDetBuilder::putDeterminant

QMC_CUDA - legacy CUDA acceleration in DiracDeterminantCUDA.  Rest of this applies if QMC_CUDA is not defined.

input options:
* 'matrix_inverter' is 'gpu' or 'host'.  Defaults to 'gpu'. (converted to matrix_inverter_kind which is DetMatInvertor::HOST or DetMatInvertor::ACCEL)
* use_batch  (option name is 'batch').  Defaults to 'yes' if ENABLE_OFFLOAD is defined, defaults to 'no' otherwise
* useGPU (option name is 'gpu').  Only available if ENABLE_CUDA or ENABLE_OFFLOAD are defined. Defaults to 'yes'.

The selections:
* use_batch is 'yes'
  * useGPU is 'yes' and  ENABLE_CUDA and ENABLE_OFFLOAD
    * `DiracDeterminantBatched<MatrixDelayedUpdateCUDA>` 
    *  "Running on NVIDIA GPU via CUDA acceleration and OpenMP offload"
  * any of the previous options not set
    * `DiracDeterminantBatched<MatrixUpdateOMPTarget>`
    * "Running on an accelerator via OpenMP offload. Only SM1 update supported. delay_rank is ignored"
* use_batch is 'no'
  * useGPU is 'yes and ENABLE_CUDA
    * `DiracDeterminant<DelayedUpdateCUDA>`
    * "Running on an NVIDIA GPU via CUDA acceleration"
  * useGPU is 'no' (or ENABLE_CUDA is not defined)
    * `DiracDeterminant<DelayedUpdate>`
    * "Running on CPU"

Looking at the classes to see how the inversion method fits in
[...]

PR #3861, opened 2022-02-21
Add unit test and micro benchmark for cuSolverInverter

markdewing:
* Rename test_dirac_matrix.cpp to test_DiracMatrix.cpp for consistency.
* Extract some of the matrix values from test_DiracMatrix.cpp to createTestMatrix.hpp for reuse.
* Add unit test test for cuSolverInverter
* Add micro benchmark for cuSolverInverter

Partial implementation of #3756

- New feature

- No

laptop

## Checklist


src/QMCWaveFunctions/tests/createTestMatrix.h
```diff
@@ -0,0 +1,69 @@
...
+  b(1, 0) = 0.07923894288;
+  b(1, 1) = 0.1496231042;
+  b(1, 2) = -0.1428117337;
+  b(2, 0) = -0.2974298429;
+  b(2, 1) = -0.04586322768;
+  b(2, 2) = 0.3927890292;
+}
+
+// Log of determinant of test matrix 1
+inline double testMatrix1_logDet() { return 3.78518913425; }
```
ye-luo, line 66: I think it s better to write a class with static functions to present the above three Matrix1 related classes as a whole.

src/QMCWaveFunctions/tests/test_cuSolverInverter.cpp
```diff
@@ -0,0 +1,78 @@
...
+namespace qmcplusplus
+{
+
+using LogValueType = std::complex<double>;
+
+TEMPLATE_TEST_CASE("cuSolverInverter", "[wavefunction][fermion]", double, float)
+{
+#ifdef QMC_COMPLEX
+  using FullPrecValueType = std::complex<double>;
+  using ValueType         = std::complex<TestType>;
```
ye-luo, line 31: Where is TestType defined?
Please use the latest type convention. Value instead of ValueType.

ye-luo (changes requested):

src/QMCWaveFunctions/tests/test_cuSolverInverter.cpp
markdewing: TestType comes from Catch.  I'm using a TEMPLATE_TEST_CASE to test float and double versions in the same test case. I added a comment about it.

src/QMCWaveFunctions/tests/createTestMatrix.h
markdewing: Good idea - fixed.

ye-luo (approved):
LGTM

ye-luo:
Test this please

prckent (commented):
It would be good if over time this could morph into a general "test Inverter", covering roc and one and well as cu, perhaps also the cpu only path. 

Hopefully it won't be the case, but straightforward extrapolation of the last six months data would indicate we can expect inverters from every vendor to contain bugs that we hit...

ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
LGTM. needs non-ANL approval.

prckent (approved):

PR #3863, opened 2022-02-21
Relax MaxKshell range in mixed precision.

ye-luo:
## Proposed changes
Noticed larger range with OneAPI compiler.

- No

dell-laptop

## Checklist


ye-luo:
Test this please

prckent (approved):
LGTM

(Do we trust the OneAPI compiler yet?)

ye-luo:
We don't have choices. If it shows up as an issue, we need file bug reports. So far no. CPU-only is doing pretty well in nightly tests.

PR #3870, opened 2022-02-23
Initial transition guide for batched drivers

### LINKED ISSUE #3865: Need high level "transition guide" for batched drivers
Need some documentation for current users to help them moving from the classic drivers to batched drivers.

ye-luo:
## Proposed changes
Provide the key info to help existing users migrating to batched drivers.
Closes #3865

- Documentation changes

- No

ryzen-box

## Checklist


markdewing:
To check my understanding:

The importance of `walkers_per_rank` is that it controls the memory usage on the device, which is usually the primary constraint.  The general performance recommendation is to have as many walkers on the GPU as will fit in the GPU memory.   The performance per walker approaches the best value asymptotically as (1/N_walkers), so it's sufficient to just get enough walkers on the GPU to give good enough performance (that is, the number of walkers per GPU doesn't need to be highly optimized.)

Multiple crowds divide up the walkers on the GPU into smaller batches in order to fill the GPU better.  Different crowds run asynchronously in different streams, allowing communication (from one crowd) to overlap with computation (on a different crowd).  (This is similar to how MPS is used, but with multiple processes instead of multiple threads/streams).   Dividing the number of walkers into crowds lowers the performance per walker for that crowd, but the increase in throughput for multiple crowds in better filling the GPU more than makes up for that loss.

For load balancing the work between crowds, the number of walkers per rank should be evenly divisible by the number of crowds.  The code will print a warning if this is not the case.

ye-luo:
> To check my understanding:
> 
> The importance of `walkers_per_rank` is that it controls the memory usage on the device, which is usually the primary constraint.

correct. GPU users needs to pay attention to this.

> The general performance recommendation is to have as many walkers on the GPU as will fit in the GPU memory. The performance per walker approaches the best value asymptotically as (1/N_walkers), so it's sufficient to just get enough walkers on the GPU to give good enough performance (that is, the number of walkers per GPU doesn't need to be highly optimized.)

No need to push to the memory capacity limit. Only needs to be enough. If users care about time-to-solution, may be 80% of the asymptotic limit is enough. If users care more about throughput, may be 95% is considered enough.

> Multiple crowds divide up the walkers on the GPU into smaller batches in order to fill the GPU better. Different crowds run asynchronously in different streams, allowing communication (from one crowd) to overlap with computation (on a different crowd). (This is similar to how MPS is used, but that uses multiple processes instead of multiple threads/streams). Dividing the number of walkers into crowds lowers the performance per walker for that crowd, but the increase in throughput for multiple crowds in better filling the GPU more than makes up for that loss.

Correct. Consider users may also mix CPU-only features, crowds benefit from more CPU cores being used.
To be more accurate, if compute per kernel is low, there is also concurrent kernel execution.

> For load balancing the work between crowds, the number of walkers per rank should be evenly divisible by the number of crowds. The code will print a warning if this is not the case.

Correct. If not hitting memory capacity equal crowd size is preferred to reduce imbalance. When there is memory capability limit, we still would like send as many walkers as possible. So imperfect load balance between crowds is OK.

docs/methods.rst
```diff
@@ -114,6 +114,39 @@ To continue a run, specify the ``mcwalkerset`` element before your VMC/DMC block
 
 In the project id section, make sure that the series number is different from any existing ones to avoid overwriting them.
 
+.. _transition_guide:
+
+Transition guide for batched drivers
+------------------------------------
+Under the Exascale Computing Project effort, we developed a new set of QMC drivers, called "batched drivers",
+with a unique feature of batching walkers on demand. Available drivers are ``vmc_batch``, ``dmc_batch`` and ``linear_batch``.
```
markdewing, line 122: What does "on demand" mean here?  What difference does is make that batching walkers is on demand or not?  (I'm trying to clarify what this means for users)

docs/methods.rst
```diff
@@ -114,6 +114,39 @@ To continue a run, specify the ``mcwalkerset`` element before your VMC/DMC block
...
+A new concept "crowd" is introduced as a suborganization of walker population in batched drivers.
+Walkers within a crowd operate their computation in lock-step, which helps the GPU efficiency.
+Walkers between crowds remain fully asynchronous unless operations involving the full population are needed.
+With this batching capability, new drivers enable feature implementations potentially to maximize the performance of a given hardware.
+For OpenMP GPU offload users, batched drivers are musts to effectively use GPUs.
+Keep in mind that batched drivers are GPU agnostic even though they incorporate the need of GPU computing.
```
markdewing, line 128: Does "GPU agnostic" mean "doesn't care which brand of GPU - Nvidia, AMD, or Intel"?  Or does it mean "doesn't care if using the GPU or CPU"?

prckent (changes requested):
As a priority, this text will need updating to give more perspective on the batched drivers (they are the future for everyone and the old drivers will eventually be removed, even for CPU runs), and then likely some clearer examples. I would prefer that we always lead with the physics settings before discussing computational settings/optimizations. Further optimization can be done in later PRs.

docs/methods.rst
ye-luo: Will make unambiguous

markdewing:
It might helpful if we discussed what we want to convey to users about the batched driver.

* Developed by ECP effort
* Single source base for CPU and GPU code (major problem with the legacy CUDA version)
  * Features can developed on the CPU and later extended to GPU for performance
* GPU is implemented primarily by OpenMP offload
   * Vendor-specific code can be used for higher performance
* Model of how driver works
* Introduction of crowd (further division of walkers)
   * For CPU, number of crowds is the number threads, and there is one walker per crowd
   * For GPU, one crowd and all the walkers in a crowd
   * More refinements of these for additional performance, but these are some concrete examples to orient people

ye-luo:
OpenMP offload and batched drivers are two separate features although the former needs the later.

- [x] Under Features of QMCPACK
  add OpenMP offload GPU port and compare it to legacy CUDA.
- [ ] Running QMCPACK, Running on GPU machines
  add OpenMP offload GPU port discussion
- [ ] Introduce batched drivers. explain concepts. to be added by this PR.
- [x] Transition guide
  highlight what user needs to pay attention.
- [x] each XXX_batch driver has its own dedicated manual section. mostly complete.

here is how the method chapter is organized.

Quantum Monte Carlo Methods
  -- some documentation applicable to all drivers.
  -- Introduce batched drivers. explain concepts like crowds. (not yet added)
  -- Transition guide for existing users.
  -- VMC, batch docs
  -- WFOpt and batch docs
  -- DMC and batch docs
  -- RMC docs

concepts applicable to multiple methods are placed before individual methods. I think the 

@prckent 
> give more perspective on the batched drivers

Will add Introduce batched drivers. More focus on concept not performance. Will only mention needed for GPU runs.

> I would prefer that we always lead with the physics settings before discussing computational settings/optimizations.

how would you like to reorganize "Quantum Monte Carlo Methods" ?

docs/features.rst
```diff
@@ -109,44 +111,74 @@ description of the available features, see  :ref:`afqmc`.
...
 
-   #. Single Slater determinants with 3D B-spline orbitals.
-      Twist-averaged boundary conditions and complex wavefunctions are
-      fully supported. Gaussian type orbitals are not supported.
 
-   #. Hybrid mixed basis representation in which orbitals are
-      represented as 1D splines times spherical harmonics in spherical
-      regions (muffin tins) around atoms and 3D B-splines in the
-      interstitial region.
+The existing implementation supports multiple GPUs per node, with MPI tasks assigned in a round-robin order to the GPUs.
```
markdewing, line 132: In my mind, "supporting multiple GPUs per node" implies supporting multiple GPU's per MPI rank.  Supporting one MPI rank for 1 GPU seems like it ought to be the bare minimum - any code that does MPI and uses a GPU should work in that mode.

ye-luo: This was an old line. Updated it and added more detail about how the primary GPU is assigned under ` Running on GPU machines `

docs/methods.rst
```diff
@@ -114,6 +114,58 @@ To continue a run, specify the ``mcwalkerset`` element before your VMC/DMC block
...
+---------------
+
+Under the Exascale Computing Project effort a new set of QMC drivers was developed
+to eliminate the divergence of CPU and GPU code paths at the QMC driver level and make the drivers CPU/GPU agnostic.
+The divergence came from the the fact that the CPU code path favors executing all the compute tasks within a step
+for one walker and then advance walker by walker. Multiple CPU threads process their own assigned walkers in parallel.
+In this way, walkers are not synchronized with each other and maximal throughout can be achieved on CPU.
+The GPU code path favors executing the same compute task over all the walkers together to maximize GPU throughput.
+This GPU code path choice also minimizes the overhead of dispatching computation and host-device data transfer.
+However, there is only one host thread responsible for handling all the interaction between the host and GPUs.
```
markdewing, line 129: What does this sentence about "only one host thread" mean?  What consequences are there for the user?

PDoakORNL: I'm not sure, In a proper CUDA/ROCM implementation of almost anything and in the case of the direct inversion, each crowd has a CUDA stream so you have num crowd independent CPU/GPU communications. If the OMPTarget runtime fails to do such a thing i.e. it doesn't have multiple worker threads doing async transfers and kernel launches that tie into to their task and data dependency abstractions, thats their failure.

PDoakORNL (changes requested):
The the relationship between threading and GPU streams real and actual needs to be clear.

docs/methods.rst
ye-luo: I was describing the legacy implementation. Updated.

PDoakORNL (approved):

ye-luo:
Test this please

prckent (approved):

PR #3872, opened 2022-02-23
InputNode and tests

PDoakORNL:
## Proposed changes

Delegated Input classes will be owned by their managing input class. For purposes of access they will be accessed through a vector of variant<refwrapper<InputClassA>, refwrapper<InputClassB>,... in the managing class. Since input classes can potentially be quite large it seems better to separate ownership and access. For that reason the ownership of the input class is through a vector of UniquePointers to InputNode a mixin interface that provides a virtual destructor for the input classes.

- New feature
- Documentation changes

- No

Leconte 

## Checklist

### DIFF
--- a/src/io/InputNode.hpp
+++ b/src/io/InputNode.hpp
@@ -1,0 +1,35 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2022 QMCPACK developers.
+//
+// File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Lab
+//
+// File created by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Lab
+//////////////////////////////////////////////////////////////////////////////////////
+
+#ifndef QMCPLUSPLUS_INPUTNODE_HPP
+#define QMCPLUSPLUS_INPUTNODE_HPP
+
+#include <memory>
+
+namespace qmcplusplus
+{
+
+/** Interface to allow input nodes to be type erased for the purposes of ownership
+ */
+struct InputNode
+{
+  virtual ~InputNode(){};
+  template<typename T, typename VOWN, typename VINPUT, typename... Args>
+  static void append(VOWN& vown, VINPUT& vinput, Args&&... args) {
+    vown.push_back(std::make_unique<T>(std::forward<Args>(args)...));
+    vinput.push_back(static_cast<T&>(*vown.back()));
+  }
+};
+
+
+  
+} // namespace qmcplusplus
+#endif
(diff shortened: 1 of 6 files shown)


src/io/InputNode.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+namespace qmcplusplus
+{
+
+/** Interface to a class to be type erased for the purposes of ownership
+ *  To allow a vector of unique_ptr to InputNode to contain a heterolist of dynamic objects
+ */
+struct InputNode
+{
+  virtual ~InputNode(){};
```
ye-luo, line 25: `= default` is better than `{}`

src/io/InputNode.hpp
```diff
@@ -0,0 +1,36 @@
...
+  template<typename T, typename VOWN, typename VINPUT, typename... Args>
+  static void append(VOWN& vown, VINPUT& vinput, Args&&... args)
```
ye-luo, line 27: It doesn't seem to be related to `InputNode` at all.
Should vown be of type `std::vector<std::unique_ptr<InputNode>>` explicitly?

src/type_traits/template_types.hpp
```diff
@@ -28,6 +28,10 @@ namespace qmcplusplus
  *  see: UPtrVector
  *  @{
  */
+
+template<typename T>
+using RefW = std::reference_wrapper<T>;
```
ye-luo, line 33: `RefWrapper` is the shortest I can take. `RefW` is so confusing.

src/io/tests/test_InputNode.cpp
```diff
@@ -0,0 +1,56 @@
...
+  std::string name{"inputB"};
+  short sA{1};
+};
+
+using ChildInput = std::variant<RefW<InputA>, RefW<InputB>>;
+
+TEST_CASE("InputNode_BasicUse", "[io]")
+{
+  UPtrVector<InputNode> input_ownership;
+  std::vector<ChildInput> inputs;
```
ye-luo, line 48: I don't feel good to keep both input_ownership and inputs around.
Also don't see `variant` brings simplification.

```
enum class {
   DENISTY,
   ESTB
} EstimatorType;

template<typename ENUM>
InputNode
{
   ENUM type;
};

UPtrVector<InputNode<EstimatorType>> input_ownership;
```

In this way, we know the type with some restrictions also. Only need input_ownership.
Also header file only need to see the enum class instead of all the class definition.

PDoakORNL: Header file for EstimatorManagerInput doesn't see all the definitions the way I wrote it forward declaration is fine.  I can make this just use variant, the memory waste isn't probably worth worrying about.

ye-luo: No memory concern. My concern was that keeping tracking of two vector separately is error-prone

PDoakORNL: I think that is legitimate and going back and dropping the separate storage works well.  This was a premature optimization.

PDoakORNL:
Input node likely premature memory optimization

Issue #3849, opened 2022-02-17
opt_crowd_size default value


ye-luo:
**Is your feature request related to a problem? Please describe.**
Currently 1 which is a poor choice, running all the batches with batch size 1.
Preferred way, if user doesn't specify, align with VMC choice.

ye-luo:
This input parameter is painful to control. Just like walkers, we don't control walkers per crowd but only `walker_per_rank`.

ye-luo:
`walkers_per_rank=1792`, 7 threads (7 crowds) in VMC
```
<opt stage="setup">
  <log>
   Reading configurations from h5FileRoot
  Using Nonlocal PP in Opt: no
  VMC Eavg = -1.4151e+03
  VMC Evar = 4.2034e+02
  Total weights = 1.7203e+06
  Execution time = 3.3972e+02
  </log>
</opt>
```

After adding add `opt_crowd_size=256`, >10x difference.
```
<opt stage="setup">
  <log>
   Reading configurations from h5FileRoot
  Using Nonlocal PP in Opt: no
  VMC Eavg = -1.4151e+03
  VMC Evar = 4.2034e+02
  Total weights = 1.7203e+06
  Execution time = 3.1634e+01
  </log>
</opt>
```

ye-luo:
A second thought on this. If we directly borrow the crowds from VMC, we neither need to care about input nor concern about additional memory usage by WFOpt.

prckent:
Option 3:

Perhaps the crowd count could be output and as well as input to the drivers. The first QMC driver executed (whatever it is) gets to pick a default or take a value from input. This value is passed out. The next driver gets this as input. If the XML for the driver's QMC section has a different value, that overrides.

Option 4:

The application has a single central function to go to for default values of these parallelization settings. This means the logic is not replicated (or even different) on a per driver basis unless there is a very good reason.

Discussion:

This raises the question of whether (and why) the optimizer will routinely need different crowd sizes from VMC/DMC.

We are going to need some logic to automagically pick walker counts and other physics/simulation settings as well as parallelization settings, so what we learn about crowd settings can factor in to how we treat these other settings e.g. If drivers need very different values, we'll definitely need to set them automagically ourselves since users will not be able to reasonably choose optimal values.

markdewing:
The optimal number of walkers per crowd could be different for the optimizer cost function and VMC, since the operations are different. However, I suspect the VMC value will work pretty well as a default (better than the current default of 1).

[...]

PR #3873, opened 2022-02-23
Split walkers among crowds in optimizer to match VMC usage

markdewing:
Addresses #3849 

The code now uses walker_per_crowd computed the same way as VMC.

Previously the batched cost function code expected every crowd would have the same number walkers. Most of the changes come from updating the code to use different number of walkers per crowd.

- Bugfix

- No

laptop, Summit

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
@@ -294,7 +294,7 @@ void QMCCostFunctionBatched::checkConfigurations()
                           bool needGrads, bool compute_nlpp, const std::string& includeNonlocalH) {
     CostFunctionCrowdData& opt_data = *opt_crowds[crowd_id];
 
-    int local_samples = samples_per_crowd_offsets[crowd_id + 1] - samples_per_crowd_offsets[crowd_id];
+    const int local_samples = samples_per_crowd_offsets[crowd_id + 1] - samples_per_crowd_offsets[crowd_id];
     int num_batches;
     int final_batch_size;
 
@@ -306,7 +306,7 @@ void QMCCostFunctionBatched::checkConfigurations()
       if (inb == num_batches - 1)
         current_batch_size = final_batch_size;
 
-      int base_sample_index = inb * walkers_per_crowd[crowd_id] + samples_per_crowd_offsets[crowd_id];
+      const int base_sample_index = inb * walkers_per_crowd[crowd_id] + samples_per_crowd_offsets[crowd_id];
 
       auto wf_list_no_leader = opt_data.get_wf_list(current_batch_size);
       auto p_list_no_leader  = opt_data.get_p_list(current_batch_size);
@@ -365,7 +365,7 @@ void QMCCostFunctionBatched::checkConfigurations()
 
         for (int ib = 0; ib < current_batch_size; ib++)
         {
-          int is = base_sample_index + ib;
+          const int is = base_sample_index + ib;
           for (int j = 0; j < nparam; j++)
           {
             DerivRecords[is][j]  = std::real(dlogpsi_array.getValue(j, ib));
@@ -377,8 +377,8 @@ void QMCCostFunctionBatched::checkConfigurations()
 
         for (int ib = 0; ib < current_batch_size; ib++)
         {
-          int is    = base_sample_index + ib;
-          auto etmp = energy_list[ib];
+          const int is = base_sample_index + ib;
+          auto etmp    = energy_list[ib];
           opt_data.get_e0() += etmp;
           opt_data.get_e2() += etmp * etmp;
 
@@ -402,8 +402,8 @@ void QMCCostFunctionBatched::checkConfigurations()
 
         for (int ib = 0; ib < current_batch_size; ib++)
         {
-          int is    = base_sample_index + ib;
-          auto etmp = energy_list[ib];
+          const int is = base_sample_index + ib;
+          auto etmp    = energy_list[ib];
           opt_data.get_e0() += etmp;
           opt_data.get_e2() += etmp * etmp;
 
@@ -530,7 +530,7 @@ QMCCostFunctionBatched::Return_rt QMCCostFunctionBatched::correlatedSampling(boo
     CostFunctionCrowdData& opt_data = *opt_crowds[crowd_id];
 
 
-    int local_samples = samples_per_crowd_offsets[crowd_id + 1] - samples_per_crowd_offsets[crowd_id];
+    const int local_samples = samples_per_crowd_offsets[crowd_id + 1] - samples_per_crowd_offsets[crowd_id];
 
     int num_batches;
     int final_batch_size;
@@ -544,7 +544,7 @@ QMCCostFunctionBatched::Return_rt QMCCostFunctionBatched::correlatedSampling(boo
         current_batch_size = final_batch_size;
       }
 
-      int base_sample_index = inb * walkers_per_crowd[crowd_id] + samples_per_crowd_offsets[crowd_id];
+      const int base_sample_index = inb * walkers_per_crowd[crowd_id] + samples_per_crowd_offsets[crowd_id];
 
       auto p_list_no_leader  = opt_data.get_p_list(current_batch_size);
       auto wf_list_no_leader = opt_data.get_wf_list(current_batch_size);
@@ -597,7 +597,7 @@ QMCCostFunctionBatched::Return_rt QMCCostFunctionBatched::correlatedSampling(boo
 
       for (int ib = 0; ib < current_batch_size; ib++)
       {
-        int is = base_sample_index + ib;
+        const int is = base_sample_index + ib;
         wf_list[ib].G += *gradPsi[is];
         wf_list[ib].L += *lapPsi[is];
         // This is needed to get the KE correct in QMCHamiltonian::mw_evaluate below
@@ -627,7 +627,7 @@ QMCCostFunctionBatched::Return_rt QMCCostFunctionBatched::correlatedSampling(boo
 
         for (int ib = 0; ib < current_batch_size; ib++)
         {
-          int is                        = base_sample_index + ib;
+          const int is                  = base_sample_index + ib;
           auto etmp                     = energy_list[ib];
           RecordsOnNode[is][ENERGY_NEW] = etmp + RecordsOnNode[is][ENERGY_FIXED];
           for (int j = 0; j < nparam; j++)
@@ -646,7 +646,7 @@ QMCCostFunctionBatched::Return_rt QMCCostFunctionBatched::correlatedSampling(boo
         auto energy_list = QMCHamiltonian::mw_evaluate(h0_list, wf_list, p_list);
         for (int ib = 0; ib < current_batch_size; ib++)
         {
-          int is                        = base_sample_index + ib;
+          const int is                  = base_sample_index + ib;
           auto etmp                     = energy_list[ib];
           RecordsOnNode[is][ENERGY_NEW] = etmp + RecordsOnNode[is][ENERGY_FIXED];
         }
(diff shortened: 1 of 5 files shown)


src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
```diff
@@ -750,18 +746,17 @@ bool QMCFixedSampleLinearOptimizeBatched::processOptXML(xmlNodePtr opt_xml,
 
   vmcEngine->enable_sample_collection();
 
-  // Code to check and set crowds take from QMCDriverNew::adjustGlobalWalkerCount
-  checkNumCrowdsLTNumThreads(opt_num_crowds_);
-  if (opt_num_crowds_ == 0)
-    opt_num_crowds_ = Concurrency::maxCapacity<>();
 
-  app_log() << " Number of crowds for optimizer: " << opt_num_crowds_ << std::endl;
+  QMCDriverNew::AdjustedWalkerCounts awc =
```
ye-luo, line 750: If you prefer not to see the warning again. We can add an option to suppress warnings in adjustGlobalWalkerCount

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -272,38 +270,40 @@ void QMCCostFunctionBatched::checkConfigurations()
 
   // Create crowd-local storage for evaluation
   outputManager.pause();
-  opt_eval_.resize(opt_num_crowds_);
-  for (int i = 0; i < opt_num_crowds_; i++)
-    opt_eval_[i] = std::make_unique<CostFunctionCrowdData>(opt_batch_size_, W, Psi, H, H_KE_node_names_, *MoverRng[0]);
+  size_t opt_num_crowds = walkers_per_crowd_.size();
```
ye-luo, line 273: If this is a convenient variable, let us add a const.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -272,38 +270,40 @@ void QMCCostFunctionBatched::checkConfigurations()
...
     int num_batches;
     int final_batch_size;
 
-    compute_batch_parameters(local_samples, crowd_size, num_batches, final_batch_size);
+    compute_batch_parameters(local_samples, walkers_per_crowd[crowd_id], num_batches, final_batch_size);
 
     for (int inb = 0; inb < num_batches; inb++)
     {
-      int curr_crowd_size = crowd_size;
+      int curr_crowd_size = walkers_per_crowd[crowd_id];
```
ye-luo, line 302: If this is a convenient variable, let us add a const.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -511,13 +511,14 @@ QMCCostFunctionBatched::Return_rt QMCCostFunctionBatched::correlatedSampling(boo
   bool compute_nlpp             = useNLPPDeriv && (includeNonlocalH != "no");
   bool compute_all_from_scratch = (includeNonlocalH != "no"); //true if we have nlpp
 
+  size_t opt_num_crowds = walkers_per_crowd_.size();
   // Divide samples among crowds
-  std::vector<int> samples_per_crowd(opt_num_crowds_ + 1);
-  FairDivide(rank_local_num_samples_, opt_num_crowds_, samples_per_crowd);
+  std::vector<int> samples_per_crowd(opt_num_crowds + 1);
+  FairDivide(rank_local_num_samples_, opt_num_crowds, samples_per_crowd);
```
ye-luo, line 517: This may cause larger than expected imbalance when crowd sizes are not equal. Need to make sure per-walker number of samples differ by one at maximum instead of the number of per-crowd samples differ by 1.

src/QMCDrivers/tests/test_QMCCostFunctionBatched.cpp
```diff
@@ -140,14 +141,14 @@ TEST_CASE("fillOverlapAndHamiltonianMatrices", "[drivers]")
 // the input/gold data (from a file created by convert_hdf_to_cpp.py)
 void fill_from_text(int num_opt_crowds, FillData& fd)
 {
-  // Not used in the function under test
-  int crowd_size = 1;
+  using Index = QMCTraits::IndexType;
```
ye-luo, line 144: Does putting int directly work? I never find Index alias useful and  we probably should think about removing it.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
markdewing: This variable can't be const because it may get changed in the following lines.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
markdewing: I added a TODO comment to the same code in checkConfigurations.  Could this change be moved to a future PR?

ye-luo: Yes. But please open an issue to track this.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -272,38 +270,43 @@ void QMCCostFunctionBatched::checkConfigurations()
...
   outputManager.resume();
 
 
+  // TODO - walkers per crowd may not be evenly divided, so the samples per crowd
+  //        might need to be divided differently for better load balancing.
+
   // Divide samples among the crowds
-  std::vector<int> samples_per_crowd(opt_num_crowds_ + 1);
-  FairDivide(rank_local_num_samples_, opt_num_crowds_, samples_per_crowd);
+  std::vector<int> samples_per_crowd(opt_num_crowds + 1);
```
ye-luo, line 285: rename this vector. It is not samples_per_crowd but offsets.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -272,38 +270,43 @@ void QMCCostFunctionBatched::checkConfigurations()
...
-                          opt_variables_type& optVars, bool needGrads, bool compute_nlpp,
-                          const std::string& includeNonlocalH) {
+                          const std::vector<int>& samples_per_crowd, const std::vector<int>& walkers_per_crowd,
+                          std::vector<ParticleGradient*>& gradPsi, std::vector<ParticleLaplacian*>& lapPsi,
+                          Matrix<Return_rt>& RecordsOnNode, Matrix<Return_rt>& DerivRecords,
+                          Matrix<Return_rt>& HDerivRecords, const SampleStack& samples, opt_variables_type& optVars,
+                          bool needGrads, bool compute_nlpp, const std::string& includeNonlocalH) {
     CostFunctionCrowdData& opt_data = *opt_crowds[crowd_id];
 
     int local_samples = samples_per_crowd[crowd_id + 1] - samples_per_crowd[crowd_id];
```
ye-luo, line 297: Put this one as const.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
ye-luo: I see. Then leave it not const. Could you rename this one as `this_batch_size`?

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
markdewing: This isn't part of the code that was changed in this PR!  It starts to feel excessively picky when minor changes get requested that aren't part of the changed code.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
markdewing: Will rename to `samples_per_crowd_offsets`

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
markdewing: How about `current_batch_size`?  I don't like using `this` in variable names because of the special meaning of "this" in C++.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
ye-luo: I should have said "could you please put this one as const?"
Existing code are subject to post-merge review as well.
When I read though the existing code trying to understand how `samples_per_crowd` is defined/used, I read this line and would like to have const added. That is also why I requested to rename the variable to reduce confusion.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
ye-luo: This is an example of surprise. Let me explain.
`curr_crowd_size` is somehow a const once its value is settled. So it is possible to made a decision and assign its value once and mark it const. It is better than set an initial value and then modify it later. Although the current way looks OK, the modification is on the second line but we do lost the const protection.

ye-luo: `current_batch_size` is good

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
ye-luo: You expressed concerns of printing duplicated warning. How would you like to handle this?
I'm actually fine with printing warning twice. First, the occurrence is low. If users decide to live with the warning, they can live with twice warning. Second, if we can reuse walker objects, then warning is also printed once only.
So I don't really feel this is critical. let me know your choice to add a warning suppression flag or leave it.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
ye-luo: Could you open an github issue? I think this the last thing blocking this PR (from my perspective).

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
markdewing: Printing the warning twice is fine for now.  If it becomes an issue later, we can add the flag.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
markdewing: The ternary operator would be one solution.
```
const int current_batch_size = inb == num_batches - 1 ? final_batch_size : walkers_per_crowd[crowd_id
```

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
markdewing: Opened #3888

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
markdewing: extra consts added in several places.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
ye-luo: That is a clean solution that we can go for.

ye-luo:
Test this please

ye-luo (commented):
LGTM. Still need non-ANL approval.

prckent (approved):
Thanks Mark & Ye.

PR #3874, opened 2022-02-24
Reorder unique determinants by excitation level

kgasperich:
## Proposed changes

This reorders the unique determinants of each particle type by excitation level.
It also identifies the reference SlaterDeterminant as the one with the largest coefficient (in magnitude) (was previously always set to the first det in the list).
For each particle group, the reference DiracDeterminant is the one that appears in the reference SlaterDeterminant.
There is also a new data member of MultiDiracDeterminant that holds the number of unique dets at each excitation level

- New feature

- No

unix workstation (gcc 10.3.0; amd zen)

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
+++ b/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
@@ -592,8 +592,8 @@ private:
   std::shared_ptr<std::vector<int>> detData;
   std::shared_ptr<std::vector<std::pair<int, int>>> uniquePairs;
   std::shared_ptr<std::vector<RealType>> DetSigns;
-  /* number of unique determinants at each excitation level (relative to reference)
-   * {1, n_singles, n_doubles, n_triples, ...}
+  /** number of unique determinants at each excitation level (relative to reference)
+   *  {1, n_singles, n_doubles, n_triples, ...}
    */
   std::shared_ptr<std::vector<int>> ndet_per_exc_lvl;
   MultiDiracDeterminantCalculator<ValueType> DetCalculator;
--- a/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
+++ b/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
@@ -80,6 +80,7 @@ public:
    *@param first index of first particle
    *@param nel number of particles in the determinant
    *@param ref_det_id id of the reference determinant
+   *@param C2nodes_ptcl mapping from overall det index to unique det index for this particle group
    *
    * Note: ciConfigList should have been populated when calling this function
    */
@@ -222,8 +223,17 @@ public:
    * END END END
    ***************************************************************************/
 
-  // create necessary structures used in the evaluation of the determinants
-  // this works with confgList, which shouldn't change during a simulation
+  /** create necessary structures used in the evaluation of the determinants
+   * sort confgList by excitation level
+   * confgList shouldn't change during a simulation after it is sorted here
+   *
+   *@param ref reference configuration
+   *@param data data structure holding information about excitation holes/particles (*this->detData)
+   *@param pairs all hole/particle pairs needed to construct excitation matrices for table method (*this->uniquePairs)
+   *@param sign sign of each excitation (parity of permutation of orbital indices) (*this->DetSigns)
+   *@param C2nodes_ptcl mapping from overall det index to unique det index for this particle group
+   *
+   */
   void createDetData(const ci_configuration2& ref,
                      std::vector<int>& data,
                      std::vector<std::pair<int, int>>& pairs,
@@ -582,6 +592,9 @@ private:
   std::shared_ptr<std::vector<int>> detData;
   std::shared_ptr<std::vector<std::pair<int, int>>> uniquePairs;
   std::shared_ptr<std::vector<RealType>> DetSigns;
+  /* number of unique determinants at each excitation level (relative to reference)
+   * {1, n_singles, n_doubles, n_triples, ...}
+   */
   std::shared_ptr<std::vector<int>> ndet_per_exc_lvl;
   MultiDiracDeterminantCalculator<ValueType> DetCalculator;
 };
--- a/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
+++ b/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
@@ -595,7 +595,7 @@ private:
   /** number of unique determinants at each excitation level (relative to reference)
    *  {1, n_singles, n_doubles, n_triples, ...}
    */
-  std::shared_ptr<std::vector<int>> ndet_per_exc_lvl;
+  std::shared_ptr<std::vector<int>> ndets_per_excitation_level;
   MultiDiracDeterminantCalculator<ValueType> DetCalculator;
 };
 
--- a/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
+++ b/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
@@ -83,7 +83,7 @@ public:
    *
    * Note: ciConfigList should have been populated when calling this function
    */
-  void set(int first, int nel, int ref_det_id);
+  void set(int first, int nel, int ref_det_id, std::vector<size_t>& C2nodes_ptcl);
 
   ///optimizations  are disabled
   inline void checkInVariables(opt_variables_type& active) override { Phi->checkInVariables(active); }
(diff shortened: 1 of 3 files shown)


prckent (commented):
What is the order of algorithm employed here? Was more use of standard c++ library functionality considered ? I was expecting to see a sort, which would benefit from future library improvements, automagic parallelization etc. Now that the code is written it is less critical but we should avail ourselves of standard functionality where practical.

I have not checked coverage yet, but we need it to be very high in the critical parts of the algorithm.

kgasperich:
> What is the order of algorithm employed here?

All of it is executed once when the MultiSlaterDetTableMethod is constructed.

The additional code has [one part that scales as O(n_unique_dets)](https://github.com/kgasperich/qmcpack/blob/478d28c51e9a902785302f68df09c6e8a0856dbc/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.cpp#L97-L114)  and [one that scales as O(n_slater_dets*n_particle_groups) ](https://github.com/kgasperich/qmcpack/blob/478d28c51e9a902785302f68df09c6e8a0856dbc/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.cpp#L118-L119) (we could combine this last part into a single loop for the entire MultiSlaterDetTableMethod if we're ok with a bit more passing around of data).

>Was more use of standard c++ library functionality considered ? I was expecting to see a sort, which would benefit from future library improvements, automagic parallelization etc. Now that the code is written it is less critical but we should avail ourselves of standard functionality where practical.

We already do the bulk of the sorting work (or at least get the information required for the sort) [when we get the excitation level, holes, particles, and parity](https://github.com/kgasperich/qmcpack/blob/478d28c51e9a902785302f68df09c6e8a0856dbc/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.cpp#L66) for each unique determinant.  In the old code, these were all pushed onto `MultiDiracDeterminant::detData` in the order in which they were encountered in the list of unique dets (which is the order in which they first appeared in the list of all Slater dets).  The new code pushes these onto separate vectors (one for each excitation level, stored in a map), and after that we just concatenate these vectors in order of excitation level (I used `std::insert`, but we could initialize with the correct size and use `std::fill` instead).

If we want to use some standard sort implementation, it might be worth adding a determinant representation based on something like [`boost::dynamic_bitset`](https://www.boost.org/doc/libs/1_78_0/libs/dynamic_bitset/dynamic_bitset.html) (or a vector of  `std::bitset<64>` a la QP). This would give us a comparison operation (for sorting) that only needs an xor and a popcount (rather than the [current method](https://github.com/QMCPACK/qmcpack/blob/541149b085d66f090a3a4875b3d36827d0118f85/src/QMCWaveFunctions/Fermion/ci_configuration2.h#L53-L63) of searching for matching elements of two vectors)

> I have not checked coverage yet, but we need it to be very high in the critical parts of the algorithm.

Anything that evaluates the value (or gradient or laplacian) of a `MultiSlaterDetTableMethod` should fail in a noticeable way if this is done incorrectly.
We don't yet have anything downstream that assumes this ordering is done, but the `test_wavefunction_determinant` is passing, so even if the ordering is somehow incorrect, at least it appears to be consistent with everything we currently assume about the determinant data (and as we modify the downstream bits to rely on the correct ordering, we will see failures in any existing tests that cover the current implementations of those functions).

anbenali:
Would it be possible to run some typeof timing to see if this affects time to solution even in the non batch form and on CPU only... the point is to see if the reordering will either cost or if the better memory alignment helps (?). If this is not relevant then discard.

ye-luo:
> Would it be possible to run some typeof timing to see if this affects time to solution even in the non batch form and on CPU only... the point is to see if the reordering will either cost or if the better memory alignment helps (?). If this is not relevant then discard.

likely not having any impact until more refactoring happens on the compute side.

prckent (approved):

prckent:
Test this please

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
```diff
@@ -581,6 +582,7 @@ class MultiDiracDeterminant : public WaveFunctionComponent
   std::shared_ptr<std::vector<int>> detData;
   std::shared_ptr<std::vector<std::pair<int, int>>> uniquePairs;
   std::shared_ptr<std::vector<RealType>> DetSigns;
+  std::shared_ptr<std::vector<int>> ndet_per_exc_lvl;
```
ye-luo, line 585: Document this variable.

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
```diff
@@ -83,7 +83,7 @@ class MultiDiracDeterminant : public WaveFunctionComponent
    *
    * Note: ciConfigList should have been populated when calling this function
    */
-  void set(int first, int nel, int ref_det_id);
+  void set(int first, int nel, int ref_det_id, std::vector<size_t>& C2nodes_ptcl);
```
ye-luo, line 86: Update documentation.

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
```diff
@@ -227,7 +227,8 @@ class MultiDiracDeterminant : public WaveFunctionComponent
   void createDetData(const ci_configuration2& ref,
                      std::vector<int>& data,
                      std::vector<std::pair<int, int>>& pairs,
-                     std::vector<RealType>& sign);
+                     std::vector<RealType>& sign,
+                     std::vector<size_t>& C2nodes_ptcl);
```
ye-luo, line 231: Please document this function with all its arguments.

ye-luo (commented):
Once the documentations are added, this PR can be merged.
Overall, I noticed that how class member objects are passed around are convoluted. I will do cleanup in separate PRs.

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
kgasperich: I added a description (maybe it would also be helpful to add member function `int max_exc_lvl(){return ndet_per_exc_lvl.size()+1;}`?

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
kgasperich: added description of `C2nodes_ptcl` to documentation

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
```diff
@@ -582,6 +592,9 @@ class MultiDiracDeterminant : public WaveFunctionComponent
   std::shared_ptr<std::vector<int>> detData;
   std::shared_ptr<std::vector<std::pair<int, int>>> uniquePairs;
   std::shared_ptr<std::vector<RealType>> DetSigns;
+  /* number of unique determinants at each excitation level (relative to reference)
+   * {1, n_singles, n_doubles, n_triples, ...}
```
ye-luo, line 596: double asteroids on the first commend line.

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
kgasperich: I added some documentation (I changed as little as possible from the previous implementation, but I don't know why we pass around so many objects that are already members of the `MultiDiracDeterminant`)

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
ye-luo: This is nice to have but kind of doubt it is absolutely needed. I expect use `ndet_per_exc_lvl` more directly.
Please also rename `exc_lvl`, use full words.

ye-luo: `ndets_per_excited_level`

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
kgasperich: fixed

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
kgasperich: fixed

ye-luo (approved):
LGTM

ye-luo:
Test this please

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.h
ye-luo: Current APIs and data encapsulations are messy and I would like to tide them up later.
Sometimes we do pass class member via argument, in such way, the function may be turned const or event static to make the function saner and independently testable.

PR #3881, opened 2022-02-27
WaveFunctionPool owns a map of TrialWaveFunction

ye-luo:
## Proposed changes
Previously. `WaveFunctionPool` holds a map of `WaveFunctionFactory*`, I think the reason to keep WaveFunctionFactory around was to keep SPOSetBuilderFactory around to access SPOSets. This is over complicated.
This PR changes WaveFunctionPool to own a map of `TrialWaveFunction*` and `WaveFunctionFactory` is only short-lived to build `TrialWaveFunction*`. Once the constructions finishes, All the SPOs will be transferred to TrialWaveFunction before WaveFunctionFactory got destroyed. In the current code, SPOSet are TWF specific and cannot be shared across two TWFs. For this reason, makes sense for TWF to own the map of pre-built SPOSets.

- No

epyc-server

## Checklist

### DIFF
--- a/src/Estimators/EstimatorManagerNew.cpp
+++ b/src/Estimators/EstimatorManagerNew.cpp
@@ -401,7 +401,7 @@ bool EstimatorManagerNew::put(QMCHamiltonian& H,
         // happens once insures golden particle set is not abused.
         ParticleSet pset_target(pset);
         operator_ests_.emplace_back(std::make_unique<OneBodyDensityMatrices>(std::move(obdmi), pset.getLattice(),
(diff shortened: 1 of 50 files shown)


ye-luo:
Test this please

prckent (approved):
With this change are there any remaining persistent factories that need similar removing/redesigning?

ye-luo:
Unforunately, `HamiltonianPool` still holds `std::map<std::string, HamiltonianFactory*>` so persistent.

PR #3887, opened 2022-03-01
Add CoordsType templates to WaveFunctionComponents

camelto2:

## Proposed changes
This PR moves the CoordsType abstraction down into WFC. Currently, TrialWaveFunction had mw_evalGrad,  mw_evalGradWithSpin and a mw_evalGrad<CT> to decide which version to call. Same for mw_calcRatioGrad. 

This simplifies the TrialWaveFunction API to only have the one implementation and moves the selection down into WaveFunctionComponent. WFC now has mw_evalGrad, mw_evalGradWithSpin, and a mw_evalGrad<CT> (same for mw_calcRatioGrad) to choose which one to call based on CT

Eventually, we will want to only have mw_evalGrad<CT> in WaveFunctionComponent, but trying to keep things small and manageable for easier review.

- New feature

- No

macOSX

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/TWFGrads.hpp
+++ b/src/QMCWaveFunctions/TWFGrads.hpp
@@ -24,7 +24,7 @@ struct TWFGrads;
 template<>
 struct TWFGrads<CoordsType::POS>
 {
-  TWFGrads(const std::size_t size) : grads_positions(size) {};
+  TWFGrads(const std::size_t size) : grads_positions(size){};
 
   TWFGrads& operator+=(const TWFGrads& rhs);
 
@@ -34,7 +34,7 @@ struct TWFGrads<CoordsType::POS>
 template<>
 struct TWFGrads<CoordsType::POS_SPIN>
 {
-  TWFGrads(const std::size_t size) : grads_positions(size), grads_spins(size) {};
+  TWFGrads(const std::size_t size) : grads_positions(size), grads_spins(size){};
 
   TWFGrads& operator+=(const TWFGrads& rhs);
 
--- a/src/QMCWaveFunctions/TWFGrads.hpp
+++ b/src/QMCWaveFunctions/TWFGrads.hpp
@@ -24,7 +24,7 @@ struct TWFGrads;
 template<>
 struct TWFGrads<CoordsType::POS>
 {
-  TWFGrads(const std::size_t size) : grads_positions(size){};
+  TWFGrads(const std::size_t size); 
 
   TWFGrads& operator+=(const TWFGrads& rhs);
 
@@ -34,7 +34,7 @@ struct TWFGrads<CoordsType::POS>
 template<>
 struct TWFGrads<CoordsType::POS_SPIN>
 {
-  TWFGrads(const std::size_t size) : grads_positions(size), grads_spins(size){};
+  TWFGrads(const std::size_t size);
 
   TWFGrads& operator+=(const TWFGrads& rhs);
 
--- a/src/QMCWaveFunctions/TWFGrads.hpp
+++ b/src/QMCWaveFunctions/TWFGrads.hpp
@@ -26,6 +26,8 @@ struct TWFGrads<CoordsType::POS>
 {
   TWFGrads(const std::size_t size) : grads_positions(size) {};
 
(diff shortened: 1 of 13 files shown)


src/QMCWaveFunctions/WaveFunctionComponent.h
```diff
@@ -347,10 +362,7 @@ class WaveFunctionComponent : public QMCTraits
                                     int iat,
                                     std::vector<PsiValueType>& ratios,
                                     std::vector<GradType>& grad_new,
-                                    std::vector<ComplexType>& spingrad_new) const
-  {
-    mw_ratioGrad(wfc_list, p_list, iat, ratios, grad_new);
-  }
+                                    std::vector<ComplexType>& spingrad_new) const;
```
ye-luo, line 365: Is this function still been called from outside? otherwise, let us move it under private.

src/QMCWaveFunctions/TWFGrads.hpp
```diff
@@ -24,15 +24,19 @@ struct TWFGrads;
 template<>
 struct TWFGrads<CoordsType::POS>
 {
-  void resize(const std::size_t size);
+  TWFGrads(const std::size_t size) : grads_positions(size){};
```
ye-luo, line 27: I saw a few places with function definition moved to cpp files. Thumb up.
Let us leave the constructor implementation in the cpp as well.

src/QMCWaveFunctions/WaveFunctionComponent.cpp
```diff
@@ -119,6 +156,19 @@ void WaveFunctionComponent::mw_ratioGrad(const RefVectorWithLeader<WaveFunctionC
     ratios[iw] = wfc_list[iw].ratioGrad(p_list[iw], iat, grad_new[iw]);
 }
 
+void WaveFunctionComponent::mw_ratioGradWithSpin(const RefVectorWithLeader<WaveFunctionComponent>& wfc_list,
+                                                 const RefVectorWithLeader<ParticleSet>& p_list,
+                                                 int iat,
+                                                 std::vector<PsiValueType>& ratios,
+                                                 std::vector<GradType>& grad_new,
+                                                 std::vector<ComplexType>& spingrad_new) const
+{
```
ye-luo, line 165: It seems that the default implementation changed but this is more safe to use.

ye-luo (commented):
thank you for the further refactoring and pushing the abstraction more extensively.

src/QMCWaveFunctions/TWFGrads.hpp
camelto2: moved constructor to cpp

src/QMCWaveFunctions/WaveFunctionComponent.cpp
camelto2: yea, I thought this was a better default

src/QMCWaveFunctions/WaveFunctionComponent.h
camelto2: right now none of the mw_ratioGradWithSpin or mw_evalGradWithSpin has actually been implemented for DiracDet or MultiDiracDet. That is for a future PR. 

I think if this goes to private, it would make sense to move the "non-spin" version as private as well and only keep the templated API as public.

ye-luo: Yes. If no existing calls from outside, move both to private and only expose the template one.

src/QMCWaveFunctions/TWFGrads.hpp
```diff
@@ -24,7 +24,7 @@ struct TWFGrads;
 template<>
 struct TWFGrads<CoordsType::POS>
 {
-  TWFGrads(const std::size_t size) : grads_positions(size){};
+  TWFGrads(const std::size_t size);
```
ye-luo, line 27: A trailing whitespace. Take care of it in the next commit?

src/QMCWaveFunctions/WaveFunctionComponent.h
camelto2: the WithSpin APIs are now private.

ye-luo (approved):
LGTM

PR #3893, opened 2022-03-04
Introduce PlatformSelector to allow both direct and indrect selection and enable offload codepath in non-offload builds.

### LINKED ISSUE #2451: New input tag to switch implementations
As mentioned in https://github.com/QMCPACK/qmcpack/pull/2423 we need a new an extensible input tag to switch execution between (say) cpu omp, gpuomp gpu omp+cuda ... . Hopefully not many will be needed, either for development or long term production. We will avoid compile time switches. Since this will be user visible we should discuss options before implementing.

ye-luo:
## Proposed changes
As described in #2451, we need a way to select the exact implementation and being able map yes/no based on the build. So introduce PlatformSelector<CPU_OMPTARGET>.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Platforms/PlatformSelector.hpp
+++ b/src/Platforms/PlatformSelector.hpp
@@ -44,5 +44,15 @@ public:
 };
 
 using CPUOMPTargetSelector = PlatformSelector<SelectorKind::CPU_OMPTARGET>;
+
+template<>
+class PlatformSelector<SelectorKind::CPU_OMPTARGET_CUDA>
+{
+public:
+  static const std::vector<std::string> candidate_values;
+  static PlatformKind selectPlatform(std::string_view value);
+};
+
+using CPUOMPTargetCUDASelector = PlatformSelector<SelectorKind::CPU_OMPTARGET_CUDA>;
 } // namespace qmcplusplus
 #endif
--- a/src/Platforms/PlatformSelector.hpp
+++ b/src/Platforms/PlatformSelector.hpp
@@ -19,28 +19,30 @@
 namespace qmcplusplus
 {
 
-enum class PlatformKind {
+enum class PlatformKind
+{
   CPU,
   OMPTARGET,
   CUDA
 };
 
-enum class SelectorKind {
-  OMPTARGET,
-  OMPTARGET_AND_CUDA
+enum class SelectorKind
+{
+  CPU_OMPTARGET,
+  CPU_OMPTARGET_CUDA
 };
 
 template<SelectorKind KIND>
 class PlatformSelector;
 
 template<>
-class PlatformSelector<SelectorKind::OMPTARGET>
+class PlatformSelector<SelectorKind::CPU_OMPTARGET>
 {
 public:
   static const std::vector<std::string> candidate_values;
-  static PlatformKind convertToPlatform(std::string_view value);
+  static PlatformKind selectPlatform(std::string_view value);
 };
 
-using OMPTargetSelector = PlatformSelector<SelectorKind::OMPTARGET>;
-}
+using CPUOMPTargetSelector = PlatformSelector<SelectorKind::CPU_OMPTARGET>;
+} // namespace qmcplusplus
 #endif
--- a/src/Platforms/PlatformSelector.hpp
+++ b/src/Platforms/PlatformSelector.hpp
@@ -1,0 +1,46 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2022 QMCPACK developers.
+//
+// File developed by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//
+// File created by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+#ifndef QMCPLUSPLUS_PLATFORM_SELECTOR_H
+#define QMCPLUSPLUS_PLATFORM_SELECTOR_H
+
+#include <vector>
+#include <string>
+
+namespace qmcplusplus
+{
+
+enum class PlatformKind {
+  CPU,
+  OMPTARGET,
+  CUDA
+};
+
+enum class SelectorKind {
+  OMPTARGET,
+  OMPTARGET_AND_CUDA
+};
+
+template<SelectorKind KIND>
+class PlatformSelector;
+
+template<>
+class PlatformSelector<SelectorKind::OMPTARGET>
+{
+public:
+  static const std::vector<std::string> candidate_values;
+  static PlatformKind convertToPlatform(std::string_view value);
+};
+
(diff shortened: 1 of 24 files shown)


ye-luo:
Test this please

src/Platforms/PlatformSelector.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+enum class PlatformKind {
+  CPU,
+  OMPTARGET,
+  CUDA
+};
+
+enum class SelectorKind {
+  OMPTARGET,
+  OMPTARGET_AND_CUDA
```
ye-luo, line 30: I probably should change these to "CPU_OMPTARGET" and "CPU_OMPTARGET_CUDA". I may also need "CPU_CUDA" and "OMPTARGET_CUDA".

src/QMCWaveFunctions/tests/test_TrialWaveFunction_diamondC_2x1x1.cpp
```diff
@@ -44,7 +44,7 @@ struct float_tag
  *  used and a constexpr based on DeterminantTypes that actually depend on legacy cuda.
  */
 template<class DiracDet, class SPO_precision>
-void testTrialWaveFunction_diamondC_2x1x1(const int ndelay)
+void testTrialWaveFunction_diamondC_2x1x1(const int ndelay, const bool offload_spo, const bool offload_jas)
```
PDoakORNL, line 47: a struct for input would make this way more readable later. 
```
struct offload_comp {
bool spo;
bool jas
};

File: src/QMCWaveFunctions/tests/test_TrialWaveFunction_diamondC_2x1x1.cpp [hunk 3]
```diff
@@ -574,25 +582,92 @@ TEST_CASE("TrialWaveFunction_diamondC_2x1x1", "[wavefunction]")
 #if defined(ENABLE_CUDA) && defined(ENABLE_OFFLOAD)
   SECTION("DiracDeterminantBatched<MatrixDelayedUpdateCUDA>")
   {
-    testTrialWaveFunction_diamondC_2x1x1<DiracDeterminantBatched<MatrixDelayedUpdateCUDA<VT, FPVT>>, float_tag>(1);
-    testTrialWaveFunction_diamondC_2x1x1<DiracDeterminantBatched<MatrixDelayedUpdateCUDA<VT, FPVT>>, float_tag>(2);
-    testTrialWaveFunction_diamondC_2x1x1<DiracDeterminantBatched<MatrixDelayedUpdateCUDA<VT, FPVT>>, double_tag>(1);
-    testTrialWaveFunction_diamondC_2x1x1<DiracDeterminantBatched<MatrixDelayedUpdateCUDA<VT, FPVT>>, double_tag>(2);
+    testTrialWaveFunction_diamondC_2x1x1<DiracDeterminantBatched<MatrixDelayedUpdateCUDA<VT, FPVT>>, float_tag>(1, false, false);
```
PDoakORNL, line 585:
```
OffloadComp off_select;
offload.spo = false;
offload.jas = false;
```
Then use this for the calls below.

src/QMCWaveFunctions/tests/test_TrialWaveFunction_diamondC_2x1x1.cpp
```diff
@@ -574,25 +582,92 @@ TEST_CASE("TrialWaveFunction_diamondC_2x1x1", "[wavefunction]")
...
+    testTrialWaveFunction_diamondC_2x1x1<DiracDeterminantBatched<MatrixDelayedUpdateCUDA<VT, FPVT>>, float_tag>(2, false, false);
+    testTrialWaveFunction_diamondC_2x1x1<DiracDeterminantBatched<MatrixDelayedUpdateCUDA<VT, FPVT>>, double_tag>(1, false, false);
+    testTrialWaveFunction_diamondC_2x1x1<DiracDeterminantBatched<MatrixDelayedUpdateCUDA<VT, FPVT>>, double_tag>(2, false, false);
+  }
+  SECTION("DiracDeterminantBatched<MatrixDelayedUpdateCUDA>_offload_spo")
+  {
+    testTrialWaveFunction_diamondC_2x1x1<DiracDeterminantBatched<MatrixDelayedUpdateCUDA<VT, FPVT>>, float_tag>(1, true, false);
```
PDoakORNL, line 592:
```
offload.spo = true;
offload.jas = false;
```

src/QMCWaveFunctions/BsplineFactory/createComplexSingle.cpp
```diff
@@ -70,8 +67,7 @@ std::unique_ptr<BsplineReaderBase> createBsplineComplexSingle(EinsplineSetBuilde
   }
 #else //QMC_COMPLEX
   app_summary() << "    Using real valued spline SPOs with complex single precision storage (C2R)." << std::endl;
-#if defined(ENABLE_OFFLOAD)
-  if (useGPU == "yes")
+  if (OMPTargetSelector::convertToPlatform(useGPU) == PlatformKind::OMPTARGET)
```
PDoakORNL, line 70: As mentioned above the name seems wrong.

PDoakORNL (commented):
Overall this seems like a good idea and a better option than putting the logic in cmake and trying to enact it through defines.
 
I don't think the string parsing belongs here long term. Eventually post input parsing the input class  should deliver a boolean i.e. `input.get_use_gpu()` or enum class value ComputeImpl::CPU, ComputeImpl::OMPTARGET, ComputeImpl::CUDA, etc. if we are going to accept input like `<some_that_computes gpu=no|yes|cpu|omptarget|cuda>`

Still in that case we're going to have the need to represent some sort of selection matrix  for implementations. This seems like fairly clean way to do it.  I think the scope is fine at this time.  Gradually this change can be pushed through the relevant factory functions/classes.

src/QMCWaveFunctions/tests/test_TrialWaveFunction_diamondC_2x1x1.cpp
ye-luo: Changed.

src/QMCWaveFunctions/BsplineFactory/createComplexSingle.cpp
ye-luo: Fixed accordingly

ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
Regarding input parsing, the string to enum needs to be done somewhere. In certain cases, put all the transition logic in the input class may not be feasible. For example, we may need to check if the input value is valid for previously built ParticleSet objects. So I think we should allow keep the original input value and delay its parsing to enum when appropriate.

ye-luo:
@PDoakORNL Does this look OK now?

ye-luo:
I didn't mean arbitrary strings. I mean the value captured by the input class is never final. The actual final value may be modified when being passed from the input class to the final use.

Issue #3875, opened 2022-02-24
Prevent Use of Legacy Estimators with Batched Code


PDoakORNL:
Abort if legacy estimator is specified when the batched drivers are used.

ye-luo:
@jtkrogel We do have MPC working in the batched code.
https://cdash.qmcpack.org/CDash/testDetails.php?test=17171347&build=293503
It seems being added to auxH. It is likely being labelled as hamiltonian non phsyical rather than an estimator.

PR #3897, opened 2022-03-04
Global batched driver switch

PDoakORNL:
**Note: After discussion and review, the switch was decided to be named driver_version  (not epoch)**

This removes the 
```xml
<qmc method=vmc_batch|dmc_batch|linear_batch>
```
and replaces them with 
```xml
    <project id="qmc_short" series="0">
     <parameter name="driver_epoch">batched</parameter>
   </project>
...
<qmc method=vmc|dmc|linear>
```

## Proposed changes
ProjectData can now be queried get_driver_epoch() and explicitly different input parsing can easily be done for batched drivers vs. legacy.

The unit tests pass.  
Most of the short and deterministic batched tests pass.  I'm not sure any of the system level batched optimizer test pass.  They seem extremely slow.

In the future this will greatly simplify preventing legacy/new input mismatches. i.e. #3875

- New feature
- Documentation changes

- Yes
For legacy no, for batched yes.
Your existing batched driver input will stop working.
You must make the change shown at the top of the PR to have them continue working.

Leconte

## Checklist


ye-luo:
I still would like to run `xxx_batch`. I thought the intended behavior was when the global switch is on. "XXX" is redirected to "XXX_batch".

PDoakORNL:
I thought the desire was to drop the vmc_batch name.  @jtkrogel  convinced me that not having the second driver name was better and it definitely is considering the legacy handling of input.  ProjectData is high level and parsed very early, we need to parse qmchamiltonian differently based on whether the driver is batched or not and this greatly simplifies that.

prckent:
xxx_batch will be history. This is the intended change. Much friendlier for users and we want to block off the old drivers in any case.

I think we should try to improve on "driver_epoch". There is a lot more than the driver being switched here, plus this tag will be visible for "eternity" given the transition plan. Ask on Slack for ideas.

jtkrogel:
Any chance we change `<parameter name="driver_epoch">batched</parameter>` to `<parameter name="driver">batched</parameter>`?  It is about as simple as is needed.

Nexus currently uses `driver='batched'` and `driver='legacy'` to switch between these.  Whatever is decided, I will update Nexus to match.

prckent:
> Nexus currently uses `driver='batched'` and `driver='legacy'` to switch between these. Whatever is decided, I will update Nexus to match.

Any arguments to not use these? Consistency is a strong argument for.

src/QMCDrivers/tests/ValidQMCInputSections.h
```diff
@@ -103,13 +103,14 @@ constexpr std::array<const char*, 2> valid_dmc_input_sections{
 constexpr int valid_dmc_input_dmc_index       = 0;
 constexpr int valid_dmc_input_dmc_batch_index = 1;
 
+/** As far as I can tell these are no longer valid */
```
markdewing, line 106: I need to update the WFOptDriverInput class, and I'll change these tests then as well.

ye-luo:
Technically, determine the behavior fully relying on a global state is convenient but bad both for the code parsing and human reading.

As a user, I request the existing input with "vmc_batch" to continue functioning. It is bad to upset users making breaking changes. The cost of keep "vmc_batch" tag to work is nothing and there is no ambiguity.

Here is what consider a sane way.
When "vmc_legacy" or "vmc_batch" is inputted as a method, it should hit exactly the requested driver.
Only when the input is "vmc", driver factory maps it to "vmc_legacy" or "vmc_batch" based on the global state.

For users new to this complexity, they may choose to just rely on the global flag. This is also a desired method to help migration.

PDoakORNL:
Not having a high level switch between the legacy architecture and the new architecture makes it quite torturous to reject legacy estimators in the qmchamiltonian section as requested in #3875. I support introducing separate code paths for input parsing as the legacy input handling is a mess that is tightly entangled with the legacy simulation objects. 

We can support the current _batch as they are now but I don't think we should make the sphaghetti to go back and check previous parsed and constructured objects like the hamiltonian pool.

ye-luo:
> Not having a high level switch between the legacy architecture and the new architecture makes it quite torturous to reject legacy estimators in the qmchamiltonian section as requested in #3875.

We do need a high level switch. We agreed on that. I was asking to keep "_batch" keep functioning.

src/QMCDrivers/QMCDriverFactory.cpp
```diff
@@ -91,17 +92,37 @@ QMCDriverFactory::DriverAssemblyState QMCDriverFactory::readSection(xmlNodePtr c
...
 
+  // Really by position if you don't write <qmc name="xxx",,,>
+  // you can write <vmc ...> so hacky
   if (curName != "qmc")
     qmc_mode = curName;
 
   const int nchars = qmc_mode.size();
+
+  // Begin to separate batch input reading from the legacy input parsing
+  if (project_data_.get_driver_epoch() == "batched")
```
ye-luo, line 103: Could you add a bool argument (force_batch) of readSection for and expand unit tests of readSection.
Also check `getEngineName()` after the driver being created in unit tests.

PDoakORNL: Since you cannot create a QMCDriverFactory without ProjectData and the global driver epoch is set there, I don't think its a good idea to also allow it to be forced directly.  There is one source of truth for this parameter and it is from the ProjectData the driver factory is constructed with.

The _batch are supported for the convenience of users and tests currently using them but are not valid input when the driver epoch is set.  Based on the driver epoch input, the qmcsystem and hamiltonian node will be parsed differently so we don't want to encourage further use of the _batch or introduce  _legacy as a tag for drivers.

PDoakORNL:
Having `driver` refer to a classification for a group of drivers when its a parameter of a project node, but still having singular drivers types: vmc, linear, dmc in the code, documentation, `name` attibute of qmc node, or when the node name of a qmc section  seems like recipe for confusion.

src/QMCDrivers/QMCDriverFactory.cpp
ye-luo: I forgot that ProjectData is required by QMCDriverFactory. Then there is no need to add a bool argument. I have not read your latest commit but I guess you have both epoch cases covered in the unit test now. What did you mean by "the qmcsystem and hamiltonian node will be parsed differently" did you mean "hamiltonian of qmcsystem" ?

PDoakORNL: We are also going to allow an `<estimators>` node for global estimator definitions.  This will only be valid if in the the batched epoch.

ye-luo:
@PDoakORNL could you run clang-format on all the changed source file ? Some diff looks strange.
I will get you back the review early this afternoon.

src/QMCDrivers/tests/test_QMCDriverFactory.cpp
```diff
@@ -57,34 +62,70 @@ TEST_CASE("QMCDriverFactory create VMC Driver", "[qmcapp]")
...
+    xmlNodePtr node                           = doc.getRoot();
+    QMCDriverFactory::DriverAssemblyState das = driver_factory.readSection(node);
+    REQUIRE(das.new_run_type == QMCRunType::VMC_BATCH);
+
+    auto particle_pool     = MinimalParticlePool::make_diamondC_1x1x1(comm);
+    auto wavefunction_pool = MinimalWaveFunctionPool::make_diamondC_1x1x1(comm, particle_pool);
+    auto hamiltonian_pool  = MinimalHamiltonianPool::make_hamWithEE(comm, particle_pool, wavefunction_pool);
+    std::string target("e");
+    MCWalkerConfiguration* qmc_system = particle_pool.getWalkerSet(target);
+
```
ye-luo, line 93: Could you move 88-92 outside the section and share them between the two sections? Try to reduce duplicated lines. The same for DMC as well.

src/Utilities/ProjectData.h
```diff
@@ -82,8 +86,13 @@ class ProjectData : public OhmmsElementBase
 
   int getSeriesIndex() const { return m_series; }
   int getMaxCPUSeconds() const { return max_cpu_secs_; }
+  const std::string& get_driver_epoch() const { return driver_epoch_; }
+
+  const std::string& getName() const { return name_; }
+  void setName(const std::string& aname) { name_ = aname; }
 
 private:
+  std::string name_;
```
ye-luo, line 95: What is this name? Isn't m_title enough?

src/Utilities/ProjectData.h
```diff
@@ -113,6 +122,8 @@ class ProjectData : public OhmmsElementBase
 
   ///max cpu seconds
   int max_cpu_secs_;
+
+  std::string driver_epoch_;
```
ye-luo, line 126: string is bad. Can we use enum? When being passed in via the constructor, a string make me think it is for m_title.

ye-luo (commented):
The driver factory change looks good.
Having concerns on the changes in ProjectData.

src/Utilities/ProjectData.h
PDoakORNL: In removing pointless and obfuscating inheritance from OhmmsElementBase I preserved the only functionality from the base class it used, having a name member.

src/Utilities/ProjectData.h
PDoakORNL: To deal with this well I have to make InputSection which can handle this kind on input enum able to deal with the application parameter which has many attributes.  I'd have more changes to InputSection coming in EstimatorManagerInput PR's and would prefer not to include it in this one.

src/Utilities/ProjectData.h
ye-luo: I didn't notice the removal of the base class. Yes it was pointless.
I tried removing name_ getName setName. They don't affect qmcpack. So let us clean this up. Some unit test correction is needed.

src/Utilities/tests/test_project_data.cpp
```diff
@@ -38,23 +38,19 @@ TEST_CASE("ProjectData", "[ohmmsapp]")
   REQUIRE(proj1.getTitle().size() > 0);
   REQUIRE(proj1.getTitle() == proj1.getName());
 
-
-  ProjectData proj2("test");
+  ProjectData proj2;
```
ye-luo, line 41: We need to keep pass title via constructor.

src/Utilities/ProjectData.h
ye-luo: We don't need to and maybe should not pull InputSection everywhere. In this use case, just compare strings after reading from XML and keep the enum around should be good enough.

tests/molecules/He_ae/det_He_opt_batch.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
 <simulation>
-  <project id="det_He_opt_batch" series="0"/>
+  <project id="det_He_opt_batch" series="0">
+  </project>
```
ye-luo, line 4: better just restore it as it was.

src/Utilities/ProjectData.cpp
```diff
@@ -14,28 +14,25 @@
...
-  myComm = OHMMS::Controller;
-  if (aname == 0)
-  {
-    m_title = getDateAndTime("%Y%m%dT%H%M");
-    setName(m_title);
-  }
-  else
-    setName(aname);
+  myComm  = OHMMS::Controller;
+  m_title = getDateAndTime("%Y%m%dT%H%M");
```
ye-luo, line 34: This is a breaking change. Please restore setting m_title via constructor. We need a way to test title without pulling xml.

ye-luo (changes requested):
See my comments around ProjectData class change.

src/Utilities/ProjectData.cpp
PDoakORNL: The way I see it you could never set the title.  It was just "asample" if you set aname.  This was never done in the code.  ProjectData doesn't need a name its type isn't polymorphic.

src/Utilities/tests/test_project_data.cpp
PDoakORNL: As noted this never worked. See below. If you passed the name you got that as the class name (never used) and title was always "asample"

PDoakORNL:
Should have addressed @ye-luo 's concerns now.

src/Utilities/tests/test_project_data.cpp
ye-luo: The old code was just bad. I meant we need a way to set the title without relying on XML.
I can image in a driver unit test, we may check certain composite names (used for output). Having a convenient way will be helpful. The intention of existing setTitle was unclear and it was never called. So I just removed it.

src/Utilities/ProjectData.cpp
ye-luo: It was never done in the main code but useful for unit testing. Just like the constructor argument of driver_epoch.

ye-luo:
Test this please

ye-luo (dismissed):

ye-luo (commented):
I wanted to get this merged before I sleep but unfortunately CI was failing on some machines for the need of including `#include <unordered_map>`.
Please also remove "OhmmsData/OhmmsElementBase.h" since the base class is removed
in ProjectData.h

Please pull my latest commits before fixing the issue.

PDoakORNL:
@ye-luo I think the include issue is fixed. Waiting to see what the CI reveals.

ye-luo:
Could you also remove ` #include "OhmmsData/OhmmsElementBase.h" ` since the base class has been removed?

ye-luo:
Test this please

ye-luo (dismissed):

prckent (changes requested):
What did we end up with? I could not determine from the discussion here.

PDoakORNL:
Test this please

ye-luo:
Test this please

ye-luo:
Test this please

ye-luo (dismissed):

src/QMCDrivers/QMCDriverFactory.h
```diff
@@ -48,12 +48,23 @@ class QMCDriverFactory
...
 
   /** default constructor **/
   //QMCDriverFactory() ;
 
-  /** read the current QMC Section */
+  /** read the current QMC Section
+   *  In the application context project data can indicate the input be read in the context of
+   *  the batched driver architecture.
+   *  param[in] cur            qmc section node
+   *  param[in] force_batch    forces input to be evaluated as if project driver type = batched
```
markdewing, line 65: There doesn't seem to be force_batch parameter.

src/QMCDrivers/QMCDriverFactory.h
```diff
@@ -48,12 +48,23 @@ class QMCDriverFactory
     QMCRunType new_run_type = QMCRunType::DUMMY;
   };
 
+  /** Application uses this constructor
+   *  param[in] project_data    this is stored as a reference and this state controls later behavior.
+   *                            For both the driver factory i.e. driver epoch. And the drivers it creates
```
markdewing, line 53: Leftover reference to 'driver epoch'

src/Utilities/ProjectData.h
```diff
@@ -113,6 +132,9 @@ class ProjectData : public OhmmsElementBase
 
   ///max cpu seconds
   int max_cpu_secs_;
+
+  // The driver epoch of the project
```
markdewing, line 136: left over mention of epoch

prckent: Thanks for catching this Mark. We should confirm grep -i epoch returns nothing.

PDoakORNL:
my refactor didn't reach into the comments.  Did a straight text based recursive grep and I don't see any epoch's related to this only the time api.

ye-luo:
Test this please

ye-luo (approved):
LGTM. @prckent good for you now?

prckent (changes requested):
I think we are missing a test for the case where driver_version is not specified but vmc_batch is called explicitly (or one of the other _batched drivers). This should work currently but in future when the driver_version tag is made a requirement will break (and be updated to expect_fail status). A quick clone of one of the deterministic tests would handle this.

prckent (approved):
The no switch + *_batch call is already tested. (This is what current users of the batched code will have in their inputs).

PR #3900, opened 2022-03-10
Prepare MSD for offload

ye-luo:
## Proposed changes
After sorting all the unique determinants, we can dispatch the computation by excitation level.
Before offloading to GPU, we make more functions static (callable inside kernel).
@TApplencourt made the initial change.

- No

epyc-server

## Checklist

### DIFF
--- a/config/build_olcf_summit_Clang.sh
+++ b/config/build_olcf_summit_Clang.sh
@@ -25,7 +25,7 @@ if [[ ! -d /ccs/proj/mat151/opt/modules ]] ; then
   exit 1
 fi
 module use /ccs/proj/mat151/opt/modules
-module load llvm/main-20220119-cuda11.0
+module load llvm/main-20220304-cuda11.0
 
 TYPE=Release
 Compiler=Clang
--- a/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
+++ b/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
@@ -53,7 +53,6 @@ void MultiDiracDeterminant::BuildDotProductsAndCalculateRatios_impl(int ref,
   for (size_t count = 0; count < nitems; ++count)
   {
     const size_t n = *it2;
-    //ratios[count]=(count!=ref)?sign[count]*det0*CustomizedMatrixDet(n,dotProducts,it2+1):det0;
     if (count != ref)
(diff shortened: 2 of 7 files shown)


anbenali (dismissed):

prckent (changes requested):
If the coverage was not confused, we seem to be missing a unit and perhaps deterministic integration test that hits mw_updateRatios_generic. This will only be called if the CI has high enough excitations. Given the upcoming offload work this would be good to add.

ye-luo:
> If the coverage was not confused, we seem to be missing a unit and perhaps deterministic integration test that hits mw_updateRatios_generic. This will only be called if the CI has high enough excitations. Given the upcoming offload work this would be good to add.

I expanded the unit tests to cover all the lower level functions called by mw_updateRatios_generic in `SmallMatrixDetCalculator.h`. I also tested rerouting mw_updateRatios to mw_updateRatios_generic.
It is not easy to construct a unit test for `mw_updateRatios_generic` which requires to instantiate a full MultiDirectDeterminant class with `>5` excitation. We currently relies on MultiSlaterDeterminant unit test to cover MultiDirectDeterminant and indeed need a direct unit test of MultiDirectDeterminant. Once I figure out that, I should have known how to construct a class with high excitation. So I won't add such unit tests in this PR.

mw_updateRatios_generic needs to run on the host because it servers arbitrary matrix sizes.
Consider it is a rare case, we won't offload it.

ye-luo:
Test this please

prckent:
Thanks Ye. Will merge after reading the response from @anbenali : I think we  need to hand construct a wavefunction with high excitations &/or add a FCI wavefunction, which will can trivially have all the excitations we would like even for a low electron count system. Have you tried to make one?

anbenali:
@prckent: We might need to increase the max excitation level in the future, but for now I do not know if it is needed. We can definitely produce one. given the effort needed for that, I would prefer to not block the merger of this PR until we get the unit tests updated for all this sections

anbenali (approved):

prckent:
@anbenali I am not going to slow down this PR, but I am making it a hard requirement that we hit that code path. Please think about generating a test. I expect at least a deterministic test. FCI on a pseudized atom in a large enough basis would do it. A unit test would be preferred. It would be prudent to PR this test ahead of offload code.

prckent (approved):

PR #3901, opened 2022-03-10
Add matrix inversion based on rocSolver for AMD GPUs

markdewing:
Implements #3756 

Assumes that ENABLE_CUDA and QMC_CUDA2HIP are enabled.

Temporary storage used during the inversion is handled automatically by the rocSolver library.  There are API's to manage this explicitly.  We should use the automatic default unless/until it's shown that explicit management is needed.

- New feature

- No

desktop, crusher

## Checklist


ye-luo:
@jakurzak to your best knowledge, does relying on implicit temporal storage managed internally by rocblas impact performance when solving 6000x6000 matrix inversion?

ye-luo:
@markdewing
1. cublas doesn't ship getri but rocblas does. Any chance to use getri? With LAPACK, it does getri in place without a second matrix. So I'm wondering if rocsolver getri could be used and save memory.
2. Did you by chance do  some benchmark? I don't need the numbers on the best GPUs. More about MI50 V100, the timings are 20% with each other on 6144x6144 matrix that type of info. If it is 5x difference then it is concerning.

src/QMCWaveFunctions/Fermion/rocSolverInverter.hpp
```diff
@@ -0,0 +1,205 @@
...
+//
+// File created by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+#ifndef QMCPLUSPLUS_ROCSOLVERINVERTER_H
+#define QMCPLUSPLUS_ROCSOLVERINVERTER_H
+
+#include "OhmmsPETE/OhmmsVector.h"
+#include "OhmmsPETE/OhmmsMatrix.h"
+// This file assumes thats QMC_CUDA2HIP is defined and that creates HIP versions of these functions (despite being labeled with "CUDA")
```
ye-luo, line 18: add a guard
```
#if !defined(QMC_CUDA2HIP)
#error
#endif
```

src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
```diff
@@ -141,11 +141,11 @@ class DelayedUpdateCUDA
   void invert_transpose(const Matrix<T>& logdetT, Matrix<T>& Ainv, std::complex<TREAL>& log_value)
   {
 #if defined(QMC_CUDA2HIP)
-    host_inverter_.invert_transpose(logdetT, Ainv, log_value);
-    initializeInv(Ainv);
+    clearDelayCount();
+    rocsolver_inverter.invert_transpose(logdetT, Ainv, Ainv_gpu, log_value);
 #else
     clearDelayCount();
```
ye-luo, line 147: Could you move clearDelayCount() out of the ifdef. So we avoid repeating the line.

ye-luo (commented):
LGTM.

markdewing:
Notes on getri vs. getrs:
1. getrs has an option to transpose during the solve.  getri does not, so there needs to be a transpose somewhere.
  Possible implementations for the transpose:
  a. There is an BLAS extension, geam, that does general matrix addition and it can perform a transpose.  Still need extra storage for the original and transposed matrix. (this is the solution I used - the extra storage is already there from the getrs implementation)
  b. Write an in-place transpose kernel
  c. Do the transpose on the host before transferring the matrix

3. Temporary storage.  getri uses more temporary storage than getrs.  For small problem sizes, this can be significant relative to the matrix size.
   a. For N=1000, matrix storage is 8MB, dgetrs temporary storage is 2.1MB and dgetri temporary storage is 17.8MB
   b. For N=4000, matrix storage is 128MB, dgetrs temporary storage is 132MB and dgetri temporary storage is 151MB

4. Performance doesn't differ much between getrs and getri. If anything, getrs is a little faster.
   a. On crusher, for N=4000, dgetrs inversion takes 82.5ms and dgetri inversion takes 83.4ms

ye-luo:
@markdewing Thank you for this in-depth analysis. So we just stick to the current choice.

prckent (approved):
Thanks for the implementation and analysis.

This is definitely a big step in the right direction on AMD. If needed, we can refine based  on experience. e.g. The large problem/no spare memory case.

markdewing:
Some timings

N  | GPU (Summit) | GPU (Crusher)
----|---------------------|---------------------
1000 | 10.9 ms       | 8.1 ms
4000 | 96.1 ms       | 82.4 ms
6000 | 214 ms        | 204 ms

Issue #3830, opened 2022-02-11
Multidetrminant offload plan


ye-luo:
- [x] Part 1: All heavy kernels were executed on the GPU. Said Kernels were in batched function but data was generated on host and transferred back and forth too many times between host and accelerator. by #2926 

- [ ] Part 2: All data needed for heavy kernels is now generated in batched code. #3807

- [ ] Part 3: All data needed for heavy kernels is generated on the GPU from the batched code. #3827

PR #3904, opened 2022-03-15
[WIP] Multideterminant Batch code [Part 4 of Offload]

anbenali:

## Proposed changes

- Offload of mw_BuildDotProductsAndCalculateRatios_impl: 

> Variable dotproducts_list calculated on the Device and never transferred back to the host (used only to compute ratios)
> Variable ratios_list (ratio and grad ratio): computed on the GPU and transferred back to CPU.

- Offload of Sherman Morrison:
>Merging mw_DeRatioByColumn and mw_InverseUpdateByColumn:
>Cleaning  excess temp variables to avoid unnecessary transfers Host->Device->Host.  

Function evaluating the orbitals:  Phi->evaluateXXX do not have a batched or GPU implementations and required to send only the host view. Following those calls require either to send the result to the device using the XXX.updateTo() function or later, for those values to be computed on the GPU directly.

This PR belongs to the general effort aiming at generating a GPU implementation of the multideterminant code. https://github.com/QMCPACK/qmcpack/issues/3830

This is part 4: Offload of the table method (Final).

- Other (please describe): GPU capability

-No

## Checklist


src/Numerics/DeterminantOperators.h
```diff
@@ -318,7 +318,7 @@ inline void mw_InverseUpdateByColumn(const int nw,
                                      const RefVector<OffloadVector<T>>& rvec_list,
                                      const RefVector<OffloadVector<T>>& rvecinv_list,
                                      int colchanged,
-                                     std::vector<T> ratiolist)
+                                     OffloadVector<T> ratiolist)
```
markdewing, line 321: It would be better to pass this vector by reference, and it could probably be const (`const OffloadVector<T>& ratiolist`)

PR #3905, opened 2022-03-15
Offload one-body Jastrow ratio calculation for NLPP

ye-luo:
## Proposed changes
Enable J1 NLPP calculation on GPU.
It is not an always win option as pre-existing kernels and multi-thread offload has made the GPU quite busy and moving more computation on to GPU may loose.
Overall, win/loose is within 5% of total walltime. I prefer keeping it on by default. As we further optimize kernels, this will win more.

- New feature

- No

epyc-server

## Checklist


prckent:
Can you please expand on your last comment? If the worst case slowdown is 5-10%, what is the best case speedup that you have seen? And on which GPU? I would imagine that the penalities will only be smaller on future GPUs and with better runtimes, faster hosts etc. i.e. We should have this on as a default.

prckent:
Q. Is this running fully async or with serialization?

prckent (approved):
LGTM (CI still needs to pass)

I think a sensible strategy is to offload everything "obvious" as a first pass and then optimize later. This falls in the obvious category for me given the legacy CUDA code and the non-trivial work here for large batch size, electron, ion count.

ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
@prckent I corrected a unit test which covers the offload code-path now. Need another approval before merging.

prckent (approved):

PR #3906, opened 2022-03-15
Add SYCL device management

ye-luo:
## Proposed changes
Handle device assignment and OpenMP interoperability. Initially developed by @TApplencourt 
I added my `CMake/IntelDPCPPConfig-modified.cmake` for now. The one provided by Intel has multiple issues. Working with Intel via ANL. It is a BSD-3 clause licensed file

- New feature

- No

dell-laptop

## Checklist


prckent (changes requested):
Discussions and justification are needed before adding additional  backend support.

Every backend is a liability.

Would have been better to discuss ahead of time.

Unfortunately I am pretty busy with review writing.  No idea about @PDoakORNL availability,  but you will have to convince both of us.

ye-luo:
> Every backend is a liability.

Understood.

We need MKL for Aurora. Just like cuBLAS/cuSolver which relies on CUDA runtime.
We also need pinned memory which relies on calling vendor native runtime APIs.
Both are performance critical and there is no acceptable workaround so far. We can talk with more details.

src/Platforms/SYCL/SYCLDeviceManager.cpp
```diff
@@ -0,0 +1,171 @@
...
+      default_device_num = sycl_default_device_num;
+    else if (default_device_num != sycl_default_device_num)
+      throw std::runtime_error("Inconsistent assigned SYCL devices with the previous record!");
+
+    default_device_queue =
+        sycl::queue(visible_devices[sycl_default_device_num].context, visible_devices[sycl_default_device_num].device);
+  }
+}
+
+#if defined(_OPENMP)
```
PDoakORNL, line 87: This would be the only code the we can object to here.  What is it needed for other than the SYCLDeviceManager constructor?

src/Platforms/SYCL/SYCLDeviceManager.h
```diff
@@ -0,0 +1,50 @@
...
+{
+
+struct syclDeviceInfo
+{
+  sycl::context context;
+  sycl::device device;
+};
+
+#if defined(_OPENMP)
+std::vector<struct syclDeviceInfo> xomp_get_sycl_devices();
```
PDoakORNL, line 32: does this need to be in the global qmcplusplus namespace?

src/Platforms/SYCL/SYCLDeviceManager.h
```diff
@@ -0,0 +1,50 @@
...
+class SYCLDeviceManager
+{
+  int sycl_default_device_num;
+  std::vector<syclDeviceInfo> visible_devices;
+  sycl::queue default_device_queue;
+
+public:
+  SYCLDeviceManager(int& default_device_num, int& num_devices, int local_rank, int local_size);
+
+  sycl::queue getDefaultDeviceQueue() const { return default_device_queue; }
```
PDoakORNL, line 46: This is an expansion of the defacto device manager API that only sycl has, is it missing in the other managers or is it just a result CUDA's implicit default stream?  Is the default queue equivalent to the default stream?

PDoakORNL (changes requested):
This expands are platform support to include SYCL for purpose of supporting intel its library/API and future hardware.  I would expect the device management API to be one of the most vendor specific elements so we should expect to have one of these for each platform. HIP being a special case. And its not surprising that the API requirement for that manager is not the same as for CUDA or OMPTarget but we should document this, why?

The global function I see it called in the constructor but is it global for some other reason?

src/Platforms/SYCL/SYCLDeviceManager.h
ye-luo: Move it to cpp and marked it static. It is only needed in the file scope.

src/Platforms/SYCL/SYCLDeviceManager.cpp
ye-luo: CUDA has a primary context notion while SYCL requires manage context/devices explicitly. Thus we need OpenMP interop to make sure a unique context is shared between SYCL and OpenMP parts.

src/Platforms/SYCL/SYCLDeviceManager.h
ye-luo: Bingo. CUDA has default context and default stream while SYCL doesn't. So we need to take the default queue to manage memory. I updated all the documentation.

PDoakORNL (approved):

prckent (commented):
Please can you confirm the original source of IntelDPCPPConfig-modified.cmake? From the Intel OneAPI package and not yet upstreamed to CMake/Kitware? I see that it is BSD3 licensed.

prckent:
Test this please

ye-luo:
> Please can you confirm the original source of IntelDPCPPConfig-modified.cmake? From the Intel OneAPI package and not yet upstreamed to CMake/Kitware? I see that it is BSD3 licensed.

The orignal source of IntelDPCPPConfig-modified.cmake is from OneAPI distribution. About upstream to CMake/Kitware, my record was, SYCL was considered C++ and library, so there is no need of being treated as a language like CUDA and HIP. Thus CMake didn't accept it. This may change in the future but not now.

prckent (approved):

PR #3909, opened 2022-03-18
Fix incorrect use of DTModes

ye-luo:
## Proposed changes
J3 in offload was broken by #3905 because virtual particle set data transfer no more depends on NEED_TEMP_DATA_ON_HOST.
This PR adds missing flags in AA table consumers who needs data on the host.
and removes implicit NEED_TEMP_DATA_ON_HOST always forced on AA, AB tables.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):

PR #3910, opened 2022-03-20
Adjust block sizes in offload kernels

ye-luo:
## Proposed changes
The full loop iterations space is spliced into sizes of ChunkSizePerTeam to increase the number of  CUDA blocks/ workgroups on GPU.

Larger block sizes allow more resident computation and thus more efficient but lose time to solution when the number of blocks. For small problem sizes, the number of blocks is not an issue and efficiency matters more. I noticed much closer kernel performance between legacy CUDA and OpenMP offload.

A larger block size only affects iterations space larger than the block size, namely when the number of blocks changes.
The increase from 128 to 512 actually improve kernel efficiency a lot for small to medium problem sizes with smaller CUDA grid sizes. For example NiO a32 has 288 splines, so need 3 blocks of size 128, 2 blocks of 256, and 1 blocks of 512. 512 should be good for both NVIDIA and AMD, 64 CUDA cores per SM and 64 Stream processors per CU.
With 128, 256, 512, I saw kernel time dropping but 1024 stays the same as 512 accross all problem sizes.

For large problem sizes, the number of walkers drop and thus keeping more blocks makes better use of the device.  So I think 512 is a sweat spot.

- Other (please describe): Performance tuning.

- No

epyc-server

## Checklist

### DIFF
--- a/config/build_olcf_summit_Clang.sh
+++ b/config/build_olcf_summit_Clang.sh
@@ -25,7 +25,7 @@ if [[ ! -d /ccs/proj/mat151/opt/modules ]] ; then
   exit 1
 fi
 module use /ccs/proj/mat151/opt/modules
-module load llvm/main-20220304-cuda11.0
+module load llvm/main-20220317-cuda11.0
 
 TYPE=Release
 Compiler=Clang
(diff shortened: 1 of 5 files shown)


prckent (approved):

prckent:
Test this please

PR #3919, opened 2022-03-25
Reflect HIP build options in printout

ye-luo:
## Proposed changes
ENABLE_ROCM is a CMake option not a macro. It should not be used in the source codes.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

src/QMCApp/QMCMain.cpp
```diff
@@ -124,18 +124,17 @@ QMCMain::QMCMain(Communicate* c)
...
+#if !defined(ENABLE_OFFLOAD) && !defined(ENABLE_CUDA) && !defined(QMC_CUDA) && !defined(ENABLE_HIP)
   app_summary() << "  CPU only build" << std::endl;
 #else
 #if defined(ENABLE_OFFLOAD)
   app_summary() << "  OpenMP target offload to accelerators build option is enabled" << std::endl;
 #endif
-#if defined(ENABLE_CUDA) || defined(QMC_CUDA)
+#if defined(ENABLE_HIP) || defined(QMC_CUDA2HIP)
+  app_summary() << "  HIP acceleration build option is enabled" << std::endl;
+#elif defined(ENABLE_CUDA) || defined(QMC_CUDA)
```
PDoakORNL, line 135: I think it would be better if these had distinct outputs
i.e. 
legacy CUDA acceleration build option is enabled
CUDA acceleration build option is enabled

src/QMCApp/QMCMain.cpp
```diff
@@ -124,18 +124,17 @@ QMCMain::QMCMain(Communicate* c)
...
   app_summary() << std::endl;
-#if !defined(ENABLE_OFFLOAD) && !defined(ENABLE_CUDA) && !defined(QMC_CUDA) && !defined(ENABLE_ROCM)
+#if !defined(ENABLE_OFFLOAD) && !defined(ENABLE_CUDA) && !defined(QMC_CUDA) && !defined(ENABLE_HIP)
   app_summary() << "  CPU only build" << std::endl;
 #else
 #if defined(ENABLE_OFFLOAD)
   app_summary() << "  OpenMP target offload to accelerators build option is enabled" << std::endl;
 #endif
-#if defined(ENABLE_CUDA) || defined(QMC_CUDA)
+#if defined(ENABLE_HIP) || defined(QMC_CUDA2HIP)
```
PDoakORNL, line 133: This is probably out of scope for this PR.

We have too many hip options.  If ENABLE_HIP is on QMC_CUDA2HIP must be on since you are by definition on a HIP system and if the CUDA is built it will be broken.
I don't see any good reason to allow QMC_CUDA2HIP and not ENABLE_HIP.

However as long as all those options exist and can be used in different combination they should probably give separate output.  Otherwise you need to have access to the build options to really know what you are running anyway.

PDoakORNL (changes requested):

src/QMCApp/QMCMain.cpp
ye-luo: That is not true. ENABLE_HIP is direct HIP source code in AFQMC.

src/QMCApp/QMCMain.cpp
ye-luo: I think it is better to expand all cases to be accurate.

src/QMCApp/QMCMain.cpp
ye-luo: Updated printout.

ye-luo:
Test this please

src/QMCApp/QMCMain.cpp
ye-luo: @PDoakORNL how about the current printout?

PDoakORNL (approved):

PR #3923, opened 2022-03-30
Move QMC_CUDA description to a legacy CUDA build section 

markdewing:
Update the installation documentation.

* Move the QMC_CUDA variable to a section about legacy CUDA build
* Add an AFQMC build section
* Remove ENABLE_CUDA from the general options.  Add to the AFQMC section.  If we still want the description for real-space, it can be added to the offload section.
* Keep ENABLE_OFFLOAD and add a link to the offload build section right below the description block for general variables.

The README.md file in the root directory should be updated to match.  Will do that if/when these changes get accepted.

- Documentation changes

- No

laptop

## Checklist

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -280,20 +280,18 @@ the path to the source directory.
 
   ::
 
-    QMC_CUDA              Enable legacy CUDA code path for NVIDIA GPU acceleration (1:yes, 0:no)
     QMC_COMPLEX           Build the complex (general twist/k-point) version (1:yes, 0:no)
     QMC_MIXED_PRECISION   Build the mixed precision (mixing double/float) version
                           (1:yes (QMC_CUDA=1 default), 0:no (QMC_CUDA=0 default)).
                           Mixed precision calculations can be signifiantly faster but should be
                           carefully checked validated against full double precision runs,
                           particularly for large electron counts.
-    ENABLE_CUDA           ON/OFF(default). Enable CUDA code path for NVIDIA GPU acceleration.
-                          Production quality for AFQMC. Pre-production quality for real-space.
-                          Use CMAKE_CUDA_ARCHITECTURES, default 70, to set the actual GPU architecture.
     ENABLE_OFFLOAD        ON/OFF(default). Enable OpenMP target offload for GPU acceleration.
     ENABLE_TIMERS         ON(default)/OFF. Enable fine-grained timers. Timers are on by default but at level coarse
                           to avoid potential slowdown in tiny systems.
                           For systems beyond tiny sizes (100+ electrons) there is no risk.
+
+See the section below, :ref:`offload_build`, for more details on how to configure the OpenMP offload build for various GPU types and compilers.
 
 - General build options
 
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -336,6 +334,22 @@ the path to the source directory.
                            if the build is on a separate filesystem from the source, as
                            required on some HPC systems.
 
+- Legacy CUDA version of QMCPACK
+
+  ::
+
+    QMC_CUDA      Enable legacy CUDA code path for NVIDIA GPU acceleration (default: OFF)
+    QMC_CUDA2HIP  AMD GPU support by translating CUDA to HIP (default: OFF)
```
ye-luo, line 342: QMC_CUDA2HIP is not for Legacy CUDA only. We can put all the GPU related option together.

GPU accelleration of QMCPACK
ENABLE_OFFLOAD. For all OpenMP offload code, and point to the section for details.

NVIDIA GPU:
ENABLE_CUDA
QMC_CUDA mention no AFQMC

AMD GPU:
EANBLE_CUDA + QMC_CUDA2HIP legacy CUDA, mention no AFQMC.
QMC_CUDA + QMC_CUDA2HIP legacy CUDA, mention no AFQMC
ENABLE_HIP AFQMC need this.

ye-luo (changes requested):

docs/installation.rst
```diff
@@ -336,6 +334,22 @@ the path to the source directory.
...
+
+- Auxiliary Field Quantum Monte Carlo
+
+  ::
+
+    BUILD_AFQMC  Build AFQMC-capable code (default: OFF)
```
ye-luo, line 348: AFQMC is a physics feature option. Let us put it at QMC_COMPLEX

docs/installation.rst
```diff
@@ -336,6 +334,22 @@ the path to the source directory.
...
+    ENABLE_HIP   AMD GPU support through HIP (default: OFF)
+    ENABLE_CUDA  NVIDIA GPU support through CUDA (default: OFF)
```
ye-luo, line 350: Although ENABLE_HIP is AFQMC exclusive, we should present as a GPU option not a AFQMC option
ENABLE_CUDA is not AFQMC only.

docs/installation.rst
markdewing: The directions for AMD AOMP compiler don't include ENABLE_CUDA or QMC_CUDA2HIP.   Should they be there?  If we just list the option it's not clear when it should be used.

markdewing:
My guess is that users will choose what type of QMCPACK they want to compile first (real space, legacy CUDA, AFQMC).  Then figure out which options affect that type.  The intent is to follow that decision tree and put AFQMC and legacy CUDA in their own sections along with the flags that affect them.

docs/installation.rst
ye-luo: We will need to update the AOMP recipe with ENABLE_CUDA or QMC_CUDA2HIP added. I think that was enabled after the original AOMP recipe was added.

ye-luo:
I except users to figure out real-space, AFQMC, real/complex for their physics problem. Then think about CPU or GPU acceleration. That is why at the GPU section, we should highlight was physics doesn't work with a GPU option.

PR #3927, opened 2022-04-04
Fix incorrect reference determinant id.

ye-luo:
## Proposed changes
@shivupa reported to me this bug. Once determinants are reordered #3874.
`ReferenceDeterminant = ref_det_id;` is wrong.
The reference one should be always id 0.
I had some difficulty to add unit test.
1) reference det is chosen with the maximal weight. It is not exposed in input file.
2) current construction of MSD relies on the builder, namely the input file.

To make a test, refactoring certain code will be needed but it is better to do that after pending MSD PRs.
I did some tests by forcing refdet_id directly in SlaterDetBuilder.cpp. No failure so far.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.cpp
+++ b/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.cpp
@@ -35,8 +35,6 @@ void MultiDiracDeterminant::createDetData(const int ref_det_id,
                                           const std::vector<size_t>& C2nodes_unsorted,
                                           std::vector<size_t>& C2nodes_sorted)
 {
-  ReferenceDeterminant = ref_det_id;
-
   auto& ref                        = configlist_unsorted[ref_det_id];
   auto& configlist_sorted          = *ciConfigList;
   auto& data                       = *detData;
@@ -463,7 +461,6 @@ MultiDiracDeterminant::MultiDiracDeterminant(const MultiDiracDeterminant& s)
       NumPtcls(s.NumPtcls),
       LastIndex(s.LastIndex),
       ciConfigList(s.ciConfigList),
-      ReferenceDeterminant(s.ReferenceDeterminant),
       is_spinor_(s.is_spinor_),
       detData(s.detData),
       uniquePairs(s.uniquePairs),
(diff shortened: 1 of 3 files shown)


kgasperich:
I agree that it looks like an error was introduced with the reordering, but I'm still not sure that this fixes everything.
It looks like my earlier PR was conflating the MultiDiracDet reference id with the MultiSlaterDet reference id.
The reference SlaterDet id will be the index of the determinant with the largest weight, but the reference DiracDet id should always be 0 (within the list of unique Dirac dets).
The part that I still don't understand is that after [we get `refdet_id` by identifying the largest coefficient](https://github.com/ye-luo/qmcpack/blob/625a354cac82c02575e3d29bcfd25df16fae17c0/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp#L515-L518), we use that same id to [identify the reference within the list of unique configs](https://github.com/ye-luo/qmcpack/blob/625a354cac82c02575e3d29bcfd25df16fae17c0/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.cpp#L38), but what we should be passing as the first argument to `MultiDiracDeterminant::createDetData` is the index within the unique configs corresponding to the Slater Det with the largest coef.
If I'm remembering the storage correctly, then this should be `C2nodes[grp][refdet_id]` (I'm assuming that for SlaterDet `i`, `C2nodes[grp][i]` gives the index within the list of unique Dirac dets from group `grp`)

ye-luo:
> The part that I still don't understand is that after [we get `refdet_id` by identifying the largest coefficient](https://github.com/ye-luo/qmcpack/blob/625a354cac82c02575e3d29bcfd25df16fae17c0/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp#L515-L518), we use that same id to [identify the reference within the list of unique configs](https://github.com/ye-luo/qmcpack/blob/625a354cac82c02575e3d29bcfd25df16fae17c0/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.cpp#L38), but what we should be passing as the first argument to `MultiDiracDeterminant::createDetData` is the index within the unique configs corresponding to the Slater Det with the largest coef.

Isn't this what we are doing?
```
auto& ref = configlist_unsorted[ref_det_id];
```
the first argument input `ref_det_id` is used with unsorted lists only.

ye-luo:
@kgasperich I see what you meant. Let me add a fix.

ye-luo:
Test this please

kgasperich:
looks good to me

shivupa:
Looks good to me as well. I was able to test it now actually. Thank you

prckent (approved):
LGTM, although I didn't spot the original problem.

+Kevin confirms OK in the discussion above.

When we have bandwidth, we need a test inspired by @shivupa 's problem to avoid any recurrence of this flavor of indexing bug.

PR #3938, opened 2022-04-08
Add elemental core sizes up to Rn for convert4qmc

tiihonej:
## Proposed changes

Add new elements up to Rn to ´QMCGaussianParserBase::gCoreTable´, because the old table was not covering the latest ccECP additions, such as W, Ir, Au and Bi. For example, trying to convert GAMESS orbitals including Au would result in the following error: ´ERROR convert4qmc: IonSet creation failed in QMCGaussianParserBase::dump, check out of bounds for valence calculation´.

I've tried to guess reasonable core sizes for those elements not presently available as ccECPs. There's an issue of ambiguity (different core sizes for one atom), which needs an entirely different implementation to be solved, but maybe the best way forward is to prefer the ccECPs as long as it's not a problem to any active user.

- Bugfix
- New feature

- No

GAMESS orbital conversion with Au now works.

## Checklist


prckent (approved):
Thanks Juha.

The fact the some guessing is required indicates that the algorithm is wrong. We can fix that if/when people hit the problem of inconsistent core sizes.

prckent:
Test this please

PR #3951, opened 2022-04-14
Add ResourceHandle

ye-luo:
## Proposed changes
When using multiple-walker resource, the current scheme is to add a unique_ptr. This stops compiler auto-generated copy constructor and requires extra code. In the way of using resources, we never need exact copying of individual resources. We always copy by the collection. For this reason, this PR introduces ResourceHandle to implement copy as no-op.

- New feature

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):

prckent:
Test this please

PR #3952, opened 2022-04-15
Connect CoulombPBCAA to multi walker resource

ye-luo:
Need #3951
## Proposed changes
1. Connect CoulombPBCAA to multi walker resource. This change reduces allocation/de-allocation of value_offload but due to its low call count, performance gain is negligible. Anyway, reducing noise to the runtime is beneficial.
2. When we have two resources from coulomb and NLPP. WFOpt shows a bug of using the same resource collection for two hamiltonians. In CostFunction, h and h_0 have different components and thus cannot use the same resource collection.

- Bugfix
- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCHamiltonians/CoulombPBCAA.cpp
+++ b/src/QMCHamiltonians/CoulombPBCAA.cpp
@@ -19,6 +19,7 @@
 #include "EwaldRef.h"
 #include "Particle/DistanceTable.h"
 #include "Utilities/ProgressReportEngine.h"
+#include <ResourceCollection.h>
 #include "Numerics/OneDimCubicSplineLinearGrid.h"
 
 namespace qmcplusplus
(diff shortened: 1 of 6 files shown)


src/QMCHamiltonians/CoulombPBCAA.cpp
```diff
@@ -538,9 +537,10 @@ std::vector<CoulombPBCAA::Return_t> CoulombPBCAA::mw_evalSR_offload(const RefVec
   const auto delta_inv   = caa_leader.rVs_offload->get_delta_inv();
   const auto Zat         = caa_leader.Zat_offload->data();
 
-  auto value_ptr = values_offload.data();
-
   {
+    values_offload.resize(nw);
+    std::fill_n(values_offload.data(), nw, 0);
```
prckent, line 542: Is this initialization strictly required? Can we avoid it by writing the penultimate line of the offload loop below not as a reduction but an assignment?

ye-luo: This is strictly required. The zeroing serves the reduction over chunks while the offload only serves one chunk. So it must be done outside the chunk loop.

ye-luo:
Test this please

prckent (approved):

PR #3957, opened 2022-04-20
Default CUDA host compiler the same as CXX compiler

ye-luo:
## Proposed changes
I noticed an issue with upstream clang which changes its default to PIE enabled.
Our nvcc setting default using gcc as the host compiler, this caused issue at linking on SLES 15 machines.
gcc doesn't have PIE enabled by default. This choice is made by the linux distro. Ubuntu has PIE enabled by default.
One fix is to pass -fPIE via CMAKE_CUDA_FLAGS but I dislike this solution.
Instead, I choose to set CMAKE_CXX_COMPILER as the default value of CMAKE_CUDA_HOST_COMPILER if user doesn't specify one.

This change needs additional changes.
1) -allow-unsupported-compiler by default. It is likely that we use newer compiler than nvcc support claimed.
2) option used by 1) requires CUDA 11. Right now all the production machine has CUDA 11. So I raised the bar.

- Build related changes

- No

epyc-server/bora/jlse/perlmutter

## Checklist


ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
Test this please

prckent (approved):

prckent:
Test this please

Issue #3922, opened 2022-03-30
MSD offload code sensitive to CUDA toolkit versions


ye-luo:
**Describe the bug**
single det doesn't seem being affected but 
deterministic-diamondC_2x1x1-gaussian_pp_MSD-vmcbatch-dmcbatch-mwalkers_sdj-1-4 and 1-16
https://cdash.qmcpack.org/CDash/viewTest.php?onlyfailed&buildid=301858
fails with CUDA >=11.3. Checked up to 11.7.
https://github.com/llvm/llvm-project/issues/54633
If you need multi det code path, be sure to use older CUDA version.

prckent:
11.2 is no longer installed; switching to 11.8. If the problem is still there the bug will just need tackling &/or the test skipping in CI

ye-luo:
Workaround #4773 applied

ye-luo:
It seems that CUDA 12.3 no more produces failure.

PR #3959, opened 2022-04-20
Force CUDA 11.2 in Offload CI

prckent:
## Proposed changes

Speculative fix for  https://github.com/QMCPACK/qmcpack/issues/3922 . I am not sure this is a sufficient change but it presumably needs merging to verify and e.g. #3921 retriggered.

- Bugfix

- No

None.

## Checklist


ye-luo:
This won't work. CUDACXX only affects the CUDA code compilation. But we need to make the clang seeing CUDA 11.2, namely make `clang++ -v` showing CUDA 11.2 instead of 11.6

PR #3961, opened 2022-04-20
spherical harmonic wrapper

camelto2:

## Proposed changes

I use the Ylm's in Numerics for the spin-orbit ECP, and will need cartesian derivatives of these for ionic derivatives. The Ylm's in Numerics require cartesian vectors as input...except they **must**  unit vectors and must be in the format [z,x,y] instead of the standard ordering. 

This just adds a sphericalHarmonic wrapper which takes in any cartesian vector so the user can simply provide whatever position they want. Internally, it calculates the unit vector and reorders it to be [z,x,y] and then calls the Ylm implementation we already have. This allows me to clean up the SOECP for better readability.

I also introduce a sphericalHarmonicGrad, which calculates (d/dx, d/dy, d/dz)Ylm. This will be need for ionic derivatives of the SOECP.

- New feature

- No

my mac

## Checklist

### DIFF
--- a/src/Numerics/Ylm.h
+++ b/src/Numerics/Ylm.h
@@ -139,5 +139,32 @@ inline std::complex<T> sphericalHarmonic(const int l, const int m, const TinyVec
   return Ylm(l, m, unit);
 }
 
+// get cartesian derivatives of spherical Harmonics. This takes a arbitrary position vector (x,y,z) and returns (d/dx, d/dy, d/dz)Ylm
+template<typename T>
+inline void sphericalHarmonicGrad(const int l,
+                                  const int m,
+                                  const TinyVector<T, 3>& r,
+                                  TinyVector<std::complex<T>, 3>& grad)
+{
+  TinyVector<T, 3> unit;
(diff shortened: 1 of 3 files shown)


src/Numerics/Ylm.h
```diff
@@ -127,5 +127,44 @@ inline void derivYlmSpherical(const int l,
   }
 }
 
+// wrapper for Ylm, which doesn't require a unit vector and is in the normal order (x,y,z)
```
ye-luo, line 130: In the function which needs (x,y,z) ordering, please add a note in the function documentation.
Here specify the ordering needed by this function.

rcclay (approved):
LGTM

src/Numerics/Ylm.h
camelto2: added and updated comments. Should be more clear

ye-luo:
Test this please

ye-luo (approved):

prckent:
Thanks for maintaining the high test coverage in Ylm.h

PR #3964, opened 2022-04-20
add n32 quadrature rule and refactor SoaSphericalTensor to Numerics

camelto2:

## Proposed changes

The pseudopotential quadrature rules that were implemented are missing the npts=32 case as defined in the [Mitas paper](https://aip.scitation.org/doi/10.1063/1.460849 ). This PR adds the N=32 icosahedron rule as an intermediate between the n=26 and n=50 rules.

This came up because @aannabe is testing a Tb pseudopotential and noticed exceptionally large variances with default settings for the isolated atom. He tracked it down to needing to increase the quadrature grid significantly. In his case, the energy and variance were converged by the npts=26 case (nrule=6), however there could be circumstances where we need more than npts=26. Having a npts=32 option could be sufficient in those circumstances and is cheaper than going all the way up to npts=50. (note: for this potential, the default choice is npts=12 which results in a variance of almost 400! whereas it converges to about 17 with both npts=26 and npts=50. This indicates that we may need to rethink our default quadrature grid choices).

Note: to be consistent with how the rules are defined, i.e. increasing rule number with increasing number of quadrature points, I changed nrule=7 to be the npts=32 case and made a new nrule=8 be the npts=50 case. This is a breaking change for inputs that expect nrule=7 to be the npts=50...but I expect those cases to be rather small. 

- New feature
- Refactoring

- Yes...only if your older inputs expect nrule=7 to have 50 quadrature points. Now it will have 32 and will need to change to nrule=8 to recover the previous behavior. I expect this will impact very few users

my mac

## Checklist


rcclay (dismissed):
LGTM.  
By the way, it looks as though there is a run-time quadrature grid check on construction, so I don't think a unit test is necessary for this small change.

ye-luo:
Could you move `TEST_CASE("CheckSphericalIntegration", "[hamiltonian]")` test at `QMCHamiltonians/tests/test_ecp.cpp`
to `Numerics/tests/test_Quadrature.cpp` and raise the max to 8?

camelto2:
> Could you move `TEST_CASE("CheckSphericalIntegration", "[hamiltonian]")` test at `QMCHamiltonians/tests/test_ecp.cpp` to `Numerics/tests/test_Quadrature.cpp` and raise the max to 8?

will do

prckent (dismissed):
OK on the "breaking" change.

Please think more about how we could change the defaults and grid settings. The current situation is not ideal even for light atoms imo - there are/were old issues on the topic.

src/Numerics/tests/CMakeLists.txt
```diff
@@ -24,13 +24,14 @@ set(UTEST_SRCS
...
+    test_one_dim_cubic_spline.cpp
+    test_Quadrature.cpp)
 
 # Run gen_gto.py to create these files.  They may take a long time to compile.
 #SET(UTEST_SRCS ${UTEST_SRCS} test_full_cartesian_tensor.cpp test_full_soa_cartesian_tensor.cpp)
 
 add_executable(${UTEST_EXE} ${UTEST_SRCS})
 
-target_link_libraries(${UTEST_EXE} catch_main containers qmcio)
+target_link_libraries(${UTEST_EXE} catch_main containers qmcio qmcwfs)
```
ye-luo, line 35: Is qmcwfs really needed?

camelto2: the internal test in quadrature calls SoaSphericalTensor to do its checks. SoaSphericalTensor is in the qmcwfs library, so that was the only was I could get it to compile without undefined symbol errors

camelto2: if there is a simpler solution let me know

aannabe:
Just for the reference, here is the data that was mentioned before:

image

As this is related to issue #557, I think it might be worth defaulting lmax=4 to N_quad = 32. At least in this case, it seems to be the appropriate choice.

src/Numerics/tests/CMakeLists.txt
ye-luo: How about move SoaSphericalTensor.h/cpp under Numerics?

camelto2: that should work. Will give it a go

camelto2: refactored SoaSphericalTensor and SoaCartessianTensor into Numerics. The Quadrature test doesn't use the SoaCartesianTensor, but it seemed like it should be in Numerics as well

ye-luo (approved):
LGTM.

ye-luo:
Test this please

Issue #3789, opened 2022-01-31
WFOpt use_nonlocalpp_deriv feature status


ye-luo:
**Is your feature request related to a problem? Please describe.**
In WFOpt, both kinetic energy and NLPP depends on WF optimizable parameters.

1. TWF::evaluateDerivatives compute both \partial_\alpha log(psi) and  \partial_\alpha KE.
so this function is currently a pure virtual function, namely, Each WFC must implement it. If a given WFC doesn't carry any optimization parameter. it still needs to implement a empty function to make zero contribution.

2. If use_nonlocalpp_deriv=yes, and NLPP algorithm=non-batched. This is a very slow code path due to repeatedly accept/reject moves.
* NLPP calls TWF::evaluateDerivativesWF which computes \partial_\alpha log(psi).
* Each WFC being part of TWF needs to implement WFC::evaluateDerivativesWF. The current base class implementation is runtime error stopper. The is intended to stop users and figure out if an zero contribution implementation is appropriate.
* We have meaningful implementation in MultiSlaterDetTableMethod, J2OrbitalSoA, J2OMPTarget, J1OrbitalSoA, J1Spin although the implmenetaiton was done quick and dirty.
* DiracDeterminantBase implements empty function to make tests without orbital optimization run. This is wrong when orbital optimization is enabled.

3. If use_nonlocalpp_deriv=yes, and NLPP algorithm=batched. This is a much faster code path.
* NLPP calls TWF::evaluateDerivRatios which computes \partial_\alpha log(psi(Rq)) log(psi(R)).
* Each WFC being part of TWF needs to implement WFC::evaluateDerivativesWF. The current base class implementation is zero contribution. I'm inclinde to change the base class implementation to runtime error. Convenience vs safety, I'm favoring safety.
* We have  meaningful implementation in J2OrbitalSoA, J2OMPTarget, J1OrbitalSoA only.
* Both single and multi Slater determinant implementations rely on base class zero contribution implementation. This is wrong when orbital optimization is enabled.

I'm inclinded to phasing out 2 and further develop 3.

We can change use_nonlocalpp_deriv to yes by default, the only breaking feature is MSD because NLPP algorithm=batched is default and case 3 adds zero contribution.

prckent:
Are we still missing implementation for J3 with use_nonlocalpp_deriv=yes and NLPP algorithm=batched?

ye-luo:
[...]

Issue #3791, opened 2022-01-31
WFOpt nonlocalpp option


ye-luo:
When this option is yes, WFOpt recompute NLPP for each given set of parameters during correlated sampling.
What is the motivation to set this to "no"? Save some computation at the cost of correctness.
It just doesn't make sense to me.

prckent:
For a few % of the time this will give a significant speedup (say, near convergence), but for most % of the time this will either slow convergence or perhaps even preclude it. Given how QMCPACK is used today I think we could safely lose the "no" option.

PR #3969, opened 2022-04-22
Turn on NLPP energy derivatives by default.

### LINKED ISSUE #3968: Unusual behavior of "OneShiftOnly" and comparison of other optimizers
**Describe the bug**
An unusual behavior was observed for the Tb atom using `OneShiftOnly` optimizer. See the comparison of optimizers for energy and variance.

energy
variance
ratio

**To Reproduce**
1. Built from a recent commit 0ac62ea27694550d1698ae75dffacd67e42ba75b
2. Compile script is attached.

**System:**
 - system name: NERSC/Cori-KNL
 - modules loaded: See attached file

**Additional context**
All optimizers use the same parameters:
```
                linopt1 = linear(
                    energy               = 0.9,
                    unreweightedvariance = 0.0,
                    reweightedvariance   = 0.1,
                    samples              = int(1e6),
                    substeps             = 3,
                    warmupSteps          = 20,
                    blocks               = 50,
                    nonlocalpp           = True,
                    usedrift             = True,
                    minmethod            = optimizer,
                    minwalkers           = 0.1,
                    timestep             = 0.5,
                    )
```
Other parameters are default ones.

Using the same resources, the execution times are given below:
```
Optimizer	Execution Time (secs)
quartic		1641
apative		895.4
OneShiftOnly	359
descent		269.5
```
[optimizers_compare.zip](https://github.com/QMCPACK/qmcpack/files/8543611/optimizers_compare.zip)

ye-luo:
## Proposed changes
Requested by #3789
Fixes #3968

- Documentation changes
- Other (please describe): Default changed.

- No

epyc-server

## Checklist


lshulen:
In changing the default to include the nonlocal pp derivs, I was surprised to see that including the NLPP energy in the cost function is still turned off by default.  Should that also be changed?

ye-luo:
> In changing the default to include the nonlocal pp derivs, I was surprised to see that including the NLPP energy in the cost function is still turned off by default. Should that also be changed?

I'm also surprised. Unfortunately, the code won't fix itself.

That is not on/off change and requires additional cleaning. nonlocalpp input actually allows the customized name of NLPP XML entry and there are look-up mechanism. I just mean need more code reading and at least a separate PR.

lshulen:
Is a subsequent PR changing the NLPP energy on by default forthcoming, or should we make an issue to document the need?

ye-luo:
There is an issue #3791

prckent:
"I was surprised to see that including the NLPP energy in the cost function is still turned off by default". So not only do we ignore the gradient (pre this PR) but the value was not included! Crazy! Unfortunately the app won't fix itself. Hopefully we have enough eyes on the code that nothing like that would get in these days (@PDoakORNL , @ye-luo , @markdewing  etc.)

lshulen (changes requested):
I'm happy with the direction.  I'd like to request a further change in the documentation and then I'll be happy to approve.  Could the description for the nonlocalpp variable to be changed to reflect current understanding of that feature?  Perhaps:
include non-local PP energy in the cost function --> Re-evaluate nonlocal pp energy in correlated sampling optimizations?

ye-luo:
> I'm happy with the direction. I'd like to request a further change in the documentation and then I'll be happy to approve. Could the description for the nonlocalpp variable to be changed to reflect current understanding of that feature? Perhaps: include non-local PP energy in the cost function --> Re-evaluate nonlocal pp energy in correlated sampling optimizations?

@lshulen Indeed. I'm reading the code again and making a few test run to confirm my understanding. I will need to update the issue and also update documentation. It will take a bit time and for now I'd like to this PR merged to minimize PRs on my table.. Does it sound OK for you?

lshulen (approved):

PR #3975, opened 2022-04-25
Mandate CMake 3.20 when using SYCL and add a compiler workaround.

ye-luo:
## Proposed changes
SYCL itself is not a CMake language. So it doesn't require a minimal CMake version.
However we need the llvm-based Intel compiler and I need to add a workaround for some compiler issues. The workaround will be based on compiler version which needs CMake 3.20 to determine the IntelLLVM compiler ID and version. For this reason, cmake 3.20 is required by ENABLE_SYCL.

- Build related changes

- No

dell-laptop

## Checklist


ye-luo:
Test this please

PR #3977, opened 2022-04-26
Remove use_stored_copy flag in orbital rotation code

markdewing:
Remove the 'use_stored_copy' flag from the functions that apply orbital rotations.

The LCAO orbital rotation stores a copy of the coefficient matrix (C is the matrix, C_copy is the copy). It multiplies the rotation matrix by C_copy and puts the result in C.
The stored copy is needed for parameter optimization.  When the rotation parameters change, they need to be reapplied to the original coefficient matrix.

Previously, use_stored_copy is false by default.  The logic in apply_rotation would perform the copy (and initialization) C_copy = *C if use_stored_copy is false (e.g., by default it uses the values in C to multiply by the rotation matrix.)
I find this logic hard to follow and would rather explicitly require initialization of C_copy via the storeParamsBeforeRotation function. (That function already exists and is already called in code prior to this PR so using it isn't adding complexity.)

One tricky part of the code path is to ensure storeParamsBeforeRotation doesn't get called after applyRotation has been called, otherwise the rotation will get applied multiple times.

This PR also adds the C_copy variable in the LCAOrbitalSet copy constructor, which is necessary because the changes in this PR don't copy/initialize C_copy as frequently.  The change may not be obvious because the clang formatting changes the layout.

- No

laptop

## Checklist

- Yes/. This PR is up to date with current the current state of 'develop'

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
@@ -71,7 +71,7 @@ bool SplineR2R<ST>::write_splines(hdf_archive& h5f)
   is very tall and skinny.
 */
 template<typename ST>
-void SplineR2R<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_copy)
+void SplineR2R<ST>::applyRotation(const ValueMatrix& rot_mat)
 {
   // SplineInst is a MultiBspline. See src/spline2/MultiBspline.hpp
   const auto spline_ptr = SplineInst->getSplinePtr();
--- a/src/QMCWaveFunctions/BsplineFactory/SplineR2R.h
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineR2R.h
@@ -95,7 +95,7 @@ public:
      [2] Toulouse & Umrigar, JCP 126, (2007)
      [3] Townsend et al., PRB 102, (2020)
   */
-  void applyRotation(const ValueMatrix& rot_mat, bool use_stored_copy) override;
+  void applyRotation(const ValueMatrix& rot_mat) override;
 
   inline void resizeStorage(size_t n, size_t nvals)
   {
--- a/src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
+++ b/src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
@@ -30,7 +30,12 @@ LCAOrbitalSet::LCAOrbitalSet(std::unique_ptr<basis_type>&& bs, bool optimize)
 }
 
 LCAOrbitalSet::LCAOrbitalSet(const LCAOrbitalSet& in)
-    : SPOSet(in), myBasisSet(in.myBasisSet->makeClone()), C(in.C), BasisSetSize(in.BasisSetSize), Identity(in.Identity)
+    : SPOSet(in),
+      myBasisSet(in.myBasisSet->makeClone()),
+      C(in.C),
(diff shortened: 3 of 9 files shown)


ye-luo:
I'm wondering if we can get ride of storeParamsBeforeRotation() if we do the following.
1. SPOSet::applyRotation is responsible to copy the original C once.
2. make c_original a shared_ptr. if SPOSet::applyRotation finds it empty, make a copy from C.
However this change the meaning of applyRotation, it applies rotation on the original coefficients instead of current coefficients.

PDoakORNL:
I agree the flag should be dropped.

But this is some nasty state, the only time C_copy seems to be used (i.e. use_stored_copy = true) is when resetParameters is called on a RotatedSPO. But only means anything for RotatedSPO sets constructed from an LCAOOribitalSet (I think).  

This is confusing overall, if we have higher level code that may want to revert to a SPOSet before rotation it makes sense for that level of the code to just make a copy of the original SPOSet, not for some SPOSet subtypes to maybe sometimes keep a copy of their "original" coefficients.

ye-luo:
@PDoakORNL ApplyRotation is valid for any SPOSet. LCAO has the MO coefficients, SplineSPO has spline coefficients to be rotated.

ye-luo (changes requested):
This change breaks the code. Let me write a bit more.

ye-luo:
original coefficients: from mean field
current coefficients: has been rotated by the past optimization steps
proposed coefficients: current optimization proposed new set of parameters, either accept or reject.

checkInParameters: storeParametersBeforeRotation stores current coefficients to c_copy;
resetParameters: apply the rotation between proposed coefficients and current coefficients.

What is proposed in this PR calls storeParametersBeforeRotation() at RotatedSPOs::buildOptVariables only once and c_copy stores the original coefficients. Then all the applyRotation except the first one is wrong.

I think the direction of this PR is right but more changes are needed to adjust the rotation matrix passed into applyRotation as the rotation from proposed coefficients to the original coefficients.

PDoakORNL:
I still feel like each rotated SPOSet is a different sposet.  An SPOSet should have just one set of coefficients. 

Handle them as separate objects at the scope that does the rotating not by having multiple sets of guts in one SPOSet.  applyRotation mutates the SPOSet, if you didn't make a "copy" of it you lost what it used to be.  If SPOSet is too high a level then have an SPOSetCoefficients abstraction and that is copied if the calling scope wants to keep state.

So not
```
spo_set.storeParametersBeforeRotation()
spo_set.applyRotation(some_rotation)
...
// don't like that rotation after all
spo_set.resetParameters()
```
Express the actual algorithm which is:
```
// You might not need one of these per SPOSet but just one per thread
auto original_sposet_coef{spo_set.getCoefficients()};
spo_set.applyRotation(some_rotation);
...
// for some reason start over
spo_set.setCoefficients(original_sposet_coef);
```

If your algorithm has three different sets of SPOSet coefficients they should be expressed in that scope not hidden in type that sometimes does and sometimes doesn't have multiple sets of state coefficients.  Here the spline spo sets don't appear to implement the state at all so the base class interface is misleading. I don't think you can ever get back your original or whatever C_copy is.  The variable names and confusion of type and variable naming conventions make these files painful to read.

PDoakORNL:
Also when you call applyRotation on an SPOSet ultimately backed by SplineR2R it does the rotation on the shared_ptr SplineInst which just seems like a terrible idea.

markdewing:
There needs to be more orbital rotation tests before changes to the code can be reliably considered.

PR #3917, opened 2022-03-25
Add chapter on performance portable implementation.

markdewing:
The batched drivers are a major change, so it seems worthwhile to highlight in its own chapter.

Having a separate chapter also displays the title in the table of contents, which will make it more visible to users.

I move the 'batched drivers' section whole.  My intention is that this PR is for discussing whether adding the chapter is a good idea, and what it should be named.   Adjusting the content in the chapter will come later.

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Documentation changes

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
laptop

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes This PR is up to date with current the current state of 'develop'
- N/A. Code added or changed in the PR has been clang-formatted
- N/A. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- Yes. Documentation has been added (if appropriate)

Issue #3793, opened 2022-02-01
WFOpt gradient_test method needs documentation.


ye-luo:
**Is your feature request related to a problem? Please describe.**
It is a useful feature as demonstrated by #3785 and worth some documentation.

In addition, aligning columns in the printout is much appreciated.

ye-luo:
Please also mention it is batched driver exclusive.

PR #3978, opened 2022-04-26
Add parameter gradient test to legacy driver

markdewing:
Add the gradient test (#3540 ) to the legacy linear optimizer.
When testing the optimizer and parameter gradients, it's helpful to be able to compare to the legacy driver to help isolate where problems occur.

- New feature

- No

## Checklist

- Not yet. Documentation has been added (if appropriate)


ye-luo (changes requested):
Please add documentation as #3793 requested.

markdewing:
I'd like #3917 to get in before I write more documentation.

prckent:
Happy to approve this once the conflict is fixed.

I assume one of the additional next steps will be to activate this code.

prckent (approved):
Note Ye's comment to add docs at some point.

ye-luo:
Please extend the existing test to cover the classic driver.

prckent:
Ping

markdewing:
Test this please

docs/methods.rst
```diff
@@ -1206,6 +1206,43 @@ the values found in the \*.sXXX.opt.h5 file. Be careful to keep the pair
...
+The gradients of the energy with respect to the variational parameters can be checked and optionally written to a file.
+The check compares the analytic derivatives with a finite difference approximation.
+These are activated by giving a ``gradient_test`` method in and ``optimize`` block, as follows:
+
+::
+
+     <qmc method="linear" move="pbyp">
+      <optimize method="gradient_test">
+      </optimize>
+      ... rest of optimizer input ...
```
ye-luo, line 1220: `rest of optimizer input`->`VMC related input`?
With gradient_test, I doubt any other optimizer parameters are still needed.

tests/molecules/He_param/CMakeLists.txt
```diff
@@ -25,6 +25,25 @@ if(NOT QMC_CUDA)
...
+  # Check the parameter gradient code for the legacy driver
+  qmc_run_and_check_custom_scalar(
+    BASE_NAME
+    He_param_grad_legacy
+    BASE_DIR
+    "${qmcpack_SOURCE_DIR}/tests/molecules/He_param"
+    PREFIX
+    He_param_grad.param
+    INPUT_FILE
+    He_param_grad.xml
```
ye-luo, line 37: Incorrect input file. This explains the coverage report complaints.
Can we rename the input file He_param_grad_legacy_driver.xml

markdewing:
I'm not sure it's worth the effort to extend the deterministic parameter gradient test to the legacy driver.

docs/methods.rst
markdewing: This is probably true, but I expect the common case is starting from an existing optimizer input block.

ye-luo (approved):

ye-luo:
Test this please

markdewing:
My original motivation for adding this to the legacy driver is to compare wavefunction and gradient values from the batched driver and the legacy driver to see where they differ. A deterministic test is a regression test for a particular wavefunction - that sort of testing seems useful for ensuring changes to the driver don't change the code unexpectedly.  And not as useful for investigating existing wavefunction code.

prckent (commented):
When does this test get run?

It looks to be a "short" test and not a "deterministic" test. Specific prefixes need to be used due to using things like -R short or -E long in the various nightly scripts.

How long does it take? What are the expected statistics?

markdewing:
The test that was added is a statistical test, not a deterministic test.   It is a copy of the existing statistical test for the parameter gradients for using the batched driver (in molecules/He_param). The values of the gradients should be the same.

There is a deterministic parameter gradient test using the batched driver for a different system (solids/diamondC_1x1x1_pp).  Copying and modifying that test for the legacy driver is what I pushed back on.

ye-luo:
Indeed, `tests/solids/diamondC_1x1x1_pp/det_qmc_param_grad.xml` was the deterministic version.
He_param was actually a bit off in the nightly. https://cdash.qmcpack.org/CDash/queryTests.php?project=QMCPACK&date=2022-05-06&filtercount=1&showfilters=1&field1=testname&compare1=63&value1=param_grad
I consider the deterministic one is more valuable than the statistically one. The statistically one is impossible to run with offload to the very slow finite difference.

markdewing:
I prefer PR's to be focused.  In this case the added test is to just enough to test the additions in this PR.  Since the added functionality is for testing, this makes the difference between functionality and tests a bit more subtle.  In my view, deterministic tests use this PR to test other functionality, but are not as good at testing this PR.   
Adding the deterministic test should be a separate PR.

ye-luo:
We may or or may not use this feature to create deterministic tests. I saw that a orthogonal topic.

This "gradient_test" is a QMCPACK feature rather than a test managed by catch. In my view, It is better to be protected by a deterministic test before being used for testing other things.

Right now the feature has been protected by a statistical test although inferior to a deterministic test. So this PR can get in but I will leave others to decide if a deterministic test is a strong demand in this PR or not.

ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
A deterministic test has been added.

Issue #3633, opened 2021-11-24
Scaling of optimizer with number of parameters


markdewing:
We would like the linear optimizer to scale to a large number of parameters. Generally the more parameters there are in the optimized wavefunction, the more accurate the wavefunction. This issue records some investigation into memory and time scaling with respect to the number of parameters.

Sources of memory usage (Np - number of parameters, Ns - number of samples)
1. Parameters and associated info in VariableSet  - O(Np)
2. Values of psi, E, and derivatives for each sample - O(Np * Ns)
3. Overlap and Hamiltonian matrices used in solving linear method eigenproblem - O(Np * Np)

The source of memory usage is more detail

**1. Parameters in VariableSet**
Some discussion in #3625 

There are often multiple copies of VariableSet, one at the each wavefunction component level.  At the top level of the optimizer, there are two copies, for dealing with constraints.

I instrumented the code, and ran on a hydrocarbon molecule with 1-body and 2-body Jastrows, and CI coefficients.  The number of coefficients was varied to get the range of total variational parameters.  There were 26 instantiations of VariableSet, most of them zero-sized or small (Jastrows). There were 4 large instantiations (CI coeff).   The space usage is about 124 bytes/parameter for an individual variable set, and 500 bytes/parameter considering all the variable sets.

(Technical aside: Many the data structures in VariableSet use the parameter name in a std::string.  The parameter names were all short enough (<16 characters) to use the short-string optimization. The short string fits in the std::string itself, and there is no additional dynamic allocation.  If the names were longer than 15 characters, we would need to measure the allocated string space as well.)

**2. Sample related storage**
I have run into this memory limit with small systems (fast to generate samples) and large sample numbers (trying to be accurate for testing).

There are two variables in QMCCostFunctionBatched of size Ns x Np: `DerivRecords` and `HDerivRecords`

The solution will be to collect the accumulated quantities during the run.  The eases the memory pressures (would be O(Np)) at the tradeoff of not being able to do correlated sampling.

**3. Overlap and Hamiltonian matrices**
In `one_shift_run` there are four matrices of size Np x Np: `ovlMat`, `hamMat`, `invMat` and `prdMat`

Just to make some concrete numbers for a sense of scale, assume the number of samples is 10000 and vary the number of parameters

Np  | VariableSet | sample-related storage | linear method matrices
----|----|----|----
10k | .005 GB |  1.6 GB | 3.2 GB
50k | .025 GB | 8.0 GB | 80 GB
[...]

PR #3981, opened 2022-04-27
Refactor optimization classes

ye-luo:
## Proposed changes
Trying to reduce code duplication between batched and classic drivers and eliminate dead routines.
I feel QMCFixedSampleLinearOptimize and QMCFixedSampleLinearOptimizeBatched are still overweight and need more thoughts and experiments to refine their underlying abstraction layers.

- No

epyc-server

## Checklist


src/QMCWaveFunctions/VariableSet.h
```diff
@@ -171,11 +171,11 @@ struct VariableSet
     }
   }
 
-  inline void getParameterTypeList(std::vector<int>& types)
+  inline void getParameterTypeList(std::vector<int>& types) const
   {
-    std::vector<index_pair_type>::iterator PTit(ParameterType.begin()), PTend(ParameterType.end());
+    auto PTit(ParameterType.begin()), PTend(ParameterType.end());
```
PDoakORNL, line 176: This file could really use an update to follow the naming conventions.
PTit, ParameterType, NameAndValue are actually all variables
pair_type, variable_type, etc are types.
It's painful to read.

src/Numerics/LinearFit.h
```diff
@@ -29,11 +30,8 @@ inline void LinearFit(std::vector<T>& y, Matrix<T>& A, std::vector<T>& coefs)
   int N = A.size(0);
   int M = A.size(1);
   if (y.size() != N)
-  {
-    app_error() << "Differernt number of rows in basis functions that in data "
-                << "points in LinearFit.  Exitting.\n";
-    abort();
-  }
+    throw std::runtime_error("Differernt number of rows in basis functions that in data "
+                "points in LinearFit.  Exitting.\n");
```
PDoakORNL, line 34: Exitting [sic] and \n are unecessary.

PDoakORNL (approved):
I'd like to see what the test coverage change here is, and have time to see if there really is any at the unit level.

But overall this looks good.

src/QMCWaveFunctions/VariableSet.h
markdewing: The VariableSet class ought to be replaced, or at least needs significant changes, as discussed in #3633 .   Updating the naming conventions as a separate step might not be worth the effort.

src/QMCDrivers/WFOpt/LinearMethod.cpp
```diff
@@ -0,0 +1,171 @@
...
+    }
+  }
+  //     returns the number of non-linear parameters.
+  //    app_log()<<"line params: "<<first<<" "<<last<< std::endl;
+}
+
+LinearMethod::Real LinearMethod::getNonLinearRescale(std::vector<Real>& dP,
+                                                     Matrix<Real>& S,
+                                                     const QMCCostFunctionBase& optTarget) const
+{
```
markdewing, line 153: Passing a "heavy" or "large" class like QMCCostFunctionBase to a function like this caught my attention.  (Maybe a better description for QMCCostFunctionBase would be that it is higher up in a logical hierarchy of concerns.) 
An immediate question is maybe getNonLinearRange should be a member function of QMCCostFunctionBase? (I think the answer is: it should not be, because this function relates to the linear method, not the cost function)

It would be better to pass just what is needed - in this case only the vector of parameter types used in getNonLinearRange.
(Passing the entire class is reasonable as an intermediate step in refactoring the code).

ye-luo:
Will take out the CUDA change.

prckent:
Thanks Mark. Without your comment this change would indeed have been missing from our changelogs.

ye-luo:
Test this please

src/Numerics/LinearFit.h
ye-luo: Just removed "Exitting.\n"

src/QMCDrivers/WFOpt/LinearMethod.cpp
ye-luo: legacy CUDA cod removal has been taken out of this PR.

ye-luo:
Test this please

prckent (approved):
LGTM

src/QMCDrivers/WFOpt/LinearMethod.h
```diff
@@ -0,0 +1,47 @@
...
+  // obtain the range of non-linear parameters
+  void getNonLinearRange(int& first, int& last, const QMCCostFunctionBase& optTarget) const;
+
+public:
+  //asymmetric generalized EV
+  Real getLowestEigenvector(Matrix<Real>& A, Matrix<Real>& B, std::vector<Real>& ev) const;
+  //asymmetric EV
+  Real getLowestEigenvector(Matrix<Real>& A, std::vector<Real>& ev) const;
+  // compute a rescale factor. Ye: Where is the method from?
+  Real getNonLinearRescale(std::vector<Real>& dP, Matrix<Real>& S, const QMCCostFunctionBase& optTarget) const;
```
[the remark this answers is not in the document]
ye-luo: Yes. I thought about that but didn't plan to do much for now. I was planning to do more design experiment in LinearMethod for example, make start()/finalize()/check_configurations() pure virtual.

PR #3985, opened 2022-04-27
Add a batched driver option to measure imbalance

ye-luo:
## Proposed changes
Add a batched driver option for measuring imbalance in a few selected locations.

- New feature

- No

Summit

## Checklist


markdewing:
Do you have any example output?

src/QMCDrivers/QMCDriverNew.cpp
```diff
@@ -570,4 +575,28 @@ void QMCDriverNew::checkLogAndGL(Crowd& crowd, const std::string_view location)
...
+  myComm->barrier();
+  std::vector<double> my_barrier_time(1, only_this_barrier.elapsed());
+  std::vector<double> barrier_time_all_ranks(myComm->size(), 0.0);
+  myComm->gather(my_barrier_time, barrier_time_all_ranks, 0);
+  if (!myComm->rank())
+  {
+    auto const count  = static_cast<double>(barrier_time_all_ranks.size());
+    const auto max_it = std::max_element(barrier_time_all_ranks.begin(), barrier_time_all_ranks.end());
+    const auto min_it = std::min_element(barrier_time_all_ranks.begin(), barrier_time_all_ranks.end());
+    app_log() << std::endl
```
PDoakORNL, line 591: Why so many std::endl?  I see the semantics of that as << '\n', and flush the stream.  Certainly there is no need to flush this output more than once.

ye-luo:
```
Warmup imbalance (slow ranks wait less):
    average wait seconds = 1.4131566310e+01
    min wait at rank 19924, seconds = 1.2588500977e-04
    max wait at rank 16959, seconds = 1.9752846956e+01

Block 0 imbalance (slow ranks wait less):
    average wait seconds = 1.3349481820e+00
    min wait at rank 23753, seconds = 8.6069107056e-05
    max wait at rank 16166, seconds = 1.4375920296e+00

Block 1 imbalance (slow ranks wait less):
    average wait seconds = 1.3463662845e+00
    min wait at rank 19878, seconds = 7.2002410889e-05
    max wait at rank 3509, seconds = 1.4509370327e+00
```

PDoakORNL (approved):
This seems quite useful in moderation.

As it is this doesn't seem like it should be too much a performance hit, but it could be. Usage will need a close watch to make sure it doesn't become something that should be only in developer builds.

Its important that which rank is max and min is reported. Rank 0 would be expected to be late if between this barrier the previous explicit or implicit MPI barrier there was io, logging, or an asymmetric collective. This might make the measurement not very useful without an additional barrier but at least that could be easily seen.  

Is it naive to think the min value just gives the overhead of having the barrier and measurement itself so it would be better to see the average of difference between the rank wait times and the minimum wait?

src/QMCDrivers/QMCDriverNew.cpp
ye-luo: It should not matter much. Clang-format doesn't work well with '\n'.

ye-luo:
Test this please

prckent:
Required:

1. Please make the messages more verbose. Imbalance of _what_? Due to _what_? The messages should mention "MPI wait time" or "MPI synchronization".

2. Modify an input file to hit this code. It would be nice to check for the output in the test as well, but at least if the code paths are active we'll be able to catch massive breakage in the nightlies & CI.

ye-luo:
@prckent  Will change from
```
Block 0 imbalance (slow ranks wait less):
```
to
```
Block 0 MPI imbalance measured by an additional barrier (slow ranks wait less):
```
> Modify an input file to hit this code. t would be nice to check for the output in the test as well, but at least if the code paths are active we'll be able to catch massive breakage in the nightlies & CI.

Do you have a particular test in mind? I don't get "catch massive breakage" part in your comment. What type of breakage in the nightlies & CI?

ye-luo:
I thought you meant using this feature to catch breakage unrelated to this feature. I will look for an appropriate test to add this.

prckent:
Hmm. Catching a large imbalance would be good "smell" in the nightlies, but I think there are higher priority items. e.g. We still have lots of code with no coverage in CI and plenty of code in need of cleaning. Better use of our time to work on those.

PR #3990, opened 2022-04-28
Update meansure_imbalance printout and add a test.

ye-luo:
## Proposed changes

Address post-merge review request of #3985 from @prckent 

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCDrivers/QMCDriverNew.cpp
+++ b/src/QMCDrivers/QMCDriverNew.cpp
@@ -589,7 +589,7 @@ void QMCDriverNew::measureImbalance(const std::string& tag) const
     const auto max_it = std::max_element(barrier_time_all_ranks.begin(), barrier_time_all_ranks.end());
     const auto min_it = std::min_element(barrier_time_all_ranks.begin(), barrier_time_all_ranks.end());
     app_log() << std::endl
-              << tag << " imbalance (slow ranks wait less):" << std::endl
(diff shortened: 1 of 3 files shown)


prckent (approved):
Thanks Ye.

PDoakORNL (commented):
The test should probably get compiled into drivers_mpi so it is tested with multiple ranks.
The looks it needs a new file but QMCDriverNewTestWrapper can be reused.

ye-luo:
> The test should probably get compiled into drivers_mpi so it is tested with multiple ranks. The looks it needs a new file but QMCDriverNewTestWrapper can be reused.

A quick try shows the need of a separate file.  test_QMCDriverNew.cpp is only for 1 rank. I will make a test_QMCDriverNew_3r.cpp

ye-luo:
@PDoakORNL Actually there is no need of extra file. Need to adjust the parsed input file "total_walkers" to "walkers_per_rank".

ye-luo:
Test this please

prckent:
I changed the "input" to have walkers_per_rank not walker_per_rank. Why did not the non-plural version work?

ye-luo:
> I changed the "input" to have walkers_per_rank not walker_per_rank. Why did not the non-plural version work?

The default value is 1 when running with 1 thread.

ye-luo:
Test this please

prckent:
OK, so another example where catching non-parsed input parameters are giving an error would help stop the user have the code do what is not intended.

PR #3997, opened 2022-05-03
Require cmake 3.17 and improve SYCL compiler discovery

ye-luo:
## Proposed changes
3.17 was released in March 20, 2020
The main reason to raise CMake minimum version is for policy [0099](https://cmake.org/cmake/help/v3.17/policy/CMP0099.html). Compiler driver needs to handle additional device linking steps and need INTERFACE_LINK_OPTIONS being properly propagated.

- Build related changes

- No

JLSE

## Checklist


ye-luo:
Test this please

CMakeLists.txt
```diff
@@ -1,15 +1,10 @@
...
 # CMake version and policies
 ######################################################################
-cmake_minimum_required(VERSION 3.15.0)
-
-# CMP0074: CMake find_package will use <PackageName>_ROOT CMake variable
-# and environment variable in search path.
-# various Find<Package>.cmake modules may not follow this policy
-# CMP0075: check_include_file_cxx and other follow CMAKE_REQUIRED_LIBRARIES
-# when doing link check
-cmake_policy(SET CMP0074 NEW)
```
PDoakORNL, line 11: I don't believe this can be dropped.  My reading of this is as of 3.23 if the policy is not set you get a warning and OLD not NEW.
Also seems true of CMP0075

PDoakORNL (changes requested):
Certainly I would argue we should keep the CMP0074 yes behavior.

CMakeLists.txt
ye-luo: See the added comment right after ` cmake_minimum_required(VERSION 3.17.0) `
CMP0075 is known to your 3.23 and introduced before 3.17. Thus NEW is set.

PDoakORNL (approved):

ye-luo:
Test this please

PR #4004, opened 2022-05-05
Fix DelayedUpdateCUDA::initializeInv H2D must be synchronized

ye-luo:
## Proposed changes
Saw this bug when reading the code. The old assumption is only valid for D2H not H2D.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
+++ b/src/QMCWaveFunctions/Fermion/DelayedUpdateCUDA.h
@@ -135,7 +135,8 @@ public:
                                    hstream),
                    "cudaMemcpyAsync failed!");
     clearDelayCount();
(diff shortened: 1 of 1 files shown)


ye-luo:
Test this please

PDoakORNL (approved):
I think counting on the behavior of async with unpinned memory causing an implicit sync isn't good in either direction. 

The semantics of the async transfer are that the dependent async operations must be given the stream. If the next operation is going to be an omp target operation or possibly on another stream then an explicit stream synchronize is obligatory.  Still it would be better to use the omp interop pragma or some other trick to make this sync unecessary.

ye-luo:
InitializeInv is only used in performance non-critical places. So just putting it on the safe side. In places sensitive to async, we will need to handle the sync better.

PR #4009, opened 2022-05-09
Workaroud OneAPI compiler bug.

ye-luo:
## Proposed changes
Add a workaround. Standalone reproducer is https://github.com/ye-luo/openmp-target/blob/master/tests/math/modf_team.cpp.
The current symptom is hang. So avoid adding the reproducer here.

- Other (please describe): bug workaround

- No

epyc-server

## Checklist

### DIFF
--- a/src/spline2/MultiBsplineEval_helper.hpp
+++ b/src/spline2/MultiBsplineEval_helper.hpp
@@ -52,8 +52,13 @@ inline void getSplineBound(T x, TRESIDUAL& dx, int& ind, int nmax)
   }
   else
   {
+#if defined(__INTEL_LLVM_COMPILER) || defined(__INTEL_CLANG_COMPILER)
+    T ipart = std::floor(x);
+    dx = x - ipart;
+#else
     T ipart;
     dx  = std::modf(x, &ipart);
+#endif
     ind = static_cast<int>(ipart);
     // upper bound
     if (ind > nmax)


correaa:
Is `defined(__INTEL_LLVM_COMPILER) || defined(__INTEL_CLANG_COMPILER)` equivalent to `defined(__INTEL_COMPILER)` ?

src/spline2/MultiBsplineEval_helper.hpp
```diff
@@ -52,8 +52,13 @@ inline void getSplineBound(T x, TRESIDUAL& dx, int& ind, int nmax)
   }
   else
   {
+#if defined(__INTEL_LLVM_COMPILER) || defined(__INTEL_CLANG_COMPILER)
```
PDoakORNL, line 55: Wish that there was some way we could make sure we remove this when the supported intel compilers no longer have broken std::modf.

ye-luo:
`__INTEL_COMPILER` is the classic compiler.

PDoakORNL (commented):
Seems fine to me, is this just a race condition surfaced by openmp?

src/spline2/MultiBsplineEval_helper.hpp
correaa: I guess one can add a check the intel compiler version until the broken feature is solved

correaa:
> `__INTEL_COMPILER` is the classic compiler.

Good to know. I am also working around some defects of intel and with this I can be more fine grained.
I still wonder what is the difference between `_CLANG_` and `_LLVM_` macros. 

So CLANG/LLVM detected `icpx`?

ye-luo:
> 



> > `__INTEL_COMPILER` is the classic compiler.
> 
> Good to know. I am also working around some defects of intel and with this I can be more fine grained. I still wonder what is the difference between `_CLANG_` and `_LLVM_` macros.
> 
> So CLANG/LLVM detected `icpx`?

If you are asking about CMake. The llvm based compiler is recognized as clang for cmake <3.20 or intelllvm >=3.20.

correaa:
> If you are asking about CMake. The llvm based compiler is recognized as clang for cmake <3.20 or intelllvm >=3.20.

No, I was asking about the code macros. But that is also good to know. Thanks.

src/spline2/MultiBsplineEval_helper.hpp
ye-luo: @PDoakORNL since the difference between these two code paths are so small. I doubt it is worth our time checking if intel fixed it from time to time. More likely we will only investigate the issue once the workaround stops working.

ye-luo:
Test this please

prckent (approved):
Necessary but unfortunate.

Issue #3941, opened 2022-04-10
Preliminary assessment of static-analysis warnings (clang-tidy)


correaa:
**Describe the bug**

After a brief conversation on Slack I ran a blanket static analysis check on the whole source with clang-tidy.

I found about <s>2'294'900 (yes, 2 x 10^6)</s> 242'406 unique warnings. (attached file)

This is an absolute upper bound, it is possible that only one order of magnitude less warnings actually deserve any attention.

From my experience from other code:
I estimate that about half can be automatically fix by clang-tidy (yes, the tool offers automatic solutions by doing text replacements in the source code. In my experience the tools offers fixes only when it is safe.)

Of the remaining, maybe half could be considered false positives or style related.

Maybe a bunch will solve unknown bugs or undefined behavior.

In my opinion, almost all warnings (~99%) have a positive impact if fixed but there is a decision to make to which I what point to stop. (not counting the checks that should be disable IMO, listed below).

**To Reproduce**

1. clone repo
2.

```bash
cd build
export OMPI_CXX=clang++
export OMPI_CC=clang
cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DBUILD_AFQMC=1 -DQMC_MIXED_PRECISION=1 -DCMAKE_BUILD_TYPE=Debug -DMPIEXEC_PREFLAGS="--allow-run-as-root;--bind-to;none" -DCMAKE_CXX_CLANG_TIDY="clang-tidy;-checks=*"  ..
make 2>&1 | grep warning | tee clang-tidy-warnings.txt
```
3. Sit tight

4. Inspect `sort clang-tidy-warnings.txt | uniq` (file attached)
[clang-tidy-warns.zip](https://github.com/QMCPACK/qmcpack/files/8458297/clang-tidy-warns.zip)

**Expected behavior**

No warnings

**System:**
 - domestic Ubuntu 21.10 compiling with clang.

**Additional context**

There is no reason to think that any of the warnings are critical.
To begin I think that this checks are for the most part useless and misleading.

(`.clang-tidy`)
```
Checks: '*,
  -readability-magic-numbers,
  -modernize-use-nodiscard,
  -altera-id-dependent-backward-branch,
  -altera-struct-pack-align,
  -altera-unroll-loops,
  -cert-err58-cpp,
  -cppcoreguidelines-avoid-non-const-global-variables,
  -cppcoreguidelines-macro-usage,
[...]

PR #4021, opened 2022-05-17
Manually triggered GitHub Action for static checks using clang-tidy

williamfgc:
## Proposed changes

Adds a `static` manual workflow on GitHub Actions, triggered from the Actions tab. 
Get results from compiling with clang-tidy (takes 2 hours 15 min) on-demand on the develop branch only.
Update docker image to add clang-tidy dependency.
Related to #3941 as we address clang-tidy recommendations.
Needs merging so an admin can trigger the workflow.

- Testing changes (e.g. new unit/integration/performance tests) static checks

- Yes

Runs on GitHub Actions on my fork, must run on QMCPACK/qmcpack

## Checklist


williamfgc:
For reference this is the resulting workflow logs on my [fork](https://github.com/williamfgc/qmcpack/actions/runs/2339592357).

correaa:
The PR looks good to me. 
Depending on the number of compilation units, `"clang-tidy;-checks=*"` might produce very large logs to the point that github could truncate them.

There will be a flood of warnings at the beginning so we might want to define a protocol on how to deal with them, for example if it is decided that nothing will be done about a certain warning (for example, disable the warning globally vs. disabling check in that line `  // NOLINT` vs disabling specific check `  // NOLINT(specific-warning)` vs adding a useful comment ` // NOLINT(specific-warning) comment: implicit conversion is necessary here because bla...`.

I think we can safely disable this three warnings to start with because they don't make sense.

```
  -llvmlibc-callee-namespace,
  -llvmlibc-implementation-in-namespace,
  -llvmlibc-restrict-system-libc-headers,
```

Finally, as exceptions mount, it is better to have `.clang-tidy` configuration file that can sit in the root directory. For example: https://gitlab.com/correaa/boost-multi/-/blob/master/.clang-tidy

correaa (approved):
Look good to me. See comments in the PR thread about possible future issues.

ye-luo:
Need some documentation. I mean if I see a CI failure, how to reproduce clang-tidy manually.

correaa:
> Need some documentation. I mean if I see a CI failure, how to reproduce clang-tidy manually.



> Need some documentation. I mean if I see a CI failure, how to reproduce clang-tidy manually.

yes, even for reproducibility sake, as soon as the command line becomes more complicated that `-DCMAKE_CXX_CLANG_TIDY="clang-tidy;-checks=*"`, we will start needing a configuration file `.clang-tidy` that is part of the repository.

markdewing:
There already is a .clang-tidy file in the src/ directory
https://github.com/QMCPACK/qmcpack/blob/develop/src/.clang-tidy

correaa:
> There already is a .clang-tidy file in the src/ directory https://github.com/QMCPACK/qmcpack/blob/develop/src/.clang-tidy

ok, in that case the command in the CI can be simply `-DCMAKE_CXX_CLANG_TIDY="clang-tidy"`.

prckent (changes requested):
We should be using a current version of Clang right from the beginning. e.g. 13, since 14 is a bit too recent to be available everywhere. LLVM10 will be out of our 2 year support window in 4 months.

Given that we have so many warnings at the moment, have you found a convenient way to review them in the GitHub interface? My experience with the current CI logs is not good for any lengthy file.

williamfgc:
Changes:
- Added docs @ye-luo 
- Added Ubuntu22 image with Clang14 (default) for a recent clang-tidy @prckent 
- Added `.clang-tidy` with `readability-*` and `bugprone-*` as seen in #3941. @correaa

docs/github_actions.rst
```diff
@@ -945,3 +945,30 @@ Generate code coverage reports once all tests have reported.
...
+
+To run the workflow:
+- Go to the Actions tab
+- Click on the `static` workflow on the left
+- Click on `Run workflow` on the right
+- Use workflow from `Branch:develop` and click on the `Run workflow` button
+
+**Note:** the current `.clang-tidy` configuration file is compatible with clang v14 and runs on the `williamfgc/qmcpack-ci:ubuntu22-openmpi` docker image. To run locally on a Linux system use: `docker run -it user williamfgc/qmcpack-ci:ubuntu22-openmpi /bin/bash` or refer to the :ref:`running_docker`` section.
+
+To build locally enaling `clang-tidy`` static checks defined in `qmcpack/.clang-tidy` use the CMake `-DCMAKE_CXX_CLANG_TIDY` option as follows:
```
ye-luo, line 964: `src/.clang-tidy` instead of `qmcpack/.clang-tidy`

williamfgc: @ye-luo good point. @PDoakORNL @quantumsteve I'm unifying the `.clang-tidy` configuration. Are you OK removing the one under `/src`? The one at the top level would have similar checks.

ye-luo: We should have only one configure file. I don't get why it should be under the root directory. If I understand correctly, clang-tidy handles C++ source code and all C++ code managed by QMCPACK is under src. Codes under external should not be subject to clang-tidy checks by QMCPACK.

docs/github_actions.rst
```diff
@@ -945,3 +945,30 @@ Generate code coverage reports once all tests have reported.
...
+
+To run the workflow:
+- Go to the Actions tab
+- Click on the `static` workflow on the left
+- Click on `Run workflow` on the right
+- Use workflow from `Branch:develop` and click on the `Run workflow` button
+
+**Note:** the current `.clang-tidy` configuration file is compatible with clang v14 and runs on the `williamfgc/qmcpack-ci:ubuntu22-openmpi` docker image. To run locally on a Linux system use: `docker run -it user williamfgc/qmcpack-ci:ubuntu22-openmpi /bin/bash` or refer to the :ref:`running_docker`` section.
+
+To build locally enaling `clang-tidy`` static checks defined in `qmcpack/src/clang-tidy` use the CMake `-DCMAKE_CXX_CLANG_TIDY` option as follows:
```
ye-luo, line 964: `qmcpack/src/.clang-tidy` missing a dot.
"enaling" typo?

PDoakORNL:
Our .clang-tidy should be in src it should not be a reachable parent for the external_codes.  Additionally be aware that subdirectories can have there own .clang-tidy which may optionally inherit from its parent or not.  For instance in src/type_traits if am experimenting with the following 
```yaml
InheritParentConfig: true
CheckOptions: [ { key: readability-identifier-naming.StructIgnoredRegexp, value: '.*(_t|_impl)$'}
                ]
```
since its rather helpful to reading template metaprogramming to have types with _t and _impl and no have a ton of //NOLINT comments.

Some of the lost cause directories could also be squelched from the clang-tidy reporting by adding `InheritParentConfig: false` .clang-tidy in their directory.

correaa:
> since its rather helpful to reading template metaprogramming to have types with _t and _impl and no have a ton of //NOLINT comments.

I use that idiom in my naming and yet I didn't have to put anything special to the clangtidy.


> Some of the lost cause directories could also be squelched from the clang-tidy reporting by adding `InheritParentConfig: false` in .clang-tidy local to those directories.  

I didn't know about this option, seems useful (but not for external code) but for directories, (like tests).

PDoakORNL:
@correaa in our case we've defined type naming rules that disallow that for types.  They're specified as JustMixedCase.

williamfgc:
Thank you all for your input. I think I addressed the review comments. Let me know if there anything pending. We can all refer to the manually generated report after merge to start addressing issues, in particular those of `bugprone-*` type.

docs/github_actions.rst
```diff
@@ -945,3 +945,30 @@ Generate code coverage reports once all tests have reported.
 Upload Coverage
 ---------------
 Upload the generated code coverage to `CodeCov <https://codecov.io/gh/QMCPACK/qmcpack/tree/develop/src>`_ where the badges on our repo will then be updated. Only done by jobs with name `*-coverage`.
+
+
+====== ======== ======== 
+Static Analysis Workflow
+====== ======== ========
```
ye-luo, line 952: I don't see breaking lines based on spaces.

docs/github_actions.rst
```diff
@@ -945,3 +945,30 @@ Generate code coverage reports once all tests have reported.
...
+
+To run the workflow:
+- Go to the Actions tab
+- Click on the `static` workflow on the left
+- Click on `Run workflow` on the right
+- Use workflow from `Branch:develop` and click on the `Run workflow` button
+
+**Note:** the current `.clang-tidy` configuration file is compatible with clang v14 and runs on the `williamfgc/qmcpack-ci:ubuntu22-openmpi` docker image. To run locally on a Linux system use: `docker run -it user williamfgc/qmcpack-ci:ubuntu22-openmpi /bin/bash` or refer to the :ref:`running_docker`` section.
+
+To build locally enaling `clang-tidy`` static checks defined in `qmcpack/src/.clang-tidy` use the CMake `-DCMAKE_CXX_CLANG_TIDY` option as follows:
```
ye-luo, line 964: `enaling` is still there.

docs/github_actions.rst
```diff
@@ -945,3 +945,30 @@ Generate code coverage reports once all tests have reported.
 Upload Coverage
 ---------------
 Upload the generated code coverage to `CodeCov <https://codecov.io/gh/QMCPACK/qmcpack/tree/develop/src>`_ where the badges on our repo will then be updated. Only done by jobs with name `*-coverage`.
+
+
+====== ======== ======== 
+Static Analysis Workflow
+====== ======== ========
+
+A manually triggered workflow on the GitHub Actions tab can generate the required checks using the [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) static analyzer. The current approach is to set checks in the `qmcpack/.clang-tidy` configuration file and run using `clang-tidy` v14 on GitHub Actions runners. The workflow is not part of CI, and it's currently used for reporting potential warnings on the GitHub Actions logs as they are addressed on the `develop` branch as part of refactoring efforts for code quality.
```
ye-luo, line 954: `qmcpack/.clang-tidy` still

ye-luo:
Test this please

ye-luo (dismissed):

prckent (approved):

prckent:
Approved assuming that I didn't break the rst.

prckent:
Test this please

prckent (approved):

correaa:
ok, i see. i didn’t know about JustMixedCase. what is the difference with
CamelCase?

Issue #3955, opened 2022-04-19
Intel 19.1.x compiler shows a frontend crash with gcc 9.3 libstdc++ header files


ye-luo:
**Describe the bug**
```
/home/yeluo/opt/qmcpack/src/Estimators/tests/test_EstimatorManagerInput.cpp(175): internal error: assertion failed at: "shared/cfe/edgcpfe/lower_name.c", line 9641
```
A workaround is to use gcc 8.

@prckent Sulfur is failing for this reason now.
https://cdash.qmcpack.org/CDash/buildSummary.php?buildid=306741

ye-luo:
CI with the same Intel compiler on sulfur seems OK. https://github.com/QMCPACK/qmcpack/runs/6217697106?check_suite_focus=true

PR #4024, opened 2022-05-20
Workaround intel 19.1.x compiler issue

ye-luo:
## Proposed changes
Expect to fix https://github.com/QMCPACK/qmcpack/issues/3955.
Intel compiler was choked by `std::vector::insert(std::variant<>)`

- Bugfix

- No

Theta

## Checklist

### DIFF
--- a/src/Estimators/EstimatorManagerInput.cpp
+++ b/src/Estimators/EstimatorManagerInput.cpp
@@ -34,13 +34,14 @@ EstimatorManagerInput::EstimatorManagerInput(std::initializer_list<EstimatorMana
 
   est_offset        = 0;
   scalar_est_offset = 0;
-  std::for_each(emil.begin(), emil.end(), [&](const EstimatorManagerInput& emi) {
+  for (const EstimatorManagerInput& emi : emil)
+  {
     std::copy(emi.estimator_inputs_.begin(), emi.estimator_inputs_.end(), estimator_inputs_.begin() + est_offset);
     est_offset += emi.estimator_inputs_.size();
     std::copy(emi.scalar_estimator_inputs_.begin(), emi.scalar_estimator_inputs_.end(),
               scalar_estimator_inputs_.begin() + scalar_est_offset);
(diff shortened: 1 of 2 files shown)


ye-luo:
Test this please

src/Estimators/EstimatorManagerInput.cpp
```diff
@@ -22,11 +22,28 @@ EstimatorManagerInput::EstimatorManagerInput(xmlNodePtr cur) { readXML(cur); }
 
 EstimatorManagerInput::EstimatorManagerInput(std::initializer_list<EstimatorManagerInput> emil)
 {
-  std::for_each(emil.begin(), emil.end(), [this](const EstimatorManagerInput& emi) {
-    std::copy(emi.estimator_inputs_.begin(), emi.estimator_inputs_.end(), std::back_inserter(estimator_inputs_));
+  // \todo the following code fusing two vectors can be written in more consice way with std::back_inserter. See history.
+  // Right now needs to use a clumsy way to make intel classic compiler 19.1.x happy when using gcc 9 on some OS distros.
```
ye-luo, line 26: @PDoakORNL Added TODO note here.
Q: does std::back_inserter potentially over allocate memory?
I'm not saying memory is a concern here. Just to understand how back_inserter works.

ye-luo:
Test this please

src/Estimators/EstimatorManagerInput.cpp
prckent: If there are concerns we can put an #ifdef for this particularly compiler, just as we have in the past. There are still alot of Intel 19.x compilers in use, so if we can make things easier for users easily we should do so.

PDoakORNL: Considering the workaround is localized the think the \todo is less trouble.

PDoakORNL: And back_inserter works like the containers push_back.

PDoakORNL (approved):
So I guess this works in later intel + stdlib combinations?

PR #4025, opened 2022-05-20
Reduce use of variadic templates

ye-luo:
## Proposed changes
Seems not worth adding another layer of calls.

- No

epyc-server

## Checklist

### DIFF
--- a/src/Estimators/EstimatorManagerInput.cpp
+++ b/src/Estimators/EstimatorManagerInput.cpp
@@ -54,7 +54,7 @@ void EstimatorManagerInput::readXML(xmlNodePtr cur)
   if (cur_name == "estimators")
     child = cur->xmlChildrenNode;
   else
-    child = cur;
+    child = cur; // the case when 'estimator's are not encapsulated by a 'estimators' node
   while (child != NULL)
   {
     std::string cname{lowerCase(castXMLCharToChar(child->name))};
@@ -67,23 +67,23 @@ void EstimatorManagerInput::readXML(xmlNodePtr cur)
       if (aname.empty() && !atype.empty())
         aname = atype;
       if (atype == "localenergy" || atype == "elocal")
-        scalar_estimator_inputs_.emplace_back(std::in_place_type<LocalEnergyInput>, child);
+        appendScalarEstimatorInput<LocalEnergyInput>(child);
       else if (atype == "cslocalenergy")
       {
-        scalar_estimator_inputs_.emplace_back(std::in_place_type<CSLocalEnergyInput>, child);
+        appendScalarEstimatorInput<CSLocalEnergyInput>(child);
         app_warning() << "CSLocalEnergyEstimator support is at best experimental with batch drivers" << std::endl;
       }
       else if (atype == "rmc")
       {
-        scalar_estimator_inputs_.emplace_back(std::in_place_type<RMCLocalEnergyInput>, child);
+        appendScalarEstimatorInput<RMCLocalEnergyInput>(child);
         app_warning() << "RMCLocalEnergyEstimator support is at best experimental with batch drivers" << std::endl;
       }
       else if (atype == "onebodydensitymatrices")
-        estimator_inputs_.emplace_back(std::in_place_type<OneBodyDensityMatricesInput>, child);
+        appendEstimatorInput<OneBodyDensityMatricesInput>(child);
       else if (atype == "spindensity")
-        estimator_inputs_.emplace_back(std::in_place_type<SpinDensityInput>, child);
+        appendEstimatorInput<SpinDensityInput>(child);
       else if (atype == "momentumdistribution")
-        estimator_inputs_.emplace_back(std::in_place_type<MomentumDistributionInput>, child);
+        appendEstimatorInput<MomentumDistributionInput>(child);
       else
         throw UniformCommunicateError(error_tag + "unparsable <estimator> node, name: " + aname + " type: " + atype +
                                       " in Estimators input.");
@@ -92,14 +92,14 @@ void EstimatorManagerInput::readXML(xmlNodePtr cur)
     {
       std::string atype(lowerCase(getXMLAttributeValue(child, "type")));
       std::string aname(lowerCase(getXMLAttributeValue(child, "name")));
-      throw UniformCommunicateError(error_tag + "<Estimators> can only contain <Estimator> nodes");
+      throw UniformCommunicateError(error_tag + "<estimators> can only contain <estimator> nodes");
     }
 
     if (cur_name == "estimators")
       child = child->next;
     else
     {
-      app_summary() << "<estimator> nodes not contained in <estimators></estimators> is a deprecated input xml idiom"
+      app_summary() << "<estimator> nodes not contained in <estimators>...</estimators> is a deprecated input xml idiom"
                     << std::endl;
       break;
     }
--- a/src/Estimators/EstimatorManagerInput.cpp
+++ b/src/Estimators/EstimatorManagerInput.cpp
@@ -67,23 +67,23 @@ void EstimatorManagerInput::readXML(xmlNodePtr cur)
       if (aname.empty() && !atype.empty())
         aname = atype;
       if (atype == "localenergy" || atype == "elocal")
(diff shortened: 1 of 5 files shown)


src/Estimators/EstimatorManagerNew.cpp
```diff
@@ -48,76 +50,57 @@ EstimatorManagerNew::EstimatorManagerNew(const QMCHamiltonian& ham, Communicate*
...
   }
+
   for (auto& scalar_input : emi.get_scalar_estimator_inputs())
   {
-    bool estimator_made = false;
-    estimator_made      = estimator_made || createScalarEstimator<LocalEnergyInput>(scalar_input, H);
-    estimator_made      = estimator_made || createScalarEstimator<CSLocalEnergyInput>(scalar_input, H);
-    estimator_made      = estimator_made || createScalarEstimator<RMCLocalEnergyInput>(scalar_input, H);
-    if (!estimator_made)
+    std::unique_ptr<ScalarEstimatorBase> scalar_est;
```
PDoakORNL, line 81: This would be better written like this
```c++
   // since we can count on these being scalar estimator inputs and the existence of each inputs Consumer type
   // we don't need to check for no type being made only deal with the monotype, which if it shows up here inidicates
   // a developer error.

    bool estimator_made = std::visit(
                                     [this, &H](auto scalar_input) {
          using T = std::decay_t<decltype(scalar_input)>;
          if constexpr (std::is_same_v<T, std::monostate>)
            return false;
          else
          {
            auto estimator = std::make_unique<typename decltype(scalar_input)::Consumer>(std::move(scalar_input), H);
            if (estimator->isMainEstimator())
              addMainEstimator(std::move(estimator));
            else
              scalar_ests_.push_back(std::move(estimator));
          }
          return true;
        },
        scalar_input);

    if (!estimator_made)
      throw UniformCommunicateError(std::string(error_tag_) +
                                    "cannot construct a scalar estimator from scalar estimator input object.");

File: src/Estimators/EstimatorManagerNew.h [hunk 2]
```diff
@@ -65,7 +65,7 @@ class EstimatorManagerNew
    *  \param[in]  twf    The fully constructed TrialWaveFunction.
    */
   EstimatorManagerNew(Communicate* comm,
-                      EstimatorManagerInput&& emi,
+                      EstimatorManagerInput emi,
```
PDoakORNL, line 68: This should not be changed. Call it with an rvale. QMCDriverNew makes the input. by either merging the global and local emi to make a new object copying the local or global emi or providing the default emi if non is in the input.

src/Estimators/EstimatorManagerInput.h
```diff
@@ -82,18 +82,6 @@ class EstimatorManagerInput
   EstimatorInputs estimator_inputs_;
   ScalarEstimatorInputs scalar_estimator_inputs_;
```
PDoakORNL, line 84: Seems clear these should be simplified to
```c++
  template<typename T>
  void appendEstimatorInput(XMLNodePtr node)
  {
    estimator_inputs_.emplace_back(std::in_place_type<T>, node);
  }
```

src/Estimators/EstimatorManagerInput.cpp
```diff
@@ -50,23 +65,23 @@ void EstimatorManagerInput::readXML(xmlNodePtr cur)
       if (aname.empty() && !atype.empty())
         aname = atype;
       if (atype == "localenergy" || atype == "elocal")
-        appendScalarEstimatorInput<LocalEnergyInput>(child);
+        scalar_estimator_inputs_.emplace_back(std::in_place_type<LocalEnergyInput>, child);
```
PDoakORNL, line 68: I think this is just a style thing.  No different from any other function that just improves the reading of the code.

PDoakORNL (changes requested):
Is there a real compiler issue here?  Or is this just a style issue?  If it's just style I think the way I've expressed most of this reads better as is and leaves the door open to more elegant code later.  I've indicated where I'm going with this when relevant.

src/Estimators/EstimatorManagerInput.cpp
```diff
@@ -22,11 +22,26 @@ EstimatorManagerInput::EstimatorManagerInput(xmlNodePtr cur) { readXML(cur); }
 
 EstimatorManagerInput::EstimatorManagerInput(std::initializer_list<EstimatorManagerInput> emil)
 {
-  std::for_each(emil.begin(), emil.end(), [this](const EstimatorManagerInput& emi) {
-    std::copy(emi.estimator_inputs_.begin(), emi.estimator_inputs_.end(), std::back_inserter(estimator_inputs_));
+  size_t est_offset        = 0;
```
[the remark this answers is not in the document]
ye-luo: This is from https://github.com/QMCPACK/qmcpack/pull/4024. Compiler issue.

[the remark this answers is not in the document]
ye-luo: Added TODO note.
Q: does std::back_inserter potentially over allocate memory?
I'm not saying memory is a concern here. Just to understand how back_inserter works.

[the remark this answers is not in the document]
PDoakORNL: Well no choice then.

src/Estimators/EstimatorManagerNew.cpp
```diff
@@ -48,76 +50,57 @@ EstimatorManagerNew::EstimatorManagerNew(const QMCHamiltonian& ham, Communicate*
...
-    bool estimator_made = false;
-    estimator_made =
-        estimator_made || createEstimator<SpinDensityInput>(est_input, pset.getLattice(), pset.getSpeciesSet());
-    estimator_made = estimator_made ||
-        createEstimator<MomentumDistributionInput>(est_input, pset.getTotalNum(), pset.getTwist(), pset.getLattice());
-    estimator_made = estimator_made ||
-        createEstimator<OneBodyDensityMatricesInput>(est_input, pset.getLattice(), pset.getSpeciesSet(),
-                                                     twf.getSPOMap(), pset);
-    if (!estimator_made)
+    if (has<SpinDensityInput>(est_input))
```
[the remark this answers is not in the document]
ye-luo: Not a compiler issue so far. I saw your concern of write the type twice. Let me think a bit more.

src/Estimators/EstimatorManagerNew.h
ye-luo: I will restore it. Here is the reason.
The constructor needs to own an Input.
If being passed by value, the move happens at the call site.
If being passed by rvalue, there is no move, emi contained inputs will still be disassembled. But the emi object remains in QMCDriverNew. Since it is passed by rvalue, the caller QMCDriverNew should be aware that emi should not be used after the call site.

Overall, I still find it unsafe to move inputs around. For each individual estimator, it tries to move in its input. Consider the input has already been copied at QMCDriverNew. It should just copy. Then EstimatorManagerInput can be more const protected.

src/Estimators/EstimatorManagerInput.cpp
ye-luo: Likely. Probably will change appendScalarEstimatorInput to a lambda instead of a member function.

src/Estimators/EstimatorManagerNew.h
PDoakORNL: We should have a call and talk this through, I think you are right that there are just too many moves with the input object.  In this case QMCDriverNew has done the combination and ideally that is just in the function scope so moving it is fine, but the complications mount.

src/Estimators/EstimatorManagerInput.h
ye-luo: Done.

src/Estimators/EstimatorManagerInput.cpp
ye-luo: Restored and simplified. FYI template lambda is C++20.

src/Estimators/EstimatorManagerInput.h
ye-luo: Done.

PDoakORNL (approved):
This looks good to me now.  Better typesafety and cleaner.

PR #4028, opened 2022-05-25
Add listener "trait" and check if listening Estimators are instantiated, remove two sources of truth for walker serialization

PDoakORNL:
## Proposed changes

In support of #4029 

The lighter weight replacement for per particle hamiltonian evaluation has crowd scope estimators register listener functions  with QMCHamiltonian. The hamiltonian operators use multi walker resources to manage the shared crowd level buffers on the operator side. It's necessary at runtime to be able to check this "trait" of the estimator and see if any have been added to the EstimatorManagerNew.

There is already a source of truth for walker serialization in QmcdriverInput which is strongly coupled to QMCDriverNew use that instead of adding an accessor to MultiWalkerDispatchers which is detail QMCDriverNew should know as little as possible about. 

- No

Leconte

## Checklist

### DIFF
--- a/src/Estimators/EstimatorManagerNew.cpp
+++ b/src/Estimators/EstimatorManagerNew.cpp
@@ -50,6 +50,12 @@ EstimatorManagerNew::EstimatorManagerNew(const QMCHamiltonian& ham, Communicate*
   addMainEstimator(std::make_unique<LocalEnergyEstimator>(ham, true));
 }
 
+bool EstimatorManagerNew::areThereListeners()
+{
+  return std::any_of(operator_ests_.begin(), operator_ests_.end(),
+                     [](auto& oper_est) { return oper_est->isListenerRequired(); });
+}
+  
 template<class EstInputType, typename... Args>
 bool EstimatorManagerNew::createEstimator(EstimatorInput& input, Args&&... args)
 {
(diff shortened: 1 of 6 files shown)


src/Estimators/EstimatorManagerNew.cpp
```diff
@@ -50,6 +50,12 @@ EstimatorManagerNew::EstimatorManagerNew(const QMCHamiltonian& ham, Communicate*
   addMainEstimator(std::make_unique<LocalEnergyEstimator>(ham, true));
 }
 
+bool EstimatorManagerNew::areThereListeners()
```
ye-luo, line 53: const function?

src/Estimators/EstimatorManagerNew.h
```diff
@@ -146,6 +146,8 @@ class EstimatorManagerNew
   std::size_t getNumEstimators() { return operator_ests_.size(); }
   std::size_t getNumScalarEstimators() { return scalar_ests_.size(); }
 
+  bool areThereListeners();
```
ye-luo, line 149: Can we keeps some documentation explaining what is a listener?

src/Estimators/EstimatorManagerNew.cpp
PDoakORNL: yes

ye-luo (approved):
LGTM.
I still don't quite get what listeners are but probably will get more understanding once actual code appears.

ye-luo:
Test this please

PR #4030, opened 2022-05-25
Suppress NVHPC warning about unreachable statement

ye-luo:
## Proposed changes
Not much I can do about this warning. GCC/Clang wants the return but NVHPC doesn't like the return.
Example
```
  virtual size_t getPerTargetPctlStrideSize() const
  {
    throw std::runtime_error(name_ + " getPerTargetPctlStrideSize not supported");
    return 0;
  }
  
"/home/yeluo/opt/cleanup/qmcpack/src/Particle/DistanceTable.h", line 371: warning #111-D: statement is unreachable
      return 0;
```
So suppress the warning of NVHPC.

- Build related changes

- No

epyc-server

## Checklist

### DIFF
--- a/CMake/NVHPCCompilers.cmake
+++ b/CMake/NVHPCCompilers.cmake
@@ -38,7 +38,7 @@ add_definitions(-Drestrict=__restrict__)
 # 998 function "AA" is hidden by "BB" -- virtual function override intended?
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} --display_error_number --diag_suppress 177 --diag_suppress 550")
 set(CMAKE_CXX_FLAGS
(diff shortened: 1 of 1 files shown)


ye-luo:
Test this please

markdewing:
An alternate solution might be to use the `[[noreturn]]` attribute to keep clang and gcc happy.
```
[[noreturn]] virtual size_t getPerTargetPctlStrideSize() const
{
    throw std::runtime_error(name_ + " getPerTargetPctlStrideSize not supported");
}
```

ye-luo:
> An alternate solution might be to use the `[[noreturn]]` attribute to keep clang and gcc happy.
> 
> ```
> [[noreturn]] virtual size_t getPerTargetPctlStrideSize() const
> {
>     throw std::runtime_error(name_ + " getPerTargetPctlStrideSize not supported");
> }
> ```

Seems better. Trying.

PR #4031, opened 2022-05-26
Reduce NVHPC warning

ye-luo:
## Proposed changes
It seems the code base is also improved.

- Other (please describe):

- No

epyc-server

## Checklist


ye-luo:
Test this please

PDoakORNL (approved):
Style wise I prefer most of these changes.  Never been a fan of unreachable returns and breaks.

PR #4033, opened 2022-05-26
Eliminate nodiscard warnings in QMC_CUDA

jakurzak:
## Proposed changes

All calls to the CUDA/HIP API in the legacy CUDA code (QMC_CUDA) are placed inside the cudaCheck() macro, which checks the return value.

- Adding missing checks for the error codes returned from the CUDA/HIP API.

- No

- AMD Radeon VII

## Checklist

- Yes, this PR is up to date with the current state of 'develop'.
- No, the code added or changed in the PR has NOT been clang-formatted - preserving the original formatting instead.
- No, this PR does not add tests. It adds more error checks in the existing code.
- No, documentation does not apply.


ye-luo:
Test this please

src/Platforms/CUDA_legacy/cuda_error.h
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+// If the cause is largely redundant with the __FILE__ and __LINE__ information
+#define cudaCheck(ans)                         \
+  {                                            \
+    cudaAssert((ans), "", __func__, __FILE__, __LINE__); \
+  }
+
+/// prints CUDA error messages. Always use cudaErrorCheck macro.
+inline void cudaAssert(cudaError_t code, const std::string& cause, const char* function, const char* filename, int line, bool abort = true)
```
PDoakORNL, line 42: It bothers me a bit that this a copy of src/Platforms/CUDA/CUDAerror.h which basically follows our file naming conventions to cuda_error.h which definitely doesn't.  The defect that the implementation belongs in a cpp file persists.

src/einspline/bspline_create_cuda.cu
```diff
@@ -261,13 +262,7 @@ create_UBspline_3d_s_cuda_conv (UBspline_3d_d* spline)
...
   size_t size = Nx*Ny*N*sizeof(float);
 
-  cudaMalloc((void**)&(cuda_spline->coefs), size);
-  cudaError_t err = cudaGetLastError();
-  if (err != cudaSuccess) {
-    fprintf (stderr, "Failed to allocate %ld memory for GPU spline coefficients.  Error %s\n",
-	     size, cudaGetErrorString(err));
-    abort();
-  }
+  cudaCheck(cudaMalloc((void**)&(cuda_spline->coefs), size));
```
PDoakORNL, line 265: It seems like we're losing most of the specificity in the error output here.  Since cudaCheck throws this can be put back.

src/einspline/multi_bspline_create_cuda.cu
```diff
@@ -215,13 +215,7 @@ create_multi_UBspline_1d_c_cuda_conv (multi_UBspline_1d_z* spline)
 		         0.0,      0.0,     -3.0,     1.0,
 		         0.0,      0.0,      1.0,     0.0 };
 
-  cudaMemcpyToSymbol(Acuda, A_h, 48*sizeof(float), 0, cudaMemcpyHostToDevice);
```
PDoakORNL, line 218: Again we're losing the legible error message.  The abort() is good to lose.

PDoakORNL (changes requested):
The contents of CUDAError.h don't need to be copied.

Informative error messsages (which may never have been seen because they just called abort()) are replaced by __FILE__, __LINE__ errors.

On the other hand we plan to delete this code entirely so there is virtue to having no contamination at all in Platforms/CUDA.

src/Platforms/CUDA_legacy/cuda_error.h
ye-luo: @PDoakORNL I requested the copy. legacy_CUDA should not overlap codes with CUDA. Otherwise, the CMake target dependency gets further complicated.

PDoakORNL: I can see that point and I too would rather continue to improve our CMake target setup.

jakurzak: So, are we okay with the copy?

jakurzak:
Please let me know how I can help move this forward.

src/Platforms/CUDA_legacy/cuda_error.h
PDoakORNL: yes

prckent (commented):
The errors that we absolutely have to trap and give an informative message to the user about are the "big allocations". These are the 3D * no. orbitals sized spline buffers used for the orbitals (I think at least multi_bspline_create* line 339 or so), and where the determinants/matrices are allocated. If the code does not run out of memory at these points there is good chance the simulation will run. I would be happier if whereever in the code we had an existing error check within a few lines of a cuda call that we could still get to that error message. The original CUDA implementation had minimal error checks and many of the additions were added based on experience.

jakurzak:
@prckent I can wrap all cudaMalloc calls in QMC_CUDA in a cudaCheckMalloc macro that prints the size of the failed allocation.

jakurzak:
I added a cudaCheckMalloc() macro that optionally takes the size and purpose of the allocation.
I used it for the three allocations that had verbose error messages.
I left the other allocations in the simple cudaCheck() macro.
This preserves the level of verbosity of the original code, while having all CUDA/HIP API calls inside error-checking macros.
Looking forward to more feedback.

prckent (dismissed):
Works for me. Thanks Jakub.

PDoakORNL (dismissed):

prckent:
Test this please

ye-luo:
Test this please

jakurzak:
I made a fix and tested on AMD and NVIDIA and looks like all CI tests are passing now.

ye-luo (approved):

PR #4034, opened 2022-05-27
Listeners to allow call backs from Operators to Estimators

PDoakORNL:
## Proposed changes

This is a basic piece of infrastructure code to reporting of per particle values in the batched regime from operators to crowd scope estimators.
 
It's just the class and test.  In all ready working code this allows calls through QMCHamiltonian mw_evaluate to deliver the needed values for EnergyDensityEstimator.

Later PRs will bring this in small pieces.

- New feature
- Documentation changes

- No

Leconte

## Checklist


src/QMCHamiltonians/Listener.hpp
```diff
@@ -0,0 +1,66 @@
...
+{
+
+/** An object of this type is a listener expecting a callback to the report function  with a vector of values,
+ *  the convention being 1 per particle, called once per walker per component.
+ *  The name is primarily for debugging purposes, if have decided against having the QMCHamiltonian use it to estable
+ *  routing. Instead the register functions in the QMCHamiltonian are specfic to the sort of value that the listener wants
+ *  to listen to.
+ */
+template<typename T>
+class ListenerVector
```
markdewing, line 38: Are there going to be other types passed to the report function? In other words, is there going to be a class named "ListenerScalar"?   Seeing the name ListenerVector, I immediately think of a vector of Listeners, but it is not.  Reading the docs, it looks like the "Vector" part means it expects a vector input.

PDoakORNL: Your point is good.  I did have a ListenerVar and ListenerMatrix but they haven't proved to be useful yet so I left them out of this PR. Maybe ListenerToVector would be better, naming these sorts of things is tricky.

PDoakORNL:
Test this please

src/QMCHamiltonians/tests/test_Listener.cpp
```diff
@@ -0,0 +1,78 @@
...
+};
+
+TEST_CASE("ListenerVector", "[hamiltonian]")
+{
+  Talker talker;
+  TestReceiver test_receiver;
+
+  talker.registerVector(test_receiver.makeListener());
+
+  talker.reportVector();
```
ye-luo, line 72: Which objects own talker or listener?
My quick guess is Estimator owns a listener while EstimatorManager owns a talker.
Estimator generates data while EstimatorManager consumes the data.
It seems now listener is giving data to talker. Sounds counter-intuitive. Maybe I got this wrong completely, just treat me as a normal developer. Seems we need to reduce confusion.

PDoakORNL: In the current design the listeners are passed by value to the QMCHamiltonian from the crowd scope Estimator, the QMCHamiltonian owns the listeners. They are valid only while the the crowd scope Estimator data object is valid.

Abstractly a listener is only likely to be valid as long as the object that produced the listener is valid. The lambda could have captured *this from the object that produced it.

The "receiver" is the crowd scope Estimator.
The "talker" is the QMCHamiltonian itself.

ye-luo (approved):
I have a bit more understanding of the Listener class and more complete picture should show up on following PRs. Efforts will be needed to improve naming to help the readability of the code.

ye-luo:
Test this please

Issue #4039, opened 2022-05-31
CUDA 11.7 doesn't compile std::map when using gcc 12.1 as a host compiler


ye-luo:
**Is your feature request related to a problem? Please describe.**
Test a compilation in CMake and issue necessary stopper and suggestions (change CMAKE_CXX_COMPILER or use CMAKE_CUDA_HOST_COMPILER).

prckent:
Sensible to issue a stopper. gcc 12.1 was released on May 6 https://www.gnu.org/software/gcc/gcc-12/ while CUDA 11.7, released on May 12, only claims support for gcc <=11 https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#system-requirements

prckent:
Closed by #4041

PR #4041, opened 2022-05-31
Add error trap if GCC 12 or later is used with CUDA 11.7

quantumsteve:

## Proposed changes

This adds and error trap if GCC 12 or later is used with CUDA 11.7. The user is encouraged to change CMAKE_CXX_COMPILER or use CMAKE_CUDA_HOST_COMPILER.

This fixes issue #4039 

- Bugfix
- Build related changes

- No

## Checklist


CMakeLists.txt
```diff
@@ -723,6 +723,13 @@ if(QMC_CUDA OR ENABLE_CUDA)
       set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --allow-unsupported-compiler")
     endif()
     enable_language(CUDA)
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 12.0)
+      if(CMAKE_CUDA_COMPILER_VERSION VERSION_EQUAL 11.7)
```
prckent, line 727: I suggest the inequality on the CUDA version as well:

VERSION_LESS_EQUAL 11.7

and update the message
CUDA versions <=  11.7 do not support GCC versions >= 12. Please change...  e.g. use an older GCC version compatible with the CUDA version.

ye-luo (changes requested):
I'd like to avoid living on versions.
It is better to test compiling. `test_map.cu`
```
#include <map>
int main()
{ }
```

prckent:
It is better to do both => Lets do both.

CMake/TestGNUCUDAVersion.cmake
```diff
@@ -0,0 +1,16 @@
+set(CUDA_GNU_COMPATIBLE TRUE)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 12.0)
+  if(CMAKE_CUDA_COMPILER_VERSION VERSION_LESS_EQUAL 11.7)
+    set(CUDA_GNU_COMPLATIBLE FALSE)	  
+  endif()
+endif()
```
ye-luo, line 6: If I use GCC 12 as the host compiler and set CMAKE_CUDA_HOST_COMPILER to GCC 11. This trap still stops me. So not good.

I think we just can't act on CMAKE_CXX_COMPILER at all since only CMAKE_CUDA_HOST_COMPILER matters.

CMake/TestGNUCUDAVersion.cmake
```diff
@@ -0,0 +1,16 @@
...
+
+file(COPY ${PROJECT_SOURCE_DIR}/CMake/try_compile_sources/test_map.cu DESTINATION ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp)
```
ye-luo, line 8: Feel better to skip COPY

CMake/TestGNUCUDAVersion.cmake
```diff
@@ -0,0 +1,16 @@
...
+try_compile(PASS_MAPCUDA ${CMAKE_BINARY_DIR} ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp/test_map.cu
```
ye-luo, line 9: By the way, only QMC_CUDA case needs this check.

CMake/TestGNUCUDAVersion.cmake
quantumsteve: Does CMake provide a way to get information about the CMAKE_CUDA_HOST_COMPILER? I'd like to avoid parsing terminal output.

ye-luo: If you mean get the version of host compiler beneath nvcc, I think the answer is no. Maybe we should just drop the version checking and only relies on the compilation check.

quantumsteve:
@prckent @ye-luo I think this is ready for review.

ye-luo:
Test this please

ye-luo (approved):

PR #4050, opened 2022-06-06
Expand NiO performance test walker counts.

ye-luo:
## Proposed changes
Previously we hard-coded walker count 16 or 32. This makes benchmark harder to reflect performance of real simulations when problem size is small. Thus introduce variable counts. There is no perfect choices but I'm sure the current choice should be better than 16/32.

Mostly make small problem size runs heavier but large problem sizes unchanged.

- No

epyc-server

## Checklist

### DIFF
--- a/tests/performance/NiO/CMakeLists.txt
+++ b/tests/performance/NiO/CMakeLists.txt
@@ -112,6 +112,10 @@ else()
       64
       64
       64)
+  # walker count preset for NIO_SIZES problems
+  # These numbers are not optimal but should be indicative of performance
+  # Walker count scan for each problem size is still required to maximize throughput
+  # for a given hardware if its memory capacity allows.
   set(WALKER_COUNTS
       1024
       512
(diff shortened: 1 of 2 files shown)


prckent (changes requested):
This is a sensible idea which we should merge.

Can you please (1) add a comment with the expected runtimes with these walker counts on a V100 (say, mixed precision build?)

(2) Add a comment with our usual disclaimer along the lines that these are not optimal but should be indicative of performance, higher throughput would be obtained with more up to memory limit etc.

ye-luo:
Added disclaimer.
128 atom and beyond. The walker count remains 16.
I'm not even considering V100 as the baseline.
Most tests runs about 20~100 sec on my RTX 3060Ti.

prckent (commented):
Update graphite as well?

ye-luo:
> Update graphite as well?

Those tests also need to be adjusted but I’d like to have more data points at least from tonight.

ye-luo:
Test this please

prckent (approved):

PR #4057, opened 2022-06-11
Expand OhmmsArray APIs.

ye-luo:
## Proposed changes
To prepare for the actual use.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Containers/OhmmsPETE/OhmmsArray.h
+++ b/src/Containers/OhmmsPETE/OhmmsArray.h
@@ -98,70 +98,70 @@ public:
   }
 
   template<typename SIZET = size_t, typename = std::is_integral<SIZET>>
-  Type_t* data(const std::array<SIZET, D>& dims)
+  Type_t* data(const std::array<SIZET, D>& offsets)
   {
-    size_t offset = dims[0];
-    for (int i = 1; i < dims.size(); i++)
-      offset = offset * Length[i] + dims[i];
+    size_t offset = offsets[0];
+    for (int i = 1; i < offsets.size(); i++)
+      offset = offset * Length[i] + offsets[i];
     return X.data() + offset;
   }
   template<typename SIZET = size_t, typename = std::is_integral<SIZET>>
-  const Type_t* data(const std::array<SIZET, D>& dims) const
+  const Type_t* data(const std::array<SIZET, D>& offsets) const
   {
-    size_t offset = dims[0];
-    for (int i = 1; i < dims.size(); i++)
-      offset = offset * Length[i] + dims[i];
+    size_t offset = offsets[0];
+    for (int i = 1; i < offsets.size(); i++)
+      offset = offset * Length[i] + offsets[i];
     return X.data() + offset;
   }
   template<typename SIZET     = size_t,
            typename           = std::is_integral<SIZET>,
            typename Allocator = ALLOC,
            typename           = qmcplusplus::IsDualSpace<Allocator>>
-  Type_t* device_data(const std::array<SIZET, D>& dims)
+  Type_t* device_data(const std::array<SIZET, D>& offsets)
   {
-    size_t offset = dims[0];
-    for (int i = 1; i < dims.size(); i++)
-      offset = offset * Length[i] + dims[i];
+    size_t offset = offsets[0];
+    for (int i = 1; i < offsets.size(); i++)
+      offset = offset * Length[i] + offsets[i];
     return X.device_data() + offset;
   }
   template<typename SIZET     = size_t,
            typename           = std::is_integral<SIZET>,
            typename Allocator = ALLOC,
            typename           = qmcplusplus::IsDualSpace<Allocator>>
-  const Type_t* device_data(const std::array<SIZET, D>& dims) const
+  const Type_t* device_data(const std::array<SIZET, D>& offsets) const
   {
-    size_t offset = dims[0];
-    for (int i = 1; i < dims.size(); i++)
-      offset = offset * Length[i] + dims[i];
+    size_t offset = offsets[0];
+    for (int i = 1; i < offsets.size(); i++)
+      offset = offset * Length[i] + offsets[i];
     return X.device_data() + offset;
   }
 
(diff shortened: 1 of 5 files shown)


src/Containers/OhmmsPETE/OhmmsArray.h
```diff
@@ -95,6 +97,74 @@ class Array
...
+  template<typename SIZET = size_t, typename = std::is_integral<SIZET>>
+  Type_t* data(const std::array<SIZET, D>& offsets)
+  {
+    size_t offset = offsets[0];
+    for (int i = 1; i < offsets.size(); i++)
+      offset = offset * Length[i] + offsets[i];
+    return X.data() + offset;
+  }
+  template<typename SIZET = size_t, typename = std::is_integral<SIZET>>
+  const Type_t* data(const std::array<SIZET, D>& offsets) const
```
PDoakORNL, line 109: As far as I can tell this is an std::array of indexes.
The term offset makes me think these are different.  You can have an offset into a buffer.  And in CUDA at least there is a notion of a 2D and 3D buffer, but I don't think this is that.

src/Containers/OhmmsPETE/OhmmsArray.h
```diff
@@ -95,6 +97,74 @@ class Array
...
+           typename           = qmcplusplus::IsDualSpace<Allocator>>
+  const Type_t* device_data(const std::array<SIZET, D>& offsets) const
+  {
+    size_t offset = offsets[0];
+    for (int i = 1; i < offsets.size(); i++)
+      offset = offset * Length[i] + offsets[i];
+    return X.device_data() + offset;
+  }
+
+  /// access data pointer at (offset_1, ..., offset_D})
```
PDoakORNL, line 139: Again different from D dimensional indexes?  For docs I'd just put them all in a doxygen group and document there the idea of indexing a D dimensional tensor to a data ptr.

src/Containers/OhmmsPETE/OhmmsArray.h
```diff
@@ -95,6 +97,74 @@ class Array
...
+  template<typename... Args>
+  Type_t* data(Args... offsets)
+  {
+    static_assert(sizeof...(Args) == D, "resize arguments must match dimensionality of Array");
```
PDoakORNL, line 143: I think the assert message needs to be updated.

src/Containers/OhmmsPETE/OhmmsArray.h
```diff
@@ -95,6 +97,74 @@ class Array
     return X.device_data();
   }
 
+  template<typename SIZET = size_t, typename = std::is_integral<SIZET>>
+  Type_t* data(const std::array<SIZET, D>& offsets)
+  {
+    size_t offset = offsets[0];
+    for (int i = 1; i < offsets.size(); i++)
```
PDoakORNL, line 104: This is a repetition written differently of the operator() indexing code.  Might be better have it all work the same.  They all calculate a linear index into the data from D indexes.

src/Containers/OhmmsPETE/tests/test_Array.cpp
```diff
@@ -73,6 +73,12 @@ TEST_CASE("array NestedContainers", "[OhmmsPETE]")
   CHECK(vec_copy(0).back() == 123);
 }
 
+TEST_CASE("Array::data", "[OhmmsPETE]")
```
PDoakORNL, line 76: This doesn't seem like quite enough.
Think about what could go wrong when the code gets modified.
Check some edge cases.
Array<T,1>
First and last indicies, 0 and max.
Array where size of the type is different from double.
populate the array and make sure that the dereference is the data at that location (the same error could be easily made in the hand indexing in the test and the data member function.

Basically the same tests can cover the operator() functions.

PDoakORNL (changes requested):
Some of the naming and the similar code in operator() functions should be addressed.  The access algorithm seems fine.

src/Containers/OhmmsPETE/OhmmsArray.h
```diff
@@ -95,6 +97,74 @@ class Array
     return X.device_data();
   }
 
+  template<typename SIZET = size_t, typename = std::is_integral<SIZET>>
+  Type_t* data(const std::array<SIZET, D>& offsets)
```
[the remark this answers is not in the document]
ye-luo: changed to `data_at()`

src/Containers/OhmmsPETE/OhmmsArray.h
ye-luo: put `indices` as the argument name

src/Containers/OhmmsPETE/OhmmsArray.h
ye-luo: put `indices` as the argument name

src/Containers/OhmmsPETE/OhmmsArray.h
ye-luo: Fixed.

src/Containers/OhmmsPETE/OhmmsArray.h
ye-luo: Rewrite `operator()` with parameter packing

src/Containers/OhmmsPETE/tests/test_Array.cpp
ye-luo: > Check some edge cases.
Array<T,1>
First and last indicies, 0 and max.
Array where size of the type is different from double.

Not understanding why the type needs to be different from double. Just changed double to float.
Also added value checking.

ye-luo:
Test this please

src/Containers/OhmmsPETE/tests/test_Array.cpp
PDoakORNL: I would say its to make sure future work doesn't somehow only work for double or have some hard coded dependence on the sizeof(double)

PDoakORNL (approved):

PR #4058, opened 2022-06-11
Adjust layout phi_vgl_v

ye-luo:
Need #4057 first

## Proposed changes
Previously phi_vgl_v object passed between SPOSet and DIracDeterminantBatched was host space only but relies on OpenMP magic to find the device pointer. This PR makes phi_vgl_v explicitly in two spaces and assume data ready on the device.
Also previously phi_vgl_v was using SoA containing which is not necessary and add difficulty to use BLAS due to the padding and collapsing of walker and orbital dimensions. This PR changes data type to Array and avoid padding.
So SPOSet data padding doesn't leak from the SPOSet object.

- Refactoring (no functional changes)

- No

epyc-server

## Checklist


src/Configuration.h
```diff
@@ -50,7 +50,8 @@ struct QMCTraits
 {
   enum
   {
-    DIM = OHMMS_DIM
+    DIM = OHMMS_DIM,
+    DIM_VGL = OHMMS_DIM + 2
```
markdewing, line 54: It might be useful to have a comment to explain what VGL means and where the +2 comes from.

src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
```diff
@@ -293,32 +293,32 @@ class MatrixDelayedUpdateCUDA
...
+    auto& cone_vec              = engine_leader.mw_mem_->cone_vec;
+    auto& czero_vec             = engine_leader.mw_mem_->czero_vec;
+    const int norb              = engine_leader.get_ref_psiMinv().rows();
+    const int lda               = engine_leader.get_ref_psiMinv().cols();
+    const int nw                = engines.size();
+    const size_t phi_vgl_stride = nw * norb;
     mw_temp.resize(norb * n_accepted);
     mw_rcopy.resize(norb * n_accepted);
-    updateRow_buffer_H2D.resize((sizeof(Value*) * 8 + sizeof(Value)) * n_accepted);
+    updateRow_buffer_H2D.resize((sizeof(Value*) * 6 + sizeof(Value)) * n_accepted);
```
markdewing, line 308: What does the "6" mean and why was it changed from "8"? 
The "6" gets reused in a few more places, it might be worth making it a named variable.

src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
```diff
@@ -579,11 +574,12 @@ class MatrixDelayedUpdateCUDA
...
     const int nw                      = engines.size();
     const int n_accepted              = psiM_g_list.size();
-    accept_rejectRow_buffer_H2D.resize((sizeof(Value*) * 14 + sizeof(Value)) * nw);
+    const size_t phi_vgl_stride       = nw * norb;
+    accept_rejectRow_buffer_H2D.resize((sizeof(Value*) * 12 + sizeof(Value)) * nw);
     engine_leader.resize_fill_constant_arrays(nw);
 
-    Matrix<Value*> ptr_buffer(reinterpret_cast<Value**>(accept_rejectRow_buffer_H2D.data()), 14, nw);
-    Value* c_ratio_inv = reinterpret_cast<Value*>(accept_rejectRow_buffer_H2D.data() + sizeof(Value*) * 14 * nw);
+    Matrix<Value*> ptr_buffer(reinterpret_cast<Value**>(accept_rejectRow_buffer_H2D.data()), 12, nw);
```
markdewing, line 581: The "12" value (that was changed from "14") gets used in a few more places - might be worth making it a named value.

src/QMCWaveFunctions/Fermion/MatrixUpdateOMPTarget.h
```diff
@@ -240,41 +242,40 @@ class MatrixUpdateOMPTarget
...
+    const size_t phi_vgl_stride = nw * norb;
 
     engine_leader.resize_scratch_arrays(norb, n_accepted);
 
     // to handle Value** of Ainv, psi_v, temp, rcopy
-    buffer_H2D.resize((sizeof(Value*) * 8 + sizeof(Value)) * n_accepted);
-    Matrix<Value*> ptr_buffer(reinterpret_cast<Value**>(buffer_H2D.data()), 8, n_accepted);
-    Value* c_ratio_inv = reinterpret_cast<Value*>(buffer_H2D.data() + sizeof(Value*) * 8 * n_accepted);
-    for (int iw = 0, count = 0; iw < isAccepted.size(); iw++)
+    buffer_H2D.resize((sizeof(Value*) * 6 + sizeof(Value)) * n_accepted);
```
markdewing, line 267: More magic "6" values changed from "8".

prckent:
I agree with Mark here - this is a good opportunity to get rid of these magic numbers. They would not be obvious to anyone not familiar with real space QMC and even then they are error prone.

ye-luo:
Test this please

markdewing (approved):

prckent (approved):

PR #4061, opened 2022-06-15
Optimizer Compatibility with Batched Drivers and Parameter Filtration

LeonOtis:
## Proposed changes

These changes make several optimizers: the adaptive three shift version of the linear method, accelerated descent, and hybrid combinations of them, usable with the batched drivers. This PR also adds the option to filter the parameters optimized by the LM based on the amount of noise in the parameter gradients. We have found that allowing the LM to optimize only a filtered subset of the parameters while leaving the rest to accelerated descent to be an effective choice in recent work ( https://arxiv.org/abs/2111.07221 ). This option is available for both the batched and legacy versions of the drivers. 

Most changes in the code are in the optimizer engines or deal with interfacing them to the batched drivers. The latter was more challenging than I expected so I'm happy to clarify anything and welcome any improvements.

- New feature

- No

Lawrencium computing cluster

## Checklist


ye-luo:
Thank you Leon! To accelerate the acceptance of the feature you developed, it will be helpful if you can think about disassembling the components of this PR into several  self-organized testable PRs. For example, if changing A requires changing B and adding C, we may have a standalone C without need to be connected to the rest of the code. We may also adjust B or part of it if that doesn't affect the existing functionality. In these break-up PRs. we may think of adding documentation and tests.

You may park this PR aside and start fresh from develop and copy over changes from this PR. With this process, we will see growing merged changes in develop and also reducing not yet emerged changes in this PR.

prckent:
Hi Leon - Thanks very much for this. As Ye wrote it would be helpful if you can break this into smaller pieces.

I do have one small question - have you thought about what it would take to get this to work both with and without OpenMP threads and with and without MPI? These are basically a hard requirement for production so it would be helpful to know if you simply didn't get around to this or if there is a tricky problem that we would have to solve in future.

(Building without MPI is a requirement for convenient tools and workstation use, while and use of threads is essential to limit memory usage e.g. for spline runs)

LeonOtis:
Thanks, Paul and Ye, for the comments. I'll break up the changes I made and try to have a PR for a first component pretty soon. I haven't gotten around to making sure things work with all the OpenMP threads and/or MPI cases. My normal use case is to just use MPI without threads so the code changes were tested with that. The optimizers with the legacy drivers should work with or without OpenMP and with or without MPI, but I will have to see what changes might be needed with the batched drivers.

markdewing:
Could you describe more about what you found difficult in interfacing with the batched drivers? It would be helpful for us to know which areas need more documentation or description.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -614,12 +936,11 @@ QMCCostFunctionBatched::EffectiveWeight QMCCostFunctionBatched::correlatedSampli
 
         TrialWaveFunction::mw_evaluateParameterDerivatives(wf_list, p_list, optVars, dlogpsi_array, dhpsioverpsi_array);
 
-
-        // Energy
         auto energy_list =
-            QMCHamiltonian::mw_evaluateValueAndDerivatives(h0_list, wf_list, p_list, optVars, dlogpsi_array,
+            QMCHamiltonian::mw_evaluateValueAndDerivatives(h_list, wf_list, p_list, optVars, dlogpsi_array,
```
markdewing, line 940: This change from h0 to h is a change in behavior of existing code, and changes how the batched code works relative to the legacy code.   If this was an intentional change, it needs to be understood more deeply.
"h0_list" in the batched code corresponds to "H_KE_Node" in the legacy code.  It includes the KE energy part of the Hamiltonian and, optionally, the nonlocal part of the psuedopotential ("nonlocalpp" parameter in the input file)

prckent (changes requested):
Noting that several tests of this feature will be required. This can be in one of the smaller PRs when this gets split up. We have learned that code we do not have inputs for can not be kept alive in practice.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
prckent: Presumably this change is contributing to the broken optimizer tests revealed in CI.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -456,7 +470,310 @@ void QMCCostFunctionBatched::engine_checkConfigurations(cqmc::engine::LMYEngine<
                                                         DescentEngine& descentEngineObj,
                                                         const std::string& MinMethod)
 {
-  APP_ABORT("LMYEngine not implemented with batch optimization");
+
```
markdewing, line 473: I wonder if there's a way to architect the code to avoid duplicating checkConfigurations into engine_checkConfigurations? The changes between them are not very large. 
The big issue is the parameters that get passed - checkConfigurations doesn't have any, and engine_checkConfigurations needs EngineObj, descentEngineObj, and MinMethod.  
Potential options
1. Combine the three arguments into a single struct (engineData?) that gets added to checkConfigurations.  the contents of this struct could be defined differently based on HAVE_LMY_ENGINE.
The added code inside of checkConfigurations is fairly well separated and could be protected by '#ifdef'
2. Fancier version - create a base class with callbacks that get called at different points in the function.   Pass that class as an argument to checkConfigurations. This is probably overkill for now.

ye-luo: Regarding parameters consumed by the optimizes, they are all defined in the top level class and keeps growing. They should be organized in structs.

LeonOtis:
> Could you describe more about what you found difficult in interfacing with the batched drivers? It would be helpful for us to know which areas need more documentation or description.

One problem I ran into was in getting the correlated sampling phase of the adaptive three shift LM to agree with the results from using the legacy drivers. There seemed to be several issues when I was testing. First, electron positions did not match between batched and legacy drivers during the correlated sampling, which could be fixed by preventing the walker initialization from being executed again in VMCBatched. After that, the local energies still didn't match between the two driver types, but I found that inserting a call to flex_evaluateLog in VMCBatched and a call to mw_evaluateLog in QMCCostFunctionBatched removed the discrepancy for a small all electron test. I also found that the correlated sampling did not give accurate local energies in tests with pseudopotentials without the change from h0_list to h_list.

The other issue I had was ensuring that different amounts of sampling could be performed for different methods in a hybrid optimization. Having QMCDriverInput and VMCDriverInput copies read the selected xml blocks and overwrite the sampling settings in QMCFixedSampleLinearOptimizeBatched was the way I addressed this.

My understanding of the batched driver code is still quite limited so I'm very happy to make corrections if those changes break other parts of qmcpack or if there are better solutions. I think the first part of this PR that I will try to break off and resubmit will be the changes for interfacing with the batched drivers.

PDoakORNL (changes requested):
There really needs to be some thought put into this. 

I really think the different optimizer methods should have separate input classes like the different estimators. Fixed input parameters should be clearly separated from calculated data members.  Separate algorithms/enginers should have separate classes. QMCCostFunct, QMCLinearOpt should not be expanded to do multiple different methods each with there own set of member variables.  Parse first into input classes. The inputs don't look that similar in the docs and should be more straight forward to validate without having a swiss optimization input army put function. Using that input construct a specialized single method/engine object possibly with a base type, for code reuse (preferably via pure functions). Types should contain data members relevant to them.  Branching should be on optional behavior of an optimization method not what amounts to different methods smashed into a single type with mutating behavior. 

I have more specific feedback but I think its probably more than is useful to give because I don't think this PR can really be brought into a mergable state.

ye-luo:
I assume this is not needed any more.

PR #4066, opened 2022-06-16
Multi Walker i.e. Crowd resources created with Crowd context

PDoakORNL:
## Proposed changes

In the course of writing up milestone 6.9 it has become clear to me that we must create the multiwalker resources one the crowd context exists.  It doesn't make sense to create the multiwalker resources in the main thread with the golden walkers and then clone them into the crowds.

Multi Walker Resource creation can be done in one step in the Crowd constructor using the golden walker elements. This allows the creation calls for the ResourceCollections and the Estimator Listeners (which are crowd scope) to be added at the same time. It removes the need for the hack with the copy constructor of the CUDALinearAlgebraHandles that made them not copies but new instances (i.e. called cudaCreateStream and cublasCreate).

To make CUDALinearAlgebraHandles function in a non surprising way std::shared_ptr is used to make sure the lifetime of the CUDA stream handle and cuBLAS handle are properly treated. 

I've added a build workaround required by the current functional clang 15 offload compiler which cannot build offload target kernels with -O0.  This breaks out the omptarget files in ParticleSet and QMCWavefunctions as separate targets.  If you actually use a debugger and have experienced using it on -O3 code you'll realize there is no way we want all of ParticleSet and QMCWavefunctions compiled with -O3 in debug.

A new benchmark is added that looks at the importance of having multiple streams for efficient direct inversion.  I had thought for a bit that was broken in the main application but it was not because of the trick with the CUDALinearAlgebraHandles copy constructor.  That wasn't a good solution though, it still works properly and for the right reasons with this change.

I have preserved the behavior of CLAH when it is cloned as part of a ResourceCollection copy because that is how all the other multiwalker resources work.  I think this is ok for the clone methods but I do not think the resources should have fake copy constructors I think the clone methods should just call new on the default constructors.

- Bugfix (Debug build)
- New feature (mw resources constructed in the context they should be)
- Refactoring, in that this PR should have the same behavior and state after crowd construction that develop currently has.
- Build related changes
- Documentation changes

- No

Leconte

## Checklist


prckent:
Test this please

ye-luo (changes requested):
Unfortunately too many changes and they needs to be break into PRs into individual topics.

src/Particle/CMakeLists.txt
```diff
@@ -53,6 +51,34 @@ target_include_directories(qmcparticle PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
...
+  add_library(qmcparticle_omptarget OBJECT ${PARTICLEOMPTARGET})
+else(USE_OBJECT_TARGET)
+  add_library(qmcparticle_omptarget ${PARTICLEOMPTARGET})
+endif(USE_OBJECT_TARGET)
+
+target_include_directories(qmcparticle_omptarget PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
+target_link_libraries(qmcparticle_omptarget PUBLIC qmcutil containers)
+
+if(ENABLE_OFFLOAD AND ${COMPILER} MATCHES "Clang")
+    target_compile_options(qmcparticle_omptarget PRIVATE $<$<CONFIG:DEBUG>:-O3>)
```
ye-luo, line 75: Separating offload source organization can be useful also for other purpose.

I'm actually wondering if you ever want to debug with -O3. -O3 change significantly numerics and often ends up changing reproducibility. I usually do `-DCMAKE_CXX_FLAGS_DEBUG=-O3`.

PDoakORNL: If you have a bug that only appears when optimization is on and you want all the other flags that come with Debug vs. Release with debug info it could be useful.

PDoakORNL:
Breaking this up into smaller PR's

src/Particle/CMakeLists.txt
ye-luo: Does the way you did end up with "-O3 -O0 -g" in the compile line?

My feeling is getting code passing tests "-O0" can be helpful but largely irrelevant. What we need is -O3 flawless right?

PDoakORNL:
No the way this works is you get '-O0 -g -O3' and the compiler uses the last optimization flag it sees.  This is the "correct" way to do this in cmake instead of modifying the CXX_DEBUG_FLAGS and the modifying them back.

For using unit tests for debugging we want as much code compiled -O0.  I do most debugging in the unit/integration test context and not in the application.  If its an application level bug I usually write a test that reproduces it if I can. 
At -O0 certain access violations are caught that are not at -O3 or -O2 and tests that fail because of this crash instead of fail a CHECK usually.

So if the tests don't pass flawlessly at -O0 something is definitely wrong.  If test pass flawlessly at -O0 then the code is "correct" as far as the C++ language goes (with an ideal compiler). If the tests pass flawlessly at -O3 something may still be wrong ie data written or read out of bounds. If tests pass at -O0 but not at -O3 you're probably looking at some kind of compiler bug (although I guess openmp SIMD directives don't actually do anything below -O2 in most compilers so it could still be our codes fault.)

PR #4069, opened 2022-06-20
Add orbital rotation test with legacy driver

markdewing:
The gen_rotated_lcao_wf.py file is extended to generate QMC averages for the parameter gradients.

A test for the legacy driver using one thread is introduced.  The same rotation is used for both up and down electrons to check the parameter gradients are the same.

The cost parameters in the input file are important (energy to 1.0 and reweightedvariance to 0.0) because the code defaults to an energy fraction of 0.9 and a reweightedvariance fraction of 0.1.  If these are left at the default, the rotation parameter gradients are around -0.29.

The error bars on the reference values were taken from a QMCPACK run 100x longer (and divided by 10).

This is an attempt to define one point where we know the code works and has a validated answer.

The test uses the legacy driver and only one thread.  It will fail with multiple threads. The batched driver fails.

- No

desktop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/tests/gen_rotated_lcao_wf.py
+++ b/src/QMCWaveFunctions/tests/gen_rotated_lcao_wf.py
@@ -5,8 +5,11 @@
 #  compute spatial and parameter derivatives
 import autograd.numpy as np
 from autograd import hessian,grad
+from stats import averager
 
-# Values used in test_RotatedSPOs_LCAO.cpp
+# Point values used in test_RotatedSPOs_LCAO.cpp
+# QMC values used to validate tests/molecules/He_param/He_orb_rot_param_grad_legacy
+
 
 def mag(r):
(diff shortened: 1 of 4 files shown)


markdewing:
> > The cost parameters in the input file are important (energy to 1.0 and reweightedvariance to 0.0) because the code defaults to an energy fraction of 0.9 and a reweightedvariance fraction of 0.1. If these are left at the default, the rotation parameter gradients are around -0.29.
> 
> Is this because GradCost that gradient_test relies on evaluates the analytic gradients of the Hamiltonian but the finite difference is done with the cost function instead of the Hamiltonian?

No, the cost parameters are taken into account in both the cost function and the gradient cost function - the gradient test output shows the analytic and finite difference gradients are consistent.  But that value doesn't agree with the value generated by the script (-0.22).  To be consistent with the script, the cost parameters need to be set different from the default values.

ye-luo (approved):

prckent (approved):

PR #4074, opened 2022-06-23
Remove omp parallel over walkers

ye-luo:
## Proposed changes
The initial thoughts was to leverage nested threading which we don't used most of the time since we rely on crowds over first level threads.
Consider the benefit is not evident and there are potentially better alternatives (CPU tasking). I'd like to remove them over walker loops.

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/DistanceTable.h
+++ b/src/Particle/DistanceTable.h
@@ -103,7 +103,6 @@ public:
   virtual void mw_evaluate(const RefVectorWithLeader<DistanceTable>& dt_list,
                            const RefVectorWithLeader<ParticleSet>& p_list) const
   {
-#pragma omp parallel for
     for (int iw = 0; iw < dt_list.size(); iw++)
       dt_list[iw].evaluate(p_list[iw]);
   }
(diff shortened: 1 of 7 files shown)


ye-luo:
I was thinking of Ncrowds<Nthreads. Then using `omp taskloop` to send tasks to ideal threads.

ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
> OMP parallel creates implicit thread scopes the safety of which depends on the nature of the variables partitioned, copied or shared in those scopes. This has many negative and invisible side effects on design, making it seems like every thing in the for loop needs its own resource because there could be from 1 to number loop elements thread scopes.

Indeed, OpenMP spec says "omp parallel" create one implicit task per thread. So it is all about tasking but users doesn't need to know that details. It seems like thread scope. It is actually the scope of the implicit tasks that code relies on.

> But I don't think we're going to get to explore the Ncrowds<Nthreads case since in the known future we are probably going to have relatively few CPU threads per rank.

Most of the machines have <=8 cores per GPU. Only perlmutter node has 16 per GPU. Unless `Ncrowds<Nthreads` really gains a lot, we need to care but so far we don't need to.

I think for us, only "crowd" scope matters and we don't really need to be bothered by threads directly.

PR #4075, opened 2022-06-24
Interfacing Optimizers with Batched Drivers

LeonOtis:
## Proposed changes

This is the first broken off part of PR #4061 . Thank you to everyone for your feedback on my original set of changes. This smaller PR is meant to focus on just the changes needed to run the optimizers (adaptive three shift LM, accelerated descent, and hybrid method) with the batched drivers without further improvements to the optimizers themselves.

I created a struct engineData to hold optimizer engine objects to be passed to checkConfigurations. This avoids duplicating code between checkConfigurations and engine_checkConfigurations. I am imagining that engine_checkConfigurations can eventually be removed whenever the legacy drivers are deleted. 

The other main changes are in QMCFixedSampleLinearOptimizeBatched. I use copies of QMCDriverInput and VMCDriverInput to read the selected xml blocks needed for the hybrid method to vary the number of samples used. As it seems the vmcEngine is intended to be run only once during an optimization iteration, I also modified the adaptive three shift LM to make its correlated sampling evaluations without generating another set of samples. This avoids some of the problematic changes I tried in #4061 and I don't think it will have much effect on the optimizer's performance in practice, but it means it will not exactly match the legacy version's behavior.

I think the optimizers should function with both MPI and OMP parallelism with these interface changes. I have also added a basic unit test for creating an engineData struct. I am happy to make more improvements and thanks again for all the feedback.

- New feature

- No

Lawrencium computing cluster

## Checklist


src/QMCDrivers/WFOpt/EngineData.h
```diff
@@ -0,0 +1,32 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2019 QMCPACK developers.
```
markdewing, line 5: Minor issue - update the copyright year to 2022.

LeonOtis: Thanks, updated now.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -360,14 +376,44 @@ void QMCCostFunctionBatched::checkConfigurations()
 
         for (int ib = 0; ib < current_batch_size; ib++)
         {
+            #ifdef HAVE_LMY_ENGINE
```
PDoakORNL, line 379: more ifdefs, this code smells and bad. 

It looks like I do this extra allocation and setup even if I'm not using the LMY engine as long as the compile switch is on.  Ifdef's should be temporary stopgap solutions and not a way to shove low quality code into the codebase. I'm also no longer sure where this method started or what exactly we are trying to accomplish.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -360,14 +376,44 @@ void QMCCostFunctionBatched::checkConfigurations()
...
+           if (data.method == "adaptive")
+           {
+              //Engines take 1.0 for weights until modified guiding function can be implemented  
+              // pass into engine
+              data.lmEngine->take_sample(der_rat_samp, le_der_samp, le_der_samp, 1.0, 1.0);
+            }
+           else if (data.method == "descent")
+            {
+                int ip = omp_get_thread_num();
+                data.descentEngine->takeSample(ip, der_rat_samp, le_der_samp, le_der_samp, 1.0, 1.0);
```
PDoakORNL, line 413: maybe more missing coverage.  Looks non trivial.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -436,6 +482,27 @@ void QMCCostFunctionBatched::checkConfigurations()
...
+  // engine finish taking samples
+  if (data.method == "adaptive")
+  {
+     //Only call sample_finish now if samples are not being stored
+    data.lmEngine->sample_finish();
+
+    if (data.lmEngine->block_first())
+    {
+      OptVariablesForPsi.setComputed();
+      app_log() << "calling setComputed function" << std::endl;
```
PDoakORNL, line 496: Pointless output.  If its really helpful for debug I doubt its a permanent addition. If it is, this is a place for to use the preprocessor
```c++
#ifndef NDEBUG
// check concurrency environment if necessary
std::cout << "returned from setComputed\n";
#endif
```

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -225,9 +225,25 @@ void compute_batch_parameters(int sample_size, int batch_size, int& num_batches,
   }
 }
 
-/** evaluate everything before optimization */
+//Blank function, can be removed after this function is removed in the base class
 void QMCCostFunctionBatched::checkConfigurations()
+{}
+
+/** evaluate everything before optimization */
+void QMCCostFunctionBatched::checkConfigurations(engineData& data)
```
PDoakORNL, line 233: This method is huge and `data` is a super generic name that I would assume point to the major data structor or container to be operated on by the function. 

Here data is actually two raw pointers and a string that should be an enum class. There are at least 3 different method here that are branched around based on matches to that string and then unchecked uses of the raw pointers.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
```diff
@@ -221,7 +224,7 @@ void QMCFixedSampleLinearOptimizeBatched::start()
   initialize_timer_.start();
   optTarget->getConfigurations("");
   optTarget->setRng(vmcEngine->getRngRefs());
-  optTarget->checkConfigurations();
+  optTarget->checkConfigurations(forCostFunc);
```
PDoakORNL, line 227: hmmm the forCostFunction seems to just contain an unknown string and two nullptrs.  I remember tons of unchecked access through those two pointers above.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
```diff
@@ -238,6 +241,11 @@ void QMCFixedSampleLinearOptimizeBatched::engine_start(cqmc::engine::LMYEngine<V
 {
   app_log() << "entering engine_start function" << std::endl;
 
+  engineData forCostFunc;
```
PDoakORNL, line 244: engineData is a misnamed member, please see [Naming](https://qmcpack.readthedocs.io/en/develop/developing.html#naming). Why is it initialized at engine start, and from here on I guess we won't know whether a method might require engine access.

I Don't like this pattern.  If QMCFixedSampleLinearOptimizeBatched composes the engines they should be present and valid as early in the life cycle as possible and the way they are composed into the object should give a clue to ownership and their lifetime.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
```diff
@@ -732,8 +767,21 @@ bool QMCFixedSampleLinearOptimizeBatched::processOptXML(xmlNodePtr opt_xml,
   vmcEngine.reset(nullptr);
 
   // Explicitly copy the driver input objects since they will be used to instantiate the VMCEngine repeatedly.
+ //Overwriting input information is also done here to account for the hybrid method 
   QMCDriverInput qmcdriver_input_copy = qmcdriver_input_;
   VMCDriverInput vmcdriver_input_copy = vmcdriver_input_;
+  
+  if(MinMethod == "hybrid")
+  {
+  qmcdriver_input_copy.readXML(hybridEngineObj->getSelectedXML());
```
PDoakORNL, line 776: This is some sort of hack, not in the spirit of the input code, and processOptXML is just a put method but even less clear.

PDoakORNL (changes requested):
@LeonOtis. 
I know that the current LMYEngine code is low quality, and that is too bad, but it should at least be incrementally improved as it is ported, its naming and style conventions should not be followed for new code that is added. And I am feeling that the quality of the batched optimizer code really needs to be addressed as well.

I'm sorry you are having to struggle with it.  I don't think that the goals for the batched version/epoch of the code have been clearly expressed to you.  I think too many of them have probably only been captured in meetings and MS reports and not the developer guide. I think it is very important especially in the Driver code that we reconsider, refactor, and whenever necessary redesign clearly rushed and poorly designed code we are porting.

I think the time is past that input and construction to a validity for the batched optimizers is straightened out and we should not bury that technical debt any deeper. Nor add code that takes what is going on there as a pattern to follow.

Many improvements can be made in this code and I hope to see a form of it at some time in the future but I can't see myself being willing to approce this PR or desiring to take collective responsibility for the code it introduces into the batched epoch/version of the code.

I will begin reviewing again the exiting batched optimizer code.  I think we should have some focused meetings to discuss what we can do to correct the course of the batched optimization code.

ye-luo:
Just talked to @LeonOtis He will keep refactoring checkConfigurations to make it with less hacks. About how to handle the input. More thinking is still needed due to the overall batched driver input design changes. We need to figure out how to accommodates the need from an optimizer like the hybrid engine optimizer

LeonOtis:
The new commits use a EngineHandle class to interface between the batched drivers and the optimizers. This avoids the repeated use of #ifdef in checkConfigurations that I tried previously.

src/QMCDrivers/WFOpt/EngineHandle.h
```diff
@@ -0,0 +1,139 @@
...
+  using Value         = QMCTraits::ValueType;
+  using FullPrecReal  = QMCTraits::FullPrecRealType;
+  using FullPrecValue = QMCTraits::FullPrecValueType;
+
+  virtual void prepareSampling(int num_params) = 0;
+  virtual void takeSample(const std::vector<FullPrecReal>& energy_list,
+                          const RecordArray<Value>& dlogpsi_array,
+                          const RecordArray<Value>& dhpsioverpsi_array,
+                          int ib)              = 0;
+  virtual void finishSampling()                = 0;
```
ye-luo, line 41: Please document all three APIs.

ye-luo (commented):
@LeonOtis I fixed the compilation issues. And requested some documentation. I do noticed that some lines are commented. Could you do a review of all the commented code. If they will be needed in your next PR, please add TODO notes. Otherwise, it is better to just delete them. Adding our thoughts/comments are definitely appreciated.

src/QMCDrivers/WFOpt/EngineHandle.h
LeonOtis: Thanks @ye-luo for fixing the builds. I've added some documentation and cleaned up some commented code, but let me know if there's more that should be changed. My intent for the next PR after this one with the interface is done is to add the remaining changes in #4061 for the optimizers themselves. Changes would be inside the LM engine and QMCFixedSampleLinearOptimizeBatched.

ye-luo:
Test this please

ye-luo (approved):
Coverage is low but it is an issue of the overall WFopt.

prckent (approved):
I think this is much improved.

Looking forward to seeing the continuation of this series of updates.

ye-luo:
@LeonOtis do you know why short-H4-hybrid test energy checks fail since this PR?

LeonOtis:
@ye-luo It looks like I forgot a one line change to the legacy driver that's needed to keep the hybrid method working with this PR. Without it, the hybrid method won't switch from descent to LM and doesn't lower the energy enough to pass the test. Thanks for catching this. I made the fix in PR #4126, which I think should be close to done? When that PR is approved, this should be fixed.

Issue #3983, opened 2022-04-27
Orbital rotation issues


markdewing:
A collection of issues and narrative saga related to orbital rotation.

### Related issues
See #3564 on getting the variational parameter indices correct (still not fixed).  #2023 is vague (no reproducer).   #637 relates to using a spline basis.

Trying to get orbital rotation correct has sent me down a long yak shaving chain.  Trying to fix one issue leads to other issues. 

I tried to start submitting fixes in #3977, but it's not clear that code is even correct, and there are no good tests to validate it.

### Testing
The existing tests (in https://github.com/QMCPACK/qmcpack/tree/develop/tests/molecules/H2_ae and https://github.com/QMCPACK/qmcpack/tree/develop/tests/molecules/H4_ae ) target excited states and require BUILD_LMYENGINE_INTERFACE to be enabled.  They also perform complete optimization.

Tests are needed that simplify it further. The first step is getting parameter derivatives correct.
For an extremely simple test case.  My favorite simple case is an He atom with STO orbitals and a Pade Jastrow.
The wavefunction excerpt:
```
   <jastrow name="Jee" type="Two-Body" function="pade">
      <correlation speciesA="u" speciesB="d">
        <var id="jud_b" name="B">0.1</var>
      </correlation>
    </jastrow>

    <sposet_collection type="MolecularOrbital">
      <!-- Use a single Slater Type Orbital (STO) for the basis. Cusp condition is correct. -->
      <basisset keyword="STO" transform="no">
        <atomicBasisSet type="STO" elementType="He">
          <basisGroup rid="R0" n="1" l="0" m="0" type="Slater">
             <radfunc exponent="2.0"/>
          </basisGroup>
          <basisGroup rid="R1" n="2" l="0" m="0" type="Slater">
             <radfunc exponent="1.0"/>
          </basisGroup>
        </atomicBasisSet>
      </basisset>
      <sposet basisset="LCAOBSet" name="spo-up" size="2" optimize="yes">
          <coefficient id="updetC" type="Array" size="2">
            1.0 0.0
            0.0 1.0
          </coefficient>
      </sposet>
    </sposet_collection>
    <determinantset type="MO" key="STO" transform="no" source="ion0">
      <slaterdeterminant>
        <determinant id="spo-up" spin="1" size="2"/>
        <determinant id="spo-up" spin="-1" size="2"/>
      </slaterdeterminant>
    </determinantset>
  </wavefunction>
```

[...]

PR #4076, opened 2022-06-24
Orbital rotation fixes for legacy driver with threads

markdewing:
These are fixes for orbital rotation with the legacy driver and multiple threads.

See #3983 for an overarching issue, and #3977 for a previous attempt (this PR is similar).  The use_stored_copy flag is no longer used, but I avoided removing it to keep the PR simple.

The change in QMCCostFunctionBase.cpp is unrelated, but small (setting the default value of do_override_output) (I know, violating my own rule about keep PR's focused on a single change.  But it is small.)

A multi-threaded version of the test was added.  The input was changed to use samplesperthread because otherwise there were too few samples per thread, and it could trigger the variance check.

The core issue is how, when, and where to keep a copy of the original coefficient matrix ("SPO internal state") for subsequent application of rotations.

### Design space
PR #3977 has some discussion about the design.
@PDoakORNL mentioned keeping a copy of the SPOSet in https://github.com/QMCPACK/qmcpack/pull/3977#issuecomment-1110289861
However, the coefficient matrix is kept in a shared pointer.  Copying the object only copies the pointer (which saves memory when copying the object), but is not the behavior we want to keep a clean copy of the coefficient matrix.   Four possible solutions:
1. An API so the SPO keeps it's own copy of the SPO internal state (this is the existing API)
2. Implement deep copy in the SPO.
3. An API that the SPO can hand over a copy the SPO internal state, and create the SPO using that state.
4. Separate object for the SPO internal state that gets set up initially, and the SPO gets created later in the process.

With 1-3, there is a further problem, related to the sequence of creation, set up, and copying.
We need to pick some point at which to save the SPO internal state, and we need to make sure we're not getting the state from a modified SPO.
Essentially, we want to get to the same object state regardless of path (creation + set up or copying a fully set-up object). (Somewhat like making an object idempotent to certain operations)

Reducing the code to a single path to get the object state is option 4. The downside of 4 is it requires a lot of changes to the initialization code.

### Rationale for choices in this PR
1. Adding a state variable to the RotatedSPO class is perhaps not the ideal choice.  However, it does seem like the best local fix for now.  
2. Given there are multiple changes necessary, it would be nice to get **something** working along with tests.  That makes it easier to change in the future.

### Other changes
- stop setting the parameters to zero in checkInVariables
- set the myVars local-to-global mapping indices on wavefunction clones via checkOutVariables. If not set correctly, the parameter derivatives in RotatedSPOs::evaluateDerivatives do not get written to the correct indices in dlogpsi and dhpsioverpsi near the end of the function. myVars does get copied via cloning, but the original doesn't get checkOutVariables called on it before the cloning happens.  So we need to make sure it does get set.  Note that if it's only set in resetParameters, the code fails on the first round of the optimizer.

- Bugfix

- No

desktop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCCostFunction.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunction.cpp
@@ -168,7 +168,12 @@ void QMCCostFunction::GradCost(std::vector<Return_rt>& PGradient,
 
 void QMCCostFunction::getConfigurations(const std::string& aroot)
 {
-  //makeClones(W,Psi,H);
+  // Set all the myVars indices in the wavefunction clones
+  for (int i = 0; i < psiClones.size(); ++i)
+  {
+    psiClones[i]->checkOutVariables(OptVariablesForPsi);
+  }
+
   if (H_KE_Node.empty())
   {
     app_log() << "  QMCCostFunction is created with " << NumThreads << " threads." << std::endl;
@@ -331,7 +336,7 @@ void QMCCostFunction::checkConfigurations()
   app_log().flush();
   setTargetEnergy(Etarget);
   ReportCounter = 0;
-  IsValid = true;
+  IsValid       = true;
   //collect SumValue for computedCost
   SumValue[SUM_WGT]       = etemp[1];
   SumValue[SUM_WGTSQ]     = etemp[1];
@@ -521,7 +526,10 @@ void QMCCostFunction::resetPsi(bool final_reset)
   //cout << "-------------------------------------- " << std::endl;
   Psi.resetParameters(OptVariablesForPsi);
   for (int i = 0; i < psiClones.size(); ++i)
+  {
+    psiClones[i]->checkOutVariables(OptVariablesForPsi);
     psiClones[i]->resetParameters(OptVariablesForPsi);
+  }
 }
 
 QMCCostFunction::EffectiveWeight QMCCostFunction::correlatedSampling(bool needGrad)
--- a/src/QMCWaveFunctions/RotatedSPOs.cpp
+++ b/src/QMCWaveFunctions/RotatedSPOs.cpp
@@ -18,7 +18,7 @@
 
 namespace qmcplusplus
 {
-RotatedSPOs::RotatedSPOs(std::unique_ptr<SPOSet>&& spos)
+RotatedSPOs::RotatedSPOs(std::unique_ptr<SPOSet>&& spos, bool creatingCopy)
     : SPOSet(spos->isOMPoffload(), spos->hasIonDerivs(), true),
       Phi(std::move(spos)),
       nel_major_(0),
@@ -26,6 +26,9 @@ RotatedSPOs::RotatedSPOs(std::unique_ptr<SPOSet>&& spos)
 {
   className      = "RotatedSPOs";
   OrbitalSetSize = Phi->getOrbitalSetSize();
+  // Ensure that a copy of the SPO data is only made once
+  if (!creatingCopy)
+    Phi->storeParamsBeforeRotation();
 }
 
 RotatedSPOs::~RotatedSPOs() {}
@@ -113,7 +116,8 @@ void RotatedSPOs::buildOptVariables(const std::vector<std::pair<int, int>>& rota
   std::vector<RealType> param(m_act_rot_inds.size());
   for (int i = 0; i < m_act_rot_inds.size(); i++)
     param[i] = myVars[i];
-  apply_rotation(param, false);
+
+  apply_rotation(param, true);
 
   if (!Optimizable)
   {
@@ -1047,7 +1051,8 @@ void RotatedSPOs::table_method_evalWF(std::vector<ValueType>& dlogpsi,
 
 std::unique_ptr<SPOSet> RotatedSPOs::makeClone() const
 {
-  auto myclone = std::make_unique<RotatedSPOs>(std::unique_ptr<SPOSet>(Phi->makeClone()));
+  bool creatingCopy = true;
+  auto myclone      = std::make_unique<RotatedSPOs>(std::unique_ptr<SPOSet>(Phi->makeClone()), creatingCopy);
 
   myclone->params          = this->params;
   myclone->params_supplied = this->params_supplied;
--- a/src/QMCWaveFunctions/RotatedSPOs.h
+++ b/src/QMCWaveFunctions/RotatedSPOs.h
@@ -22,7 +22,7 @@ class RotatedSPOs : public SPOSet
 {
 public:
   //constructor
-  RotatedSPOs(std::unique_ptr<SPOSet>&& spos);
+  RotatedSPOs(std::unique_ptr<SPOSet>&& spos, bool creatingCopy = false);
   //destructor
   ~RotatedSPOs() override;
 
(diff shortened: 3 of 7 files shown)


src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -18,14 +18,17 @@
...
+RotatedSPOs::RotatedSPOs(std::unique_ptr<SPOSet>&& spos, bool creatingCopy)
     : SPOSet(spos->isOMPoffload(), spos->hasIonDerivs(), true),
       Phi(std::move(spos)),
       nel_major_(0),
       params_supplied(false)
 {
   className      = "RotatedSPOs";
   OrbitalSetSize = Phi->getOrbitalSetSize();
+  // Ensure that a copy of the SPO data is only made once
+  if (!creatingCopy)
```
PDoakORNL, line 30: There needs to be detailed documentation, its quite cryptic.  I guess that sometimes the caller of this constructor might have another copy of the object with the unique_ptr and that other times they don't so they going to invoke the maybe noop storeParams call.  It's difficult to reason about, and it seems like it should have to be at this layer at all.

What I really think is:

Several things are going wrong here.  One is that RotatedSpoSet wants to both compose and inherit from SPOSet.  This is a bad smell and probably means not just this class is unclear but so is the base class. May there is a data structure and a call interface here they probably don't both need to be type erased. Until that is cleared up the rest is going to be pretty confusing.  But I'm going to assume that needs to be a different PR.

It also seems like some versions of the derived SPOSet that is composed can storeParams and others can't / don't.  This makes reasoning about this code unhappy. We are managing a state machine that only exists in one of the cases.  Calls to do this have polluted the base class API, flags for this appear in the composing class.  Something needs to be seriously refactored or redesigned here.

src/QMCWaveFunctions/RotatedSPOs.h
```diff
@@ -165,15 +165,10 @@ class RotatedSPOs : public SPOSet
 
   void checkInVariables(opt_variables_type& active) override
   {
-    //reset parameters to zero after coefficient matrix has been updated
-    for (int k = 0; k < myVars.size(); ++k)
-      myVars[k] = 0.0;
-
     if (Optimizable)
     {
       if (myVars.size())
```
PDoakORNL, line 170: This function needs documentation. You're going to have to write it.  What is it doing?  Explain why its now ok that it doesn't call storeParams (does it also get called only once in the state cycle and that's before the RotatedSPOSet constructor, that seems impossible.

PDoakORNL (changes requested):
I spent an hour and a half reading through the code that might be impacted by this, and honestly I have no idea.  Write some docs in the source to make this change seem ok, please make the actual code I was just lost in make some sense, maybe it just needs docs, but I think it actually just needs a redesign.

src/QMCDrivers/WFOpt/QMCCostFunction.cpp
```diff
@@ -521,7 +526,10 @@ void QMCCostFunction::resetPsi(bool final_reset)
   //cout << "-------------------------------------- " << std::endl;
   Psi.resetParameters(OptVariablesForPsi);
   for (int i = 0; i < psiClones.size(); ++i)
+  {
+    psiClones[i]->checkOutVariables(OptVariablesForPsi);
```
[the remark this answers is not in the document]
markdewing: The overall documentation for optimization parameter functions is in TrialWaveFunction.h
https://github.com/QMCPACK/qmcpack/blob/7d8618dd7e70b6899432c173562c29a9e51e97a0/src/QMCWaveFunctions/TrialWaveFunction.h#L135-L167

src/QMCWaveFunctions/RotatedSPOs.cpp
markdewing: I don't think inheriting from SPOSet and storing an SPOset is the problem.  Conceptually, orbital rotation is modifying an existing set of SPO's, so modeling it that way in the code reflects the problem domain.  The variable Phi stores the original SPOSet.
What might help is creating another class that derives from SPOSet and introduces applyRotations and related code.  Maybe call it ModifiableSPOSet (or RotatableSPOSet).  And then the SPOSet's that support rotation (LCAO and Splines) can derive from that.

src/QMCDrivers/WFOpt/QMCCostFunction.cpp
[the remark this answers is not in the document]
PDoakORNL: There is no way for someone coming to the code the first time to know that.  RotatedSPOSet should make sense and have clearly defined behavior without having to read documentation belonging several layers out.

markdewing:
I think i found another solution - calling storeParamsBeforeRotation in SPOSetBuilder.  WIll start a fresh PR.

src/QMCWaveFunctions/RotatedSPOs.cpp
PDoakORNL: What might be more clear is just have a function that applies the rotation and returns a new SPOSet that has been rotated.

Isn't this to preserve a weird implementation detail where the optimizer always calculates the total rotation from the original SPOSet?  Getting what is essentially a new set of Coefficients returned means you can make sure that surprising behavior is know where it might needed to reason about why the rotation at each application is what it is. 
```c++
 SPOSet RotateOrbitals(const SPOSet& sposet)
```
then later I read code like this
```c++
sposet_opt = OptimizingRotator.rotate(rotation, sposet_orig)
```
```c++
//or if your update of the rotation as I would expect
sposet_opt =  OptimizingRotator.rotate(delta_rotation, sposet_opt)
```

My understanding is that SPOSet is too heavy for this to be feasible right now, which means it really is a SPOSet and too much other stuff. It's sad to bake this state machine further into the SPOSet infrastructure and reinforce an API member that shouldn't be in SPOSet.

A RotatedSPOSet doesn't have a is-a relationship to an SPOSet. It requires the definition of an SPOSet to be changed to have an API member `storeParamsBeforeRotation` that is part of being a "SPOSet" but actually is a noop in all cases (7+BsplineSet derived classes) except for LCAOOrbitalSet.

Even if we are going to have such an awkward object shouldn't it have a const reference to the original SPOSet and the rotation that has been applied.  And then for optimization you could add habe it be derived from the derived type of the original SPOSet and operate on its own coefficients instead of making an SPOSet capable of holding two sets of coefficient data occassionaly for some derived types.

markdewing: BsplineSet needs to implement storeParamsBeforeRotation if optimizing orbital rotations is going to work on splines.

I tried implementing an SPOSetRotatable class that derives from SPOSet and adds storeParamsBeforeRotation and applyRotation.  The idea is that only SPO sets that implement rotation would derive from SPOSetRotatable (so LCAOrbitalSet and BsplineSet).  And RotatedSPOs would take an SPOSetRotatable object as input (instead of the current SPOSet).   I couldn't get it to compile (issues related to dynamic casting SPOSet to SPOSetRotatable in setup).

At the most concrete level, some piece of code needs to multiply the rotation matrix by the coefficient matrix.  It makes sense to put this at the LCAOrbitalSet level. Or some other piece of code will need to know about LCAOrbitalSet internals to do that multiplication.

Applying a delta rotation introduces more effective state into an object - the number of times the delta rotation has been applied (and the particular values of each rotation)

ye-luo: RotatedSPO should carry some states "delta rotation" and "total rotation"(not there yet) but it should not know what is the actual underlying SPOSet. However, it also needs some help from the SPOSet. To make "total rotation" meaningful, SPOSet is responsible for the original coefficients.

I have doubts in the necessity of SPOSetRotatable intermediate base class.

PR #4059, opened 2022-06-13
Add tests for Rotated SPOs using LCAO

markdewing:
The test system is a helium atom with two orbitals and an optional Jastrow factor.

A python script is used to generate reference values for the values of the wavefunction and derivatives at one point in space.  The script uses automatic differentiation (via autograd) to make computing the spatial and parameter derivatives straightforward.

This is the starting point for some testing of orbital rotation that should make changes and fixes more reliable (see #3977 ). 

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Testing changes (e.g. new unit/integration/performance tests)

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
laptop

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- Yes. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- N/A. Documentation has been added (if appropriate)

PR #4083, opened 2022-06-28
Orbital rotation fixes for legacy driver and threads, again.

markdewing:
So #3977 and #4076 have failed to be an acceptable move forward in fixing orbital rotation.
Trying again...
The trick this time is to instruct the SPOSet to store the orbital rotations at setup time, after we know that orbital rotations are being used in the code, but before any cloning is done.
The copy constructor on the LCAOrbitalSet class is fixed so the pristine copy of the coefficient matrix gets propagated through all the subsequent clones. And the tricky state management goes away.

- Bugfix

- No

desktop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCCostFunction.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunction.cpp
@@ -168,7 +168,13 @@ void QMCCostFunction::GradCost(std::vector<Return_rt>& PGradient,
 
 void QMCCostFunction::getConfigurations(const std::string& aroot)
 {
-  //makeClones(W,Psi,H);
+  // Set all the myVars indices in the wavefunction clones.
+  // If not set, the parameter derivative pieces in evaluateDerivatives
+  // end up in the wrong indices in dlogpsi and dhpsioverpsi.
+  for (int i = 0; i < psiClones.size(); i++)
+    psiClones[i]->checkOutVariables(OptVariablesForPsi);
+
+
   if (H_KE_Node.empty())
   {
     app_log() << "  QMCCostFunction is created with " << NumThreads << " threads." << std::endl;
--- a/src/QMCWaveFunctions/RotatedSPOs.h
+++ b/src/QMCWaveFunctions/RotatedSPOs.h
@@ -165,15 +165,10 @@ public:
 
   void checkInVariables(opt_variables_type& active) override
   {
-    //reset parameters to zero after coefficient matrix has been updated
-    for (int k = 0; k < myVars.size(); ++k)
-      myVars[k] = 0.0;
-
     if (Optimizable)
     {
       if (myVars.size())
         active.insertFrom(myVars);
-      Phi->storeParamsBeforeRotation();
     }
   }
 
--- a/src/QMCWaveFunctions/SPOSetBuilder.cpp
+++ b/src/QMCWaveFunctions/SPOSetBuilder.cpp
@@ -95,6 +95,10 @@ std::unique_ptr<SPOSet> SPOSetBuilder::createSPOSet(xmlNodePtr cur)
     app_error() << "Orbital optimization via rotation doesn't support complex wavefunction yet.\n";
     abort();
 #else
+    // At this point in the code we know that orbital rotation is used, so the target SPO
+    // can make a copy of the coefficient matrix.
+    sposet->storeParamsBeforeRotation();
+
     // create sposet with rotation
     auto& sposet_ref = *sposet;
     auto rot_spo     = std::make_unique<RotatedSPOs>(std::move(sposet));
--- a/tests/molecules/He_param/He_orb_rot_param_grad_legacy.xml
+++ b/tests/molecules/He_param/He_orb_rot_param_grad_legacy.xml
@@ -90,7 +90,7 @@
       <parameter name="steps"> 10 </parameter>
       <parameter name="substeps"> 20 </parameter>
       <parameter name="timestep"> 0.5 </parameter>
-      <parameter name="samples"> 1000 </parameter>
+      <parameter name="samplesperthread"> 1000 </parameter>
       <cost name="energy">                   1.0 </cost>
       <cost name="reweightedvariance">       0.00 </cost>
     </qmc>
--- a/src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
+++ b/src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
@@ -30,7 +30,12 @@ LCAOrbitalSet::LCAOrbitalSet(std::unique_ptr<basis_type>&& bs, bool optimize)
 }
 
 LCAOrbitalSet::LCAOrbitalSet(const LCAOrbitalSet& in)
-    : SPOSet(in), myBasisSet(in.myBasisSet->makeClone()), C(in.C), BasisSetSize(in.BasisSetSize), Identity(in.Identity)
+    : SPOSet(in),
+      myBasisSet(in.myBasisSet->makeClone()),
+      C(in.C),
+      BasisSetSize(in.BasisSetSize),
+      C_copy(in.C_copy),
+      Identity(in.Identity)
 {
   Temp.resize(BasisSetSize);
   Temph.resize(BasisSetSize);
--- a/src/QMCWaveFunctions/RotatedSPOs.cpp
+++ b/src/QMCWaveFunctions/RotatedSPOs.cpp
@@ -113,7 +113,7 @@ void RotatedSPOs::buildOptVariables(const std::vector<std::pair<int, int>>& rota
   std::vector<RealType> param(m_act_rot_inds.size());
   for (int i = 0; i < m_act_rot_inds.size(); i++)
     param[i] = myVars[i];
-  apply_rotation(param, false);
+  apply_rotation(param, true);
 
   if (!Optimizable)
   {
--- a/tests/molecules/He_param/CMakeLists.txt
+++ b/tests/molecules/He_param/CMakeLists.txt
@@ -103,6 +103,24 @@ if(NOT QMC_CUDA)
     SCALAR_VALUES
     HE_ORB_ROT_PARAM)
 
+  qmc_run_and_check_custom_scalar(
+    BASE_NAME
+    He_orb_rot_param_grad_legacy
+    BASE_DIR
+    "${qmcpack_SOURCE_DIR}/tests/molecules/He_param"
+    PREFIX
+    He_orb_rot_param_grad_legacy.param
+    INPUT_FILE
+    He_orb_rot_param_grad_legacy.xml
+    PROCS
+    1
+    THREADS
+    16
+    SERIES
+    0
+    SCALAR_VALUES
+    HE_ORB_ROT_PARAM)
+
 
   else()
     message(VERBOSE "Skipping He_param tests because parameter output is not supported by mixed precison build (QMC_MIXED_PRECISION=1)")


src/QMCWaveFunctions/SPOSetBuilder.cpp
```diff
@@ -95,6 +95,10 @@ std::unique_ptr<SPOSet> SPOSetBuilder::createSPOSet(xmlNodePtr cur)
     app_error() << "Orbital optimization via rotation doesn't support complex wavefunction yet.\n";
     abort();
 #else
+    // At this point in the code we know that orbital rotation is used, so the target SPO
+    // can make a copy of the coefficient matrix.
+    sposet->storeParamsBeforeRotation();
```
ye-luo, line 100: Can we move this line to the constructor of RoratedSPOs?

tests/molecules/He_param/He_orb_rot_param_grad_legacy.xml
```diff
@@ -90,7 +90,7 @@
       <parameter name="steps"> 10 </parameter>
       <parameter name="substeps"> 20 </parameter>
       <parameter name="timestep"> 0.5 </parameter>
-      <parameter name="samples"> 1000 </parameter>
+      <parameter name="samplesperthread"> 1000 </parameter>
```
ye-luo, line 93: Could you use the total number of samples and input it still as samples? samplesperthread was never a good input even with legacy drivers.

src/QMCWaveFunctions/SPOSetBuilder.cpp
markdewing: Then that line needs to be guarded under a flag as in #4076 to distinguish between the original creation and the making of copies.  The point of moving storeParamsBeforeRotation here is to remove the need for the creatingCopy flag.

ye-luo: Probably we need to make adjustment to the behavior makeClone, making sure it doesn't change the state of orbitals. Is this the reason that causes the threading bug?
@markdewing could you leave some comment about your understanding?
We can revisit the issue after this PR.

markdewing:
I feel stuck in a catch-22 situation here
 - the orbital rotation code needs some redesign, but we can't be confident in any redesign without tests
 - being confident in the tests requires the code to be fixed at least enough to agree with the tests

My primary goal of this PR and the previous one are to get the test cases into the code base.  The code fixes are a secondary consideration - the goal is code changes that are minimal and sufficient to pass the tests, rather than the right design.

ye-luo:
@markdewing I can feel your pain.  Fix threading with tests added and then fix the design seem pretty cumbersome especially with simulation tests. How about trying an alternative route by adding unit tests, fix the design and then fix the threading?

markdewing:
PR #4059 added unit tests for wavefunction values.
The problems come about from the interaction of the optimizer steps in the driver and RotatedSPOs.  So unit tests that catch these type of issues are going to have to call the drivers.  At that point, might as well use the simulation tests.

The problems with threading stem from cloning and saving the pristine coefficient matrix.  Redesign is largely to address that issue, and so testing the threading seems pretty integral to testing that the design fixes the threading issues.

src/QMCWaveFunctions/SPOSetBuilder.cpp
markdewing: The point of putting the storeParamsBeforeRotation here is then the SPOset is completely set up before any clones get made.  Then makeClone can make all the copies it wants and we don't have to worry about whether some of them are set up correctly (that is, have the pristine copy of the coefficient matrix initialized).

The other (so far unstated) requirement is that we know that the SPOSet is used for orbital rotation before making the copy of the coefficient matrix, in order to avoid wasting memory in cases that aren't going to use it.

ye-luo:
I mean bring the unit tests beyond individual calls to the call sequence of an optimization.
1. RotatedSPO trigger initialization
2. Set current rotation matrix parameter delta into Rotated SPO.
4. make the the rotation of the underlying SPOSet. Using current rotation
5. Set a second rotation matrix parameter delta into Rotated SPO.
6. make the the rotation of the underlying SPOSet

With this unit test, we change the behavior of SPOSet::storeParamsBeforeRotation/applyRotation APIs to rely on the original coeficients.
1. RotatedSPO trigger initialization
2. Set current rotation matrix parameter delta into Rotated SPO.
4. make the the rotation of the underlying SPOSet. Underlying using total rotation
5. Set a second rotation matrix parameter delta into Rotated SPO.
6. make the the rotation of the underlying SPOSet. Underlying Using total rotation

src/QMCWaveFunctions/SPOSetBuilder.cpp
ye-luo: > The point of putting the storeParamsBeforeRotation here is then the SPOset is completely set up before any clones get made. 

If we move storeParamsBeforeRotation() to the constructor of RotatedSPO, it is still get called before any clones get made. However, if makeClone was invoking that constructor or doing something funny in the copy constructor, things get messed up.
About wasting memory, we need to change C_copy to shared_ptr like C. Since only RotatedSPO constructor calls storeParamsBeforeRotation. C_copy won't be created when rotation is not enabled.

src/QMCWaveFunctions/RotatedSPOs.h
```diff
@@ -165,15 +165,10 @@ class RotatedSPOs : public SPOSet
...
   {
-    //reset parameters to zero after coefficient matrix has been updated
-    for (int k = 0; k < myVars.size(); ++k)
-      myVars[k] = 0.0;
-
     if (Optimizable)
     {
       if (myVars.size())
         active.insertFrom(myVars);
-      Phi->storeParamsBeforeRotation();
```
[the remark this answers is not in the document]
ye-luo: Correct.
> At each optimizer step, update the rotation parameters and apply the updated (total) rotation matrix to the original coefficients. The rotation parameters always reflect the total rotation.

I'm in favor of this way. The optimization step still generates a rotation matrix delta but RotatedSPO needs to combine it with the accumulated rotation and apply the combined rotation on to the original coefficients by applyRotation.

> At each optimizer step, save a copy of the previous coefficients and set the rotation parameters to zero. The rotation parameters are a delta update at each step. 

This is the way we have in the LCAO code and I would like to move away. applyRotation apply the rotation matrix delta onto the current coefficients.

For the final optimized coefficients, there should be no difference between these ways. However, keeping the "total rotation" instead of final orbitals has the advantages
1. smaller storage. We never want to destroy the orbitals we started with, they are on disk anyway. Then we only need to store the rotation matrix. In a spline case, rotation matrix is just 1k x 1k but storing the final spline coefficients can be GBs.
2. We can do more analysis directly with "total rotation".

So to move from 2 to 1. We need to adjust LCAOrbitalSet and SplineR2R internals.
This also requires updating I/O to make sure the global rotation gets properly handled on disks in the meanwhile, the optimizer only touches the delta rotation matrix.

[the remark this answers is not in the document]
markdewing: As far as I know, nothing needs to get updated (except the changes in this PR) for the optimizer to always work with the total rotation.
* LCAOrbitalSet::applyRotation applies whatever rotation you give it.  The changes in this PR are to ensure that it operates on the original coefficient matrix
* RotatedSPOs::apply_rotation (which calls LCAOrbitaSet::applyRotation) builds a fresh rotation matrix from the current values of the parameters that are passed in.
* RotatedSPOs::resetParameters takes the new values of the parameters passed to it and passes them to RotatedSPOs::apply_rotation

I don't see where a delta rotation enters in anywhere (that is, after this PR removes setting the parameters to zero in checkInVariables).   Treating the rotation parameters as a total rotation means they behave like any other parameters, and I don't think we need to do anything special to handle them.

src/QMCWaveFunctions/SPOSetBuilder.cpp
markdewing: The RotatedSPOs::makeClone function does call the RotatedSPOs constructor.

Changing C_copy to shared_ptr will most likely cause the rotation to be applied <number of threads> times.  There will need to be some mechanism to ensure that the rotation only gets applied once.  This is something to address in the future.

src/QMCWaveFunctions/RotatedSPOs.h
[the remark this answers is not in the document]
ye-luo: If parameters are not set to zero, should the RoratedSPO rotation parameter derivatives to be calculated with LCAO using the original coefficients instead of current coefficients?
Probably this is even better. RotatedSPO carries one original SPOSet for derivatives and another SPOSet with rotation for sampling. Then we don't need to fiddle with SPOSet internals.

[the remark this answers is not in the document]
markdewing: I don't think so.
The evaluateDerivatives function makes no reference to the rotation matrix, so the only way it knows the effects of the rotation parameters on the wavefunction is through the modified coefficient matrix in LCAOrbitalSet.
Also, the tests in #4059 and #4069 use a non-zero value for the rotation and they match the reference values which assume the rotation is applied.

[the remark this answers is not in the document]
ye-luo: If the derivatives are taken with respect to the total rotation parameters, I'm expecting the reference values to be different from current values which are derivatives of the delta rotation parameters.

markdewing:
The derivative of the rotation parameters is based on [Jacobi's formula](https://en.wikipedia.org/wiki/Jacobi%27s_formula) : d/dt det(A) = det(A) tr(A^1 dA/dt)    ( where t is one of the rotation parameters)
In the derivative dA/dt, the matrix A is a rotation matrix, which can be expressed as an exponential of a skew-symmetric matrix, A = exp(X).  The first order approximation to the matrix exponential is used (A = exp(X) ~= 1 + X), so dA = dX/dt.  The matrix X is linear in the rotation parameters (i.e., entries, are k_1,k_2, etc), so the derivative effectively creates a matrix that selects entries (all the entries are zero except the ones that involve that rotation parameter).  That selection is implemented in the last loop in RotatedSPOs::evaluateDerivatives.

More precisely, the matrix A is the product of the rotation matrix, the coefficient matrix, and the orbitals (exp(X) * C * phi).   In QMCPACK the rotation is always folded into the coefficient matrix, C' = exp(X)* C and treated as a constant (for the purposes of taking a derivative). So A = exp(Y) * C' * phi, where Y is the rotation matrix at zero.  The derivative for dA/dt is d/t (exp(Y) * C' * phi).  C' and phi have no dependence on the rotation parameter, so dA/dt = (d/dt exp(Y) ) * C' * phi. Note that since the derivative of the exponential is evaluated at Y=0, using a first order approximation is exact.

The main consequence is when comparing derivatives of psi (wrt rotation parameters) to analytic forms for the rotation matrix. The analytic expressions have to be set up carefully to ensure they match what QMCPACK is doing.

markdewing:
Okay, I think I finally understand the fundamental issue with orbital rotation - the rotation parameters cannot be combined via addition (i.e. treated like a vector).

The update of parameters assumes that the parameters can be added, as in the following code.  This assumption is broken with rotation parameters.
https://github.com/QMCPACK/qmcpack/blob/66743b34fd6d28fbcc8e4746eb7b170cc3be1aa8/src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimize.cpp#L1343-L1344

The current code rotation code works around this by, at each optimizer step, resetting the variables to zero (in checkInVariables) and applying the incremental rotation from that step (in resetParameters).  This leads to a few problems
1. The rotation part of the wavefunction depends on the history of calls to resetParameters.  (I would think that resetParameters sets the wavefunction to a known state that depends only on the input to resetParameters). 
2. There is no way to save the optimized rotation for later use, unless a sequence of values is stored (which is different than other types of variational parameters).
3. The calls to checkInVariables and resetParameters are used to indicate the start/stop of each optimizer step.  This is an additional requirement on the use of these calls from the optimizer.

I would prefer to always keep the total rotation, as that is compatible with existing methods to save and restore variational parameters.  To accommodate this, we need some API for the optimizer to hand off update of the parameters to the wavefunction components (pass the current parameters and the delta).  Most components would just add them.  The rotation component could exponentiate the parameters and apply the rotation (as the code does now), and then take matrix log to recover the rotation parameters.  Alternately, the Baker-Campbell-Hausdorff formula could be applied.

src/QMCDrivers/WFOpt/QMCCostFunction.cpp
```diff
@@ -168,7 +168,13 @@ void QMCCostFunction::GradCost(std::vector<Return_rt>& PGradient,
 
 void QMCCostFunction::getConfigurations(const std::string& aroot)
 {
-  //makeClones(W,Psi,H);
+  // Set all the myVars indices in the wavefunction clones.
+  // If not set, the parameter derivative pieces in evaluateDerivatives
+  // end up in the wrong indices in dlogpsi and dhpsioverpsi.
+  for (int i = 0; i < psiClones.size(); i++)
+    psiClones[i]->checkOutVariables(OptVariablesForPsi);
```
PDoakORNL, line 175: Trying to verify this is a good way to do this I fell right into having to reason about `QMCostFunction` `CloneManager` and `psiClones` multiplicity, in the amount of time I have I failed.
The naming is perfectly backwards for OptVariablesForPsi. I think there is a psiClones per "thread." I think they are just getting indexing set up and this isn't a checkout in the sense its used elsewhere in the code where psiClones now is using a memory resource or the like from OptVariableForPsi?

We're commited to just deleting this code ASAP correct?
Based on QMCCostFunctionBatched I assume so.

PDoakORNL (commented):
So I came to figure out if this was ready, make a comment about checkOutVariables method which may not be relevant (I think it isn't).  Still not sure what the status is.  I think @markdewing and @ye-luo have gotten this PR ready but  how can we get it in shape for an approving review?

src/QMCDrivers/WFOpt/QMCCostFunction.cpp
markdewing: checkOutVariables is called on Psi in QMCCostFunctionBase::put.  I don't know why that isn't sufficient to set the indices before makeClones is called in VMC::run.  I'll have to look into it.

markdewing: The DiracDeterminant constructor calls buildOptVariables.  RotatedSPO::buildOptVariables calls myVars.clear().  A better solution might be to make sure buildOptVariables only initializes the parameters once, rather than every time a clone is made.

markdewing: It appears that if I protect the body of RotatedSPOs::buildOptVariables against reinitialization, some of the other changes in this PR aren't necessary for the test to pass.

markdewing:
More minimal solution coming.

PR #4084, opened 2022-06-28
FreeOrbital SPOSet

Paul-St-Young:
## Proposed changes

This PR consolidates the ElectronGas folder into a single SPOSet object "FreeParticle".
This merges real and complex code paths into one file.

While mostly a refactor, this PR has the side effect of enabling simulation of the 2D electron gas, because it reuses `KContainer`, where `ndim` was recently introduced #3852.

I updated all the unit tests. One test needed to be dropped: "det_heg.rpa.rs5.xml" in the real build.
The dropped test uses twisted boundary condition, which the real code was never meant to support.

- Yes

Electron gas with twisted boundary condition can no longer be executed by the real code.
One must compile the complex code to use a twist.

workstation

## Checklist


ye-luo:
Test this please

ye-luo:
Replace ElectronGas with something without ifdef is long wanted.

prckent:
Thanks.

An easy to address point:

Why not retain the naming of ElectronGas or similar? Do you have a more general non-electron gas usage despite all the electron gas plumbing and formulae? 

Repurposing "Particle" here is not ideal because we already have things like ParticleSets which refer to very different things. We are trying to remove these sources of confusion from the codebase.

prckent:
I would like to see deterministic tests added for this code, with very good coverage obtained. This should be easy to achieve given the existing stochastic tests. (On my phone, so apologies if I missed them)

Paul-St-Young:
> Why not retain the naming of ElectronGas or similar?

Because the single-particle orbitals for the electron gas can be free, localized (Wigner crystal), or something else.
Do you prefer something like `ElectronGasFree`, `ElectronGasWigner` etc?

> I would like to see deterministic tests added for this code

Can you give more detail on the kind of tests you are looking for?
Do you mean different particle numbers, different system sizes, or something else?

Paul-St-Young:
Now I remember: the free orbitals can be used in a simulation with boson or boltzmann statistics.
That's why I did not want to add "ElectronGas" in front of "Free".

prckent:
Paul: the deterministic tests I was referring to use a fixed random seed and a very short run to enable fully deterministic and reproducibile results. Typically they are just a few moves, just enough to hit different code paths. These tests don't demonstrate correctness but do demonstrate consistency. We run them in the CI and recommend users run all the deterministic tests when installing qmcpack (ctest -L deterministic). They are the only way we have found to reliably keep features alive, e.g. ensuring that someone doesn't accidentally break this code . If we have a deterministic test then it will fail in CI and the attempt to break this code in future will fail. We also run sanitizers with the deterministic tests to catch memory leaks etc.

The standard way to make these is to adopt a short test. See e.g. https://github.com/QMCPACK/qmcpack/blob/develop/tests/solids/diamondC_1x1x1_pp/det_qmc_short.in.xml and https://github.com/QMCPACK/qmcpack/blob/develop/tests/solids/diamondC_1x1x1_pp/CMakeLists.txt .The former has

```
<?xml version="1.0"?>
<simulation>
   <project id="det_qmc_short" series="0">
      <application name="qmcapp" role="molecu" class="serial" version="1.0"/>
   </project>
   <random seed="71"/>

   <qmc method="vmc" move="pbyp">
      <estimator name="LocalEnergy" hdf5="no"/>
      <parameter name="walkers"             >    1               </parameter>
      <parameter name="blocks"              >    3               </parameter>
      <parameter name="steps"               >    3               </parameter>
      <parameter name="subSteps"            >    2               </parameter>
      <parameter name="timestep"            >    0.3             </parameter>
      <parameter name="warmupSteps"         >    3               </parameter>
   </qmc>
```

ye-luo:
How about FreeParticleOrbitals?

prckent:
Good news, coverage is actually still very decent :-) https://app.codecov.io/gh/QMCPACK/qmcpack/compare/4084/tree/src/QMCWaveFunctions/ElectronGas (You might need to authenticate with your GitHub ID). For simplicity I suggest we put this item on the list of things to verify _after_ this PR is in. The goal would be to craft an extra test or remove some unused legacy code to get the coverage >>80%. If there are vestigial functions from some old usage these should definitely go.

ye-luo:
I think we need to start short and deterministic tests with batched driver.

Paul-St-Young:
> How about FreeParticleOrbitals?

Can we further simplify it down to FreeOrbitals?

@ye-luo  I added short and deterministic tests of the SJ wavefunction using batched driver.
Running SJB with the batched driver causes a Seg. Fault at `SlaterDetWithBackflow::ratioGrad()`.
In practice this is OK because the backflow wavefunction should be ran with the all-particle driver for reasonable efficiency.
Is there plan to support all-particle moves in the batched driver in the near future?

@prckent Thanks for the detailed explanation of the deterministic tests!
Given that there are already quite a few of these tests for the HEG, do we need to add more in this PR?

prckent:
> > How about FreeParticleOrbitals?
> 
> Can we further simplify it down to FreeOrbitals?

My vote is for FreeOrbitals.

> @prckent Thanks for the detailed explanation of the deterministic tests! Given that there are already quite a few of these tests for the HEG, do we need to add more in this PR?

No :-)

ye-luo:
OK with FreeOrbitals as well.

Paul-St-Young:
name change enacted

src/Particle/LongRange/KContainer.h
```diff
@@ -81,7 +81,7 @@ class KContainer : public QMCTraits
    */
   void findApproxMMax(const ParticleLayout& lattice, unsigned ndim);
   /** construct the container for k-vectors */
-  void BuildKLists(const ParticleLayout& lattice, bool useSphere);
+  void BuildKLists(const ParticleLayout& lattice, PosType twist, bool useSphere);
```
ye-luo, line 84: Please document `twist` argument.
Also use `const PosType& twist`.

src/Particle/LongRange/LRBreakup.h
```diff
@@ -334,7 +334,8 @@ int LRBreakup<BreakupBasis>::SetupKVecs(mRealType kc, mRealType kcont, mRealType
 {
   //Add low |k| ( < kcont) k-points with exact degeneracy
   KContainer kexact;
-  kexact.updateKLists(Basis.get_Lattice(), kcont, Basis.get_Lattice().ndim);
+  PosType twist(0);
+  kexact.updateKLists(Basis.get_Lattice(), kcont, Basis.get_Lattice().ndim, twist);
```
ye-luo, line 338: With default value being added in updateKLists. This change can be dropped.

src/Particle/LongRange/StructFact.h
```diff
@@ -71,6 +71,9 @@ class StructFact : public QMCTraits
   /// accessor of StorePerParticle
   bool isStorePerParticle() const { return StorePerParticle; }
 
+  /// accessor of k_lists_
+  KContainer getKLists() const { return k_lists_; }
```
ye-luo, line 75: return type `const KContainer&`

src/Particle/SimulationCell.cpp
```diff
@@ -59,7 +59,8 @@ void SimulationCell::resetLRBox()
       app_log() << "--------------------------------------- " << std::endl;
     }
 
-    k_lists_.updateKLists(LRBox_, LRBox_.LR_kc, LRBox_.ndim);
+    QMCTraits::PosType twist(0);
+    k_lists_.updateKLists(LRBox_, LRBox_.LR_kc, LRBox_.ndim, twist);
```
ye-luo, line 63: We can avoid this change with a default value in updateKLists

ye-luo (commented):
Only minor changes needed before acceptance.

src/Particle/LongRange/KContainer.h
```diff
@@ -72,16 +72,21 @@ class KContainer : public QMCTraits
    * @param kc cutoff radius in the K
    * @param useSphere if true, use the |K|
    */
-  void updateKLists(const ParticleLayout& lattice, RealType kc, unsigned ndim, bool useSphere = true);
+  void updateKLists(const ParticleLayout& lattice,
+                    RealType kc,
+                    unsigned ndim,
+                    const PosType& twist = PosType(),
```
ye-luo, line 78: Still missing documentation for twist. Say something about how twist affects KLists.

src/Particle/LongRange/KContainer.h
ye-luo: @Paul-St-Young did you miss this one?

src/Particle/LongRange/StructFact.h
```diff
@@ -71,6 +70,9 @@ class StructFact : public QMCTraits
   /// accessor of StorePerParticle
   bool isStorePerParticle() const { return StorePerParticle; }
 
+  /// accessor of k_lists_
+  const KContainer getKLists() const { return k_lists_; }
```
ye-luo, line 74: Still incorrect. "const KContainer&"

ye-luo (commented):
Seems only a few requested change got missed. Very close to merge.

ye-luo:
Test this please

src/Particle/LongRange/tests/test_kcontainer.cpp
```diff
@@ -0,0 +1,194 @@
...
+    { 0,  2,  0},
+    { 2,  0,  0},
+  };
+
+  for (int ik=0;ik<mxk;ik++)
+  {
+    for (int ldim=0;ldim<ndim;ldim++)
+    {
+      CHECK(klists.kpts[ik][ldim] == gvecs[ik][ldim]);
+      CHECK(klists.kpts[ik][ldim]*blat == klists.kpts_cart[ik][ldim]);
```
ye-luo, line 147: CI failure are from missing Approx. Please double check all the CHECK. Also this file seems needing clang-format.

src/QMCWaveFunctions/ElectronGas/FreeOrbital.cpp
```diff
@@ -0,0 +1,265 @@
...
+        {
+          (hess[j1])(la, lb)      = -coskr * (kvecs[ik])[la] * (kvecs[ik])[lb];
+          (hess[j2])(la, lb)      = -sinkr * (kvecs[ik])[la] * (kvecs[ik])[lb];
+          (hess[j1])(lb, la)      = (hess[j1])(la, lb);
+          (hess[j2])(lb, la)      = (hess[j2])(la, lb);
+          ((ggg[j1])[la])(lb, la) = sinkr * (kvecs[ik])[la] * (kvecs[ik])[lb] * (kvecs[ik])[la];
+          ((ggg[j2])[la])(lb, la) = -coskr * (kvecs[ik])[la] * (kvecs[ik])[lb] * (kvecs[ik])[la];
+          ((ggg[j1])[la])(la, lb) = ((ggg[j1])[la])(lb, la);
+          ((ggg[j2])[la])(la, lb) = ((ggg[j2])[la])(lb, la);
+          ((ggg[j1])[lb])(la, la) = ((ggg[j1])[la])(lb, la);
```
ye-luo, line 219: My eyes hurt with excessive () around hess and ggg

src/QMCWaveFunctions/ElectronGas/FreeParticle.cpp
```diff
@@ -0,0 +1,277 @@
...
+      dp[j2] = coskr * kvecs[ik];
+      for (int la = 0; la < OHMMS_DIM; la++)
+      {
+        (hess[j1])(la, la)      = -coskr * (kvecs[ik])[la] * (kvecs[ik])[la]; 
+        (hess[j2])(la, la)      = -sinkr * (kvecs[ik])[la] * (kvecs[ik])[la]; 
+        ((ggg[j1])[la])(la, la) = sinkr * (kvecs[ik])[la] * (kvecs[ik])[la] * (kvecs[ik])[la];
+        ((ggg[j2])[la])(la, la) = -coskr * (kvecs[ik])[la] * (kvecs[ik])[la] * (kvecs[ik])[la];
+        for (int lb = la + 1; lb < OHMMS_DIM; lb++)
+        {
+          (hess[j1])(la, lb)      = -coskr * (kvecs[ik])[la] * (kvecs[ik])[lb];
```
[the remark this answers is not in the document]
Paul-St-Young: I agree this code is confusing. I wrote only the complex variant.
I consider the real variant an optimization at twist=0.
Given that it passes the same tests as the complex variant, it's probably correct.

Regarding the use of `OHMMS_DIM`, I want to switch it to `LRBreakupParameters::ndim` as soon as 2D tests are added.

ye-luo:
Test this please

ye-luo (dismissed):

Paul-St-Young:
@PDoakORNL unit tests for the KContainer have been added.
Please see if this is satisfactory now.

prckent (dismissed):
I think this is much improved.

Looking forward to seeing the continuation of this series of updates.

prckent (approved):
This now LGTM assuming that all the tests pass. @ye-luo should still sign off. Thanks for all the updates.

ye-luo (approved):

prckent:
Test this please

PR #3646, opened 2021-12-08
Use HIP as language in CMake

quantumsteve:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

Describe what this PR changes and why.  If it closes an issue, link to it here
with [a supported keyword](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

Fixes #3581

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Code style update (formatting, renaming)
- Refactoring (no functional changes, no api changes)
- Build related changes

### Does this introduce a breaking change?

- Yes (requires CMake 3.21 or later)

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- No. Documentation has been added (if appropriate)

PR #4085, opened 2022-06-29
Add missing dependency in CMake.

ye-luo:
## Proposed changes
qmcwfs_omptarget needs platform_LA
platform_LA also needs to be PUBLIC due to exposure via header files.

- Bugfix
- Build related changes

- No

epyc-server

## Checklist


ye-luo:
Test this please

ye-luo:
Showed up when updating #3646 but the fix needs to be in upstream.

PDoakORNL:
LGTM

PR #4086, opened 2022-06-30
Update OperatorBase and BareKineticEnergy operator with mw_evaluatePerParticle APIs

PDoakORNL:
## Proposed changes

This PR adds mw_evaluatePerParticle and mw_evaluatePerParticleWithToperator calls to the OperatorBase API.  It adds overridden versions of these methods to BareKineticEnergy and tests of them.

These are part of the development described in ECP Milestone 6.9.

I also suppress a bunch of unecessary output from the test_hamiltonian_ham tests that has no functional effect on the tests.
 
- New feature
- Build related changes
- Documentation changes

- No

Leconte

## Checklist

### DIFF
--- a/src/QMCHamiltonians/BareKineticEnergy.cpp
+++ b/src/QMCHamiltonians/BareKineticEnergy.cpp
@@ -25,7 +25,6 @@
 #include "Particle/MCWalkerConfiguration.h"
 #endif
 #include "type_traits/ConvertToReal.h"
-#include "Listener.hpp"
 
 namespace qmcplusplus
 {
--- a/src/QMCHamiltonians/BareKineticEnergy.cpp
+++ b/src/QMCHamiltonians/BareKineticEnergy.cpp
@@ -493,6 +493,15 @@ void BareKineticEnergy::mw_evaluatePerParticle(const RefVectorWithLeader<Operato
   }
 }
 
+void BareKineticEnergy::mw_evaluatePerParticleWithToperator(const RefVectorWithLeader<OperatorBase>& o_list,
+					   const RefVectorWithLeader<TrialWaveFunction>& wf_list,
+					   const RefVectorWithLeader<ParticleSet>& p_list,
+					   const std::vector<ListenerVector<RealType>>& listeners,
+					   const std::vector<ListenerVector<RealType>>& ion_listeners) const
+{
+  mw_evaluatePerParticle(o_list, wf_list, p_list, listeners, ion_listeners);
+}
+  
 #if !defined(REMOVE_TRACEMANAGER)
 Return_t BareKineticEnergy::evaluate_sp(ParticleSet& P)
 {
--- a/src/QMCHamiltonians/BareKineticEnergy.cpp
+++ b/src/QMCHamiltonians/BareKineticEnergy.cpp
@@ -493,15 +493,16 @@ void BareKineticEnergy::mw_evaluatePerParticle(const RefVectorWithLeader<Operato
   }
 }
 
-void BareKineticEnergy::mw_evaluatePerParticleWithToperator(const RefVectorWithLeader<OperatorBase>& o_list,
-					   const RefVectorWithLeader<TrialWaveFunction>& wf_list,
-					   const RefVectorWithLeader<ParticleSet>& p_list,
-					   const std::vector<ListenerVector<RealType>>& listeners,
-					   const std::vector<ListenerVector<RealType>>& ion_listeners) const
+void BareKineticEnergy::mw_evaluatePerParticleWithToperator(
+    const RefVectorWithLeader<OperatorBase>& o_list,
+    const RefVectorWithLeader<TrialWaveFunction>& wf_list,
+    const RefVectorWithLeader<ParticleSet>& p_list,
+    const std::vector<ListenerVector<RealType>>& listeners,
+    const std::vector<ListenerVector<RealType>>& ion_listeners) const
 {
   mw_evaluatePerParticle(o_list, wf_list, p_list, listeners, ion_listeners);
 }
-  
+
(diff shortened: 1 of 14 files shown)


src/QMCHamiltonians/BareKineticEnergy.cpp
```diff
@@ -402,6 +403,106 @@ void BareKineticEnergy::evaluateOneBodyOpMatrixForceDeriv(ParticleSet& P,
     }
 }
 
+void BareKineticEnergy::createResource(ResourceCollection& collection) const
+{
+  auto new_res        = std::make_unique<MultiWalkerResource>();
```
markdewing, line 408: Everywhere else in the code the "MultiWalkerResource" has a class name attached (e.g. CoulombPBCAAMultiWalkerResource.  Should this also have it?  Alternately, could the name be additionally qualified? (BareKineticEnergy::MultiWalkerResource).   
The issue is just reading through the code it looks like this is a generic class, but it is specific to BareKineticEnergy.

PDoakORNL: You're definitely right, I did a bunch of these and became less thoughtful. I'll stick with the current convention we've used so far. Fully qualifying is tempting but I'm not a big fan of Classes whose meaning changes based on name scopes inside of a project. Maybe we should consider an abbreviation as these types are obviously going to continue to appear.

src/QMCHamiltonians/tests/test_PairCorrEstimator.cpp
```diff
@@ -66,6 +66,8 @@ namespace qmcplusplus
 {
 TEST_CASE("Pair Correlation", "[hamiltonian]")
 {
+  outputManager.pause();
```
ye-luo, line 69: This change adds difficulty for troubleshooting issues.
We probably need to extend the catch_main.cpp to allow passing an option to make the printout through if the default is off. It should not be hard coded.

If you just need to work on one test, pass the test name as an argument.

Probably just revert such changes to keep this PR moving.

src/QMCHamiltonians/BareKineticEnergy.cpp
```diff
@@ -25,6 +25,7 @@
 #include "Particle/MCWalkerConfiguration.h"
 #endif
 #include "type_traits/ConvertToReal.h"
+#include "Listener.hpp"
```
ye-luo, line 28: Not needed?

src/QMCHamiltonians/BareKineticEnergy.h
```diff
@@ -18,9 +18,15 @@
...
 
 #include "QMCHamiltonians/OperatorBase.h"
+#include <ResourceCollection.h>
+#include <ResourceHandle.h>
 
 namespace qmcplusplus
 {
+
+template<typename T>
+class ListenerVector;
```
ye-luo, line 28: The full header file has been included in the base class. So why this is needed?

If the intention is to reduce header file inclusion. Then base class file can carry forward declaration instead of here.

src/QMCHamiltonians/BareKineticEnergy.h
```diff
@@ -138,6 +165,30 @@ class BareKineticEnergy : public OperatorBase
...
+  /** acquire a shared resource from a collection
+   */
+  void acquireResource(ResourceCollection& collection, const RefVectorWithLeader<OperatorBase>& o_list) const override;
+
+  /** return a shared resource to a collection
+   */
+  void releaseResource(ResourceCollection& collection, const RefVectorWithLeader<OperatorBase>& o_list) const override;
+
+private:
+  struct MultiWalkerResource : public Resource
```
ye-luo, line 181: I think we can keep `struct MultiWalkerResource;` in the header and move the definition to cpp.

Regarding the name. I used to define the resource struct outside the resource owner class, then carrying the owner class name was necessary. However here, the resource class is defined inside the owner class scope. So not carrying the owner class name should be fine.

ye-luo (changes requested):

src/QMCHamiltonians/tests/test_PairCorrEstimator.cpp
PDoakORNL: The tests spew so much unecessary information because of the need to setup objects.  If a test fails and if there is a need to debug remove this.  The default should not to print a huge amount of application output irrelevant to the test.  Also when running tests -V in terminal it really slows them down.  It's almost always the same output because we use the same input over and over for mocking.

PDoakORNL: I will revert the changes in historical tests but I always include this in my tests as we are not testing the ability to blast stdout.

src/QMCHamiltonians/BareKineticEnergy.h
PDoakORNL: I have a good number of PR's coming I'll look at how this usage works for them and make it consistent along the way.  I think @markdewing has a point.  Will leave as is for now.

src/QMCHamiltonians/tests/test_PairCorrEstimator.cpp
ye-luo: That is why I put up #4090. I will put up a PR removing all the rest pause in tests after that one and this one get merged.

ye-luo (approved):

PDoakORNL:
Please don't remove output_manager.pause() from tests I have written. In those cases that output is irrelevant to the test and repetitive. If a unit test has output it should be relevant to the test and brief.

The only good thing about the global outputManager is that it can be disabled and I can forget objects write to it as a side effect of constructing an object.

ye-luo:
Test this please

ye-luo:
Test this please

ye-luo (approved):

PR #4087, opened 2022-06-30
Test more orbital rotation scenarios

markdewing:
### Preliminary discussion
From conversations with Ye, it may be the case the mathematics of RotatedSPOs::evaluateDerivatives assumes the current rotation angle is zero. See some of the discussion in #4083 on parameter derivatives with respect to the total rotation or delta rotation.

The equation of orbital rotation is:
C_1 = exp(X) * C_0
where C_0 is the original MO coefficient matrix, X is the skew-symmetric matrix containing the rotation parameters, exp(X) is the rotation matrix, and C_1 is the MO coefficient matrix after the rotations have been applied.
The question is whether the derivatives are computed as
 * exp(0) * C_1 ( This is delta rotation )
or 
* exp(X) * C_0 ( This is total rotation )

(where the "0" in exp(0) means a matrix where all the rotation parameters are set to zero.  The resulting rotation matrix will be the identity matrix, but the derivatives with respect to those parameters may be non-zero)

Rotation matrices have the property that exp(X +Y) = exp(X)exp(Y). So the value of the wavefunction doesn't depend on how the rotation is applied, and will be the same in both schemes.
It's not clear the derivatives have the same property (or at least not as implemented in RotatedSPOs::evaluateDerivatives.)

The added tests check total and delta rotation scenarios, and they pass.
However, these cases only cover a 2x2 rotation matrix, which may very well be a special case. 

### Description of the actual PR code:

Add cases where the rotation has already been applied to the MO coefficients in the input block.

Existing case:
* No rotation in MO coeff (identity matrix), rotation in input parameters (0.1, 0.2)

Added cases:
* All the rotation in the MO coeff (0.1, 0.2), zero rotation in input parameters
* Half the rotation in the MO coeff (0.05, 0.1), half the rotation in the input parameters (0.05, 0.1)

## Proposed changes

- Bugfix

- No

laptop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/tests/test_RotatedSPOs_LCAO.cpp
+++ b/src/QMCWaveFunctions/tests/test_RotatedSPOs_LCAO.cpp
@@ -162,7 +162,7 @@ TEST_CASE("Rotated LCAO WF1", "[qmcapp]")
 }
 
 // No Jastrow, rotation angle of 0
-TEST_CASE("Rotated LCAO WF zero angle", "[qmcapp]")
+TEST_CASE("Rotated LCAO WF0 zero angle", "[qmcapp]")
 {
   Communicate* c;
   c = OHMMS::Controller;
@@ -267,7 +267,7 @@ TEST_CASE("Rotated LCAO WF zero angle", "[qmcapp]")
 }
 
 // Rotation angle of 0 and add Jastrow factory
-TEST_CASE("Rotated LCAO WF with jastrow", "[qmcapp]")
+TEST_CASE("Rotated LCAO WF2 with jastrow", "[qmcapp]")
 {
   Communicate* c;
   c = OHMMS::Controller;
@@ -364,9 +364,9 @@ TEST_CASE("Rotated LCAO WF with jastrow", "[qmcapp]")
 
 // Test the case where the rotation has already been applied to
 // the MO coefficients in the input file.
-// Should give the same results as the "Rotated LCAO WF zero angle" test case
+// Should give the same results as the "Rotated LCAO WF1 zero angle" test case
 
-TEST_CASE("Rotated LCAO WF, MO coeff rotated, zero angle", "[qmcapp]")
+TEST_CASE("Rotated LCAO WF1, MO coeff rotated, zero angle", "[qmcapp]")
 {
   Communicate* c;
   c = OHMMS::Controller;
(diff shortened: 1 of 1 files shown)


src/QMCWaveFunctions/tests/test_RotatedSPOs_LCAO.cpp
```diff
@@ -361,4 +361,188 @@ TEST_CASE("Rotated LCAO WF with jastrow", "[qmcapp]")
   CHECK(dhpsioverpsi[1] == ValueApprox(10.047601212881027));
   CHECK(dhpsioverpsi[2] == ValueApprox(2.0820644399551895));
 }
+
+// Test the case where the rotation has already been applied to
+// the MO coefficients in the input file.
+// Should give the same results as the "Rotated LCAO WF zero angle" test case
```
ye-luo, line 367: In the added two test cases, the wavefunction/orbitals are the same as "Rotated LCAO WF1" with (0.1, 0.2).
Since the derivatives are computed as the these of the current rotation, the reference numbers here are the same as "Rotated LCAO WF1". 
But why the comment says
"Should give the same results as the "Rotated LCAO WF zero angle" test case"

"Rotated LCAO WF zero angle" is different WF compared to "Rotated LCAO WF1" and the derivatives cannot be compared at a given electron configuration.

Or maybe that is just typo? Maybe we should rename "WF1" as "WF (0.1, 0.2)" to help reading.

markdewing: It's a typo. I'll either fix it to be consistent or find a better name (as you suggest).

ye-luo:
Test this please

ye-luo (approved):
LGTM. I swapped WF0 and WF1 order.

ye-luo:
> These test the total rotation scenario, and they pass.
However, these cases only cover a 2x2 rotation matrix, which may very well be a special case.

I think added tests test the current rotation. It is better to keep the PR description accurate.

src/QMCWaveFunctions/tests/test_RotatedSPOs_LCAO.cpp
```diff
@@ -361,4 +361,188 @@ TEST_CASE("Rotated LCAO WF with jastrow", "[qmcapp]")
...
+
+  const char* wf_input = R"(<wavefunction target='e'>
+    <sposet_collection type="MolecularOrbital">
+      <!-- Use a single Slater Type Orbital (STO) for the basis. Cusp condition is correct. -->
+      <basisset keyword="STO" transform="no">
+        <atomicBasisSet type="STO" elementType="He" normalized="no">
+          <basisGroup rid="R0" l="0" m="0" type="Slater">
+             <radfunc n="1" exponent="2.0"/>
+          </basisGroup>
+          <basisGroup rid="R1" l="0" m="0" type="Slater">
```
PDoakORNL, line 481: This is too much repeating text and it took me awhile to spot the difference in the XML especially.  

Seems like its just the <opt_vars> lines and the values in the checks that change. Collapsing this would easy test maintenance (hopefullly this test does not require such a huge state machine setup forever) easier and I think make it clearer what the test was about.

PDoakORNL (changes requested):
Reducing the amount of copy paste seems worthwhile.

markdewing:
I created a function to parameterize the creation of the XML input for rotation with no jastrow.  Unfortunately, the diff algorithm gets confused and makes a complete mess of the diff.

PDoakORNL (approved):
LGTM. Certainly the 2x2 is a case that better always work how its expected to.

ye-luo:
Test this please

PR #4090, opened 2022-07-01
Suppress QMCPACK printout in unit tests by default.

ye-luo:
## Proposed changes
Failing unit test was not able to report all output to cdash.
"The rest of the test output was removed since it exceeds the threshold of 100000 bytes."
https://cdash.qmcpack.org/CDash/testDetails.php?test=19984254&build=325108
So it is better to suppress output by default.
Option "--turn-on-printout" added to make the printout through if needed.

- No

epyc-server

## Checklist

### DIFF
--- a/docs/unit_testing.rst
+++ b/docs/unit_testing.rst
@@ -32,9 +32,11 @@ Some of the tests reference input files. The unit test CMake setup places those 
 
 Command line options are available on the unit test executables.  Some of the more useful ones are
 
--  List command line options.
+-  ``-h`` List command line options.
 
--  List all the tests in the executable.
+-  ``-l`` List all the tests in the executable.
+
+-  ``--turn-on-printout`` See all the QMCPACK printouts which are suppressed by default in unit tests.
 
 A test name can be given on the command line to execute just that test.  This is useful when iterating
 on a particular test or when running in the debugger.   Test names often contain spaces, so most command line environments require enclosing the test name in single or double quotes.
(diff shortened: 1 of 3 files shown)


prckent:
This is an issue with cdash configuration, not QMCPACK. Why change QMCPACK, or anything at all?

ye-luo:
I tend to agree with @PDoakORNL most of the printout are unrelated with what one is working on.
When I run into a failure, going through the long printout is always painful and doesn't add valuable info.
Catch already reports where the failure is from.

> When run by hand I would like to see the unit test output.

There is an option being added to turn on the printout and you will be able to see all the printout.

PDoakORNL (approved):
Ok I admit I did not look at this closely enough.  Yes this is a good idea and then we can forget about outputManager in most tests.

prckent (changes requested):
OK, but docs are required. i.e. *.md entries

ye-luo:
> OK, but docs are required. i.e. *.md entries

Updated rst docs. I'm not aware of "*.md" files. Let me know I missed anything.
When you run unit tests manually, "QMCPACK printout is suppressed. Use --turn-on-printout to see all the printout." is on the first line if this option is not used.

prckent:
Please indicate how to route turn-on-printouts to these tests via ctest invocation.

ye-luo:
> Please indicate how to route turn-on-printouts to these tests via ctest invocation.

Simply cannot. Probably when one needs that, running individual tests manually is likely the best thing to do.
When a failure happens, adding these printout doesn't add useful info. What catch prints should be sufficient.

To add the control at ctest, the way I can think of is adding environment variable. That is quite involving code change, need to see a good reason for that.

prckent (approved):

ye-luo:
Test this please

Issue #4096, opened 2022-07-06
GCC12 NoMPI builds broken by HEG update


prckent:
**Describe the bug**

All builds without MPI using GCC12.1 failed to compile code related to yesterdays HEG update.

https://cdash.qmcpack.org/CDash/viewBuildError.php?buildid=326504

```
In file included from /scratch/pk7/QMCPACK_CI_BUILDS/qmcpack/src/Particle/ParticleIO/LatticeIO.cpp:25:
/scratch/pk7/QMCPACK_CI_BUILDS/qmcpack/src/QMCWaveFunctions/ElectronGas/HEGGrid.h:31:40: error: 'constexpr const std::array<int, 31> qmcplusplus::HEGGrid<T>::n_within_shell' has incomplete type
   31 |   static constexpr std::array<int, 31> n_within_shell{{1,   7,   19,  27,  33,  57,  81,  93,  123, 147, 171,
      |             
```
**To Reproduce**

Latest develop.

**Expected behavior**

Build successfully

**System:**

sulfur and nitrogen

PR #4097, opened 2022-07-06
Add missing <array> include.

ye-luo:
## Proposed changes
Fixes https://github.com/QMCPACK/qmcpack/issues/4096
In MPI builds, std::array has been pulled implicitly but not in non-MPI builds.
This is the reason why only non-MPI builds failed.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):
Impressively quick fix, thanks Ye.

PR #4101, opened 2022-07-06
Only create gridded Jastrow and pseudopotential files at debug level verbosity

prckent:
## Proposed changes

Change the defaults to only create *.pp.dat, J1.*, J2.*, uk.* files when verbosity is set to the debug level.

i.e. After this change, `qmcpack input.xml` will result in a much cleaner run directory by default while `qmcpack -verbosity=debug input.xml` will restore historical behavior. 

In previous discussions, these files are rarely used and only in debugging type situations. They are also not mentioned in the manual.

Note that the einspline.tile_* files are still created since these are mentioned in the manual and the excitations tutorial. 

- New feature

- Yes

Someone relying on these files will have to specify -verbosity=debug.

Sulfur, nightly gcc12 mpi config

## Checklist

### DIFF
--- a/src/QMCHamiltonians/ECPComponentBuilder.cpp
+++ b/src/QMCHamiltonians/ECPComponentBuilder.cpp
@@ -21,6 +21,7 @@
 #include "Utilities/IteratorUtility.h"
 #include "Utilities/SimpleParser.h"
 #include "Message/CommOperators.h"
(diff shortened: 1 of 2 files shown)


ye-luo:
Test this please

ye-luo:
Test this please

ye-luo (approved):

Issue #3056, opened 2021-03-27
Change blocks_between_recompute default?


prckent:
Currently when run at full precision the inverse matrices are never recomputed by default, no matter how long the run. This is guaranteed to eventually become incorrect and is unsafe. Mixed precision uses 1 block for the default recomputation period.

I think we should change the default to always be 1, or at least a small number. An additional improvement would be to check and print the accuracy of the matrices when the matrices are recomputed each recompute period (e.g. deviation from MM^-1=I). This could be used as a guide by the user to help set the recompute period.

```
blocks_between_recompute Recompute the accuracy critical determinant part of the wavefunction
from scratch: =1 by default when using mixed precision. =0 (no recompute) by default when not
using mixed precision. Recomputing introduces a performance penalty dependent on system size.
```

PR #4109, opened 2022-07-12
Safe default for blocks_between_recompute at full precision

prckent:
## Proposed changes

Change the default number of blocks between computing the wavefunction (primarily the inverse Slater matrices) for non-mixed precision builds to 10 from the current value of zero (never). Set the input parameter in a couple of tests for coverage.

QMCPACK should not have unsafe defaults, and this removes one. Closes https://github.com/QMCPACK/qmcpack/issues/3056 . Currently if this input is not set, the results are guaranteed to be wrong after a sufficiently long run due to accumulation of numerical error. Unfortunately while the error clearly is larger in larger systems, how it accumulates is not obvious, particularly for different quality wavefunctions where e.g. occasional near electron coalescence presumably risks worsening the error. Potentially some of the reported problems with production runs could be due to this. While there are other more plausible reasons, this is an easy fix to one possibility. (Thanks to @jtkrogel for discussion on this issue.)

The new value of ten should be benign in terms of performance in nearly all circumstances while being numerically safe. 
Experts can set a larger value or zero if they wish. 

A future improvement could involve investigation of the error, checking the accumulated error in at least the inverses, or changing the default depending on electron count and steps per block. Ideally the code would check numerical consistency and restart a section or block if e.g. the local energy was too poor. The same analysis could be applied to mixed precision builds as well.

- Bugfix
- New feature
- Documentation changes

- No

gcc 12 mpi, nightly sulfur config.

## Checklist

### DIFF
--- a/docs/methods.rst
+++ b/docs/methods.rst
@@ -401,10 +401,11 @@ Additional information:
 - ``storeconfigs`` If ``storeconfigs`` is set to a nonzero value, then electron configurations during the VMC run are saved to
   files.
 
-- ``blocks_between_recompute`` Recompute the accuracy critical determinant part of the wavefunction
-  from scratch: =1 by default when using mixed precision. =0 (no
-  recompute) by default when not using mixed precision. Recomputing
-  introduces a performance penalty dependent on system size.
+- ``blocks_between_recompute`` Recompute the accuracy critical determinant part of the wavefunction from scratch: =1 by
+  default when using mixed precision. =10 by default when not using mixed precision. 0 can be set for no recomputation
+  and higher performance, but numerical errors will accumulate over time. Recomputing introduces a performance penalty
+  dependent on system size, but protects against the accumulation of numerical error, particularly in the inverses of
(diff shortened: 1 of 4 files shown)


docs/methods.rst
```diff
@@ -266,10 +266,11 @@ Additional information:
 - ``storeconfigs`` If ``storeconfigs`` is set to a nonzero value, then electron configurations during the VMC run are saved to
   files.
 
-- ``blocks_between_recompute`` Recompute the accuracy critical determinant part of the wavefunction
-  from scratch: =1 by default when using mixed precision. =0 (no
-  recompute) by default when not using mixed precision. Recomputing
-  introduces a performance penalty dependent on system size.
+- ``blocks_between_recompute`` Recompute the accuracy critical determinant part of the wavefunction from scratch: =1 by
```
ye-luo, line 269: There are multiple times of the same documentation. I think you only updated one.

ye-luo (approved):
Test this please

ye-luo:
Test this please

PR #4110, opened 2022-07-12
Orbital rotation initializes parameter only once, fixes for legacy driver with threads

markdewing:
See #4083, #3977, and #4076 for previous attempts.

This is the minimal change to get the added test to pass (parameter derivative for the legacy driver when multiple threads are used).

As a result of investigations from the comment in https://github.com/QMCPACK/qmcpack/pull/4083#pullrequestreview-1036223817 , calling buildOptVariables multiple times is the culprit in resetting the global indexing in myVars.

- Bugfix

- No

desktop

## Checklist


prckent (approved):
LGTM

Not relevant here, but I'll note that the test names don't follow our usual prefixes, and have not for some time...

ye-luo:
Test this please

PR #4113, opened 2022-07-13
intel/offload: Update cmake for intel compiler for GPU offloading

stephan-rohr:
The Intel compiler requires debug flags '-g -O0' to be explicetly
specified via CMake flags 'CMAKE_CXX_FLAGS_<target>' when offloading
code to GPU. If those flags are not specified, offloaded code/kernels
may be optimized and debugging experience may differ from what is
expected by users.

## Proposed changes

The Intel compiler requires 'CMAKE_CXX_FLAGS_DEBUG' and 'CMAKE_CXX_FLAGS_RELWITHDEBINFO' to contain debug flags in order to enable debugging of GPU offloaded code. 

- Build related changes

- No

## Checklist

### DIFF
--- a/CMake/IntelCompilers.cmake
+++ b/CMake/IntelCompilers.cmake
@@ -23,6 +23,8 @@ if(QMC_OMP)
           "spir64"
           CACHE STRING "Offload target architecture")
       set(OPENMP_OFFLOAD_COMPILE_OPTIONS "-fopenmp-targets=${OFFLOAD_TARGET}")
+      set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -O0")
+      set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -g -O2")
     endif(ENABLE_OFFLOAD)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fiopenmp")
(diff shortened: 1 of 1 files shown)


ye-luo (changes requested):
Users should be responsible what they are doing.
They can pass '-g -O0' to `CMAKE_CXX_FLAGS_<BUILD_TYPE>` via command-line.
Hard-coded options is just bad and we should not do that.
We will likely restructure the handling of OpenMP offload in CMake after the clang 15 release. That will allow us to customize offload compilation pass more precisely.

prckent:
@ye-luo I disagree here. I think only an extra comment is needed about why these settings are there. We have plenty of CMake hacks already for vendor software stack quirks/weaknesses, and provided they are contained they are OK for me.

I will never remember to set these flags!

ye-luo:
`We have plenty of CMake hacks already for vendor software stack quirks/weaknesses`
We only do that when there is no better alternatives and it is absolutely needed. It is a huge burden to maintain hot fixes.
I view that more valuable for users but developers are more capable of better using CMake.
Debugging with `-O0` may actually be a bad thing on GPU. Low optimization level means more register/scratch usage, the `-O0` compiled binary being debugged can have totally different behavior from `-O3` binary and have non-reproducible issues in debug builds. So it is an option and should not be hard-coded.

ye-luo:
https://qmcpack.readthedocs.io/en/develop/installation.html#notes-for-openmp-target-offload-to-accelerators-experimental
You may add a note under Intel oneAPI. Please mention the last verified date and compiler version.

PR #4120, opened 2022-07-18
CoulombPBCAA operator support for listener

PDoakORNL:
## Proposed changes

This contains changes to `CoulombPBCAA` to support Listeners. And the corresponding test.  If this is too small a chunk CoulombPBCAB could be added.  Documentation is partial here but should make sense when the changes to QMCHamiltonian are PR'd.  That can't really be done until all the lower level changes go in.

- New feature
- Build related changes
- Documentation changes

- No

Leconte

## Checklist


src/QMCHamiltonians/tests/test_coulomb_pbcAA.cpp
```diff
@@ -257,13 +267,99 @@ void test_CoulombPBCAA_3p(DynamicCoordinateKind kind)
...
 }
 
 TEST_CASE("Coulomb PBC A-A BCC 3 particles", "[hamiltonian]")
 {
   test_CoulombPBCAA_3p(DynamicCoordinateKind::DC_POS);
   test_CoulombPBCAA_3p(DynamicCoordinateKind::DC_POS_OFFLOAD);
 }
+
+
+TEST_CASE("CoulombAAListener", "[hamiltonian]")
```
ye-luo, line 281: Can we fuse this test with the above test and make it testing with and without listener?

PDoakORNL: I'm hesitant to merge further tests here.  They are already broad integration tests with unclear goals.  I'd rather see refactoring so that these are more targeted unit tests. There certainly is room to factor out some shared setup code.  I'll look at doing that.

src/QMCHamiltonians/CoulombPBCAA.h
```diff
@@ -101,6 +100,17 @@ struct CoulombPBCAA : public OperatorBase, public ForceBase
...
 
+  /**
+   * Evaluate the contribution of this component of multiple walkers per particle reporting
+   * to registered listeners from Estimators.
+   */
+  void mw_evaluatePerParticle(const RefVectorWithLeader<OperatorBase>& o_list,
+                              const RefVectorWithLeader<TrialWaveFunction>& wf_list,
+                              const RefVectorWithLeader<ParticleSet>& p_list,
+                              const std::vector<ListenerVector<RealType>>& listeners,
+                              const std::vector<ListenerVector<RealType>>& ion_listeners) const override;
```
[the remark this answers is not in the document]
ye-luo: I consider moving particles and non-moving ones. The moving one is usually electron particle set. Others may be ions and plus something else if added.

src/QMCHamiltonians/CoulombPBCAA.cpp
```diff
@@ -213,6 +221,93 @@ void CoulombPBCAA::mw_evaluate(const RefVectorWithLeader<OperatorBase>& o_list,
     OperatorBase::mw_evaluate(o_list, wf_list, p_list);
 }
 
+void CoulombPBCAA::mw_evaluatePerParticle(const RefVectorWithLeader<OperatorBase>& o_list,
+                                          const RefVectorWithLeader<TrialWaveFunction>& wf_list,
+                                          const RefVectorWithLeader<ParticleSet>& p_list,
+                                          const std::vector<ListenerVector<RealType>>& listeners,
+                                          const std::vector<ListenerVector<RealType>>& listeners_ions) const
```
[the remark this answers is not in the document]
ye-luo: I left my comment noting my concern. Let's move on.

src/QMCHamiltonians/tests/test_coulomb_pbcAA.cpp
ye-luo: I think both mw_evaluate functions w/ w/o listeners should be available to use at the same time. That is why I think we should not avoid putting it as a fully separate unit test.

ye-luo:
Test this please

ye-luo (approved):

src/QMCHamiltonians/tests/test_coulomb_pbcAA.cpp
PDoakORNL: I agree but I'm used to crudely think of unit tests being roughly per function.  But I see your point here, if we want to skip reporting we need to be able to call both with the same setup.  That should work fine let me add a call to mw_evaluate(). I'm not sure that justifies dropping one of the other tests though.

PR #4122, opened 2022-07-19
Orbital rotation test for wavefunction with jastrow

markdewing:
Add a test of the parameter gradient for a wavefunction that includes orbital rotation and a Jastrow factor.

Update the python script to more closely match how QMCPACK computes the derivatives.

- No

desktop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/tests/gen_rotated_lcao_wf.py
+++ b/src/QMCWaveFunctions/tests/gen_rotated_lcao_wf.py
@@ -39,8 +39,18 @@ def orb2(R):
     snorm2 = sto_norm2(zeta)
     return snorm2* y00 * r* np.exp(-zeta*r)
 
+# Return the 2x2 rotation matrix
+def rot_mat_size2(theta):
+    return np.array([[ np.cos(theta), np.sin(theta) ],
+                     [ -np.sin(theta), np.cos(theta) ]])
+
 def rot_orb(R, theta):
-    return orb1(R) * np.cos(theta) + orb2(R) * np.sin(theta)
+    global Coeff
+    c00 = Coeff[0,0] * np.cos(theta) + Coeff[1,0] * np.sin(theta)
+    c01 = Coeff[0,1] * np.cos(theta) + Coeff[1,1] * np.sin(theta)
+    return orb1(R) * c00 + orb2(R) * c01
+
+    #return orb1(R) * np.cos(theta) + orb2(R) * np.sin(theta)
 
 
 def jastrow(r12, B):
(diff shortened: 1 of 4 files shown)


src/QMCWaveFunctions/tests/gen_rotated_lcao_wf.py
```diff
@@ -294,11 +305,23 @@ def run_qmc(VP):
 
 
 def run_qmc_parameter_derivatives():
-    global use_jastrow
-    use_jastrow=False
-    theta1 = 0.1
-    theta2 = 0.1
-    VP = np.array([theta1, theta2])
+    global use_jastrow, Coeff
+    use_jastrow=True
```
PDoakORNL, line 309: I had forgotten you could even use globals like this in python.  Why not enclose the explosion of functions in one (or more classes) and scope Coeff and use_jastrow as members?  I'd feel much better about the maintenance and ability to extend this code.  I also notice that the print function mutates the global variable use_jastrow as a side effect and is by default false, currently the net effect of calling it is to set use_jastrow=True but this is just waiting to burn someone in the future.

PDoakORNL (changes requested):
It just generates test data but globals and side effects will make this dangerous to revisit.

src/QMCWaveFunctions/tests/gen_rotated_lcao_wf.py
markdewing: If this code were intended to scale, I would agree about using a class.  My goal is simplicity of the code so that it is straightforward to read. Good software engineering practices can work against the simplicity (and readability) of code.

I originally used a class to hold the wavefunction when I started writing Python code for these simple wavefunctions.  However, I've moved away from using a class because it was more difficult to use numba and other acceleration and GPU offload technologies, and ended up moving everything to separate functions.  (The use of acceleration tech is not a requirement in this case, but I wanted to explain where my thinking on the style came from).

It's not intended that print_point_values and run_qmc_parameter_derivatives are used at the same time, since the output is used in different tests. And both functions should set use_jastrow explicitly, so even if both functions get run, the output won't be unexpected.

PDoakORNL: If I see a variable of unclear scope I stop reading and immediately look to resolve its scope. I get your point about writing this in a functional style so follow through and just pass the Coeff and jastrow and write pure functions. Python improves things some by forcing the `global` declaration but if you write that why not just have an argument.

I don't like to see code that depends on side effects on global variables ever, once you know a code is written in that manner you can't reason about the output of the code without reading through the entire control flow. It isn't clear.

prckent: I think that the idea that we'll have people learning python reading this code is a good motivational argument to avoid globals and generally keep very clean structured code even for uses like this. I don't think we have given much thought to this in the past, but it would be consistent to apply the same arguments to our python "utilities" as we do to the main C++, and for the same reasons.

markdewing: I updated it to put all the wavefunction related code in a class.   One of the potential future changes is extracting run_qmc so it can be used with other wavefunctions.   One possible future change is to extract run_qmc to use on other wavefunctions.  It's easier to make a generic version of run_qmc using a class for the wavefunction versus passing all the values as parameters.

PDoakORNL (approved):

prckent:
Test this please

ye-luo:
Test this please

PR #4124, opened 2022-07-20
Update OpenMP/SYCL interop

ye-luo:
## Proposed changes
No more need of explicitly handling  Level0 or OpenCL.

- New feature

- No

dell-laptop

## Checklist

### DIFF
--- a/src/Platforms/SYCL/SYCLDeviceManager.cpp
+++ b/src/Platforms/SYCL/SYCLDeviceManager.cpp
@@ -17,33 +17,47 @@
 #include <string>
 #include <algorithm>
 #include "config.h"
-#include "OutputManager.h"
-#include "determineDefaultDeviceNum.h"
-#if defined(_OPENMP)
-#include <omp.h>
-#endif
-
+#include "Platforms/Host/OutputManager.h"
+#include "Platforms/Host/determineDefaultDeviceNum.h"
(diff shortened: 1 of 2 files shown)


src/Platforms/SYCL/SYCLDeviceManager.cpp
```diff
@@ -16,34 +16,58 @@
...
+    : context_(context), device_(device), interop_(interop)
+#else
+syclDeviceInfo::syclDeviceInfo(const sycl::context& context, const sycl::device& device)
+    : context_(context), device_(device)
 #endif
+{}
+
+syclDeviceInfo::~syclDeviceInfo()
+{
+  //#pragma omp interop destroy(interop_)
```
ye-luo, line 35: This still needs a compiler fix.

PDoakORNL (approved):
LGTM, but to be clear eventually omp interop is going to provide a more "portable" way to do some of this?

ye-luo:
The old way and the new ways are both allowed in OpenMP spec. Requesting context/device and then rebuild queues or directly request queues are both valid. Before this PR, I need to call out Level0 or OpenCL which are clearly less portable since I need to deal with handles explicitly which somehow rely on additional header files.

ye-luo:
Test this please

PR #4126, opened 2022-07-21
Parameter Filtration in LM Optimizer

LeonOtis:
## Proposed changes

This is the remainder of the changes originally proposed in #4061 . This PR allows the wave function parameters optimized by the linear method to be filtered based on the amount of noise in the parameter gradients. The LM will then optimize only a subset of the parameters while the others can be handled by accelerated descent in a hybrid optimization. 

The main changes are in the LM engine and the QMCFixedSampleLinearOptimizeBatched driver. There are a few additional user inputs for turning on parameter filtration, which are currently being handled similarly to the many other driver inputs. The parameter filtration should now be compatible with both MPI and OMP parallelism. Thanks again for all the feedback so far and I am always happy to make further improvements.

- New feature

- No

Lawrencium computing cluster

## Checklist


markdewing:
Test this please

src/QMCDrivers/WFOpt/EngineHandle.h
```diff
@@ -118,14 +118,22 @@ class LMYEngineHandle : public EngineHandle
...
 
-  void prepareSampling(int num_params) override
+  void prepareSampling(int num_params,int num_samples) override
   {
+      int num_threads = omp_get_max_threads();
     der_rat_samp.resize(num_params + 1, 0.0);
     le_der_samp.resize(num_params + 1, 0.0);
+    if(lm_engine_.getStoringSamples())
+    {
+        sample_count.resize(num_threads,0);
```
PDoakORNL, line 133: I'd encourage you to read the section on [thread safety of stl containers](https://en.cppreference.com/w/cpp/container)

src/QMCDrivers/tests/test_EngineHandle.cpp
```diff
@@ -46,7 +46,8 @@ TEST_CASE("EngineHandle construction", "[drivers]")
 
 
   int fake_num = 5;
-  handle->prepareSampling(fake_num);
+  int fake_sample_num = 100;
+  handle->prepareSampling(fake_num,fake_sample_num);
```
PDoakORNL, line 50: this needs to be tested in a threaded context.

PDoakORNL: in fact all of the engine methods need to be or it needs to be.  Or they need to be called only from one thread.

src/QMCDrivers/WFOpt/EngineHandle.h
```diff
@@ -143,9 +151,22 @@ class LMYEngineHandle : public EngineHandle
           le_der_samp[0] * static_cast<FullPrecValue>(dlogpsi_array.getValue(j, ib));
     }
 
+    int ip = omp_get_thread_num();
```
PDoakORNL, line 154: There looks to be a potential disaster data race above this on der_rat_samp and le_der_samp.

The only id you should care about in a threaded section is the crowd id, it may or may not have any relation to the omp thead id. Crowds may be called at the same time on any collection of thread_id's the concurrency context you're in is identified by the crowd id and you should avoid needing to know it.

PDoakORNL (changes requested):
Unless I'm reading this very wrong there are serious fundamental problems with the concurrency logic of the batched linear optimization driver and this PR.  Please mark this as [WIP] until the concurrency issues in it are fixed.

One of the maintainers should help you work through this in detail.

Have you run this with opt_num_crowds > 1?  I would expect something interesting to happen.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -282,20 +281,22 @@ void QMCCostFunctionBatched::checkConfigurations(EngineHandle& handle)
...
                           const std::vector<int>& samples_per_crowd_offsets, const std::vector<int>& walkers_per_crowd,
                           std::vector<ParticleGradient*>& gradPsi, std::vector<ParticleLaplacian*>& lapPsi,
                           Matrix<Return_rt>& RecordsOnNode, Matrix<Return_rt>& DerivRecords,
                           Matrix<Return_rt>& HDerivRecords, const SampleStack& samples, opt_variables_type& optVars,
                           bool needGrads, bool compute_nlpp, const std::string& includeNonlocalH,
-                          EngineHandle& handle) {
+                          EngineHandle& handle, int NumOptimizables) {
     CostFunctionCrowdData& opt_data = *opt_crowds[crowd_id];
 
     const int local_samples = samples_per_crowd_offsets[crowd_id + 1] - samples_per_crowd_offsets[crowd_id];
```
markdewing, line 295: It would be better to keep the arrays in LMYEngine sized as the full number of samples (per rank, in rank_local_num_samples_) and use samples_per_crowd_offsets to index into those arrays.   That should get rid of all (or most) of the thread calls in takeSample.

markdewing: takeSample has `ib` passed an argument, which is the crowd-local index.  It needs to be updated to also pass `is` ( supposed to be short for 'Index into Samples', not the word `is`), which is the index into samples on each rank.

PDoakORNL: Both of those argument names could be long enough to carry information.  Two letters excepting those that are common words 'is', 'as', etc are fine for an iterated index in a small loop but not great as argument names.  `local_index` and `index_offset` come to mind.  Then only index offset would need non trivial docs.

PDoakORNL: And I think sizing the arrays to the full number of samples outside of the concurrent lambda would do a good deal here to get a hacked first version, but every method called in the concurrent context and its calls etc needs to be evaluated for thread safety.  It's quite likely they will work fine in a single threaded context.

LeonOtis:
Thanks for the feedback. The new set of changes have the history arrays in the LM engine sized according to the number of samples on an MPI rank and different threads should store data in them using the index offsets. Tests I've done with multiple threads have worked, but there could be cases I've overlooked. I'm happy to do further rounds of fixes as needed.

src/QMCDrivers/WFOpt/EngineHandle.h
```diff
@@ -38,19 +38,21 @@ class EngineHandle
...
+  virtual void prepareSampling(int num_params,int num_samples) = 0;
   /** Function for passing derivative ratios to optimizer engines
    *
    * \param[in] energy_list         Vector of local energy values
    * \param[in] dlogpsi_array       Parameter derivatives of log psi
    * \param[in] dhpsioverpsi_array  Parameter derivatives of local energy
-   * \param[in] ib                  Sample index
+   * \param[in] crowd_id            Crowd number
+   * \param[in] local_index         Crowd local index
+   * \param[in] index_offset        Offset for sample's index
```
markdewing, line 49: I'm not sure this is best described as an offset. Maybe "sample_index"?  This is the index of the sample on this MPI rank.   The extent of the index is [0, rank_local_num_samples_), so maybe rank_local_sample_index?  The rank_local part is a bit wordy, so sample_index should be sufficient.

LeonOtis: Ok, I have changed the naming to sample_index to make things clearer.

markdewing:
Do you have an example input file for this feature?  Ideally there should be a test added that also serves as an example.  However, parameter filtering gets used when scaling up the number of parameters than can be handled and a test that is small enough to be a good test might not be a good example.  It also sounds like parameter filtering gets used in combination with hybrid optimization.  it would be good to have a more complete example to see how the feature works in use.  Attaching files to the PR is good enough for now.

LeonOtis:
[example_input.zip](https://github.com/QMCPACK/qmcpack/files/9244172/example_input.zip)

Sure, here are input files for a quick example with the carbon trimer. The actual optimization results won't be very meaningful with low sample numbers and filtering an already small number of parameters, but the example_filter_hybrid.xml file shows the new inputs in the context of a hybrid optimization. Parameter filtration can be used for the LM on its own just by using the LM section of the hybrid input as a standalone, but I would generally expect the results to be inferior to allowing descent to optimize any parameters the LM leaves alone. I think this PR should already have some changes to the documentation to explain the new inputs, but I can add more if needed.

LeonOtis:
Added a small fix to the legacy driver to correct a problem @ye-luo noticed with PR #4075. At this point, what else is needed before this PR can be merged? I'm happy to revise the handling of threading more if desired. I'm also willing to come up with tests if needed though I wanted to be sure the threading issues were settled before trying to design any.

ye-luo:
@LeonOtis could you make a new branch from develop and cherry pick the last fix commit you pushed and then make a separate PR.

LeonOtis:
@ye-luo Ok, I've opened PR #4161 for that small change.

LeonOtis:
I've now added a unit test for parameter filtration. Are there any other changes I should make before this PR can be merged? @ye-luo @markdewing

ye-luo:
@LeonOtis I will make a review of this PR within this week. Q: with the current develop, does the "hybrid" method working with batched driver? When you made the fix to the legacy driver, I noticed there were no corresponding tests for the batched driver. If the current develop already works with "hybrid" and batched driver, could you introduce some tests.

LeonOtis:
@ye-luo The hybrid method should work with the batched driver. I've now added a test for it on H4, similar to what I previously added for the legacy driver.

ye-luo:
> @ye-luo The hybrid method should work with the batched driver. I've now added a test for it on H4, similar to what I previously added for the legacy driver.

Thank you. Could you make a separate PR with the new commits to develop? Since it works with the develop already.

LeonOtis:
> > @ye-luo The hybrid method should work with the batched driver. I've now added a test for it on H4, similar to what I previously added for the legacy driver.
> 
> Thank you. Could you make a separate PR with the new commits to develop? Since it works with the develop already.

Ok, I've opened a new PR #4172 that just adds the test.

src/QMCDrivers/tests/test_EngineHandle.cpp
ye-luo: It prepares storage for each sample.
This is called outside the thread context and thread agnostic.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
```diff
@@ -673,6 +684,15 @@ bool QMCFixedSampleLinearOptimizeBatched::processOptXML(xmlNodePtr opt_xml,
...
   block_lm    = (block_lmStr == "yes");
 
+  filter_paramStr = lowerCase(filter_paramStr);
+  filter_param_   = (filter_paramStr == "yes");
+
+  filter_infoStr = lowerCase(filter_infoStr);
+  filter_info_   = (filter_infoStr == "yes");
+
+  store_samplesStr = lowerCase(store_samplesStr);
+  store_samples_   = (store_samplesStr == "yes");
```
ye-luo, line 694: Why not m_param.add directly filter_param_, filter_info_, store_samples_?

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
```diff
@@ -1131,23 +1163,150 @@ bool QMCFixedSampleLinearOptimizeBatched::adaptive_three_shift_run()
...
+  }
+  else
+  {
+    app_log() << "Skipping initialization at first" << std::endl;
+    EngineObj->store_blocked_lm_info(nblocks, nkept);
+  }
+
   // initialize the engine if we do not use block lm or it's the first part of block lm
-  EngineObj->initialize(nblocks, 0, nkept, previous_update, false);
+  //  EngineObj->initialize(nblocks, 0, nkept, previous_update, false);
```
ye-luo, line 1194: Remove?

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.h
```diff
@@ -233,16 +233,37 @@ class QMCFixedSampleLinearOptimizeBatched : public QMCDriverNew, LinearMethod
...
+  std::string filter_infoStr;
+
+  ///threshold for filtering parameters for the lm
+  double ratio_threshold_;
+
+  ///whether to store samples for the lm
+  bool store_samples_;
+
+  ///whether to store samples for the lm
+  std::string store_samplesStr;
```
ye-luo, line 255: Could you put all the LMY specific option variables in a structure?

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -370,7 +369,7 @@ void QMCCostFunctionBatched::checkConfigurations(EngineHandle& handle)
           RecordsOnNode[is][LOGPSI_FIXED] = opt_data.get_log_psi_fixed()[ib];
           RecordsOnNode[is][LOGPSI_FREE]  = opt_data.get_log_psi_opt()[ib];
 
-          handle.takeSample(energy_list, dlogpsi_array, dhpsioverpsi_array, ib);
+          handle.takeSample(energy_list, dlogpsi_array, dhpsioverpsi_array, ib, is);
```
ye-luo, line 372: Could you adjust the takeSample API and move the call out of the ib loop?
takeSample(energy_list, dlogpsi_array, dhpsioverpsi_array, base_sample_index);
In the takeSample implementation, handle the ib loop.

ye-luo (commented):
@LeonOtis I did some cleaning, please pull before continuing editing. Once my requests are addressed, I think this PR can get in. I close my eyes on run_three_shift and everything inside formic.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
LeonOtis: Ok, I've put the ib loop inside takeSample.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
LeonOtis: Removed.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
LeonOtis: I find when I try m_param.add directly for those booleans, the user input must be "1" to for them to be set to true. Using the strings seemed to allow for more intuitive user input and follows what's done for the blocked LM and excited state settings.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.h
LeonOtis: I've grouped the optimizer input parameters together in an unnamed struct. I had to leave the string parameters outside the struct to compile successfully, but they are listed just above the struct if this change is just for organizing the file.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
ye-luo: Thanks. I will keep a note to update bool handling.

src/QMCDrivers/WFOpt/EngineHandle.h
```diff
@@ -81,7 +82,7 @@ class DescentEngineHandle : public EngineHandle
   //Retrieve der_rat_samp vector for testing
   const std::vector<FullPrecValue>& getVector() const { return der_rat_samp; }
 
-  void prepareSampling(int num_params) override
+  void prepareSampling(int num_params, int num_samples) override
   {
     engine_.prepareStorage(omp_get_max_threads(), num_params);
```
ye-luo, line 87: I think we should avoid using number of threads and thread id.

src/QMCDrivers/WFOpt/EngineHandle.h
```diff
@@ -92,20 +93,24 @@ class DescentEngineHandle : public EngineHandle
...
+      le_der_samp[0]  = energy_list[local_index];
+
+      int num_params = der_rat_samp.size() - 1;
+      for (int j = 0; j < num_params; j++)
+      {
+        der_rat_samp[j + 1] = static_cast<FullPrecValue>(dlogpsi_array.getValue(j, local_index));
+        le_der_samp[j + 1]  = static_cast<FullPrecValue>(dhpsioverpsi_array.getValue(j, local_index)) +
+            le_der_samp[0] * static_cast<FullPrecValue>(dlogpsi_array.getValue(j, local_index));
+      }
+      int ip = omp_get_thread_num();
```
ye-luo, line 111: would like to move away from thread_num

ye-luo (dismissed):
There is still an issue of explicitly using threads. That is not a good pattern. However, I'd like that being addressed in a later PR.

ye-luo:
Ping @PDoakORNL @markdewing since you gave review comments as well.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
ye-luo: I actually fixed the ParameterSet handling boolean in the develop.
@PDoakORNL seems out of office. As we are waiting for his review, could you try removing the input strings?

LeonOtis: Ok, thanks. Removing the strings works now and I've pushed a commit.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.h
```diff
@@ -196,54 +196,60 @@ class QMCFixedSampleLinearOptimizeBatched : public QMCDriverNew, LinearMethod
...
-
-  //type of the previous optimization method, updated by processOptXML before run
-  OptimizerType previous_optimizer_type_;
+  // Parameters in this struct are used by one or more of the adaptive LM, descent, or hybrid optimizers
 
-  //type of the current optimization method, updated by processOptXML before run
-  OptimizerType current_optimizer_type_;
+  //String inputs are listed outside the struct
+  ///name of the current optimization method, updated by processOptXML before run
+  std::string MinMethod;
```
PDoakORNL, line 203: the naming of all these private variables does not follow the coding standards of QMCPACK.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.h
```diff
@@ -196,54 +196,60 @@ class QMCFixedSampleLinearOptimizeBatched : public QMCDriverNew, LinearMethod
...
 
-  //whether to use hybrid method
-  bool doHybrid;
+  //LMY related input
+  struct
```
PDoakORNL, line 206: What purpose does this anonymous struct serve?  
To my knowledge anonymous structs are not part of the c++ standard although I guess they compile for most compilers.
Does this make the code easier to read or better in anyway?  If you want a nested data structure name it, declare a class member of that nested type with a name following the naming convention i.e. `lmy_in_` and access it properly for example `lmy_in_.num_shifts`

src/QMCDrivers/WFOpt/EngineHandle.h
```diff
@@ -92,20 +93,24 @@ class DescentEngineHandle : public EngineHandle
...
+      engine_.takeSample(ip, der_rat_samp, le_der_samp, le_der_samp, 1.0, 1.0);
```
PDoakORNL, line 112: How this relates to all the replica based accesses in the DescentEngine itself is really cryptic. Since this oenpmp thread id which isn't guaranteed to be any particular number especially in the batched context yet it is used as the replica id in the engine, which seems like it is very important.

src/QMCDrivers/WFOpt/EngineHandle.h
```diff
@@ -81,7 +82,7 @@ class DescentEngineHandle : public EngineHandle
   //Retrieve der_rat_samp vector for testing
   const std::vector<FullPrecValue>& getVector() const { return der_rat_samp; }
 
-  void prepareSampling(int num_params) override
+  void prepareSampling(int num_params, int num_samples) override
```
PDoakORNL, line 85: num samples is unused.

PDoakORNL (changes requested):
I think some of the thread safety issues I previously was worried have been cleaned up but the subtypes of EngineHandle seem to be trouble and perhaps only one is really relevant to this PR.

DescentEngineHandle/Descent engine seems incoherent, if this is just here to get things compiling keep the unfinished code in your private branch and remove it from the PR throw a runtime exception if anyone tries to use it.

That would get us to smaller issues like the anonymous struct.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -280,6 +278,7 @@ void QMCCostFunctionBatched::checkConfigurations(EngineHandle& handle)
   std::vector<int> samples_per_crowd_offsets(opt_num_crowds + 1);
   FairDivide(rank_local_num_samples_, opt_num_crowds, samples_per_crowd_offsets);
 
+  handle.prepareSampling(NumOptimizables, rank_local_num_samples_);
```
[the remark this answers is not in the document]
ye-luo: I'm not getting this part "don't infact have the same sematics for the prepareSampling and takeSample Methods."

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.h
ye-luo: I also found this anonymous struct strange and it basically means no struct being used. My original request to introduce a struct was to collect several categories of input parameters.

PDoakORNL: The other issue here is the incoherence of encapsulation and the equivalence of the "engines" in this class but that is outside the scope of this PR.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
```diff
@@ -98,7 +97,10 @@ QMCFixedSampleLinearOptimizeBatched::QMCFixedSampleLinearOptimizeBatched(
       block_first(true),
       block_second(false),
       block_third(false),
-      MinMethod("OneShiftOnly"),
+      filter_param_(false),
```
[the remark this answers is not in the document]
ye-luo: Fixed

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.h
ye-luo: Name the struct as LMYOptions.

src/formic/utils/lmyengine/matrix_builder.h
```diff
@@ -219,6 +219,27 @@ namespace cqmc {
 
       }
 
+    void resetParamNumber(int new_num)
+     {
+     _num_params = new_num;
+     int NumThreads = omp_get_max_threads();
```
PDoakORNL, line 225: The is an abstraction to get the number of threads, see src/Concurrency/Info.hpp but the optimization has no guarantee it will run on that many threads so this just isn't an effective way to allocate concurrenct workspace.  It's possible this will work for awhile but it is at least a memory use bomb.

PDoakORNL (changes requested):
Ok another iteration to go here.  Some of these broken Engines/EngineHandles and elements of formic could just be documented including todo's  But by the end of this series of PR's the illogical and/or wrong code needs to be gone.  Especially since sizable amounts of it appear to be new.

src/formic/utils/lmyengine/matrix_builder.h
ye-luo: #4209 tracks the issue

src/formic/utils/lmyengine/engine.cpp
```diff
@@ -1529,6 +1530,468 @@ void cqmc::engine::LMYEngine<S>::get_brlm_update_alg_part_two(const formic::VarD
...
+     weight_history.resize(numSamples);
+     
+ }
+
+//Function for storing a sample (local energy and the sets of derivative ratios)
+ template<typename S>
+ void cqmc::engine::LMYEngine<S>::store_sample(std::vector<S> & der_rat_samp,std::vector<S> & le_der_samp,std::vector<S> & ls_der_samp,double vgs_samp,double weight_samp,int sample_index)
+ {
+
+    int myThread = omp_get_thread_num();
```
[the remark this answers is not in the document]
ye-luo: Removed.

PDoakORNL (approved):

ye-luo:
Test this please

PR #3994, opened 2022-04-29
Add Automatic Rebase Github Action Job

walshmm:
## Proposed changes
In reference to issue: https://github.com/QMCPACK/qmcpack/issues/3862
This pr contains configuration for a job to automatically rebase a PR granted the PR body contains the phrase:

!-> Feel free to automatically rebase this PR. <-!

Original WIP PR: #3878

## What type(s) of changes does this code introduce?

- Other (please describe): Github Actions Pipeline Changes

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?

remote github runners

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- Yes. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

Issue #3862, opened 2022-02-21
Explore GitHub options for automatic update develop onto approved and tested PR branch


williamfgc:
**Is your feature request related to a problem? Please describe.**
Currently, when PRs come simultaneously they get out-of-sync with any changes to the target `develop` branch.
The goal is to update automatically an already approved and tested PR branch (currently merging or rebasing `develop`). 

**Describe the solution you'd like**
Tested and approved PRs shouldn't go out of sync, and ideally be rebased from the latest `develop` branch.

**Describe alternatives you've considered**
Options to explore: GitHub Actions that could do these (if any?) or a GitHub [probot](https://probot.github.io/)

**Additional context**
This could be a bit invasive to the PR owner/contributors as it will make any pre-existing local branch non fast forward with the remote PR branch.

prckent:
I think we only want this for PRs that have been approved for Automerge and where the tests have been run successfully once. Therefore the admins will be able to approve for automerge a handful of PRs and they will be updated and merged in sequence. Currently this requires human steps for every PR.

ye-luo:
I'm not sure if we'd like to do rebase. merge develop into the feature branch is more common.

walshmm:
#3878 is ready to be looked at, but I will leave it in a draft state as I'd want to clean up commits before it got merged.

prckent:
Closing, outdated.

PR #4128, opened 2022-07-22
Reapprove PR after Auto Rebase Completes

walshmm:
## Proposed changes
In reference to issue: https://github.com/QMCPACK/qmcpack/issues/3862
Continuation of this PR: https://github.com/QMCPACK/qmcpack/pull/3994

Allows the rebase action to reapprove pr after autorebase is complete.  It approves specifically the SHA it pushes, so there shouldnt be an accidental approval.

- Other (please describe): Github Actions Pipeline Changes

- No

remote github runners

Tested on this PR: https://github.com/walshmm/qmcpack/pull/11
You can follow its listed history to see it was rebased and reapproved via bot before being automerged.

## Checklist

### DIFF
--- a/external_codes/github_actions/auto-rebase.sh
+++ b/external_codes/github_actions/auto-rebase.sh
@@ -111,8 +111,9 @@ UNTRIMMED_COMMITTER_TOKEN=${!USER_TOKEN:-$GITHUB_TOKEN}
 COMMITTER_TOKEN="$(echo -e "${UNTRIMMED_COMMITTER_TOKEN}" | tr -d '[:space:]')"
 
 git remote set-url origin https://x-access-token:$COMMITTER_TOKEN@github.com/$GITHUB_REPOSITORY.git
-git config --global user.email "$USER_EMAIL"
-git config --global user.name "$USER_NAME"
+# CHANGED FROM ORIGNAL: use bot credentials
+git config --global user.email "QMCPACKbot@gmail.com"
+git config --global user.name "QMCPACK-Bot"
 
 git remote add fork https://x-access-token:$COMMITTER_TOKEN@github.com/$HEAD_REPO.git
 
@@ -128,6 +129,11 @@ git rebase origin/$BASE_BRANCH
 
 # push back
 git push --force-with-lease fork fork/$HEAD_BRANCH:$HEAD_BRANCH
+
+# CHANGE FROM ORIGINAL: add empty commit signed by bot 
+git config --global user.signingkey "$QMCPACK_BOT_GPG_KEY"  
+git commit --allow-empty -S -m "Rebased Signed Off by QMCPACK-Bot"
+git push fork HEAD
 
 COMMIT_SHA=$(git rev-parse --verify HEAD)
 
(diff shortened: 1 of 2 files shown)


external_codes/github_actions/auto-rebase.sh
```diff
@@ -127,4 +127,20 @@ git checkout -b fork/$HEAD_BRANCH fork/$HEAD_BRANCH
 git rebase origin/$BASE_BRANCH
 
 # push back
-git push --force-with-lease fork fork/$HEAD_BRANCH:$HEAD_BRANCH
\ No newline at end of file
+git push --force-with-lease fork fork/$HEAD_BRANCH:$HEAD_BRANCH
```
PDoakORNL, line 130: what happens if this push fails.  Say an attacker knows the what the hash of the rebase will be and already pushed it.  Will this action still continue and use the COMMIT_SHA?

PDoakORNL (commented):
So here's the objection a cyber security guy is going to make and I don't mean to make this difficult but we'd like to argue to be able to do CI or nightlies on Summit, Frontier or Perlmutter someday.

We're hoping that Github's monitoring for SHA-1 collision attacks which are "feasible" now will prevent the following scenario, but we don't actually know what they are doing.  Certainly if it relates to the ordering of pushes with colliding hashes that's probably not good enough.

The scenario would be something like an attacker submits a PR knowing already what the hash will be once it is rebased and secures approval with the rebase action turned on. Since the attacker can determine what the hash will be before the action rebases the PR they have time to prepare a colliding commit with a payload in it.

Then they then have several possible ways to use this.  One is in my review comment.  Another is to between the actions rebase and automated approval make a forced push to remove the commit with that hash from the repo and replace it with the payload commit.

My first thought on how to prevent this would be to have this action also add a sufficient amount of truly random "salt" to a file in the commit such that predetermining the rebased PR hash would be impossible and instead the attacker would have only the time between the rebase and the approval to prepare the collision.

I'm pretty sure that force pushing to an already approved commit removes the approval.

external_codes/github_actions/auto-rebase.sh
walshmm: I can look into creating a bot account and signing commits as outlined in this article:
https://httgp.com/signing-commits-in-github-actions/

This would essentially be adding an additional unknown salt to the commit hash.

But also according to this article git is aware of and actively monitor commits for collisions:
https://github.blog/2017-03-20-sha-1-collision-detection-on-github-com/

markdewing: The script has "set -e" specified at the top, so the script will exit if this command exits with a non-zero error code.  And not submit the following approval code. Under what conditions would this push fail and what state would the PR be left in?

PDoakORNL:
Test this please.

PDoakORNL (approved):
I think the looks good now, but we wil need an outside of ORNL reviewer.  I'm also not up to date on whether the bot account accesses are setup properly so @walshmm should weigh in on that.

walshmm:
> I think the looks good now, but we wil need an outside of ORNL reviewer. I'm also not up to date on whether the bot account accesses are setup properly so @walshmm should weigh in on that.

I believe Paul granted it basic access, which should be enough because all the bot is doing is signing a commit.  The pushing and rebasing is done by the github actions token.  Post merge I would need to provide the secrets it uses so they can be stored and accessed via GithubActions, then probably open some test PR to give it a once over.

ye-luo (approved):
@PDoakORNL primarily looked at it and approved it.

ye-luo:
Test this please

PR #4131, opened 2022-07-26
Compute logarithm of rotation matrix

markdewing:
Add functions to compute the matrix log of a rotation matrix and extract the entries of the resulting antisymmetric matrix to a list of parameters.

This should allow us to recover rotation parameters from a rotation matrix.

- New feature

- No

desktop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/RotatedSPOs.cpp
+++ b/src/QMCWaveFunctions/RotatedSPOs.cpp
@@ -119,7 +119,7 @@ void RotatedSPOs::buildOptVariables(const RotationIndices& rotations)
   app_log() << "nparams_active: " << nparams_active << " params2.size(): " << params.size() << std::endl;
   if (params_supplied)
     if (nparams_active != params.size())
-      APP_ABORT("The number of supplied orbital rotation parameters does not match number prdouced by the slater "
+      throw std::runtime_error("The number of supplied orbital rotation parameters does not match number prdouced by the slater "
                 "expansion. \n");
 
   myVars.clear();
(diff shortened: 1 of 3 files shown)


prckent:
Mark - the code cov browser is timing out for me currently, but would it be reasonable to get the coverage a bit higher?

markdewing:
I'll have to look at the report when it's available, because I don't see how it's easily possible to increase the coverage. The only code that isn't hit are error checks (check for info after the call to geev and a check for a large imaginary part in the last loop).  And I'm not sure how to make those trigger.  Probably have to input a matrix that's not a rotation matrix.

I've run the round trip test over a grid of values. But that's the kind of test to run once, not as part of the unit tests that get run every time.

prckent:
Test this please

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -62,6 +62,20 @@ void RotatedSPOs::constructAntiSymmetricMatrix(const RotationIndices& rot_indice
   }
 }
 
+void RotatedSPOs::extractParamsFromAntiSymmetricMatrix(const RotationIndices& rot_indices,
+                                                       const ValueMatrix& rot_mat,
+                                                       std::vector<ValueType>& param)
+{
+  assert(rot_indices.size() == param.size());
+  // Assumes rot_mat is of the correct size and initialized to zero upon entry
```
PDoakORNL, line 70: Document these assumptions in the method declaration documentation.

PDoakORNL (changes requested):
Basically good to go.  Just fix the error handling move a little documentation.

PDoakORNL (approved):

ye-luo:
Test this please

PR #4135, opened 2022-07-27
Magnetization Density Observable for Spinors

rcclay:

## Proposed changes

This introduces a magnetization density estimator (legacy implementation) for use with non-collinear QMC calculations.  Use is very similar to existing density estimator.  Documentation on what is calculated and how to use in QMCPACK is added to the manual.  

- New feature
- Documentation changes

- No

Intel Xeon
## Checklist


PDoakORNL (commented):
I am against adding any new feature using the legacy design. This is going to need to be ported to the batched version before the legacy driver can be dropped.

prckent:
@rcclay This needs to be implemented and tested in the batched code. Please ask questions as needed about this.

jtkrogel:
@rcclay thanks for working on this.  Any chance for supporting real space and/or grid count based ways of specifying the histogram cells as in `SpinDensity` (i.e. `dr` and `grid`)?

Also, one question: since the spin variables are sampled in non-collinear QMC, why are additional/different integrals performed over these variables?  I thought this would only be needed for off-diagonal estimators like the momentum distribution and 1RDM.

jtkrogel:
Also, what are the physical systems you are testing the developments on?

rcclay:
@PDoakORNL I am working on the batched implementation of this right now--I'm taking full responsibility for that, and this legacy implementation will be superceded by the batch implementation before it becomes useless (before end of FY or sooner).  However, this was the fastest way for me to stand up this **unique** capability and let ORNL/SNL folks start using it and stress testing it in the field.  There's going to be a lot of potential theory and application stumbling blocks associated with this estimator that will need to be sussed out independently of the code/performance aspects, and it's probably best for people to get a head start on that. 

@prckent I agree.  

@jtkrogel Right now, this is a dumb generalization of the DensityEstimator.  Can you hit me with a wishlist of features/gridding schemes so that I can fold that into the batched implementation?  Also, will the old binning scheme work for your team to start doing science with it, or does this need to be moved up the priority queue?

Re: additional spin integral.  Basically, if I'm in an Sz basis, Sx and Sy are nondiagonal.  The way that looks when you shift to this over complete continuous spin representation is you pick up an integral.  I've attached the integral performed:  
image

PR #4136, opened 2022-07-28
Update SPOSet WFopt related APIs

ye-luo:
## Proposed changes
Once TrialWF is built, it calls checkInVariables to query variational parameters under the TrialWF tree regardless of if running WFOpt or not. This the desired behavior.
Consequentially, checkInVariables cannot be implemented as a feature blocker. Update SPOSet and its derived classes.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/SPOSet.h
+++ b/src/QMCWaveFunctions/SPOSet.h
@@ -119,8 +119,12 @@ public:
   /// reset parameters to the values from optimizer
   virtual void resetParameters(const opt_variables_type& optVariables) {}
 
-  /// check in/out parameters to the global list of parameters used by the optimizer
+  /** check in parameters to the global list of parameters used by the optimizer.
+   * This is a query function and should never be implemented as a feature blocker.
+   * If an SPOSet derived class doesn't support optimization, use the base class fallback.
+   */
   virtual void checkInVariables(opt_variables_type& active) {}
+  /// check out parameters to the global list of parameters used by the optimizer
   virtual void checkOutVariables(const opt_variables_type& active) {}
 
   virtual void evaluateDerivatives(ParticleSet& P,
--- a/src/QMCWaveFunctions/SPOSet.h
+++ b/src/QMCWaveFunctions/SPOSet.h
@@ -117,7 +117,7 @@ public:
     APP_ABORT(o.str());
   }
   /// reset parameters to the values from optimizer
-  virtual void resetParameters(const opt_variables_type& optVariables) = 0;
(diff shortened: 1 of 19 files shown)


PDoakORNL (commented):
The obvious question is does something else need to be added to block this combination or was this just a message for developers.

ye-luo:
There was a switch optimize at the determinant level and spo level. When there is a mismatch, need an error.
I'm moving away from this way because the determinant level doesn't include any parameter and should not own a switch.
Then there is no check needed since everything is decided by the spo. Simple.

PDoakORNL (dismissed):
LGTM

markdewing:
It looks like this PR also removes an unused feature for projecting the derivatives - multiplyDerivsByOrb and the "project" arguments.

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -117,10 +117,14 @@ class SPOSet : public QMCTraits
     APP_ABORT(o.str());
   }
   /// reset parameters to the values from optimizer
-  virtual void resetParameters(const opt_variables_type& optVariables) = 0;
+  virtual void resetParameters(const opt_variables_type& optVariables) {}
 
-  /// check in/out parameters to the global list of parameters used by the optimizer
+  /** check in parameters to the global list of parameters used by the optimizer.
+   * This is a query function and should never be implemented as a feature blocker.
```
markdewing, line 123: I don't quite understand what's happening by looking at this comment and just the PR.  `resetParameters` is the function that changed from needing to be implemented by all the derived classes to one with a default implementation. `checkInVariables` already has a default implementation.  So this comment references some change that isn't in this PR?

ye-luo: [LCAOrbitalSet.h](https://github.com/QMCPACK/qmcpack/pull/4136/files#diff-3008c4a5fdfa852f307be230c2d1e896b22871f8a9bf40fc2cdfd29be81da657)
I removed the blocker. So I added this comment to explain why.
I don't see a good reason to require implementing empty reset parameters.
There is a bigger picture I'm moving towards. https://github.com/ye-luo/qmcpack/wiki/Optimization-API-changes
It is still quite preliminary.

markdewing: I missed the removal in LCAOrbitalSet. D'oh.

PDoakORNL (approved):
LGTM

ye-luo:
Test this please

PR #4139, opened 2022-07-28
Introduce OptimizableObject

ye-luo:
## Proposed changes

Introduce OptimizableObject base class for all optimizable objects. Changed WFC but not yet SPOSet.
The final goal is to enable component selection in the WFopt driver.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/OptimizableObject.h
+++ b/src/QMCWaveFunctions/OptimizableObject.h
@@ -25,19 +25,15 @@ using opt_variables_type = optimize::VariableSet;
 class OptimizableObject
 {
 public:
-  OptimizableObject(const std::string& name, bool optimizable) : name_(name), optimizable_(optimizable) {}
+  OptimizableObject(const std::string& name) : name_(name) {}
 
   const std::string& getName() const { return name_; }
   bool isOptimized() const { return is_optimized_; }
-  bool isOptimizable() const { return optimizable_; }
 
 private:
   /** Name of the optimizable object
    */
   const std::string name_;
-  /** If true, this object and/or its underlying components is optimizable_
-   */
-  const bool optimizable_;
   /** If true, this object is actively modified during WFOpt
    */
   bool is_optimized_ = false;
@@ -49,16 +45,25 @@ public:
    * This is a query function and should never be implemented as a feature blocker.
    * If an OptimizableObject derived class doesn't support optimization, use the base class fallback.
    */
-  virtual void checkInVariables(opt_variables_type& active) {}
+  virtual void checkInVariables(opt_variables_type& active)
+  {
+    throw std::logic_error("BUG!! OptimizableObject::checkInVariables should not be called!");
+  }
 
   /** check out variational optimizable variables
    * @param active a super set of optimizable variables
    */
-  virtual void checkOutVariables(const opt_variables_type& active) {}
+  virtual void checkOutVariables(const opt_variables_type& active)
+  {
+    throw std::logic_error("BUG!! OptimizableObject::checkOutVariables should not be called!");
+  }
 
   /** reset the parameters during optimizations
    */
-  virtual void resetParameters(const opt_variables_type& active) {}
+  virtual void resetParameters(const opt_variables_type& active)
+  {
+    throw std::logic_error("BUG!! OptimizableObject::resetParameters should not be called!");
+  }
 
   /** print the state, e.g., optimizables */
   virtual void reportStatus(std::ostream& os) {}
--- a/src/QMCWaveFunctions/OptimizableObject.h
+++ b/src/QMCWaveFunctions/OptimizableObject.h
@@ -43,16 +43,16 @@ private:
   bool is_optimized_ = false;
 
 public:
-  /** check in optimizable parameters
+  /** check in variational parameters to the global list of parameters used by the optimizer.
    * @param active a super set of optimizable variables
    *
-   * Add the paramemters this WaveFunctionComponent manage to active.
+   * This is a query function and should never be implemented as a feature blocker.
+   * If an SPOSet derived class doesn't support optimization, use the base class fallback.
    */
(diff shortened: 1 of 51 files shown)


src/QMCWaveFunctions/OptimizableObject.h
```diff
@@ -0,0 +1,67 @@
...
+  /** If true, this object is actively modified during WFOpt
+   */
+  bool is_optimized_ = false;
+
+public:
+  /** check in variational parameters to the global list of parameters used by the optimizer.
+   * @param active a super set of optimizable variables
+   *
+   * This is a query function and should never be implemented as a feature blocker.
+   * If an SPOSet derived class doesn't support optimization, use the base class fallback.
```
markdewing, line 50: Mention of "SPOSet".  This should probably be OptimizableObject?

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -24,6 +24,7 @@
 #include "Particle/ParticleSet.h"
 #include "Particle/VirtualParticleSet.h"
 #include "QMCWaveFunctions/OrbitalSetTraits.h"
+#include "OptimizableObject.h"
```
PDoakORNL, line 27: Doesn't seem needed.

src/QMCWaveFunctions/WaveFunctionComponent.cpp
```diff
@@ -22,8 +22,8 @@ namespace qmcplusplus
 // for return types
 using PsiValueType = WaveFunctionComponent::PsiValueType;
 
-WaveFunctionComponent::WaveFunctionComponent(const std::string& class_name, const std::string& obj_name)
-    : Optimizable(true),
+WaveFunctionComponent::WaveFunctionComponent(const std::string& class_name, const std::string& obj_name, bool optimizable)
+    : OptimizableObject(obj_name, optimizable),
```
PDoakORNL, line 26: As far as I can tell we want to be able to ask a WaveFunctionComponent whether it isOptimizable() and If it is Optimizable we want to be able to call an Optimization interface corresponding to the component. 
I'd prefer something like

```
class OptimizableVars {
  virtual void checkInVariables(opt_variables_type& active) = 0;

  /** check out variational optimizable variables
   * @param active a super set of optimizable variables
   */
  virtual void checkOutVariables(const opt_variables_type& active) =0;

  /** reset the parameters during optimizations
   */
  virtual void resetParameters(const opt_variables_type& active) =0;

  /** print the state, e.g., optimizables */
  virtual void reportStatus(std::ostream& os) =0;
};

class WFC {
WFC(...) 
virtual bool isOptimizable() { return false; }
virtual OptimizableVars& getOptimizableVars() { throw LogicError; }
}

UPtrVec<OptimizableVars> extractOptimizationVars(RefVector<WFCS> wfcs) { 
  RefVector<OptimizableVars> opt_vars;
  for (auto& wfc : wfcs) {
    if wfc.isOptimizable()
      opt_vars.push_back(wfc.getOptimizationVars);
  }
  return opt_vars;
}
```

src/QMCWaveFunctions/OptimizableObject.h
```diff
@@ -0,0 +1,67 @@
...
+  bool isOptimizable() const { return optimizable_; }
+
+private:
+  /** Name of the optimizable object
+   */
+  const std::string name_;
+  /** If true, this object and/or its underlying components is optimizable_
+   */
+  const bool optimizable_;
+  /** If true, this object is actively modified during WFOpt
```
PDoakORNL, line 41: maybe you mean were.

src/QMCWaveFunctions/OptimizableObject.h
PDoakORNL: But the fall back is to do nothing.  Which means that by default classes deriving from OptimizableObject aren't optimizable.  I really think it would be better to get a vector of optimization inteface references belonging to actual optimizable objects and have that be the API.

A vector of "OptimizableObjects" that also includes many objects that cannot be optimized but have a default implementation of noop for optimization is less desirable.

ye-luo: Good catch.

ye-luo:
Test this please

PDoakORNL (changes requested):
I like the direction this is going in, but I don't like OptimizableObject.  If its just a stepping stone it's tolerable in the short term.

Other smaller changes especially to std algorithms I think would help with reading this.

src/QMCWaveFunctions/WaveFunctionComponent.cpp
ye-luo: The final product will be largely like this. The scan of optimizable parameters will be from TWF tree. That is exactly what you suggested TWF::extractOptimizationVars() and WFC::getOptimizableVars().

checkOutVariables and resetParameters are intended to be called not via TWF but directly from `RefVector<OptimizableObjects>` When this change is done, I can change `checkOutVariables/resetParameters` to pure virtual functions. The functionality of checkInVariables will be replaced with the survey function (extractOptimizationVars).

I'm going threw a few intermediate steps to make sure everything works as expected.

ye-luo: I updated OptimizableObject. isOptimizable() has been moved to WFC as virtual.
All the API base implementation are std::logic_error. Their change to pure virtual needs to wait until I finish updating WFOpt drivers.

src/QMCWaveFunctions/SPOSet.h
ye-luo: Need this to propagate typedef. Will connect SPOSet to OptimizableObject later.

PDoakORNL (approved):

PDoakORNL:
Test this please.

ye-luo:
Test this please

PR #4142, opened 2022-07-29
Fix regression affecting linemin and quartic

ye-luo:
## Proposed changes
Found via misbehaving test short-H4-opt-linear-linemin-16-1
Caused by da0b85b8f8a2f82b9b51471c1445e785527cdc1b

- Bugfix

- No

epyc-server

## Checklist


prckent (approved):
Good catch. While these are little used, this points to us needing a more optimizer testing in CI. We could have made a release with this code.

I am not enjoying how the internal names do not resemble the names of the input parameters in this old code...

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimize.cpp
```diff
@@ -122,7 +124,7 @@ QMCFixedSampleLinearOptimize::QMCFixedSampleLinearOptimize(MCWalkerConfiguration
   m_param.add(num_shifts, "num_shifts");
   m_param.add(cost_increase_tol, "cost_increase_tol");
   m_param.add(target_shift_i, "target_shift_i");
-
+  m_param.add(param_tol, "alloweddifference");
```
PDoakORNL, line 127: So if we had a unit test of an input class would the loss of alloweddifference as a parsed input value have been caught?  I tried to dig through some history here and it was unclear what caused the regression.

prckent: It would be easy for these things to sneak through. This value has a default which is not so different from what is used normally, i.e. it is a sensible default. For optional input parameters we would need to make sure in unit testing that we can both change the value and that something changes based on it.

ye-luo: I moved it from one class to another without moving the initialization and input support. Strangely I didn't get uninitialized value warning. Likely compiler guarantee the initial value 0.

PR #4144, opened 2022-07-29
Fix queryOrbitalsForSType due to reordering of ions.

### LINKED ISSUE #4092: CuspCorrection construction failure
**Describe the bug**
Cusp Correction scheme leads to outrageously wrong VMC energy (in the 10kHa).  Cusp corrected parameters are corrected (when compared to older version of the code) and the Updet and Downdet files are consistent. using generated files with latest versions of QMCPACK gives the very wrong energy. 
When tested with 8c175d587a35c5230e3deef395a50b07c352e32d on laptop and Cooley (ALCF machine), the same correction were produced but the construction of the correction leads to the right energies. 
Also, when no cusp correction scheme is used, VMC energy produces the same DFT energy (with a big variance).

  
**To Reproduce**
1. release version or git commit hash being built: 4e14d0bc327034cc3ac0c1bbed8978f320e8be2d-dirty
2. cmake command: 
3. full program/test invocation command: qmcpack cusp.xml | tee cusp.out
4. additional steps: Download input files [here](https://anl.box.com/s/worprbgpaa7f98ln6d2kroxtcylkkie2)

**Expected behavior**
files named updet.CuspInfo.xml and downdet.CuspInfo.xml are created but a 100 blocks VMC run makes energy diverge.

**System:**
 - Cooley (ALCF Haswell); Theta (ALCF  KNL); Laptop (Intel core i9)
 - other systems where this is reproducible: "my Laptop" "Cooley" "Theta"

**Additional context**
To get proper VMC energy use 8c175d587a35c5230e3deef395a50b07c352e32d
To get error, use latest versions.

ye-luo:
## Proposed changes
Fixes #4092. 3.13 and 3.14 releases are affected.
As @markdewing pointed out, it was caused by reordering and it seem I missed changing one function.
Still need a test, the current reproducer seems too heavy being used as a unit test.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/LCAO/SoaLocalizedBasisSet.cpp
+++ b/src/QMCWaveFunctions/LCAO/SoaLocalizedBasisSet.cpp
@@ -99,9 +99,9 @@ void SoaLocalizedBasisSet<COT, ORBT>::queryOrbitalsForSType(const std::vector<bo
                                                             std::vector<bool>& is_s_orbital) const
 {
   const auto& IonID(ions_.GroupID);
-  int idx = 0;
   for (int c = 0; c < NumCenters; c++)
   {
+    int idx = BasisOffset[c];
     int bss = LOBasisSet[IonID[c]]->BasisSetSize;
     std::vector<bool> local_is_s_orbital(bss);
     LOBasisSet[IonID[c]]->queryOrbitalsForSType(local_is_s_orbital);


prckent (approved):
Unfortuante that none of the tests or analysis tools caught this.

markdewing:
The existing ethanol unit test could be modified with a reordering of the atoms.

markdewing:
I found the files used to create that test - I can update the unit test.

ye-luo:
> I found the files used to create that test - I can update the unit test.

@markdewing Thanks. FYI, if atoms get reordered, the MO coefficients needs to be reordered accordingly.

ye-luo:
Test this please

PR #4146, opened 2022-08-01
[WIP] Orbital rotation optimization prototype

markdewing:
This implements option 2 in https://github.com/QMCPACK/qmcpack/issues/3983#issuecomment-1197380631
It saves a history of applied rotations.

One purpose of this PR is allow others to test orbital rotation (@anbenali ).   Variational parameters can be saved and restored from the vp.h5 file, though this is still a prototype.  RHF wavefunctions, where the up and down coefficients should be constrained to the be the same is not implemented. For testing I've used separate coefficient matrices and let them vary separately.

The problem with previous code is that checkInVariables is used to signal the beginning of an optimization cycle.  This function does not get called on all the clones, so it is not a good choice for this signal.
    
The rotation code depends on two things in the call to checkInVariables:
1. Sets the parameters to zero in order to get the parameter change when resetParameters gets called.
2. Stores the rotated coefficient matrix for later use.
    
The new code addresses these as:
1. Computes the delta between old parameters (in myVars) and the new parameters (in active).  This also has the effect that calling resetParameters multiple times with the same parameters will have no effect.
2. Always applies the rotation from the existing coefficient matrix - no copy is saved.   This should work for both the LCAO and spline code. (The spline code doesn't implement saving the coefficient matrix - with this change it does not need to).
    
One problem that arises is the coefficient matrix is stored as a shared pointer, and is shared among the clones.  If multiple threads call resetParameters, the rotation gets applied multiple times.

The hack to fix this here is to use the fact that checkInVariables (the irony!) is only called on one clone, and use that clone to perform the rotation.  This should be changed to something better before going into the code.

The mechanism to save and restore extra parameter information in the HDF file is to add two new functions in TWF (saveExtraParameters/readExtraParameters) and various pieces in the wavefunction hierarchy.
The hdf_archive object is pulled out of saveAsHDF/readFromHDF so it can be passed to save/readExtraParameters.

The SPOSet version of save/readExtraParameters has an id parameter to distinguish between up and down electrons.

- New feature

- No

laptop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
@@ -57,7 +57,8 @@ QMCCostFunctionBase::QMCCostFunctionBase(ParticleSet& w,
       m_wfPtr(NULL),
       m_doc_out(NULL),
       includeNonlocalH("no"),
-      debug_stream(0)
+      debug_stream(0),
+      do_override_output(true)
 {
   GEVType = "mixed";
   //paramList.resize(10);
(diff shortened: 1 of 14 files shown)


ye-luo (commented):
It will be better to split this PR into 1) introduce the new API with empty implementation. 2) do the change in RoratedSPO.
Regarding 1) read/saveExtraParameters can be added to OptimizableObject.
SPOSet should use its object name instead of an integer as its identifier.

PR #4152, opened 2022-08-04
Adding batched per particle value reporting for NonLocalECPotential

PDoakORNL:
## Proposed changes

This is another MS6.9 PR.  Now adding per particle energy reporting to the NonLocalECPotential.
I removed the makeClone method for NonLocalECPComponent which makes no sense since it is never type erased, now a proper constructor takes its place which allows use of std::make_unique which makes more sense.

There is some questionable raw pointer use in NonLocalECPComponent that should be addressed at some point. I don't think this makes it worse but it still carries the potential to create a double free if there is a problem during copy construction or cloning.

Rather than add to the misnamed and legacy only integration testing in `test_ecp.cpp`. I added a closer to actual unit test test.

It also shows a real lack of sensistivity to the randomization of the pp grid which should perhaps become an issue.

- New feature
- Build related changes
- Documentation changes

- No

Leconte

## Checklist

### DIFF
--- a/src/QMCHamiltonians/Listener.hpp
+++ b/src/QMCHamiltonians/Listener.hpp
@@ -59,6 +59,17 @@ private:
   const std::string name_;
 };
 
+template<typename T>
+struct ListenerOption
+{
+    ListenerOption(const std::vector<ListenerVector<T>>& le, const std::vector<ListenerVector<T>>& li)
+        : electrons(le), ions(li)
+    {}
+    const std::vector<ListenerVector<T>>& electrons;
+    const std::vector<ListenerVector<T>>& ions;
+};
+
+
 
 } // namespace qmcplusplus
 
--- a/src/QMCHamiltonians/Listener.hpp
+++ b/src/QMCHamiltonians/Listener.hpp
@@ -59,6 +59,13 @@ private:
   const std::string name_;
 };
 
+/** Convenience container for common optional element to mw_eval.._impl.
+ *  This can allow the per_particle and reduced mw_eval to share the same
+ *  implementation method.
+ *
+ *  see NonLocalECPotential for example of usage.
+ *  I anticipate it could be used elsewhere.
+ */
 template<typename T>
 struct ListenerOption
 {
--- a/src/QMCHamiltonians/NonLocalECPComponent.cpp
+++ b/src/QMCHamiltonians/NonLocalECPComponent.cpp
@@ -243,7 +243,6 @@ void NonLocalECPComponent::mw_evaluateOne(const RefVectorWithLeader<NonLocalECPC
   else
   {
     // Compute ratios without VP. This is working but very slow code path.
-#pragma omp parallel for
     for (size_t i = 0; i < p_list.size(); i++)
     {
       NonLocalECPComponent& component(ecp_component_list[i]);
--- a/src/QMCHamiltonians/NonLocalECPComponent.cpp
+++ b/src/QMCHamiltonians/NonLocalECPComponent.cpp
@@ -30,9 +30,9 @@ NonLocalECPComponent::NonLocalECPComponent(const NonLocalECPComponent& nl_ecpc, 
     : NonLocalECPComponent(nl_ecpc)
 {
   for (int i = 0; i < nl_ecpc.nlpp_m.size(); ++i)
-    this->nlpp_m[i] = nl_ecpc.nlpp_m[i]->makeClone();
+    nlpp_m[i] = nl_ecpc.nlpp_m[i]->makeClone();
   if (nl_ecpc.VP)
-    this->VP = new VirtualParticleSet(pset, nknot);
+    VP = new VirtualParticleSet(pset, nknot);
 }
 
 NonLocalECPComponent::~NonLocalECPComponent()
--- a/src/QMCHamiltonians/NonLocalECPComponent.cpp
+++ b/src/QMCHamiltonians/NonLocalECPComponent.cpp
@@ -24,22 +24,22 @@ namespace qmcplusplus
 {
 NonLocalECPComponent::NonLocalECPComponent() : lmax(0), nchannel(0), nknot(0), Rmax(-1), VP(nullptr) {}
 
+// unfortunately we continue the sloppy use of the default copy constructor followed by reassigning pointers.
+// This prevents use of smart pointers and concievably sets us up for trouble with double frees and the destructor.
+NonLocalECPComponent::NonLocalECPComponent(const NonLocalECPComponent& nl_ecpc, const ParticleSet& pset) : NonLocalECPComponent(nl_ecpc)
+{
+  for (int i = 0; i < nl_ecpc.nlpp_m.size(); ++i)
+    this->nlpp_m[i] = nl_ecpc.nlpp_m[i]->makeClone();
+  if (nl_ecpc.VP)
+    this->VP = new VirtualParticleSet(pset, nknot);
+}
+
 NonLocalECPComponent::~NonLocalECPComponent()
 {
   for (int ip = 0; ip < nlpp_m.size(); ip++)
     delete nlpp_m[ip];
   if (VP)
     delete VP;
-}
-
-NonLocalECPComponent* NonLocalECPComponent::makeClone(const ParticleSet& qp)
-{
-  NonLocalECPComponent* myclone = new NonLocalECPComponent(*this);
-  for (int i = 0; i < nlpp_m.size(); ++i)
-    myclone->nlpp_m[i] = nlpp_m[i]->makeClone();
-  if (VP)
-    myclone->VP = new VirtualParticleSet(qp, nknot);
-  return myclone;
 }
 
 void NonLocalECPComponent::initVirtualParticle(const ParticleSet& qp)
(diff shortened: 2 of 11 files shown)


src/QMCHamiltonians/Listener.hpp
```diff
@@ -59,6 +59,17 @@ class ListenerVector
   const std::string name_;
 };
 
+template<typename T>
+struct ListenerOption
```
ye-luo, line 63: Need some documentation about this class.

src/QMCHamiltonians/NonLocalECPComponent.cpp
```diff
@@ -24,6 +24,16 @@ namespace qmcplusplus
 {
 NonLocalECPComponent::NonLocalECPComponent() : lmax(0), nchannel(0), nknot(0), Rmax(-1), VP(nullptr) {}
 
+// unfortunately we continue the sloppy use of the default copy constructor followed by reassigning pointers.
+// This prevents use of smart pointers and concievably sets us up for trouble with double frees and the destructor.
+NonLocalECPComponent::NonLocalECPComponent(const NonLocalECPComponent& nl_ecpc, const ParticleSet& pset) : NonLocalECPComponent(nl_ecpc)
```
ye-luo, line 29: NonLocalECPComponent doesn't need inheritance and thus a specialized copy constructor is better. Yes a bit unfortunate that we would like to to minimal change.

src/QMCHamiltonians/NonLocalECPotential.h
```diff
@@ -2,37 +2,57 @@
...
+
+  struct NonLocalECPotentialMultiWalkerResource : public Resource
+  {
+    NonLocalECPotentialMultiWalkerResource() : Resource("NonLocalECPotential") {}
+
+    Resource* makeClone() const override;
+
+    ResourceCollection collection{"NLPPcollection"};
+
+    int num_walker = 0;
```
ye-luo, line 50: Really need this? It is a high risk state. Can we keep the forward declaration and move the full definition back to cpp.

src/QMCHamiltonians/NonLocalECPotential.cpp
```diff
@@ -261,9 +284,10 @@ void NonLocalECPotential::mw_evaluateImpl(const RefVectorWithLeader<OperatorBase
     if (Tmove)
       O.tmove_xy_.clear();
 
-    for (int ipp = 0; ipp < O.PPset.size(); ipp++)
-      if (O.PPset[ipp])
-        O.PPset[ipp]->randomize_grid(*O.myRNG);
+    if (!keep_grid)
+      for (int ipp = 0; ipp < O.PPset.size(); ipp++)
+	if (O.PPset[ipp])
+	  O.PPset[ipp]->randomize_grid(*O.myRNG);
```
ye-luo, line 290: Formatting.

src/QMCHamiltonians/tests/test_NonLocalECPotential.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+  ParticleSet elec(simulation_cell);
+  elec.setName("elec");
+  elec.create({2, 1});
+  elec.R[0][0] = 0.4;
+  elec.R[0][1] = 0.0;
+  elec.R[0][2] = 0.0;
+  elec.R[1][0] = 1.0;
+  elec.R[1][1] = 0.0;
+  elec.R[1][2] = 0.0;
```
ye-luo, line 101: Please write in more compact ways `elec.R[1] = {1.0, 2.0, 3.0}`

src/QMCHamiltonians/tests/test_NonLocalECPotential.cpp
```diff
@@ -0,0 +1,221 @@
...
+  CHECK(std::accumulate(local_pots.begin(), local_pots.begin() + local_pots.cols(), 0.0) == Approx(value));
+  CHECK(std::accumulate(local_pots2.begin(), local_pots2.begin() + local_pots2.cols(), 0.0) == Approx(value));
+  CHECK(std::accumulate(ion_pots.begin(), ion_pots.begin() + ion_pots.cols(), 0.0) == Approx(10.5313520432));
+  CHECK(std::accumulate(ion_pots2.begin(), ion_pots2.begin() + ion_pots2.cols(), 0.0) == Approx(10.5313520432));
+
+  CHECK(!testing::TestNonLocalECPotential::didGridChange(nl_ecp));
+
+  elec.R[0][0] = 0.5;
+  elec.R[0][1] = 0.0;
+  elec.R[0][2] = 2.0;
```
ye-luo, line 203: Please write in more compact ways `elec.R[1] = {1.0, 2.0, 3.0}`

ye-luo (changes requested):

prckent:
test_NonLocalECPotential.cpp is giving NaNs

src/QMCHamiltonians/NonLocalECPComponent.cpp
PDoakORNL: The alternative is to have the incorrect refPS for the NonLocalECPComponent contained VP set. Seems like that would result in the wrong distance tables being used and infact a possible race on the differerent tables.

PDoakORNL: The real solution to this is that no object constructed via cloning of the golden walker elements should retain any reference to another golden element and arguments needed for an evaluation from other elements need to be explicitly in the API.

src/QMCHamiltonians/NonLocalECPotential.h
PDoakORNL: Mistake on my part.  I've moved the definition back into the cpp, but it differs from the treatment of practically every other MultiWalkerResource.  Since it lets us make changes without causing as big a recompilation it is in principle good.

PDoakORNL:
Test this please.

prckent:
Test this please

markdewing:
Test this please

src/QMCHamiltonians/NonLocalECPComponent.cpp
```diff
@@ -24,6 +24,17 @@ namespace qmcplusplus
 {
 NonLocalECPComponent::NonLocalECPComponent() : lmax(0), nchannel(0), nknot(0), Rmax(-1), VP(nullptr) {}
 
+// unfortunately we continue the sloppy use of the default copy constructor followed by reassigning pointers.
+// This prevents use of smart pointers and concievably sets us up for trouble with double frees and the destructor.
+NonLocalECPComponent::NonLocalECPComponent(const NonLocalECPComponent& nl_ecpc, const ParticleSet& pset)
```
ye-luo, line 29: There is no polymorphism around NonLocalECPComponent and thus makeClone was unnecessary.

src/QMCHamiltonians/NonLocalECPComponent.cpp
```diff
@@ -24,6 +24,17 @@ namespace qmcplusplus
...
+    : NonLocalECPComponent(nl_ecpc)
+{
+  for (int i = 0; i < nl_ecpc.nlpp_m.size(); ++i)
+    this->nlpp_m[i] = nl_ecpc.nlpp_m[i]->makeClone();
+  if (nl_ecpc.VP)
+    this->VP = new VirtualParticleSet(pset, nknot);
```
ye-luo, line 35: Please remove the above two "this->"

src/QMCHamiltonians/NonLocalECPotential.cpp
```diff
@@ -368,24 +398,64 @@ void NonLocalECPotential::mw_evaluateImpl(const RefVectorWithLeader<OperatorBase
...
+        {
+          auto& ve_samples = O_leader.mw_res_->ve_samples;
+          auto& vi_samples = O_leader.mw_res_->vi_samples;
+          // CAUTION! This may not be so simple in the future
+          int iw = j;
+          ve_samples(iw, batch_list[j].get().electron_id) += pairpots[j];
+          vi_samples(iw, batch_list[j].get().ion_id) += pairpots[j];
+        }
+
+#ifndef DEBUG_NLPP_BATCHED
```
ye-luo, line 420: ifdef and remove "if(false)"?

src/QMCHamiltonians/NonLocalECPotential.cpp
```diff
@@ -368,24 +398,64 @@ void NonLocalECPotential::mw_evaluateImpl(const RefVectorWithLeader<OperatorBase
...
+      Vector<Real> ve_sample(ve_samples.begin(iw), num_electrons);
+      Vector<Real> vi_sample(vi_samples.begin(iw), O_leader.NumIons);
+      for (const ListenerVector<Real>& listener : listeners->electrons)
+      {
+        listener.report(iw, O_leader.getName(), ve_sample);
+      }
+      for (const ListenerVector<Real>& listener : listeners->ions)
+      {
+        listener.report(iw, O_leader.getName(), vi_sample);
+      }
```
ye-luo, line 454: remove {} around single lines.

ye-luo (approved):

ye-luo:
Test this please

PR #4153, opened 2022-08-05
Math::FFTW3 and Math::BLAS_LAPACK should not be exposed inside modules

ye-luo:
## Proposed changes
It is a CMake internal change, not affecting users.
Having FindMKL/FindFFTW modules directly accessing the global Math::BLAS_LAPACK and Math::FFTW3 goes against modularization. Thus change the behavior. Create a module specific target and make the global one depends on it.

PS: there are also minor changes replace set properties with corresponding target_ command.

- Build related changes

- No

epyc-server

## Checklist


ye-luo:
Test this please

ye-luo:
Note: our FindFFTW3.cmake was not well written. Will need another pass.

quantumsteve (approved):
Improvement to FindMKL and FindFFTW.

ye-luo:
Test this please

PR #4158, opened 2022-08-06
Fix excited states in splines when spin-up/down sets are built from the same spin species and occupation is specified on the first sposet

ye-luo:
## Proposed changes
I noticed this bug when I tried to delete the implicit sposet sharing by spin-up/down electron.

When spin-up/down sets are built from the same spin specie and the occupation is specified on the first sposet, the spin-up set will be built correctly but the spin-down set will misidentify itself the same as the up set and thus the excited set will be used for both up/down spin electrons. Current test reference numbers are all bad.

@Hyeondeok-Shin 
**Need to update reference number of all the excited states tests using splines**

A quick scan shows examples in the manual load two sets of orbitals from different spin. So it seems fine @Hyeondeok-Shin please double check the manual entries.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
all the reference values for the deterministic tests has been updated. short/long test will be updated following #4162

prckent (approved):
Thanks Ye & Hyeondeok.

I normally make a comment about avoiding these bugs in future, but if the implemented logic and tests are both wrong and consistent with each other there is not much more we can do, except things like having few options and therefore fewer lines and fewer chances for issues.

prckent:
From the blame this bug looks to have existed for a long time....

ye-luo:
Test this please

PR #4159, opened 2022-08-08
Avoid creating driver scope walkers in the population object of WFOpt

ye-luo:

## Proposed changes
pset, twf, ham objects are maintained in CostFunctionCrowdData.

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCDrivers/DMC/DMCBatched.cpp
+++ b/src/QMCDrivers/DMC/DMCBatched.cpp
@@ -380,7 +380,7 @@ void DMCBatched::process(xmlNodePtr node)
                                 qmcdriver_input_.get_walkers_per_rank(), dmcdriver_input_.get_reserve(),
                                 qmcdriver_input_.get_num_crowds());
 
-    Base::startup(node, awc);
+    Base::startup(awc);
   }
   catch (const UniformCommunicateError& ue)
   {
--- a/src/QMCDrivers/DMC/DMCBatched.cpp
+++ b/src/QMCDrivers/DMC/DMCBatched.cpp
@@ -380,7 +380,7 @@ void DMCBatched::process(xmlNodePtr node)
                                 qmcdriver_input_.get_walkers_per_rank(), dmcdriver_input_.get_reserve(),
                                 qmcdriver_input_.get_num_crowds());
 
-    Base::startup(awc);
+    Base::initializeQMC(awc);
   }
   catch (const UniformCommunicateError& ue)
   {
--- a/src/QMCDrivers/MCPopulation.cpp
+++ b/src/QMCDrivers/MCPopulation.cpp
@@ -257,14 +257,6 @@ void MCPopulation::measureGlobalEnergyVariance(Communicate& comm,
   variance = weight_energy_variance[2] / weight_energy_variance[0] - ener * ener;
 }
 
-void MCPopulation::set_variational_parameters(const opt_variables_type& active)
-{
-  for (auto it_twfs = walker_trial_wavefunctions_.begin(); it_twfs != walker_trial_wavefunctions_.end(); ++it_twfs)
-  {
-    (*it_twfs).get()->resetParameters(active);
-  }
-}
-
 void MCPopulation::checkIntegrity() const
 {
   // check active walkers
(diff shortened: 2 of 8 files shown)


markdewing:
The function 'set_variational_parameters' was added in #2562 and the call the QMCDriverNew::startup was added in #2853.   Either those calls weren't necessary then, or something has changed in the code so they aren't needed now.

Where do walkers get created and variational parameters updated for the call to VMCBatch::run in QMCFixedSampleLinearOptimizedBatched::generateSamples?

ye-luo:
QMCDriverNew::startup creates walker objects in population_ but QMCFixedSampleLinearOptimizedBatched doesn't use those objects. Thus skip creation. Instead QMCFixedSampleLinearOptimizedBatched create and maintain those objects in CostFunctionCrowdData.

The resetParameter to the gold reference and crowd member of happens at
https://github.com/QMCPACK/qmcpack/blob/687b3db43c1fadd063c0d461995d777c5b49809b/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp#L482

markdewing:
In #2817, the creation of vmcEngine creates a new MCPopulation object rather than reusing and referencing the one in the optimizer driver.  That explains why "set_variational_parameters" is no longer needed.

However, QMCDriverNew::startup does more than that.  Working through the list of what it does:
* Sets up `population_`  - already determined this isn't used by the optimizer
* Sets up `crowds_` - not used in the optimizer
* Sets up rng step contexts - resizes Rng and step_contexts_.   The `step_contexts_` variable is not used in the optimizer, and the Rng's get taken from the VMC driver, not the optimizer driver (in `QMCFixedSampleLinearOptimizer::start`, the call to `optTarget->setRng` uses the Rng's from vmcEngine)

So it does look like there is no need to call QMCDriverNew::startup on the optimizer driver.

markdewing:
It's troubling from a design point of view that there is this function in the base class, `startup`, that isn't necessary for one the drivers, and the only way to determine that is to go through it line-by-line.  I think part of the problem is `QMCDriverNew` is used for two purposes  - one is the interface to 'algorithms to execute', used by QMCMain at the high level for dispatching the setup and execution, and the other is the common base for QMC algorithms (things that VMC, DMC, RMC, etc, share).  A better design would be to separate these two purposes.   (Not suggesting that we make that change soon - it would be a low priority - but wanted to record some analysis of this issue)

markdewing (dismissed):

ye-luo:
@markdewing thanks for your analysis. It is accurate. I updated the docs and removed unused xml argument in startup. Please re-approve this PR.

ye-luo:
Test this please

markdewing (dismissed):

prckent:
Can we change startup to something more descriptive, e.g. startupSomething /setupQMC / initializeQMC / ... ?

PDoakORNL:
QMCDriverNew was not designed to be more than the base for VMC/DMC and there mostly as a convenience/as an iterative move from the legacy.  The interface and the shared implementation used by VMC/DMC batched should certainly be split.  And I'm not particularly happy with the startup and process methods.  No one should consider this implementation detail "good" or something the not be refactor improved over time.

In my opinion there is a shared interface used by the QMCMain here that the optimizers should also inherit but that is it.  The optimizers should be getting constructucted with a QMCOptimizationDriverInput that may be paired with its own QMCDriverInput it should have child nodes of QMCDriverInput and VMCInput for the enclosed driver used in its optimization loops.

PDoakORNL (dismissed):
LGTM this follows through on one of the important todo's for the batched driver.

Next should be removal of the xml from the process call, perhaps removal of the process state transformation altogether.

ye-luo:
Agree with @markdewing. As @PDoakORNL mentioned, this startup() only a stop gap intended for VMC/DMC only. We need to think about how to share codes between VMC and DMC without making it to the base/interface class.

ye-luo:
I renamed the function from startup to initializeQMC and moved it under "protected" but this is likely not final.

ye-luo:
Test this please

prckent (approved):

PR #4160, opened 2022-08-08
Require named SPOSet

ye-luo:
Review can be done independent from  #4158 but need its fix of tests before merging.

## Proposed changes
The major change is in SPOSet.h
1. require it named
2. isOptimizable isOptimizable hasIonDerivs getClassName are virtual functions instead of relying on member variables.
3. make it inherited from OptimizableObject. This change will be soon reverted as OptimizableObject will be used in actual optimizable derived classes. The change will come with a later PR. 
4. in SPOSet builders, implicit sharing SPOSet for spin-up/down has been removed. Users may hit memory overhead if they are still using legacy input style instead of sposet_collection. I think it is time to phase out the legacy input style with SPOSet constructed inside determinants.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/EinsplineSpinorSetBuilder.cpp
+++ b/src/QMCWaveFunctions/EinsplineSpinorSetBuilder.cpp
@@ -132,10 +132,12 @@ std::unique_ptr<SPOSet> EinsplineSpinorSetBuilder::createSPOSetFromXML(xmlNodePt
   H5OrbSet aset(H5FileName, spinSet, numOrbs);
   const auto iter = SPOSetMap.find(aset);
   if ((iter != SPOSetMap.end()) && (!NewOcc))
-    app_warning() << "!!!!!!! Identical SPOSet detected in EinsplineSpinorSetBuilder! "
-                     "Implicit sharing one SPOSet for spin-up and spin-down electrions has been removed. "
-                     "Create a single SPO set outside determinantset instead. "
-                     "Use sposet_collection to construct an explicit sposet for explicit sharing."
+    app_warning() << "!!!!!!! Identical SPOSets are detected by EinsplineSpinorSetBuilder! "
+                     "Implicit sharing one SPOSet for spin-up and spin-down electrons has been removed. "
+                     "Each determinant creates its own SPOSet with dedicated memory for spline coefficients. "
+                     "To avoid increasing the memory footprint of spline coefficients, "
+                     "create a single SPOset outside the determinantset using 'sposet_collection' "
+                     "and reference it by name on the determinant line."
                   << std::endl;
 
   if (FullBands[spinSet] == 0)
--- a/src/QMCWaveFunctions/EinsplineSpinorSetBuilder.cpp
+++ b/src/QMCWaveFunctions/EinsplineSpinorSetBuilder.cpp
@@ -49,6 +49,8 @@ std::unique_ptr<SPOSet> EinsplineSpinorSetBuilder::createSPOSetFromXML(xmlNodePt
   std::string spo_prec("double");
   std::string truncate("no");
   std::string hybrid_rep("no");
+  std::string spo_object_name;
+
   ScopedTimer spo_timer_scope(*timer_manager.createTimer("einspline::CreateSpinorSetFromXML", timer_level_medium));
 
   {
@@ -90,6 +92,8 @@ std::unique_ptr<SPOSet> EinsplineSpinorSetBuilder::createSPOSetFromXML(xmlNodePt
   {
     OhmmsAttributeSet oAttrib;
     oAttrib.add(spinSet, "spindataset");
+    oAttrib.add(spo_object_name, "name");
+    oAttrib.add(spo_object_name, "id");
     oAttrib.put(cur);
   }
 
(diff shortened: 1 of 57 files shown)


src/QMCWaveFunctions/EinsplineSpinorSetBuilder.cpp
```diff
@@ -126,14 +130,13 @@ std::unique_ptr<SPOSet> EinsplineSpinorSetBuilder::createSPOSetFromXML(xmlNodePt
...
-  iter = SPOSetMap.find(aset);
+  const auto iter = SPOSetMap.find(aset);
   if ((iter != SPOSetMap.end()) && (!NewOcc))
-  {
-    app_log() << "SPOSet parameters match in EinsplineSetBuilder:  "
-              << "cloning EinsplineSet object.\n";
-    return std::unique_ptr<SPOSet>{iter->second->makeClone()};
-  }
+    app_warning() << "!!!!!!! Identical SPOSet detected in EinsplineSpinorSetBuilder! "
+                     "Implicit sharing one SPOSet for spin-up and spin-down electrions has been removed. "
```
prckent, line 136: electrons

prckent (commented):
What are the relevant considerations regarding backwards compatibility (or not) here?

ye-luo:
> What are the relevant considerations regarding backwards compatibility (or not) here?

The deprecation warning has been printed for quite a few releases. With the change from this PR, those input files will continue to run but with memory penalty. We continue issue warnings to users reminding them to change input style. I have not decided to turn the warning to error.

prckent (approved):
LGTM but please wait for @PDoakORNL 

Happy to see the appearance of accessor functions

src/QMCWaveFunctions/ElectronGas/FreeOrbitalBuilder.cpp
```diff
@@ -13,10 +13,12 @@ FreeOrbitalBuilder::FreeOrbitalBuilder(ParticleSet& els, Communicate* comm, xmlN
 std::unique_ptr<SPOSet> FreeOrbitalBuilder::createSPOSetFromXML(xmlNodePtr cur)
 {
   int npw, norb;
+  std::string spo_object_name;
   PosType twist(0.0);
   OhmmsAttributeSet attrib;
   attrib.add(norb, "size");
   attrib.add(twist, "twist");
+  attrib.add(spo_object_name, "name");
```
PDoakORNL, line 21: If id is intended to be an alternate input tag for the spo_object_name it is missing here.

src/QMCWaveFunctions/ElectronGas/FreeOrbital.h
```diff
@@ -24,9 +24,11 @@ namespace qmcplusplus
 class FreeOrbital : public SPOSet
 {
 public:
-  FreeOrbital(const std::vector<PosType>& kpts_cart);
+  FreeOrbital(const std::string& my_name, const std::vector<PosType>& kpts_cart);
   ~FreeOrbital();
 
+  std::string getClassName() const override { return "FreeOrbital"; }
```
PDoakORNL, line 30: If this is just used for error reporting/debugging, which is what I see, its fine if we need runtime reflection I don't think we should do it via string matching.

src/QMCWaveFunctions/SPOSetBuilder.cpp
```diff
@@ -110,23 +111,12 @@ std::unique_ptr<SPOSet> SPOSetBuilder::createSPOSet(xmlNodePtr cur)
...
-      sposet_ref.setName(spo_object_name + "_before_rotation");
-
-    // overwrite sposet
     sposet = std::move(rot_spo);
 #endif
   }
 
-  if (!spo_object_name.empty() && sposet->getName().empty())
-    sposet->setName(spo_object_name);
+  //if (!spo_object_name.empty() && sposet->getName().empty())
```
PDoakORNL, line 118: Delete this

PDoakORNL (changes requested):
Just couple of proofreading changes and a question about `name` and `id` tags otherwise this looks like a good step.

src/QMCWaveFunctions/ElectronGas/FreeOrbitalBuilder.cpp
ye-luo: In a couple of places, `name` and `id` are interchangeable. In this PR, I added both because there are existing input files. It is unclear how we would like to move forward. My current thought is to disable "id" support and allow "name" only. Supporting both requires do a lot of extra things, for example trapping errors when "name" and "id" differ.

For now, we need both both. Here is what I will do. 
```
attrib.add(spo_object_name, "name");
attrib.add(spo_object_name, "id", {}, TagStatus::DEPRECATED)
```
and later change to TagStatus::DELETED.

I will do a separate to address this change.

src/QMCWaveFunctions/ElectronGas/FreeOrbital.h
ye-luo: Agreed. If we need to make selection based on derived class types, we should do enum not strings. It is only intended for printing.

src/QMCWaveFunctions/SPOSetBuilder.cpp
ye-luo: Please allow me to do it in my next PR.

PDoakORNL (approved):

williamfgc:
Test this please

ye-luo:
Test this please

PR #4164, opened 2022-08-09
Fix cusp correction bug

ye-luo:
## Proposed changes
Breakage since #4053. No release got affected.
The modified unit test fails on current develop but pass with this PR.
I also added a bit more safety guard

- Bugfix

- No

epyc-server

## Checklist


src/QMCWaveFunctions/LCAO/SoaCuspCorrection.cpp
```diff
@@ -111,25 +104,22 @@ void SoaCuspCorrection::evaluate_vgl(const ParticleSet& P,
...
       LOBasisSet[c]->evaluate_vgl(dist[c], displ[c], myVGL[0], myVGL[1], myVGL[2], myVGL[3], myVGL[4]);
-    }
-  }
 
   const auto v_in  = myVGL[0];
   const auto gx_in = myVGL[1];
   const auto gy_in = myVGL[2];
   const auto gz_in = myVGL[3];
   const auto l_in  = myVGL[4];
-  for (size_t i = 0; i < psi.rows(); ++i)
```
ye-luo, line 132: This is the line of bug.

src/QMCWaveFunctions/tests/test_soa_cusp_corr.cpp
```diff
@@ -528,9 +528,9 @@ TEST_CASE("Ethanol MO with cusp", "[wavefunction]")
   SPOSet::ValueMatrix all_values;
   SPOSet::GradMatrix all_grad;
   SPOSet::ValueMatrix all_lap;
-  all_values.resize(13, 13);
-  all_grad.resize(13, 13);
-  all_lap.resize(13, 13);
+  all_values.resize(10, 13);
```
markdewing, line 531: Where does the "10" come from?
The following call to "evaluate_notranspose" uses 0,7 for first,last.  For ethanol, that should be 0,14. I think the code was copied from the HCN test and those values were not updated to match.  However, it could be left as-is to test a case where there are fewer electrons than orbitals.

ye-luo: You can consider it as a non-ground state calculation with only 10 electron in one spin.

ye-luo: From evaluate_notranspose API with first, last, there is no restriction requiring the container size to match exactly. So 0.7 can be considered as a corner case test.

ye-luo: @markdewing looks good?

prckent (approved):
Assuming @markdewing gives assent.

markdewing (approved):

src/QMCWaveFunctions/tests/test_soa_cusp_corr.cpp
markdewing: Yes, keeping these values to test non-standard cases is good.

ye-luo:
Test this please

PR #4168, opened 2022-08-11
Update OptimizableObject and its usage

ye-luo:
## Proposed changes
Add OptimizableObject as a base class for classes directly owning variational parameters.
No more as base class of WFC/SPOSet.
Following this API design.
https://github.com/ye-luo/qmcpack/wiki/Optimization-API-changes

- No

epyc-server

## Checklist


src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.cpp
```diff
@@ -23,19 +23,18 @@ namespace qmcplusplus
...
-    : WaveFunctionComponent("MultiSlaterDetTableMethod", ""),
-      RatioTimer(*timer_manager.createTimer(ClassName + "::ratio")),
-      offload_timer(*timer_manager.createTimer(ClassName + "::offload")),
-      EvalGradTimer(*timer_manager.createTimer(ClassName + "::evalGrad")),
-      RatioGradTimer(*timer_manager.createTimer(ClassName + "::ratioGrad")),
-      PrepareGroupTimer(*timer_manager.createTimer(ClassName + "::prepareGroup")),
-      UpdateTimer(*timer_manager.createTimer(ClassName + "::updateBuffer")),
-      AccRejTimer(*timer_manager.createTimer(ClassName + "::Accept_Reject")),
-      EvaluateTimer(*timer_manager.createTimer(ClassName + "::evaluate")),
+    : OptimizableObject("multidet_coefs"),
```
ye-luo, line 26: Unlike Jastrow factor functors. There is only one set of multi det coefficients and thus the name is hard-coded. This is the tag to select this component.
"multidet_coefs" is just bad.
"msd_coefs"? Let me know your preference.

ye-luo: Probably it should be "ci_coefs"

ye-luo:
@PDoakORNL in this PR OptimizableObject is moved out of base classes and added as a base class only on their derived classes with actual optimizable parameters.

markdewing:
Is there a plan to get rid of checkInVariables and resetParameters in favor of using only checkInVariablesExclusive and resetParametersExclusive? (In other words, is having both functions in the API an intermediate state of the code?)

markdewing:
Eventually I would like an API that removes the need to call checkOutVariables to propagate the global-to-local index mapping.

ye-luo:
> Is there a plan to get rid of checkInVariables and resetParameters in favor of using only checkInVariablesExclusive and resetParametersExclusive? (In other words, is having both functions in the API an intermediate state of the code?)

Yes. SPO/WFC level of checkInVariables and resetParameters will be removed. Not there yet ;)

ye-luo:
> Eventually I would like an API that removes the need to call checkOutVariables to propagate the global-to-local index mapping.

I found checkOutVariables is still needed, it settings internal state for later evaluateDerivatives. In case of ROHF, there is only one set of sposet but there are two objects. So checkOutVariables needs to be called on both objects and thus needs to be called from TWF tree.

src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
```diff
@@ -397,20 +402,20 @@ bool QMCCostFunctionBase::put(xmlNodePtr q)
...
+    app_log() << "   '" << obj.getName() << "'" << (obj.isOptimized() ? " optimized" : " fixed") << std::endl;
 
   //build optimizables from the wavefunction
   OptVariablesForPsi.clear();
-  Psi.checkInVariables(OptVariablesForPsi);
+  for (OptimizableObject& obj : opt_obj_refs)
+    if (obj.isOptimized())
+      obj.checkInVariablesExclusive(OptVariablesForPsi);
   OptVariablesForPsi.resetIndex();
+  app_log() << "Totally " << OptVariablesForPsi.size() << " parameters being optimzied." << std::endl;
```
PDoakORNL, line 417:
```
<< "This will be radical"
```

Or "In total" for "Totally"

src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
```diff
@@ -1058,6 +1063,27 @@ bool QMCCostFunctionBase::isEffectiveWeightValid(EffectiveWeight effective_weigh
   return true;
 }
 
+UniqueOptObjRefs QMCCostFunctionBase::extractOptimizableObjects(TrialWaveFunction& psi) const
+{
+  const auto& names(variational_subset_names);
+  /// survey all the optimizable objects
+  const auto opt_obj_refs = psi.extractOptimizableObjectRefs();
+  for (OptimizableObject& obj : opt_obj_refs)
+    obj.setOptimization(names.empty() || std::find_if(names.begin(), names.end(), [&](const std::string& name) {
```
PDoakORNL, line 1072: This took me a while to parse out, I think it would be easier to read (and cleaner) without the  `&` capture parameter.  Just use `[&obj]`

src/QMCWaveFunctions/OptimizableObject.h
```diff
@@ -43,31 +43,35 @@ class OptimizableObject
...
+   * With OptimizableObject,
+   * class A: public OptimizableObject {}
+   * class B: public OptimizableObject
+   * {
+   *   A objA;
+   *   checkInVariablesExclusive() { // should not call objA.checkInVariablesExclusive(); }
+   * };
+   * A vector will be created by calling extractOptimizableObjects().
+   * All the checkInVariables() will be called through this vector and thus
+   * checkInVariablesExclusive implementation should only handle non-OptimizableObject members.
```
PDoakORNL, line 63: So to me the use of inclusive and exclusive here is not very clear.  I immediately thought of it terms of thread when I saw the Exclusive on this method. i.e. only call if you are certain you will be the only caller. Exclusive/Inclusive I might also have the sense of a range that does or doesn't include the endpoints or an operation that checks in any variable not already checked in (XOR/OR).

The difference here is whether the call traverses the implicit tree of optimizables below the object called with checkInVariables or not. So its easier for me to think of it as a "Traversing" checkInVariables (the legacy call) or a NonTraversing the new `checkInVariablesExclusive` call. Or as a recursive vs. nonrecursive call.

src/QMCWaveFunctions/Fermion/BackflowTransformation.h
```diff
@@ -156,15 +156,17 @@ class BackflowTransformation : public OptimizableObject
...
-
   bool isOptimizable() const;
-
-  void resetParameters(const opt_variables_type& active) override;
+  void checkInVariables(opt_variables_type& active);
+  void checkOutVariables(const opt_variables_type& active);
+  void resetParameters(const opt_variables_type& active);
+  void reportStatus(std::ostream& os) final;
+
+  // extractOptimizableObjectRefs is not enabled in BackflowTransformation.
```
PDoakORNL, line 165: I assume this is part of why this object is final?  Not sure whether this needs to be documented more at this time but it leaves me wondering why BackFlowTransformation is a special case.  This probably really amounts to BackFlow needing better documentation which is beyond this PR.

src/QMCWaveFunctions/OptimizableObject.h
PDoakORNL: It does seem clear in the new design the traversal occurs with extractOptimizableObjects so the checkInVariables must be non traversing.

I don't think this is a bad design especially since the "tree" is adhoc and each "node" used to class handle calling its leaves/branches in an adhoc way.  

Now that "tree" traversal is in classes extract method and the checkIn/reset/whatever are in methods that just do that which is better.

PDoakORNL (approved):
I think the documentation is coming along.  I think the "Exclusive" name is confusing but I'm not going to push for "NonTraversing" or "Local" that doesn't need to be fixed this PR.

src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
ye-luo: Will put "In total" when I finalize #4169

src/QMCWaveFunctions/OptimizableObject.h
ye-luo: I was taking the notion from inclusive/exclusive timer.

src/QMCWaveFunctions/Fermion/BackflowTransformation.h
ye-luo: Right now there is a compiler warning. Adding final in the class stops the warning. BackFlowTransformation should never be inherited and thus I think adding final is OK.

ye-luo:
@markdewing Does this PR look good to you?

markdewing (approved):

ye-luo:
Test this please

PR #4171, opened 2022-08-12
Add convertStrToVec and astring support in ParameterSet

ye-luo:
## Proposed changes
To load a vector of things, use astring to load the whole parameter content.
Then use convertStrToVec to generate the vector.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Utilities/string_utils.h
+++ b/src/Utilities/string_utils.h
@@ -187,10 +187,7 @@ inline std::ostream& operator<<(std::ostream& os, const astring& rhs)
   return os;
 }
 
-inline bool operator==(const astring& lhs, const astring& rhs)
-{
-  return lhs.s == rhs.s;
-}
+inline bool operator==(const astring& lhs, const astring& rhs) { return lhs.s == rhs.s; }
 } // namespace qmcplusplus
 
 #endif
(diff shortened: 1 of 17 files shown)


ye-luo:
Test this please

src/Utilities/string_utils.h
```diff
@@ -139,11 +142,30 @@ inline bool string2bool(const std::string& s)
...
   else
   {
-    APP_ABORT("string2bool received non-boolean string: " + s);
+    throw std::runtime_error("string2bool received non-boolean string: " + s);
     return false;
   }
 }
 
+template<class T>
+inline std::vector<T> convertStrToVec(const std::string& s)
```
markdewing, line 151: What format does this function expect the string to have?  Elements separated by spaces, commas, etc?

ye-luo: it is through `istringstream>>` so by white space. I should add a comment. Should I rename the function to `convertStrToVecBySpaces`?

markdewing: No, the separator shouldn't be in the name. It should be documented, though.

ye-luo: I will add a line of documentation in  #4168

PDoakORNL (approved):
This looks workable

PR #4173, opened 2022-08-15
Test cusp correction reordering fix

markdewing:
Rearrange the atoms to ungroup them in the existing ethanol cusp correction test case to test the fix in #4614
Add the GAMESS input file for this test case.

Output the particle index mapping if such a reordering is necessary. The output for this case is:
```
  Input particle set is not grouped by species.  Remapping particle position indices internally.
    Species : input particle index -> internal particle index
    H : 2 -> 1
    H : 3 -> 2
    H : 4 -> 3
    H : 6 -> 4
    H : 8 -> 5
    C : 1 -> 6
    C : 5 -> 7
    O : 7 -> 8
```

- No

desktop

## Checklist


markdewing:
Test this please

src/Particle/ParticleIO/XMLParticleIO.cpp
```diff
@@ -244,6 +245,18 @@ bool XMLParticleParser::readXML(xmlNodePtr cur)
...
+                if (!ungrouped_message_printed)
+                {
+                  app_log() << "  Input particle set is not grouped by species.  Remapping particle position indices "
+                               "internally."
+                            << std::endl;
+                  app_log() << "    Species : input particle index -> internal particle index" << std::endl;
+                  ungrouped_message_printed = true;
+                }
+                app_log() << "    " << group_species_name << " : " << iat << " -> " << storage_index - 1 << std::endl;
+              }
```
ye-luo, line 259: This is printout is sensitive to the ordering of element groups.
It is better to only set `input_ungrouped` to true here and then after the ig loop, handling the printout based on map_storage_to_input.

PS:
1) it is better to consume storage_index before storage_index++ to avoid adding confusing -1.
2) is isHighActive() necessary, should app_log() have been set accordingly?

markdewing: Moved the output code to come after the loop.

1. I would prefer to put code in order of importance - this code is less important than the code that actually stores the reordering.  But the if-test is now before the increment.
2. The isHighActive is not necessary, it is implicit in app_log.

ye-luo (commented):
LGTM. Need non-ANL approval.

prckent (commented):
A small suggestion while we are discussing it: guard this output about remapping based on verbosity.

If someone runs with 250 atoms,  we don't need 250 new lines of output unless verbosity=high. For other settings, printing the the remapping was done should still occur, with a note to set verbosity=high for details.

markdewing:
app_log should be paused when verbosity is low.   I should document this because I had forgotten the app_summary/app_log split until Ye mentioned it.
Maybe this output should be at the debug level?  The particle reorder mapping shouldn't affect the user and shouldn't be visible in any output.

prckent:
Debug OK with me.

ye-luo:
Test this please

prckent (approved):
LGTM

PR #4177, opened 2022-08-16
Implement NLPP parameter derivatives in evaluateDerivRatios in JeeI

ye-luo:
## Proposed changes
J3(eeI) NLPP derivatives

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
+++ b/src/QMCWaveFunctions/Jastrow/JeeIOrbitalSoA.h
@@ -434,6 +434,7 @@ public:
 
   void evaluateRatios(const VirtualParticleSet& VP, std::vector<ValueType>& ratios) override
   {
+    assert(VP.getTotalNum() == ratios.size());
     for (int k = 0; k < ratios.size(); ++k)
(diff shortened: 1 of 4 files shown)


ye-luo:
Test this please

ye-luo:
@prckent this one is ready for review. Reference numbers are generated in the same way as MSD PR earlier.

prckent (approved):
Thank Ye. Very good to have these.

Do any of the statistical tests go through this path and will need updating? For the tests we have I would think the change is subtle.

ye-luo:
> Thank Ye. Very good to have these.
> 
> Do any of the statistical tests go through this path and will need updating? For the tests we have I would think the change is subtle.

That is also what I found. The actual change in the total energy is likely small if the NLPP fraction of total energy is already small. However we should not have wrong derivatives, period.

rcclay (commented):
Code LGTM.  I would only request a little bit of documentation on the new code.  Specifically function documentation above PolynomialFunctor3D::evaluateDerivatives(...).

ye-luo:
> Disregard

Will add some documentation.

PR #4180, opened 2022-08-16
Change cuspInfo file name input behavior

### LINKED ISSUE #4179: Cusp correction file generation when file doesn't exist
When input file contains `cuspInfo="ethanol_updet.cuspInfo.xml"` but the file doesn't exist, the current code ignores this entry completely and generates a new file "sposet_name+`cuspInfo.xml`". To me this is a surprising behavior. I would expect the latter behavior only when cuspInfo is empty (not given). If cuspInfo is given, a file named as cuspInfo should be created so my following runs can pick it up.

Any comments to defend the current behavior?

ye-luo:
## Proposed changes
Closes #4179
The old code behavior:
if the file with file name from cuspInfo doesn't exist or invalid, create a new file with default filename in the current work directory.
So a second run still regenerate cusp correction unless pointing to the new files or move the new files as cuspInfo indicated. 

changed behavior:
if the file exists but is invalid, error out.
if the file doesn't exist, generate the file named as cuspInfo or default if cuspInfo is not provided.

So when the file doesn't exist, the first run create the file and the second run picks up the file.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/LCAO/CuspCorrectionConstruction.cpp
+++ b/src/QMCWaveFunctions/LCAO/CuspCorrectionConstruction.cpp
@@ -20,13 +20,13 @@ namespace qmcplusplus
 {
 // Modifies orbital set lcwc
 void applyCuspCorrection(const Matrix<CuspCorrectionParameters>& info,
-                         int num_centers,
-                         int orbital_set_size,
                          ParticleSet& targetPtcl,
                          ParticleSet& sourcePtcl,
                          LCAOrbitalSetWithCorrection& lcwc,
                          const std::string& id)
 {
+  const int num_centers = info.rows();
+  const int orbital_set_size = info.cols();
   using RealType = QMCTraits::RealType;
 
   NewTimer& cuspApplyTimer =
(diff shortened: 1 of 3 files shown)


prckent (commented):
I think the filename prefix should be the projectname. Doesn't look like this is the case?
Should have the same naming and logic as for spline storage.

ye-luo:
> I think the initial part of the filename should be the projectname. Doesn't look like this is the case? Should have the same naming and logic as for spline storage.

I don't think cusp currection should be named after the project name. It is related to the sposet not the project.
The spline storage(dump) doesn't allow input file names and they are hard-coded based on the attributes of sposet.

prckent:
The reason for adding the project name is to avoid problems when someone runs multiple versions of a system with different geometries in the same directory but without renaming the SPOset.

ye-luo:
> The reason for adding the project name is to avoid problems when someone runs multiple versions of a system with different geometries in the same directory but without renaming the SPOset. Of course this assumes that they rename the project as well...

When cuspInfo is not given, we don't read any cuspInfo file but only write one based on fresh reconstruction. This won't give you wrong results but some race when writing the file. Note that in an ensemble run, there is also potential race in writing and the solution in some places is adding a group id.
When cuspInfo file name is giving, users are responsible to name the files differently.

I think the question you raised is applicable to spline dump as well. How to handle file output collision in multiple regular jobs in the same work directory or how to handle ensemble jobs. My quick takes is we just don't support "multiple regular jobs in the same work directory", the ensemble job case needs to be looked at but it is beyond the scope of this PR and need to develop a policy we agree upon. use #4185

prckent (approved):

ye-luo:
Test this please

PR #4186, opened 2022-08-17
Add ompReductionTinyVector and rename ompReduction to ompReductionComplex

ye-luo:
## Proposed changes
Add user defined reduction for TinyVector types.
Clang is quite flexible on this but not GCC.
This makes gcc 13 develop to work in offload real builds. All deterministic tests pass.
Complex builds cause gcc compiler internal error.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):
Not 100% clear to me that ompReductionTinyVector.hpp is where it should be, but it will be easy to move if someone has a better idea.

ye-luo:
Test this please

PR #4192, opened 2022-08-19
Consolidate boolean handling in AttributeSet/ParameterSet

### LINKED ISSUE #2758: ParameterSet cannot handle boolean
**Describe the bug**
ParameterSet::add(bool) doesn't parse bool correctly.
The most funny logic is fliping value if the input entry is empty.
https://github.com/QMCPACK/qmcpack/blob/41b802bd38cb639ce040c555b4c20c4ecf490427/src/OhmmsData/OhmmsParameter.h#L171

**Expected behavior**
parse yes/no correctly

ye-luo:
## Proposed changes
Restrict input to "yes"/"no"/"true"/"false" values. block unknown or empty value.
Closes #2758

- Bugfix

- No

epyc-server

## Checklist


prckent (changes requested):
std::string yes should be renamed to input_string or somesuch. This avoids yes == "yes" etc.

Are the bool values converted to lower case anywhere? What about supporting "No" and "Yes"? A polite user friendly feature would be to support these variants, even if we don't advertise.

ye-luo:
> std::string yes should be renamed to input_string or somesuch. This avoids yes == "yes" etc.
> 
> Are the bool values converted to lower case anywhere? What about supporting "No" and "Yes"? A polite user friendly feature would be to support these variants, even if we don't advertise.

Added upper case handling.

prckent (approved):
Thanks.

ye-luo:
Test this please

PR #4194, opened 2022-08-20
Make use_nonlocalpp_deriv and nonlocalpp fully obselete

### LINKED ISSUE #3791: WFOpt nonlocalpp option

ye-luo:
## Proposed changes
Closes #3791.
The code can internally determine if NLPP exists or not. If it exists, always add it in correlated sampling calculation, namely always setting nonlocalpp="yes".
This is achieved by checking dependingOnTWF attribute of Hamiltonian components (H.getTWFDependentComponents()), not just specialized for NLPP.
To consolidate this concept, BareKineticEnergy has been made to depend on TWF explicitly and 
mw_evaluateWithParameterDerivatives and valuateValueAndDerivatives has been implemented to call TWF::evaluateDerivatives(). Thus explicit call to TWF::evaluateDerivatives() in QMCCostFunction implementations are no more needed.

Also selecting use_nonlocalpp_deriv=no is no more respected. As long as nonlocalpp or use_nonlocalpp_deriv appears in the parameters, deprecation warning will be printed.

Documentation update and removal of these two parameter in test xml will be done in a later PR.

- New feature

- No

epyc-server

## Checklist


src/QMCHamiltonians/HamiltonianFactory.cpp
```diff
@@ -133,6 +132,9 @@ bool HamiltonianFactory::build(xmlNodePtr cur, bool buildtree)
   if (psi_it == psiPool.end())
     APP_ABORT("Unknown psi \"" + psiName + "\" for target Psi");
   TrialWaveFunction* targetPsi = psi_it->second.get();
+  // KineticEnergy must be the first element in the hamiltonian array.
```
prckent, line 135: Where is this checked/enforced? How can it not be the case?

prckent (approved):

prckent:
Test this please

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -385,12 +365,9 @@ void QMCCostFunctionBatched::checkConfigurations(EngineHandle& handle)
...
 
-          if (includeNonlocalH != "no")
-          {
-            OperatorBase* nlpp = h_list[ib].getHamiltonian(includeNonlocalH);
-            if (nlpp)
-              RecordsOnNode[is][ENERGY_FIXED] -= nlpp->getValue();
-          }
+          const auto twf_dependent_components = h_list[ib].getTWFDependentComponents();
+          for (const OperatorBase& component: twf_dependent_components)
+            RecordsOnNode[is][ENERGY_FIXED] -= component.getValue();
```
markdewing, line 370: Previously, this value was only included if nlpp was set.  Now it checks for TWF-dependent components, which includes the kinetic energy.

PR #4201, opened 2022-08-23
Add isRotationSupported() and refine SPOSet.cpp error messages

ye-luo:
## Proposed changes
Better safeguards targeting internal logic consistency.

- New feature

- No

epyc

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/SPOSetBuilder.cpp
+++ b/src/QMCWaveFunctions/SPOSetBuilder.cpp
@@ -98,6 +98,9 @@ std::unique_ptr<SPOSet> SPOSetBuilder::createSPOSet(xmlNodePtr cur)
     // create sposet with rotation
     auto& sposet_ref = *sposet;
     app_log() << "  SPOSet " << sposet_ref.getName() << " is optimizable\n";
+    if (!sposet_ref.isRotationSupported())
+      myComm->barrier_and_abort("Orbital rotation not supported with '" + sposet_ref.getName() + "' of type '" +
+                                sposet_ref.getClassName() + "'.");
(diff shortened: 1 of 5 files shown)


src/QMCWaveFunctions/SPOSetBuilder.cpp
```diff
@@ -98,6 +98,9 @@ std::unique_ptr<SPOSet> SPOSetBuilder::createSPOSet(xmlNodePtr cur)
     // create sposet with rotation
     auto& sposet_ref = *sposet;
     app_log() << "  SPOSet " << sposet_ref.getName() << " is optimizable\n";
+    if (!sposet_ref.isRotationSupported())
```
PDoakORNL, line 101: Why not throw a UniformCommunicateError here?

PDoakORNL (changes requested):
This looks fine to me except for the SPOSetBuilder error.

src/QMCWaveFunctions/SPOSetBuilder.cpp
ye-luo: At this point, I can already safely stop the code and mpi communicator is available here. I think UniformCommunicateError is better suited in classes without mpi communicator.
returning UniformCommunicateError may got stuck if another exception happens in destruction due to earlier issues. So I feel better to stop early whenever possible.

PDoakORNL: Seems reasonable.

PDoakORNL (approved):

ye-luo:
Test this please

PR #4196, opened 2022-08-22
Fix LCAOrbitalSet::evaluateDetRatios and clean up Numerics/MatrixOperators.h

ye-luo:
## Proposed changes
A bug found in LCAOrbitalSet::evaluateDetRatios.
Also removed many not necessary APIs in Numerics/MatrixOperators.h

## What type(s) of changes does this code introduce?
- Bugfix
- Build related changes
- Testing changes (e.g. new unit/integration/performance tests)

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
epyc

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'

PR #4202, opened 2022-08-24
[WIP]Prototype orbital rotation

markdewing:
Replaces #4146 and #4147
Includes the changes in OptimizableObject and the wavefunction optimization API changes.
The PR isn't intended to be checked in. The intent to update the previous prototypes and make it visible for testing.

Implements options 2 and 3 in https://github.com/QMCPACK/qmcpack/issues/3983#issuecomment-1197380631

The options are switched by a flag (`use_global_rot_` in RotatedSPOs). There's no input option for it - the constructor needs to be changed in the source and recompiled.
* false - use the history list (like #4146)
* true - use global rotation (like #4147)

Loading a vp.h5 will restore the rotation state based on what type of data is stored in the file (history list or global rotation)

Status of various system types:
* LCAO
  * STO basis functions with single determinant - works
  * pp/ECP with GTO basis functions with single determinant  - work s(with the #4196 changes, which are included here)
  * GTO basis functions with cusp correction. Does not work. Probably will require recomputing the cusp correction after every rotation.
* B-splines
   * multideterminant  - doesn't crash, but the parameter gradients don't match in the gradient test
   * single determinant - crashes - will require fixes similar to #4196

- New feature

- No

desktop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
@@ -53,7 +53,8 @@ QMCCostFunctionBase::QMCCostFunctionBase(ParticleSet& w, TrialWaveFunction& psi,
       msg_stream(0),
       m_wfPtr(NULL),
       m_doc_out(NULL),
-      debug_stream(0)
+      debug_stream(0),
+      do_override_output(true)
 {
   GEVType = "mixed";
   //paramList.resize(10);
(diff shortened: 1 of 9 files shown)


markdewing:
Updated with some fixes so it works with splines SPO's and single determinants.
Replaces the in-place multiplication of the rotation matrix and the coefficient matrix.

Note on the gradient test: the default settings for the finite difference delta (1e-5) and using "float" precision for the splines will produce concerning relative differences (few percent) between the finite difference and analytic gradients for the rotation parameters.  Switching the spline precision to "double" for the gradient test produces much better relative differences.  Or increasing delta will lower the relative difference.

markdewing:
Replaced by #4351

PR #4205, opened 2022-08-24
Update batch driver to fix parameter derivative

### LINKED ISSUE #4203: Changes to parameter derivative evaluation break parameter derivative tests
PR #4194 breaks some of the parameter gradient tests.

Such as https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=short-He_orb_rot_jastrow_param_grad_batch-1-16-jud_b&date=2022-08-24

Run the affected tests with ctest -R short-He_orb_rot*

Interestingly, the gradient test passes (the analytic vs numerical gradients are fine).  It's the comparison to the external value that fails.

It also only affects the batch driver, not the legacy driver.

**System:**
Happens on desktop and CDash nightly testing

markdewing:
Fix #4203
The change from the local potential to the total potential was done in the legacy driver code, but not the batched driver code.

Prior to #4194 , the nlpp energy contribution was subtracted, if present.   With the new code, the list of wavefunction-dependent Hamiltonians includes the kinetic energy in addition to the nlpp contribution.  The initialization of the ENERGY_FIXED field was changed to account for this.

- Bugfix

- No

desktop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
@@ -363,7 +363,7 @@ void QMCCostFunctionBatched::checkConfigurations(EngineHandle& handle)
           RecordsOnNode[is][ENERGY_NEW]   = etmp;
           RecordsOnNode[is][ENERGY_TOT]   = etmp;
           RecordsOnNode[is][REWEIGHT]     = 1.0;
-          RecordsOnNode[is][ENERGY_FIXED] = h_list[ib].getLocalPotential();
+          RecordsOnNode[is][ENERGY_FIXED] = RecordsOnNode[is][ENERGY_TOT];
 
           const auto twf_dependent_components = h_list[ib].getTWFDependentComponents();
           for (const OperatorBase& component: twf_dependent_components)
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
@@ -174,7 +174,7 @@ void QMCCostFunctionBatched::getConfigurations(const std::string& aroot)
   auto components = H.getTWFDependentComponents();
   app_log() << " Found " << components.size() << " wavefunction dependent components in the Hamiltonian";
   if (components.size())
-    for(const OperatorBase& component: components)
+    for (const OperatorBase& component : components)
       app_log() << " '" << component.getName() << "'";
   app_log() << "." << std::endl;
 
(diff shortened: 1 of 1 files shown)


src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -352,49 +351,33 @@ void QMCCostFunctionBatched::checkConfigurations(EngineHandle& handle)
...
+          RecordsOnNode[is][ENERGY_FIXED] -= component.getValue();
       }
     }
   };
 
   ParallelExecutor<> crowd_tasks;
   crowd_tasks(opt_num_crowds, evalOptConfig, opt_eval_, samples_per_crowd_offsets, walkers_per_crowd_, dLogPsi,
-              d2LogPsi, RecordsOnNode_, DerivRecords_, HDerivRecords_, samples_, OptVariablesForPsi, needGrads,
-              handle);
+              d2LogPsi, RecordsOnNode_, DerivRecords_, HDerivRecords_, samples_, OptVariablesForPsi, needGrads, handle);
```
ye-luo, line 380: Sorry for the mess. From this line and below are only formatting.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -352,49 +351,33 @@ void QMCCostFunctionBatched::checkConfigurations(EngineHandle& handle)
...
-        for (int ib = 0; ib < current_batch_size; ib++)
-        {
-          const int is = base_sample_index + ib;
-          auto etmp    = energy_list[ib];
-          opt_data.get_e0() += etmp;
-          opt_data.get_e2() += etmp * etmp;
-
-          RecordsOnNode[is][ENERGY_NEW]   = etmp;
-          RecordsOnNode[is][ENERGY_TOT]   = etmp;
-          RecordsOnNode[is][ENERGY_FIXED] = h_list[ib].getLocalPotential();
```
ye-luo, line 387: This also doesn't seem right. and can share code with above. So I made the change.

src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp
```diff
@@ -352,49 +351,33 @@ void QMCCostFunctionBatched::checkConfigurations(EngineHandle& handle)
...
-        {
-          const int is = base_sample_index + ib;
-          auto etmp    = energy_list[ib];
-          opt_data.get_e0() += etmp;
-          opt_data.get_e2() += etmp * etmp;
-
-          RecordsOnNode[is][ENERGY_NEW]   = etmp;
-          RecordsOnNode[is][ENERGY_TOT]   = etmp;
-          RecordsOnNode[is][REWEIGHT]     = 1.0;
-          RecordsOnNode[is][ENERGY_FIXED] = h_list[ib].getLocalPotential();
```
ye-luo, line 366: My oversight. Thank @markdewing for root causing the failure.

prckent (commented):
This fixes a bug but did not require a test update => a test is missing or an existing one is too in sensitive.

How can we avoid this class of error in future?

ye-luo:
> This fixes a bug but did not require a test update => a test is missing or an existing one is too insensitive.
> 
> How can we avoid this class of error in future?

A short test was failing caused by the bug fixed in this PR. Right now deterministic tests don't cover all parts of WFopt drivers. So running short tests is necessary. WFOpt needs a lot of refactoring to be unit testable.

prckent:
Yes, we are sorely lacking unit through deterministic integration tests in the optimizer. I fear these breakages will keep happening until we have something running in the CI.

prckent (approved):

PR #4213, opened 2022-08-29
Fix counting Jas test failure

ye-luo:
## Proposed changes

Fix https://cdash.qmcpack.org/CDash/testDetails.php?test=21311124&build=342868

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):
I think this points to a weakness of the current class/function setup - we ought to force an optimizable object to implement this function.

prckent:
Test this please

ye-luo:
> I think this points to a weakness of the current class/function setup - we ought to force an optimizable object to implement this function.

Actually extractOptimizableObjectRefs is required by WFC, OptimizableObject doesn't require this and probably should not. CountingJastrow inherit both classes and thus looks like it is being required by OptimizableObject. So far, my bug trap actually works as intended.

prckent:
Compile time trap would be better though.

ye-luo:
Test this please

PR #4221, opened 2022-09-06
Improve gradient test

markdewing:
Allow specifying the finite difference delta from the input file. Motivation for this addition comes from checking the orbital rotation gradients for Bspline SPOs.  The Bspline SPOs can be used with a precision of float or double.  The orbital rotation gradients with float precision were giving a large relative difference, and it turned out the default finite difference delta of 1e-5 was too small.  As the value of delta increased, the relative error decreased, which indicated it was in the numerical noise region.  As an alternate check, the precision of the SPOs was switched to double, and then the relative differences from the default delta were fine.

For improvements in the output formatting, the columns and headers are aligned, as requested in #3793.  The precision of the gradients was increased.  The precision of the relative difference was decreased (pretty much only the exponent matters.).
Making the format of the relative difference more uniform makes it harder to quickly scan for problematic values.  A simple threshold scheme was added that prints !, !!, or !!! after the relative difference to highlight problematic values. The thresholds are set to 10^-3, 10^-2, and 10^-1.  

The documentation now includes sample output.

- New feature

- No

desktop

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/GradientTest.cpp
+++ b/src/QMCDrivers/WFOpt/GradientTest.cpp
@@ -47,8 +47,7 @@ void GradientTest::run(QMCCostFunctionBase& costFunc, const std::string& root_na
   }
 
   // Numerical gradient
-  double finite_diff_delta = 1e-5;
-  costFunc.GradCost(numeric_grad, params, finite_diff_delta);
+  costFunc.GradCost(numeric_grad, params, input_.finite_diff_delta());
 
   // Analytic gradient
   costFunc.GradCost(analytic_grad, params);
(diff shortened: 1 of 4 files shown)


src/QMCDrivers/WFOpt/GradientTest.cpp
```diff
@@ -60,20 +59,62 @@ void GradientTest::run(QMCCostFunctionBase& costFunc, const std::string& root_na
     param_deriv_index_++;
   }
 
-  app_log() << "Param_Name  Value    Numeric            Analytic       Percent" << std::endl;
+  size_t max_name_len = std::string("Param_Name").size();
```
ye-luo, line 62: It is better to define a `std::string_view` for "Param_Name" considering it being used twice here and below.

src/QMCDrivers/WFOpt/GradientTestInput.h
```diff
@@ -25,9 +25,11 @@ class GradientTestInput
 
 protected:
   bool do_param_output_ = false;
+  double finite_diff_delta_ = 1e-5;
 
 public:
   bool do_param_output() { return do_param_output_; }
+  double finite_diff_delta() { return finite_diff_delta_; }
```
ye-luo, line 32: The overall convention for function naming is starting with a verb. `get_finite_diff_delta` ? also both member functions should be marked const.

ye-luo (commented):
Overall looks good to me. Minor change requested.

PDoakORNL (approved):
LGTM

ye-luo:
Test this please

PR #4224, opened 2022-09-09
Enable samples input tag in batched drivers

ye-luo:
## Proposed changes
Add `samples` as a way of control independently from run details (MPI ranks, threads)

- New feature

- No

epyc-server

## Checklist


prckent (changes requested):
Please put up an issue on the topic of new input tags.

There needs to be a clear discussion on the end point. I have ideas that I believe can work, as do you + no doubt others.  This will take some time.

Before tackling the optimizer we need to tackle at least VMC and also make sure that the new input tags are sufficiently verbose and self-explanatory. e.g. we will need prefixes total_ input items that are "global" and also to make sure that they don't conflict with historically used terms. Importantly we need a reasonably complete plan for all the main QMC driver input revisions before going ahead with user facing changes. I think a very simple solution is possible but discussion is needed ahead of implementation.

src/QMCDrivers/DMC/DMCBatched.h
```diff
@@ -47,19 +47,24 @@ class DMCBatched : public QMCDriverNew
   struct StateForThread
   {
     const QMCDriverInput& qmcdrv_input;
-    const DMCDriverInput& dmcdrv_input;
     const DriftModifierBase& drift_modifier;
     const MCPopulation& population;
     SFNBranch& branch_engine;
     IndexType recalculate_properties_period;
-    IndexType step            = -1;
+    const size_t steps_per_block;
```
PDoakORNL, line 54: since `IndexType` is signed and `steps_per_block` is unsigned and we're likely to compare `step` and `steps_per_block` and this potentially an issue.  As is doing arithmatic operations with mixed signed and unsiged types.  I think steps_per_block should be signed.

PDoakORNL (changes requested):
This looks good to me outside of the unsigned steps_per_block.

prckent:
I would like for all the input tags that refer to the "overall total amount of work" to have the normalization included in their name. This is to reduce the risk for confusion and mistakes over what the user might expect. What we choose for "samples" will carry through to other things such as some eventual "total_steps" (to include sum over blocks as well as steps and walkers). While not exactly the same thing, for walkers we have "total_walkers", i.e. the total number of walkers pushed in VMC/DMC, equivalent to the total number of walkers summed over all threads, crowds, mpi processes etc.  For samples we could use "total_samples" or perhaps "total_samples_per_cycle". I do not think we should use "samples" since it does not include the normalization. What do you prefer?

ye-luo:
> I would like for all the input tags that refer to the "overall total amount of work" to have the normalization included in their name. This is to reduce the risk for confusion and mistakes over what the user might expect. What we choose for "samples" will carry through to other things such as some eventual "total_steps" (to include sum over blocks as well as steps and walkers). While not exactly the same thing, for walkers we have "total_walkers", i.e. the total number of walkers pushed in VMC/DMC, equivalent to the total number of walkers summed over all threads, crowds, mpi processes etc. For samples we could use "total_samples" or perhaps "total_samples_per_cycle". I do not think we should use "samples" since it does not include the normalization. What do you prefer?

Since `samples_per_thread` has been removed in the batched drivers. `samples` is the only input option in use and there is no intention to add more. It does exactly mean `total_samples`, for maximizing backward compatibility and considering its briefness. I prefer keeping `samples`.

ye-luo:
> > It does exactly mean `total_samples`, for maximizing backward compatibility and considering its briefness. I prefer keeping `samples`.
> 
> OK, so lets run with `samples`. Can always add more options if confusion persists. The documentation needs to be updated to make the normalization clear. I am happy to write that change. However unfortunately I have been stalling this PR for long enough that this PR has conflicts. Would you prefer I update the docs here?

Let me review the current change and bring it up to date. Will also update docs in this PR.

src/QMCDrivers/DMC/DMCBatched.h
ye-luo: "steps_per_block" should be unsigned by nature.
```
for (int step = 0; step < steps_per_block_; ++step)
```
I have not heard of unsigned loop upper bound being an issue.

ye-luo:
@prckent I updated the code. After reviewing the code, "samples" should be interpreted as a requested number. VMC may produce equal or more. DMC is similar to VMC when using fixed population or a rough estimate when using dynamic population.

src/QMCDrivers/DMC/DMCBatched.h
prckent: The issue is the signed/unsigned comparison, which is not a great habit. We probably have several of them in the code which makes me wonder if compiler warnings have been disabled for this case -- usually there will be a warning.

ye-luo: since we guarantee the counter `step` positive, there should be no concern. If a real case that matters, I would prefer changing to size_t or int depending on the actual need.

src/QMCDrivers/QMCDriverNew.h
```diff
@@ -299,6 +299,11 @@ class QMCDriverNew : public QMCDriverInterface, public MPIObjectBase
                                                                     const RealType reserve_walkers,
                                                                     int num_crowds);
 
+  static size_t determineStepsPerBlock(IndexType global_walkers,
```
prckent, line 302: Comment with definition needed.
Can you also make everything const so that it is clear e.g. none of the arguments are changed?

src/QMCDrivers/tests/QMCDriverNewTestWrapper.h
```diff
@@ -176,6 +176,20 @@ class QMCDriverNewTestWrapper : public QMCDriverNew
   };
 
   void testMeasureImbalance() { measureImbalance("Test"); }
+  void testDetermineStepsPerBlock()
+  {
+    CHECK(QMCDriverNew::determineStepsPerBlock(6, 36, 2, 3) == 2);
```
prckent, line 181: Thanks for adding these. I think sufficient to avoid needing to add integration tests.

prckent (commented):
Interesting to support samples in DMC. Do you know a use case or were you aiming for consistency?

ye-luo:
> Interesting to support samples in DMC. Do you know a use case or were you aiming for consistency?

Only a bare minimal implementation aiming for consistency.

src/QMCDrivers/QMCDriverNew.h
ye-luo: It is a static member function and arguments are passed by value. So everything is stateless and const by nature.

ye-luo:
Test this please

prckent (approved):

prckent:
@PDoakORNL I think the signed comparison issue you pointed to is adequately protected. OK to merge?

src/QMCDrivers/DMC/DMCBatched.h
PDoakORNL: Something being positive definite is not a good reason to use an unsigned type.  If you are going to do any arithmetic with it save some grief and use a signed type.

PDoakORNL (dismissed):
This sort of thing is not limited to here in the code but I don't like it much. Feel free to override my request for changes its bikeshed topic and its a typical sort of scientific code defect.

But if you mind is not set on using unsiged types because the number is positive definite please read:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-mix

Maybe you can write better code than our future LLM masters.

ye-luo:
> This sort of thing is not limited to here in the code but I don't like it much. Feel free to override my request for changes its bikeshed topic and its a typical sort of scientific code defect.
> 
> But if you mind is not set on using unsiged types because the number is positive definite please read: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-mix
> 
> Maybe you can write better code than our future LLM masters.

I actually agree with what C++ core guidelines say "Do not mix". My issue here is actually the data type of steps, it should be unsigned as well. The reason we keep it signed was the fact that we are using -1 to indicate no input given. I actually feel the need to change the default value to 0. In a broader scope, all the use -1 in input handling should be revisited.

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -276,11 +279,11 @@ void VMCBatched::process(xmlNodePtr node)
...
 
-int VMCBatched::compute_samples_per_rank(const QMCDriverInput& qmcdriver_input, const IndexType local_walkers)
+size_t VMCBatched::compute_samples_per_rank(const size_t num_blocks,
+                                            const size_t samples_per_block,
+                                            const size_t local_walkers)
 {
-  int nblocks = qmcdriver_input.get_max_blocks();
-  int nsteps  = qmcdriver_input.get_max_steps();
-  return nblocks * nsteps * local_walkers;
+  return num_blocks * samples_per_block * local_walkers;
```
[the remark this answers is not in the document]
ye-luo: > The only location compute_samples_per_rank is called it is assigned into a signed integer.
I noticed. I will fix it now. `compute_samples_per_rank` itself is fine, everything should be unsigned as we are handling all kinds of sizes.

prckent (approved):
auto addresses the typing issue raised by Peter.

We should be careful in future about computations such as `nsteps*nblocks*total_walkers`, which could come close to (u)int32 ranges in extreme cases.

prckent:
Test this please

PR #4208, opened 2022-08-25
mw_ratioGradWithSpin for DiracDeterminantBatched

camelto2:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

This PR adds the mw_ratioGradWithSpin API for DiracDeterminantBatched. Additionally, added the mw_evaluateVGLandRatiosGradWithSpin APIs to SPOSet and SpinorSet which was needed for mw_ratioGradWithSpin. Unit test is updated to test the mw_ APIs. I also added the checks for mw_evalGradWithSpin to the unit test, but left it commented out until I follow up with a subsequent PR implementing that for DiracDeterminantBatched. 

Once this and the subsequent PR is merged, DiracDeterminantBatched should be complete for spinors. 

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- New feature

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
macbook pro, M1Pro

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- Yes. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #4226, opened 2022-09-09
SpinorSet resources 

camelto2:
## Proposed changes

The goal of the PR is to eventually enable offloading for mw_evaluateVGLandDetRatios for SpinorSet and other related APIs. In order to enable this, SpinorSet needs the create/acquire/releaseResource APIs, and also needs to make sure the up/down sposets also create/acquire/release their Resources as well. This PR adds the resource APIs and a SpinorSetMultiWalkerResource which will get registered to the ResourceCollection. 

It is tested by adding a check in the SpinorSet unit tests to make sure the SpinorSetMultiWalkerResource is either nullptr or not depending on whether or not a ResourceCollectionTeamLock is in place. 

As of right now, the resource doesn't actually contain any useful data. This will be added in a future PR when I actually implement the offload stuff which will take care of [this](https://github.com/QMCPACK/qmcpack/pull/4208#discussion_r955299844) and [this](https://github.com/QMCPACK/qmcpack/pull/4208#discussion_r955312073) 

- New feature

- No

my mac

## Checklist


ye-luo (approved):
I moved the multi walker resource definition to the cpp file and also make the resource unique_ptr private

ye-luo (approved):

PR #4231, opened 2022-09-12
SpinorSet offload VGLandDetGradRatios

camelto2:
## Proposed changes

This PR optimizes the existing mw_evaluateVGLandDetRatiosGrad implemented in #4208 so that we remove the data transfers to the host and instead update the spinor data on the device. 

This also adds an extractSpinComponentRefList API reduce repeated code for getting the up/down spo reference vectors. I updated a few other functions which had repeated code to use the new API

Note that there is a unit test for this API already in place. However, the problem is that the existing unit tests don't utilize splines which are the only SPOSets that are currently offloaded. 

- New feature

- No

my mac

## Checklist


ye-luo:
Test this please

ye-luo (approved):

PR #4237, opened 2022-09-14
Fixed bug in handling of coefs array.

jptowns:
## Proposed changes

In the current orbital rotation in SplineR2R, the spline coefs are overwritten in place, which is bad. This PR fixes this problem by creating a temporary copy of the coefs, applying the rotation, and then overwriting the coefs array with the correct updated coefs.

This PR changes the applyRotation method in SplineR2R

- Bugfix

- No

RHEL8 Intel workstation

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
@@ -60,18 +60,19 @@ bool SplineR2R<ST>::write_splines(hdf_archive& h5f)
   spl_coefs has a complicated layout depending on dimensionality of splines. 
   Luckily, for our purposes, we can think of spl_coefs as pointing to a 
   matrix of size BasisSetSize x (OrbitalSetSize + padding), with the spline 
-  index adjacent in memory. NB: The orbital index is SIMD aligned and
-  therefore may include padding.
+  index adjacent in memory. The orbital index is SIMD aligned and therefore 
+  may include padding.
     
-  In other words, due to SIMD alignment, Nsplines may be larger than the 
-  actual number of splined orbitals, which means that in practice rot_mat 
-  may be smaller than the number of 'columns' in the coefs array. 
-  Therefore, we put rot_mat inside "tmpU". The padding of the splines 
-  is at the end, so if we put rot_mat at top left corner of tmpU, then 
-  we can apply tmpU to the coefs safely regardless of padding.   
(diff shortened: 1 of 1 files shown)


src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
```diff
@@ -99,6 +100,7 @@ void SplineR2R<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
   }
 
   // Apply rotation the dumb way b/c I can't get BLAS::gemm to work...
+  const auto new_coefs = new ValueType[coefs_tot_size];
```
markdewing, line 103: It would be better to avoid explicit memory management, and use a std::vector for the temporary.

jptowns: Fixed.

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
```diff
@@ -110,10 +112,16 @@ void SplineR2R<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
...
       }
-      *(spl_coefs + cur_elem) = newval;
+      new_coefs[cur_elem] = newval;
     }
   }
 
+  // Update the coefs
+  for (auto i = 0; i < coefs_tot_size; i++)
+  {
+    *(spl_coefs + i) = new_coefs[i];
```
PDoakORNL, line 122: any reason not to use:
```c++
std::copy(new_coefs.begin(), new_coefs.end(), spl_coefs);
```
This would allow dropping the coefs_tot_size variable as well.

PDoakORNL (changes requested):
One other easy improvement

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
jptowns: Done.

markdewing (approved):

markdewing:
Test this please

prckent:
Test this please

PDoakORNL (approved):

PDoakORNL:
I'm not following codecov claiming a change in coverage for the PR.  Looks like it should be the same to me.

Better direct unit test coverage of the SplineXtoX classes would be great but is a concern beyond this PR.

ye-luo:
Test this please

PR #4238, opened 2022-09-14
Fix -DQMC_SYMLINK_TEST_FILES=OFF 

quantumsteve:

## Proposed changes

At the Crusher Hackathon I ran into trouble using `-DQMC_SYMLINK_TEST_FILES=OFF`. Files were being placed inside an additional directory and not where they were expected to be located.

I believe the issue is that [file(COPY ...)]( https://cmake.org/cmake/help/latest/command/file.html#copy) asks for a destination directory and therefore has different behavior than `cmake -E copy...`. Instead of going back to execute_process, I think [configure_file](https://cmake.org/cmake/help/latest/command/configure_file.html) should give the same behavior as `cmake -E copy ...` with the added benefit of updating files that change in the source directory or qmc_dir.

To test:

1. build qmcpack with `-DQMC_SYMLINK_TEST_FILES=OFF`
2. run `cmake -R deterministic-unit_test_wavefunction_trialwf`. This test should fail in develop and pass in this branch.

- Bugfix
- Build related changes

- No

local workstation

## Checklist

### DIFF
--- a/CMake/macros.cmake
+++ b/CMake/macros.cmake
@@ -98,7 +98,7 @@ function(MAYBE_SYMLINK SRC_DIR DST_DIR)
   if(QMC_SYMLINK_TEST_FILES)
     file(CREATE_LINK ${SRC_DIR} ${DST_DIR} SYMBOLIC)
   else()
-    file(COPY ${SRC_DIR} DESTINATION ${DST_DIR})
+    configure_file(${SRC_DIR} ${DST_DIR} COPYONLY)
   endif()
 endfunction()
 
--- a/CMake/macros.cmake
+++ b/CMake/macros.cmake
@@ -94,11 +94,11 @@ function(COPY_DIRECTORY_MAYBE_USING_SYMLINK SRC_DIR DST_DIR ${ARGN})
 endfunction()
 
 # Symlink or copy an individual file
-function(MAYBE_SYMLINK SRC_DIR DST_DIR)
+function(MAYBE_SYMLINK SRC_FILE DST_FILE)
   if(QMC_SYMLINK_TEST_FILES)
-    file(CREATE_LINK ${SRC_DIR} ${DST_DIR} SYMBOLIC)
+    file(CREATE_LINK ${SRC_FILE} ${DST_FILE} SYMBOLIC)
   else()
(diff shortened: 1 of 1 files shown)


markdewing (commented):
I think the function argument names (which I used originally) might be misleading with the _DIR suffix.  I think they're always files and never a directory(?).  Using SRC_FILE and DST_FILE instead would make the function clearer.   Could you change those names, as long as you're working on the PR.

CMake/macros.cmake
```diff
@@ -94,11 +94,11 @@ function(COPY_DIRECTORY_MAYBE_USING_SYMLINK SRC_DIR DST_DIR ${ARGN})
...
 
 # Symlink or copy an individual file
-function(MAYBE_SYMLINK SRC_DIR DST_DIR)
+function(MAYBE_SYMLINK SRC_FILE DST_FILE)
   if(QMC_SYMLINK_TEST_FILES)
-    file(CREATE_LINK ${SRC_DIR} ${DST_DIR} SYMBOLIC)
+    file(CREATE_LINK ${SRC_FILE} ${DST_FILE} SYMBOLIC)
   else()
-    file(COPY ${SRC_DIR} DESTINATION ${DST_DIR})
+    configure_file(${SRC_FILE} ${DST_FILE} COPYONLY)
```
ye-luo, line 101: Could you add a comment saying `configure_file` is used as a copy and explain why `file(copy` cannot be used. It is quite confusing when reading.

quantumsteve: Comment added. I also listed a few other options I considered but rejected because the require newer CMake versions.

May we set `-DQMC_SYMLINK_TEST_FILES=OFF` in a PR or nightly job to catch future errors?

markdewing (approved):

ye-luo:
Test this please

quantumsteve:
> @quantumsteve How would we best test for this bug in future testing? Was -DQMC_SYMLINK_TEST_FILES=OFF broken everywhere?

Yes, it was broken everywhere. Since tests already fail, adding `-DQMC_SYMLINK_TEST_FILES=OFF` to one CI job should be sufficient to highlight issues in the future.

PR #4243, opened 2022-09-16
Nexus: fix convert4qmc hdf5 issue

jtkrogel:
## Proposed changes

Fix an issue with expected filenames.  `convert4qmc` no longer produces an hdf5 file when given one from PySCF.  This leads to incorrect paths being provided to QMCPACK for the orbitals.  Fix this by inserting a symlink in place of the file convert4qmc used to generate.

- Bugfix

- No

Laptop

## Checklist


prckent (commented):
Can you please put a comment about why this is done?

jtkrogel:
Done.

prckent (approved):
LGTM

nexus/lib/qmcpack_converters.py
```diff
@@ -864,6 +864,19 @@ def check_sim_status(self):
         output = open(os.path.join(self.locdir,self.outfile),'r').read()
         #errors = open(os.path.join(self.locdir,self.errfile),'r').read()
 
+        # Recent versions of convert4qmc no longer produce the orbs.h5 file.
```
ye-luo, line 867: Do you know how recent? If not, put 3.15 should be fine.

markdewing:
Where does .orbs.h5 get used?  Is it possible to propagate the name of the h5 file where it is needed, rather than make symbolic link?

markdewing:
It looks like there are cases in convert4qmc that can still produce an `.orbs.h5` file (-dirac and -hdf5 flags).  Will cause this code to try make symlink to the same file? (It'll probably issue a failure message for `ln`, but still continue.)

nexus/lib/qmcpack_converters.py
jtkrogel: It's not very recent.  @anbenali knows better.

jtkrogel:
Yes, it would try and fail to make a symlink in those cases but proceed fine anyway.  I can put in a file existence check before proceeding with the attempted symlink creation.

jtkrogel:
More safety checks added.  Now the symlink will only be made if an h5 file has been provided and one has not been created by convert4qmc.

nexus/lib/qmcpack_converters.py
anbenali: convert4qmc produces an HDF5 only when a txt file is provided (gamess) and the tag -hdf5 is produced.

anbenali (approved):

prckent:
Test this please

PR #4236, opened 2022-09-14
Raise GCC minimum to 9

ye-luo:
Need to decide if we'd like to merge this before or after release.
Needed by #4235

## Proposed changes
It is way beyond 2-year rule.

## What type(s) of changes does this code introduce?
- Build related changes

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
bora, theta

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'

PR #4244, opened 2022-09-16
Upgrade to Clang15 and GCC9 on self-hosted CI

### LINKED ISSUE #4242: Bump minimal CI compilers versions
QMCPACK has a two-year rule support for minimal compiler versions supported. Some GCC compilers toolchains in CI predate this two year limit.

Bump minimum CI version on sulfur, nitrogen and migrate to Ubuntu 22 containers for GitHub Actions CI

**Additional context**
Related to PR #4236 

- [ ] use and test gcc-9 on sulfur and nitrogen, RHEL8 uses gcc-8 by default

williamfgc:
## Proposed changes
Bump from RHEL8 system GCC8 to GCC9 on sulfur and nitrogen CI
Trying Clang15 on sulfur built pointing at GCC9
Fixes #4242 
Related to PR #4236 to make GCC9 a minimal requirement

- No

sulfur, needs to pass CI

## Checklist


williamfgc:
Test this please

williamfgc:
TO DO: need to update job names to GCC9 on yaml file after current CI passes with configure logs catching GCC9.

ye-luo:
Test this please

tests/test_automation/github-actions/ci/run_step.sh
```diff
@@ -375,7 +383,7 @@ case "$1" in
 
     if [[ "${GH_JOBNAME}" =~ (AFQMC-Offload) ]]
     then
-       export LD_LIBRARY_PATH=/opt/llvm/15.0.0-rc3/lib:/usr/lib64/openmpi/lib:${LD_LIBRARY_PATH}
+       export LD_LIBRARY_PATH=/opt/llvm/15.0.0/lib:/usr/lib64/openmpi/lib:${LD_LIBRARY_PATH}
```
prckent, line 386: Does not need fixing here, but clearly this openmpi setting is in the wrong location; there should be an "if MPI" section.

prckent:
Ye beat me to the error bar update. This is fine.

ye-luo:
> TO DO: need to update job names to GCC9 on yaml file after current CI passes with configure logs catching GCC9.

How about just call tests GCC without version numbers? Then we don't need to update job yaml all the time.
We can figure out the version from the configure stage cmake printout anyway.

prckent:
I prefer to have the version number in the job name since it is a regular reminder of the different compiler versions that we are testing and helpful for people who otherwise might be unsure where to look. It is also consistent with the nightlies. (In those we can figure out the version of the compiler and set the cdash submit name automatically.)

tests/test_automation/github-actions/ci/run_step.sh
williamfgc: If I recall correctly AFQMC always needs MPI that was the logic behind this if. We can change it.

williamfgc:
After merging we can disable the old `GCC8-*` required statuses and move to the new `GCC9-*`. Thanks!

ye-luo:
@prckent please force merge this PR. There is no more GCC8 configuration support and ornl-ci expect to fail.

prckent:
I'll have to update the required checks when these new ones report on an open PR.

PR #4247, opened 2022-09-19
Directly set c++17 for nvcc.

ye-luo:
## Proposed changes
Current clang develop caused nvcc detection failure in CMake. Explicitly instructing the C++ standard used by nvcc works around the issue. See discussion in https://gitlab.kitware.com/cmake/cmake/-/issues/23971

- Build related changes

- No

epyc-server

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -731,10 +731,16 @@ if(QMC_CUDA OR ENABLE_CUDA)
     endif()
     set(CMAKE_CUDA_STANDARD_REQUIRED TRUE)
     set(CMAKE_CUDA_EXTENSIONS OFF)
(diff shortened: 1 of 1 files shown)


prckent (changes requested):
Please link the Kitware discussion. Also a mention of the affected versions would be ideal, assuming we know what they are.

ye-luo:
Test this please

ye-luo:
> Please link the Kitware discussion. Also a mention of the affected versions would be ideal, assuming we know what they are.

Kitware gitlab issue link was provided. All CMake versions that QMCPACK supports are affected. The issue shows up with clang since Sep 08 nightly compiler. icpx+nvcc also potentially has this issue.

prckent (approved):

prckent:
I was thinking that more explanation in the actual cmake file would help, not here.

ye-luo:
Test this please

PR #4257, opened 2022-09-27
Nexus:  Improve QE Fermi level handling for spin polarized systems

aannabe:
## Proposed changes

In Quantum Espresso, if one uses `nspin = 2` (unrestricted alpha and beta) without specifying `tot_magnetization`, QE will write only one Fermi level to `data-file-schema.xml`. In this case, the same E_Fermi is used for alpha and beta electrons and must be duplicated in the conversion. This is only relevant when using the grand-canonical occupations in QMC.

- Bugfix

- No

OLCF

## Checklist


ye-luo:
Next time please use your own fork to initiate PRs.

markdewing:
It is always going to be the case that ns == 2 and len(Ef_list) == 1 or 2?
If there are cases where len(Ef_list) == 0, this code will stop with an IndexError.
If there is a chance of other cases occurring, it might be worth adding checks.

aannabe:
Agreed, added a safety check in case they are different.

markdewing (approved):

PR #4234, opened 2022-09-13
[WIP] MatrixUpdateOMP evalGradWithSpin

camelto2:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

add implementation of MatrixUpdateOMPTarget::mw_evalGradWithSpin. Currently it isn't used, but will be used by DiracDeterminantBatched::mw_evalGradWithSpin. 

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- New feature

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
my mac

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- No. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #4269, opened 2022-10-06
mw_evalGradWithSpin for DiracDeterminantBatched

camelto2:

## Proposed changes

This PR incorporates #4234 to implement mw_evalGradWithSpin for DiracDeterminantBatched. In this implementation, we recompute the spin gradients from scratch similar to how it was done for the legacy implementation. 

Ideally, we would be storing and using the accept/reject to avoid recomputing the spin gradients from the SPOSets each time we need it. I would eventually like to change to this approach, but for now this should be correct and not result in too large of a penalty. After discussing offline with @ye-luo, this was the preferred approach for now. 

Also in #4234, @PDoakORNL mentioned he would help with the CUDA implementation after this is working and tested for the offload. So for now, I just added a runtime_error for the CUDA version. 

- New feature

- No

mac

## Checklist


src/QMCWaveFunctions/Fermion/MatrixUpdateOMPTarget.h
```diff
@@ -197,6 +199,69 @@ class MatrixUpdateOMPTarget
...
+    {
+      ptr_buffer[0][iw] = engines[iw].get_psiMinv().device_data() + rowchanged * engine_leader.get_psiMinv().cols();
+      ptr_buffer[1][iw] = dpsiM_row_list[iw];
+      ptr_buffer[2][iw] = dspin_row_list[iw];
+    }
+
+    constexpr unsigned DIM = GT::Size;
+    grads_value_v.resize(nw, DIM);
+    spingrads_value_v.resize(nw);
+    auto* __restrict__ grads_value_v_ptr     = grads_value_v.data();
```
PDoakORNL, line 230: never like to see the cargo cult __restrict__ all over but I notice it elsewhere in this file now, its not part of C++.  Except in legacy files it is used `#defined` so it can be easily dropped. In principle an ideal optimizer should not need it. Do we have any evidence that omp target does something with this?  That's the only justification I can see for this being here.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
```diff
@@ -197,6 +197,80 @@ typename DiracDeterminantBatched<DET_ENGINE>::Grad DiracDeterminantBatched<DET_E
...
+  dspin_v_list.reserve(nw);
+
+  SPOSet::ValueVector val(wfc_leader.Phi->size());
+  SPOSet::GradVector grad(wfc_leader.Phi->size());
+  for (int iw = 0; iw < wfc_list.size(); iw++)
+  {
+    auto& det = wfc_list.getCastedElement<DiracDeterminantBatched<DET_ENGINE>>(iw);
+    phi_list.push_back(*det.Phi);
+
+    psi_values.emplace_back(val);
```
PDoakORNL, line 237: So these are all getting constructed with the leader's copy of val? If that isn't incorrect i.e. these are only output variables why not just resize to nw.
Am I missing something?

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
```diff
@@ -197,6 +197,80 @@ typename DiracDeterminantBatched<DET_ENGINE>::Grad DiracDeterminantBatched<DET_E
...
+    std::vector<Grad>& grad_now,
+    std::vector<ComplexType>& spingrad_now) const
+{
+  assert(this == &wfc_list.getLeader());
+  auto& wfc_leader = wfc_list.getCastedLeader<DiracDeterminantBatched<DET_ENGINE>>();
+  ScopedTimer local_timer(RatioTimer);
+
+  const int nw = wfc_list.size();
+  RefVectorWithLeader<SPOSet> phi_list(*Phi);
+  std::vector<SPOSet::ValueVector> psi_values, lap_values, dspin_values;
```
PDoakORNL, line 214: So why more copies of these variables?

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
```diff
@@ -197,6 +197,80 @@ typename DiracDeterminantBatched<DET_ENGINE>::Grad DiracDeterminantBatched<DET_E
...
+    lap_values.emplace_back(val);
+    dspin_values.emplace_back(val);
+
+    psi_v_list.emplace_back(psi_values.back());
+    grad_v_list.emplace_back(grad_values.back());
+    lap_v_list.emplace_back(lap_values.back());
+    dspin_v_list.emplace_back(dspin_values.back());
+  }
+  auto& phi_leader = phi_list.getLeader();
+  phi_leader.mw_evaluateVGLWithSpin(phi_list, p_list, iat, psi_v_list, grad_v_list, lap_v_list, dspin_v_list);
```
PDoakORNL, line 248: So  I guess this is all being done to get dspin_values via dspin_v_list?  I don't see where grad_values, lap_values etc. are ever used. There is no indication of in and out variables int the SPOSet docs.

PDoakORNL (changes requested):
I have a number of questions and see no testing.

src/QMCWaveFunctions/Fermion/MatrixUpdateOMPTarget.h
camelto2: I was basically following what was done elsewhere in the file. I'm not sure exactly what omp target does with it. @ye-luo?

camelto2:
> I have a number of questions and see no testing.

The test was already in test_DiracDeterminantBatched for the spinor_update test. I had added the test and reference values in a previous PR, but commented them out so that when I added mw_evaluateGradWithSpin I could uncomment. So the MatrixUpdateOMPTarget is indirectly tested because it is called inside the DiracDeterminantBatched::mw_evaluateGradWithSpin

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
camelto2: Yes, mw_evaluateVGLWithSpin was taking RefVectors as arguments, so I constructed the vectors of values first and then use emplace_back(xxx_values.back()) to add the references to the RefVector. 

All this is to recompute the spin gradients...and this approach is just reusing existing functionality to compute the spin gradients since we aren't storing them and updating them along the way. I was reusing the existing mw_evaluateVGLWithSpin since it already accomplished what I needed. If I didn't want to recalculate the GL and just have an API to give the spin gradients only, then inside SpinorSet I would need a mw_evaluateValue for each of the up/down SPOSets that make up the spinor in order to construct the spin gradient.  That API is defined in SPOSet, but it is just a simple loop over the legacy single walker APIs evaluateValue. There isn't a mw_evaluateValue in SplineC2C that utilizes the offload...so it just seemed simpler to use the existing API. 

I spoke with @ye-luo on slack about this...basically there are two approached to this. First is we store the spin gradients and then need to either add to the existing structures (like OffloadMWVGLArray for example) or add new ones to accommodate for the extra spin gradients...and then need to update the values whenever we call mw_accept_rejectMove. The other option (the one chosen here) is to just recompute the spingradients from the SPOSet from scratch every time we call TrialWaveFunction::mw_evalGrad and the spin gradients are being used. Ye thought there may not be a significant penalty to just recomputing, and it seemed simpler so I tried that here. 

I would actually prefer the first approach where we store and update when we accept/reject moves, but I was having trouble deciding the best way to store the spin gradients so that we only allocate on the device when we need it =

camelto2: Doing this recompute is quite a bit simpler than the other option. I've been working on a version where I store the spin gradients and update with accept move, and that has required a lot more code changes the way I've been doing it. 

It looks like the recalculation of things that aren't being used was also done elsewhere in this file...see DiracDeterminantBatched::mw_calcRatio. There is a call to Phi->mw_evaluateVGLandDetRatioGrads even though an implementation of something like mw_evaluateVandDetRatio would be more approriate so there isn't any computation of things not being used.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
camelto2: So here, val is just a temporary ValueVector. In order to have the right size of OrbitalSetSize (from the wfc_leader.Phi->size() call) to push into the std::vector<ValueVector> and then push back into the reference vector. 

All of this was done to just make it so that I can pass the RefVectors into the mw_evaluateVGLWithSpin call. I was having trouble getting things initialized to the right size using the constructors, so I turned to reserving the vector and RefVector sizes and then empalce_back. 

This method of getting the RefVectors of the right size was done in some unit tests, https://github.com/QMCPACK/qmcpack/blob/develop/src/QMCWaveFunctions/tests/test_einset.cpp#L520-L543

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
```diff
@@ -197,6 +197,80 @@ typename DiracDeterminantBatched<DET_ENGINE>::Grad DiracDeterminantBatched<DET_E
...
+  {
+    auto& det = wfc_list.getCastedElement<DiracDeterminantBatched<DET_ENGINE>>(iw);
+    // capacity is the size of each vector in the VGL so this advances us to the g then makes
+    // an offset into the gradients
+    dpsiM_row_list[iw] = det.psiM_vgl.device_data() + psiM_vgl.capacity() + NumOrbitals * WorkingIndex * DIM;
+    dspin_row_list[iw] = dspin_values[iw].data();
+    engine_list.push_back(det.det_engine_);
+  }
+
+  DET_ENGINE::mw_evalGradWithSpin(engine_list, dpsiM_row_list, dspin_row_list, WorkingIndex, grad_now, spingrad_now);
```
ye-luo, line 266: Suspicious here. dspin_row_list collect host pointers but de-reference them on the device. dspin_row_list can be replaced with a matrix by copying values and then transfer the whole matrix to device.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
```diff
@@ -197,6 +197,80 @@ typename DiracDeterminantBatched<DET_ENGINE>::Grad DiracDeterminantBatched<DET_E
...
+  std::vector<SPOSet::ValueVector> psi_values, lap_values, dspin_values;
+  std::vector<SPOSet::GradVector> grad_values;
+  RefVector<SPOSet::ValueVector> psi_v_list, lap_v_list, dspin_v_list;
+  RefVector<SPOSet::GradVector> grad_v_list;
+
+  phi_list.reserve(nw);
+  psi_values.reserve(nw);
+  grad_values.reserve(nw);
+  lap_values.reserve(nw);
+  dspin_values.reserve(nw);
```
ye-luo, line 223: can you extract them from wfc_list instead of creating them on the fly?

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
PDoakORNL: Ok I see, it was far enough away that I missed it. I guess this could use a doc comment, in the spirit of self documentation consider:
``` c++
int num_orbitals = wfc_leader.Phi->size();
...
phi_list.emplace_back(num_orbitals);
```
Or If you wanted instead to move the construction outside of the for loop you could write the behavior of std::vector's constructors are pretty well known.
```
SPOSet::ValueVector sized_vector(wfc_leader.Phi->size());
std::vector<SPOSet::ValueVector> psi_values(wfc_list.size(),sized_vector);
```

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
PDoakORNL: Is it ok to update the state variables that would be extracted.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
camelto2: this has been fixed. this doesn't work with RefVector (which is std::vector<std::reference_wrapper<T>>, so I left those as they were

PDoakORNL: we have a helper function `makeRefVector` in `type_traits/template_types.hpp`

camelto2: Oh ok great, I wasn't aware of that. Will fix

camelto2: done

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp
camelto2: this has been addressed by moving mw_dspin to DualMatrix

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -281,7 +284,7 @@ class SPOSet : public QMCTraits
                                       const RefVector<ValueVector>& psi_v_list,
                                       const RefVector<GradVector>& dpsi_v_list,
                                       const RefVector<ValueVector>& d2psi_v_list,
-                                      const RefVector<ValueVector>& dspin_v_list) const;
+                                      DualMatrix<ComplexType>& mw_dspin) const;
```
ye-luo, line 287: Update the documentation and mention that the device side is up-to-date.

ye-luo, line 287: Update the documentation and mention that the device side is up-to-date.

src/QMCWaveFunctions/SpinorSet.cpp
```diff
@@ -157,45 +157,17 @@ void SpinorSet::mw_evaluateVGLWithSpin(const RefVectorWithLeader<SPOSet>& spo_li
...
-    mw_up_d2psi_work.emplace_back(tmp_val_vec);
-    up_d2psi_v_list.emplace_back(mw_up_d2psi_work.back());
-    mw_dn_d2psi_work.emplace_back(tmp_val_vec);
-    dn_d2psi_v_list.emplace_back(mw_dn_d2psi_work.back());
-  }
+  ValueVector sized_val(OrbitalSetSize);
+  GradVector sized_grad(OrbitalSetSize);
+  std::vector<ValueVector> mw_up_psi_work(nw, sized_val), mw_dn_psi_work(nw, sized_val);
+  std::vector<GradVector> mw_up_dpsi_work(nw, sized_grad), mw_dn_dpsi_work(nw, sized_grad);
+  std::vector<ValueVector> mw_up_d2psi_work(nw, sized_val), mw_dn_d2psi_work(nw, sized_val);
```
ye-luo, line 164: It is bad to create and destroy the work space. The per-walker vector can be collected.

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.h
```diff
@@ -74,6 +74,7 @@ class DiracDeterminantBatched : public DiracDeterminantBase
     std::vector<Grad> grad_new_local;
     // multi walker of spingrads
     std::vector<Value> spingrad_new_local;
+    DualMatrix<ComplexType> mw_dspin;
```
ye-luo, line 77: Documentation.

src/QMCWaveFunctions/SPOSet.h
camelto2: added documentation

src/QMCWaveFunctions/SpinorSet.cpp
camelto2: fixed to collect from per-walker work arrays that already existed

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.h
camelto2: added documentation

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -276,7 +276,8 @@ class SPOSet : public QMCTraits
    * @param psi_v_list the list of value vector pointers in a walker batch
    * @param dpsi_v_list the list of gradient vector pointers in a walker batch
    * @param d2psi_v_list the list of laplacian vector pointers in a walker batch
-   * @param dspin_v_list the list of spin gradients vector pointers in a walker batch
+   * @param mw_dspin is a dual matrix of spin gradients [nw][norb]
+   * Note that at the end of the function, the host and device data are synchronized
```
ye-luo, line 280: Based on the consumer pattern and to achieve optimal performance, I would assume only the device copy up-to-date instead of requiring both.

camelto2: Now it says that device side is up to date

src/QMCWaveFunctions/SPOSet.cpp
```diff
@@ -109,11 +109,9 @@ void SPOSet::mw_evaluateVGLWithSpin(const RefVectorWithLeader<SPOSet>& spo_list,
                                     const RefVector<ValueVector>& psi_v_list,
                                     const RefVector<GradVector>& dpsi_v_list,
                                     const RefVector<ValueVector>& d2psi_v_list,
-                                    const RefVector<ValueVector>& dspin_v_list) const
+                                    DualMatrix<ComplexType>& mw_dspin) const
 {
-  assert(this == &spo_list.getLeader());
-  for (int iw = 0; iw < spo_list.size(); iw++)
-    spo_list[iw].evaluateVGL_spin(P_list[iw], iat, psi_v_list[iw], dpsi_v_list[iw], d2psi_v_list[iw], dspin_v_list[iw]);
+  throw std::runtime_error(getClassName() + "::mw_evaluateVGLandDetRatioGradsWithSpin() is not supported. \n");
```
ye-luo, line 114: Need to improve the safe-guard in a later PR.

camelto2: oops, added the wrong name in the error message. Will fix

ye-luo:
Test this please

ye-luo:
Test this please

ye-luo (approved):

PR #4277, opened 2022-10-10
Add HIP_ARCH to ctest_script.cmake

prckent:
## Proposed changes

Add HIP_ARCH for convenient scripting

- No

nitrogen2 (gfx90a)

## Checklist


ye-luo:
I feel this is not needed and try to avoid expanding the list of bypass in ctest_scripts.
Please pass options `-DQMC_OPTIONS='-DHIP_ARCH=gfx90a'.`
If you need multiple entry `-DQMC_OPTIONS='-DHIP_ARCH=gfx90a;-DENABLE_OFFLOAD=ON'`

ye-luo:
With the current bypassing scheme, you will keep hitting this type of issue requiring updating ctest_scripts.
I reopened https://github.com/QMCPACK/qmcpack/issues/350

prckent:
Updating QMC_OPTIONS is a pain in the existing scripts and this gets them running fastest; I certainly don't have time to update them now. That said I do agree that something less error prone is desirable. An additional problem imo is that we don't have a good way to catch arguments that are effectively silently dropped. It wouldn't surprise me if we have a few cases where the "wrong" test configuration is being run due to missing options or typos etc.

ye-luo:
It should not be an all-in all-out approach. You may start adding new flags via QMC_OPTIONS and gradually phase out the old way. Probably start with adding QMC_OPTIONS variable parallel to CTCFG and at the ctest line add -DQMC_OPTIONS="$QMC_OPTIONS"

PR #4288, opened 2022-10-20
Update handling of array sizes in SplineR2R and SplineC2C for rotation cases

markdewing:
In the case of orbital rotation, the orbital set size from the H5 file may not match the number of SPOs used, which may not match the number of occupied orbitals.

In BsplineSet.h, the slices of the matrices need to have a size consistent with the size of the matrix.

In SplineR2R.cpp, use the size of psi rather than myV.  In the call to evaluate3d it doesn't matter because all the inputs are sized the same as myV, but in the following call to assign_*, the size discrepancy can cause invalid accesses.

Add a Python script for evaluating SPOs given the HDF file of coefficients.  It uses autograd to evaluate the spatial derivatives.

- Bugfix

- No

desktop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
@@ -86,7 +86,7 @@ void SplineC2C<ST>::evaluateValue(const ParticleSet& P, const int iat, ValueVect
 #pragma omp parallel
   {
     int first, last;
-    FairDivideAligned(myV.size(), getAlignment<ST>(), omp_get_num_threads(), omp_get_thread_num(), first, last);
+    FairDivideAligned(2 * psi.size(), getAlignment<ST>(), omp_get_num_threads(), omp_get_thread_num(), first, last);
 
     spline2::evaluate3d(SplineInst->getSplinePtr(), ru, myV, first, last);
     assign_v(r, myV, psi, first / 2, last / 2);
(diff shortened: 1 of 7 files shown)


src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
```diff
@@ -86,7 +86,7 @@ void SplineC2C<ST>::evaluateValue(const ParticleSet& P, const int iat, ValueVect
 #pragma omp parallel
   {
     int first, last;
-    FairDivideAligned(myV.size(), getAlignment<ST>(), omp_get_num_threads(), omp_get_thread_num(), first, last);
+    FairDivideAligned(2 * psi.size(), getAlignment<ST>(), omp_get_num_threads(), omp_get_thread_num(), first, last);
```
PDoakORNL, line 89: I'm not completely sure this magic number is from the same cause as all the other bare 2's in this file, but who knows. The line of documentation in the header actually manages to use the word "double" in an ambiguous way until you read the .c

PDoakORNL (changes requested):
Haven't look at this file in a while. I feel kinda nauseous.  Considering the centrality of spline evaluation in the code its sad that it is basically C from the nineties with almost no documentation.

src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
prckent: Can we replace these 2s with a const named something about complex? I also noted the bare 2s when reading the file.

PDoakORNL (dismissed):

PDoakORNL:
test this please

prckent (dismissed):

markdewing:
Test this please

markdewing:
Test this please

ye-luo (approved):

ye-luo:
Test this please

Issue #3868, opened 2022-02-22
How to resurrect true 2D capability?


Paul-St-Young:
**Problem**
`OHMMS_DIM` is planned for obsolescence. With its removal, all code snippets for true 2D calculation will be history.

**Proposed Solution**
Enable true 2D calculation via xml input.

**To-do List:**
- [ ] 2D potential energy
- [ ] 2D Jastrow wavefunction (kinetic energy)
- [ ] 2D determinant wavefunction
- [ ] 2D VMC driver
- [ ] 2D DMC driver

prckent:
Please take a look at #3867 and think generally about how the dimensionality might be hidden/abstracted. #3867 will eventually get spin-orbit and non-spin orbit to use the same batched drivers. (The old cpu serial drivers are also heading for retirement since they will never run on GPUs efficiently). Realistically I don't think we can have multiple drivers for each dimensionality, at least the way we have done them historically - the humans simply don't scale. If you can come up with a design that is compatible with a generic driver you will also get wavefunction optimization, reptation etc. for "free", and 1D might already be nearly supported. For sure other people will have ideas on this.

Paul-St-Young:
@prckent from a preliminary look, I don't think the MCCoords abstraction poses any difficulty at all.
For true 2D, one can:
1. fill the z-component of all vectors in `walker_deltas.positions` with zeros.
2. check that the z-component of all vectors in `drifts` are zero.

PR #2806, opened 2020-12-04
Generalized 2 particle cusp

shivupa:
## Proposed changes
This generalizes the hard coded cusp condition to -2 q_i q_j mu_ij / (d \pm 1). In the 1D case, the denominator is always taken to be the plus case ([see footnote 45](https://arxiv.org/abs/2003.06506)). 

## What type(s) of changes does this code introduce?
- Bugfix
- New feature

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
Workstation (i7 2600s)
## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- Yes/No? This PR adds tests to cover any new code, or to catch a bug that is being fixed
Should a test be added with various particle masses? 
- N/A. Documentation has been added (if appropriate)

PR #4289, opened 2022-10-21
make two-body Jastrow work in 2D

Paul-St-Young:
## Proposed changes

Follow up on #3868, enable Jastrow in 2D.
The WavefunctionTester was modified to run in 2D (in addition to 3D) to test these changes.

- New feature

- No

workstation

## Checklist


ye-luo (changes requested):
Could you please add a unit test. WaveFunctionTest is not the optimal way of making a test because it cannot provide enough safeguard. If the value, gradients are set zero, WaveFunctionTest cannot catch it.

prckent (changes requested):
Thanks Paul. Echoing Ye's comments, please stay away from wftester for formal testing. We have established routes for that via unit testing (by far preferred), followed by integration testing. 

Background: wftester is a historical clearly non-looked after part of the code that we should not be increasing dependencies on. Other than this PR from you and a PR from me, there are no existing uses or tests. Perhaps we'll rewrite it, remove it, or trim it. Some days I have been tempted to just delete it, since it would be faster to start afresh with clean code. If there is something that is unclear how to write from a unit test please ask.

Paul-St-Young:
@ye-luo @prckent The WavefunctionTester is by far the easiest route to test particle and parameter derivatives of a complicated wavefunction. When I doubt the correctness of a run, it's easy replace the <qmc method="vmc/dmc"> section with <qmc method="wftest"> for a quick test. I know that @rcclay and my self have made many uses of this feature, for example when debugging backflow. @markdewing also made the FiniteDifference class fairly clean.

WavefunctionTester is also my go-to when developing a new wavefunction component, for example when resurrecting 2D electron gas orbitals #4084.
Since we have no other facility for generally checking the correctness of analytical derivatives (either for single-particle orbital, or for the many-body wavefunction), I don't see why we should remove this feature.
Having to re-code finite-difference for every unit test is not productive IMO.

I am happy to put in a simple regression unit test, but I would like to keep the wftest to check for correctness.

src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
```diff
@@ -223,7 +224,7 @@ std::unique_ptr<WaveFunctionComponent> RadialJastrowBuilder::createJ2(xmlNodePtr
 #if OHMMS_DIM == 1
         RealType dim_factor = 1.0 / (OHMMS_DIM + 1);
 #else
-        RealType dim_factor = (ia == ib) ? 1.0 / (OHMMS_DIM + 1) : 1.0 / (OHMMS_DIM - 1);
+        RealType dim_factor = (ia == ib) ? 0.5 / (ndim - 1) : 1.0 / (ndim - 1);
```
markdewing, line 227: For 2 dimensions, the factor for the case ia==ib changes from 1/3 to 1/2.  Is that change intended?

prckent:
Fair point on not recoding the finite derivatives.

src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
Paul-St-Young: Yes, this is an intentional change.
The same-spin cusp should always be half of that of the opposite-spin cusp.
Did I miss something?

prckent:
@Paul-St-Young You use `<parameter name="ratio">deriv</parameter>`?

src/QMCDrivers/WaveFunctionTester.cpp
```diff
@@ -69,6 +70,12 @@ WaveFunctionTester::WaveFunctionTester(const ProjectData& project_data,
 
   deltaR.resize(w.getTotalNum());
   makeGaussRandom(deltaR);
```
prckent, line 72: Any reason not to make makeGaussRandom dimension aware and avoid the changes and logging on later lines?

src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
markdewing: The previous code was added in #2806 , and it references equation 9 of https://arxiv.org/abs/2003.06506 (and footnote 45 for the 1-D case).  According to that formula, it looks like the same-spin cusp is half of the opposite spin cusp for d=3, but not for d=2.

prckent (changes requested):
Just noticed: the tests are not being run in CI. Please can you rename them with the prefix deterministic- so that "ctest -L deterministic" will pick them up. I think the name is sufficient. Thanks.

edit: for consistency with the existing heg tests, how about "deterministic-heg2d-...". That avoids introducing a new abbreviation that people will have to learn the meaning of.

prckent:
Test this please

ye-luo:
> @ye-luo @prckent The WavefunctionTester is by far the easiest route to test particle and parameter derivatives of a complicated wavefunction. When I doubt the correctness of a run, it's easy replace the `<qmc method="vmc/dmc">` section with `<qmc method="wftest">` for a quick test. I know that @rcclay and my self have made many uses of this feature, for example when debugging backflow. @markdewing also made the FiniteDifference class fairly clean.
> 
> WavefunctionTester is also my go-to when developing a new wavefunction component, for example when resurrecting 2D electron gas orbitals #4084. Since we have no other facility for generally checking the correctness of analytical derivatives (either for single-particle orbital, or for the many-body wavefunction), I don't see why we should remove this feature. Having to re-code finite-difference for every unit test is not productive IMO.
> 
> I am happy to put in a simple regression unit test, but I would like to keep the wftest to check for correctness.

You can use wftest for your development but the final test being added needs to be a unit test. wftest does random sampling which is not desired in unit tests. I don't see where you get the impression of the need of finite-difference in unit test. Unit test should check against stored values. See https://github.com/QMCPACK/qmcpack/blob/develop/src/QMCWaveFunctions/tests/test_polynomial_eeI_jastrow.cpp

src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
Paul-St-Young: Yes, you are right. I must have made a mistake in my own derivation.
Thanks for catching this!

Paul-St-Young:
> Just noticed: the tests are not being run in CI. Please can you rename them with the prefix deterministic- so that "ctest -L deterministic" will pick them up. I think the name is sufficient. Thanks.
> 
> edit: for consistency with the existing heg tests, how about "deterministic-heg2d-...". That avoids introducing a new abbreviation that people will have to learn the meaning of.

Thanks fo advice! I have changed the name of the test.

prckent:
Test this please

src/QMCDrivers/WaveFunctionTester.cpp
Paul-St-Young: The reason is that I wanted the change to be localized in the WaveFunctionTester.

Eventually, what you suggested is nicer. Although, I would have to touch a lot of files to make this API change.
I wonder if it's worth making another PR for that.

prckent: Hi Paul - In that case I agree that making another PR would be a good way of splitting this up

ye-luo:
@Paul-St-Young It seems failing with address sanitizer https://github.com/QMCPACK/qmcpack/actions/runs/3315476545/jobs/5476115646

prckent:
Paul - Are you able to reproduce any of these problems locally? e.g. Our gcc9 " no mpi and no omp " build is crashing with a segfault.

Paul-St-Young:
Yes, I was able to reproduce the crash.
The problem is only in the real build.

The input has 2 up electrons in 2 PW orbitals.
The 2 complex PW orbitals 0 and e^ikr are fine, but there are 3 real PW orbitals: 0, sin(kr), and cos(kr).
The extra memory was not allocated in DiracDeterminant.

How may I resize DiracDeterminant from within an SPOSet builder?

ye-luo:
DiracDeterminant query the size from SPOSet. Make sure it is correct at the first place.

prckent:
I agree with Ye on this one. We should create diracdeterminants with the correct size to begin with. Cleaner and clearer. The old trope of changing the size of things after initial construction hurts the accessibility of the code and is not good for maintenance.

Paul-St-Young:
@prckent @PDoakORNL Yes, I agree with your view.
That that `resize` in the snippet above is from the constructor of DiracDeterminant, so it's not changing the size of things after initial construction.
The question is whether to initialize the determinants to `(NumPtcls, NumPtcls)` or  `(NumPtcls, norb)`.

ye-luo:
The determinant matrices must be strictly square. `(NumPtcls, NumPtcls)`. In the FreeOrbital input, you specified size 2 in the input. So the object should work for exactly  or up to 2 orbitals. Based on what you mentioned that FreeOrbital creates 3 orbitals in the real case, then a truncation must be done at the construction of FreeOrbital object.

Paul-St-Young:
Thanks @ye-luo! Your explanation was very helpful.

I think the fundamental problem was that I set up a system that truly needs a complex wavefunction.
The 2 up 2 dn system at Gamma should have a complex single determinant wavefunction, because there is no -k PW to pair up with the +k PW to make sin and cos.

I added some checks in FreeOrbitalBuilder to abort in a situation as described above.

I also edited the example to contain 5 up 5 dn (closed-shell in 2D) and now things run through.

src/QMCDrivers/WaveFunctionTester.cpp
```diff
@@ -581,8 +593,8 @@ void WaveFunctionTester::computeNumericalGrad(RealType delta,
                                               ParticleSet::ParticleGradient& G_fd, // finite difference
                                               ParticleSet::ParticleLaplacian& L_fd)
 {
-  FiniteDifference fd(FiniteDifference::FiniteDiff_LowOrder);
-  //FiniteDifference fd(FiniteDifference::FiniteDiff_Richardson);
+  FiniteDifference fd(ndim, FiniteDifference::FiniteDiff_LowOrder);
+  //FiniteDifference fd(ndim, FiniteDifference::FiniteDiff_Richardson);
```
ye-luo, line 597: Remove this commented line

src/QMCDrivers/WaveFunctionTester.cpp
```diff
@@ -728,9 +740,9 @@ bool WaveFunctionTester::checkGradientAtConfiguration(MCWalkerConfiguration::Wal
   {
     delta = deltaParam;
   }
-  FiniteDifference fd(FiniteDifference::FiniteDiff_LowOrder);
+  FiniteDifference fd(ndim, FiniteDifference::FiniteDiff_LowOrder);
   //RealType delta = 1.0;
-  //FiniteDifference fd(FiniteDifference::FiniteDiff_Richardson);
+  //FiniteDifference fd(ndim, FiniteDifference::FiniteDiff_Richardson);
```
ye-luo, line 745: Remove commented lines.

src/QMCWaveFunctions/ElectronGas/FreeOrbitalBuilder.cpp
```diff
@@ -31,6 +34,14 @@ std::unique_ptr<SPOSet> FreeOrbitalBuilder::createSPOSetFromXML(xmlNodePtr cur)
   app_log() << "twist cartesian = " << tvec << std::endl;
 #else
   npw = std::ceil((norb + 1.0) / 2);
+  if (2*npw-1 != norb)
+  {
+    std::ostringstream msg;
+    msg << "norb = " << norb << " npw = " << npw;
+    msg << " cannot be ran in real PWs (sin, cos)" << std::endl;
+    msg << "either use complex build or change the size of SPO set" << std::endl;
```
ye-luo, line 42: Do you have recommendation to print about how to "change the size of SPO set"?

src/QMCWaveFunctions/ElectronGas/FreeOrbitalBuilder.cpp
```diff
@@ -31,6 +34,14 @@ std::unique_ptr<SPOSet> FreeOrbitalBuilder::createSPOSetFromXML(xmlNodePtr cur)
   app_log() << "twist cartesian = " << tvec << std::endl;
 #else
   npw = std::ceil((norb + 1.0) / 2);
```
ye-luo, line 36: Remove int npw in line 15 and put
`int npw =` here and line 31
Also please do a clang-formatting.

src/QMCWaveFunctions/Jastrow/RadialJastrowBuilder.cpp
```diff
@@ -217,13 +218,13 @@ std::unique_ptr<WaveFunctionComponent> RadialJastrowBuilder::createJ2(xmlNodePtr
...
       if (cusp < -1e6)
-      {
+      { // see eq. (9) in https://arxiv.org/abs/2003.06506
         RealType qq       = species(chargeInd, ia) * species(chargeInd, ib);
         RealType red_mass = species(massInd, ia) * species(massInd, ib) / (species(massInd, ia) + species(massInd, ib));
 #if OHMMS_DIM == 1
         RealType dim_factor = 1.0 / (OHMMS_DIM + 1);
 #else
-        RealType dim_factor = (ia == ib) ? 1.0 / (OHMMS_DIM + 1) : 1.0 / (OHMMS_DIM - 1);
+        RealType dim_factor = (ia == ib) ? 1.0 / (ndim + 1) : 1.0 / (ndim - 1);
```
ye-luo, line 227: Remove `#if OHMMS_DIM == 1`

ye-luo (commented):
Minor changes requested.

ye-luo (changes requested):
Please add random seeds in both determinsitic test xml.

Paul-St-Young:
@ye-luo I think I addressed all of your comments. Please see if this is satisfactory now.
The macos error seems to be an h5py install error, so not related to this PR

prckent:
I reran the failed macos test and it passed. Seems to have failed due to a transient infrastructure problem, nothing to do with this PR.

ye-luo (approved):
LGTM
@prckent do you still have remaining concerns?

prckent (approved):

PR #4295, opened 2022-10-27
Extract run_qmc to separate module

markdewing:
Generalize the run_qmc function to an arbitrary number of electrons and extract it to a separate module.
Future validation code will use different wavefunctions and we want to reuse the run_qmc function.

The implementation of psi in the wavefunction needs to be split into two parts.  For generality, calling from run_qmc uses an array for all the electron positions.  Autograd is easier to use if each position is a separate argument.  The solution is to make psi a wrapper that extracts individual positions and calls the internal implementation.

- No

desktop

## Checklist


ye-luo:
Test this please

prckent:
Aside: if you did Gaussian distributed moves to compute delta/the new trial position, run_qmc would be doing conventional VMC.

prckent (approved):
LGTM

prckent:
Test this please

Issue #2709, opened 2020-09-14
CSVMC effectively untested due to WALKER_MAX_PROPERTIES


prckent:
**Describe the bug**

The CSVMC tests e.g. https://cdash.qmcpack.org/CDash/testDetails.php?test=9832855&build=141686 fail with default builds due to the WALKER_MAX_PROPERTIES workaround/problem. This means they are effectively untested since all the nightly/weekly tests use the default.

If a "reasonable" WALKER_MAX_PROPERTIES is identified we could update the builds, otherwise we'll have to wait for WALKER_MAX_PROPERTIES to be redesigned/refactored away. The next person interested in CSVMC should experiment and add comments here.

**To Reproduce**

Current develop, ctest -R csvmc

**Expected behavior**

Test should run to completion and pass/fail.

**System:**

All the systems and builds reporting to cdash regularly

**Additional context**

markdewing:
See also #2509

jtkrogel:
CSVMC is dead

Issue #2509, opened 2020-06-03
Constant size matrix errors for CSVMC


jtkrogel:
The CSVMC tests are currently failing due to constant size matrix errors.  A stopgap would be to recompile the code at all nightly test sites to use a large enough `WALKER_MAX_PROPERTIES` to clear these particular tests, but I think this issue should remain open (or at least another one created and remain open) until archaic compile time sizing is no longer required.

See:
https://cdash.qmcpack.org/CDash/testDetails.php?test=8433143&build=120036
https://cdash.qmcpack.org/CDash/testDetails.php?test=8433144&build=120036
https://cdash.qmcpack.org/CDash/testDetails.php?test=8433142&build=120036

ddinge:
I'm seeing this error on an ARM machine with gnu compiler and open mpi. I don't think increasing "WALKER_MAX_PROPERTIES" will fix it. I tried changing it from the previous 2048 to 4096 and it made no difference. Resizes where the first dimention goes from 1 to 2 in the message below seem to generate the error. Resizes with 1,x are fine.

Error message "You cannot resize a constant size matrix beyond its initial max dimensions ( 2,14 > 1,2048 )"

I'm a newby to the code so this is a phenomenological observation not based on a knowledge of the code logic.   Perhaps someone can chime in on what the code is attempting to do here.

prckent:
@ddinge  There is nothing wrong with your compiler here. The code is unfortunately behaving as it is meant to, if not how we would like it to in future.

The code needs a redesign in this area to fix. There was a problem in the legacy dynamic sizing of properties that was temporarily patch/bypassed some months back by defining a maximum number of properties. This is problematic for some qmc methods (CSVMC) and for some observables as you have discovered but it OK for vanilla VMC and DMC runs. Basically the legacy code/design needs replacing to better to suit current needs.

markdewing:
The issue is ConstantSizeMatrix can only be resized along dimensions independently.  Even if there is enough underlying storage to accommodate the resized matrix, it still fails.
One solution is to compare the total size when resizing:
```
--- a/src/Containers/MinimalContainers/ConstantSizeMatrix.hpp
+++ b/src/Containers/MinimalContainers/ConstantSizeMatrix.hpp
@@ -160,13 +160,23 @@ public:
 
   void resize(size_t m, size_t n)
   {
-    if (n > n_max_ || m > m_max_)
+    if (n*m > n_max_* m_max_)
     {
       std::ostringstream error;
[...]

PR #4307, opened 2022-11-04
Allow ConstantSizeMatrix to increase along a dimension

markdewing:
Allow increasing one dimension in ConstantSizeMatrix, as long the other dimension decreases to compensate and the new capacity is less than or equal to the existing capacity.

This fixes the WALKER_MAX_PROPERTIES error in the csvmc tests (#2509 and #2709).
(I got annoyed seeing the csvmc failures on cdash in the nightly tests.)
 
- Bugfix

- No

desktop

## Checklist

### DIFF
--- a/src/Containers/MinimalContainers/ConstantSizeMatrix.hpp
+++ b/src/Containers/MinimalContainers/ConstantSizeMatrix.hpp
@@ -160,13 +160,26 @@ public:
 
   void resize(size_t m, size_t n)
   {
-    if (n > n_max_ || m > m_max_)
+    if (n * m > n_max_ * m_max_)
     {
       std::ostringstream error;
       error << "You cannot resize a constant size matrix beyond its initial max dimensions ( " << m << "," << n << " > "
             << m_max_ << "," << n_max_ << " )\n";
       throw std::domain_error(error.str());
     }
+
+    size_t capacity = n_max_ * m_max_;
+    if (n > n_max_)
+    {
+      n_max_ = n;
+      m_max_ = capacity / n_max_;
+    }
+    if (m > m_max_)
+    {
+      m_max_ = m;
+      n_max_ = capacity / m_max_;
+    }
+
     n_ = n;
     m_ = m;
   }
@@ -182,8 +195,6 @@ private:
   size_t m_max_;
   size_t n_max_;
   std::vector<T, ALLOC> data_;
-
-
 };
 
 extern template class ConstantSizeMatrix<float>;
--- a/src/Containers/MinimalContainers/ConstantSizeMatrix.hpp
+++ b/src/Containers/MinimalContainers/ConstantSizeMatrix.hpp
@@ -24,14 +24,14 @@ class ConstantSizeMatrix
 {
 public:
   ConstantSizeMatrix(size_t m, size_t n, size_t m_max, size_t n_max, T val = T())
-      : m_(m), n_(n), m_max_(m_max), n_max_(n_max), data_(n_max * m_max, val)
+      : m_(m), n_(n), m_max_(m_max), n_max_(n_max), capacity_(n_max * m_max), data_(n_max * m_max, val)
   {
     if (n_max <= 0 || n > n_max || m_max <= 0 || m > m_max)
       throw std::runtime_error("Cannot construct ConstantSizeMatrix with an invalid size");
   }
 
   ConstantSizeMatrix(const ConstantSizeMatrix& csm)
-      : m_(csm.m_), n_(csm.n_), m_max_(csm.m_max_), n_max_(csm.n_max_), data_(csm.n_max_ * csm.m_max_)
+      : m_(csm.m_), n_(csm.n_), m_max_(csm.m_max_), n_max_(csm.n_max_), capacity_(csm.capacity_), data_(csm.capacity_)
   {
     //I don't just do an = here because of the posible semantics of allocator propagation.
     data_.assign(csm.begin(), csm.end());
@@ -151,7 +151,7 @@ public:
   T* data() { return data_.data(); }
   const T* data() const { return data_.data(); }
 
-  size_t capacity() { return n_max_ * m_max_; }
+  size_t capacity() { return capacity_; }
   size_t n_capacity() { return n_max_; }
 
   size_t size() const { return n_ * m_; }
@@ -160,9 +160,7 @@ public:
 
   void resize(size_t m, size_t n)
   {
-    size_t max_capacity = capacity();
-
-    if (n * m > max_capacity)
+    if (n * m > capacity())
     {
       std::ostringstream error;
       error << "You cannot resize a constant size matrix beyond its initial max dimensions ( " << m << "," << n << " > "
@@ -173,12 +171,13 @@ public:
     if (n > n_max_)
     {
       n_max_ = n;
-      m_max_ = max_capacity / n_max_;
+      m_max_ = capacity() / n_max_;
     }
+
     if (m > m_max_)
     {
       m_max_ = m;
-      n_max_ = max_capacity / m_max_;
+      n_max_ = capacity() / m_max_;
     }
 
     n_ = n;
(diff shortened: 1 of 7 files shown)


prckent (commented):
Excellent - I thought we were stuck with this "feature" without major plumbing.

Can you arrange to get test coverage for the n>n_nmax_, m>m_max_ cases? For these fundamental classes we should strive for highest coverage.

src/Containers/MinimalContainers/ConstantSizeMatrix.hpp
```diff
@@ -160,13 +160,26 @@ class ConstantSizeMatrix
...
-    if (n > n_max_ || m > m_max_)
+    if (n * m > n_max_ * m_max_)
     {
       std::ostringstream error;
       error << "You cannot resize a constant size matrix beyond its initial max dimensions ( " << m << "," << n << " > "
             << m_max_ << "," << n_max_ << " )\n";
       throw std::domain_error(error.str());
     }
+
+    size_t capacity = n_max_ * m_max_;
```
quantumsteve, line 171: calculate capacity before line 163

markdewing: Done.  I also changed the name of the variable from `capacity` to `max_capacity` to avoid confusion with the `capacity()` function.

PDoakORNL (changes requested):
The reason this matrix is constant size is because the walker properties matrix has/had a very complex state relation with the walker DataSet. In the batched code this has been reduced but walker properties but they may never change in size after the intial walkers are created. 

Having them change dimensionality but not size might be okay but you would need to guarantee this state update propagated to every walker living or dead or have Dataset corruption on walker transfers.

PDoakORNL:
I'd tentatively say this looks ok. This class wasn't intended to be permanent but just to keep legacy abuse of the properties buffer from making the walker a constant source of memory corruption in the batched code. But I'd hate to see new code depend on hacking it. What really needs to happen is the notion of walker "properties" needs to be re examined.  It can be runtime determined but this can't be through a adhoc series of state transformations because no one wants to deal with input responsibly.  It also doesn't seem necessary to just throw out type safety anywhere near it because someone decided to just stuff it all into a byte buffer with a bunch of other datatypes 10+ years ago.

markdewing:
It looks like CSVMC is the only test case that runs into this condition of resizing the first dimension of the properties.
It looks like there's only one code path that will resize the first dimension, and that gets called from QMCDriver::process.  It calls W.resetWalkerProperty with the number of hamiltonians, which I think CSVMC is the only case (?) that has multiple wavefunctions and hamiltonians.

src/Containers/MinimalContainers/tests/test_ConstantSizeMatrix.cpp
```diff
@@ -28,8 +28,16 @@ TEST_CASE("ConstantSizeMatrix basics", "[containers]")
...
 
-  CHECK_THROWS(cmat.resize(2, 9));
+  CHECK_NOTHROW(cmat.resize(2, 9));
+  CHECK(cmat.capacity() == 32);
+
+  // If the new dimensions are not commensurable with the old ones,
+  // the capacity could decrease, even though the size of the underlying
+  // storage does not decrease.
+  CHECK_NOTHROW(cmat.resize(3, 9));
+  CHECK(cmat.capacity() == 30);
```
PDoakORNL, line 40: This change in the semantics of capacity() is no good for other code such as in Walker::registerData(). They are assume to match "std::vector::capacity() returns number of elements that can be held in the allocated storage." This remains 32 even if you have changed n_max and m_max such that you can't index all the allocated storage.  You could resize again and get access to all the allocated storage and the capacity.

The constant size for this object is the max number of elements so that a constant offset to the next element cant be maintained when it is packed into byte buffer and sent over the wire.

My thought on returning n_max_ * m_max_ is that was the area we committed to copying onto the wire.  The actual std::vector::capacity() could always be in blocks of 32 or whatever it just must be  => size().

I think a member variable for the "constant size" i.e. the capacity of the container not the underlying vector is necessary for this change.

PDoakORNL (approved):
Ok I see one change than needs to be made.  Also test coverage needs to be better, although I don't think its as bad as the coverage tool suggests.  `n > n_max` looks covered to me.

PDoakORNL (dismissed):
meant for that to be a request changes.

prckent:
A quick poll on Slack did not find any current users, but it did find occasional users. Unfortunately the architectural problems are real.

In the past I see that I wrote things like `I would prefer not to fix the nightlies since it will hide the very clear user-facing problem`. A greater problem is that we are very good at ignoring failing tests in the nightlies, since too many of the stochastic tests fail. The best tests are those run in CI, i.e. the deterministic set. These can't be ignored. Therefore if we can get a deterministic test for CSVMC it would be worth patching the ConstantSizeMatrix issue. Otherwise we will still be able to break the code and not notice/ignore.

markdewing:
Added a `capacity_` member variable that is kept fixed, even if n_max or m_max are changed.

It looks like n_max is the only one that matters as it's used as a leading dimension for 2-d array access.   In theory, the class could just take a leading dimension and a capacity, and reshaping it would just change the leading dimension.

It seems worth fixing this error in the nightly tests.  For one, there is a chicken-and-egg problem that creating the deterministic test requires the constant matrix to be fixed first.  Additionally, when I look a the nightly tests, I guess the failure mode based on the suffix of the test.  If it's a measurement test (test name ends in "-kinetic" or whatever), and there's no corresponding base test failure, it's likely a statistical failure.
I think that's why the csvmc test failure annoys me - it mentally falls into the category of 'real failure' when scanning CDash.

src/Containers/MinimalContainers/ConstantSizeMatrix.hpp
```diff
@@ -181,9 +194,8 @@ class ConstantSizeMatrix
   size_t n_;
   size_t m_max_;
   size_t n_max_;
+  size_t capacity_;
```
quantumsteve, line 197: is `capacity` always equal to `data_.size()`?

prckent:
@markdewing Lets stretch for a deterministic test which will give CSVMC a high chances of staying alive. One would be small enough to add to this PR... at least if the macros don't need updating or unless CSVMC turns out to have deterministic problems itself... Please take a look.

src/Containers/MinimalContainers/ConstantSizeMatrix.hpp
markdewing: Yes, I think so.

prckent:
Test this please

markdewing:
Test this please

src/Containers/MinimalContainers/tests/test_ConstantSizeMatrix.cpp
```diff
@@ -28,8 +28,16 @@ TEST_CASE("ConstantSizeMatrix basics", "[containers]")
...
 
   CHECK_THROWS(cmat.resize(1, 33));
+  CHECK_THROWS(cmat.resize(34, 2));
 
-  CHECK_THROWS(cmat.resize(2, 9));
+  CHECK_NOTHROW(cmat.resize(2, 9));
+  CHECK(cmat.capacity() == 32);
+
+  // If the new dimensions are not commensurable with the old ones,
+  // the capacity could decrease, even though the size of the underlying
```
PDoakORNL, line 37: This comment is confusing since directly below you show the capacity doesn't decrease. I think what you mean is the maximum size consistent with the resize arguments can decrease.

markdewing: The comment is no longer valid - the capacity does not decrease.  I'll remove it.

prckent (approved):
Approving. Thanks Mark.

While the walker_max_properties situation is not ideal, there is more extensive cleanup and rework needed in csvmc, not related to this PR. e.g. Whoever batches it will also need to reconsider the input handling and what we want to have in a modern driver, particularly one that handles multiple wavefunctions. Since csvmc has so few users, I don't think updating these should be any concern.

We'll also need some  documentation and a real-world example worked-through, since I couldn't find a CSVMC one.

In the interim this update keeps the code alive in CI, and the asan fix has improved it.

PR #4315, opened 2022-11-06
remove spurious assignment

correaa:
## Proposed changes

Remove spurious assignments that can be easily replaced by copy- (actually duplicate-) constructor.
The exact meaning of (re)assignment in BMPI3 communicator can change in the near future, so it is better to avoid it.
As well as two-phase construction in general.

- No

## Checklist


ye-luo (approved):

ye-luo:
Test this please

PR #4316, opened 2022-11-07
add function NodeComm to avoid two-step initialization

correaa:
## Proposed changes

Add functions to avoid two-step initialization (and simplify Communicate interface).

- No

## Checklist


correaa:
please test this even if it is marked as Draft

ye-luo:
All the MPI communicators in QMCPACK should be initialized from the QMCPACK global comm instead of from the mpi3 owned global one. I don’t see the point of added functionality.

ye-luo:
Probably you can check in the use side of code. That may help me understand your intention.

correaa:
The idea is not to add a functionality, but to soon remove the uses of `initializeAsNodeComm`, an ill-conceived function IMO.

The current implementation does a two-step initialization and depends on BMPI3 supporting two-step initialization also.

Also, `initializeAsNodeComm` repeats code from the constructor for example.
The alternative I am proposing doesn't.

correaa:
> Probably you can check in the use side of code. That may help me understand your intention.

Good point, that is the spirit and that is what I am doing, `initializeAsNodeComm` is used only once in the code, and right after declaring the variable it modifies. IMO another indication that it is a spurious use. (It is used in another instance but only to set up test and in the same way).

I am testing on my side as well, a few more commits will come.
If you are interested please monitor how I propose to modify the code.

correaa:
To clarify, these changes are needed because I am going to deprecate `mpi3::communicator` _(re)assignment_. Which after a long thought I got to the conclusion that it doesn't play well with the MPI model. QMCPACK depended _accidentally_ (not in a fundamental way) on that feature and this PR fixes it.

markdewing:
test this please

src/QMCApp/QMCMain.cpp
```diff
@@ -66,8 +66,8 @@ QMCMain::QMCMain(Communicate* c)
       traces_xml(NULL)
 #endif
 {
-  Communicate node_comm;
-  node_comm.initializeAsNodeComm(*OHMMS::Controller);
+  Communicate node_comm{OHMMS::Controller->NodeComm()};
+// node_comm.initializeAsNodeComm(*OHMMS::Controller);
```
markdewing, line 70: Commented-out code

correaa: good catch, I don't know why the tests didn't catch this. Should I push the change or is it already applied to the PR?

correaa (commented):
if everything looks right, I think the PR is finished. This will allow me to update the BMPI3 subtree.

src/QMCApp/QMCMain.cpp
markdewing: Push the change.
We don't have any tools or tests to catch commented-out code. 
(I do this all the time, too - comment out code temporarily while developing, and sometimes miss it when checking  in the code)

markdewing (approved):

markdewing:
test this please

src/QMCApp/QMCMain.cpp
correaa: Ah, sorry I was confused (the code reviews in github) always trip me. *I* did the change (and forgot to remove the comment), which I didn't remember doing but now I do. 

Maybe a way to avoid commented code is to 1) use .clang-tidy or .clang-format or cppcheck to complain about badly formatted comment (they can do it) and 2) get used to badly format commented code on purpose during development (for example not leaving a space after `//`).

PR #4328, opened 2022-11-18
Use std::chrono clock in profile timers

markdewing:
Replace timer based on gettimeofday with std::chrono timer (system_clock) in the profile timers.

Reasons:
1. I would like to add the event profile code, and it works better with std::chrono timers
2. Storing the start time as a double (since the epoch) loses precision - the elapsed time ends up being a small difference between large numbers.  This isn't as major an issue with aggregated profile data, but is noticeable in the event data.
3. Moves to a C++ standard API

The runtime manager was not switched in this PR, but could be changed in the future (and the the FakeCPUClock removed)

The test uses "chrono_literals" ( https://en.cppreference.com/w/cpp/header/chrono#Literals ), which uses C++ 11 user-defined literals ( https://en.cppreference.com/w/cpp/language/user_literal ).   The "1.1s" literal syntax can be a little disorienting if one has not encountered it before.

- New implementation of existing functionality

- No

desktop

## Checklist

- Ye. Code added or changed in the PR has been clang-formatted

### DIFF
--- a/src/Utilities/Clock.h
+++ b/src/Utilities/Clock.h
@@ -19,9 +19,33 @@
 
 #include <sys/time.h>
 #include <stddef.h>
+#include <chrono>
 
 namespace qmcplusplus
 {
+
+
+// Implements a std::chrono clock
+// See https://github.com/korfuri/fake_clock
+
+class FakeChronoClock
+{
+public:
+  using duration   = std::chrono::nanoseconds;
+  using rep        = duration::rep;
+  using period     = duration::period;
+  using time_point = std::chrono::time_point<FakeChronoClock>;
+  static time_point now() noexcept
+  {
+    fake_chrono_clock_value += fake_chrono_clock_increment;
+    return fake_chrono_clock_value;
+  }
+
+  static time_point fake_chrono_clock_value;
+  static duration fake_chrono_clock_increment;
+};
+
+
 /** functor for fake clock
  * calling FakeCPUClock()() returns the clock value
  */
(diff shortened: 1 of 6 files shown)


src/Utilities/Clock.h
```diff
@@ -19,9 +19,33 @@
...
+#include <chrono>
 
 namespace qmcplusplus
 {
+
+
+// Implements a std::chrono clock
+// See https://github.com/korfuri/fake_clock
+
+class FakeChronoClock
```
ye-luo, line 31: Can we use FakeChronoClock to replace all FakeCPUClock?

markdewing: The runtime manager was not switched in this PR, but could be changed in the future (and the the FakeCPUClock removed)
I would prefer not to put that change in this PR, in order to keep PR's smaller.

ye-luo: Hopefully the switch will happen relatively soon.

src/Utilities/NewTimer.cpp
```diff
@@ -101,19 +101,19 @@ void TimerType<CLOCK>::stop()
...
 
-      per_stack_total_time[current_stack_key] += elapsed;
+      per_stack_total_time[current_stack_key] += elapsed.count();
       per_stack_num_calls[current_stack_key] += 1;
 
       if (manager)
         manager->pop_timer(this);
     }
 #else
-    double elapsed = CLOCK()() - start_time;
```
ye-luo, line 115: I disliked the wired `()()`, the new function is more readable.

ye-luo (approved):
Still need non-ANL approval

src/Utilities/Clock.h
prckent: Indeed.

prckent:
Test this please

prckent (approved):

PR #4332, opened 2022-11-21
Use h5d_append to simplify ObservableHelper

quantumsteve:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

Describe what this PR changes and why.  If it closes an issue, link to it here
with [a supported keyword](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

I used h5d_append to remove the `value1_d` and `space1_d` state variables in ObservableHelper. Passing `hdf_archive` to `ObservableHelper::write` lets us remove `ObservableHelper::open`.

Variadic templates and initializer lists don't play nicely together, which is why `std::vector<std::string>{...} is added inside `std::make_unique(...)` and `std::vector<T>::emplace_back(...)`.   

Some writes still occur inside registerCollectables and registerObservables, so I didn't try removing `hdf_archive &` from the arguments list. 

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Code style update (formatting, renaming)
- Refactoring (no functional changes, no api changes)
- Testing changes (e.g. new unit/integration/performance tests)

### Does this introduce a breaking change?

- No 

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- Yes. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- Yes/No. Documentation has been added (if appropriate)

Issue #4322, opened 2022-11-09
change converter test error check


ye-luo:
**Describe the bug**
Right now the criteria is whether the standard error file was empty or not.
On Cray machine, the MPI runner may print additional things. I think when OpenMPI is enabled with GPU, additional printout also tricked the testing.
So I feel it is better to change convert4qmc to print a completion message like QMCPACK to signal a healthy run.
The pass criteria is to see that specific line and return error code 0 from the converter run.

**Expected behavior**
Make testing easy.

prckent:
Good point

PR #4338, opened 2022-11-23
Drop special case constructor from Observable Helper

PDoakORNL:
## Proposed changes
Following on from some experimenting to review #4332 which got merged prior to my review. OperatorHelper is looking better but it would be nice to remove some ugly.

Remove single group path term convenience constructor. Remove unnecessary and potentially problematic default constructor. Remove need for a bunch of explicit construction of temporary `std::vector<std::string>`

Stop using unique pointer vectors in a unnecessary way in `OperatorEstBase` simplify things further.

- Refactoring (no functional changes, no api changes)  Well at least not much.
- Documentation changes

- No

Leconte

## Checklist

### DIFF
--- a/src/QMCHamiltonians/MomentumEstimator.cpp
+++ b/src/QMCHamiltonians/MomentumEstimator.cpp
@@ -14,6 +14,10 @@
 
 
 #include "MomentumEstimator.h"
+
+#include <set>
+#include <string>
+
 #include "QMCWaveFunctions/TrialWaveFunction.h"
 #include "CPU/e2iphi.h"
 #include "CPU/BLAS.hpp"
@@ -21,7 +25,6 @@
 #include "Utilities/SimpleParser.h"
 #include "Particle/DistanceTable.h"
 #include "Numerics/DeterminantOperators.h"
-#include <set>
 
 namespace qmcplusplus
 {
@@ -101,8 +104,8 @@ void MomentumEstimator::registerCollectables(std::vector<ObservableHelper>& h5de
     //descriptor for the data, 1-D data
     std::vector<int> ng(1);
     //add nofk
-    ng[0] = nofK.size();
-    h5desc.push_back({{"nofk"}});
+    using namespace std::string_literals;
+    h5desc.push_back({{"nofk"s}});
     auto& h5o = h5desc.back();
     h5o.set_dimensions(ng, my_index_);
     h5o.addProperty(const_cast<std::vector<PosType>&>(kPoints), "kpoints", file);
--- a/src/QMCHamiltonians/MomentumEstimator.cpp
+++ b/src/QMCHamiltonians/MomentumEstimator.cpp
@@ -102,7 +102,7 @@ void MomentumEstimator::registerCollectables(std::vector<ObservableHelper>& h5de
     std::vector<int> ng(1);
     //add nofk
     ng[0] = nofK.size();
-    h5desc.emplace_back("nofk");
+    h5desc.push_back({{"nofk"}});
     auto& h5o = h5desc.back();
     h5o.set_dimensions(ng, my_index_);
     h5o.addProperty(const_cast<std::vector<PosType>&>(kPoints), "kpoints", file);
--- a/src/QMCHamiltonians/MomentumEstimator.cpp
+++ b/src/QMCHamiltonians/MomentumEstimator.cpp
@@ -104,6 +104,7 @@ void MomentumEstimator::registerCollectables(std::vector<ObservableHelper>& h5de
     //descriptor for the data, 1-D data
     std::vector<int> ng(1);
     //add nofk
+    ng[0] = nofK.size();
     using namespace std::string_literals;
     h5desc.push_back({{"nofk"s}});
     auto& h5o = h5desc.back();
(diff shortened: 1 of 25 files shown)


src/QMCHamiltonians/StaticStructureFactor.cpp
```diff
@@ -106,7 +106,7 @@ void StaticStructureFactor::addObservables(PropertySetType& plist, BufferType& c
 
 void StaticStructureFactor::registerCollectables(std::vector<ObservableHelper>& h5desc, hdf_archive& file) const
 {
-  h5desc.emplace_back(std::vector<std::string>{name_, "kpoints"});
+  h5desc.push_back({{name_, "kpoints"}});
```
markdewing, line 109: Is there a reason for when to use a plain character string vs. when to use the string literal 's' operator? (as in SkAllEstimator.cpp)

PDoakORNL: Whether its required or not is a question of when the complier can successfully infer it or not.

[Cpp core guidelines has a bit to say]
(http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slstr12-use-the-s-suffix-for-string-literals-meant-to-be-standard-library-strings)

Which makes we think we should use it here and consistently which was an element of #4332.

There are a bunch of type resolution implications as well.  These are generally invisible since std::string has  a const char * constructor.

src/QMCHamiltonians/PairCorrEstimator.cpp
```diff
@@ -164,7 +164,7 @@ void PairCorrEstimator::registerCollectables(std::vector<ObservableHelper>& h5li
   int offset = my_index_;
   for (int i = 0; i < gof_r_prefix.size(); ++i)
   {
-    h5list.emplace_back(gof_r_prefix[i]);
+    h5list.push_back({{name_, gof_r_prefix[i]}});
```
ye-luo, line 167: Suspicious change. Why two element now?

src/QMCHamiltonians/ObservableHelper.h
```diff
@@ -35,10 +35,11 @@ using value_type = QMCTraits::FullPrecRealType;
 class ObservableHelper
 {
 public:
+  ObservableHelper() = delete;
```
ye-luo, line 38: This is not really necessary. Once there is an explicit constructor. The one with empty argument should be disabled by default.

src/QMCHamiltonians/ObservableHelper.h
```diff
@@ -35,10 +35,11 @@ using value_type = QMCTraits::FullPrecRealType;
...
   /**
-   * default constructor
+   * Favored constructor
+   * \param[in] title         is the ordered hdf5 group path elements of the observable
    */
-  ObservableHelper(const std::string& title = "dummy");
```
ye-luo, line 41: I actually prefer getting ride of the constructor with `std::vector<std::string>`, I rarely saw the need of vector and saw more cumbersome caused by vector.

can we just have
```
ObservableHelper(const std::string& title);
```
Then we can use `emplace_back("abc")` instead of `push_back({{"abc"}})`

src/QMCHamiltonians/ObservableHelper.h
PDoakORNL: It's true I put this in to make sure there wasn't a hidden use of the default constructor somewhere.  I will remove it.

src/QMCHamiltonians/PairCorrEstimator.cpp
PDoakORNL: I started to change this so it wouldn't just dump its variables in the the global group.  But we shouldn't change it in this PR.

src/QMCHamiltonians/ObservableHelper.h
PDoakORNL: The point of #4322 was to remove that state games that made the single string constructor useful.  Basically the library was keeping that group path due to the sequence of calls from the different observables.  Most of the ported estimators use a multi element path.  Several hamiltonian operators just dump every hdf5 dataset into the global group with some gloppy name and should be using the `{name_, "variable_name"}`

src/QMCHamiltonians/MomentumEstimator.cpp
```diff
@@ -101,8 +104,8 @@ void MomentumEstimator::registerCollectables(std::vector<ObservableHelper>& h5de
     //descriptor for the data, 1-D data
     std::vector<int> ng(1);
     //add nofk
-    ng[0] = nofK.size();
```
ye-luo, line 104: Is this thre reason of CI failure?

PDoakORNL: Thanks for the catch @ye-luo .  I got a little happy with ctrl-k

ye-luo (approved):
LGTM @markdewing are you satisfied with the change?

markdewing (approved):

markdewing:
Test this please

ye-luo:
Test this please

quantumsteve:
@PDoakORNL Thank you for further improving `ObservableHelper`.

PR #4339, opened 2022-11-23
Nexus: enable reading Jastrow factors from existing files

jtkrogel:
## Proposed changes

This PR enables reading Jastrow factors from existing files and including them into new QMCPACK input files generated by Nexus.  The pre-existing capability to specify an existing ESHDF5 file for the orbitals has also been hardened to include file existence checks by default.  

- New feature

- No

Laptop

## Checklist

### DIFF
--- a/nexus/lib/qmcpack_input.py
+++ b/nexus/lib/qmcpack_input.py
@@ -1940,9 +1940,9 @@ class radfunc(QIxml):
 #end class radfunc
 
 class slaterdeterminant(QIxml):
-    attributes = ['optimize','delay_rank']
+    attributes = ['optimize','delay_rank','gpu','matrix_inverter']
     elements   = ['determinant']
-    write_types = obj(optimize=yesno)
+    write_types = obj(optimize=yesno,gpu=yesno)
 #end class slaterdeterminant
 
 class determinant(QIxml):
@@ -4596,7 +4596,8 @@ def generate_bspline_builder(type           = 'bspline',
                              spo_up         = 'spo_u',
                              spo_down       = 'spo_d',
                              sposets        = None,
-                             system         = None
+                             system         = None,
+                             orbitals_cpu   = None,
                              ):
     tilematrix = identity(3,dtype=int)
     if system!=None:
@@ -4620,7 +4621,7 @@ def generate_bspline_builder(type           = 'bspline',
             spindatasets   = True
             )
         )
-    if sort!=None:
+    if sort is not None:
         bsb.sort = sort
     #end if
     if truncate and buffer!=None:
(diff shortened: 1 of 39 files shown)


nexus/tests/unit/test_simulation_module.py
```diff
@@ -2387,7 +2387,10 @@ def test_execute():
     assert(os.path.exists(outfile))
     assert(os.path.exists(errfile))
     assert(open(outfile,'r').read().strip()=='run')
-    assert(open(errfile,'r').read().strip()=='')
+    err_contents = open(errfile,'r').read().strip()
+    # workaround for openmpi issue on some systems
```
markdewing, line 2391: Is there more context about this issue, or a link to some more discussion?
It appears to be a very specific problem that apparently doesn't affect actual MPI execution?

prckent: Agreed. It would be good to have a note of which MPI and when this occurs, example good and bad output. Since this is a brittle fix and others like it could occur, perhaps move to a dedicated file/function of workarounds? Check for unexpected output and print a warning?

jtkrogel: This is the second time I've seen this behavior over the last 3 years.  The string in question is written to stderr whenever mpirun is executed and does not affect the correctness of what is run.  It is machine configuration dependent and appears to be related to OpenMPI.  This is essentially a quality of life fix that allows me to continue development unaffected by this apparently benign behavior.  In any event, the issuance of this string does not constitute an error in Nexus and so it is correct to guard against it.

ye-luo: Whatever MPI library decide to print is out of our control. See https://github.com/QMCPACK/qmcpack/issues/4322. If I remember correctly, it is related to OpenMPI built with GPU support but there is no GPU on the compute node.

markdewing: @jtkrogel Thanks for the description - this is the kind of context I was requesting.

Searching this issue, it appears many people have seen this message with MPI, but there doesn't appear to be any definitive explanation of why it happens.   The MIT-MAGIC-COOKIE message is associated with X11 authentication, but several reports indicate it happens with no X11 visibly involved.

Could the comment be changed to something more precise, like "Handle spurious error message from OpenMPI" (and optionally link to these comments on the PR).

jtkrogel: Done

jtkrogel:
Should be ready to merge now.

prckent (approved):

markdewing (approved):

PR #4299, opened 2022-10-29
Per particle logger

PDoakORNL:
## Proposed changes

This PR finishes connecting per particle Hamiltonian reporting to estimators through the entire application to the per particle Hamiltonian logger.

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- New feature
- Build related changes
- Testing changes (e.g. new unit/integration/performance tests)
- Documentation changes

### Does this introduce a breaking change?

- No

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes This PR is up to date with current the current state of 'develop'
- Yes/No. Code added or changed in the PR has been clang-formatted
- Yes This PR adds tests to cover any new code, or to catch a bug that is being fixed
- Yes/No. Documentation has been added (if appropriate)

PR #4340, opened 2022-11-23
Extending modern string utils to include split and strip

PDoakORNL:
## Proposed changes

Adds a strip and split function that support `string-view`
Existing code does not use this but it is required for #4299 

- New feature
- Documentation changes

- No

Leconte

## Checklist

### DIFF
--- a/src/Utilities/ModernStringUtils.hpp
+++ b/src/Utilities/ModernStringUtils.hpp
@@ -12,8 +12,12 @@
 #ifndef QMCPLUSPLUS_MODERNSTRINGUTILS_HPP
 #define QMCPLUSPLUS_MODERNSTRINGUTILS_HPP
 
+#include <charconv>
+#include <stdexcept>
 #include <string>
 #include <string_view>
+#include <type_traits>
+#include <vector>
 
 namespace qmcplusplus
 {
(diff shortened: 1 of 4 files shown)


prckent:
Test this please

src/Utilities/tests/test_ModernStringUtils.cpp
```diff
@@ -31,4 +31,61 @@ TEST_CASE("ModernStringUtils_strToLower", "[utilities]")
         "\xab"
         "not_just_ascii");
 }
+
+TEST_CASE("ModernStringUtils_split", "[utilities]")
```
markdewing, line 35: There should be some tests for edge cases - empty string, string with no delimiters, string with empty delimiter, string that is only delimiters.

src/Utilities/tests/test_ModernStringUtils.cpp
```diff
@@ -31,4 +31,61 @@ TEST_CASE("ModernStringUtils_strToLower", "[utilities]")
...
+}
+
+TEST_CASE("ModernStringUtils_string2Real", "[utilities]")
+{
+  std::string_view svalue{"101.52326626"};
+  double value = string2Real<double>(svalue);
+  CHECK(value == Approx(101.52326626));
 } // namespace qmcplusplus
+
+TEST_CASE("ModernStringUtils_strip", "[utilities]")
```
markdewing, line 68: This should also have some more edge case tests - empty string, string that is only whitespace.

PDoakORNL: Added

src/Utilities/for_testing/NativeInitializerPrint.hpp
```diff
@@ -39,4 +39,16 @@ std::ostream& operator<<(std::ostream& out, const NativePrint<std::vector<T>>& n
   return out;
 }
 
+template<class T>
+std::ostream& operator<<(std::ostream& out, const NativePrint<Vector<T>>& np_vec)
```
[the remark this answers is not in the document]
PDoakORNL: Added file scope documentation.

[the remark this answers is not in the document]
ye-luo: Thanks

src/Utilities/ModernStringUtils.hpp
```diff
@@ -30,6 +34,36 @@ namespace qmcplusplus
...
+ *  calls c++ string to real conversion based on T's precision template<typename T>
+ */
+template<typename T>
+inline T string2Real(const std::string_view svalue) {
+  static_assert(std::is_floating_point_v<T>);
+      T result;
+// full support for floating point from char not present until stdlibc++ aligned with gcc11
+// there is still not from_char for floats as of libc++ 15
+#if _GLIBCXX_RELEASE > 10
+  auto [prt, ec] = std::from_chars(svalue.data(), svalue.data() + svalue.size(), result, std::chars_format::general);
```
ye-luo, line 56: What is the benefit of this special case? Faster? I don't mind keep only the universal working way although a bit slow.

PDoakORNL: It's c++ 17 and the point of this file is to use the best practice for the current standard requirement of the code.  But here we have to make an except due to incomplete support for the std library.

The new way does not require a mutable string to be constructed and in general fits with the string_view and current c++ io framework the other is a c function.

ye-luo: C++17 language yes. C++17 standard library, if it is a hassle due to implementation issue, I prefer using a dumb method. Is your new way safer because of error handling? That could shift my preference.

PDoakORNL: Yes it is able to clearly indicate an error.  atof really can't do anything you get 0.0 or UB.  

So with the c++17 way if it can't do the conversion or its out of range for the floating point type we can throw.  This means we can deal with bad input better.

markdewing (approved):

ye-luo:
Test this please

PR #4349, opened 2022-11-30
Fix subset parameter optimization in legacy drivers

ye-luo:
## Proposed changes
The bug affects runs with threads. Need to bring meta data up-to-date.
This is more of a stopgap. I tend to change all the myVars in to shared pointers between clones and we only operates through the gold copy.

- Bugfix

- No

epyc-server

## Checklist


prckent (approved):
LGTM

When more work is done, I think we should add some testing of these features. We have established that getting options handled consistently is tricky and need to ensure that we don't make the same mistakes again. Maybe @Hyeondeok-Shin check with @ye-luo ?

PR #4351, opened 2022-12-01
[WIP] Next orbital rotation prototype

markdewing:
Supersedes  #4202 

Changes from that version:
- For splines, the coefficients can be saved in order to enable the global rotation method.  The multiplication was changed slightly to copy the old spline coefficients to the temporary storage first, then store the results of the multiplication in the actual spline storage.
- The coefficient copy uses a shared pointer to reduce storage.  The LCAO implementation was changed.  This reduces the extra storage needed.  Only one thread should apply the rotation
- Hack to specify the rotation method in the input file.  Default is history list.  Use 'optimize="global"' to use global rotation method.

Updates:
 - Works for the single-determinant case.  Implemented `evaluteDerivativesWF` in RotatedSPOs and a path to calling it.
 - Works with batched algorithm for NLPP in the single-determinant case.  Implemented `evaluateDerivRatios` in RotatedSPOs and a path to calling it, along with extra arguments.

- New feature

- No

## Checklist


jptowns:
This may not be the right place for this, but want to get out in the open so that we remember to circle back at some point. During optimization (oneshift at least but possibly for other methods), a rotation with a proposed set of wfn parameters is applied before deciding to accept/reject the proposed parameters. This is done, e.g. in line 1365 in QMCFixedSampleLinearOptimize.cpp in this line: `const RealType newCost = optTarget->Cost(false);`. If the parameters are accepted, it's fine. But if th eparameters are rejected, we need to apply the old rotation, otherwise we're stuck with the (now rejected) parameters. I thikn that the simplest possible fix would be to simply add another optTarget->Cost() call after rejection. But we should probably study the optimizer(s) more closely to see where else this may crop up. 

Otherwise, this PR looks good to me.

prckent:
Q. How much of this remains to be cleaved off and merged? Can we close this now? What is needed to run (say) FeO in splines at gamma in the batched code?

markdewing:
The key changes to enable orbital rotation are the functions checkInVariablesExclusive and resetParametersExclusive in RotatedSPOs.
What remains
* #4468 
* The bspline version of that PR
* The changes to checkInVariablesExclusive and resetParametersExclusive
That should be enough to make the optimization work.

What's left  after that is the saving and restoring of the rotation parameters

PR #4354, opened 2022-12-07
Nexus: support Polaris at ALCF

jtkrogel:
## Proposed changes

This PR adds support for Polaris in Nexus.  This includes support for job bundling (which on Polaris requires parsing/splitting/distributing nodelist files).

- New feature

- No

Workstation, Polaris

## Checklist


nexus/lib/machines.py
```diff
@@ -3371,6 +3379,69 @@ def write_job_header(self,job):
...
+            job.run_options.add(**opt)
+        #end if
+    #end def post_process_job
+
+    def write_job_header(self,job):
+        if job.queue is None:
+            job.queue = 'prod'
+        #end if
+        if job.filesystems is None:
+            job.filesystems = 'home:eagle'
```
markdewing, line 3405: Should the default add 'grand' as well?

prckent (commented):
What is the quirk that requires specialize_bundle_commands?

nexus/lib/machines.py
jtkrogel: Sure.  Will add.

jtkrogel: Done.

prckent:
ntest_nexus_machines is failing in CI after grand was added. It was fine before.

prckent (approved):

PR #4355, opened 2022-12-07
Improve determinant list signed integer handling

### LINKED ISSUE #4353: Multidet with the highest orbital in a single int occupied
**Describe the bug**
If the highest orbital is occupied in a single bitstring (e.g. orbital 64) then we end up overflowing the signed int + throwing away the sign?
This results in a different occupation string than expected.

### Logging in `SlaterDetBuilder.cpp`
When logging near line 1190 in `SlaterDetBuilder.cpp`, for a determinant that appears with an int `9223372036854775809` in the hdf5 the absolute value of the signed int appears:
```
grp   1    K 0   a 9223372036854775807         0111111111111111111111111111111111111111111111111111111111111111
```

### Toy cpp example with the same behavior (https://gist.github.com/shivupa/199da68de90eb634960f0a7d93584bac):
```
Highest occ orbtial bug!
 bitkind 8
 std::pow(2,7) + 1 = 129
 uint8_t 129 bitset<bitkind> 10000001
 int8_t -127 bitset<bitkind> 10000001
 uint_t from int_t 127 bitset<bitkind> 01111111

 bitkind 64
 std::pow(2,63) + 1 = 9223372036854775809
 uint64_t 9223372036854775809 bitset<bitkind> 1000000000000000000000000000000000000000000000000000000000000001
 int64_t -9223372036854775807 bitset<bitkind> 1000000000000000000000000000000000000000000000000000000000000001
 uint_t from int_t 9223372036854775807 bitset<bitkind> 0111111111111111111111111111111111111111111111111111111111111111
```

I believe we should use `uint64_t` instead of `int64_t` in `src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp`

I have a branch that just switches the `int64_t` to `uint64_t` [here](https://github.com/shivupa/qmcpack/tree/highest_occ_orb_bug).

**To Reproduce**
1. release version or git commit hash being built 
current dev

I am not sure exactly what a test should entail. Should we do H2 FCI (w/ aug-cc-pvqz to have enough orbitals) in QP2 to trigger this?

**Expected behavior**
The assert for the correct number of particles in a bitstring should never be hit.

**System:**
 - Laptop (i7-7700HQ)

**The behavior is consistent with what I am describing, but I am skeptical that this hasn't been hit by other users especially CIPSI users.**

shivupa:
## Proposed changes
This makes the requirements for signed integers in the multidet.h5 more explicit. Also fixes a typo in a test.
Closes #4353

- Documentation changes

- No

## Checklist


prckent (dismissed):

shivupa:
Won't that affect current users who already have hdf5s generated with the signed ints? Or is that very quick to regenerate?

prckent:
@shivupa Lets just make these all unsigned. It is the most logical solution and what should have been done in the first place. What changed my mind is your question on the userbase -- realistically the number of users of multidet wavefunctions and this route is low and only expert users. So they can regenerate if they are hit by this bug. If someone is really hurting we or they could write a converter.

shivupa:
Sounds good I'll update this shortly

prckent (changes requested):
Setting "changes requested" status to reflect discussions

shivupa:
OK unsigned ints are now used. Also `determinant_tools.py` can now transform old hdf5 files to unsigned integers, and I used this to change the test inputs to the unsigned format.

docs/intro_wavefunction.rst
```diff
@@ -873,6 +873,9 @@ Additional information:
 - ``algorithm`` algorithms used in multi-Slater determinant implementation. ``table_method`` table method of Clark et al. :cite:`Clark2011` .
   ``precomputed_table_method`` adds partial sum precomputation on top of ``table_method``.
 
+- When the determinants are read from an HDF5 file in the ``detlist`` child, the hdf5 representation of the multideterminant wavefunction must use unsigned integers to represent the determinants. The ``utils/determinants_tools.py`` script described in further detail below will check that the determinants are stored using the correct type.
```
kgasperich, line 876: do the ints need to be 64-bit unsigned, or is the converter flexible enough to handle other sizes?

utils/determinants_tools.py
```diff
@@ -153,13 +156,37 @@ def csf_to_string(csf, bit_kind_size, size):
...
+        bit_kind_size = f[alpha_group].dtype.itemsize * 8
+        print(f[alpha_group].dtype)
 
         # We sanitize the determinants. Aka we transform any negative integers, into the proper unsigned one.
-        i_alpha= (sanitize(d,bit_kind_size) for d in f['MultiDet/CI_Alpha'])
-        i_beta= (sanitize(d,bit_kind_size) for d in f['MultiDet/CI_Beta'])
+        i_alpha= (sanitize(d,bit_kind_size) for d in f[alpha_group])
+        i_beta= (sanitize(d,bit_kind_size) for d in f[beta_group])
+        if args.transform:
+            alpha_dets = np.array([i for i in i_alpha], dtype=np.uint)
```
kgasperich, line 177: Should this be [np.uint64](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.uint64)? (I think on many/most platforms uint and uint64 will be the same, but if we always want uint64 then it might save some future headaches if that's specified here)

utils/determinants_tools.py
```diff
@@ -153,13 +156,37 @@ def csf_to_string(csf, bit_kind_size, size):
...
+            f.create_dataset(alpha_group, data=alpha_dets)
+            beta_dets = np.array([i for i in i_beta], dtype=np.uint)
+            del f[beta_group]
+            f.create_dataset(beta_group, data=beta_dets)
+            # regenerate iterators
+            i_alpha= (sanitize(d,bit_kind_size) for d in f[alpha_group])
+            i_beta= (sanitize(d,bit_kind_size) for d in f[beta_group])
+
+        if f[alpha_group].dtype == 'int64':
+            raise TypeError("This HDF5 file uses signed 64 bit integers (int64) to represent the determinants. QMCPACK expects unsigned integers (uint64) this can cause issues when used in qmcpack.")
```
kgasperich, line 188: If we're not going to apply the signed->unsigned transformation here, then we should at least include the `--transform` flag in the error message.
I'd actually lean toward just removing the `--transform` option and doing the conversion if signed ints are detected.

kgasperich (changes requested):

utils/determinants_tools.py
shivupa: Fixed!

utils/determinants_tools.py
shivupa: I actually like that I removed the transform option and always transform if we find signed ints

docs/intro_wavefunction.rst
shivupa: I am not sure about other paths into the code (xml), but at least through HDF5 you need 64 bit.  I've fixed the docs as well

prckent:
Test this please

utils/determinants_tools.py
```diff
@@ -153,13 +153,33 @@ def csf_to_string(csf, bit_kind_size, size):
...
-        bit_kind_size = f['MultiDet/CI_Alpha'].dtype.itemsize * 8
+        bit_kind_size = f[alpha_group].dtype.itemsize * 8
+        print(f[alpha_group].dtype)
 
         # We sanitize the determinants. Aka we transform any negative integers, into the proper unsigned one.
-        i_alpha= (sanitize(d,bit_kind_size) for d in f['MultiDet/CI_Alpha'])
-        i_beta= (sanitize(d,bit_kind_size) for d in f['MultiDet/CI_Beta'])
+        i_alpha= (sanitize(d,bit_kind_size) for d in f[alpha_group])
+        i_beta= (sanitize(d,bit_kind_size) for d in f[beta_group])
+        if f[alpha_group].dtype == 'int64':
```
kgasperich, line 171: sorry, in my earlier review I was thinking that this was part of the required QP2->QMCPACK workflow, and that was why I thought we should do the conversion automatically.

I think that if we don't want to handle dets as signed ints in SlaterDetBuilder, then we should provide a tool to do the conversion, but maybe your first way was better (add the `--transform` option to do the transformation, and in SlaterDetBuilder throw an error if the hdf5 uses signed ints (and tell the user how to do the conversion using the python script)).

shivupa:
touched base with kevin need a bit to update this

utils/determinants_tools.py
```diff
@@ -153,13 +153,33 @@ def csf_to_string(csf, bit_kind_size, size):
...
+        alpha_group = 'MultiDet/CI_Alpha'
+        if alpha_group not in f:
+            alpha_group = 'MultiDet/CI_0'
+        beta_group = 'MultiDet/CI_Beta'
+        if beta_group not in f:
+            beta_group = 'MultiDet/CI_1'
 
-        bit_kind_size = f['MultiDet/CI_Alpha'].dtype.itemsize * 8
+        bit_kind_size = f[alpha_group].dtype.itemsize * 8
+        print(f[alpha_group].dtype)
```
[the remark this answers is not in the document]
shivupa: Yeah sorry removed!

utils/determinants_tools.py
shivupa: Reverted this. I ended up figuring out how to check the type of a dataset in an hdf5 file.

shivupa:
> Aborting at QMCPACK expects the HDF5 CI vectors to be stored as unsigned 64 bit integers. This HDF5 uses signed 64 bit integers. The determinants_tools.py script can transform this file using the 'transform' flag.
Nice it works but I forgot to update an hdf5 file. Updated now.

shivupa:
This touches hdf5 parts of the code. Can someone with hdf5 knowledge take a look maybe @quantumsteve ?

kgasperich (dismissed):

src/io/hdf/hdf_archive.h
```diff
@@ -146,6 +146,21 @@ class hdf_archive
    */
   bool is_group(const std::string& aname);
 
+  /** check if aname is a dataset of type T
+   * @param aname group's name
+   * @return true, if aname is a dataset of type T
+   */
+  template<typename T>
+  bool is_dataset_of_type(const std::string& aname)
```
ye-luo, line 154: Need a unit test to cover both true and false case.

ye-luo (changes requested):
Request a unit test.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -1138,23 +1138,43 @@ bool SlaterDetBuilder::readDetListH5(xmlNodePtr cur,
...
-        hin.read(temps[grp], "CI_Beta");
-      else
-        APP_ABORT("Unknown HDF5 CI format");
+      {
+        ds_tag          = "CI_Beta";
+        is_correct_type = hin.is_dataset_of_type<uint64_t>(ds_tag);
+      }
     }
+
+    if (!is_correct_type)
```
quantumsteve, line 1166: Why repeat the same check on line 1151?

shivupa: The first check just checks that the numbered "CI_X" tag wasn't found (e.g. "CI_0"). For historical reasons we have to check if "CI_Alpha/Beta" exists as well. So if `!correct_type` happens the first time, we still have to check again.

src/io/hdf/hdf_wrapper_functions.h
```diff
@@ -133,6 +133,25 @@ inline bool h5d_read(hid_t grp, const std::string& aname, T* first, hid_t xfer_p
   return ret != -1;
 }
 
+template<typename T>
+inline bool h5d_check_type(hid_t grp, const std::string& aname)
+{
+  if (grp < 0)
+    return true;
+  hid_t h1 = H5Dopen(grp, aname.c_str(), H5P_DEFAULT);
```
quantumsteve, line 141: Where do you call `H5Dclose( h1 );`?

src/io/hdf/hdf_wrapper_functions.h
```diff
@@ -133,6 +133,25 @@ inline bool h5d_read(hid_t grp, const std::string& aname, T* first, hid_t xfer_p
...
+  if (h1 < 0)
+    return false;
+  T temp;
+  hid_t h5d_type_id     = get_h5_datatype(temp);
+  hid_t datatype        = H5Dget_type(h1);
```
quantumsteve, line 146: Should we check that `H5Dget_type` doesn't return `H5I_INVALID_HID`?
https://docs.hdfgroup.org/hdf5/develop/group___h5_d.html#title25

src/io/hdf/hdf_wrapper_functions.h
```diff
@@ -133,6 +133,25 @@ inline bool h5d_read(hid_t grp, const std::string& aname, T* first, hid_t xfer_p
...
+  htri_t equality_check = H5Tequal(datatype, h5d_type_id);
```
quantumsteve, line 147: Should we differentiate between `equality_check == 0` (false) and `equality_check < 0` (failure)
https://docs.hdfgroup.org/hdf5/develop/group___h5_t.html#gaa92250f289b557b63cba974defa20b0f

src/io/hdf/hdf_wrapper_functions.h
shivupa: Thanks I forgot this

src/io/hdf/hdf_wrapper_functions.h
shivupa: I believe the above return if the dataset tag doesn't exist prevent us from encountering `H5I_INVALID_HID`. However I like assuming that the tag should exist + throwing an error if something goes wrong along the way since this is more explicit. Fixed

src/io/hdf/hdf_wrapper_functions.h
shivupa: Same with the above comment. I think now we can hit an error only if the equality check is done with an `H5I_INVALID_HID` type for the first arg. However I also decided to throw here. The test has been updated with a "REQUIRE_THROWS_AS" test case.

src/io/hdf/hdf_wrapper_functions.h
ye-luo: Fixed?

ye-luo (commented):
What is the current status?

src/io/hdf/hdf_wrapper_functions.h
shivupa: done

prckent (approved):
LGTM

All comments appear to have been addressed.

prckent:
Test this please

PR #4357, opened 2022-12-07
Input section updates

PDoakORNL:
Changes to InputSection that are needed for PerParticleLogger or EnergyDensityEstimator.  Addition documentation will be in another PR but I prefer to wait until real examples of the use and need for these InputSection expansions are needed.

## Proposed changes

Add ability for child to provide a custom stream parser for a particular tag.
Add ability to delegate parsing of a tag to another input class/section.

- New feature
- Build related changes
- Documentation changes

- No

Leconte

## Checklist

- Yes Some Documentation has been added (if appropriate)


ye-luo:
> I prefer to wait until real examples of the use and need for these InputSection expansions are needed.

When will the real example appear? Did you mean `PerParticleLogger or EnergyDensityEstimator`?
I was a bit confused. If there is no use in the near term. I'm wondering why this feature is added in the first place.
Perhaps I misunderstood "real example".

src/Estimators/InputSection.h
```diff
@@ -173,8 +197,8 @@ class InputSection
 private:
   // Query functions
   bool isAttribute(const std::string& name) const { return attributes.find(name) != attributes.end(); }
-  bool isParameter(const std::string& name) const { return parameters.find(name) != parameters.end(); }
   bool isDelegate(const std::string& name) const { return delegates.find(name) != delegates.end(); }
```
ye-luo, line 200: I don't know what is Delegate. So upcoming documentation needs to explain this and give examples.

src/Estimators/OperatorEstBase.h
```diff
@@ -133,7 +120,21 @@ class OperatorEstBase
 
   bool isListenerRequired() { return requires_listener_; }
 
+  DataLocality get_data_locality() { return data_locality_; }
```
ye-luo, line 123: `get_data_locality() const`

src/Estimators/InputSection.h
PDoakORNL: This is for sure.  Without example its hard to explain.  But bringing the concrete example in results in too much.  This could be another instance where the name is too general and we may want to return to the naming after the docs are expanded.

ye-luo:
Test this please

Issue #4362, opened 2022-12-12
randomize_grid does not randomize the PP grid as expected


markdewing:
I would expect the function `NonLocalECPComponent::randomize_grid` to randomize the grid origin across the entire sphere. 
But it does not - it's missing some coverage around the "poles".   See the attached graph.

The root cause is the initialization of `cth` from a random number.  It is currently `cth = myRNG() - 0.5`.  I think it should be `cth = 1.0 - 2*myRNG()`.

I'm not sure this issue has a practical impact, since the rotation is only meant to change the origin of the spherical integration grid and not cover the sphere by itself.   If the grid is sufficiently dense, the other points will cover the missing regions from the origin. (But I didn't actually check this)

There is another minor annoyance in that when initializing the current code with all zeros for the rng values, the result is not the identity rotation.  With the corrected initialization, using zeros does result in the identity rotation. 

I tried to find the origin of the formula in `randomize_grid`.  Nothing online seemed to match, and I couldn't recreate it through multiplying 2x2 rotation matrices (though that search was not exhaustive).

Plot of random rotations applied an input vector (0,0, 1)
![random_rotation](https://user-images.githubusercontent.com/266185/207135506-83fc5607-e674-4aa4-8532-cb310446f91b.png)

The script used to generate the plot

[check_grid_randomize.tar.gz](https://github.com/QMCPACK/qmcpack/files/10211349/check_grid_randomize.tar.gz)

prckent:
Thanks Mark. Indeed this seems like a bug, since we shouldn't rely on grid density or symmetry to get the correct result. It should be correct even if we use one point, which it will not be based on your plot.

The functions in question are at https://github.com/QMCPACK/qmcpack/blob/develop/src/QMCHamiltonians/NonLocalECPComponent.cpp#L868 (we have two variants)

It is also interesting that we use 3 random numbers for this.

jtkrogel:
This is how the code looked when randomize_grid was introduced in Mar. 2005 (commit: a9b5aadbe8b37609c7f65222df8d1daec5f97f93, file NonLocalPPotential.cpp):

```
  void NonLocalPPotential::RadialPotentialSet::randomize_grid(){
    const double twopi=6.28318531;
    double phi,psi,sph,cph,sth,cth,sps,cps;
    Tensor<double,3> rmat;
    /* Rotation matrix with Euler angles defined as: 
       1) counterclockwise rotation around z (phi)
       2) clockwise rotation around y (theta)
[...]

PR #4363, opened 2022-12-13
Refactor grid randomization

markdewing:
Fixes the issue raised in #4362.  There will be a lot of CI failures.  Not sure the best way to approach the fix and updating the deterministic tests.  This is a start to at least introduce a possible fix.

Does some refactoring to put the inputs to the random rotation matrix into their own class.  The generation of the rotation matrix is extracted from the randomize_grid function, to make it testable in isolation.

A single unit test is added to the rotation matrix - will need more in the future. 

- Bugfix

- Yes

desktop

## Checklist


src/QMCHamiltonians/RandomRotationState.h
```diff
@@ -0,0 +1,62 @@
...
+{
+
+// The angular values for a random rotation matrix
+// phi : 0 to 2*pi
+// psi : 0 to 2*pi
+// cth : -1 to 1  (cth = cos(theta))
+class RandomRotationState
+{
+public:
+  using RealType   = QMCTraits::RealType;
```
PDoakORNL, line 30: Delete Type from name, the capitalization already indicates this is a type.

src/QMCHamiltonians/RandomRotationState.h
```diff
@@ -0,0 +1,62 @@
...
+  using TensorType = QMCTraits::TensorType;
+
+  RandomRotationState() : rng_vals_({0.0, 0.0, 0.0}) {}
+  RandomRotationState(RandomGenerator& rng) : rng_vals_({rng(), rng(), rng()}) {}
+
+  TensorType getRandomRotationMatrix() const
```
PDoakORNL, line 36: I think this should just be a function that takes an RNG and returns a random rotation matrix.

src/QMCHamiltonians/NonLocalECPComponent.cpp
```diff
@@ -866,13 +865,9 @@ void NonLocalECPComponent::evaluateOneBodyOpMatrixdRContribution(ParticleSet& W,
 }
 
 ///Randomly rotate sgrid_m
-void NonLocalECPComponent::randomize_grid(RandomGenerator& myRNG)
+void NonLocalECPComponent::randomize_grid(const RandomRotationState& rs)
```
camelto2, line 868: Note that there is also NonLocalECPComponent::randomize_grid(std::vector<T>& sphere, RandomGenerator& myRNG). From what I can tell, this is only used by the legacy NonLocalECPotential_CUDA.cpp. For consistency, should that randomize_grid call be changed as well since the legacy code hasn't been removed yet?

That probably breaks more tests and since we are dropping eventually dropping the legacy, maybe it isn't worth changing. But it seems to me things should be done consistently for whatever is actively in the code base.

prckent: Indeed, we will likely want to fix this everywhere.

The code used by the legacy CUDA code paths will also need to be updated.

There will have to be some discussion around updating the tests. This will be a significant effort even for the smallest possible update and likely need parallelizing over multiple people.

prckent:
I experimented with an ugly bash-awk-sed combination to update the test values for all CPU build variants as well as legacy CUDA. It looks like nearly all can be automated this way. Estimators such as the density will have to be updated by hand. Should I pursue this or is anyone else close to a clean solution?

prckent:
Test this please

prckent:
I have a fully updated set of reference values in #4383 that passes CI. They could be merged independently of this refactor or pulled into this PR.

The conflict in this PR also needs fixing.

What are the preferred next steps? Who has bandwidth?

src/Numerics/RandomRotationMatrix.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include "Configuration.h"
+#include "Utilities/RandomGenerator.h"
+#include <array>
+
+namespace qmcplusplus
+{
+
+template<typename T>
+inline Tensor<T, 3> getRotationMatrix(T rng1, T rng2, T rng3)
```
ye-luo, line 24: Documentation needed. Prefer `generateRotationMatrix`

src/Numerics/RandomRotationMatrix.h
```diff
@@ -0,0 +1,55 @@
...
+                     -cph * cth * sps - sph * cps,
+                     -sph * cth * sps + cph * cps,
+                      sth * sps,
+                      cph * sth,
+                      sph * sth,
+                      cth);
+  // clang-format on
+}
+
+inline QMCTraits::TensorType getRandomRotationMatrix(RandomGenerator& rng)
```
ye-luo, line 49: This API needs RandomGenerator. It should be moved to QMCHamiltonian. Better to keep files under `src/Numerics` away from RNG classes.
It is also better to assign a more self-explanatory function name once being moved to a better location.

markdewing:
I think #4383 should be merged independently.

src/Numerics/RandomRotationMatrix.h
markdewing: I split up the file into RotationMatrix3D (in Numerics) and moved this function to QMCHamiltonians/RandomRotationMatrix.h

markdewing:
Test this please

src/QMCHamiltonians/NonLocalECPComponent.h
```diff
@@ -140,9 +141,9 @@ class NonLocalECPComponent : public QMCTraits
 
   void resize_warrays(int n, int m, int l);
 
-  void randomize_grid(RandomGenerator& myRNG);
+  void randomize_grid(const TensorType& rmat);
```
ye-luo, line 144: This function no more handles randomization. Hence a rename and also adopt naming convention. rotateQuadratureGrid?

src/QMCHamiltonians/NonLocalECPotential.cpp
```diff
@@ -552,7 +561,10 @@ void NonLocalECPotential::evaluateOneBodyOpMatrix(ParticleSet& P,
   for (int ipp = 0; ipp < PPset.size(); ipp++)
     if (PPset[ipp])
       if (!keepGrid)
-        PPset[ipp]->randomize_grid(*myRNG);
+      {
+        TensorType rmat = generateRandomRotationMatrix(*myRNG);
+        PPset[ipp]->randomize_grid(rmat);
```
ye-luo, line 566: Could you fuse them into one line and keep the code concise?

src/Numerics/RotationMatrix3D.h
```diff
@@ -0,0 +1,53 @@
...
+// File developed by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
+//
+// File created by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+#ifndef QMCPLUSPLUS_ROTATION_MATRIX_3D_H
+#define QMCPLUSPLUS_ROTATION_MATRIX_3D_H
+
+#include "Configuration.h"
```
ye-luo, line 16: Is this header file needed? How about incude Tensor header file.

src/QMCHamiltonians/SOECPotential.cpp
```diff
@@ -38,7 +38,10 @@ SOECPotential::Return_t SOECPotential::evaluate(ParticleSet& P)
   value_ = 0.0;
   for (int ipp = 0; ipp < PPset.size(); ipp++)
     if (PPset[ipp])
-      PPset[ipp]->randomize_grid(*myRNG);
+    {
+      TensorType rmat = generateRandomRotationMatrix(*myRNG);
+      PPset[ipp]->rotateQuadratureGrid(rmat);
+    }
```
ye-luo, line 44: fuse into one line and remove {}.

tests/solids/bccMo_2x1x1_SOREP/CMakeLists.txt
```diff
@@ -80,14 +80,14 @@ if(QMC_COMPLEX)
 #Deterministic test
 
   if(NOT QMC_MIXED_PRECISION)
-    list(APPEND DET_BCC_MO_VMC_SCALARS "totenergy" "-136.88596684 0.000002")
+    list(APPEND DET_BCC_MO_VMC_SCALARS "totenergy" "-136.88580235 0.000002")
```
ye-luo, line 83: I thought #4383 already applied a fix and updated reference values. I'm wondering why this file needs to be touched.

ye-luo (commented):
Is it time to remove the WIP tag?

tests/solids/bccMo_2x1x1_SOREP/CMakeLists.txt
markdewing: The fix in #4383 did not touch SOECPotential.cpp, and that reference value didn't get updated.

ye-luo: Clear. Thanks.

markdewing:
Test this please

ye-luo (dismissed):
LGTM. Still need non-ANL approval.

markdewing:
Test this please

prckent (changes requested):
The python code needs a description at the top saying what it is, what it was used for, where the output went, when it would need to be rerun etc.

prckent (approved):

prckent:
Test this pleas

prckent:
Test this please

PR #4364, opened 2022-12-14
[WIP] convert h5 orbs from spherical to Cartesian

kgasperich:
## Proposed changes

Added a utility to convert pyscf h5 orbitals from spherical to Cartesian AO basis.

I still need to check the normalization for the different shells, but all of the framework is in place.

The PyscfToQmcpack converter doesn't seem to correctly handle cases where different atoms of the same element have different basis sets, so I didn't account for that in this converter either (it will be simple to update this converter when PyscfToQmcpack is updated).

- New feature

- No

### DIFF
--- a/src/QMCTools/PyscfSphToCart.py
+++ b/src/QMCTools/PyscfSphToCart.py
@@ -1,0 +1,445 @@
+#!/usr/bin/env python3
+######################################################################################
+## This file is distributed under the University of Illinois/NCSA Open Source License.
+## See LICENSE file in top directory for details.
+##
+## Copyright (c) 2016 Jeongnim Kim and QMCPACK developers.
+##
+## File developed by: Kevin Gasperich, kgasperich@anl.gov, Argonne National Laboratory
+##
+## File created by:   Kevin Gasperich, kgasperich@anl.gov, Argonne National Laboratory
+#######################################################################################
+
+import h5py
+import numpy as np
+import itertools
+import shutil
+
+def gms_cart_order(l):
+    def sort1(s):
+        # return single xyz string in gamess order
+        return ''.join(sorted((i*s.count(i) for i in 'xyz'), key=lambda x:(-len(x),x)))
+
+    # unsorted, unordered xyz strings
+    sxyz0 = map(''.join, itertools.combinations_with_replacement('xyz', r=l))
+
+    # unsorted, ordered xyz strings
+    sxyz1 = map(sort1, sxyz0)
+
+    # sorted, ordered xyz strings
+    return sorted(sxyz1, key=lambda s: (sorted((-s.count(i) for i in 'xyz')), s))
+
+def pyscf_cart_order(l):
+    return sorted(map(''.join, itertools.combinations_with_replacement('xyz', r=l)))
+
+def pyscf_to_gms_order(l):
+    xg = gms_cart_order(l)
+    xp = pyscf_cart_order(l)
+    m = np.zeros((len(xp),len(xg)))
+    #p2g = [(xp.index(''.join(sorted(xi))),i) for i,xi in enumerate(xg)]
+    for i,xi in enumerate(xg):
+        m[i,xp.index(''.join(sorted(xi)))] = 1
+    return m
+
+s2cdict = {
+         2:[((0,1),12),
+            ((1,4),12),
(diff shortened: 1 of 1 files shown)


src/QMCTools/PyscfSphToCart.py
```diff
@@ -0,0 +1,502 @@
...
+    print(xg)
+    print(xp)
+    m = np.zeros((len(xp),len(xg)))
+    #p2g = [(xp.index(''.join(sorted(xi))),i) for i,xi in enumerate(xg)]
+    for i,xi in enumerate(xg):
+        m[i,xp.index(''.join(sorted(xi)))] = 1
+    print(m)
+    return m
+
+s2cdict = {
```
markdewing, line 56: Could you add some documentation about what the numbers mean in this structure?

kgasperich: yes, I'm still moving some factors around, but I plan to add more documentation

kgasperich: I added a description there.

It's not my ideal solution (I should just add a function to generate these for arbitrary l/m), but I wanted to get something working first, and this was simple. 
I squared everything so that I could keep an array of values with finite decimal expansions.

I haven't noticed any issues with accumulation of floating-point errors yet, but I should probably consider whether that might become a problem for high enough angular momenta.
(This actually led me to your jupyter notebooks yesterday: I `grep`ped for `factorial` while trying to decide whether to take a product of ratios `prod(a_i/b_i)` or a ratio of products `prod(a_i)/prod(b_i)` and found a comment that mentioned the `GaussianOrbitals` notebook from the qmc_algorithms repo)

kgasperich:
I think the only thing left to do here is to clean up the tests.
Here is a summary of the testing steps:
1. run small pyscf calculation in each of several basis sets (sp, sd, sf, sg, ...) [[Done](https://github.com/kgasperich/qmcpack/blob/823a4563173775cac9f7bfabecc17a5d441cf31b/tests/pyscf/CMakeLists.txt#L83-L87)*]
   1. *need to change the string back to `'pdfghi'` [here](https://github.com/kgasperich/qmcpack/blob/823a4563173775cac9f7bfabecc17a5d441cf31b/examples/solids/pyscf-inputs/he2_1x1x1/he2_1x1x1_sph2cart.py#L61);
    it was shortened to save time while the test workflow is ironed out
3. for each basis, call converter to generate cartesian hdf5 input [[Done](https://github.com/kgasperich/qmcpack/blob/823a4563173775cac9f7bfabecc17a5d441cf31b/tests/pyscf/CMakeLists.txt#L89-L96)]*
   1. *as above, need to add `sf sg sh si` back to the loop
4. for each basis, use convert4qmc to generate xml inputs from hdf5 [[needs to be added as cmake test](https://github.com/kgasperich/qmcpack/blob/823a4563173775cac9f7bfabecc17a5d441cf31b/examples/solids/pyscf-inputs/he2_1x1x1/test.sh#L24-L25)]
5. run short VMC with same seed for sph and cart inputs [[needs to be added as cmake test](https://github.com/kgasperich/qmcpack/blob/823a4563173775cac9f7bfabecc17a5d441cf31b/examples/solids/pyscf-inputs/he2_1x1x1/test.sh#L26)]
6. verify that sph/cart outputs agree [[needs to be done in a cleaner way](https://github.com/kgasperich/qmcpack/blob/823a4563173775cac9f7bfabecc17a5d441cf31b/examples/solids/pyscf-inputs/he2_1x1x1/test.sh#L27-L35) (maybe with [check_scalars.py](https://github.com/kgasperich/qmcpack/blob/sph2cart/tests/scripts/check_scalars.py)) and added as a cmake test]

src/QMCTools/PyscfSphToCart.py
anbenali: @prckent @kgasperich 

I tested this and it works and solves the #5205 issue.

ye-luo:
Could you expand the documentation mentioning spherical vs Cartesian and this tool that helps converting under [LCAO](https://qmcpack.readthedocs.io/en/develop/intro_wavefunction.html#linear-combination-of-atomic-orbitals-lcao-with-gaussian-and-or-slater-type-basis-sets) to make this tool visible.

Issue #4157, opened 2022-08-05
Read/write walker weights in restart


ye-luo:
**Is your feature request related to a problem? Please describe.**
Need to record walker weights.

quantumsteve:
@ye-luo Is this roughly what you had in mind?
https://github.com/QMCPACK/qmcpack/compare/develop...quantumsteve:qmcpack:4157_walker_weights

Wanted to check I'm on the right track before going further.

PR #4366, opened 2022-12-14
Use std::filename::path and avoid sprintf in HDFWalkerInput

quantumsteve:

## Proposed changes

Some linting before starting #4157. I tried using std::filesystem::path in more places in `snprintf` instead of the more dangerous `sprintf`. 

- No

## Checklist


src/Particle/HDFWalkerInput_0_4.cpp
```diff
@@ -74,20 +70,22 @@ void HDFWalkerInput_0_4::checkOptions(xmlNodePtr cur)
...
       }
       else
       {
-        int pid      = myComm->rank() % i_info.nprocs;
-        char* h5name = new char[froot.size() + 10];
-        sprintf(h5name, "%s.p%03d", froot.c_str(), pid);
-        FileStack.push(h5name);
-        delete[] h5name;
+        int pid = myComm->rank() % i_info.nprocs;
+        // 2 chars for '.p', 11 chars for pid, 1 char for \0
```
markdewing, line 84: The source of the "11 characters" is a little confusing at first glance because the format string is "%03d".  It looks like it should be zero-padded with a width of 3.  However, the width specifier is a minimum width - printf formatting expands the field to accommodate larger values.  The type of pid is an int, and so the maximum number of characters need to represent the int with the longest string is 11 (For instance, -2**31)

quantumsteve: Yes, I made that mistake in the first commit and am thankful that GCC caught it. clang didn't generate any warnings. 

The `+10` buffer currently used could also potentially overflow, though I suspect other things would also break with 10,000,000+ processes.

src/Particle/HDFWalkerInput_0_4.h
```diff
@@ -81,11 +82,11 @@ struct HDFWalkerInput_0_4
...
 
   /** read walkers for small number of MPI tasks */
-  bool read_hdf5(std::string h5name);
+  bool read_hdf5(std::filesystem::path h5name);
   /** read walkers. Master reads and scatter the walkers */
-  bool read_hdf5_scatter(std::string h5name);
+  bool read_hdf5_scatter(std::filesystem::path h5name);
   /** read walkers using PHDF5 */
-  bool read_phdf5(std::string h5name);
+  bool read_phdf5(std::filesystem::path h5name);
```
ye-luo, line 89: For the above three APIs, can we have `const std::filesystem::path&`?

quantumsteve: These functions all call `h5name.concat(hdf::config_ext);` which modifies the value of h5name.

ye-luo: That is the confusing part. I feel the argument should be renamed as h5name_noext and kept intact and then h5name is used for the full name.

src/Particle/HDFWalkerInput_0_4.h
```diff
@@ -57,9 +58,9 @@ struct HDFWalkerInput_0_4
   //propoery list to handle parallel io
   hid_t h_plist;
   //the last file which was processed
-  std::string FileName;
+  std::filesystem::path FileName_noext;
   //list of files to process
-  std::stack<std::string> FileStack;
+  std::stack<std::filesystem::path> FileStack;
```
ye-luo, line 63: Does each element include extension or not? No need of updating the variable name, updating the comments can be sufficient.

quantumsteve: No, if the extension `.config.h5` is present it is removed. I updated the comments to reflect this.

ye-luo:
Test this please

PR #4369, opened 2022-12-15
Add flag to turn off NLPP grid randomization

markdewing:
Add input flag to turn off the non-local pseudopotential grid randomization.

One motivation is for testing of parameter derivatives against an external script.  It's easier if there is no rotation of the grid.

Another motivation is this may be one approach to splitting up the necessary updates in #4362.

- New feature

- No

desktop

## Checklist


prckent:
I am not convinced that we need this. Convince us.

I am not generally keen on adding more input flags.

markdewing:
Added in #4394, don't need this PR

Issue #1294, opened 2019-01-03
Draft build recipe and bug workaround for QE6.3 on Summit IBM P9


prckent:
I plan to add some build instructions for QE6.3 to the manual since QE does not build out of the box due to minor architecture issues.  The unreleased develop version has fixes by Pietro Delugas but our converter patches v6.3. Suggestions welcome.

```
#!/usr/bin/bash
module load gcc/6.4.0 #6.4.0 was default on 2019-01-03
module load essl
module load netlib-lapack #because ESSL does not provided needed LAPACK functionality
module load hdf5
module load fftw
export FFTW_HOME=$OLCF_FFTW_ROOT
module -t list
./download_and_patch_qe6.3.sh
cd qe-6.3
cd archive
mv fox.tgz fox.tgz_orig
wget https://gitlab.com/QEF/q-e/raw/develop/archive/fox.tgz # Latest FoX on develop is patched for PPC64 architectures
cd ..
./configure --with-hdf5=$OLCF_HDF5_ROOT
cp -p make.inc make.inc_orig
sed 's/libhdf5.a /libhdf5.a -L/g' make.inc_orig >make.inc # Incorrect awk in install/configure drops library dir for zlib
make pw
```

prckent:
The above does not use ESSL yet.

prckent:
Now part of #1301

Issue #1292, opened 2019-01-03
Nexus tests are not created


prckent:
Nexus tests are missing from cdash and nightly testing. Seemingly they are no longer added as part of the configure despite a lack of warnings. e.g. https://cdash.qmcpack.org/CDash/viewConfigure.php?buildid=37253
```
Adding test example_He_from_gamess-1-1
Adding Nexus tests
-- Configuring done
```
Presumably this is a slight misconfiguration somewhere. e.g. due to the recent renaming.

PR #4375, opened 2022-12-20
Add VPs to SOECP

camelto2:
## Proposed changes

This PR adds VirtualParticleSets to SOECPs. This essentially makes it so that we no longer have to add an additional keyword to run with SOECPs since it can now use the default batched path with virtual particles. 

Note that while this is only implemented in the legacy APIs, this actually enables SOECPs to be used in the batched code since the default behavior of OperatorBase is to fallback to a loop over single walker APIs. So effectively, we have a path to run using the batched code, although it won't be as efficient as the mw implementation (coming soon). 

I did a little refactoring in SOECP, and had to change the makeMoves API in VirtualParticleSet so that we also store the current electron spin...otherwise the NonLocalECP part was failing when spinors were enabled. 

I also updated the short/long/determinanistic tests for bccMo to remove the algorithm=non-batched so that it uses the batched by default.

Also, I'm trying to make it so that I don't use raw pointers for the VirtualParticleSet (that's how it is still done in NonLocalECPComponent as well) and use smart pointers, but I ran into some hiccups as I was trying to change that and other things to use smart pointers. I'm still playing with that, but I wanted to go ahead and get this up so the rest of the code could be looked at. 

- New feature

- No

my mac

## Checklist

### DIFF
--- a/src/Particle/VirtualParticleSet.cpp
+++ b/src/Particle/VirtualParticleSet.cpp
@@ -116,6 +116,31 @@ void VirtualParticleSet::makeMoves(int jel,
   update();
 }
 
+/// move virtual particles to new postions and update distance tables
+void VirtualParticleSet::makeMovesWithSpin(int jel,
+                                           const PosType& ref_pos,
+                                           const std::vector<PosType>& deltaV,
+                                           const RealType& ref_spin,
+                                           const std::vector<RealType>& deltaS,
+                                           bool sphere,
+                                           int iat)
+{
+  if (sphere && iat < 0)
+    throw std::runtime_error(
+        "VirtualParticleSet::makeMoves is invoked incorrectly, the flag sphere=true requires iat specified!");
+  onSphere      = sphere;
+  refPtcl       = jel;
+  refSourcePtcl = iat;
+  assert(R.size() == deltaV.size());
+  assert(spins.size() == deltaS.size());
+  for (size_t ivp = 0; ivp < R.size(); ivp++)
+  {
+    R[ivp]     = ref_pos + deltaV[ivp];
+    spins[ivp] = ref_spin + deltaS[ivp];
+  }
+  update();
+}
+
 void VirtualParticleSet::mw_makeMoves(const RefVectorWithLeader<VirtualParticleSet>& vp_list,
                                       const RefVector<const std::vector<PosType>>& deltaV_list,
                                       const RefVector<const NLPPJob<RealType>>& joblist,
--- a/src/Particle/VirtualParticleSet.cpp
+++ b/src/Particle/VirtualParticleSet.cpp
@@ -99,7 +99,7 @@ void VirtualParticleSet::releaseResource(ResourceCollection& collection,
 
 /// move virtual particles to new postions and update distance tables
 void VirtualParticleSet::makeMoves(int jel,
-                                   const PosType& ref_pos,
+                                   const ParticleSet& p,
                                    const std::vector<PosType>& deltaV,
                                    bool sphere,
                                    int iat)
@@ -112,15 +112,17 @@ void VirtualParticleSet::makeMoves(int jel,
   refSourcePtcl = iat;
   assert(R.size() == deltaV.size());
   for (size_t ivp = 0; ivp < R.size(); ivp++)
-    R[ivp] = ref_pos + deltaV[ivp];
+  {
+    R[ivp]     = p.R[jel] + deltaV[ivp];
+    spins[ivp] = p.spins[jel]; //no spin deltas in this API
+  }
   update();
 }
 
 /// move virtual particles to new postions and update distance tables
 void VirtualParticleSet::makeMovesWithSpin(int jel,
-                                           const PosType& ref_pos,
+                                           const ParticleSet& p,
                                            const std::vector<PosType>& deltaV,
-                                           const RealType& ref_spin,
                                            const std::vector<RealType>& deltaS,
                                            bool sphere,
                                            int iat)
@@ -135,8 +137,8 @@ void VirtualParticleSet::makeMovesWithSpin(int jel,
   assert(spins.size() == deltaS.size());
   for (size_t ivp = 0; ivp < R.size(); ivp++)
   {
-    R[ivp]     = ref_pos + deltaV[ivp];
-    spins[ivp] = ref_spin + deltaS[ivp];
+    R[ivp]     = p.R[jel] + deltaV[ivp];
+    spins[ivp] = p.spins[jel] + deltaS[ivp];
   }
   update();
 }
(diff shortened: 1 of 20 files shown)


src/Particle/VirtualParticleSet.h
```diff
@@ -75,16 +75,19 @@ class VirtualParticleSet : public ParticleSet
...
      * @param deltaV Position delta for virtual moves.
      * @param sphere set true if VP are on a sphere around the reference source particle
      * @param iat reference source particle
      */
-  void makeMoves(int jel,
-                 const PosType& ref_pos,
-                 const std::vector<PosType>& deltaV,
-                 bool sphere = false,
-                 int iat     = -1);
+  void makeMoves(int jel, const ParticleSet& p, const std::vector<PosType>& deltaV, bool sphere = false, int iat = -1);
```
ye-luo, line 83: Only exposing ref_pos is preferred. Please add two argument ref_pos and ref_spinor instead of particleset. Also add function comment.

src/QMCHamiltonians/tests/test_ecp.cpp
```diff
@@ -536,21 +536,32 @@ TEST_CASE("Evaluate_soecp", "[hamiltonian]")
...
-      if (sopp != nullptr && dist[iat] < sopp->getRmax())
+      const auto& dist  = myTable.getDistRow(jel);
+      const auto& displ = myTable.getDisplRow(jel);
+      for (int iat = 0; iat < ions.getTotalNum(); iat++)
       {
-        Value1 += sopp->evaluateOne(elec, iat, psi, jel, dist[iat], RealType(-1) * displ[iat]);
+        if (sopp != nullptr && dist[iat] < sopp->getRmax())
+        {
+          Value1 += sopp->evaluateOne(elec, iat, psi, jel, dist[iat], RealType(-1) * displ[iat]);
+        }
```
ye-luo, line 551: Could you remove two layers of `{}`

ye-luo (changes requested):

src/Particle/VirtualParticleSet.h
camelto2: done

src/QMCHamiltonians/tests/test_ecp.cpp
camelto2: done

src/Particle/VirtualParticleSet.cpp
```diff
@@ -112,7 +113,35 @@ void VirtualParticleSet::makeMoves(int jel,
   refSourcePtcl = iat;
   assert(R.size() == deltaV.size());
   for (size_t ivp = 0; ivp < R.size(); ivp++)
-    R[ivp] = ref_pos + deltaV[ivp];
+  {
+    R[ivp]     = ref_pos + deltaV[ivp];
+    spins[ivp] = ref_spin; //no spin deltas in this API
```
ye-luo, line 118: I'm not getting why `spins` needs to be updated. When not using spinor, Should the whole run not depending on the value of spins? Will there be any issue if we keep makeMoves untouched?

ye-luo:
Test this please

src/Particle/VirtualParticleSet.cpp
camelto2: Not for normal calculations. But for calculations that do include spinors yes it does matter. We still have NonLocalECPs used in spinor calculations, so if those evaluateRatios don't use VPs that have up to date spin values, the ratio evaluations are wrong

camelto2: Basically, if we don't have spinors then the original behavior of makeMoves would be fine. But if we do have spinors, then I need it to also have up to date spins (but no actual displacement from their original values). 

I wasn't sure of the best way to handle this.

camelto2: For reference, if I don't keep the spins up to date in NonLocalECPComponent, then I get the following

> mymac:build_gcc cmelton$ ctest -R SOREP-vmc                              
Test project /Users/cmelton/Software/qmcpack/build_gcc
      Start 1286: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16
 1/24 Test #1286: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16 ......................   Passed   81.97 sec
      Start 1287: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-kinetic
 2/24 Test #1287: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-kinetic ..............   Passed    0.03 sec
      Start 1288: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-totenergy
 3/24 Test #1288: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-totenergy ............***Failed    0.02 sec
      Start 1289: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-eeenergy
 4/24 Test #1289: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-eeenergy .............   Passed    0.02 sec
      Start 1290: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-samples
 5/24 Test #1290: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-samples ..............   Passed    0.02 sec
      Start 1291: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-potential
 6/24 Test #1291: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-potential ............***Failed    0.02 sec
      Start 1292: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-ionion
 7/24 Test #1292: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-ionion ...............   Passed    0.02 sec
      Start 1293: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-localecp
 8/24 Test #1293: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-localecp .............   Passed    0.02 sec
      Start 1294: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-nonlocalecp
 9/24 Test #1294: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-nonlocalecp ..........***Failed    0.02 sec
      Start 1295: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-soecp
10/24 Test #1295: short-bccMo_2x1x1_SOREP-vmc_sdj-1-16-soecp ................   Passed    0.02 sec

ye-luo: > so if those evaluateRatios don't use VPs that have up to date spin values, the ratio evaluations are wrong

evaluateRatios should be fixed to handle VP isSpinor() true and false cases. It should not blindly use spins. VP isSpinor() should carry identical value as the ParticleSet it constructs from.

ye-luo: Thanks for the info and I have a better picture now. I will make a follow up PR to consolidate APIs.

ye-luo (approved):

PR #4376, opened 2022-12-22
Remove use of bmpi3 deprecated functions

correaa:
## Proposed changes

remove artificial uses of deprecated BMPI3 functions, like assignments.

- No

Ubuntu 22.04

## Checklist


src/QMCApp/qmcapp.cpp
```diff
@@ -46,7 +46,7 @@ int main(int argc, char** argv)
   using namespace qmcplusplus;
 #ifdef HAVE_MPI
   mpi3::environment env(argc, argv);
-  OHMMS::Controller->initialize(env);
+  OHMMS::Controller = new Communicate(env.world());
```
ye-luo, line 49: Memory leak. First allocated https://github.com/QMCPACK/qmcpack/blob/f1bfdccbd4abd1d57892c00dbb570ecbf2949370/src/Message/Communicate.cpp#L35

correaa: the memory leak is there anyway because the global controller is initialized with `new` anyway. The correct way would be to use a `std::unique_ptr` but fixing the leak was not the idea here.

correaa:
let's continue in the other PR https://github.com/QMCPACK/qmcpack/pull/4377. 
This one is mixed with the update in the library.

PR #4379, opened 2023-01-04
Protect complex reduction in real builds.

ye-luo:
## Proposed changes
Getting llvm-link error in real builds with icpx due to unprotected reduction on complex in offload regions. 

- Bugfix

- No

sunspot

## Checklist


ye-luo:
Test this please

prckent (approved):
Q. Since we will likely gain more of these workarounds, how about adding // COMPILER WORKAROUND or similar so that we can find these in future?

prckent:
Test this please

ye-luo:
> Q. Since we will likely gain more of these workarounds, how about adding // COMPILER WORKAROUND or similar so that we can find these in future?

Added on https://github.com/QMCPACK/qmcpack/pull/4380 which needs a new approval

PR #4382, opened 2023-01-05
Restore old interop scheme using level_zero.

ye-luo:
## Proposed changes
Using OpenMP interop to directly create SYCL queue has a lot of limitations. Revert to the old way exposing only the Level0 device and context object.

- Other (please describe):

- No

sunspot

## Checklist


ye-luo:
Test this please

PDoakORNL (approved):
We can only leverage the performance portability from openmp as much as is usable.

PR #4384, opened 2023-01-05
Sycl queue created instead of copy

ye-luo:
## Proposed changes
SYCL queue copy constructor only does shallow copy. Underlying level 0 objects are shared and caused performance penalty due to undesired entanglement.
Make actual queue creation from device and context instead of copying.

- Bugfix

- No

sunspot

## Checklist


ye-luo:
Test this please

prckent (approved):
LGTM

PR #4385, opened 2023-01-06
Nexus: Handle pickle protocols between python versions

kayahans:
## Proposed changes

Handles the pickle (https://docs.python.org/3/library/pickle.html) protocol differences that can arise due to using different python versions, (such as using multiple computers having different python versions, but sharing the same filesystem). In the current code, if python 3.8 is used to generate pickle (.p) files, nexus uses protocol 5. If python 3.7 is used to rerun the same script, protocol 5 is not accessible in that python version (https://docs.python.org/3/library/pickle.html#data-stream-format)

- No

Andes, Summit

## Checklist


prckent (changes requested):
Please add exception handling on the pickle5 import and print a message about why it is needed if the import fails.

The mixed python situation is fairly common. e.g. Old system python, user installed recent python.

prckent:
While tweaking this code, add some error handling on the open call? Or is the file always guaranteed to be there and readable at this point?

kayahans:
Import error exception is added, please check. File is always guaranteed to be there and readable per nexus workflows.

prckent (approved):
LGTM

I tweaked the message since there are many ways to install python modules and pip might not be the relevant one.

prckent:
Test this please

ye-luo (approved):

ye-luo:
Test this please

PR #4378, opened 2022-12-22
Update reference ParticleSet handling in VirtualParticleSet

ye-luo:
## Proposed changes
It is better to update the reference ParticleSet whenever makeMoves happen.

## What type(s) of changes does this code introduce?
- Refactoring

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
laptop

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted

PR #4388, opened 2023-01-07
Fix refPS assignment

ye-luo:
## Proposed changes
Fix a bug introduced in #4378
https://cdash.qmcpack.org/CDash/testDetails.php?test=24412236&build=387147

Also add a few missing resource lock in performance tests.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):
Q. How do you think this could this have been caught sooner? Deterministic test with offload in CI? Can we exercise this code more with a unit test or at least not full run?

ye-luo:
> Q. How do you think this could this have been caught sooner? Deterministic test with offload in CI? Can we exercise this code more with a unit test or at least not full run?

Should use unit test. Should not counting on deterministic simulation test to exactly cover this case.

PR #4389, opened 2023-01-08
Update intel classic compiler version minimum to 2021.1

ye-luo:
## Proposed changes
Too old intel classic compilers are too painful to deal with. See https://github.com/QMCPACK/qmcpack/pull/4359 CI failure.
Raised to >=2021.1 released in the end of 2020 (as part of oneapi 2021.1)
@williamfgc Update CI will be needed to merge this PR.

## What type(s) of changes does this code introduce?
- Build related changes

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
bora

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'

Issue #4326, opened 2022-11-14
Intel OneAPI compiler needed in sulfur CI


prckent:
One day we will have to drop the classic Intel compiler. (Re: #4319 )

Before whenever that is, we should have a OneAPI build fully working in self-hosted CI.
`$source /opt/intel/oneapi/setvars.sh  intel64`

PR #4393, opened 2023-01-10
Bump icc version to 2021 from oneAPI 2023 on sulfur CI

williamfgc:
## Proposed changes

Uses classic compiler icc inside oneAPI (to be deprecated) on sulfur CI
Related to #4389 , #4326 , #4359 
Move eventually to oneAPI icx compiler

- No

sulfur 

## Checklist


ye-luo (approved):
LGTM. @prckent please force merge.

prckent (approved):
I'll approve and merge this, but on the next update I realized we should do a couple of things:

- This job should be called IntelClassic and not have the numerical component in it to avoid confusion. 
- After sourcing compilervars.sh the compilers will be first on the path, so the OMPI lines should be export OMPI_CC=`which icc`  etc. The script will either fail or not use the expected compiler next minor version update from Intel due to hardcoded paths.  The GITHUB_ENV lines also need changing.
- For coverage of different numerical libraries, this build should use MKL from the OneAPI install and not OpenBLAS.

PR #4394, opened 2023-01-10
Test NLPP parameter derivative

markdewing:
This test provides a use-case for #4369.  I tried adding a "debug_" prefix to the flag, but "debug_randomize_grid" looks like "debug" is a verb.  Having a default of "true" for normal operation seemed strange. 
In this PR, the sense of the flag is flipped and named "disable_randomize_grid".  This seems to be more natural as normal operation has the flag off.

There are alternatives to using this flag, but I didn't pursue them:
1. Apply the same random grid rotation in the script.  This adds an extra step to the calculation, which make debugging harder.  It also makes the values depend on the precise numbers coming from the random number generator (If the RNG changes, then the test needs to change)
2. Could address that second point by using FakeRandom, but that introduces other complications into the build.

To simplify the comparison script, there are no long-range sums.   There is no e-e or i-i term in the Hamiltonian, and the pseudopotential has pbc="no".

The wavefunction uses a multideterminant wavefunction.  In the future, this test can be copied and changed to use a single determinant wavefunction to test the fixes needed in that case.

- No

desktop

## Checklist


prckent:
I agree with your `disable_randomize_grid` logic. Can merge it here or in the older PR.

prckent (commented):
Are you proposing to move all the existing potentials to tests/pseudopotentials_for_tests in a later PR?

markdewing:
I moved this pseudopotential because it made it easier to add it to the list in QMCHamiltonians/tests/CMakeLists.txt.   The existing placement of pseudopotential files seems to be a mix of in-place, and linked to pseudopotentials_for_tests.  I could move other existing files if you want, in a future PR.

prckent (approved):
LGTM

Browsing via the web it looks like most potentials have already been relocated. If you can move any remaining for consistency, please do.

prckent:
Test this please

PR #4400, opened 2023-01-18
[WIP] Remove mpi walker deduplication step one

PDoakORNL:
Currently we have a bunch of complexity and more difference than necessary between the MPI and no MPI population control.  This is due to an optimization to account for the case where more than one copy of the same walker is sent to the same rank.  It is not proven this is an important case at all and it was only ported over to the batched design when a rewrite of the population control without it didn't match the behavior of legacy.

## Proposed changes
Simplify the DMC population control and provide sensible walker IDs that encode the parent and rank the walker originated from.  To accomplish this walker copies spawned on the originating node instead of being simply a copy of the (usually) retained copied walker.  Which is then spawned a variable number of times on the receiving node.

Right now I have added the spawning on the sender or non mpi node which results in no walker with num_copies > 1 arriving to the swap code.  This means much of the code in swapWalkersSimple is now dead code to minimize changes I will remove that in the next PR in this series.

- Bugfix
- New feature
- Build related changes
- Documentation changes
- Other (please describe):

- Yes
- No

## Checklist


ye-luo (changes requested):
The amount of changes related to variable renaming, added APIs due to marking variable private dominates this PR. It prevents from reviewing actual functional changes in this PR.

PDoakORNL:
When I have time I will get back to this and do the refactor naming first.  But WalkerID's in the per particle logger will be unreliable at best until then.

PR #4402, opened 2023-01-19
Evaluate parameter derivatives for NLPP

markdewing:
Add `evaluateDerivativesWF` for single determinants to RotatedSPOs.  This enables the Hamiltonian parameter derivative from NLPP for single determinant for non-batched NLPP.  The function is a reduced version of `evaluateDerivatives`. It only needs the derivative of the wavefunction.

Add `evaluateDerivRatios` for single determinants to RotatedSPOs.  This enables the Hamiltonian parameter derivative from NLPP for single determinants for the batched NLPP algorithm.   The function combines functionality from `evaluateDetRatios` and `evaluateDerivativesWF`. Two parameters were added to `evaluateDerivRatios` - the original particle set, and the index of the electron being moved.  The derivatives of the rotation matrix need an inverse of the entire matrix of orbitals.  I don't know if there are shortcuts to get away without needing these, but this is the most straightforward combining of existing code.

These changes were pulled out of #4351

- New feature

- No

laptop

## Checklist


ye-luo:
The original particleset and the reference particle exist as `VirtualParticleSet::getRefPS()` and `refPtcl`. Preferably `refPtcl` should be marked private and accessed via an API (A later topic/PR). evaluateDerivRatios can be restored as it was.

markdewing:
Test this please

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -206,6 +214,18 @@ class SPOSet : public QMCTraits
                                  const ValueVector& psiinv,
                                  std::vector<ValueType>& ratios);
 
+
+  /// Determinant ratios and parameter derivatives of the wavefunction for virtual moves
+  virtual void evaluateDerivRatios(const VirtualParticleSet& VP,
```
ye-luo, line 219: For both added APIs, I feel it will be better to pass the inverse matrix in instead of recomputing inside. Consider that in a later PR.

ye-luo (commented):
Need to think about expanding coverage. But I would recommend merging the current PR as it is. LGTM.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -306,6 +306,131 @@ void RotatedSPOs::log_antisym_matrix(ValueMatrix& mat)
...
+  const ParticleSet& P = VP.getRefPS();
+  int iel = VP.refPtcl;
+
+  Phi->evaluate_notranspose(P, FirstIndex, LastIndex, psiM_all, dpsiM_all, d2psiM_all);
+
+  for (int i = 0; i < nel; i++)
+    for (int j = 0; j < nel; j++)
+      psiM_inv(i, j) = psiM_all(i, j);
+
+  Invert(psiM_inv.data(), nel, nel);
```
prckent, line 342: Noting the full cubic cost of matrix inversion here. However, getting something working first takes precedence over possible alternatives. Plus considering that the  initial electron counts will be smallish, this might not be so bad.

src/QMCWaveFunctions/SPOSet.h
prckent: Agree, as per my comment above.

prckent (approved):
LGTM

Thanks Mark! Exciting development.

prckent:
Test this please

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -306,6 +306,131 @@ void RotatedSPOs::log_antisym_matrix(ValueMatrix& mat)
...
+    Phi->evaluateValue(VP, iat, tmp_psi);
+
+    for (int j = 0; j < nmo; j++)
+      psiM_all(iel - FirstIndex, j) = tmp_psi[j];
+
+    for (int i = 0; i < nel; i++)
+      for (int j = 0; j < nel; j++)
+        psiM_inv(i, j) = psiM_all(i, j);
+
+    Invert(psiM_inv.data(), nel, nel);
```
markdewing, line 368: This is the inverse that is likely to get expensive, as it gets call nknot times.

Maybe combining a Sherman-Morrison update with the matrix multiply below will lead to some simplifications.

ye-luo: I believe the match can be formulated without any update.

PR #4404, opened 2023-01-20
 add docker tests for spack build centos image

walshmm:
## Proposed changes

Adds jobs to Github Actions to run ~~GCC9~~ GCC11 No MPI tests on a spack built Centos-Stream docker image.

- No

Ubuntu 20.04

## Checklist

### DIFF
--- a/config/docker/dependencies/centos-stream/Dockerfile
+++ b/config/docker/dependencies/centos-stream/Dockerfile
@@ -1,0 +1,80 @@
+# Build stage with Spack pre-installed and ready to be used
+FROM spack/centos-stream as builder
+
+
+# What we want to install and how we want to install it
+# is specified in a manifest file (spack.yaml)
+
+RUN which tar
+RUN tar -cf tarfilename.tar -T /dev/null
+
+RUN spack compilers
+
+RUN yum install texinfo -y
+
+RUN mkdir /opt/spack-environment \
+&& export gcc_vnew=8.5.0 \
+&& export cmake_vnew=3.24.2 \
+&& export libxml2_v=2.9.13 \
+&& export boost_vnew=1.79.0 \
+&& export hdf5_vnew=1.12.2 \
+&& export fftw_vnew=3.3.10 \
+&&  (echo "spack:" \
+&&   echo "  specs:" \
+&&   echo "  - gcc@${gcc_vnew}%gcc@${gcc_vnew}" \
+&&   echo "  - texinfo" \
+&&   echo "  - binutils" \
+&&   echo "  - git" \
+&&   echo "  - ninja" \
+&&   echo "  - cmake@${cmake_vnew}" \
+&&   echo "  - libxml2@${libxml2_v}%gcc@${gcc_vnew}" \
+&&   echo "  - boost@${boost_vnew}%gcc@${gcc_vnew}" \
+&&   echo "  - util-linux-uuid%gcc@${gcc_vnew}" \
+&&   echo "  - python%gcc@${gcc_vnew}" \
+&&   echo "  - hdf5@${hdf5_vnew}%gcc@${gcc_vnew} +fortran +hl ~mpi" \
+&&   echo "  - fftw@${fftw_vnew}%gcc@${gcc_vnew} -mpi" \
+&&   echo "  - openblas%gcc@${gcc_vnew} threads=openmp" \
+&&   echo "  - py-lxml" \
+&&   echo "  - py-matplotlib" \
+&&   echo "  - py-pandas" \
+&&   echo "  - py-scipy" \
+&&   echo "  - py-h5py ^hdf5@${hdf5_vnew}%gcc@${gcc_vnew} +fortran +hl ~mpi" \
+&&   echo "  concretizer:" \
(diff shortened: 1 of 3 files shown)


walshmm:
The runner seems to think my image doesnt have cmake :thinking: odd...

prckent (commented):
Why gcc9 from the toolset after using 8.5.0 for all the spack dependencies? You could get gcc9 or later from spack...

walshmm:
> Why gcc9 from the toolset after using 8.5.0 for all the spack dependencies? You could get gcc9 or later from spack...
@prckent 

It seems that when gcc is built from source it doesnt include an assembler or linker, at least not the ones QMCPACK was expecting(though I guess that begs the question as to how spack managed to use it--if it used it).  
As for why gcc8 at all... that's mostly arbitrary.  I was having issues discerning what versions were available.  

I can probably reorg the dockerfile such that spack uses the toolset gcc for compilation of dependencies, or if there is a way to acquire an assembler/linker (as and ld in the bin folder) then I can actually use spack gcc.

PDoakORNL:
If you want binutils i.e. ld, as, etc with a spack gcc you need to specify that as a variant.  I find usually you do, problems aren't frequent but they happen.

prckent (commented):
Do you have bandwidth to bump the gcc version? 11 or newer would be best given our support window.

With spack I usually add the newly built gcc as a compiler, as per the nightly setup scripts. You may have to add the latest binutils as well.

walshmm:
Ah now I recall why it was gcc 8.
``` 
spack compilers
```
returns 
```
==> Available compilers
-- gcc centos8-x86_64 -------------------------------------------
gcc@8.5.0
```
and fails if I try gcc 11.3.0

prckent:
Try
```
spack install gcc@11.3.0
spack load gcc@11.3.0
spack compiler find
spack compilers
```

walshmm:
Currently in progress, gcc compile is taking some time.

walshmm:
I had to do something slightly different
```
RUN spack install gcc@11.3.0
RUN spack load gcc@11.3.0
RUN spack compiler add $(spack location -i gcc@11.3.0)
RUN spack compilers
```

but now it seems configuration for 
```
cmake: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by cmake)
```
is bugging out.  Trying to find solutions that dont require me to build this image again... it takes a very long time.  Hoping its just a configuration issue because the file it needs is in the spack install of gcc.

prckent:
Sadly, spack does not handle the C++ library well yet.

Fix:
```
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:`which gcc|sed 's/bin\/gcc/lib64/g'`
```

PDoakORNL:
This is one of the reasons I just use the lld linker from llvm for whatever reason this seems to work at least when you build llvm yourself.  But you should also be able to add to the compiler.yaml
`extra_rpaths` for the compiler.  Then you don't have to use LD_LIBRARY_PATHS which I try to avoid.

Here LD_LIBRARY_PATHS may be a simpler solution but I don't think its how we should suggest users build.

walshmm:
Alright so the build is configured correctly now I believe.  The failure its exhibiting in the build step is reproducible on my local, so either this is a real build issue or a dependency mismatch in the docker file.  
Note: I was able to successfully build and run tests with gcc 9 from the toolkit package, so perhaps it has to do with gcc?

walshmm:
The last remaining test is failing successfully, I think I just need to clean up this pr and add documentation now.

quantumsteve:
Does this option build the missinglinker?
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/gcc/package.py#L122

config/docker/dependencies/centos-stream/Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+FROM spack/centos-stream as builder
+
+
+# What we want to install and how we want to install it
+# is specified in a manifest file (spack.yaml)
+
+RUN which tar
+RUN tar -cf tarfilename.tar -T /dev/null
+
+RUN spack install gcc@11.3.0
```
quantumsteve, line 11:
```suggestion
RUN spack install gcc@11.3.0+binutils
```

prckent: I hope that this will clear up a lot of the weirdness (imo) in the current configuration. For example, on RHEL we only occasionally have needed to setup an extra LD_LIBRARY_PATH, nothing like the extent of the workarounds here. (Can send the current versions of the setup scripts)

walshmm: It didnt seem to fix the workarounds, I'll be reverting back what was previous

markdewing (approved):

ye-luo (changes requested):
Temporary hold. Asking what merge method to use.

ye-luo (approved):
Will do squash merge.

ye-luo:
Test this please

ye-luo:
Test this please

PR #4408, opened 2023-01-23
Fix wrong output size in SplineC2ROMPTarget::mw_evaluateVGLandDetRatioGrads

ye-luo:
## Proposed changes
There is bug when using offload real build.
This caused writing to undesired memory locations and wrong numbers (crazy high variance/energy >>1)

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
@@ -247,14 +247,14 @@ void SplineC2ROMPTarget<ST>::evaluateValue(const ParticleSet& P, const int iat, 
     const size_t ChunkSizePerTeam = 512;
     const int NumTeams            = (myV.size() + ChunkSizePerTeam - 1) / ChunkSizePerTeam;
 
-    const auto padded_size = myV.size();
-    offload_scratch.resize(padded_size);
+    const auto spline_padded_size = myV.size();
+    offload_scratch.resize(spline_padded_size);
 
     // Ye: need to extract sizes and pointers before entering target region
-    const auto orb_size       = psi.size();
-    const auto* spline_ptr    = SplineInst->getSplinePtr();
-    auto* offload_scratch_ptr = offload_scratch.data();
-    auto* psi_ptr             = psi.data();
+    const auto requested_orb_size = psi.size();
+    const auto* spline_ptr        = SplineInst->getSplinePtr();
+    auto* offload_scratch_ptr     = offload_scratch.data();
+    auto* psi_ptr                 = psi.data();
     const auto x = r[0], y = r[1], z = r[2];
     const auto rux = ru[0], ruy = ru[1], ruz = ru[2];
     const auto myKcart_padded_size   = myKcart->capacity();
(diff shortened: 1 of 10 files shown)


src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
```diff
@@ -942,31 +958,32 @@ void SplineC2ROMPTarget<ST>::mw_evaluateVGLandDetRatioGrads(const RefVectorWithL
...
           spline2offload::evaluate_vgh_impl_v2(spline_ptr, ix, iy, iz, a, b, c, da, db, dc, d2a, d2b, d2c,
                                                offload_scratch_iw_ptr + first,
-                                               offload_scratch_iw_ptr + padded_size + first,
-                                               offload_scratch_iw_ptr + padded_size * 4 + first, padded_size, first,
-                                               index);
+                                               offload_scratch_iw_ptr + spline_padded_size + first,
+                                               offload_scratch_iw_ptr + spline_padded_size * 4 + first,
+                                               spline_padded_size, first, index);
         const size_t first_cplx = first / 2;
-        const size_t last_cplx  = omptarget::min(last / 2, orb_size);
```
ye-luo, line 949: The `min()` here is wrong. orb_size refers to output real orbitals.

src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
```diff
@@ -942,31 +958,32 @@ void SplineC2ROMPTarget<ST>::mw_evaluateVGLandDetRatioGrads(const RefVectorWithL
...
 
-        ValueType* restrict out_phi    = phi_vgl_ptr + iw * orb_size;
+        ValueType* restrict out_phi    = phi_vgl_ptr + iw * requested_orb_size;
         ValueType* restrict out_dphi_x = out_phi + phi_vgl_stride;
         ValueType* restrict out_dphi_y = out_dphi_x + phi_vgl_stride;
         ValueType* restrict out_dphi_z = out_dphi_y + phi_vgl_stride;
         ValueType* restrict out_d2phi  = out_dphi_z + phi_vgl_stride;
 
         const size_t first_real = first_cplx + omptarget::min(nComplexBands_local, first_cplx);
-        const size_t last_real  = last_cplx + omptarget::min(nComplexBands_local, last_cplx);
```
ye-luo, line 969: `last_real` can be larger than `requested_orb_size` and caused overflow.

ye-luo:
Test this please

prckent (commented):
Thanks Ye. We discussed this a bit last week, but so we have a record here: why didn't asan catch this?

ye-luo:
> Thanks Ye. We discussed this a bit last week, but so we have a record here: why didn't asan catch this?

Because we don't have a test case exposing this bug. The added test triggers asan failure when I backported it to develop.

ye-luo:
Test this please

ye-luo:
I also noticed that asan runs on the github actions without QMC_DATA. We probably need to update the image to include NiO a4(S1) and a16(S4) h5 files.

prckent (approved):
LGTM. Coverage would be higher but for #4409 .

We (I) should mention in the release notes that QMC_DATA and these NiO files are encouraged.

PR #4414, opened 2023-01-24
Create ConstantSPOSet for Unit Testing

rcclay:

## Proposed changes

This adds an SPOSet object that returns constant values for value, gradient, and laplacians.  These constant values are set at construction.  Basically, just set and keep Nelec x Norb matrices.  evaluateValue, evaluateVGL, and evaluate_notranspose just return slices of these matrices.  Similar spirit as FakeSPOSet, but with the ability to use arbitrarily sized and valued matrices.  

- New feature

- No

Intel Xeon, GNU 9.2.0, OpenMPI  1.10.1
## Checklist


prckent:
Since this is a new testing thing, please aim for really high coverage.

rcclay:
@prckent Fair point.  Will do.

src/QMCWaveFunctions/tests/test_ConstantSPOSet.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+
+#include "catch.hpp"
+#include "Configuration.h"
+#include "QMCWaveFunctions/OrbitalSetTraits.h"
+#include "QMCWaveFunctions/ConstantSPOSet.h"
+
+namespace qmcplusplus
+{
+//Ray:  Figure out how to template me on value type.
```
markdewing, line 20: For reference, this test uses TEMPLATE_TEST_CASE: https://github.com/QMCPACK/qmcpack/blob/19b99c950d1bb518332c69bda4adf823adab97d1/src/Numerics/tests/test_RotationMatrix3D.cpp#L20-L23

rcclay: Thank you!   I'll try to make more efficient use of template_test_case in the future.  For now, real vs. complex test cases are covered by letting the imaginary parts be zero.

ye-luo:
We have a class called "ConstantOrbital". The added one further caused confusion.

> Similar spirit as FakeSPOSet, but with the ability to use arbitrarily sized and valued matrices.

Can we fuse FakeSPOSet and this one and avoid increasing maintenance complexity.

rcclay:
@ye-luo For first point, it's definitely confusing, but I'd argue it's the fault of ConstantOrbital's name.  It's really a WaveFunctionComponent.  I think it was overlooked in renaming Orbitals to WaveFunctionComponents.  

For second point, I think the two could probably be collapsed into one class, but @markdewing is better to comment on it than me.  In the mean time, I need to look to see what all would be affected.

ye-luo:
> @ye-luo For first point, it's definitely confusing, but I'd argue it's the fault of ConstantOrbital's name. It's really a WaveFunctionComponent. I think it was overlooked in renaming Orbitals to WaveFunctionComponents.

agreed. ConstantOrbital was bad in the first place.

> For second point, I think the two could probably be collapsed into one class, but @markdewing is better to comment on it than me. In the mean time, I need to look to see what all would be affected.

Please discuss with @markdewing I think FakeSPO needs some extension to accommodate your needs. it is too rigid right now. The outcome of your discussion will be a good foundation for writing future tests involving SPOSet.

prckent:
Besides ConstantOrbital, we also have a LinearOrbital that is also a WaveFuctionComponent. The latter does get invoked by the test_dmc test.  Can we rename these?

rcclay:
For this PR, I'm going to:

- [ ] Replace FakeSPO instances with ConstantSPOSet.  
- [ ] Change API away from setting value, gradient, etc. matrices in inititalization towards getter/setter functions.

markdewing:
I have one concern about the "Constant" part of the name.  With ConstantOrbital and LinearOrbital, the first part of the name refers to mathematical properties, and the value and derivatives are consistent.  The "Constant" in ConstantSPO means "fixed" or "evaluated at a single point".   I'm okay with either meaning, but I'm wondering if the name "Constant" with two different uses would be confusing?  Or if the differences are clear enough from reading the code?  (Since ConstantOrbital needs to be renamed eventually for the "Orbital" part, the first word could be changed as well?)

rcclay:
ReferenceSPOSet?  ContainerSPOSet?

ye-luo:
If all the use cases of previous FakeSPOSet are covered. I would just claim it FakeSPOSet.

src/QMCWaveFunctions/ConstantSPOSet.h
```diff
@@ -0,0 +1,133 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
```
PDoakORNL, line 2: Mock classes should go in the testing directories not module roots.  See `src/Estimators/tests/FakeOperatorEstimator.h`  Generally they should not be header only. Making the tests build slower benefits nobody.  The targets they are built as part of shouldn't be linked into the application code.

If the class is with testing code and only linked to a test program or sub target of a test program it may not be necessary to call it Fake... or Mock...

src/QMCWaveFunctions/tests/test_ConstantSPOSet.cpp
```diff
@@ -0,0 +1,193 @@
...
+TEST_CASE("ConstantSPOSet", "[wavefunction]")
+{
+  //For now, do a small square case.
+  const int nelec = 2;
+  const int norb  = 2;
+
+  using RealType    = QMCTraits::RealType;
```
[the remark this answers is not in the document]
PDoakORNL: I'd suggest using src/QMCWaveFunctions/WaveFunctionTypes.hpp to define consistent types.  QMCTraits bakes in the multiexecutable design and is just bad. OrbitalSetTraits obfuscates the handling of precision.

example from `DiracDeterminantBatched`
```
  using WFT           = WaveFunctionTypes<VALUE, FP_VALUE>;
  using Value         = typename WFT::Value;
  using FullPrecValue = typename WFT::FullPrecValue;
  using PsiValue      = typename WFT::PsiValue;
  using LogValue      = typename WFT::LogValue;
  using Grad          = typename WFT::Grad;
  using Hess          = typename WFT::Hess;
  using Real          = typename WFT::Real;
```
Vector types gain no value from the typedefs at all since they can just be written
```
Vector<Value> row0;
Vector<Grad> grow0;
Vector<Value> lrow0l;
```

src/QMCWaveFunctions/tests/test_ConstantSPOSet.cpp
```diff
@@ -0,0 +1,193 @@
...
+  auto sposet_vgl = std::make_unique<ConstantSPOSet>("constant_spo", spomat, gradspomat,laplspomat);
+  psiV=0;
+  psiG=0;
+  psiL=0;
+  sposet_vgl->evaluateVGL(elec,1,psiV,psiG,psiL);
+   
+  CHECK(psiV[0] == row1[0]);
+  CHECK(psiV[1] == row1[1]);
+  
+  CHECK(psiG[0][0] == grow1[0][0]);
```
PDoakORNL, line 138: Use `src/Utils/forTesting/checkMatrix.hpp`?  This is painful to look at and it has pretty decent reporting if there is an bad element.

PDoakORNL (changes requested):
Class needs to be moved, there is some redundancy with FakeSpoSet but I don't think addressing that is hugely important in this PR, it could just be a good first issue for someone to work on.

src/QMCWaveFunctions/tests/test_ConstantSPOSet.cpp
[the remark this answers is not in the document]
rcclay: WaveFunctionTypes is the file I need, thank you.  I'll make those changes

src/QMCWaveFunctions/tests/test_ConstantSPOSet.cpp
rcclay: Also exact tool I need.  Thank you.  Will fix.

rcclay:
@ye-luo I took a stab at merging FakeSPO and ConstantSPOSet, and I think this might be better as a separate PR.  FakeSPO returns values for evaluateValue, evaluateVGL, evaluateAlltoOne, etc., through a mixture of returning matrix values set at construction, and on the fly-computation.  That would be fine by itself, but there's also some logic going on between plain evaluate_notranspose like calls  evaluateV* for ratio evaluations using an active particle that will take some disentangling, and maybe some tweaking of the tests.

src/QMCWaveFunctions/ConstantSPOSet.h
rcclay: My intended use case here is for Estimator/Hamiltonian evaluation, secondarily as a test of WaveFunctionComponents, so they would be linked to stuff in both of those directories.  Specifically, want to get rid of some of the SPOSet gymnastics done in things like test_soecp.cpp (and soon to be in magnetization density).  What do you recommend in this case?  Same location and naming convention?

PDoakORNL: Right now we do share tests classes across modules, all the minimal pools for instance.  At the moment the easier thing to do in that case is to leave the test class header only. Include by full path to the file from the project root if crossing modules.
ex:
```
#include "Particle/tests/MinimalParticlePool.h"
```

PDoakORNL: The only one that seems to make a target that gets linked by other tests is container_testing from the Containers module. But I think its overkill for this class.

src/QMCWaveFunctions/tests/test_ConstantSPOSet.cpp
rcclay: Do we have anything like checkMatrix for vectors?

rcclay:
Hi all--
Recent changes:

1. Moved ConstantSPOSet stuff to QMCWaveFunction/tests
2. Split implementation from header. 
3. Replaced matrix comparisons with checkMatrix().  
4. Added documentation to ConstantSPOSet header and setter functions.

src/QMCWaveFunctions/tests/test_ConstantSPOSet.cpp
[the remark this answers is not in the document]
rcclay: Done

src/QMCWaveFunctions/tests/test_ConstantSPOSet.cpp
rcclay: ParticleGradient and GradTypes are vectors of vectors, so for now I'm just checking them with loops.  Need to write some checker functions to handle vectors and vectors of vectors.

src/QMCWaveFunctions/tests/ConstantSPOSet.h
```diff
@@ -0,0 +1,85 @@
...
+#include "QMCWaveFunctions/SPOSet.h"
+
+namespace qmcplusplus
+{
+/** Constant SPOSet for testing purposes.  Fixed N_elec x N_orb matrices storing value, gradients, and laplacians, etc.,
+   *  These values are accessed through standard SPOSet calls like evaluateValue, evaluateVGL, etc.
+   *  Exists to provide deterministic and known output to objects requiring SPOSet evaluations.      
+   *
+   */
+struct ConstantSPOSet : public SPOSet
```
ye-luo, line 25: Class instead of struct.

rcclay: Whoops.  Done.

src/QMCWaveFunctions/tests/ConstantSPOSet.h
```diff
@@ -0,0 +1,85 @@
...
+  void evaluate_notranspose(const ParticleSet& P,
+                            int first,
+                            int last,
+                            ValueMatrix& logdet,
+                            GradMatrix& dlogdet,
+                            ValueMatrix& d2logdet) override;
+
+protected:
+private:
+  int numparticles_;
```
ye-luo, line 75: `const int numparticles_;` as this is set at the constructor.
Could you add a comment about this variable? I feel the name can be very confusing and documentation should help.

src/QMCWaveFunctions/tests/ConstantSPOSet.h
```diff
@@ -0,0 +1,85 @@
...
+  void evaluateVGL(const ParticleSet& P, int iat, ValueVector& psi, GradVector& dpsi, ValueVector& d2psi) override;
+
+  void evaluate_notranspose(const ParticleSet& P,
+                            int first,
+                            int last,
+                            ValueMatrix& logdet,
+                            GradMatrix& dlogdet,
+                            ValueMatrix& d2logdet) override;
+
+protected:
```
ye-luo, line 73: Remove it please.

src/QMCWaveFunctions/tests/ConstantSPOSet.h
rcclay: Done.

src/QMCWaveFunctions/tests/ConstantSPOSet.h
rcclay: Done

ye-luo (approved):

ye-luo:
@PDoakORNL good to be merged?

PDoakORNL (approved):

PDoakORNL:
I'm satisfied

ye-luo:
Test this please

PR #4415, opened 2023-01-24
refactor SOECPComponent spin integral

camelto2:

## Proposed changes

Refactor SOECPComponent evaluation. No new unit tests since this just rearranges the way I do evaluateOne and therefore the reference values stay the same. 

For some background on the motivation, I was working on evaluateValueAndDerivatives for SOECP both with and without VPs which enables the correct parameter derivatives during WF optimization when SOECPs are present. As I was implementing things, I ran into a snag due to how I initially wrote SOECPComponent. The solution was to either add a resize functionality to VPs in order to keep my previous code for evaluateOne untouched, or I could refactor evaluateOne and leave VPs alone. I felt it was better to leave VPs and their data sized correctly upon initialization, so I decided to refactor evaluateOne. 

To keep PRs small, I'm doing the refactor first and will then follow up with the evaluateValueAndDerivative PR next. 

- No

M1 mac

## Checklist


src/QMCHamiltonians/SOECPComponent.cpp
```diff
@@ -140,105 +154,84 @@ SOECPComponent::ComplexType SOECPComponent::getAngularIntegral(RealType sold,
...
-  }
-
+  int count = 0;
   RealType smin(0.0);
   RealType smax(TWOPI);
   RealType dS = (smax - smin) / sknot; //step size for spin
 
-  RealType sold = W.spins[iel];
-  ComplexType sint(0.0);
+  auto addSpatialQuadrature = [&](const int is, const RealType r, const PosType& dr, const RealType ds,
```
rcclay, line 201: Can you add a function comment here explaining what it's doing?  High level, plus what member variables get modified and how.  Also, there's an implied layout of quadrature grids w.r.t. real space and spin quadrature components for example.

rcclay (changes requested):
Looks pretty much good to go.  Other than lambda function comment, would recommend checking that the member variables are following code naming conventions.  total_knots_ vs. total_knots, etc.

src/QMCHamiltonians/SOECPComponent.cpp
camelto2: done. added a description of the lambda and a more detailed comment in the header describing buildTotalQuadrature

camelto2: also, changed all private data members of SOECPComponent to have the trailing _. Note I had to touch both the ECPotentialBuilder and test_ecp since both of those are friend classes that touched some of these members

rcclay (approved):
LGTM.

src/QMCHamiltonians/SOECPComponent.cpp
```diff
@@ -122,129 +128,122 @@ SOECPComponent::ComplexType SOECPComponent::lmMatrixElements(int l, int m1, int
...
-  buildQuadraturePointDeltas(r, dr, deltaV, snew - sold, deltaS);
+  if (sknot_ < 2)
+    APP_ABORT("Spin knots must be greater than 2\n");
 
-  if (VP)
+  if (sknot_ % 2 != 0)
+    APP_ABORT("Spin knots uses Simpson's rule. Must have even number of knots");
+
+  for (int ip = 0; ip < nchannel_; ip++)
   {
```
ye-luo, line 145: Remove {} around one line.

src/QMCHamiltonians/SOECPComponent.cpp
```diff
@@ -122,129 +128,122 @@ SOECPComponent::ComplexType SOECPComponent::lmMatrixElements(int l, int m1, int
...
+  if (VP_)
+  {
+    VP_->makeMovesWithSpin(W, iel, deltaV_, deltaS_, true, iat);
+    Psi.evaluateRatios(*VP_, psiratio_);
   }
   else
   {
-    //quadrature sum for angular integral
-    for (int j = 0; j < nknot; j++)
+    for (int iq = 0; iq < total_knots_; iq++)
```
ye-luo, line 159: Remove {} around the for loop.

ye-luo (approved):
We can keep the code more compact. Can be addressed in the following PRs. No need to do here.

ye-luo:
Test this please

rcclay (approved):

PR #4416, opened 2023-01-24
Fix restart tests for FreeBSD

prckent:
## Proposed changes

Use /usr/bin/env to find bash. On my install it is not /bin/bash, so the unmodified tests fail on attempting to run the check.sh scripts. We use env in a few other places already but are not consistent. With this fix ctest -L deterministic works for me on FreeBSD 13.1

- Bugfix
- Build related changes

- No

FreeBSD 13.1 virtual machine

## Checklist


markdewing (approved):

ye-luo:
To my best knowledge, we are not relying on "/usr/bin/env" even it exists in quite many python scripts. The final robust solution we picked was to specify the interpreter in the command line. I feel better to use `find_program(BASH bash /bin /usr/bin`. If bash is not found, we error out immediately.

prckent:
This PR makes things work ahead of 3.16.0 release.

I did look through the repo and noticed that we have a variety of #shell invocations on top of whatever CMake is invoking at different points. It would help to have a discussion, put the conclusion in the developer guide, and then make things consistent. There are also inconsistencies in how python is handled between QMCPACK and NEXUS that can trip up a user...

ye-luo:
> There are also inconsistencies in how python is handled between QMCPACK and NEXUS that can trip up a user...

could you elaborate the issue? QMCPACK dependency on python is short lived. Only during building and testing. Are you saying when running nexus tests, different python installations can be potentially pulled in? If this issue is not documented, let us put up an issue.

If you refer to users adopting a different python for nexus in production use, I assume they are at responsible.

ye-luo:
Test this please

PR #4417, opened 2023-01-26
Change catch REQUIRE to CHECK on all the FP values.

ye-luo:
## Proposed changes
This change is made by script. All the REQUIRE with APPROX are changed to CHECK.

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):
Good change. By not aborting immediately we will hopefully get better and quicker insights into failures.

https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md#natural-expressions

PR #4418, opened 2023-01-26
Parameter derivatives for SOECPotential

### LINKED ISSUE #4297: Jastrow optimization with SOECP no longer working
**Describe the bug**
Since QMCPACK 3.15.0, Jastrow optimization with SOECP is giving failure with following message,

terminate called after throwing an instance of 'std::logic_error'
  what():  Bug!! SOECPotential::evaluateValueAndDerivativesmust be overloaded when the OperatorBase depends on a wavefunction.

As it normally works on QMCPACK 3.14.0, a trigger looks to be located between 3.14.0 and 3.15.0.

**To Reproduce**
Try VTe2 optimization located on
https://anl.box.com/s/okkc0r0iuvhe5fzbd63kte62jdwqk8gr

**Expected behavior**
Jastrow optimization should work normally with SOECP.

**System:**
 - system name 
 theta, summit

camelto2:

## Proposed changes

This PR implements parameter derivatives for the SOECPotential. This closes #4297 

Note that I had to modify the SOECP unit test a bit in order to have a wave function that could be optimized. The new reference values are all obtained from the standalone code soecp_val_reference.cpp in QMCHamiltonians/tests. This implements everything from scratch and uses finite differences to obtain the reference values. 

- New feature

- No

M1 mac

## Checklist


prckent:
Test this please

src/QMCHamiltonians/SOECPComponent.cpp
```diff
@@ -193,6 +190,93 @@ SOECPComponent::RealType SOECPComponent::evaluateOne(ParticleSet& W,
...
+                                                                     const PosType& dr,
+                                                                     const opt_variables_type& optvars,
+                                                                     const Vector<ValueType>& dlogpsi,
+                                                                     Vector<ValueType>& dhpsioverpsi)
+{
+#ifndef QMC_COMPLEX
+  throw std::runtime_error("SOECPComponent::evaluateValueAndDerivatives should not be called in real build\n");
+#else
+  if (sknot_ < 2)
+    APP_ABORT("Spin knots must be greater than 2\n");
```
ye-luo, line 207: Not "greater equal"?
Could you move this check where sknot_ is set?
APP_ABORT better using exception.

camelto2:
Tests were failing with the real build for my initial PR because I have data that needs to interface with WF APIs and thus need to be ValueType....but part of the SOECP eval is inherently complex. Problem line was

wvec_[iq] = lsum * psiratio_[iq] * spin_quad_weights_[iq];

lsum has to be ComplexType (due to the ldots matrix element that is carried out) but wvec_ needs to be ValueType for the subsequent gemv call to get the dhpsioverpsi correct. 

I wasn't sure of the cleanest way to fix this...basically none of the spinor and SOECP stuff should ever be used if it is a real build in the first place but it seems we don't have perfect guards in place. I tried to modify the CMakeLists to only build SOECPotential/SOECPComponent if QMC_COMPLEX, but that left a lot undefined symbols due to the ECPotential/ECPComponent builders. So I resorted to adding an ifdef for now in the evaluateValueAndDerivatives...which I don't like because I know we are trying to move away from that but that was the simplest solution I came up with

src/QMCHamiltonians/SOECPComponent.cpp
camelto2: fixed

ye-luo:
Test this please

ye-luo (approved):

PR #4422, opened 2023-01-27
Check for OpenMP compiler support

### LINKED ISSUE #4372: Test for OpenMP compiler support

Some compilers don't support openmp even in the CPU only context. We should test for this and abort with a friendly message if the default QMC_OMP=1 is set; currently the first compilation test will break and give an unclear error.

Add a compilation check right at the top of our CMake.

PhilipFackler:
## Proposed changes
Added find_package call and meaningful error message.
Closes #4372 

- Build related changes

- No

Ubuntu 22.04

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -357,14 +357,18 @@ if(alignment_remainder)
 endif()
 message(STATUS "QMC_SIMD_ALIGNMENT is set to ${QMC_SIMD_ALIGNMENT}")
 
-#---------------------------------------------------------
-# Check for OpenMP compiler support if requested
-#---------------------------------------------------------
+#-------------------------------------------------------------------------------
+# Check for OpenMP compiler support
+# 
+# The CMake OpenMP package is *only* used to to verify compiler support.
+# No feature of the package is used and we customize OpenMP options directly
+# on a per vendor/compiler basis.
+#-------------------------------------------------------------------------------
 if(QMC_OMP)
   find_package(OpenMP)
   if(NOT OpenMP_FOUND)
-      message(FATAL_ERROR "No compiler support for OpenMP found. You will need to run CMake configure with -DQMC_OMP=OFF"
-                          "Switching to a compiler with OpenMP support is recommended.")
+      message(FATAL_ERROR "No compiler support for OpenMP found. Switching to a compiler with OpenMP support is recommended."
+                          "Alternatively, you will need to run CMake configure with -DQMC_OMP=OFF")
   endif()
 endif(QMC_OMP)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -363,8 +363,8 @@ message(STATUS "QMC_SIMD_ALIGNMENT is set to ${QMC_SIMD_ALIGNMENT}")
 if(QMC_OMP)
   find_package(OpenMP)
   if(NOT OpenMP_FOUND)
-      message(FATAL_ERROR "No compiler support for OpenMP..."
-                          "need to use -DQMC_OMP=OFF")
+      message(FATAL_ERROR "No compiler support for OpenMP found. You will need to run CMake configure with -DQMC_OMP=OFF"
+                          "Switching to a compiler with OpenMP support is recommended.")
   endif()
 endif(QMC_OMP)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -357,6 +357,17 @@ if(alignment_remainder)
...
 message(STATUS "QMC_SIMD_ALIGNMENT is set to ${QMC_SIMD_ALIGNMENT}")
 
+#---------------------------------------------------------
+# Check for OpenMP compiler support if requested
+#---------------------------------------------------------
+if(QMC_OMP)
+  find_package(OpenMP)
+  if(NOT OpenMP_FOUND)
+      message(FATAL_ERROR "No compiler support for OpenMP..."
+                          "need to use -DQMC_OMP=OFF")
```
williamfgc, line 367: Since OpenMP is a critical part of QMCPACK simulations I wonder if it's best to suggest finding compiler support for it. @prckent and @ye-luo thoughts?

prckent: Will push an edit

prckent (commented):
Have you been able to test this? e.g. Does the fruit company's compiler result in this error?

prckent (commented):
I haven't tested this code, but I think that this OpenMP availability test will need to be higher up. e.g. if the libstdc++ check has -fopenmp on the compile line, it will already have failed before we reach this point. The OpenMP test should be one of the first run.

CMakeLists.txt
```diff
@@ -357,6 +357,17 @@ if(alignment_remainder)
 endif()
 message(STATUS "QMC_SIMD_ALIGNMENT is set to ${QMC_SIMD_ALIGNMENT}")
 
+#---------------------------------------------------------
+# Check for OpenMP compiler support if requested
+#---------------------------------------------------------
+if(QMC_OMP)
+  find_package(OpenMP)
```
ye-luo, line 364: Please add a comment before this line saying
```
# OpenMP package currently is only used to to verify compiler support. No feature of this package is been used and we customize OpenMP options directly in per vendor configurations.
```

prckent:
This test needs to go before the Testlibstdc++ check, otherwise that can fail for the wrong reason. After the move, the current AppleClang 14.0.0.14000029 fails on the cxx17 check. According to https://en.cppreference.com/w/cpp/compiler_support/17 this is the correct result since memory resources are not yet implemented. Hopefully this will change shortly.

with QMC_OMP=ON or not specified:
```
-- QMCPACK adds -fstandalone-debug for Debug builds
-- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) 
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
-- Could NOT find OpenMP (missing: OpenMP_C_FOUND OpenMP_CXX_FOUND) 
CMake Error at CMakeLists.txt:301 (message):
  No compiler support for OpenMP found.  Switching to a compiler with OpenMP
  support is recommended.  Alternatively, you will need to run CMake
  configure with -DQMC_OMP=OFF


-- Configuring incomplete, errors occurred!
```
with QMC_OMP=OFF:
```
-- C++17 standard library support not found or incomplete
compiler is AppleClang 14.0.0.14000029
Compiler detected is <Clang> namely clang++ or a vendor variant (icpx, amdclang++, armclang++).
  If not using libcxx, ensure a GCC toolchain version equal or greater than 9.0 gets picked up. Check with '<Clang> -v'. Or use the --gcc-toolchain compiler option (added to both CMAKE_C_FLAGS and CMAKE_CXX_FLAGS) to point to a newer GCC installation.
  Output of test compile is in cpp17_compile_fail.txt
CMake Error at CMake/TestCxx17Library.cmake:55 (message):
  stopping
Call Stack (most recent call first):
  CMakeLists.txt:310 (include)


-- Configuring incomplete, errors occurred!
```

ye-luo:
@prckent I moved the openmp check up. Could you verify if it works for you? If yes. Please proceed to merge this PR.

ye-luo (approved):

prckent:
Can confirm your fix reproduces my local fix.

prckent:
Test this please

PR #4426, opened 2023-01-30
Orbital rotation part 1

markdewing:
Partially supersedes #4351.  This PR doesn't have the VP file save/restore code, but that will be in a subsequent PR.

Changes:
* The LCAO coefficient copy uses a shared pointer to reduce storage.  The rotation must be performed by only one thread.
* The bpsline coefficients are saved in order to enable the global rotation method.  The multiplication was changed slightly to copy the old spline coefficients to temporary storage first, then store the results of the multiplication in the actual spline storage.
* The global rotation method was made the default when optimize="yes" is specified in the input (versus #4351 where the history list method was the default.)
* The single-determinant derivatives were updated to enable Restricted Hartree-Fock wavefunctions (where the same orbitals are used for up and down electrons)
* Some tests for the global rotation code

- New feature

- No

desktop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/RotatedSPOs.cpp
+++ b/src/QMCWaveFunctions/RotatedSPOs.cpp
@@ -92,7 +92,7 @@ void RotatedSPOs::extractParamsFromAntiSymmetricMatrix(const RotationIndices& ro
   }
 }
 
-void RotatedSPOs::resetParametersExclusive(const opt_variables_type& active)
+void RotatedSPOs::resetParametersExclusive(const opt_variables_type& active, bool isPrimaryObject)
 {
   std::vector<RealType> delta_param(m_act_rot_inds.size());
 
@@ -119,7 +119,7 @@ void RotatedSPOs::resetParametersExclusive(const opt_variables_type& active)
     for (int i = 0; i < m_full_rot_inds.size(); i++)
       old_param[i] = myVarsFull[i];
 
-    if (use_this_copy_to_apply_rotation_)
+    if (isPrimaryObject)
       applyDeltaRotation(delta_param, old_param, new_param);
 
     // Save the the params
@@ -128,7 +128,7 @@ void RotatedSPOs::resetParametersExclusive(const opt_variables_type& active)
   }
   else
   {
-    if (use_this_copy_to_apply_rotation_)
+    if (isPrimaryObject)
       apply_rotation(delta_param, false);
 
     // Save the parameters in the history list
@@ -1409,8 +1409,6 @@ std::unique_ptr<SPOSet> RotatedSPOs::makeClone() const
   myclone->myVarsFull      = this->myVarsFull;
   myclone->history_params_ = this->history_params_;
   myclone->use_global_rot_ = this->use_global_rot_;
-
-  // use_this_copy_to_apply_rotation_ is deliberately not copied
 
   return myclone;
 }
--- a/src/QMCWaveFunctions/RotatedSPOs.cpp
+++ b/src/QMCWaveFunctions/RotatedSPOs.cpp
@@ -88,7 +88,7 @@ void RotatedSPOs::extractParamsFromAntiSymmetricMatrix(const RotationIndices& ro
                                                        std::vector<ValueType>& param)
 {
   assert(rot_indices.size() == param.size());
-  // Assumes rot_mat is of the correct size and initialized to zero upon entry
+  // Assumes rot_mat is of the correct size upon entry
 
   for (int i = 0; i < rot_indices.size(); i++)
   {
@@ -126,7 +126,7 @@ void RotatedSPOs::resetParametersExclusive(const opt_variables_type& active)
       old_param[i] = myVarsFull[i];
 
     if (use_this_copy_to_apply_rotation_)
-      apply_delta_rotation(delta_param, old_param, new_param);
+      applyDeltaRotation(delta_param, old_param, new_param);
 
     // Save the the params
     for (int i = 0; i < m_full_rot_inds.size(); i++)
@@ -178,7 +178,7 @@ void RotatedSPOs::buildOptVariables(const RotationIndices& rotations, const Rota
   const size_t nmo = Phi->getOrbitalSetSize();
 
   // create active rotations
-  m_act_rot_inds  = rotations;
+  m_act_rot_inds = rotations;
 
   if (use_global_rot_)
     m_full_rot_inds = full_rotations;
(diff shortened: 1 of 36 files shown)


src/QMCWaveFunctions/SPOSet.h
```diff
@@ -117,7 +117,9 @@ class SPOSet : public QMCTraits
   // Single Slater creation
   virtual void buildOptVariables(const size_t nel) {}
   // For the MSD case rotations must be created in MultiSlaterDetTableMethod class
-  virtual void buildOptVariables(const std::vector<std::pair<int, int>>& rotations) {}
+  virtual void buildOptVariables(const std::vector<std::pair<int, int>>& rotations,
```
ye-luo, line 120: In what scenario, rotations and full_rotations are different?
I'd like to get rid of the two buildOptVariables interfaces in SPOSet.h because it makes sense only to RotatedSPO.
Instead, in the current use of this API, we can dynamic_cast and figure out if RotatedSPO is used.
Let us explore changing this in upstream first.

markdewing: The only case where rotations and full_rotations are the same is with 1 electron and 2 orbitals - that leads to 1 parameter and a 2x2 rotation matrix.  I think they are different in all other cases - there are multiple occupied orbitals and/or multiple unoccupied orbitals and the rotations internal to each category are not listed in "rotations", but need to be present in "full_rotations".

ye-luo: Oh. Need to document both arguments which are very confusing by just looking at names.

ye-luo:
I added an issue discussing the input change for orbital rotation. https://github.com/QMCPACK/qmcpack/issues/4428

src/QMCWaveFunctions/RotatedSPOs.h
```diff
@@ -203,28 +226,15 @@ class RotatedSPOs : public SPOSet, public OptimizableObject
...
-  {
-    std::vector<RealType> param(m_act_rot_inds.size());
-    for (int i = 0; i < m_act_rot_inds.size(); i++)
-    {
-      int loc  = myVars.where(i);
-      param[i] = myVars[i] = active[loc];
-    }
-    apply_rotation(param, true);
-  }
+  void resetParametersExclusive(const opt_variables_type& active) override;
```
PDoakORNL, line 237: So I guess this touches something that shares data with other threads/spos?  Could use some documentation.  Otherwise I just don't know what exclusive means.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -19,7 +19,13 @@
 namespace qmcplusplus
 {
 RotatedSPOs::RotatedSPOs(const std::string& my_name, std::unique_ptr<SPOSet>&& spos)
-    : SPOSet(my_name), OptimizableObject(my_name), Phi(std::move(spos)), nel_major_(0), params_supplied(false)
+    : SPOSet(my_name),
+      OptimizableObject(my_name),
+      Phi(std::move(spos)),
+      nel_major_(0),
+      params_supplied(false),
+      use_this_copy_to_apply_rotation_(false),
```
PDoakORNL, line 27: Why not just initialize this in the class declaration? The reason to do it here would be if it was dependent on a constructor argument.

src/QMCWaveFunctions/RotatedSPOs.h
```diff
@@ -340,11 +350,26 @@ class RotatedSPOs : public SPOSet, public OptimizableObject
...
+  void set_use_global_rotation(bool use_global_rotation) { use_global_rot_ = use_global_rotation; }
+
 private:
   /// true if SPO parameters (orbital rotation parameters) have been supplied by input
   bool params_supplied;
   /// list of supplied orbital rotation parameters
   std::vector<RealType> params;
+
+  /// Flag to ensure rotation is applied to shared coefficients only on one thread
+  bool use_this_copy_to_apply_rotation_;
```
PDoakORNL, line 363: So if they are only applied on one thread what is the synchronization mechanism to ensure they aren't used on other threads before they are updated or during the update?  Assuming you'll be covered by some other rank scope synchronization then that needs to be documented.

I guess a future PR will make this clear?

PDoakORNL (changes requested):
A small change or two and concurrency design question that may need to be tabled.  Also I'm probably forgetting but are we going to really carry two methods of doing this one indefinitely or is the historylist likely to get cut?

markdewing:
@ye-luo  I can split this into smaller PR's if that would be better.

src/QMCWaveFunctions/RotatedSPOs.h
markdewing: It's documented in OptimizableObject.h.

src/QMCWaveFunctions/RotatedSPOs.cpp
markdewing: Habit. Will change it.

ye-luo:
> @ye-luo I can split this into smaller PR's if that would be better.

I'm thinking of
1) Remove buildOptVariables in SPOSet.h
2) update LCAO
3) update SpineR2R
4) adjust input, need to settle #4428
5) update RotatedSPO.

Try to maximize concurrency if dependency is satisfied.
4) may take a bit long
Can 1,2,3 PRed independently?
5 probably needs to wait for 1,2,3

src/QMCWaveFunctions/RotatedSPOs.h
PDoakORNL: Ok I'm was just confused by exclusive not having read that recently.

src/QMCWaveFunctions/RotatedSPOs.h
markdewing: resetPsi explicitly iterates over all the clones - which is why no lock is needed, and why some solution like this flag is needed.  Another possible solution would be to add a flag to resetParameterExclusive to indicate the primary copy is being updated vs. one of the clones.

markdewing: I updated the PR to use an alternate method of passing a flag (isPrimaryObject) through resetParameterExclusive.

markdewing: @PDoakORNL and @ye-luo : Could I get your feedback on this latest commit (using the isPrimaryObject instead of use_this_copy... ).  I can put it in a separate PR for actual check-in.

ye-luo: I prefer neither isPrimaryObject nor use_this_copy but to get rid of this line.
https://github.com/QMCPACK/qmcpack/blob/608db3e62129a0c3c7cdbc8db7be7ace282f747a/src/QMCDrivers/WFOpt/QMCCostFunctionBatched.cpp#L444
Which needs fixing in a few optimizable components. All the optimization related operation should be done only through the gold copy. Updated parameters should be propagated automatically via shared_ptr among clones.

PDoakORNL: I'd prefer this design direction as well.  I'm not fond of the shared pointer, from my perspective there isn't any ambiguity, the MultiBSplineSet or whatever basis is owned by the gold SPOSet. All the other SPOSet's should have lifetimes within the enclosing scope of the gold SPOSet lifetime. We don't need to do reference counting. The shared pointer is a quick workaround, fine for now but not ideal.

It doesn't really express the semantics and so doesn't indicate the undesirability of updating that pointed to object from what is basically a thread scoped object.

markdewing:
Needs more redesign, I guess...

PR #4430, opened 2023-01-31
Set development version 3.16.9

prckent:
## Proposed changes

Set development version 3.16.9 per release process https://github.com/QMCPACK/qmcpack/wiki/Release-process


prckent:
I'll use admin privs to merge this - we don't need the full CI to run.

PR #4437, opened 2023-02-03
Fix orbital rotation RHF parameter derivatives

markdewing:
Accumulate the derivatives in the case where multiple wavefunction pieces are controlled by the same variational parameter.

The fixes the single determinant case.  The multi-determinant case should work as it uses accumulation already.

The input file has no initial rotation (no "opt_vars" elements).  These don't work correctly in the RHF case, as the rotation ends up being applied twice.  (I don't think there's an easy way to fix this, and it doesn't seem like an important case to fix.  For normal usage, initial rotations will be read in from a VP file.)

Some of the other single-determinant derivative functions related to ECP's aren't changed, and I don't know if they need to be - will need to look into that separately.

- Bugfix

- No

desktop

## Checklist

### DIFF
--- a/tests/molecules/He_param/He_orb_rot_param_grad_rhf.xml
+++ b/tests/molecules/He_param/He_orb_rot_param_grad_rhf.xml
@@ -1,0 +1,90 @@
+<?xml version="1.0"?>
+<simulation>
+  <project id="He_orb_rot_param_grad_rhf" series="0">
+         <parameter name="driver_version">batch</parameter>
+  </project>
+
+  <!-- Location of atoms -->
+
+  <particleset name="ion0" size="1">
+    <group name="He">
+      <parameter name="charge">2</parameter>
+    </group>
+    <attrib name="position" datatype="posArray">
+      0.0 0.0 0.0
+    </attrib>
+  </particleset>
+
+  <!-- Randomly create electrons around the atomic position -->
+
+  <particleset name="e" random="yes" randomsrc="ion0">
+    <group name="u" size="1">
+      <parameter name="charge">-1</parameter>
+    </group>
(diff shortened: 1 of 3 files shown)


markdewing:
Test this please

ye-luo (approved):

tests/molecules/He_param/He_orb_rot_param_grad_rhf.xml
```diff
@@ -0,0 +1,90 @@
...
+          <coefficient id="updetC" type="Array" size="2">
+            1.0 0.0
+            0.0 1.0
+          </coefficient>
+      </sposet>
+    </sposet_collection>
+    <determinantset type="MO" key="STO" transform="no" source="ion0">
+      <slaterdeterminant>
+        <determinant id="spo-up" spin="1" size="2"/>
+        <determinant id="spo-up" spin="-1" size="2"/>
```
ye-luo, line 54: I think there are certain consistency issues. It is problematic to let the determinant know there are two orbitals beneath RotatedSPO. That should be a detail internal to RotatedSPO. We will clean up this mess as we change input xml.

prckent: Is there no requirement for unique IDs? We should fix that if not.

prckent (approved):

tests/molecules/He_param/He_orb_rot_param_grad_rhf.xml
ye-luo: id was from old input style to name sposets when they were built inside determinants. I don't think any of `id`, `spin`, `size` are actually effective.

PR #4439, opened 2023-02-04
Refactor low level routines used by Spline SPO offload

ye-luo:
## Proposed changes
Preparation for more re-use of routines.

- No

epyc-server

## Checklist


prckent (approved):
Hopefully this gets through CI.

SoAFields3D should be called something like splinegradients in a later revision

Please comment on status with llvm and aomp offload in subsequent PRs. I assume you are checking and it would be good to note that.

ye-luo:
Test this please

ye-luo:
> SoAFields3D

I still need to get it moved to another place, perhaps a separate header file.

PR #4443, opened 2023-02-07
Fix spline block offset caused by #4439

ye-luo:
## Proposed changes
Affecting >512 splines. Caught by NiO a64 in complex build. 384 complex orbitals, 768 splines.

https://cdash.qmcpack.org/CDash/testDetails.php?test=25095055&build=396449
The error message is invalid memory access but the root cause was bad orbital values caused NaN in electron positions.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/SplineC2COMPTarget.cpp
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineC2COMPTarget.cpp
@@ -160,7 +160,7 @@ void SplineC2COMPTarget<ST>::evaluateDetRatios(const VirtualParticleSet& VP,
 
         PRAGMA_OFFLOAD("omp parallel for")
         for (int index = 0; index < last - first; index++)
-          spline2offload::evaluate_v_impl_v2(spline_ptr, ix, iy, iz, index, a, b, c,
+          spline2offload::evaluate_v_impl_v2(spline_ptr, ix, iy, iz, first + index, a, b, c,
                                              offload_scratch_iat_ptr + first + index);
         const size_t first_cplx = first / 2;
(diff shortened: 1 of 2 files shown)


ye-luo:
Test this please

prckent:
The lack of unit tests for large electron counts/basis sizes is something that I have been concerned about. Can you expand the unit tests in test_MO.cpp to exclude one with number of splines larger than all of the chunk sizes/teaming sizes? Say 997 splines or some slightly awkward prime? Such a test likely would have caught this issue.

ye-luo:
test_MO.cpp is LCAO. The lack of faking spline orbitals is an issue.

prckent:
I didn't find a test_spline equivalent to test_MO, although there are spline tests elsewhere. We could use fake spline data or even real data for a small system that is replicated so that there are sufficient orbitals.  Both would have caught this problem since the expected orbital values would not have been obtained.

prckent:
The LCAO code could have the same issues, either already or in the future.

ye-luo:
> I didn't find a test_spline equivalent to test_MO, although there are spline tests elsewhere. We could use fake spline data or even real data for a small system that is replicated so that there are sufficient orbitals. Both would have caught this problem since the expected orbital values would not have been obtained.

I don't have a way in mind to replicate small system orbitals to a large system. The current bug can be viewed as repeating orbitals beyond 512.

Fake data is more workable and it is also useful to create benchmarks rather than relying on large h5 files.

prckent:
>Fake data is more workable and it is also useful to create benchmarks rather than relying on large h5 files.
Agreed

prckent (approved):

PR #4451, opened 2023-02-09
Cleanup ForceBase and BareForce classes

PhilipFackler:
## Proposed changes

Renamed and made private/protected all data members
Added necessary get/set functions
Refactored functions from h to cpp
Removed unused code

- No

Ubuntu 22.04

## Checklist


src/QMCHamiltonians/QMCHamiltonian.h
```diff
@@ -364,7 +364,7 @@ class QMCHamiltonian
   bool has_L2() { return l2_ptr != nullptr; }
 
   /** compute D matrix and K vector for L2 potential propagator
-    * @param r single particle coordinate
+    * @param r single fffficle coordinate
```
markdewing, line 367: particle

PhilipFackler: Not sure how that happened! Fixed it!

src/QMCHamiltonians/ForceCeperley.cpp
```diff
@@ -216,9 +216,9 @@ std::unique_ptr<OperatorBase> ForceCeperley::makeClone(ParticleSet& qp, TrialWav
...
 //    ///cerr << "ForceBase::setObs storing forces";
 //    int index = myIndex;
-//    for(int iat=0; iat<Nnuc; iat++) {
+//    for(int iat=0; iat<n_nuc_; iat++) {
 //      for(int x=0; x<OHMMS_DIM; x++) {
-//        plist[index] = forces(iat,x) + storeForces(iat,x);
+//        plist[index] = forces_(iat,x) + storeForces(iat,x);
 //        //cerr << " " << index << ":" << plist[index];
 //        index++;
 //      }
```
williamfgc, line 224: Since these are commented out blocks it might be worth cleaning as well.

src/QMCHamiltonians/BareForce.h
```diff
@@ -0,0 +1,58 @@
...
+#include "QMCHamiltonians/ForceBase.h"
+#include "QMCHamiltonians/OperatorBase.h"
+
+namespace qmcplusplus
+{
+struct BareForce : public OperatorBase, public ForceBase
+{
+public:
+  BareForce(ParticleSet& ions, ParticleSet& elns);
+  std::string getClassName() const override;
```
williamfgc, line 29: For these simple getters and setters I tend to use `noexcept` , but this is an inherited function as it could be part of future PRs.

williamfgc (changes requested):
Thanks Philip, just minor request to remove commented out code

src/QMCHamiltonians/ForceCeperley.cpp
PhilipFackler: Done

src/QMCHamiltonians/BareForce.h
PhilipFackler: Are you saying you would add `noexcept` even though it's a virtual function override?

williamfgc (approved):

src/QMCHamiltonians/BareForce.h
williamfgc: That adding `noexcept` would need to be done at the base class level so it will propagate, but in a future PR.

williamfgc:
Test this please

src/QMCHamiltonians/ForceBase.h
```diff
@@ -79,45 +57,50 @@ struct ForceBase
...
-public:
-  BareForce(ParticleSet& ions, ParticleSet& elns);
-  std::string getClassName() const override { return "BareForce"; }
-  void resetTargetParticleSet(ParticleSet& P) override;
+  const ParticleSet::ParticlePos& getForces() const noexcept;
+  void setForces(const ParticleSet::ParticlePos& forces);
+  void setForces(Real val);
 
-  Return_t evaluate(ParticleSet& P) override;
+  const ParticleSet::ParticlePos& getForcesIonIon() const noexcept;
```
ye-luo, line 68: To my compiler knowledge, having accessor functions defined in the header files allows being in-lined by compilers. Usually they are just simple return and having them defined in the header file helps readability.
I'm not aware of any advantages of having the definition in cpp.

prckent: Simple accessor functions should be kept in header files for this reason. 

[ The source base has a good number of large functions headers that could be relocated, but small and potentially important-for-performance accessors should be kept. We do not want to have to rely on e.g. LTO for efficiency.]

prckent (approved):

src/QMCHamiltonians/ForceBase.h
williamfgc: My two cents is that in general inlining is not good for caching and bloats the executable, keeping implementation in ccp make translation units confined without compromising performance. Not a problem for QMCPACK users, but for developers modifying header implementations bloat compilation times, depending on the header is not uncommon that ~ 300 targets need to be rebuilt adding little benefit.

prckent: Happily, should we arrive at a debate, this is testable.

ye-luo: > My two cents is that in general inlining is not good for caching and bloats the executable, keeping implementation in cpp make translation units confined without compromising performance. Not a problem for QMCPACK users, but for developers modifying header implementations bloat compilation times, depending on the header is not uncommon that ~ 300 targets need to be rebuilt adding little benefit.

True. I only expect accessor functions defined in header files
```
const ParticleSet::ParticlePos& getForcesIonIon() const noexcept
{
  return forces_ion_ion_; // return a member.
}

ye-luo: Please put all the get function definition in the header file and I will approve this PR.

williamfgc: One liners should be fine, not sure if the time critical [aspect](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f5-if-a-function-is-very-small-and-time-critical-declare-it-inline) applies, but readability is a good point.

PR #4463, opened 2023-02-14
Add delta rotation functions for orbital rotation

markdewing:
Add a function to compute the new rotation matrix (and corresponding parameters). 
This also needs a function to compute the indices for the full rotation matrix.

(At one time during the implementation of orbital rotation, it was important that the first part of the list of full rotation indices corresponded to the other list of rotation indices.  That might have been due to incorrect code.  I think with the correct code the ordering shouldn't matter.)

These will eventually support the global rotation method for orbital rotation optimization.

- New feature
:

- No

desktop

## Checklist


prckent:
Test this please

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -40,6 +40,25 @@ void RotatedSPOs::createRotationIndices(int nel, int nmo, RotationIndices& rot_i
...
+
+  // Add core-core rotations - put them at the end of the list
+  for (int i = 0; i < nel; i++)
+    for (int j = i + 1; j < nel; j++)
+      rot_indices.emplace_back(i, j);
+
+  // Add active-active rotations - put them at the end of the list
+  for (int i = nel; i < nmo; i++)
+    for (int j = i + 1; j < nmo; j++)
+      rot_indices.emplace_back(i, j);
```
ye-luo, line 59: Is it possible to precompute the size of rot_indices and reserve it? I feel nmo * nel scale emplace_back can be quite costly due to resize.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -172,6 +191,41 @@ void RotatedSPOs::apply_rotation(const std::vector<RealType>& param, bool use_st
...
+  assert(old_param.size() == full_rot_inds.size());
+  assert(new_param.size() == full_rot_inds.size());
+
+  const size_t nmo = new_rot_mat.rows();
+  assert(new_rot_mat.rows() == new_rot_mat.cols());
+
+  ValueMatrix old_rot_mat(nmo, nmo);
+  old_rot_mat = ValueType(0);
+
+  constructAntiSymmetricMatrix(full_rot_inds, old_param, old_rot_mat);
```
ye-luo, line 211: Is the set 0 necessary? constructAntiSymmetricMatrix doesn't sound like need of setting zero before calling it.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -172,6 +191,41 @@ void RotatedSPOs::apply_rotation(const std::vector<RealType>& param, bool use_st
...
+  constructAntiSymmetricMatrix(act_rot_inds, delta_param, delta_rot_mat);
+  exponentiate_antisym_matrix(delta_rot_mat);
+
+  // Apply delta rotation to old rotation.
+  BLAS::gemm('N', 'N', nmo, nmo, nmo, 1.0, delta_rot_mat.data(), nmo, old_rot_mat.data(), nmo, 0.0, new_rot_mat.data(),
+             nmo);
+
+  ValueMatrix log_rot_mat(nmo, nmo);
+  log_rot_mat = new_rot_mat;
+  log_antisym_matrix(log_rot_mat);
```
ye-luo, line 226: Can this function do out-of-place transformation?

src/QMCWaveFunctions/RotatedSPOs.cpp
markdewing: This function isn't going to run that frequently, and the resize algorithm should double the capacity every time it runs out, so this really shouldn't be a big issue.

But yes, the size can be precomputed, and it should be nmo*(nmo-1)/2.

src/QMCWaveFunctions/RotatedSPOs.cpp
markdewing: constructAntiSymmetric matrix does need the input matrix to be zero.  Not all entries are referenced, and the unreferenced entries will not get set by the function.
It looks like ValueMatrix uses Vector as an underlying container and that will set the value to zero on initialization/resize.

src/QMCWaveFunctions/RotatedSPOs.cpp
markdewing: It could be modified to do the transformation out-of-place.

prckent:
Happy to merge this since getting to usable-for-science oo in the batch code is such a priority.

src/QMCWaveFunctions/RotatedSPOs.cpp
ye-luo: `unreferenced entries will not get set by the function.` that means `constructAntiSymmetric` assumes the previous value should be kept untouched. Is a desired safe behavior? I'm wondering if constructAntiSymmetric should set 0 internally. Whether ValueMatrix set 0 or not at initialization is not relevant when we define the behavior of constructAntiSymmetric I assume.

markdewing: There's a comment in constructAntiSymmetricMatrix that states the assumption that the rotation matrix is initialized to zero upon entry.   Based on your other review comments on this PR, I got the impression you were concerned about performance implications, and in this case it is redundantly setting the matrix to zero.    But constructAntiSymmetricMatrix could set the array to zero to ensure correctness.

ye-luo: In this case, correctness is favored.

prckent:
Test this please

prckent (approved):
Concerns addressed.

Ye: Cancel the automerge if you have additional comments.

ye-luo:
> Concerns addressed.
> 
> Ye: Cancel the automerge if you have additional comments.

@markdewing addressed all my concern.

PR #4468, opened 2023-02-15
Used shared pointer for LCAO coefficient matrix copy

markdewing:
A copy of the coefficient matrix is used for orbital rotation.
Change the copy to use a shared pointer.  This means the rotation only needs to be performed on one thread (and should be performed only on one thread).

The storage for the coefficient copy is used regardless of the setting of `use_stored_copy`. That means `storeParamsBeforeRotation` needs to be called once sometime before `applyRotation` is called.

The `use_stored_copy` flag changes which values get used - true uses the values in the copy and false uses the values in the existing coefficients.

- New feature

- No

desktop

## Checklist


markdewing:
Test this please

prckent (approved):

PR #4469, opened 2023-02-16
Store copy of bspline coefficients for orbital rotation

markdewing:
This is the bspline version of #4468.

Bsplines did not previously keep a copy of the coefficients.   The copy of the coefficients performs two functions. First it allocates storage for the matrix multiplication in applyRotation (it replaces `new_coefs` for this purpose). Second, it allows using the global rotation method with bsplines.

- New feature

- No

desktop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
@@ -51,6 +51,16 @@ bool SplineR2R<ST>::write_splines(hdf_archive& h5f)
   return h5f.writeEntry(bigtable, o.str().c_str()); //"spline_0");
 }
 
+template<typename ST>
+void SplineR2R<ST>::storeParamsBeforeRotation()
+{
+  const auto spline_ptr     = SplineInst->getSplinePtr();
+  const auto coefs_tot_size = spline_ptr->coefs_size;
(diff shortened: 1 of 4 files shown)


markdewing:
Test this please

prckent (approved):
Couple of points as this enters into production:
* The coeffs copy is potentially huge. Some extra error handling on the allocation and informative messages will be needed.
* Should print a note of the size of the coeffs copy in all cases.

ye-luo:
@markdewing did you get a chance to build an abstract class for the all the sposet supporting rotation as we talked last friday?

prckent:
Ideally we would just apply a rotation matrix to the calculated sposets in a completely abstract manner, which would allow any sposet to be optimized (e.g. the hybrid basis, plane waves, something new). However I assume the current route will get us something working most quickly, which is the key priority currently. We can't learn about the state of the optimizer or do science (priority) until at least one complete set of pipes is in.

markdewing:
I did not try the abstract class.
I did try moving to using makeIndependentCopy to keep a copy of the SPO.  There were two issues.  First, the spline classes were very resistant to being copied (mostly they didn't keep around the grid and boundary condition info necessary to recreate the class).  But I finally got that working.   Secondly, the initial code I wrote using the SPO copies didn't work.  And I didn't want to spend time debugging it.
I think the existing interface is good enough for now.

ye-luo:
OK. We can explore such an abstraction after what you have gets in.

PR #4440, opened 2023-02-06
Construct crowd data before each use in the batch optimizer

markdewing:
For batched optimization, construct the crowd data in correlatedSampling in addition to checkConfigurations. The previous code would construct it in checkConfigurations and reuse the crowd data in correlatedSampling.
This simplifies changes to the variational parameters in the wavefunction, as the reset parameters call only needs to be performed on the gold copy.

Suggested in https://github.com/QMCPACK/qmcpack/pull/4426#discussion_r1095028360

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Refactoring (no functional changes, no api changes)

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
desktop

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- No. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #4470, opened 2023-02-16
Enable orbital rotation

markdewing:
Change how rotation interacts with the optimizer to enable orbital rotation.

The method for tracking the rotation - history list versus global rotation - uses the "method" attribute on the "rotated_spo" tag.  The choices are "global" and "history", and the default is "global".

Note that the history method with the legacy driver and more than one thread does not work.  I couldn't figure out how to easily get all the information in one place to make the check and throw an error.  Given the default method is global rotation, and that does work with the legacy driver and multiple threads, this is going to take some effort to encounter.
* Batched driver
   * With the changes in #4440, the batched driver only ever operates on one copy.
* Legacy driver
   * global rotation - the state is kept in the RotatedSPOs, so applying resetParametersExclusive multiple times results in the same rotated coefficients.
   * history list - the state is kept in the coefficients of the underlying SPOSet, so the rotation gets applied once per-thread.

The rationale for keeping both methods around:
 * The history list is conceptually simpler and is more likely to be correct (though it might have numerical issues if the spline coefficients are kept as floats)
 * The matrix exponential uses an N^3 eigenvalue decomposition.  I expect this to be one of the first performance issues encountered when scaling up the number of rotation parameters.  The matrix log also uses an eigenvalue decomposition.  So the history list method uses one eigenvalue decomposition per applied rotation, where the global method uses two. 
 * The global rotation method is convenient because there is always just one set of rotation parameters to describe the state.
 * I suspect one future option for performance (or numerics) is to use the history method during the optimization run, but maintain the rotation matrix separate from the coefficients.  Then the matrix log can be performed once at the end to recover a single set of rotation parameters.

The previous code in checkInVariablesExclusive and resetParametersExclusive computed the delta rotation by setting the parameters to 0 in checkInVariablesExclusive.  Then when the driver calls resetParametersExclusive, the passed parameters represent the change in rotation.  The problem is the code is not guaranteed to pair every call to resetParametersExclusive with a previous call to checkInVariablesExclusive.  For instance, when computing the gradient, or when using multiple threads.

The new code tracks the current values of the parameters in myVars and computes the difference on entry to resetParameters.  As long as the parameters in myVars in RotatedSPOs match the values the optimizer has in currentParameters, the change in rotation will be computed correctly.

- New feature

- No

desktop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.cpp
+++ b/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.cpp
@@ -790,6 +790,7 @@ void MultiDiracDeterminant::buildOptVariables(std::vector<size_t>& C2node)
 
   // create active rotation parameter indices
   std::vector<std::pair<int, int>> m_act_rot_inds;
+  std::vector<std::pair<int, int>> other_rot_inds;
 
   for (int i = 0; i < nmo; i++)
     for (int j = i + 1; j < nmo; j++)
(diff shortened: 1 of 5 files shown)


prckent:
Thanks Mark.

An aside for any potential orbital optimization users: we emphasize that the legacy cpu driver implementations only exist to help the developers mature the implementations in the batched drivers and prove out that they work. *we are going to delete or wall off the legacy drivers very very soon*. So if the batch driver implementation does not work, the correct course of action is to get them fixed or enhanced, not to try the legacy drivers.

prckent:
Mark - Based on the codecov results it looks like we are missing some coverage of the history method. Can you add some?

prckent:
Test this please

prckent (approved):

PR #4474, opened 2023-02-20
Add Magnetization Density Input Class

rcclay:

## Proposed changes

This adds a MagnetizationDensityInput class, which handles file parsing and input validation.  Uses SpinDensityInput and OneBodyDensityMatricesInput as templates for a gridded estimator with additional control over spin integration.  

- New feature

- No

Intel Xeon, RHEL7
## Checklist


ye-luo:
Could you document this estimator in the manual? what is this estimator, how does the input block look like, what are the options.

prckent:
Can you state here what this will allow us to delete in future? e.g. I have never heard a valid reason for why historically we have had multiple density-like estimators.

rcclay:
> Could you document this estimator in the manual? what is this estimator, how does the input block look like, what are the options.

I'm planning to put the documentation in when the estimator itself is PR'd and testable.  This is just the input class/parser.  I can add documentation now if you prefer, but I'd rather not advertise it in the manual until it's functional.

PDoakORNL:
Well It's my intention that the `DensityEstimator` never get ported to the batched version. It seems to provide nothing that summing over the spin density estimator doesn't except a memory savings. I explored other ways of saving memory in the SpinDensity operator that can be further pursued if there is ever an issue there. The EnergyDensityEstimator warrants a separate "estimator."  

But there doesn't need to be a one to one relationship between input classes and estimators.  From a validation standpoint it may make sense to have split input classes even if the simulation class is more generic.

  I'll have time to do an actual review for this PR this afternoon.

rcclay:
> Can you state here what this will allow us to delete in future? e.g. I have never heard a valid reason for why historically we have had multiple density-like estimators.

At the moment, this doesn't help or hurt the code over current state.  Could be straightforward to merge this with other density-like classes, but probably by doing "if option A, do this, if option B, do that, etc."  With a bit more thinking, a better way is probably to specify a real space grid and all related handling, and an operation mapping R, H(R), psi(R) -> value, (x,y,z).  So then the difference between spin density, total density, magnetization density, etc., would just be in defining how that reduction step happens--how to determine a grid point, what data values and types get assigned to it, etc.

src/Estimators/MagnetizationDensityInput.h
```diff
@@ -0,0 +1,129 @@
...
+class MagnetizationDensityInput
+{
+public:
+  enum class Integrator
+  {
+    SIMPSONS,
+    MONTECARLO
+  };
+
+  inline static const std::unordered_map<std::string, std::any> lookup_input_enum_value{{"integrator-simpsons",
```
PDoakORNL, line 33: clang-format?  or is responsible for this?

src/Estimators/MagnetizationDensityInput.cpp
```diff
@@ -0,0 +1,119 @@
...
+  TinyVector<int, DIM> grid;
+
+  //dr or grid must be specified to perform the grid math.  Input already checked before we get here.  
+  if (have_dr_)
+    for (int d = 0; d < DIM; ++d)
+    {
+      grid[d] = (int)std::ceil(std::sqrt(dot(lattice.Rv[d], lattice.Rv[d])) / dr_[d]);
+      //Now we know the lattice and can compute grid points from dr.  We check it's validity here.
+      if (grid[d] < 1)
+        throw UniformCommunicateError(
```
PDoakORNL, line 52: This can happen I think only for a negative dr value and the `checkParticularValidity` means this can never happen.  Am I wrong?  If I'm not I'd just drop the check, and add a comment to this function that it has a precondition that all `dr_  > 0`

PDoakORNL (dismissed):
This looks good but I think the coverage issue can be fixed if you want a perfect score.

src/Estimators/MagnetizationDensityInput.h
rcclay: clang-format is responsible for that.  Improper application of clang-format not ruled out.

src/Estimators/MagnetizationDensityInput.cpp
rcclay: I agree with that.  I'll remove that redundant check.

src/Estimators/MagnetizationDensityInput.h
rcclay: Double checked the clang-format output.  Running from src, that's the output that should be consistent with .clang-format.  I can manually format it if you like.

PDoakORNL (approved):
LGTM
Future work could look at merging some/all of the SpinDensityInput and this Input. Although I would rather see two classes and some repeated boiler plate than a input class with a bunch of if spin if magnetization branching.

rcclay:
@ye-luo Did you want the documentation added to this PR or to the actual magnetization density PR?

ye-luo:
Please update the manual in this PR. It helps reviewing the added code (input parsing).

rcclay:
@ye-luo Documentation added.

ye-luo:
Test this please

ye-luo (approved):

Issue #4099, opened 2022-07-06
Unify compile-time complex and real builds as run-time options in a single build


williamfgc:
**Is your feature request related to a problem? Please describe.**

Currently QMCPACK separates `real` and `complex` functionality at compile-time. Thus, stakeholders need to create two separate executables to test, deploy and develop functionality.

**Describe the solution you'd like**
The long term goal is to unify these real/complex compile-time build options into a single build that can take equivalent runtime options.

**Describe alternatives you've considered**
Since these involve invasive changes in the entire codebase tackling compile-time conditionals  e.g. `#ifdef QMC_COMPLEX`, there are a few steps that need to be accounted for.  If possible, break down the effort into smaller tasks. 

- [x] Explore how many places need to be modified. This [gist](https://gist.github.com/williamfgc/28ffbd23f6888ed89530e3f68c992c93) contains information extracted for `#ifdef QMC_COMPLEX`
- [ ] Initially replace `#ifdef QMC_COMPLEX` by a proper variable at the `qmcapp.cpp` main level and propagate through the `qmcplusplus` namespace. This should not be an invasive and breaking change.
- [ ] Allow the runtime option at the top-most level (e.g. configuration xml file header)
- [ ] Explore more invasive changes: at the template level in QMCPACK internal functionality and CMake's build system configuration level
- [ ] Explore testing options (separate real and complex tests).

**Additional context**
This issue tracks efforts and discussion related to this integration.

prckent:
I suggest to add the runtime option next. Then adapt code starting at the lowest levels of the app that connects to it, as per the other tasks.

There are many `if (QMC_COMPLEX)` in the cmake test config that can be converted to dedicated complex mode tests.

williamfgc:
@prckent @ye-luo what would be a good place for a runtime configuration option tracking real, complex, and precision (full, mixed): 

- [ProjectData](https://github.com/QMCPACK/qmcpack/blob/develop/src/Utilities/ProjectData.h) . This is a member of [QMCAppBase](https://github.com/QMCPACK/qmcpack/blob/develop/src/QMCApp/QMCAppBase.h#L67)
- [QMCState](https://github.com/QMCPACK/qmcpack/blob/develop/src/Utilities/qmc_common.h)
- [QMCMainState](https://github.com/QMCPACK/qmcpack/blob/develop/src/QMCApp/QMCMainState.h)
- Other

prckent:
[...]

PR #4480, opened 2023-02-23
Add ProjectData ref member to WaveFunctionPool from QMCMain

williamfgc:
## Proposed changes

Propagate `ProjectData` information from top QMCMain to WaveFunctionPool as a const references member
Needed to pass complex/real info at runtime
Modified unit tests accordingly which is most of the current PR work
Related to #3411 #4099

- No

Ubuntu 22, must pass CI variants

## Checklist


ye-luo:
If passing `complex/real` is the need, please pass a bool not the whole ProjectData.

williamfgc:
@ye-luo just thinking ahead if we move more compile-time configurations to runtime variants to ProjectData: e.g. Mixed precision, AFQMC (?). It's also currently done for `QMCDriver`.

ye-luo:
QMCDriver depends on ProjectData because ProjectData holds the counter of QMC runs 000, 001. It is different story from all the pools.

prckent:
The two options intended to be made compile time are QMC_COMPLEX and QMC_MIXED_PRECISION, with QMC_COMPLEX being the most useful and requiring the most work, particularly around the spline handling. AFQMC is a feature flag not a numerics flag and will remain a compile time option.

ye-luo:
We could create a struct with all the flags inside ProjectData and pass it to pools instead of the whole ProjectData.

williamfgc:
I've been thinking more about options and one could be just having two bool members (the scope) as part of the pool classes to keep it simple. ProjectData will only be at the Pool constructor level so we know it can only come from it Thoughts? 

```
WavefunctionPool(const ProjectData& project_data, ...) : is_complex_(project_data.IsComplex() ), is_half_precision_(project_data.isHalfPrecision(), ...
```

ye-luo:
With a struct for flags, we don't need change these APIs again when adding or removing flags happens.
The global mixed precision control may go away.
PS: it is called mixed precision not half precision. Also make sure make the stored flags in pools marked const.

williamfgc:
> With a struct for flags, we don't need change these APIs again when adding or removing flags happens.
The struct would be something like `struct RuntimeOptions { bool is_complex_; bool is_mixed_precision_; }`?  Would a `std::array<bool, 2>` be sufficient? Technically, break API but these are not going to change very often (if ever). Thoughts?

ye-luo:
> > With a struct for flags, we don't need change these APIs again when adding or removing flags happens.
> 
> The struct would be something like `struct RuntimeOptions { bool is_complex_; bool is_mixed_precision_; }`? Would a `std::array<bool, 2>` be sufficient? Technically, breaks API due to the number of elements, but these are not going to change very often (if ever). Thoughts?

`struct RuntimeOptions { bool is_complex_; bool is_mixed_precision_; }` is my choice. With `std::array<bool, 2>` , it is unclear about the meaning of bool switches.
Pass RuntimeOptions by value and mark its object as a const member in Pool classes. Since we use struct and keep member public, here is no need to put the underscore at the end of variable names.

PDoakORNL:
I think from our API perspective the struct RuntimeOptions is a fine way to minimize significant change to the API. I'd argue that  adding elements to the struct is only an ABI break. We offer zero assurance that there will be binary level compatibility between different modules with different commits of the code.

This is basically just a really light input class.

PDoakORNL:
What I'm trying to get a feel for is how do we deal with the eventual issue that we have many API's essentially defined by Real and Value currently hidden by the fact that they derive from QMCTraits and that has preprocessor swapping of type aliases.   We need to dispatch based on that.
So somehwhere there will be many:
```
  if ( RuntimeOptions.isMixed)
     someMagicType = std::make_unique<ParticleSet<float,double>>;
  else
     someMagicType = std::make_unique<ParticleSet<double,double>>;
```
I guess `SomeMagicType` could be a variant.

In other codes I've done something like the following:

```
enum class RTO {
FLOAT
DOUBLE,
COMPLEX,
DOUBLE_COMPLEX
}

RTO runtime_options;
...
//Input is parsed
//based on that runtime options gets a enum value that can be decoded at runtime to a particular statically typed code path. 
runIt = [&](auto& qmc_main) {
   qmc_main.execute(cur)
...
};

switch runtime_options {
   case rto::FLOAT:
       QMCMain<float, double> qmcmain;
       runIt(qmcmain);
       break;
...
   case rto::DOUBLE_COMPLEX:
       QMCMain<double, std:;Complex<double>> qmcmain;
       runIt(qmcmain);
       break;
}
```

ye-luo:
> So I think this idea is fine, but I'm trying to get a feel for is how do we deal with the eventual issue that we have many API's essentially defined by Real and Value currently hidden by the fact that they derive from QMCTrait.
> 
> So I think somewhere there will be many:
> 
> ```
>   if ( RuntimeOptions.isMixed)
>      someMagicType = std::make_unique<ParticleSet<float,double>>;
>   else
>      someMagicType = std::make_unique<ParticleSet<double,double>>;
> ```
> 
> I guess `SomeMagicType` could be a variant.
> 
> In other codes I've done something like the following:
> 
> ```
> enum class RTO {
> FLOAT
> DOUBLE,
> COMPLEX,
> DOUBLE_COMPLEX
> }
> 
> RTO runtime_options;
> ...
> //Input is parsed
> //based on that runtime options gets a enum value that can be decoded at runtime to a particular statically typed code path. 
> runIt = [&](auto& qmc_main) {
>    qmc_main.execute(cur)
> ...
> };
> 
> switch runtime_options {
>    case rto::FLOAT:
>        QMCMain<float, double> qmcmain;
>        runIt(qmcmain);
>        break;
> ...
>    case rto::DOUBLE_COMPLEX:
>        QMCMain<double, std:;Complex<double>> qmcmain;
>        runIt(qmcmain);
>        break;
> }
> ```
> 
> Which I know we're trying to avoid.

A second thought. Do we really need to make the whole particleSet template or maybe we can just make ParticleSet::R as a variant between Vector<float> and Vector<double>.

williamfgc:
@ye-luo approach is what I'd follow. Preferably, only certain affected member functions (if private even better as all templates would be in one translation unit) would follow this pattern instead of an entire class. It would difficult putting them in containers

PDoakORNL:
Well I'm interested in what that will end up looking like.  This is a pretty general issue.  Template parameters extended up to the scope of main is not really what is desired.

ye-luo:
Test this please

ye-luo (approved):

PR #4482, opened 2023-02-24
Fix walker initialization in batched drivers when the request number is smaller than the carry-over

ye-luo:
## Proposed changes
When the number of requested walkers is higher than the number of walkers from the previous section,
the remaining walkers are initialized as the gold copy which has a walker configuration from before any QMC section runs, namely far from equilibrium. It is much better to initialize them by repeating the walker configurations of previous QMC section.

- Bugfix

- No

epyc-server

## Checklist


prckent (dismissed):
Agreed.

ye-luo:
Test this please

ye-luo:
> Aside: we could consider adding an input flag for whether to renormalize the population to the target population at the start of a section. Could make the logic in different scenarios easier for the user and for us.

Please wait for my next PR and after that. We can decide what behavior to choose.

prckent (approved):

prckent:
>Please wait for my next PR and after that. We can decide what behavior to choose.
Will do.
There is also #4483

Issue #4483, opened 2023-02-27
Continuing a DMC run results in "Fatal Error. Aborting at SimpleFixedNodeBranch::resetRun cannot initialize WalkerController"


krongch2:
Continuing a DMC run results in "Fatal Error. Aborting at SimpleFixedNodeBranch::resetRun cannot initialize WalkerController"

**Describe the bug**
I tried to continue a DMC run for this H2O example: https://github.com/QMCPACK/qmcpack/tree/develop/examples/molecules/H2O
and ran into the following error: 
```
Fatal Error. Aborting at SimpleFixedNodeBranch::resetRun cannot initialize WalkerController
```
All the files including inputs and outputs within the directory are uploaded here: https://drive.google.com/drive/folders/1QJ2mZvzO0F1wCGVN22hB4VQ8WOmMp19N?usp=share_link

**To Reproduce**
1. First, I edited the `simple-H2O.xml` file so that the field `checkpoint="-1"` becomes `checkpoint="0"` to ensure that the `config.h5` file is generated after the first DMC run.
2. Then, I ran the input file `simple-H2O.xml` by submitting the following script with `bsub dmc.bsub.in`. 
```
#!/bin/bash
#BSUB -P mat221
#BSUB -J dmc
#BSUB -o dmc.out
#BSUB -e dmc.err
#BSUB -W 00:20
#BSUB -nnodes 1
#BSUB -alloc_flags "smt1"

module load gcc/9.3.0 spectrum-mpi cuda essl netlib-lapack hdf5/1.10.7 fftw; module use /gpfs/alpine/mat151/world-shared/opt/modules; module load llvm/release-15.0.0-cuda11.0
export OMP_NUM_THREADS=7
jsrun -a 1 -b rs -c 7 -d packed -g 1 -n 6 -r 6 --smpiargs="-disable_gpu_hooks" /gpfs/alpine/mat151/world-shared/opt/qmcpack/release-3.15.0/build_summit_Clang_cpu_cplx/bin/qmcpack_complex simple-H2O.xml
```
The QMCPACK executable was compiled by @ye-luo and is located in `/gpfs/alpine/mat151/world-shared/opt/qmcpack/release-3.15.0/build_summit_Clang_cpu_cplx/bin/qmcpack_complex` on Summit. 
This step autogenerates `H2O.s002.cont.xml` for the subsequent run. 

[...]

PR #4486, opened 2023-02-27
Error abort for unsupported non-DMC restarts after DMC in legacy drivers

ye-luo:
## Proposed changes

Close https://github.com/QMCPACK/qmcpack/issues/4483

- Bugfix

- No

epyc-server

## Checklist


prckent (approved):
LGTM

prckent:
Test this please

src/QMCDrivers/SimpleFixedNodeBranch.cpp
```diff
@@ -782,6 +783,11 @@ void SimpleFixedNodeBranch::read(const std::string& fname)
     }
   }
 
+  if (BranchMode[B_DMC] && !bmode[B_DMC])
```
ye-luo, line 786: When troubleshooting https://github.com/QMCPACK/qmcpack/issues/4487, I noticed that the added blocker blocks DMC-> any QMC  because bmode[B_DMC] is always 0.

Issue #3590, opened 2021-11-05
Unit test and documentation needed for qdens-radial tool


mcbennet:
Unit tests and documentation need to be added for the `qdens-radial` tool  (Added in #3587). I can take care of this after the workshop presentation next week, if someone can assign me.

Issue #3092, opened 2021-04-12
nexus qdens test is failing on some systems


prckent:
**Describe the bug**

https://cdash.qmcpack.org/CDash/testSummary.php?project=1&name=ntest_nexus_qdens&date=2021-04-10

Listed as failing with gcc10 and pgi on nitrogen, also failing with gcc10 on sulfur. However also working with Intel2019.1 on sulfur and other system. Seems most likely to be a configuration issue+more error trapping needed, but the error message is cryptic. Ideas?

```
    Start 1557: ntest_nexus_qdens

1557: Test command: /qmcpack/nexus/bin/nxs-test "-R" "qdens" "--ctest" "--pythonpath=git_QMCPACK_prckent/qmcpack/nexus/lib"
1557: Test timeout computed to be: 3600
1557: Test name     : qdens
1557: Test sublabel : test_density
1557: Test exception: "AssertionError: "
1557: Test backtrace:
1557:   File "git_QMCPACK_prckent/qmcpack/nexus/bin/nxs-test", line 478, in run
1557:     self.operation()
1557:   File "git_QMCPACK_prckent/qmcpack/nexus/bin/nxs-test", line 1626, in qdens
1557:     nunit_all()
1557:   File "git_QMCPACK_prckent/qmcpack/nexus/bin/nxs-test", line 362, in nunit_all
1557:     run_external_unit_test(test_name,unit_test)
1557:   File "git_QMCPACK_prckent/qmcpack/nexus/bin/nxs-test", line 388, in run_external_unit_test
1557:     unit_test()
1557:   File "git_QMCPACK_prckent/qmcpack/nexus/tests/unit/test_qdens.py", line 68, in test_density
1557:     assert(check_value_eq(set(os.listdir(dmc_path)),set(files_aft)))
1557: 
1557: Test status: fail
1/1 Test #1557: ntest_nexus_qdens ................***Failed    0.59 sec
```
**To Reproduce**

ctest -R ntest_nexus_qdens

**Expected behavior**

Test passes

**System:**

As described.

**Additional context**

jtkrogel:
This test is checking whether the expected set of files is being produced as output from qdens.

What do you get if you just run `nxs-test -R qdens --verbose` by hand on nitrogen or sulfur?

prckent:
Hmm:
```
1557: Test status: fail
1/1 Test #1557: ntest_nexus_qdens ................***Failed    0.60 sec

[...]

PR #4490, opened 2023-03-01
Fix units in qdens-radial and add documentation

aannabe:
## Proposed changes

While I was trying to compare the charge densities obtained from `pwscf` versus `qmcpack`, I noticed that the charge density object in `qdens-radial`  uses `change_distance_units` to go from (e/Bohr^3) to (e/Ang^3) so that the overall integrated charge density is equal to N_e. Although the right density is obtained, I believe this is due to a double error, one in the definition of `change_distance_units` and one in the usage. In addition, the `change_distance_units` seems to be incorrectly transforming the grid points, thus `change_density_units` must be called instead (leaving the grid points untouched).

In addition to the above changes, `plot_radial_density` was modified to show the density units and a documentation section was added for `qdens-radial` thus closing #3590. The documentation text was borrowed from qmc-workshop_2021 with some minor modifications.

- Bugfix
- Documentation changes

- No

NERSC/Perlmutter. All nexus tests are passing in Perlmutter. However, `qdens-radial` test is failing for me on Debian. The failure is due to a difference in charge(error) of 2e-16(1e-13) between output and reference. Perhaps #3092 is related.

## Checklist

### DIFF
--- a/docs/analyzing.rst
+++ b/docs/analyzing.rst
@@ -1352,3 +1352,105 @@ sufficient number of blocks remaining following any reblocking.
 
 When twist averaging, the group tag (e.g. ``g000`` or similar) will be
 replaced with ``avg`` in the names of the outputted files.
+
+.. _qdens-radial:
+
+Using the qdens-radial tool to estimate atomic occupations
+----------------------------------------------------------
+
+Once the ``*Density*.xsf`` files are obtained from ``qdens``, one can use the ``qdens-radial`` tool to calculate the on-site populations.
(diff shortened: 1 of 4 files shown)


ye-luo (approved):
LGTM

ye-luo:
Test this please

aannabe:
Can #3590 be closed? Is anything else to be added in that regard?

prckent:
I think you can go ahead and close it. A new issue can be made if needed.

PR #4507, opened 2023-03-12
Introduce QMC_GPU_ARCHS option

ye-luo:
## Proposed changes

Introduce QMC_GPU_ARCHS for setting actual GPU architectures (not programming model).
Currently It is hard-coded using sm_70 for NVIDIA and gfx906 on AMD in a few places.
Intel uses JIT and so not affected.
When the hard-coded default are not the intended GPUs, users may succeed to compile but fail to run if they didn't specify the correct GPU architectures via CMake options. This is bad user experience.

The added QMC_GPU_ARCHS should serve both CUDA/HIP and OpenMP offload as the single source of input.
If QMC_GPU_ARCHS is not set by user, it first tries to be derived from CMAKE_CUDA/HIP_ARCHITECTURES and thus existing scripts continue to work.
If still not set, use amdgpu-arch(LLVM15+) and nvptx-arch(LLVM16+) to detect existing GPUs on the node.
If it still cannot be figured out, CMake stops.
There is also consistency check added between QMC_GPU_ARCHS and CMAKE_CUDA/HIP_ARCHITECTURES if they are both specified.

QMC_GPU_ARCHS allows specify multiple architectures to enable a single executable running multiple GPU architectures.

- New feature
- Documentation changes

- No

epyc-server

## Checklist


ye-luo:
Test this please

CMake/ClangCompilers.cmake
```diff
@@ -110,7 +75,7 @@ set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fno-omit-frame-pointer")
 # unfortunately this removes standalone-debug altogether for offload builds
 # but until we discover how to use the ${OPENMP_OFFLOAD_COMPILE_OPTIONS} more selectively
 # this is the only way to avoid a warning per compilation unit that contains an omp symbol.
-if (NOT OFFLOAD_TARGET MATCHES "nvptx64")
+if(NOT OFFLOAD_TARGET MATCHES "nvptx64")
```
ye-luo, line 78: Needs to update this line.

prckent (commented):
This would be a good simplification for users Q. Which compilers ship amdgpu-arch and nvptx-arch? {LLVM, vendor, gcc})

ye-luo:
> This would be a good simplification for users Q. Which compilers ship amdgpu-arch and nvptx-arch? {LLVM, vendor, gcc})

amdgpu-arch is shipped with ROCm and llvm15+
nvptx-arch is shipped by llvm16+.

prckent (commented):
LGTM. Good to improve the user experience with more auto detection.

prckent (dismissed):

prckent:
Test this please

ye-luo:
@prckent I make a correction to the multi-gpu node handling in which I need to remove the duplicates in the return value of `amdgpu-arch` and `nvptx-arch`.

prckent (approved):
Assuming that Summit and Polaris scripts are also OK.

PR #4505, opened 2023-03-09
Complex Spline rotation api implementation

jptowns:
## Proposed changes

- Implements SplineC2C::applyRotation in accordance with @markdewing's recent PR on SplineR2R.
- Removed rotation test from 'test_einset'.
- Added new test 'test_spline_applyrotation'. This test is much more rigorous than the old test in 'test_einset'. This test includes more rotations, multiple rotations, and, if QMC_COMPLEX is set, includes a complex rotation. Checks are done on the entire set of values (psiM, dpsiM, d2psiM) rather than a subset.

NB: This does not make possible complex orbopt. Optimizer stuff will have to change in order for that to happen and so is outside the scope of this PR.

## What type(s) of changes does this code introduce?

- New feature
- Testing changes (e.g. new unit/integration/performance tests)

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
RHEL8 workstation using gcc/openBLAS

## Checklist

- No. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- Yes. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #4508, opened 2023-03-13
Simplify SplineR2R rotation

markdewing:
There's been some discussion in #4505 about the matrix multiply to apply the rotation. This PR simplifies the SplineR2R version.

Use OrbitalSetSize in the matrix multiplication loop bounds, since it should the actual number splines (versus Nsplines which is the number of splines rounded up to an alignment boundary for padding).

The use of tmpU can be removed, and the value of the rotation matrix used directly in the matrix multiplication loop.

Using a GEMM call gets more complicated because the rotation matrix is double precision real, and there will be mismatches if the splines are single precision or complex, which would necessitate make a copy one of the two matrices in those cases.   We would need to measure the performance of this section of code to see if it's worth that effort.  My guess is that this will not be a performance-critical operation (or there will be other places that will be more important when scaling up the number of rotation parameters and spline coefficients).

- No

desktop

## Checklist


jptowns:
This is a nice change. The spline rotation stuff touches a bunch of einspline stuff buried super deep in the code, so in terms of code maintainability I think there should be some comments explaining the size mismatch between spl_coefs and rot_mat. 

If I understand correctly, in a production calculation rotation happens once per sposet construction. In optimization, I /think/ rotation happens /at least/ twice per sposet per optimization loop: once to apply a new rotation at the start of the VMC sampling section, and then again in the accept/reject step. Regardless, its not a bottleneck.

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
```diff
@@ -100,29 +100,17 @@ void SplineR2R<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
...
-  std::fill(tmpU.begin(), tmpU.end(), 0.0);
-  for (auto i = 0; i < rot_mat.size1(); i++)
-  {
-    for (auto j = 0; j < rot_mat.size2(); j++)
-    {
-      tmpU[i][j] = rot_mat[i][j];
-    }
-  }
-
   // Apply rotation the dumb way b/c I can't get BLAS::gemm to work...
```
ye-luo, line 103: Please add a note about mismatch. What is padded and what is not.

ye-luo: Also
```
assert(OrbitalSetSize == rot_mat.rows())
assert(OrbitalSetSize == rot_mat.cols())
```

ye-luo: I didn't see the node as they are not close to the changed part and github review hides them.

ye-luo:
LGTM.

ye-luo:
Test this please

prckent (approved):

ye-luo:
Test this please

PR #4511, opened 2023-03-13
Detect amdgpu using gfx instead amdgcn.

ye-luo:
## Proposed changes
Fix a regression of memory association on AMD GPUs caused by the change in #4507.
Detect amdgpu using gfx instead amdgcn.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
I noticed that our CI missing OpenMP offload on AMD GPU.

ye-luo:
Test this please

prckent (approved):

Issue #4510, opened 2023-03-13
Spack CI constant failure


williamfgc:
**Describe the bug**
The recent CI report a constant failure related to numerical issues on Debug mode: 
- https://github.com/QMCPACK/qmcpack/actions/runs/4407886471/jobs/7723357699
- https://github.com/QMCPACK/qmcpack/actions/runs/4407661797/jobs/7721631980

**To Reproduce**
1. current develop branch
2. cmake command: `cmake -GNinja -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DQMC_MPI=0 -DCMAKE_BUILD_TYPE=Debug /__w/qmcpack/qmcpack`
3. build: `ninja`
4. run tests: `ctest --output-on-failure -L deterministic -j 2`

**Expected behavior**
Tests should pass

**System:**
 - GitHub Actions runners, local Ubuntu 22 running on container 

**Additional context**
Numerical differences might be related to debug mode?
Try configuring with RelWithDebInfo

camelto2:
I'll point out that it does seem to be specific to Debug mode in the CI, but these tests actually fail on my M1 Mac, and I am not building with debug mode on.

prckent:
deterministic-He_ae-opt-1-16-check  fails indicating that the wavefunction optimization has produced a different jastrow factor.

It could be that the test is just a bit sensitive, but my worry has always been that there is a threading related bug here. These are not simple single driver QMC tests and involve a wavefunction optimization. Perhaps something is not initialized or there is a botched reduction operation. Some experimentation will be needed to clear this up. e.g. Does a test using 1 thread but same statistics as current test prove much more reliable.

ye-luo:
We can try thread sanitizer to see if there is race.

ye-luo:
Orange pi (ARM) also failed. https://cdash.qmcpack.org/CDash/viewTest.php?onlyfailed&buildid=403751

markdewing:
An aside on the Orange Pi failure - the run for deterministc-He_ae-opt-1-16 switches to 8 threads for the second optimizer loop.
https://cdash.qmcpack.org/CDash/testDetails.php?test=25483891&build=404150
```
<vmc stage="main" blocks="5">
  Cannot make clones again. Use existing 8 clones
  Initial partition of walkers 0 2 4 6 8 10 12 14 16 
```
```
  Number of samples loaded to each thread : 2 2 2 2 2 2 2 2
```
[...]

PR #4515, opened 2023-03-15
Protect the available thread count against library calls changing it

ye-luo:
## Proposed changes
Fixes https://github.com/QMCPACK/qmcpack/issues/4510

- Bugfix

- No

OrangePi5
## Checklist

### DIFF
--- a/src/Concurrency/UtilityFunctions.hpp
+++ b/src/Concurrency/UtilityFunctions.hpp
@@ -25,28 +25,34 @@ namespace qmcplusplus
 namespace Concurrency
 {
 template<Executor TT = Executor::OPENMP>
-class OverrideMaxCapacity;
-
-template<>
-class OverrideMaxCapacity<Executor::OPENMP>
+class ThreadCountProtector
 {
-private:
-  int original_max_threads_;
-
-public:
-  OverrideMaxCapacity(int max_threads)
-  {
-    original_max_threads_ = omp_get_max_threads();
-    omp_set_num_threads(max_threads);
-  }
-
-  ~OverrideMaxCapacity() { omp_set_num_threads(original_max_threads_); }
+  ThreadCountProtector() {}
 };
 
-template<Executor TT>
-class OverrideMaxCapacity
+template<>
+class ThreadCountProtector<Executor::OPENMP>
+{
+private:
+  const int original_max_threads_;
+
+public:
+  ThreadCountProtector() : original_max_threads_(omp_get_max_threads()) {}
+
+  ~ThreadCountProtector() { omp_set_num_threads(original_max_threads_); }
+};
+
+template<Executor TT = Executor::OPENMP>
+class OverrideMaxCapacity : private ThreadCountProtector<TT>
 {
   OverrideMaxCapacity(int max_threads) {}
+};
+
+template<>
+class OverrideMaxCapacity<Executor::OPENMP> : private ThreadCountProtector<Executor::OPENMP>
+{
+public:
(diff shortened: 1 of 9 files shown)


ye-luo:
UtilityFunctions.hpp actually contain utility classes. @PDoakORNL Do you prefer changing it or not.

prckent (commented):
* Why is this for linear algebra only? I think the code is general but the current use is restricted to linear algebra since OpenBLAS is the only library we have found that misbehaves (is buggy) in this way (changing the thread count of the calling process).

* I prefer OMPThreadCountProtector as the name since it is more specific and nearly fully describes what this class does.

ye-luo:
> * Why is this for linear algebra only? I think the code is general but the current use is restricted to linear algebra since OpenBLAS is the only library we have found that misbehaves (is buggy) in this way (changing the thread count of the calling process).
> 
>     * I prefer OMPThreadCountProtector as the name since it is more specific and nearly fully describes what this class does.

ThreadCountProtector is a template class with Executor::OPENMP as one of the template parameter values. Thus its name should avoid OMP and it can be extended to other threading mode if needed following the design of Concurrency.
For this reason, LAThreadProtect is wrapper which restrict its use to LA and we don't need to propagate HAVE_OPENBLAS definition via config.h globally.

ye-luo:
if you prefer LAThreadProtector can be renamed to LAOMPThreadCountProtector as it is coded to use OpenMP.

prckent:
Depending on where the OMP coding is, could be ThreadCountProtector or OMPThreadCountProtector, your call. The LA part of the current name seems unneeded since it is not restricted to linear algebra. This is some defensive programming to protect us from misbehaving libraries - if HDF5 or libXML2 behaved similarly we could use the same code.

ye-luo:
The LA part is intended for fine-grain control and has interaction with macros and CMake. Activate the protection or not is a compile time decision and I try to keep it contained inside LAThreadProtector.cpp. Adding HDF5ThreadProtector wrapping on top of ThreadCountProtector is cheap. However if three libraries share exactly the same implementation and only HDF5 library that needs the protection, it will cause paying the price at every place using this shared implementation even if there is no need to pay the cost around libxml or LA. So right now ThreadCountProtector is shared at lower level but shim layer LAThreadProtector is specialized.

PDoakORNL (approved):
This looks good to me

PDoakORNL:
Just for the record I think this is an insane thing for a library to just do. Can we file a bug on openBLAS?

ye-luo:
> Just for the record I think this is an insane thing for a library to just do. Can we file a bug on openBLAS?

We did.
https://github.com/xianyi/OpenBLAS/issues/3940

ye-luo:
> The LA part is intended for fine-grain control and has interaction with macros and CMake. Activate the protection or not is a compile time decision and I try to keep it contained inside LAThreadProtector.cpp. Adding HDF5ThreadProtector wrapping on top of ThreadCountProtector is cheap. However if three libraries share exactly the same implementation and only HDF5 library that needs the protection, it will cause paying the price at every place using this shared implementation even if there is no need to pay the cost around libxml or LA. So right now ThreadCountProtector is shared at lower level but shim layer LAThreadProtector is specialized.

@prckent make sense?

prckent (dismissed):

ye-luo:
Test this please

prckent (approved):

PDoakORNL:
`UtilityClasses` would be more correct. `ScopedThreadCounts` would be more descriptive. It really doesn't seem like making two files is helpful.  I don't think its essential for merging this.

prckent:
Test this please

PR #4517, opened 2023-03-15
[WIP] Save/load rotation parameters from optimization

markdewing:
Save the rotation parameters to the VP h5 file.

The state of the rotation matrix requires more state than the ones in the main parameter list provide.  The extra state is stored in sections that store either the global rotation parameters or a list of applied rotations.

Just to make it clean, the values of rotation parameters in the main parameter list are not used and have no meaning beyond their use during optimization.  (That is, it's not rotation parameters on the main list + extra state that is needed to construct the rotation matrices, only the extra state)

Marked "WIP" because there are no tests.

There's also a fix to guard the derivatives when the rotation parameters are excluded via "variational_subset".
This should be a separate PR (and expanded to the to the other derivative routines that get used with PP).

- New feature

- No

desktop

## Checklist

- Yes (mostly). Code added or changed in the PR has been clang-formatted


ye-luo:
`readExtraParameters/saveExtraParameters` the reason for "extra" is due to the fact that other variational parameter I/O is handled via the single aggregated buffer. I think we needs to move away from this pattern.
Here is my suggestion. Add `read/writeParameters` in `OptimizableObject` for read/write both variational parameters and also extra parameters and handle additional data setup. Then we can gradually move off the I/O via the global buffer.

PR #4518, opened 2023-03-16
qmc-fit: Added equation of states and morse fits with jackknife

kayahans:
Equation of states and morse potential fitting capabilities are added to qmc-fit. Morse fit was available in numerics.py, but it was not user-friendly, therefore it is wrapped in qmc-fit. Other equation of state fits (Vinet, Birch, Murnaghan) didn't have jackknife fitting capabilities, hence such capability is added in this PR. The code is tested with nxs-test, but the existing qmc-fit tests do not cover these new capabilities. 

- New feature

- No

macOS Monterey 12.6.3, Python 3.9.13

## Checklist


prckent (commented):
* What are your plans to document this?
* How is this tested? I think some run calling these capabilities is required, e.g. via extensions of qmc-fit tests.

kayahans:
- I am open to suggestions about how to document these. I think the argparse helper provides a lot of information such as available parameters and defaults, but for the manual I can write a small section as well with example data. 
- I have tested it on my data for bilayer binding energies and also Daniel Wines's data for bulk moduli which yielded within 10 % match to the experimental bulk moduli of C, Ge and Si. I will discuss with Jaron on how to run tests for the added functionality.

prckent:
Great. Can you add an initial manual entry in this PR? The rest can be done subsequently imo.


As background to my comments and thinking, I have some general concerns about our python-based tools and NEXUS-related contributions that we need to find a way to tackle. None are specific to this PR!

- They need to be documented in the QMCPACK manual so that people can discover them. Something that exists only in the NEXUS or tool source code is just not going to be found and used. The added functionality in this PR is very useful, we'd like poeple to use it.
- We need proper testing, in particular we need to be running tests of these tools in CI and our nightlies. Otherwise we won't know when they break in future or be able to catch changes. Ideally we would have a coverage report so that reviewers could see this new code was run. For example, problems due to API deprecations and changes would be caught.
- We need to be serious about python dependencies and having a proper installation for these tools along with the QMCPACK executables. If we had those, making packages for operating systems would be straightforward and result in a fully usable installation. Currently we only install the compiled code.

prckent (approved):
Confirmed html and PDF generation are OK locally.

Thanks Kayahan.

prckent:
Test this please

kayahans:
Thanks Paul, I will be adding the unit test as well.

prckent:
Test this please

ye-luo:
Test this please

ye-luo:
Keep this moving and I added  #4539

PR #4523, opened 2023-03-21
Pass RuntimeOptions to TrialWaveFunction

williamfgc:
## Proposed changes

Pass it from WaveFunctionPool to
- `TrialWaveFunction` constructor
- `WaveFunctionFactory::buildTWF`
Modify tests accordingly using ProjectData
Related to #3411 and #4099

- No

Ubuntu 22

## Checklist

### DIFF
--- a/src/QMCDrivers/tests/test_QMCCostFunctionBase.cpp
+++ b/src/QMCDrivers/tests/test_QMCCostFunctionBase.cpp
@@ -11,7 +11,7 @@
 
 #include "catch.hpp"
 #include "QMCDrivers/WFOpt/QMCCostFunctionBase.h"
-#include "Utilities/ProjectData.h"
+#include "Utilities/RuntimeOptions.h"
 
 
 namespace qmcplusplus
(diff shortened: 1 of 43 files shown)


src/QMCDrivers/tests/test_QMCCostFunctionBase.cpp
```diff
@@ -55,7 +56,8 @@ TEST_CASE("updateXmlNodes", "[drivers]")
   const SimulationCell simulation_cell;
   MCWalkerConfiguration w(simulation_cell);
   QMCHamiltonian h;
-  TrialWaveFunction psi;
+  ProjectData project_data;
```
ye-luo, line 59: Can we just use RuntimeOptions instead of ProjectData? Maybe also put RuntimeOptions in a separate header file.

ye-luo (changes requested):

src/QMCDrivers/tests/test_QMCCostFunctionBase.cpp
williamfgc: ProjectData objects currently aligns with the current build picking the appropriate RuntimeOptions for complex/real or full/mixed precision. I might have to move this at the RuntimeOptions constructor level.

ye-luo (approved):

ye-luo:
> LGTM.
> 
> I don't think it's necessary for this PR but I don't think RuntimeOptions belongs in Utilities, lets think on where these sort of files should go. We need to consider minimizing intermodule entanglements. Platform?

runtime options are QMC particleset and WF logics. It should not be under Platforms. I also cannot find a better location and thus Utilities should be fine.

ye-luo:
Test this please

PR #4525, opened 2023-03-23
[WIP] Store VP in groups by wavefunction component

markdewing:
Store variational parameters as suggested by Ye in
https://github.com/QMCPACK/qmcpack/pull/4517#issuecomment-1470684859

Each optimizable component handles reading/writing its own parameters.   The default implementation in OptimizableObject operates on the variables returned checkInVariablesExclusive.
Objects that need to need store additional information (like RotatedSPOs) can override {save,read}VariationalParameters.  That will be implemented in a future PR - for now this is a refactor of the existing scheme.

The VP file changes internal organization and the major version is changed to "2".  Each component stores its parameters in a separate group. 
The ability to read version 1 files is still present.

Marked "WIP" because there are no tests yet.

- No

desktop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/OptimizableObject.cpp
+++ b/src/QMCWaveFunctions/OptimizableObject.cpp
@@ -1,0 +1,100 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2023 QMCPACK developers.
+//
+// File developed by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
+//
+// File created by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+#include "OptimizableObject.h"
+#include "VariableSet.h"
+#include "io/hdf/hdf_archive.h"
+
+/**@file OptimizableObject.cpp
+ *@brief OptimizableObject implementation
+ */
+
+namespace qmcplusplus
+{
+
+using opt_variables_type = optimize::VariableSet;
+
+void OptimizableObject::openHDFToSave(const std::string& filename, hdf_archive& hout)
+{
+  hout.create(filename);
+  std::vector<int> vp_file_version{2, 0, 0};
+  hout.write(vp_file_version, "version");
+}
+
+int OptimizableObject::openHDFToRead(const std::string& filename, hdf_archive& hin)
+{
+  if (!hin.open(filename, H5F_ACC_RDONLY))
+  {
+    std::ostringstream err_msg;
+    err_msg << "Unable to open VP file: " << filename;
+    throw std::runtime_error(err_msg.str());
+  }
+  std::vector<int> vp_file_version(3);
+  hin.read(vp_file_version, "version");
+  return vp_file_version[0];
+}
+
+void OptimizableObject::saveVariationalParameters(hdf_archive& hout, opt_variables_type& opt_vars)
+{
+  opt_variables_type dummy;
+  checkInVariablesExclusive(dummy);
+  hid_t name_grp = hout.push(name_);
+
+  hid_t grp = hout.push("name_value_lists");
+
+  std::vector<qmcplusplus::QMCTraits::ValueType> param_values;
+  std::vector<std::string> param_names;
+  for (auto pair_it = dummy.begin(); pair_it != dummy.end(); pair_it++)
+  {
+    const std::string& param_name = pair_it->first;
+    param_names.push_back(param_name);
+    //auto opt_it = opt_vars.find(param_name);
+    //if (opt_it == opt_vars.end())
+    //  throw std::runtime_error("Inconsistent parameter " + param_name + " in " + name_);
+    //param_values.push_back(opt_it->second);
+    param_values.push_back(pair_it->second);
+  }
+
+  hout.write(param_names, "parameter_names");
+  hout.write(param_values, "parameter_values");
+  hout.pop(); // name_value_lists
+  hout.pop(); // component name in name_
+}
+
+void OptimizableObject::readVariationalParameters(hdf_archive& hin, opt_variables_type& opt_vars)
+{
+  hid_t name_grp = hin.push(name_);
+  hid_t grp      = hin.push("name_value_lists", false);
+  if (grp < 0)
+  {
+    std::ostringstream err_msg;
+    err_msg << "The group name_value_lists in not present in file";
+    throw std::runtime_error(err_msg.str());
+  }
+
+  std::vector<qmcplusplus::QMCTraits::ValueType> param_values;
+  hin.read(param_values, "parameter_values");
+
+  std::vector<std::string> param_names;
+  hin.read(param_names, "parameter_names");
+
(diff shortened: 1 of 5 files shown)


src/QMCWaveFunctions/OptimizableObject.cpp
```diff
@@ -0,0 +1,100 @@
...
+  return vp_file_version[0];
+}
+
+void OptimizableObject::saveVariationalParameters(hdf_archive& hout, opt_variables_type& opt_vars)
+{
+  opt_variables_type dummy;
+  checkInVariablesExclusive(dummy);
+  hid_t name_grp = hout.push(name_);
+
+  hid_t grp = hout.push("name_value_lists");
```
quantumsteve, line 52:
```suggestion
  hid_t grp = hout.push(hdf_path(name_) / "name_value_lists");
```

src/QMCWaveFunctions/OptimizableObject.cpp
```diff
@@ -0,0 +1,100 @@
...
+    //if (opt_it == opt_vars.end())
+    //  throw std::runtime_error("Inconsistent parameter " + param_name + " in " + name_);
+    //param_values.push_back(opt_it->second);
+    param_values.push_back(pair_it->second);
+  }
+
+  hout.write(param_names, "parameter_names");
+  hout.write(param_values, "parameter_values");
+  hout.pop(); // name_value_lists
+  hout.pop(); // component name in name_
```
quantumsteve, line 70:
```suggestion
  hout.pop(); // component name in name_ / name_value_lists
```

src/QMCWaveFunctions/OptimizableObject.cpp
```diff
@@ -0,0 +1,100 @@
...
+using opt_variables_type = optimize::VariableSet;
+
+void OptimizableObject::openHDFToSave(const std::string& filename, hdf_archive& hout)
+{
+  hout.create(filename);
+  std::vector<int> vp_file_version{2, 0, 0};
+  hout.write(vp_file_version, "version");
+}
+
+int OptimizableObject::openHDFToRead(const std::string& filename, hdf_archive& hin)
```
quantumsteve, line 33:
```suggestion
int OptimizableObject::openHDFToRead(const std::filesystem::path& filename, hdf_archive& hin)
```

src/QMCWaveFunctions/OptimizableObject.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+  opt_variables_type dummy;
+  checkInVariablesExclusive(dummy);
+  hid_t name_grp = hout.push(name_);
+
+  hid_t grp = hout.push("name_value_lists");
+
+  std::vector<qmcplusplus::QMCTraits::ValueType> param_values;
+  std::vector<std::string> param_names;
+  for (auto pair_it = dummy.begin(); pair_it != dummy.end(); pair_it++)
```
quantumsteve, line 56: Should be have a specialization `h5data_proxy<std::map<Key,T>>`?

https://github.com/QMCPACK/qmcpack/blob/464286d39f207853a76da801301e708c4538b3a7/src/io/hdf/hdf_stl.h#L27

markdewing: VariableSet (the type of opt_vars) is not a map, though some of the API's act like it.  I'm not sure it's worth getting a map specialization for h5data_proxy working, and updating VariableSet to work with it.  VariableSet should also be replaced in the future with something that is not map-like, as part of scaling to larger numbers of variational parameters.

src/QMCWaveFunctions/OptimizableObject.h
```diff
@@ -72,6 +75,23 @@ class OptimizableObject
   virtual void reportStatus(std::ostream& os) {}
 
   void setOptimization(bool state) { is_optimized_ = state; }
+
+  /** Open HDF file for saving variational parameters */
+  static void openHDFToSave(const std::filesystem::path& filename, hdf_archive& hout);
```
ye-luo, line 80: If I open file directly and bypass openHDFToSave, VPs can still be read/written. I don't feel OptimizableObject is responsible of versioning VP schemes. If that is really needed, it can be in a separate class with proper encapsulation. In the initial design, prefer to do minimal.

src/QMCWaveFunctions/OptimizableObject.h
```diff
@@ -72,6 +75,23 @@ class OptimizableObject
...
+
+  /** Save variational parameters.
+   * The hout parameter should be opened by openHDFToSave
+   * The default implementation saves the parameters listed in checkInVariablesExclusive,
+   * The parameter values are taken from opt_vars */
+  virtual void saveVariationalParameters(hdf_archive& hout, opt_variables_type& opt_vars);
+
+  /** Read variational parameters.
+   * The hin parameter should be opened by openHDFToRead */
+  virtual void readVariationalParameters(hdf_archive& hin, opt_variables_type& opt_vars);
```
ye-luo, line 94: Use either read/write or load/save pairs and I was thinking of only one argument (hdf_archive&)

ye-luo:
Here is an example
```
ObjA: public OptimizableObject
{
  opt_variables_type my_var_;
  # the following two connects my_var_w with global which is used by the optimizer
  checkInVariablesExclusive(opt_variables_type& global) override
  checkOutVariablesExclusive(opt_variables_type& global) override
  # the following two connects my_var_ with h5 file.
  readVariationalParameters(hdf_archive& hin);
  writeVariationalParameters(hdf_archive& hout);
}
```

src/QMCWaveFunctions/WaveFunctionFactory.cpp
```diff
@@ -158,7 +158,16 @@ std::unique_ptr<TrialWaveFunction> WaveFunctionFactory::buildTWF(xmlNodePtr cur,
...
+    hdf_archive hin;
+    int major_version = OptimizableObject::openHDFToRead(vp_file_to_load, hin);
+    if (major_version == 1)
+      dummy.readFromHDF(vp_file_to_load);
+    else
+    {
+      UniqueOptObjRefs opt_obj_refs = targetPsi->extractOptimizableObjectRefs();
+      for (auto opt_obj : opt_obj_refs)
+        opt_obj.get().readVariationalParameters(hin, dummy);
+    }
```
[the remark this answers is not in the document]
markdewing: Versioning data files is pretty important.   Three digits is the standard for semantic versioning, and we ought to have a really good reason to use a different versioning scheme.

Only the major version was returned because that's all that is needed for now.  If future implementations need more of the version, the code can be changed.

It sounds like the design you're asking for is #4517, but with the name of {read,save}ExtraParameters changed to drop the "extra" part of the name.

The big difference between that PR and this one is how the implementation default is handled for OptimizableObjects that don't handle reading/writing their own variational parameters.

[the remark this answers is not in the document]
ye-luo: > It sounds like the design you're asking for is https://github.com/QMCPACK/qmcpack/pull/4517, but with the name of {read,save}ExtraParameters changed to drop the "extra" part of the name.

yes. Sort of. The reason to drop "extra" is to have per OptimizableObject I/O not just doing the "extra" parameters but also the variational parameters this OptimizableObject object owns. So we can have the global buffer I/O as only a secondary method and rely on more precise per component I/O as the primary method.

PR #4528, opened 2023-03-23
Nexus: added capability to run self-consistent DFT+U+V in QE > 7.1 

kayahans:
## Proposed changes

QE has significantly updated hubbard input syntax in ver. => 7.1. This PR adds capability to run DFT+U and +V calculations using Nexus with QE=>7.1. In addition to providing the Hubbard card in pw.x input, Pwscf class in Nexus is also modified to receive 'hubbard_parameters' as a result from a previous calculation. Thanks to this update, self-consistent DFT+U+V calculations can be performed in Nexus. For the self consistent DFT+U+V calculations, Nexus delegates work to the linear response code, hp.x, in the QE distribution. A temporary example of self-consistent U+V calculations is given for C-diamond. Currently, QE7.1 distribution is not supported under external_codes in QMCPACK, but I think pw2qmcpack.x converter from earlier versions should be able to work using wavefunctions generated with ver. =>7.1. 

I have noted some quirks with the DFT+U+V code in QE7.1, but I have learned that these are fixed in the development version and will be published in QE7.2 soon: 
1. QE complains and terminates if the atomic wavefunction manifold names are not capitalized. Thus I changed (2s->2S and 2p->2P) in C.BFD.upf to get the example to work.
2. Negative U (and possibly V) values are not allowed in pw.x, although they are hypothetically possible. Thus, I temporarily keep positive_only = True in pwscf_input.py to have self-consistent calculations running. 

- New feature

- No

macOS Monterey 12.6.3, Python 3.9.13

## Checklist


kayahans:
After discussing with Ye, I understood that the pw2qmcpack.x can handle the conversion of new Hubbard manifolds. I made some final commits to get it ready for PR.

prckent:
Wondering if there is anything holding back merging this?

QE 7.3 was out on 9 January 2024. Any issues?

Does this PR also mean that the current DFT+U format is also fully supported?

kayahans:
@prckent I think it is ready for PR. If you are asking whether the current (older than 7.1) DFT+U format is still in place, the answer is yes. I think Nexus does not perform QE version check, but adds the keywords cumulatively. Nevertheless, at line 2073 of nexus/lib/pwscf_input.py of this PR I wrote some code which complains if both hubbard input formats are defined in the same nexus pwscf object.  

For QE 7.3, I didn't see any item in the release notes that could have a conflict. They started supporting non-collinear DFT+U+V calculations in 7.3 version, but that should use the pre-existing flags (noncolin=.true.).  
https://gitlab.com/QEF/q-e/-/releases

prckent (approved):
Addressing a question just now, looks like we missed merging this and that nothing is holding it back.

@jtkrogel should comment on any compatibility concerns or python issues.

I have been using the latest QE and benefiting from the automatic parallelization.

prckent:
Test this please

jtkrogel:
Sorry for my very long silence here.  I would like this PR to go in as soon as is practical.  A number of current projects need this functionality.

I think all that is needed is for the branch to be merged with current develop.  Can we proceed?

prckent:
Test this please

prckent:
Test this please

PR #4529, opened 2023-03-24
Remove unnecessary hdf5 error checking

shivupa:
## Proposed changes
Hdf5 wrapper `read` has built-in error checking which is duplicated in other places if the lower level `readEntry` is used. This PR replaces some `readEntry` with `read`.

- No

Workstation 

## Checklist

### DIFF
--- a/src/QMCTools/LCAOHDFParser.cpp
+++ b/src/QMCTools/LCAOHDFParser.cpp
@@ -326,17 +326,9 @@ void LCAOHDFParser::getMO(const std::string& fname)
   char name[72];
   sprintf(name, "%s", "/Super_Twist/eigenset_0");
   setname = name;
-  if (!hin.readEntry(CartMat, setname))
-  {
-    setname = "SPOSet::putFromH5 Missing " + setname + " from HDF5 File.";
-    APP_ABORT(setname.c_str());
-  }
+  hin.read(CartMat, setname);
   sprintf(name, "%s", "/Super_Twist/eigenval_0");
-  if (!hin.readEntry(EigVal_alpha, setname))
-  {
-    setname = "SPOSet::putFromH5 Missing " + setname + " from HDF5 File.";
-    APP_ABORT(setname.c_str());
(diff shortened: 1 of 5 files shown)


src/QMCTools/LCAOHDFParser.cpp
```diff
@@ -326,17 +326,9 @@ void LCAOHDFParser::getMO(const std::string& fname)
   char name[72];
   sprintf(name, "%s", "/Super_Twist/eigenset_0");
   setname = name;
-  if (!hin.readEntry(CartMat, setname))
-  {
-    setname = "SPOSet::putFromH5 Missing " + setname + " from HDF5 File.";
-    APP_ABORT(setname.c_str());
-  }
+  hin.read(CartMat, setname);
```
ye-luo, line 329: Fine with this change. Using the built-in message in read/write doesn't reduce infomation to users.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -1150,8 +1150,7 @@ bool SlaterDetBuilder::readDetListH5(xmlNodePtr cur,
       APP_ABORT("Unknown HDF5 CI format");
     }
 
-    if (!hin.readEntry(temps[grp], ds_tag))
-      APP_ABORT("Unknown HDF5 CI format");
+    hin.read(temps[grp], ds_tag);
```
ye-luo, line 1153: In this case, the built-in message won't be as accurate as the explicit error message. So the old one is better but please change the APP_ABORT to runtime_error

shivupa: Yeah I thought about overloading read to take an informative message to pass to runtime_error. We could replace some `readEntry` in some other places as well if we did that. 

I reverted this changed though and throw runtime_error as requested.

ye-luo (approved):

PR #4530, opened 2023-03-27
fix code hightlighting default in devloping section. upd copyright

PDoakORNL:
## Proposed changes

A fix and a couple of updates to the development guide. 
First much of the C++ code was actually receiving generic python highlighting which looks just terrible.
In Restructured text, the format sphinx uses, with sphinx being the code what renders the manual now both locally and for `readthedocs` you actually need to specify the language you want syntax highlighting for in a block of code.  You can also set a default for a context.  Here I set the default for the `developing.rst` to be c++.

I've also revised some of the text around encapsulation. And removed reference to an exception being made for the use of c style casts.  I don't think there is or should be one.

- Bugfix
- Documentation changes

- No

mac laptop

## Checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,7 +18,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'QMCPACK Manual'
-copyright = '2021, QMCPACK Developers'
+copyright = '2023, QMCPACK Developers'
 author = 'QMCPACK Developers'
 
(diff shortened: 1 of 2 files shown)


ye-luo (approved):

prckent (approved):
Confirmed PDF generation still OK

ye-luo:
Test this please

PR #4540, opened 2023-04-04
[WIP] Remove Observable Helper

PDoakORNL:
## Proposed changes

Make  `OperatorEstBase::write` virtual and override for estimators that do not simply write their data block to hdf5 actually implemented.  Integration test added.

My only explanation here is that we had inadequate testing, and there is plenty more "corner" cases like this totally uncovered to be found.

This is a draft because for whatever reason I have not put together code to successfully use hdfarchive to read a previous written dataset.  Anyone who can point out my issue please do so.

- Bugfix
- New feature
- Documentation changes

- No

Mac laptop

## Checklist

### DIFF
--- a/src/Estimators/SpinDensityNew.cpp
+++ b/src/Estimators/SpinDensityNew.cpp
@@ -231,5 +231,31 @@ void SpinDensityNew::registerOperatorEstimator(hdf_archive& file)
   }
 }
 
+void SpinDensityNew::write(hdf_archive& file)
+{
+  if (h5desc_.empty())
+    return;
+
+  auto writeFullPrecData = [&](auto& fp_data) {
+    std::size_t offset = 0;
+    for (auto& h5d : h5desc_)
+    {
+      h5d.write(fp_data.data() + offset, file);
+      offset += derived_parameters_.npoints;
+    }
+  };
+
+#ifdef MIXED_PRECISION
+  std::vector<QMCT::FullPrecRealType> expanded_data(data_.size(), 0.0);
+  std::copy_n(data_.begin(), data_.size(), expanded_data.begin());
+  assert(!data_.empty());
+  writeFullPrecData(expanded_data);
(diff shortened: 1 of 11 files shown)


camelto2:
In the ObservableHelper::set_dimensions call already has an offset into the data as a function argument (the lower_bound variable in ObservableHelper). The problem with SpinDensityNew seems to be in SpinDensityNew::registerObservableEstimator. In that we register both the SpinDensity/u and SpinDensity/d to the hdf5 and set the dimensions. However, the offset is passed in as 0 in both cases, even though the data_ is twice the size to account for up and down. Thats why it writes the up twice...it started from index 0 in the underlying data for both u and d. 

It seems that all of this is fixed by using oh.set_dimensions(ng, s * derived_parameters_.npoints) in that API. 

Maybe I'm wrong, but if you set the correct starting index using this set_dimensions API, you don't need to make OperatorEstBase::write virtual since it is already taken care of.

PDoakORNL:
You're right about that.  But I think now a better solution would be just to drop ObservableHelper altogether its adding 0 value here.

src/Estimators/SpinDensityNew.cpp
```diff
@@ -195,41 +209,64 @@ void SpinDensityNew::collect(const RefVector<OperatorEstBase>& type_erased_opera
...
+      file.pop();
+    }
+    ++append_index_;
+    file.pop();
+  };
 
-  std::vector<int> ng(1, derived_parameters_.npoints);
+#ifdef MIXED_PRECISION
+  std::vector<QMCT::FullPrecRealType> expanded_data(data_.size(), 0.0);
+  std::copy_n(data_.begin(), data_.size(), expanded_data.begin());
```
quantumsteve, line 261: Can we do the cast during construction of `expanded_data`?
```suggestion
  std::vector<QMCT::FullPrecRealType> expanded_data(data_.cbegin(), data_.cend());
```

PDoakORNL:
continue work in private

PR #4541, opened 2023-04-05
Read and write variational parameters

markdewing:
Implement just the read/write parts of #4517, and add a test.

The "extra" part of the name was removed.  In the future, the variational parameter reading and writing to HDF for all objects may be handled through these functions (and the VariableSet read/write functions removed).
But for now, the default set of parameters is handled through the VariableSet functions, and only objects that need extra data (orbital rotation) need to implement these functions.

The minor file version was incremented to indicate these files could have data from other objects present.

- New feature

- No

desktop

## Checklist


src/QMCWaveFunctions/tests/test_OptimizableObject.cpp
```diff
@@ -13,16 +13,59 @@
...
+    hout.write(var2, "var2");
+    hout.write(extra_data, "extra_data");
+    hout.pop();
+  }
+
+  void readVariationalParameters(hdf_archive& hin) override
+  {
+    hid_t group = hin.push("FakeOptimizableObject", false);
+    if (group < 0)
+      throw std::runtime_error("Expected group not present in file");
```
ye-luo, line 53: I feel the error should be handled by `push()`.

src/QMCWaveFunctions/VariableSet.cpp
```diff
@@ -244,11 +244,10 @@ void VariableSet::print(std::ostream& os, int leftPadSpaces, bool printHeader) c
   }
 }
 
-void VariableSet::saveAsHDF(const std::string& filename) const
+void VariableSet::writeToHDF(const std::string& filename, qmcplusplus::hdf_archive& hout) const
 {
-  qmcplusplus::hdf_archive hout;
   hout.create(filename);
-  std::vector<int> vp_file_version{1, 0, 0};
+  std::vector<int> vp_file_version{1, 1, 0};
```
ye-luo, line 250: Could you add a note saying what 1.0.0 is and what 1.1.0 changes.

src/QMCWaveFunctions/tests/test_OptimizableObject.cpp
markdewing: That might be better, but this PR is not the place to make that change.

ye-luo: Could you change 51-53 lines to
```
hin.push("FakeOptimizableObject", false);
```
since we expect to add the guard in `push()` in a later PR.

markdewing: It can be changed later in that PR.  Presumably other places in the code would need a similar change as well.

ye-luo: I still feel better to remove the check here and we don't need to revisit this file later. But this PR doesn't need to be blocked.

ye-luo (commented):
LGTM.

markdewing:
Test this please

rcclay (approved):
LGTM

PR #4545, opened 2023-04-10
skip ppconvert compilation when isnan unsupported

correaa:
## Proposed changes

`ppconvert` produces in NaN values in normal use (oxygen test).

`ppconvert` apparently uses `NaN` as part of its internal implementation. My conclusion is that `ppconvert` simply cannot be compiled with `-fast-math`. This is in addition and independent on the problem with `icc`.

According to my tests, variables are set to NaN internally or result in NaN, even with normal usage on the oxygen example.

At the same time removing all the conditions still produces passing tests for the (one and only) test of ppconvert with oxygen. So, it is hard to know if this really has a practical consequence. Is `ppconvert` regularly used by anyone that can contribute more test cases?

So, there are two options, 1) forbid compilation with fast math to allow the utilization of NaNs that happen in normal use (safer), or 2) remove all the `std::isnan` and its side effects.

This PR implements 1) forbid compilation with fast math to allow the utilization of NaNs

The safeguard used `__STDC_IEC_559__` and it seems to work with GCC, but it is not clear if it works with other compilers.

```
  #ifndef __STDC_IEC_559__
  #error This program depends on NaN markings and therefore relies on IEEE 754 Standard, ensure this with the necessary compilation flags.
  #endif

if (std::isnan(Vup))
   Vup = 0.0;
...
```

- Bugfix

- No

Fedora 37, with GCC 12

## Checklist

- Yes. This PR adds tests to cover any new code or to catch a bug that is being fixed


prckent (commented):
Since we compile ppconvert by default, this will have the side effect of giving an error whenever fast-math is defaulted on, complicating the QMCPACK installation. This problem needs to be minimized and instructions given. Since we want to support QMCPACK compilation with fast-math, it might be better to simply disable ppconvert. Another option would be to properly understand where the NaN's come from here - the check looks like some defensive programming (good) to handle bad numerics - perhaps these can be caught another way.

correaa:
I think it is better to do `ppconvert` special:

Add the option -no-fast-math to the ppconvert target in CMake and leave the `#error` pragma in the PR to prevent anyone to changing the compilation option accidentally.
(defensive compilation 😅)

prckent:
Lets go the defensive compilation route :-) Good pragmatic suggestion.

ye-luo:
Can we add a small test in CMake right after https://github.com/QMCPACK/qmcpack/blob/cdd0d8795558e519045842e17235de17a859bf3b/src/QMCTools/ppconvert/CMakeLists.txt#L8
if the test fails, disable ppconvert.

The test produces a NaN and check std::isnan functionality if it returns true, skip ppconvert.

correaa:
I added this to cmake 

```
target_compile_options(ppconvert PUBLIC -fno-fast-math)  # ppconvert generates and checks for NaN in normal use; needs IEEE compliance
```

Let's see what MacOS says.

src/QMCTools/ppconvert/src/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@ target_sources(
          "${CMAKE_CURRENT_SOURCE_DIR}/XMLWriterClass2.cc" "${CMAKE_CURRENT_SOURCE_DIR}/NLPPClass.cc"
          "${CMAKE_CURRENT_SOURCE_DIR}/ParserClass.cc")
 
+target_compile_options(ppconvert PUBLIC -fno-fast-math)  # ppconvert generates and checks for NaN in normal use; needs IEEE compliance
```
ye-luo, line 11: It can potentially choke non GCC/Clang compiler I guess. Does this work with all C++ compilers?

prckent:
@correaa When you say "MacOS" please specify the compiler and arch. e.g. While we would like to use apple clang, at least recently it didn't have enough C++17 for us, and unfortunately no OpenMP, so we end up using a gcc or llvm via macports, brew, or selfinstall.

src/QMCTools/ppconvert/src/CMakeLists.txt
correaa: I guess it works in all compilers that support `-ffast-math`.

correaa:
> @correaa When you say "MacOS" please specify the compiler and arch. e.g. While we would like to use apple clang, at least recently it didn't have enough C++17 for us, and unfortunately no OpenMP, so we end up using a gcc or llvm via macports, brew, or selfinstall. Depending how installed these can pull in different includes.

I am talking about the MacOS in the test, https://github.com/QMCPACK/qmcpack/actions/runs/4656325824/jobs/823984055 . I don't have a Mac to try.

src/QMCTools/ppconvert/src/CMakeLists.txt
ye-luo: So it can be a hassle and I don't think we should take this route.

correaa: Do you think `-fno-fast-math` is not supported in all the compilers that matter?

ye-luo:
```
$ nvc++ -fno-fast-math main.cpp 
nvc++-Error-Unknown switch: -fno-fast-math
```

correaa: good catch, is there a CI test for nvc++ currently?

ye-luo: We do have intel classic. As I suggested, write a test in cmake checking `std::isnan` and turn off ppconvert if the test fails.

correaa: The problem is that these kind of checks are fragile, depend on running the code (never is going to work for crosscompilation), and  basically, we are saying that `ppconvert` is going to be compiled or not compiled "randomly" depending on the compiler and flags that leak from the global settings. For that matter we can only compile it with GCC and clang, at least we will know when it is compiled and when it is not.

ye-luo: ppconvert should be a host tool. With crosscompilation, the results of test failing is not producing a ppconvert, no harm.

markdewing:
Since NaN is a bit pattern, it should be possible to make detection function that checks the bit pattern, and that function would be robust against the fast math flag (though dependent on implementation details of the floating point format, but every platform we care about is going to be IEEE 754, so that shouldn't be a problem)

Some discussion of options here:
https://stackoverflow.com/questions/570669/checking-if-a-double-or-float-is-nan-in-c
Some answers give functions to check bit patterns.  We could grab one of those to use as our own isnan function.

Alternately, the `fpclassify` function looks interesting (https://www.gnu.org/software/libc/manual/html_node/Floating-Point-Classes.html ). 
The glibc implementation of fpclassify does check bit patterns.
The problem is std::fpclassify in the include files gets redirected to a compiler built-in version, which does the faster self-comparison check for NaN (and hence fails with fast-math).

The glibc implementation appears to be available as `__fpclassify`.   In one test, `__fpclassify(x) == FP_NAN` works robustly against the fast-math flag.   I haven't tested how wide-spread that function might be available.

There may be other other ways to manipulate the defines so `/usr/include/math.h` uses `__fpclassify` instead of the built-in.

correaa:
I like this approach.
I think it is the least intrusive.

prckent:
I think too much effort is being spent on this, please do not do anything involved. CMake check is trivial. As noted above, ppconvert is not a priority, cleaning up QMCPACK compilation warning is. ppconvert really has few users.

ye-luo:
I found `-O3` without `-ffast-math` may also cause non-effective `isnan` on mac.
I feel the best way is simply do a check run with `isnan()`.
The check may cause false negative in cross-compilation but ppconvert is not that important.

ye-luo:
@prckent I made significant change that I should not review my own changes.

correaa:
> I found `-O3` without `-ffast-math` may also cause non-effective `isnan` on mac.

Yes, that is why my change above failed in MacOS.

> I dropped `__STDC_IEC_559__` which is a gcc thing only.

I though it was more general than gcc. Good to know.

prckent (approved):
Thanks all.

prckent:
Test this please

Issue #4556, opened 2023-04-17
Inconsistent results from qmca when using different scalar.dat versions


prckent:
**Describe the bug**
While checking #4549 I noticed that at some recent point we have reordered the scalar.dat output. The columns are correctly labeled. Unfortunately qmca doesn't completely handle this situation and gives inconsistent results when files of different versions are mixed. They are nearly OK which suggests the correct data is being used for the most part.

**To Reproduce**

Analyze different versions of scalar files simultaneously. In the following notice the debug* results are reported consistently but the orig* ones are not. Quantities other than the energy are suspect, e.g. samples, although I noticed this first by the energy.

Archive of the respective files attached. Used qmca from develop.

```
$ qmca -q e */*/*.scalar.dat
debug/vmc_perf_pbe_u_None_3x3x1_1x1x1_4990_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.290314 +/- 0.029507 
debug/vmc_perf_pbe_u_None_3x3x1_1x1x1_5000_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.336379 +/- 0.093084 
debug/vmc_perf_pbe_u_None_3x3x1_1x1x1_5010_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.176166 +/- 0.030095 
orig/vmc_perf_pbe_u_None_3x3x1_1x1x1_4990_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -198.202521 +/- 0.026945 
orig/vmc_perf_pbe_u_None_3x3x1_1x1x1_5000_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.273603 +/- 0.026549 
orig/vmc_perf_pbe_u_None_3x3x1_1x1x1_5010_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -198.168970 +/- 0.031373 
$ qmca -q e d*/*/*.scalar.dat
debug/vmc_perf_pbe_u_None_3x3x1_1x1x1_4990_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.290314 +/- 0.029507 
[...]

PR #4557, opened 2023-04-18
Nexus: use data hashing in heuristic equilibration detection algorithm

jtkrogel:
## Proposed changes

The heuristic equilibration detection algorithm used by `qmca` when `-e` is unspecified uses a random number to avoid stastistical bias in the selection of the equilibration length.  This approach generates an estimate of the mean that varies each time `qmca` is run.   While statistically correct, users expect deterministic behavior.  This PR seeds the RNG based on the hash of the data being analyzed.  This preserves use of the RNG for unbiasedness while presenting deterministic behavior to the user.

Addresses #4556.  The reported means are now consistent when multiple files are used (or when the same file is used repeatedly):

```
>qmca -q e */*/*.scalar.dat
debug/vmc_perf_pbe_u_None_3x3x1_1x1x1_4990_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.290314 +/- 0.029507 
debug/vmc_perf_pbe_u_None_3x3x1_1x1x1_5000_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.336379 +/- 0.093084 
debug/vmc_perf_pbe_u_None_3x3x1_1x1x1_5010_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.176166 +/- 0.030095 
orig/vmc_perf_pbe_u_None_3x3x1_1x1x1_4990_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -198.198560 +/- 0.027040 
orig/vmc_perf_pbe_u_None_3x3x1_1x1x1_5000_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.274630 +/- 0.026488 
orig/vmc_perf_pbe_u_None_3x3x1_1x1x1_5010_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -198.168970 +/- 0.031373 

>qmca -q e d*/*/*.scalar.dat
debug/vmc_perf_pbe_u_None_3x3x1_1x1x1_4990_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.290314 +/- 0.029507 
debug/vmc_perf_pbe_u_None_3x3x1_1x1x1_5000_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.336379 +/- 0.093084 
debug/vmc_perf_pbe_u_None_3x3x1_1x1x1_5010_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.176166 +/- 0.030095 

>qmca -q e o*/*/*.scalar.dat
orig/vmc_perf_pbe_u_None_3x3x1_1x1x1_4990_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -198.198560 +/- 0.027040 
orig/vmc_perf_pbe_u_None_3x3x1_1x1x1_5000_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -197.274630 +/- 0.026488 
orig/vmc_perf_pbe_u_None_3x3x1_1x1x1_5010_ccecparep_legacy_500_500_1.5_50_True_hf/vmc  series 0  LocalEnergy           =  -198.168970 +/- 0.031373 
``` 

- New feature

- No

Workstation

## Checklist


markdewing:
There are still some ways this might not reproduce the value.

* The hash function (https://docs.python.org/3/reference/datamodel.html#object.__hash__ ) includes a random salt for string and byte values.   The data here is probably only numeric (numpy array?), so this most likely won't be an issue.  Since Python definitions don't have a declared type,  the type of `x` is not immediately obvious.
* The hash value is not guaranteed to be the same between Python versions (but it usually is).
* Changing the data (removing the last line, for instance), could result in a larger than expected change in the output.

These are probably sufficiently unlikely scenarios that having a consistent output via hashing is the more useful solution, but I did want to get them listed and considered.

prckent:
This would work for me. Have you thought about printing out the equilibration length? Giving an indication that equilibration length was taken into account was a good suggestion. (Definitely could be done in another PR)

prckent:
Test this please

prckent:
Test this please

prckent:
Test this please

prckent:
Test this please

PR #4581, opened 2023-05-02
Handle disabled variational parameters in rotation derivatives

markdewing:
Disabled variational parameters will return -1 from a call to `myVars.where`, leading to an out-of-bounds array access.  This PR adds a check to avoid this situation.

Variational parameters can be disabled due to choosing a variational subset.  

Also in the case of a RHF wavefunction where there are different numbers of spin up/down electrons, there will be a different number of rotation parameters for each spin.  At a higher level of managing the parameters, this case is not being handled correctly, but this fix will stop it from segfaulting on invalid array access in the meantime.

- Bugfix

- No

desktop

## Checklist


markdewing:
The RHF case described in #4582

markdewing:
Test this please

rcclay (approved):

Issue #4547, opened 2023-04-10
handle h5 group check in hdf_archive


ye-luo:
**Is your feature request related to a problem? Please describe.**
See comment
https://github.com/QMCPACK/qmcpack/pull/4541/files#r1158964516

markdewing:
The comment is asking for the error to be handled in the call to `push`.

To be clear, the desired behavior is:
   If the group name is not present and `createit` is false, then the `push` call should throw an exception.

ye-luo:
I don't feel right returning hid_t type outside hdf_archive which should hid details of the hdf5 library. So  I'm thinking of
1. push(name, createit). Throw an exception when creation is requested and failed or whgen creation is not requested and open group failed.
2. is_group can be used to inspect group existence.
3. if the group tid_t is actually needed use top() after a successful push().

PR #4585, opened 2023-05-03
Update hdf_archive push

markdewing:
Addresses #4547 

Changes to hdf_archive:
* When 'push' is called with the createit flag set to false, and the group does not exist, throw an exception.
* Also throw an exception if the file is not open.
* Remove the return value (use `top` to get the value if it's really needed)
* Store the hdf archive filename so it can be included in the error messages in the exceptions.

In most cases, I added try/catch blocks to preserve the existing error handling flow.  But in a few places I let the underlying exception propagate.

In LCAOrbitalBuilder.cpp, the logic around 'PBC' was updated to match what was described in the commented-out code.

- No

desktop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
+++ b/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
@@ -1060,15 +1060,7 @@ bool SlaterDetBuilder::readDetListH5(xmlNodePtr cur,
   }
 
 
-  try
-  {
-    hin.push("MultiDet", false);
-  }
-  catch (...)
-  {
-    std::cerr << "Could not open Multidet Group in H5 file" << std::endl;
-    abort();
-  }
+  hin.push("MultiDet", false);
 
   hin.read(H5_ndets, "NbDet");
   if (ndets != H5_ndets)
@@ -1110,15 +1102,8 @@ bool SlaterDetBuilder::readDetListH5(xmlNodePtr cur,
       abort();
     }
 
-    try
-    {
-      coeffin.push("MultiDet", false);
-    }
-    catch (...)
-    {
-      std::cerr << "Could not open Multidet Group in H5 file" << std::endl;
-      abort();
-    }
+    coeffin.push("MultiDet", false);
+
     coeffin.read(OptCiSize, "NbDet");
     CIcoeffopt.resize(OptCiSize);
 
--- a/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
+++ b/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
@@ -825,7 +825,7 @@ bool SlaterDetBuilder::readDetList(xmlNodePtr cur,
             }
           } // if(name=="det")
           csf = csf->next;
-        }   // csf loop
+        } // csf loop
         if (DetsPerCSF.back() == 0)
         {
           APP_ABORT("Found empty CSF (no det blocks).");
@@ -1059,7 +1059,12 @@ bool SlaterDetBuilder::readDetListH5(xmlNodePtr cur,
     abort();
   }
 
-  if (!hin.push("MultiDet"))
+
+  try
+  {
+    hin.push("MultiDet", false);
+  }
+  catch (...)
   {
     std::cerr << "Could not open Multidet Group in H5 file" << std::endl;
     abort();
@@ -1105,7 +1110,11 @@ bool SlaterDetBuilder::readDetListH5(xmlNodePtr cur,
       abort();
     }
 
-    if (!coeffin.push("MultiDet"))
+    try
+    {
+      coeffin.push("MultiDet", false);
+    }
+    catch (...)
     {
       std::cerr << "Could not open Multidet Group in H5 file" << std::endl;
       abort();
--- a/src/io/hdf/hdf_archive.cpp
+++ b/src/io/hdf/hdf_archive.cpp
@@ -43,7 +43,6 @@ void hdf_archive::close()
   }
   if (file_id != is_closed)
     H5Fclose(file_id);
-  filename_.clear();
   file_id = is_closed;
 }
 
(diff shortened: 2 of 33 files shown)


src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -1059,7 +1059,12 @@ bool SlaterDetBuilder::readDetListH5(xmlNodePtr cur,
...
 
-  if (!hin.push("MultiDet"))
+
+  try
+  {
+    hin.push("MultiDet", false);
+  }
+  catch (...)
   {
     std::cerr << "Could not open Multidet Group in H5 file" << std::endl;
```
ye-luo, line 1069: Is this error handling still needed? There are similar pattern below.

src/io/hdf/hdf_archive.cpp
```diff
@@ -138,6 +138,7 @@ void hdf_archive::set_access_plist(boost::mpi3::communicator& comm, bool request
 
 bool hdf_archive::create(const std::filesystem::path& fname, unsigned flags)
 {
+  filename_ = fname;
```
ye-luo, line 141: adding `assert(filename_.empty())`?

src/io/hdf/hdf_archive.cpp
```diff
@@ -149,6 +150,7 @@ bool hdf_archive::create(const std::filesystem::path& fname, unsigned flags)
 
 bool hdf_archive::open(const std::filesystem::path& fname, unsigned flags)
 {
+  filename_ = fname;
```
ye-luo, line 153: Add a similar assert.
I also noticed missing wiping out filename_ at `close()`.

markdewing: If there are operations than happen after a file is closed, it would be useful to have the name of the file that was open in the error message.  So clearing the filename_ variable at close doesn't seem helpful.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
markdewing: I don't know if it's still needed.  This changes the error handling flow the least.

markdewing: I removed the try/catch in this file.  The existing error handling simply aborts on error, and moving to an unhandled exception seems similar.

src/io/hdf/hdf_archive.cpp
markdewing: I assumed that there was a 1:1 correspondence between an hdf_archive instance and an HDF file.   But I see from the code that doesn't necessarily have to be the case.  I added a check to hdf_archive, and it turns out there are places in the code where an hfd_archive instance is reused.

src/io/hdf/hdf_archive.h
```diff
@@ -202,11 +204,16 @@ class hdf_archive
...
     group_id.pop();
+    group_names.pop_back();
     herr_t err = H5Gclose(g);
     if (err < 0)
       throw std::runtime_error("H5Gclose failed with error.");
   }
 
+  /** Return a string representation of the current group stack
+   */
+  std::string group_path_as_string();
```
ye-luo, line 215: `std::string group_path_as_string() const;`

ye-luo (dismissed):
Thanks @markdewing

ye-luo:
Test this please

ye-luo:
@markdewing there are hdf_archive uses in AFQMC causing CI failure.

markdewing:
Test this please

markdewing:
Some of ORNL CI failures had this message; `ERROR Failed to open group "state_0" because file is not open`.
This message would have been more useful if the name of the file that hdf_archive was trying to access were printed out.
I change the name of the variable from `filename_` to `possible_filename_` to indicate that it is debugging information, not part of the class state.

I think part of the issue is in `HDFWalkerInput_0_4.cpp` - the return value from the `open` call is not checked, and so the subsequent `push` calls fail.   I don't know why the open failed, though.  Possibly a resource issue on the machine?

src/io/hdf/hdf_archive.h
```diff
@@ -73,8 +73,11 @@ class hdf_archive
   ///Track group names corresponding to group_id
   std::vector<std::string> group_names;
 
-  /// Name of file
-  std::string filename_;
+  /** Name of file that hdf_archive thinks is open.
+   *  This may not correspond to the actual file because the open call failed,
+   *  or the file was closed. This information is useful for debugging.
```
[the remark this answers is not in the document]
ye-luo: Maybe your intention is to just introduce debug code temporarily.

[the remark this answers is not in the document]
markdewing: This is IO code - there are always going to be issues (sometimes related to the environment, sometimes related to the usage of the class) and having as much diagnostic information as possible is going to be helpful.    Eventually we can clean up the usage of the class, but there will always be something new related to the environment.

Diagnostic code doesn't have to cover every possible situation to be provide useful information.

markdewing:
Don't know which line it is - there are multiple open calls, and there is no backtrace.

The group "state_0" is defined as `main_state` in HDFVersion.h.  So it could be any of the `push(hdf::main_state)` calls.

src/io/hdf/hdf_archive.cpp
```diff
@@ -149,6 +151,7 @@ bool hdf_archive::create(const std::filesystem::path& fname, unsigned flags)
 
 bool hdf_archive::open(const std::filesystem::path& fname, unsigned flags)
 {
```
ye-luo, line 153: Similar to `create()`. make it on the restrictive side.
```
bool hdf_archive::open(const std::filesystem::path& fname, unsigned flags, bool allowed_to_fail = false)
```
This default behavior change may require updating a few test. We can temporarily choose default true and then flip to true in a separate PR. Trying not to grow the size of this PR. Sorry about dealing the legacy code.

ye-luo:
> Don't know which line it is - there are multiple open calls, and there is no backtrace.
> 
> The group "state_0" is defined as `main_state` in HDFVersion.h. So it could be any of the `push(hdf::main_state)` calls.

Unfortunately this surfaced the loose file handling we have.

markdewing:
Yeah, it would be nice to tighten up all the file handling code that uses hdf_archive.  But I don't want to fix it all in this PR :)

I can't reproduce the CI failures on my own machine, but I haven't tried the exact offload builds.   Given that a number of builds failed, it could be something with those machine's environment.

markdewing:
It turns out the problem does reproduce on my local machine.  I just wasn't running enough of the test suite.

src/io/hdf/hdf_archive.cpp
```diff
@@ -189,11 +192,12 @@ bool hdf_archive::is_group(const std::string& aname)
...
 }
 
-hid_t hdf_archive::push(const std::string& gname, bool createit)
+void hdf_archive::push(const std::string& gname, bool createit)
 {
   hid_t g = is_closed;
   if (Mode[NOIO] || file_id == is_closed)
-    return is_closed;
+    throw std::runtime_error("Failed to open group \"" + gname +
+                             "\" because file is not open.  Expected file: " + possible_filename_);
```
ye-luo, line 200: probably there is a behavior change with `Mode[NOIO] ` which is intended to treat every operation as no-op.
So probably
```
if (Mode[NOIO]) return;
if (file_id == is_closed))
  throw
```

ye-luo:
Test this please

ye-luo (approved):
LGTM.

rcclay (approved):

PR #4590, opened 2023-05-04
Remove versions from requirements.txt to fix rtd builds

prckent:
## Proposed changes

Theory: too old versions in current file are source of rtd build failures. Will need to restore specific versions that are supported by spack for reproducibility and local testing.

- Bugfix
- Documentation changes

- No

None. CI needs to run. Unsure if merge required for rtd to see config change.

## Checklist


prckent:
Results in newer  Sphinx v5.3.0 being run, but it is still tripped by the urllib3 v2.0 / openssl dependency issue that is being discussed on multiple repos.

PR #4593, opened 2023-05-05
Orbital rotation optimizer test

markdewing:
Use a carbon atom with STO wavefunction to test the linear optimizer with orbital rotation.

The wavefunction is UHF (up and down coefficients vary separately) and has no Jastrow, so could be compared to other HF basis sets.

The test takes about 40 seconds on a Broadwell machine (E5-2695v4).   Shorter runs with fewer samples were noisier (I can show graphs if interested), and there was more overlap between the range of initial values and the range of optimized values.

The pink runs with 360K samples show that the optimizer convergences in just a few cycles for this problem.
The dark purple runs with 320K samples correspond to the input file being checked in.

The run with 4.48M samples is a single run.  It was used to establish the value for the test.  The error bounds were estimated from the graph.  (These bounds should be able to distinguish the initial value from the optimized value, and hopefully are wide enough to prevent too many random failures)

c_sto_en

- No

desktop

## Checklist


markdewing:
Based on discussion in slack, should use published results with available exponents for HF values.

PR #4609, opened 2023-05-22
Add Be STO VMC test

markdewing:
Add test for Be atom with a restricted HF wavefunction using STO basis functions.

The mw_evaluateVGL code had a buffer overrun (caught by sanitizer) for the case using batched VMC with drift.   Previously it set output_size to 7 (the size of the basis), but phi_v_list only is of size 2 (the number MOs needed in the determinant)

The data (exponents, coefficients, and energies) are taken from https://doi.org/10.1006/ADND.1993.1003 .

The runs in this test are very short to facilitate regression testing.  The energies were testing against much longer VMC test (1.3 billion samples), and also agreed.

The test uses `qmc_run_and_check_custom_scalar` instead of the usual `qmc_run_and_check` for two reasons:
1. `qmc_run_and_check_custom_scalar` uses named parameters which makes the call more legible
2. The scalar estimators are directly named.  (`qmc_run_and_check` converts the name to `check_scalars.py` flag which then converts that to the estimator name used in the scalar.dat file.  This avoids those naming indirections)

- Bugfix

- No

desktop

## Checklist


markdewing:
Test this please

prckent (commented):
Thank Mark!

For the record, please could you note which sanitizer and build option caught the bug. An interesting discovery.

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
```diff
@@ -406,12 +406,12 @@ void LCAOrbitalSet::mw_evaluateVGL(const RefVectorWithLeader<SPOSet>& spo_list,
...
   mw_evaluateVGLImplGEMM(spo_list, P_list, iat, phi_vgl_v);
 
-  const size_t output_size = phi_vgl_v.size(2);
-  const size_t nw          = phi_vgl_v.size(1);
+  const size_t nw = phi_vgl_v.size(1);
 
   //TODO: make this cleaner?
   for (int iw = 0; iw < nw; iw++)
   {
+    const size_t output_size = psi_v_list[iw].get().size();
```
ye-luo, line 414: good catch.

prckent (approved):
Also, while not strictly

markdewing:
The affected test used the batch driver with drift.  It failed with nan's for all the energies in the release mode build.  I built in debug mode with the address sanitizer, and that isolated the issue.

prckent:
Noting that happily these all appeared to work last night. The results are split between the nightly (qmcpack runs) and unstable categories (statistics). e.g. https://cdash.qmcpack.org/CDash/viewTest.php?onlydelta&buildid=417665

PR #3358, opened 2021-08-13
SPOSet::makeClone() returns a std::unique_ptr

quantumsteve:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

Describe what this PR changes and why.  If it closes an issue, link to it here
with [a supported keyword](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

SPOSet::makeClone() and derived classes now return a std::unique_ptr<SPOset>, improving memory safety.

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Refactoring (no functional changes, no api changes)

### Does this introduce a breaking change?

- No

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted

PR #4617, opened 2023-06-01
Fix orbital images code

jtkrogel:
## Proposed changes

Replace odd usage of shared_ptr in OrbitalImages with a regular vector.  In the code prior to this change, the shared_ptr was uninitialized, causing a crash.  This update restores the functionality of the OrbitalImages code.

- Bugfix

- No

Andes

## Checklist

### DIFF
--- a/src/QMCHamiltonians/OrbitalImages.cpp
+++ b/src/QMCHamiltonians/OrbitalImages.cpp
@@ -156,7 +156,7 @@ bool OrbitalImages::put(xmlNodePtr cur)
 
   //second pass parameter read to get orbital indices
   //  each parameter is named after the corresponding sposet
-  sposet_indices->resize(sposet_names.size());
+  sposet_indices.resize(sposet_names.size());
   for (int n = 0; n < other_elements.size(); ++n)
   {
     xmlNodePtr element = other_elements[n];
@@ -166,7 +166,7 @@ bool OrbitalImages::put(xmlNodePtr cur)
       const std::string name(getXMLAttributeValue(element, "name"));
       for (int i = 0; i < sposet_names.size(); ++i)
         if (name == sposet_names[i])
-          putContent((*sposet_indices)[i], element);
+          putContent(sposet_indices[i], element);
     }
   }
 
@@ -219,7 +219,7 @@ bool OrbitalImages::put(xmlNodePtr cur)
 
     sposets.push_back(spo_it->second->makeClone());
     auto& sposet = sposets.back();
-    std::vector<int>& sposet_inds = (*sposet_indices)[i];
+    std::vector<int>& sposet_inds = sposet_indices[i];
     if (sposet_inds.size() == 0)
       for (int n = 0; n < sposet->size(); ++n)
         sposet_inds.push_back(n);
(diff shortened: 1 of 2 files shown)


markdewing:
It looks like this was introduced in #3358

jtkrogel:
Yes, correct.  This is a reversion of that change.

quantumsteve:
Does OrbitalImages have any testing?

jtkrogel:
No.  In the past this was objected to because it is not really an estimator.  The code as it stands is slated for deletion anyway along with the legacy drivers.

ye-luo:
`sposet_indices` should remain `shared_ptr`. I think the right fix is to initialize sposet_indices in the constructor.
Add `sposet_indices(std::make_shared<std::vector<std::vector<int>>>)` to `OrbitalImages::OrbitalImages` definition

prckent:
Is this code scheduled for deletion though? It is one of those pieces of nice to have but rarely used functionality.

Since it has been proven useful recently we could set about keeping it. A first step would be a test since it has zero coverage. The test would need to be custom since this "estimator" is just a way to plot the orbital sets and after doing so quits (aborts) the whole application. Then the question is how to get some version of this functionality in the batched code.

Comments?

ye-luo:
Could like to know what feature is missing in SPOSetScanner compared to OrbitalImages. I guess OrbitalImages at the estimator level requires the whole estimator + QMC driver machinery which is an overkill.

prckent:
@jtkrogel Is Ye's requested code change clear and are you OK with it?

@ye-luo The fixed version of this code is being used for debugging currently, so we need to get it merged.

OrbitalImages is undocumented functionality, but is actually nearly functional. How the functionality is implemented could be changed, but that should be done at a later time. It is not something that we have to worry about backwards compatibility for, but whatever is functional should be documented.

ye-luo:
My comment on SPOSetScanner vs OrbitalImages is only meant to raise my concern about long term strategy. Not concerning the limited change of this PR.

prckent:
Thanks Ye, understood. The use of an observable for this historically was definitely imaginative but it got the job done :-)

jtkrogel:
I don't really get the need for any type of pointer in this case - why is it problematic for an instance to contain a vector?  This is meant to be run single mpi single thread anyway (since we are just printing orbital info from memory) and the code aborts once the files are written:
``` 
app_log() << "OrbitalImages::evaluate  orbital images written successfully, exiting.\n" << std::endl;
APP_ABORT("  Not a fatal error, just exiting.");
```
Essentially, there is no follow-on impact on the rest of the code since there is no follow-on action performed.

AFAIK OrbitalImages is set for deletion (old "estimator") along with the rest of the legacy code, so I don't think its worth putting in the effort to first make and then soon delete a test.

In my mind this is just a patch fix restoring "original" functionality of the legacy code prior to deletion (e.g. as requested by @Paul-St-Young).

Followon discussion about the relative features in SPOSetScanner vs OrbitalImages is worthwhile separately since I think the two were created without awareness of the existence of the other.

ye-luo:
> I don't really get the need for any type of pointer in this case - why is it problematic for an instance to contain a vector?

When directly using a vector, each clone needs to own a copy. Using shared_ptr, each clone doesn't need a copy. This is appropriate for metadata like `sposet_indices`. Keeping a single source of truth and avoiding making unnecessary copies is a practice we need to keep in mind when developing features. A lot of times, making copies simply introduces source of troubles, forgetting copy, updated on copy but not all the clones.

ye-luo:
> app_log() << "OrbitalImages::evaluate  orbital images written successfully, exiting.\n" << std::endl;
> APP_ABORT("  Not a fatal error, just exiting.");

This is a basically a hack and not following good software practice. SPOSetScanner is integrated at the SPOSet builder level. Thus we can make clean exit instead of a hack. Yes, I introduced SPOSetScanner without being aware of OrbitalImages which I viewed as a necessary feature but unfortunately coded in a wrong way. I think bug fixes on OrbitalImages is fine but we better avoid keep developing it further.

Paul-St-Young:
Woah, I did not mean to spur such a long discussion...

This quick-fix for a rarely used legacy feature is just a quality-of-life improvement for a legacy developer such as myself.

There is no need to spend the effort to make it nice IMO.

ye-luo:
> Woah, I did not mean to spur such a long discussion...
> 
> This quick-fix for a rarely used legacy feature is just a quality-of-life improvement for a legacy developer such as myself.
> 
> There is no need to spend the effort to make it nice IMO.

There is only a one line change needed https://github.com/QMCPACK/qmcpack/pull/4617#issuecomment-1572657158

ye-luo:
The original issue has been fixed by #4627

PR #4623, opened 2023-06-02
HSPotential Helper replace ifdefs with const expr templates

walshmm:

## Proposed changes

This is a refactor of `HSPotential_Helpers.cpp` and `HSPotential_Helpers.h`  in accordance with this gist https://gist.github.com/williamfgc/28ffbd23f6888ed89530e3f68c992c93#refactoring-strategy
It removes the QMC_COMPLEX ifdefs and replaces them with templates and const expr as to generate complex/real versions in one compile.

- No

## Checklist

### DIFF
--- a/src/AFQMC/Hamiltonians/HSPotential_Helpers.cpp
+++ b/src/AFQMC/Hamiltonians/HSPotential_Helpers.cpp
@@ -18,6 +18,7 @@
 #include "AFQMC/Numerics/ma_operations.hpp"
 #include "AFQMC/Numerics/csr_blas.hpp"
 #include "AFQMC/Numerics/detail/utilities.hpp"
+#include "type_traits/QMCMacros.h"
 
 /********************************************************************
 *  You get 2 potentials per Cholesky std::vector
@@ -78,7 +79,7 @@ void count_over_cholvec(double cut,
       if (abs(*vik) > cut)
       {
         count[2 * (*cik)] += size_t(2); // Lik + 0
-        if constexpr (is_complex<T>::value)
+        if constexpr (is_complex_v<T>)
           count[2 * (*cik) + 1] += size_t(2); // Lik - 0
       }
       ++cik;
@@ -89,7 +90,7 @@ void count_over_cholvec(double cut,
       if (abs(*vki) > cut)
       {
         count[2 * (*cki)] += size_t(2); // Lki + 0
-        if constexpr (is_complex<T>::value)
+        if constexpr (is_complex_v<T>)
           count[2 * (*cki) + 1] += size_t(2); // Lki - 0
       }
       ++cki;
@@ -102,7 +103,7 @@ void count_over_cholvec(double cut,
     if (abs(*vik) > cut)
     {
       count[2 * (*cik)] += size_t(2); // Lik + 0
-      if constexpr (is_complex<T>::value)
+      if constexpr (is_complex_v<T>)
         count[2 * (*cik) + 1] += size_t(2); // Lik - 0
     }
     ++cik;
@@ -113,13 +114,24 @@ void count_over_cholvec(double cut,
     if (abs(*vki) > cut)
     {
       count[2 * (*cki)] += size_t(2); // Lki + 0
-      if constexpr (is_complex<T>::value)
+      if constexpr (is_complex_v<T>)
         count[2 * (*cki) + 1] += size_t(2); // Lki - 0
     }
     ++cki;
     ++vki;
   }
 }
+
+#define declare_type(T) \
+  template void count_over_cholvec<T>(double cut, \
+                          std::vector<std::size_t>& count, \
+                          int c0, \
+                          int c1, \
+                          SpVType_shm_csr_matrix::reference const& Lik, \
+                          SpVType_shm_csr_matrix::reference const& Lki);
+  QMC_FOREACH_TYPE(declare_type)
+#undef declare_type
+
 template<class T>
 void count_over_cholvec(double cut,
                         std::vector<std::size_t>& count,
@@ -143,7 +155,7 @@ void count_over_cholvec(double cut,
   {
     if (abs(*vi + ma::conj(*vi)) > cut)
       ++count[2 * (*ci)]; // Lii + Lii*
-    if constexpr (is_complex<T>::value) {
+    if constexpr (is_complex_v<T>) {
       if (abs(*vi - ma::conj(*vi)) > cut)
         ++count[2 * (*ci) + 1]; // Lii - Lii*
     }
@@ -151,6 +163,16 @@ void count_over_cholvec(double cut,
     ++vi;
   }
 }
+
+#define declare_type(T) \
+  template \
+  void count_over_cholvec<T>(double cut, \
+                          std::vector<std::size_t>& count, \
+                          int c0, \
+                          int c1, \
+                          SpVType_shm_csr_matrix::reference const& Lii);
+  QMC_FOREACH_TYPE(declare_type)
+#undef declare_type
 
 // In this case, c0/c1 refer to the ranges in the expanded list of CVs, e.g. 2*n/2*n+1
 template<class T>
@@ -192,7 +214,7 @@ void count_nnz(double cut,
         nik++;
         nki++;
       }
-      if constexpr (is_complex<T>::value){
+      if constexpr (is_complex_v<T>){
         if (abs(*vik - ma::conj(*vki)) > cut && // Lik - Lki* and Lki - Lik*
             2 * (*cik) + 1 >= c0 && 2 * (*cik) + 1 < c1)
         {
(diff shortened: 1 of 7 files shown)


prckent (dismissed):
LGTM

Will need a non-ORNL merge.

We should talk if there is anything tricky under AFQMC given our focus on the real space code.

prckent:
Note: we are at the mercy of the AFMC test coverage here. Recommend you have 
QMC_DATA/NiO/NiO_afm_fcidump.h5
QMC_DATA/NiO/NiO_afm_wfn.dat
QMC_DATA/NiO/NiO_nm_choldump.h5
which are only used by AFQMC tests when developing.

src/AFQMC/config.0.h
```diff
@@ -58,6 +58,16 @@ using SPValueType = SPRealType;
...
 using SPComplexType = std::complex<SPRealType>;
 
+template <class T>
+struct is_complex : std::false_type {};
+template <>
+struct is_complex<ComplexType> : std::true_type {};
+template <>
+struct is_complex<SPComplexType> : std::true_type {};
+
+{};
```
quantumsteve, line 68:
```suggestion
template<typename T>
struct is_complex : std::false_type {};

template <class T>
struct is_complex<std::complex<T>> : std::true_type {};

template< class T>
inline constexpr bool is_complex_v = is_complex<T>::value;
```

src/AFQMC/Hamiltonians/HSPotential_Helpers.cpp
```diff
@@ -78,9 +78,8 @@ void count_over_cholvec(double cut,
       if (abs(*vik) > cut)
       {
         count[2 * (*cik)] += size_t(2); // Lik + 0
-#if defined(QMC_COMPLEX)
-        count[2 * (*cik) + 1] += size_t(2); // Lik - 0
-#endif
+        if constexpr (is_complex<T>::value)
```
quantumsteve, line 81:
```suggestion
        if constexpr (is_complex_v<T>)
```
Requires the helper function I suggested adding in config.0.h

src/AFQMC/Hamiltonians/HSPotential_Helpers.cpp
```diff
@@ -35,7 +35,7 @@ namespace
 {
 //inline double const& ma::conj(double const& d){return d;}
 //inline float const& ma::conj(float const& f){return f;}
-
+template<class T>
```
quantumsteve, line 38: Since these functions are in the implementation file we need to explicitly instantiate them.

William's branch has a couple macros to make this easier.

ye-luo: I feel the template parameter should be bool instead of T.

src/type_traits/QMCMacros.h
```diff
@@ -0,0 +1,29 @@
...
+  MACRO(float)                       \
+  MACRO(double)
+
+#define QMC_FOREACH_COMPLEX_TYPE(MACRO) \
+  MACRO(std::complex<float>)            \
+  MACRO(std::complex<double>)
+
+#define QMC_FOREACH_TYPE(MACRO) \
+  QMC_FOREACH_REAL_TYPE(MACRO)  \
+  QMC_FOREACH_COMPLEX_TYPE(MACRO)
```
quantumsteve, line 27: We may want to pick one the other based on `QMC_COMPLEX`.

src/AFQMC/Hamiltonians/HSPotential_Helpers.cpp
quantumsteve: If we go that route, make it an enum or enum class. Names would be clearer than `true` and `false` and if in the future we want to add more template options (mixed precision?) there is less to change.

prckent: Good point @quantumsteve . We will  want to add mixed precision support in future to do away with QMC_MIXED_PRECISION. If we can pick a pattern that make both r/c and sp/dp combinations simpler to support, we should.

src/AFQMC/Hamiltonians/HSPotential_Helpers.h
```diff
@@ -23,15 +23,17 @@ namespace afqmc
 {
 namespace HamHelper
 {
+template<class T>
```
ye-luo, line 26: I think making `template<bool IS_COMPLEX>` reduce the complexity of explicit template instantiation cases.

src/AFQMC/Hamiltonians/HSPotential_Helpers.cpp
PhilipFackler: There's another problem here. Something that was overlooked here is that the type of the `SpVType_shm_csr_matrix` function parameter for these functions would need to become a template parameter since it is dependent on the `SPValueType`. Following that we could use the `CSR_matrix_type::value_type` to determine if we're dealing with complex or not.

To Ye's suggestion then, we would still need to make explicit instantiations for just the two sp value types (where the rest of the specialization of `csr_matrix` remains the same: `int, std::size_t, shared_allocator`).

src/AFQMC/config.0.h
```diff
@@ -58,6 +58,16 @@ using SPValueType = SPRealType;
 using ComplexType   = std::complex<RealType>;
 using SPComplexType = std::complex<SPRealType>;
 
+template <class T>
+struct is_complex : std::false_type {};
+
+template <class T>
+struct is_complex<std::complex<T>> : std::true_type {};
```
quantumsteve, line 65: is_complex is already in the codebase
https://github.com/QMCPACK/qmcpack/blob/2db0b25e4ad64b06f11cfb5ba9644c553094571a/src/type_traits/complex_help.hpp#L17

src/AFQMC/Hamiltonians/HSPotential_Helpers.h
```diff
@@ -23,15 +23,17 @@ namespace afqmc
...
 std::vector<std::size_t> count_nnz_per_cholvec(double cut, TaskGroup_& TG, SpVType_shm_csr_matrix& V2, int NMO);
 
+template<class T>
```
PDoakORNL, line 29: in `config.h` replace 
```
using SpCType_shm_csr_matrix =
    ma::sparse::csr_matrix<SPComplexType, int, std::size_t, shared_allocator<SPComplexType>, ma::sparse::is_root>;
using SpVType_shm_csr_matrix =
    ma::sparse::csr_matrix<SPValueType, int, std::size_t, shared_allocator<SPValueType>, ma::sparse::is_root>;
using CType_shm_csr_matrix =
    ma::sparse::csr_matrix<ComplexType, int, std::size_t, shared_allocator<ComplexType>, ma::sparse::is_root>;
using VType_shm_csr_matrix =
    ma::sparse::csr_matrix<ValueType, int, std::size_t, shared_allocator<ValueType>, ma::sparse::is_root>;
```
with
```
template<typename T>
using CSRMtrxShared<T> = ma::sparse::csr_matrix<SPComplexType, int, std::size_t, shared_allocator<SPComplexType>, ma::sparse::is_root>;
```
In most cases this will just read
`CSRMatrixShared<Value>` or `CSRMatrixShared<Complex>` which is much easier to read quickly and surfaces the important parameter for both read and specializing code.

then one can write
```
template<typename T>
...(...
     CSRMatrix<T>  WorthlessTwoLetterVariableName,
     ...)
{
....
if constexpr (IsComplex_t<T>::value) {
  ...
}
...
}
```

src/AFQMC/Hamiltonians/HSPotential_Helpers.cpp
PDoakORNL: Does anyone feel like they could write some documentation for this function?  Anyway I make a suggestion for changing the matrix type aliases in another comment which are pretty disgusting anyway which will make instantiating the two template cases much less awkward.

src/AFQMC/Hamiltonians/HSPotential_Helpers.cpp
```diff
@@ -376,20 +410,21 @@ void add_to_vn(SpVType_shm_csr_matrix& vn,
...
-      assert(map_[2 * (*ci) + 1] - c_origin < vn.size(1));
-      vn.emplace_back({ik, (map_[2 * (*ci) + 1] - c_origin)},
-                      static_cast<SPValueType>(half * im * (*vi - ma::conj(*vi)))); // Lii - Lii*
+    if constexpr (is_complex_v<T>) {
+      if (abs(*vi - ma::conj(*vi)) > cut && 2 * (*ci) + 1 >= c0 && 2 * (*ci) + 1 < c1)
+      {
+        assert(map_[2 * (*ci) + 1] >= 0);
+        assert(map_[2 * (*ci) + 1] - c_origin < vn.size(1));
+        vn.emplace_back({ik, (map_[2 * (*ci) + 1] - c_origin)},
+                        static_cast<SPValueType>(half * im * (*vi - ma::conj(*vi)))); // Lii - Lii*
```
PDoakORNL, line 419: this should be `static_cast<decltype(vn)::value_type>`

PDoakORNL (changes requested):
I think this can be made a bit simpler and more readable by reexamining the typealias choices made in config.h.

csr_matrix lacks meaningful documentation but pretty clearly is primarily templated on a VALUE type

PR #4631, opened 2023-06-16
Handle the special case of empty spin channels in Nexus

### LINKED ISSUE #4630: Hydrogen atomic run is aborted due to absence of down spin electrons
**Describe the bug**
A Hydrogen atom was simulated in a PBC box using QE. The QMCPACK run aborts due to no down spin electrons present in the input. Line 180 in `src/Particle/ParticleIO/XMLParticleIO.cpp` throws an error:
``
Fatal Error. Aborting at Some 'group' XML element node doesn't contain a 'size' attribute!
``

**To Reproduce**
1. Version: QMCPACK 3.16.9 (using batched VMC drivers)
2. The files are copied to OLCF and can be found at: `/gpfs/alpine/mat151/proj-shared/bugs/boxed_H`

**Expected behavior**
I realize this is an exactly solvable system, but in general, one would expect to be able to run systems without down spin electrons. For example, isovalent elements (Li, Na, K, ...) or the excited state of Be (2s^1 2p^1) will probably show a similar problem.

**System:**
 - NERSC/Perlmutter

**Additional context**
The problem was encountered when trying to study the `rcut` parameter in H for use in hybrid orbital representation. I checked the `pwscf.pwscf.h5` file and both spin information were written by QE. So everything seems to be OK on the QE side.

aannabe:
## Proposed changes

This PR should fix #4630. Nexus will no longer write the spin channels if they are empty.

- Bugfix

- No

NERSC/Perlmutter: All Nexus tests are passing.

## Checklist

### DIFF
--- a/nexus/lib/qmcpack_input.py
+++ b/nexus/lib/qmcpack_input.py
@@ -4456,14 +4456,19 @@ def generate_particlesets(electrons   = 'e',
     eup  = elns.up_electron
     edn  = elns.down_electron
 
+    particleset_groups = []
+    if eup.count > 0:
+        particleset_groups.append(group(name=uname,charge=-1,mass=eup.mass,size=eup.count))
+    #end if
+    if edn.count > 0:
+        particleset_groups.append(group(name=dname,charge=-1,mass=edn.mass,size=edn.count))
+    #end if
+
     particlesets = []
     eps = particleset(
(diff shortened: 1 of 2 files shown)


ye-luo (approved):

ye-luo:
Test this please

PR #4634, opened 2023-06-19
[WIP] Refactor TrialWaveFunction::ratio function to template form

williamfgc:
## Proposed changes

Refactors the `public virtual TrailWaveFunction::ratio` function to devirtualize the public call and move implementation to a derived class private area.
Add MACRO to define virtual functions in private/protected scope.
Introduce tcc files for template function implementations in private/protected scope.
Tests pass locally for complex and real builds. 
Do not merge, need to pass CI.
Part of the refactoring effort towards #4099 
See [gist](https://gist.github.com/williamfgc/28ffbd23f6888ed89530e3f68c992c93#refactoring-strategy)
 for the refactoring strategy this PR is based upon.

- No

Ubuntu 22.04

## Checklist


ye-luo:
It doesn't really make sense to have single WFC object to simultaneously have `ration()` support `float`, `double`, `complex<float>`, `complex<double>` four return types. The internal storage type should limit the API type to either a real or complex.

conceptually `TWF=WFC[0] * WFC[1] * ...`

```
template<TAPI>
WFC
{
   virtual TAPI ratio(P, iat);
};

template<TAPI, T> #The implementation can have internal template type T.
Determinant: public WFC<T>
{
}

template<TAPI>
TWF
{
  std::vector<WFC<TAPI>> Z;
};

template<TAPI>
OperatorBase
{
  evaluate(TWF<TAPI>)
};
```

ye-luo:
> @ye-luo thanks, this is something think about later, changing the class would make PRs intractable as I'd have to modify all consumers. I'm trying keep refactoring PRs as minimal as possible. This is the current pattern for `ratio` just it's defined at compile-time. Thoughts?

I appreciate your effort but I don't see the value of this PR even as an intermediate step because I feel it doesn't bring us closer to where we want to go and add maintenance hell.

In my pseudo code, you may start with changing WFC but keeping TWF and OperatorBase with hard-coded types. You don't even need to fix all the compilation but just make TWF.cpp WFC.cpp and one of WFC derived class.cpp to compile as a proof of concept.

williamfgc:
@ye-luo I need to understand better the proposed pattern, the current classes are not template classes. The overall value and goal is on replacing the compile-time defined aliases `ValueType` defined [here](https://github.com/QMCPACK/qmcpack/blob/develop/src/type_traits/QMCTypes.h#L39) one at a time which is what's preventing a unified real/complex build as it propagates throughout the code. This is a "minimal  proof-of-concept PR" for what's needed to change for one current function `ValueType ratio` --> `template <class T> T ratio` in which implementations of T for real and complex must coexist at runtime. We might need to discuss offline to come up with a strategy.

ye-luo:
> @ye-luo I need to understand better the proposed pattern, the current classes are not template classes. The overall value and goal is on replacing the compile-time defined aliases `ValueType` defined [here](https://github.com/QMCPACK/qmcpack/blob/develop/src/type_traits/QMCTypes.h#L39) one at a time which is what's preventing a unified real/complex build as it propagates throughout the code. This is a "minimal proof-of-concept PR" for what's needed to change for one current function `ValueType ratio` --> `template <class T> T ratio` in which implementations of T for real and complex must coexist at runtime. We might need to discuss offline to come up with a strategy.

Let us discuss prones and cons offline.

williamfgc:
Superseeded by #4684

PR #4638, opened 2023-06-21
Allow the use of "cell" and "corner" parameters in spindensity qdens analysis

aannabe:
## Proposed changes

The `spindensity` estimator can define arbitrary regions in space for density to be sampled. So far, `qdens` ignored these parameters if they were present in the qmcpack input, resulting in an incorrect `*.xsf` density output. This PR implements this missing feature in `qdens` and adds the documentation. This is currently handled for the `*.xsf` type file only, as I am unfamiliar with the `chgcar` format.

- Bugfix
- New feature
- Documentation changes

- No

NERSC/Perlmutter: All Nexus tests are passing.

## Checklist

### DIFF
--- a/docs/analyzing.rst
+++ b/docs/analyzing.rst
@@ -1394,9 +1394,18 @@ input flags are:
                           File containing atomic structure (default=None).
     -g GRID, --grid=GRID  Density grid dimensions (default=None).
     -c CELL, --cell=CELL  Simulation cell axes (default=None).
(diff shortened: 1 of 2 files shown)


aannabe:
Added an extra commit that handles cases where there is only a single spin channel in the `*.stat.h5` file, such as in the H atom.

ye-luo (approved):

ye-luo:
Test this please

PR #4640, opened 2023-06-22
Fix MSD when there is only 1 det in a given spin channel.

### LINKED ISSUE #4639: Multiple determinants with detlist of size 1 fails for precomputed_table_method
The input file contains a multideterminant with a detlist of size 1.   With 'algorithm="precomputed_table_method"', the resulting run has an acceptance ratio of zero.  With 'algorithm="computed_table_method"', the acceptance ratio looks much better and energy appears correct. 

The attached input has a detlist of size 2, but the second determinant has a coefficient of 0.0, which is less than the cutoff.  The same behavior results if the detlist is specified as size 1.

The test case was taken from the Be_STO test, and should get the same energy (-14.573)
The output of the qmcpack run is in out.txt.  The output of qmca is in output_from_qmca.txt.

[be_sto_multdet.tar.gz](https://github.com/QMCPACK/qmcpack/files/11835767/be_sto_multdet.tar.gz)

ye-luo:
## Proposed changes
Fixes #4639

- Bugfix

- No

laptop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
+++ b/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
@@ -155,38 +155,40 @@ void MultiDiracDeterminant::mw_buildTableMatrix_calculateRatios_impl(
       data_offset += (*ndets_per_excitation_level_)[ext_level] * (3 * ext_level + 1);
     };
 
+    mw_updateRatios_det0(det0_list, ratios_deviceptr_list);
+
     if (max_ext_level >= 1)
     {
-      mw_updateRatios<1>(det_offset, data_offset, data, sign, det0_list, table_matrix_deviceptr_list,
-                         nb_cols_table_matrix, ratios_deviceptr_list);
+      mw_updateRatios<1>(det_offset, data_offset, data, sign, table_matrix_deviceptr_list, nb_cols_table_matrix,
+                         ratios_deviceptr_list);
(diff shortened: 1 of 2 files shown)


src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
```diff
@@ -1128,22 +1142,17 @@ void MultiDiracDeterminant::mw_updateRatios(const size_t det_offset,
...
 
-  PRAGMA_OFFLOAD("omp target teams distribute parallel for collapse(2) map(always, to: det0_list_ptr[:nw])")
+  PRAGMA_OFFLOAD("omp target teams distribute parallel for collapse(2)")
   for (size_t iw = 0; iw < nw; iw++)
     for (size_t count = 0; count < ndet_ext; ++count)
     {
-      size_t det_id = det_offset + count;
-      ValueType ratios_local;
-      ///Initialization here to avoid one additional transfer and allow the use of collapse(2)
-      ratios_list_ptr[iw][0] = det0_list_ptr[iw];
```
ye-luo, line 1141: This was the bug and my original comment explains part of the cause.
mw_updateRatios<EXT_LEVEL> is only called when there is excitation.
Simply move the assignment out.

ye-luo:
Test this please

markdewing (approved):
Verified the fix on my build.  Thanks!

Issue #4637, opened 2023-06-20
Batched driver input does not catch/abort on legacy estimator input


jtkrogel:
**Describe the bug**
When the batched drivers are used, estimators are expected either in the driver input sections or in a dedicated <estimators> block.  Currently legacy estimator input (in <hamiltonion>) is simply ignored.  This needs to be fixed before the batched drivers become the default.

Issue originally reported by Dan Staros.

**To Reproduce**
Run with legacy estimator inputs along with the batched drivers

**Expected behavior**
An informative error message (instructions on correct placement of estimators in <estimators> block).

PDoakORNL:
I think this would be possible to catch, but what needs to happen long term is that the higher levels of the input system need to adopt something like the input objects and InputSection code.  Then no nonsense input of any sort whether typo or 
input in the wrong parent node can fall through.

PDoakORNL:
You can add almost any input you want in the wrong place or with a slightly wrong spelling of the tag or name and its just ignored by the code if its outside of the "<Estimators>" block in the batched code. The only sustainable way to deal with this class of issue is to rewrite the rest of the input system. So this might be a useful for transitioning users to batched but we should decide on a limit to how much of this we will do in the legacy input code.

jtkrogel:
@PDoakORNL what you are saying is true.  However, the placement of estimators is a major change going from legacy to batched and as such it is going to catch a lot of users.  I would suggest we put in a guard (error message + abort) when <estimator/>'s appear in <hamiltonian/> for the batched code.

ye-luo:
Please provide the input xml so we can see if the input style is supported or not and if yes, are we missing support.

jptowns:
Bump in 2024. I just tried to run a spindensity with batched drivers but using the legacy style input. Currently the manual has only legacy input description, and has no information on batched. QMCPACK not only does /not/ crash with legacy estimator input in the hamiltonian, but the output actually shows that the estimators were added. However, the corresponding stat.h5 files are empty. 

Example input/output attached (kindof a big problem so not good for fast debugging, but you'll get a flavor at least)
[estimator_example.tgz](https://github.com/QMCPACK/qmcpack/files/14439783/estimator_example.tgz)

PR #4643, opened 2023-06-22
Don't allow batched version to accept hamiltonian estimator input

PDoakORNL:
## Proposed changes

Prevent legacy estimator inputs in batched input.
This addresses issue #4637
 
- New feature

- Yes if a user has these previous ignored input elements in the hamiltonian an exception is thrown and the code halts.

osx laptop

## Checklist


src/QMCHamiltonians/HamiltonianFactory.cpp
```diff
@@ -176,8 +177,18 @@ bool HamiltonianFactory::build(xmlNodePtr cur)
         targetH->addOperator(std::move(hs), "Grid", true);
       }
     }
+    else if (cname == "flux")
+    {
+      if (potName.size() < 1)
```
quantumsteve, line 182:
```suggestion
      if (potName.empty())
```

src/QMCApp/QMCMain.cpp
```diff
@@ -446,7 +446,18 @@ bool QMCMain::validateXML()
...
+      switch (my_project_.getDriverVersion())
+      {
+      case ProjectData::DriverVersion::LEGACY:
+        batched = false;
+        break;
+      case ProjectData::DriverVersion::BATCH:
+        batched = true;
+        break;
+      }
+      ham_pool_->put(cur, batched);
```
ye-luo, line 460: Could the following work?
```
ham_pool_->put(cur, my_project_.getDriverVersion() == ProjectData::DriverVersion::BATCH);
```
If it works, please change the following place as well.

src/QMCHamiltonians/tests/MinimalHamiltonianPool.h
```diff
@@ -45,7 +45,7 @@ class MinimalHamiltonianPool
     doc.parseFromString(hamiltonian_xml);
 
     xmlNodePtr root = doc.getRoot();
-    hpool.put(root);
+    hpool.put(root,true);
```
ye-luo, line 48: Please apply clang-format. There should be a space after the comma.

src/QMCHamiltonians/HamiltonianFactory.cpp
ye-luo: I'm thinking of adding `legacy_estimator`? there is no guarantee we make the legacy ones working but at least users may have a chance to force it.

src/QMCHamiltonians/HamiltonianFactory.cpp
```diff
@@ -176,8 +177,18 @@ bool HamiltonianFactory::build(xmlNodePtr cur)
...
+	potName = "flux";
+      targetH->addOperator(std::make_unique<ConservedEnergy>(), potName, false);
+    }
     else if (cname == "estimator")
```
ye-luo, line 186: Change here like `cname == "estimator" || cname == "legacy_estimator"` and the error check as
`if (batched && cname == "estimator")`

tests/solids/diamondC_2x1x1_pp/qmc_short_vmcbatch_dmcbatch.in.xml
```diff
@@ -80,7 +80,7 @@
          <pairpot type="pseudo" name="PseudoPot" source="ion0" wavefunction="psi0" format="xml">
             <pseudo elementType="C" href="C.BFD.xml"/>
          </pairpot>
-         <!-- <estimator type="flux" name="Flux"/> -->
+         <!-- <flux name="Flux"/> -->
```
ye-luo, line 83: Delete this line.

tests/solids/diamondC_2x1x1_pp/qmc_short_vmcbatch_dmcbatch_4r.in.xml
```diff
@@ -80,7 +80,7 @@
          <pairpot type="pseudo" name="PseudoPot" source="ion0" wavefunction="psi0" format="xml">
             <pseudo elementType="C" href="C.BFD.xml"/>
          </pairpot>
-         <!-- <estimator type="flux" name="Flux"/> -->
+         <!-- <flux name="Flux"/> -->
```
ye-luo, line 83: Delete this line.

src/QMCApp/QMCMain.cpp
PDoakORNL: Done

PDoakORNL (commented):
Addressed review

src/QMCHamiltonians/HamiltonianFactory.cpp
PDoakORNL: I disagree with this change, legacy estimators that are going to continue to be allowed should be explicitly changed as the flux estimator has been.  They are not "estimators" by the definition we will be using coming forward.  They should be renabled individually if necessary. Further discussion of this should return to #4637 which this PR addresses, if this is the functionality you think is correct argue for it in #4637 with @jtkrogel  and amend this PR or make another one.

PDoakORNL (commented):
I consider this PR complete and do not have the bandwidth to continue I consider it a straight forward solution to #4637.

prckent:
Taking a look at this, we are not going to add legacy_estimator. We need to focus on estimators for the batched code and not get entangled with supporting more legacy. This will take time, resources and complexity we should simply not engage with. Users should use updated estimators for the batched code or simply run an older version until their needed estimator is available.

@jtkrogel Are you OK with `<flux name="Flux"/>` and the changes here? My opinion is that this is mergable now.

prckent:
Test this please

jtkrogel:
I don't understand the need for changing the notation around "flux" (`<flux/>`).  It is a statistical estimator (`<estimator/>`).  It should be ported (or not) to the batched code under the same conditions as all the rest.

For the purpose of this PR, the only requirement I see is that `<estimator/>`'s are banished from `<hamiltonian/>`.  The priority around moving other estimators to the batched code framework is separate.

Perhaps we just learned that "flux" is a higher priority estimator for porting than others.  I don't think that should hold up the rest of the work here.

ye-luo:
I feel unnecessary to introduce `<flux name="Flux"/>`. If we decide to allow such estimators still in the Hamiltonian input section, having `legacy_estimator` input style is more generic. Users needs to opt-in if they really want to try such a feature and the experience is allowed to be a hit and miss. In the current situation, we block everything and porting is slow.

src/QMCHamiltonians/HamiltonianFactory.cpp
ye-luo: > I'm not in favor of that as it will inhibit refactoring QMCHamiltonian. If there is a batched version the legacy version should be disabled for a batched run.

If there is a batched version, the old input through Hamiltonian should be blocked. What if not? I would prefer giving user a chance to try out things without modifying the source code.

src/QMCHamiltonians/HamiltonianFactory.cpp
ye-luo: Changing to `flux` or `legacy_estimator`, I prefer the latter a more generic way.
We can only allow using `legacy_estimator` when batched driver is in use and the selected estimator doesn't have an actual batched implementation.

PDoakORNL:
Stake holders could not reach consensus on this work which is now completely out of date.  
#4637 is political issue not amenable to a source level solution.

Issue #4633, opened 2023-06-17
how to continue a dmc run from a vmc or dmc outputs


YanyongW91:
**Calculation Purpose**
Basically I want to divide an complete time-extrapolation DMC calculation (vmc --> dmc [timestep: 0.04] --> dmc [timestep: 0.02] --> dmc [timestep: 0.01] ) into four jobs by series.

First I do the first-step vmc run, with ` <qmc method="vmc" move="pbyp" checkpoint="0">` in the input xml file according to the manual, then I get the corresponding dmc.s000.config.h5, dmc.s000.qmc.xml, dmc.s000.random.h5 that needed for the next dmc run.

**Describe the bug**
However, after I added `</qmcsystem><mcwalkerset fileroot="dmc.s000" node="-1" nprocs="16" version="3 16" collected="yes"/>` in the input xml for the next dmc run, it rerun the vmc part from the very beginning (producing s004, s005, s006, s005 files. respectively), while what I want is to continue the run for the next dmc block, not repeat the vmc run.

**System:**
qmcpack version: `3.16.0`, cpu
platform: nersc perlmutter
job script setting (number of cpus, processes..)  keeps the same

**Additional questions**
If I follow the same way but modify the input xml file with `</qmcsystem><mcwalkerset fileroot="dmc.s001" node="-1" nprocs="16" version="3 16"` to start from a dmc run (`dmc.s001.config.h5, dmc.s001.qmc.xml, dmc.s001.random.h5` provided), it reports error:
`Fatal Error. Aborting at SimpleFixedNodeBranch::resetRun cannot initialize WalkerController`
and the final lines in the dmc.out file are
```
QMC Execution time = 5.9797e-01 secs
BRANCH resetRun
 SimpleFixedNodeBranch::resetRun detected changes in <parameter>'s 
 BranchMode : 0001001111 0001001111
```
Please help! If there's practical example for rerun, that would be helpful!

YanyongW91:
Thanks! I do need this function due to time limits on perlmutter. What I missed is that I need to comment out already completed vmc or dmc section. I also checked that when setting checkingpoints to a limit number of blocks, qmcpack still only write out the random.h5 and config.h5 files only if it has finished an independent vmc or dmc section run, which is against what the manual stated. I don't know where goes wrong.

prckent:
[...]

PR #4646, opened 2023-06-26
Simplify checkpointing, change defaults, fix legacy, implement in batched

prckent:
## Proposed changes

This PR aims to simplify and improve the checkpointing process.

As mentioned in https://github.com/QMCPACK/qmcpack/issues/4633 the documented every N blocks checkpointing feature was not working as expected. In legacy, this required storeconfigs to also be specified to work (undocumented/unexpected).  Checkpointing was also not implemented in the batched drivers.

Here:

- checkpoint={-1,0,N} driver parameter option sufficient to control checkpointing
- Default changed to checkpoint=0. Configs will be written at the end of each QMC section by default. -1 disables. N sets write period in addition to end of section write.
- Removed storeconfigs option completely.
- Implemented checkpointing in batched code.
- Updated docs appropriately

Have verified by eye that the checkpoints look sensible and are updated periodically. Open to testing suggestions.

Batched code does not create a info.xml as part of checkpoint.

Also remove unused FastGrad option.

- Bugfix
- New feature

- No

sulfur gcc build only. PR for CI.

## Checklist

- Yes/ This PR is up to date with current the current state of 'develop'
- (Existing restart tests use code paths). This PR adds tests to cover any new code, or to catch a bug that is being fixed

### DIFF
--- a/docs/methods.rst
+++ b/docs/methods.rst
@@ -36,7 +36,7 @@ Quantum Monte Carlo Methods
   +----------------+--------------+--------------+-------------+---------------------------------+
   | ``profiling``  | text         | yes/no       | no          | Activate resume/pause control   |
   +----------------+--------------+--------------+-------------+---------------------------------+
-  | ``checkpoint`` | integer      | -1, 0, n     | -1          | Checkpoint frequency            |
+  | ``checkpoint`` | integer      | -1, 0, n     | 0           | Checkpoint frequency            |
   +----------------+--------------+--------------+-------------+---------------------------------+
   | ``record``     | integer      | n            | 0           | Save configuration ever n steps |
   +----------------+--------------+--------------+-------------+---------------------------------+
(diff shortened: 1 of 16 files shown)


ye-luo:
Based on the documentation the default value of checkpoint is -1, should it be better to have 0 as default, namely end of driver section?

prckent:
I agree. Pushed change and will update PR text

ye-luo:
@prckent your fix just bring me attention that we don't have the corresponding tests for batched drivers.

ye-luo (approved):

prckent:
Agree we are missing some tests here. Which were you thinking of? batched dmc restart tests? or something else?

I don't think that we have _any_ tests on checkpoints saved within a section (as opposed to at the end), whether legacy or batched. The restart tests use the "checkpoints" saved at the end of a section. It is not obvious how to reliably catch the checkpoints written within a section for testing.

ye-luo:
> Agree we are missing some tests here. Which were you thinking of? batched dmc restart tests? or something else?
> 
> I don't think that we have _any_ tests on checkpoints saved within a section (as opposed to at the end), whether legacy or batched. The restart tests use the "checkpoints" saved at the end of a section. It is not obvious how to reliably catch the checkpoints written within a section for testing.

That types of tests can only be made on the unit test side. We can simply run a driver with coordinates dumping in the middle, then load it back without using a driver and check against fixed numbers.

prckent:
Test this please

PR #4657, opened 2023-06-30
use int64_t when aggregating accept/reject counts.

### LINKED ISSUE #4652: Unexpected VMC acceptance ratio behavior
**Describe the bug**
The linear molecule C4N2 was simulated using QE and batched VMC with PBC:
C4N2

As the total number of walkers is increased via increasing the number of nodes in the job while leaving the QMC input the same, the following unexpected behavior occurs in the VMC acceptance ratio in the large walker limit for 500 steps/block:
acceptance

To check if it was related to #4648, I ran with `crowd_serialize_walkers=yes`, and this didn't resolve the problem, as can be seen above. However, decreasing the number of steps to 300/block or 50/block resolves the issue.

I also checked to total energies and they seem unaffected by all of this:
energy

**To Reproduce**
Steps to reproduce the behavior: Relase Version 3.16.9
Input and output files are copied to OLCF at:
`/gpfs/alpine/mat151/proj-shared/bugs/issue_acceptance`

**Expected behavior**
The acceptance ratio should stay the same as the total number of walkers is increased.

**System:**
- NERSC/Perlmutter. CPU run with batched-VMC

**Additional context**
In this case, it was desired to keep the number of steps as high as possible since `spindensity` was being accumulated, which dumps large amounts of data every block. So by increasing the number of steps, the statistics can be improved without large files.

ye-luo:

## Proposed changes
Fixes #4652

- Bugfix

- No

## Checklist

### DIFF
--- a/src/Estimators/EstimatorManagerNew.cpp
+++ b/src/Estimators/EstimatorManagerNew.cpp
@@ -14,6 +14,7 @@
 #include <numeric>
 #include <algorithm>
 #include <variant>
+#include <cstdint>
 
 #include "EstimatorManagerNew.h"
 #include "SpinDensityNew.h"
@@ -57,7 +58,7 @@ bool EstimatorManagerNew::areThereListeners() const
   return std::any_of(operator_ests_.begin(), operator_ests_.end(),
                      [](auto& oper_est) { return oper_est->isListenerRequired(); });
 }
-  
+
 template<class EstInputType, typename... Args>
 bool EstimatorManagerNew::createEstimator(EstimatorInput& input, Args&&... args)
 {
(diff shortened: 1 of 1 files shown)


ye-luo:
@aannabe could you test my fix? Thanks.

ye-luo:
hold a second.

ye-luo:
@aannabe should be ready for testing.

aannabe:
@ye-luo Thanks; it resolved the issue. With the largest walkers (143360), I now get the correct values:
```
vmc_J123  series 0 
  LocalEnergy           =        -42.170104 +/-         0.000045    1.6 
  Variance              =            0.7814 +/-           0.0014    1.0 
  Kinetic               =          32.01800 +/-          0.00034    1.0 
  LocalPotential        =         -74.18811 +/-          0.00034    1.0 
  AcceptRatio           =         0.5352537 +/-        0.0000018    1.0
```

prckent (approved):
Thanks Ye for the prompt fix and Gani for checking. I think this works for now. It fixes the current problem and partly protects against the problem in general. We would still be vulnerable to a edge cases e.g. someone running an atom on a workstation for a week, so I think the general fix would be to declare a rule that all the counters should be 64 bit ints.

prckent:
Test this please

PR #4658, opened 2023-07-03
Improve some assignment operators

prckent:
## Proposed changes

A few too many consts / non-enough consts found using CodeCheck/Understand.

- Bugfix

- No

gcc12, mac laptop only.

## Checklist

### DIFF
--- a/src/Utilities/ResourceHandle.h
+++ b/src/Utilities/ResourceHandle.h
@@ -27,11 +27,7 @@ public:
 
   ResourceHandle() = default;
   ResourceHandle(RS& res) { Base::emplace(res); }
-  ResourceHandle& operator=(const RS& res)
-  {
-    Base::emplace(res);
-    return *this;
-  }
+
   bool hasResource() const { return Base::has_value(); }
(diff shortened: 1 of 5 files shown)


prckent:
Test this please

src/Utilities/ResourceHandle.h
```diff
@@ -27,7 +27,7 @@ class ResourceHandle : private std::optional<std::reference_wrapper<RS>>
 
   ResourceHandle() = default;
   ResourceHandle(RS& res) { Base::emplace(res); }
-  ResourceHandle& operator=(RS& res)
+  ResourceHandle& operator=(const RS& res)
```
ye-luo, line 30: This is not a conventional assignment operator. Resources needs be modifiable and thus it should not carry const. I noticed this function not being used anywhere, we can probably just delete it.

prckent: Removed

prckent:
Test this please

ye-luo (approved):

PR #4661, opened 2023-07-06
Fix hdf path empty string handling

prckent:
## Proposed changes

Access to potentially empty/zero length strings was not fully protected. Not likely to be hit in QMCPACK execution but does fix asan debug builds of unit tests.

Also increase ion-ion error bar for NiO48, tripped in debug builds (gcc11).

Code could be written more concisely but is easier to read + place breakpoints in with current format.

- Bugfix

- No

ntirogen2, llvm15 and gcc11

## Checklist


src/io/hdf/hdf_path.cpp
```diff
@@ -23,8 +23,18 @@ hdf_path& hdf_path::append(const hdf_path& p)
...
+  else
+  {
+    if (path_.size() > 0)
+    {
+      if (path_.back() != '/')
+        path_.push_back('/');
+    }
+    else
+    {
+      path_.push_back('/');
```
ye-luo, line 35: This line adds a new execution behavior. Do you have an example explaining that this is line is needed?

src/io/hdf/hdf_path.cpp
```diff
@@ -23,8 +23,18 @@ hdf_path& hdf_path::append(const hdf_path& p)
...
+    }
+  }
```
ye-luo, line 37: The whole code block can be simpler.
```
  if (p.has_root_directory())
    path_.clear();

  if (!path.empty() && path_.back() != '/')
    path_.push_back('/');

  return concat(p);
```

src/io/hdf/hdf_path.cpp
```diff
@@ -70,7 +80,18 @@ hdf_path& hdf_path::replace_subgroup(std::string_view p)
   return append(p);
 }
 
-bool hdf_path::has_root_directory() const { return (!path_.empty() && path_[0] == '/'); }
```
ye-luo, line 73: The old code seems solid to me. Based on C/C++ rule, if the first condition failed, the second should not run.

ye-luo (changes requested):

src/io/hdf/hdf_path.cpp
prckent: Will fix.

prckent:
I'll make this longer than strictly necessary since it is imo interesting and there are some subtleties. I am not sure if I am correct here, but I do know that the PR works for me while the old code does not.

I tried out your suggested combination and unfortunately it does not work:

```
hdf_path& hdf_path::append(const hdf_path& p)
{
if (p.has_root_directory())
    path_.clear();

if (!path_.empty() && path_.back() != '/')
    path_.push_back('/');

return concat(p);
}

bool hdf_path::has_root_directory() const
{
 return (!path_.empty() && path_[0] == '/');
}
```

My possibly incorrect understanding is that we do not have any guarantee of expression evaluation order for functiona() && functionb(), and since functionb has an unsafe reference we get into troubles -- it is just unsafe to write checkifsafe() && maybedangeroustest(). And even if it is OK under some circumstances we should keep the code simpler and avoid murky features. In this case we just get an unsafe memory reference, so the debug/asan build fails.

I tried the above with just my replacement for has_root_directory

```
bool hdf_path::has_root_directory() const
{
  bool have_root = false;
  if (!path_.empty())
  {
    if (path_[0] == '/')
    {
      have_root = true;
    }
  }
  return have_root;
}
```
and we still get a failure, as I was expecting.

if we also add the following

```
hdf_path& hdf_path::append(const hdf_path& p)
{
  if (p.has_root_directory())
    path_.clear();
  else
  {
   if (path_.size() > 0)
    {
      if (path_.back() != '/')
        path_.push_back('/');
    }
    else
    {
     path_.push_back('/');
    }
  }

  return concat(p);
}
```

Then it works. I don't see the change in behavior from the original here. In the old code we add a '/' if has_root_directory() is false. The key change is that in the new code we avoid asking for the last character of the string (calling path_.back()) if the string is zero length. And in all cases we then add on the user provided p.

prckent:
Just a sec.

prckent:
Paging @quantumsteve for comment on the intended behaviors.

quantumsteve:
@prckent Would you add a failing test [here](
https://github.com/QMCPACK/qmcpack/blob/eda7abeaa098b7f941f8ad28ac9196c00c385036/src/io/hdf/tests/test_hdf_path.cpp#L36)

prckent:
The existing tests fail. Below is output from current develop
```
$ ctest -R unit_test_io_hdf5 --output-on-failure
Test project /home/pk7/projects/qmc/git_QMCPACK_prckent/qmcpack/build
    Start 1: deterministic-unit_test_io_hdf5
1/1 Test #1: deterministic-unit_test_io_hdf5 ...Subprocess aborted***Exception:   0.07 sec
QMCPACK printout is suppressed. Use --turn-on-printout to see all the printout.
/builddir/build/BUILD/gcc-11.3.1-20221121/obj-x86_64-redhat-linux/x86_64-redhat-linux/libstdc++-v3/include/bits/basic_string.h:1137: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::back() [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference = char&]: Assertion '!empty()' failed.

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test_io_hdf5 is a Catch v2.13.9 host application.
Run with -? for options

-------------------------------------------------------------------------------
hdf_path_append
-------------------------------------------------------------------------------
../src/io/hdf/tests/test_hdf_path.cpp:36
...............................................................................

../src/io/hdf/tests/test_hdf_path.cpp:36: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

===============================================================================
test cases:  15 |  14 passed | 1 failed
assertions: 109 | 108 passed | 1 failed



0% tests passed, 1 tests failed out of 1

Label Time Summary:
deterministic      =   0.07 sec*proc (1 test)
quality_unknown    =   0.07 sec*proc (1 test)
unit               =   0.07 sec*proc (1 test)

Total Test time (real) =   0.12 sec

The following tests FAILED:
          1 - deterministic-unit_test_io_hdf5 (Subprocess aborted)
Errors while running CTest
$ 
```

prckent:
```
$ export OMP_NUM_THREADS=1
$ ./test_io_hdf5  --turn-on-printout
==================================================
--- Memory usage report : Before running tests ---
==================================================
Available memory on node 0, free + buffers :  191132 MiB
Memory footprint by rank 0 on node 0       :      29 MiB
==================================================
/builddir/build/BUILD/gcc-11.3.1-20221121/obj-x86_64-redhat-linux/x86_64-redhat-linux/libstdc++-v3/include/bits/basic_string.h:1137: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::back() [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference = char&]: Assertion '!empty()' failed.

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test_io_hdf5 is a Catch v2.13.9 host application.
Run with -? for options

-------------------------------------------------------------------------------
hdf_path_append
-------------------------------------------------------------------------------
../src/io/hdf/tests/test_hdf_path.cpp:36
...............................................................................

../src/io/hdf/tests/test_hdf_path.cpp:36: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

===============================================================================
test cases:  15 |  14 passed | 1 failed
assertions: 109 | 108 passed | 1 failed

Aborted (core dumped)
```

src/io/hdf/hdf_path.cpp
quantumsteve, line 35:
```suggestion
    if (path_.empty())
      path_.push_back('/');
    else if (path_.back() != '/')
      path_.push_back('/');
```
Is this code more clear?

quantumsteve: If `p` is an absolute path, it overwrites the string.

if `path_` is empty, we add `/` to create an absolute path and call concat.

otherwise we check to avoid a `//` and call concat.

quantumsteve: I tried to follow the same rules as `std::filesystem::path::append`
https://en.cppreference.com/w/cpp/filesystem/path/append

src/io/hdf/hdf_path.cpp
quantumsteve: I'm also struggling to understand why short-circuit evaluation isn't working here.
https://en.wikipedia.org/wiki/Short-circuit_evaluation

src/io/hdf/hdf_path.cpp
prckent: Both clearer and more concise, at least to me. I have pushed the change & verified it is OK locally, as expect.

ye-luo: > if path_ is empty, we add / to create an absolute path and call concat.

I'm not following this rule. If `path_` is empty and a user appends `p` which is a relative path, I don't see why adding a root `/` is valid.

quantumsteve:
```
If p.is_absolute() || (p.has_root_name() && p.root_name() != root_name()), then replaces the current path with p as if by operator=(p) and finishes.
```
https://en.cppreference.com/w/cpp/filesystem/path/append

src/io/hdf/hdf_path.cpp
```diff
@@ -23,8 +23,13 @@ hdf_path& hdf_path::append(const hdf_path& p)
 {
   if (p.has_root_directory())
     path_.clear();
-  else if (path_.back() != '/')
```
ye-luo, line 26: I can understand that accessing back() without protecting the empty string case can cause invalid access. I don't understand why asan still fails after adding a check of `empty`. The even more strange thing is that I cannot reproduce the asan failure locally. I was using 11.2 but Paul has 11.3. I'm building 11.4 now.

ye-luo:
> ```
> $ export OMP_NUM_THREADS=1
> $ ./test_io_hdf5  --turn-on-printout
> ==================================================
> --- Memory usage report : Before running tests ---
> ==================================================
> Available memory on node 0, free + buffers :  191132 MiB
> Memory footprint by rank 0 on node 0       :      29 MiB
> ==================================================
> /builddir/build/BUILD/gcc-11.3.1-20221121/obj-x86_64-redhat-linux/x86_64-redhat-linux/libstdc++-v3/include/bits/basic_string.h:1137: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::back() [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::reference = char&]: Assertion '!empty()' failed.
> 
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> test_io_hdf5 is a Catch v2.13.9 host application.
> Run with -? for options
> 
> -------------------------------------------------------------------------------
> hdf_path_append
> -------------------------------------------------------------------------------
> ../src/io/hdf/tests/test_hdf_path.cpp:36
> ...............................................................................
> 
> ../src/io/hdf/tests/test_hdf_path.cpp:36: FAILED:
>   {Unknown expression after the reported line}
> due to a fatal error condition:
>   SIGABRT - Abort (abnormal termination) signal
> 
> ===============================================================================
> test cases:  15 |  14 passed | 1 failed
> assertions: 109 | 108 passed | 1 failed
> 
> Aborted (core dumped)
> ```

The error message still lacks clarity. Usually with asan, and -g from debug build, I should see which line the segfault is from.

src/io/hdf/hdf_path.cpp
ye-luo: > ```
> If p.is_absolute() || (p.has_root_name() && p.root_name() != root_name()), then replaces the current path with p as if by operator=(p) and finishes.
> ```
> 
> https://en.cppreference.com/w/cpp/filesystem/path/append

not following. in my question, `p` is not absolute and has no root_name.

quantumsteve: I think you're right that `hdf_path` isn't following `std::filesystem::path` in that case. Hopefully nothing depends on this behavior and it's an easy change :crossed_fingers: 

std::filesystem::path::append doesn't add a `/` at the beginning.  
https://godbolt.org/z/981zhK4sv

src/io/hdf/hdf_path.cpp
prckent: FYI, these builds were with RHEL 9.2 entirely using RHEL provided packages for the compiler, libraries etc.

prckent:
Replaced by #4662

PR #4663, opened 2023-07-08
Use SYCL in-order queue.

ye-luo:
## Proposed changes
It seems that matrix inversion in MKL fails when using default out-of-order queue. Switching to in-order queues makes `test_syclSolverInverter.cpp` to pass reliably. In-order queues add less burden on GPU runtime and should have lower latency. QMCPACK uses multiple queues and doesn't really need out-of-order queue to gain kernel parallelism.

- Bugfix

- No

sunspot

## Checklist


ye-luo:
Test this please

prckent (approved):

PR #4674, opened 2023-07-13
Introduce QMC_DISABLE_HIP_HOST_REGISTER. ON by default.

ye-luo:
## Proposed changes
Since rocm 5.6. hipHostRegister becomes terribly slow https://github.com/ye-luo/miniqmc/wiki/hip-hsa-issues.
Introduce QMC_DISABLE_HIP_HOST_REGISTER. which can be used to make hipHostRegister no-op.
I set it ON by default for now. Will go through a few nights of performance test runs.

My guess is that it makes runs slower because we lose more from not using pinned memory than slower hipHostRegister in close  to production benchmarks.

- Other (please describe):

- No

epyc-server

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -842,6 +842,7 @@ if(ENABLE_ROCM)
     target_link_libraries(ROCM::libraries INTERFACE roc::rocsolver roc::rocrand)
   endif()
   message("Project HIP_FLAGS: ${CMAKE_HIP_FLAGS}")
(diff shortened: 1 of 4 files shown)


prckent (dismissed):

prckent:
It will be interesting to see if this helps or hinders the full range of system sizes run in the nightlies.

prckent:
Test this please

ye-luo:
Test this please

prckent (approved):
LGTM

PR #4677, opened 2023-07-18
Introduce qmcplusplus::isnan

ye-luo:
## Proposed changes
Introduce custom isnan to make it always work.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Platforms/CPU/CMakeLists.txt
+++ b/src/Platforms/CPU/CMakeLists.txt
@@ -2,14 +2,17 @@
 #// This file is distributed under the University of Illinois/NCSA Open Source License.
 #// See LICENSE file in top directory for details.
 #//
-#// Copyright (c) 2020 QMCPACK developers.
+#// Copyright (c) 2023 QMCPACK developers.
 #//
 #// File developed by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
 #//
 #// File created by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
 #//////////////////////////////////////////////////////////////////////////////////////
 
-add_library(platform_cpu_runtime INTERFACE)
+# To enable a custom qmcplusplus::isnan, remove fast-math. Only the file scope CMAKE_CXX_FLAGS gets affected.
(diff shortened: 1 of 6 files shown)


ye-luo:
Test this please

PDoakORNL:
So we need to support some platform without `std::isnan`?

ye-luo:
> So we need to support some platform without `std::isnan`?

When `-fffast-math` is enabled, std::isnan always return false and I cannot implement safe-guards. qmcplusplus::isnan remains effective.

prckent (changes requested):
This will be useful to have since unfortunately there are sightings of NaN in optimized builds where std::isnan is not helpful. AFAICT checking a!=a is the preferred way to do this.

Can you please add comments to the changes in math.hpp,.cpp explaining the functions?

prckent (approved):

prckent:
Test this please

src/Platforms/CPU/CMakeLists.txt
```diff
@@ -2,14 +2,17 @@
...
+#// Copyright (c) 2023 QMCPACK developers.
 #//
 #// File developed by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
 #//
 #// File created by: Ye Luo, yeluo@anl.gov, Argonne National Laboratory
 #//////////////////////////////////////////////////////////////////////////////////////
 
-add_library(platform_cpu_runtime INTERFACE)
+# To enable a custom qmcplusplus::isnan, remove fast-math. Only the file scope CMAKE_CXX_FLAGS gets affected.
+string(REPLACE " -ffast-math" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
```
markdewing, line 13: Do all compilers use a flag with the same name to perform the optimization that affects isnan?

ye-luo: All the compilers that have `-ffast-math` behave the same. I explicitly tested nvhpc which doesn't support `-ffast-math` and passes the added tests.

prckent: In any case this PR still expands the range of cases for which we can catch a NaN. Hopefully the code path won't get much use...

PR #4682, opened 2023-07-21
Remove xxx_unit cmake targets

### LINKED ISSUE #3700: Get rid of USE_FAKE_RNG macro
USE_FAKE_RNG can be considered a hack to force fake RNG in unit tests.
It introduces internal forks in the source tree and caused chain reaction.

I would consider a base+derived class implementation. As long as RNGs are explicitly initialized for the tested object, there is no issue selecting true or fake RNGs.

I thought of templates on RNG but then it doesn't change the nature of two forks as all the types are enforced. So I prefer not to take this route.

ye-luo:
## Proposed changes
A big thank to @quantumsteve 
No more redundant compilation of qmcham qmcdriver qmcestimators.
Closes #3700.

- Build related changes

- No

epyc-server

## Checklist

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -107,11 +107,6 @@ if(IS_GIT_PROJECT)
   message("Git commit hash: ${GIT_CONFIG_COMMIT_HASH}")
 endif()
 
-# For unit tests, enable use for the fake RNG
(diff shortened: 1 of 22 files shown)


ye-luo:
Test this please

prckent (approved):
LGTM

PR #4684, opened 2023-07-24
[WIP] Implement SPOSetT<T> template class

williamfgc:
## Proposed changes

Introduce a shim template class to asses the effort to change `SPOSet` to `SPOSet<T>`.
`SPOSetT<T>` currently does not have consumers nor tests.
Testing compilation in CI
The challenge is to adapt `SPOSetT<T>` to `SPOSet` current consumers.
Related to #4099 

- No

sulfur 

## Checklist


williamfgc:
Changed to the option suggested by @ye-luo to only concretize class type.

williamfgc:
@ye-luo thanks for chiming in. We just thought having a separate SPOSetT hierarchy would help us with CI (at least build errors we introduce are caught) and for us to work in parallel. Once we have that we can start integrating with current SPOSet consumers without rewriting those and once done remove SPOSet and rename SPOSetT->SPOSet<T>. Thoughts?

ye-luo:
> @ye-luo thanks for chiming in. We just thought having a separate SPOSetT hierarchy would help us with CI (at least build errors we introduce are caught) and for us to work in parallel. Once we have that we can start integrating with current SPOSet consumers without rewriting those and once done remove SPOSet and rename SPOSetT->SPOSet. Thoughts?

probably I didn't fully understand the current status of this PR. I saw multiple files (SPOSetT and derived classes) got added and was thinking of how to make the change mergeable without the need of porting every SPOSet derived classes before merging. By making SPOSet a base of SPOSetT probably allows us to keep ported and not yet ported classes to co-exist without any duplication. All the high level classes still just see SPOSet.

src/Particle/RealSpacePositionsTOMPTarget.h
```diff
@@ -178,6 +177,7 @@ class RealSpacePositionsTOMPTarget : public DynamicCoordinatesT<T>
         const size_t mw_pos_stride = mw_new_pos.capacity();
 
         PRAGMA_OFFLOAD("omp target teams distribute parallel for \
+                    is_device_ptr(mw_pos_ptr, mw_rosa_ptr) \
```
williamfgc, line 181: @ye-luo @correaa FYI this passes the right values

williamfgc:
Test this please

williamfgc:
Test this please

williamfgc:
Test this please

markdewing:
Is there any chance of breaking this into separate PR's that are more reviewable? (That is, are the changes to individual classes separable, or they all linked?)
Is there a change in build time as a result of the extra templates?

quantumsteve:
> Is there any chance of breaking this into separate PR's that are more reviewable? (That is, are the changes to individual classes separable, or they all linked?)

Changing `SPOSet` to `SPOSetT<T>` meant modify all derived classes and builders. Beyond that we tried to use macros to contain this change to a minimum (but still large :crying_cat_face:) number of files.

markdewing:
The changes in the Numerics and Particle directories clearly can't depend on SPOSet.  And a lot of the changes in QMCDrivers are just minor changes like include files.

src/Containers/OhmmsSoA/VectorSoaContainer.h
```diff
@@ -220,6 +220,7 @@ struct VectorSoaContainer
   template<typename T1>
   void copyOut(Vector<TinyVector<T1, D>>& out) const
   {
+    static_assert(std::is_same_v<T, T1>);
```
quantumsteve, line 223: #4802 should eliminate this static_assert

src/QMCWaveFunctions/tests/test_RotatedSPOsT.cpp
```diff
@@ -0,0 +1,911 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source
+// License. See LICENSE file in top directory for details.
+//
+// Copyright (c) 2022 QMCPACK developers.
+//
+// File developed by: Joshua Townsend, jptowns@sandia.gov, Sandia National
```
markdewing, line 7: It looks like someone is running clang-format with a column limit of 80 (vs the qmcpack default of 120). (Or is using an editor that does auto-wrapping).  While technically within the guidelines (line length is < 120), it is creating extra differences between files.

markdewing:
It looks like the "new" files have been run through clang-format that is more restrictive than the default one, or been modified in an editor that does some formatting.
What I've noticed so far:
 - Copyright header gets wrapped to 80 columns
 - Comments in general get wrapped to 80 columns
 - Space gets added between comment and text

I think it would be better if the copyright headers had a consistent format in every file. 

Unfortunately, these aren't new files - they're renamed and changed versions of existing files.
The other changes wouldn't be a problem in a small PR, but for one this large, every change is making it harder for git to detect moved files (It uses some threshold for the number of pairs of changes to check, or a percentage of file similarity).

quantumsteve:
`cmake -G Ninja -DQMC_MPI=OFF ../`

develop
```
$ time ninja
[646/646] Linking CXX executable bin/ppconvert

real	1m55.924s
user	64m36.053s
sys	5m4.636s
```

ref-add-SPOSetT

```
$ time ninja
[652/652] Linking CXX executable bin/ppconvert

real	2m5.956s
user	70m52.674s
sys	5m15.397s
```

src/QMCWaveFunctions/tests/test_RotatedSPOsT.cpp
```diff
@@ -0,0 +1,911 @@
...
+  using Type = Catch::Detail::Approx;
+};
+template<typename T>
+struct ValueApproxHelper<std::complex<T>>
+{
+  using Type = Catch::Detail::ComplexApprox;
+};
+
+template<typename T>
+using ValueApprox = typename ValueApproxHelper<T>::Type;
```
markdewing, line 49: This code might be useful elsewhere, and could get moved to src/external_codes/complex_approx.hpp or src/Configuration.h
(Not necessary for this PR, though)

williamfgc:
@markdewing thanks for the catch. The latest commit should have addressed clang-format related issues. I will double-check. Thanks!

williamfgc:
Test this please

williamfgc:
> The changes in the Numerics and Particle directories clearly can't depend on SPOSet. And a lot of the changes in QMCDrivers are just minor changes like include files.

A lot of these changes are at the `SPOSet<T>` consumer level (e.g. builders) and likely 'cause GitHub Actions CI was breaking as at some point we ran it in every commit the current branch has. It's probably the smallest PR we can get with such refactoring.

williamfgc:
Test this please

PDoakORNL:
What's going on with the `MCWalkerConfiguration` further bleeding into the batched drivers and mw is undesirable.  I've gone out of my way to keep this from happening. We want to drop that class when the legacy drivers and estimators go away.  It should never be a source for type information for non legacy code.

The explicit definition of MCPWalker from Walker instead of pulling in Walker_t from elsewhere was intended to decouple classes that were tightly bound via Walker type aliases. Then they can be built and tested without creating cascading rebuilds. There seems to be a clash now between QMCTraits and ParticleTraits which is making a bit of a mess that is just getting fixed by adding includes that shouldn't really be included.

Similar issues seem behind the many forward declarations that were successfully decoupling headers that have been reverted by this PR.

I'm not sure if we should put this in and then work to refactor away the new includes and MCPWalker type aliases or just stage this in such a way we don't add all these design regressions.

prckent (commented):
I assent to merge this subject to fixing the conflicts and there not being any new issues there.

If we are concerned about possible offload errors, we could checks this code is OK on Frontier and Aurora, so that we don't risk breaking develop for FOM runs. Our CI coverage of offload is not great (only NV, limited coverage).

Our earlier decision was that this would be a large PR instead of a lot of small ones since the large PR route was much more practical. As Peter said, we can work to refactor away some of the new design dependencies. Breaking this PR up would not be consistent with what we agreed on, and from a project perspective there simply is not time to rework this PR. There are very week weeks remaining in the year to get unified complex-real builds completed.

Next step: William & co to fix the conflicts.

williamfgc:
Test this please

src/Particle/ParticleSet.h
```diff
@@ -19,682 +19,12 @@
...
-    GroupID.resize(numPtcl);
-    G.resize(numPtcl);
-    L.resize(numPtcl);
-    Mass.resize(numPtcl);
-    Z.resize(numPtcl);
-
-    coordinates_->resize(numPtcl);
-  }
-};
+using ParticleSet = ParticleSetT<QMCTraits::ValueType>;
```
ye-luo, line 27: ParticleSet should be real-only.

ye-luo (commented):
First, I appreciated the effort. It is a very heavy workload.
After looking through the change. I feel two aspects are mixed when rushing to make the code compile and run.
1. real/complex. This is what we would like to explore.
2. full/mixed precision. This is what I would like to leave aside.
by looking at all the changes. I'm surprised to see ParticleSet being template as ValueType. When making such massive changes. Certain choices may not be made optimal. Merging this PR as it is will be a challenge for post-ECP develops as there is very limited understanding of this PR and some modifications are not necessary and needs undo. So I still would like to explore a way to make piece wise changing possible.

I have additional thoughts about how to handle mixed precision and also simplify the code. It will be a separate discussion.

williamfgc:
> full/mixed precision. This is what I would like to leave aside. by looking at all the changes. I'm surprised to see ParticleSet being template as ValueType.

This was done to satisfy the requirement to not specialize on a single function basis, but rather class level since `ParticleSet` is directly consumed by `SPOSet` and derived classes. If we start specializing it will look like lots of trailing `#ifdef` for handling mixed precision, and templates for real/complex. We think this is a good-enough compromise to move forward and not complicate current tasks given timelines. We didn't want to make complicated design changes (other than what we were asked) as we are relying on the current CI and many parts are not necessarily tested.

> Merging this PR as it is will be a challenge for post-ECP develops as there is very limited understanding of this PR and some modifications are not necessary and needs undo. So I still would like to explore a way to make piece wise changing possible.

The PR itself is about undoing things, and there is plenty of pending things (e.g. we are still depending of #ifdef for template class definitions). I'm afraid any other attempt to make this piece wise won't add value and we we will end up with something very similar to what we have (or worse, combining current non-generic #ifdef patterns with template generic patterns). This is a good opportunity to identify pieces towards major re-architecture redesign, refactoring and add more testing. This PR can be summarized as SPOSet -> SPOSet<T> + the most straight-forward/shortest-path way to satisfy CI requirements. My concern is on trying to spend time to maximize impact given that resources are limited.

src/Particle/DistanceTable.h
```diff
@@ -16,355 +16,13 @@
...
-
-  /// return stride of per target pctl data. full table data = stride * num of target particles
-  [[noreturn]] virtual size_t getPerTargetPctlStrideSize() const
-  {
-    throw std::runtime_error(name_ + " getPerTargetPctlStrideSize not supported");
-  }
-};
+using DistanceTable = DistanceTableT<QMCTraits::ValueType>;
+using DistanceTableAA = DistanceTableAAT<QMCTraits::ValueType>;
+using DistanceTableAB = DistanceTableABT<QMCTraits::ValueType>;
```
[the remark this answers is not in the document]
prckent: Yes. This needs to be fixed.

src/Particle/ParticleSet.h
quantumsteve: Is [MCWalkerConfiguration](https://github.com/QMCPACK/qmcpack/blob/864c22d19c26a8b41b0461bca5beb818cbbbd766/src/Particle/MCWalkerConfiguration.h#L55), which inherits from ParticleSet, real-only?

ye-luo: it should.

quantumsteve: Since `ParticleSet` contains `GradType` and `QMCTraits::GradType` depends on `ValueType`. I don't see a straightforward way to make ParticleSet real-only.

https://github.com/QMCPACK/qmcpack/blob/864c22d19c26a8b41b0461bca5beb818cbbbd766/src/type_traits/QMCTypes.h#L44

src/Particle/DistanceTable.h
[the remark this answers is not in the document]
quantumsteve: That can't be fixed without removing the member variable `const ParticleSetT<T>& origin_;`

https://github.com/QMCPACK/qmcpack/blob/864c22d19c26a8b41b0461bca5beb818cbbbd766/src/Particle/DistanceTable.h#L51

williamfgc:
Test this please

williamfgc:
@prckent @ye-luo @quantumsteve this is a new checkpoint that CI tests are passing and conflicts are resolved.

src/Particle/DistanceTable.h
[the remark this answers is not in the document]
williamfgc: Given that in the current state in `develop` `DistanceTable` depends tightly on `ParticleSet` and `ValueType`, I think it falls outside this PR.

src/Particle/ParticleSet.h
williamfgc: Given that in the current state in `develop` `ParticleSet` depends on `ValueType`, I think it falls outside the scope of this PR.

PR #4685, opened 2023-07-24
[WIP] Make SPOSet a template.

ye-luo:
## Proposed changes
Make SPOSet a class template.

Not sure how to do partial compilation with ninja. With make,
```
cd src/QMCWaveFunctions
make -j32 SPOSet.o
```

- New feature

- No

laptop

## Checklist


jtkrogel:
Is there any possibility for odd interactions with classes like CompositeSPOSet?  That class merges other SPOSets into a single one.  Since with this PR each SPOSet might have a distinct value type, I was wondering if that might produce odd corner cases and/or break that class for some value type combinations.

ye-luo:
> Is there any possibility for odd interactions with classes like CompositeSPOSet? That class merges other SPOSets into a single one. Since with this PR each SPOSet might have a distinct value type, I was wondering if that might produce odd corner cases and/or break that class for some value type combinations.

The type/precision restriction from the base class only applies to the API. The derived class has all the freedom to use any type internally, for example single precision splines orbitals in the full precision build. Making SPOSet template doesn't change the fact that this restriction applies to CompositeSPOSet when it talks to its owned SPOSets but the latter have all the freedom to make their internal precision individually.

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -44,31 +46,47 @@ opt_variables_type& getMyVars(SPOSet& spo);
...
+  using HessVector        = typename OrbitalSetTraits<VALUE>::HessVector;
+  using HessMatrix        = typename OrbitalSetTraits<VALUE>::HessMatrix;
+  using GGGVector         = typename OrbitalSetTraits<VALUE>::GradHessVector;
+  using GGGMatrix         = typename OrbitalSetTraits<VALUE>::GradHessMatrix;
+  using SPOMap            = std::map<std::string, const std::unique_ptr<const SPOSetT>>;
+  using OffloadMWVGLArray = Array<Value, 3, OffloadPinnedAllocator<Value>>; // [VGL, walker, Orbs]
+  using OffloadMWVArray   = Array<Value, 2, OffloadPinnedAllocator<Value>>; // [walker, Orbs]
   template<typename DT>
   using OffloadMatrix = Matrix<DT, OffloadPinnedAllocator<DT>>;
+  using QTFull        = QMCTraits::QTFull;
```
ye-luo, line 80: Probabaly still need a fix to avoid exposing QTFull.

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -121,56 +139,57 @@ class SPOSet : public QMCTraits
...
   /** Evaluate the derivative of the optimized orbitals with respect to the parameters
    *  this is used only for MSD, to be refined for better serving both single and multi SD
    */
   virtual void evaluateDerivativesWF(ParticleSet& P,
                                      const opt_variables_type& optvars,
-                                     Vector<ValueType>& dlogpsi,
-                                     const QTFull::ValueType& psiCurrent,
-                                     const std::vector<ValueType>& Coeff,
+                                     Vector<Value>& dlogpsi,
+                                     const QTFull::ValueType& psiCurrent, //FIXME it is still a compile time type
```
ye-luo, line 191: The only use of QTFull here.

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
```diff
@@ -136,7 +136,7 @@ void SplineR2R<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
       for (IndexType j = 0; j < OrbitalSetSize; j++)
       {
         const auto cur_elem = Nsplines * i + j;
-        FullPrecValueType newval{0.};
+        QTFull::ValueType newval{0.};
```
ye-luo, line 139: Also here.

ye-luo:
Test this please

Issue #4681, opened 2023-07-20
Fix all app_error + APP_ABORT needed


ye-luo:
**Describe the bug**
```
      app_error() << e.what() << std::endl;
      APP_ABORT("Unhandled Exception");
```
doesn't print messages properly when using MPI. Such pattern needs to be fixed throughout the code base.

See fix, documentation in https://github.com/QMCPACK/qmcpack/pull/4680

prckent:
This would be a good target for the robots if any instances remain. If not, close (@ye-luo ?).

PR #4686, opened 2023-07-25
Correct app_error usage

ye-luo:
## Proposed changes
Address #4681 in a few places.

a) bug fix in Concurrency/ParallelExecutorOPENMP.hpp
b) clean up the MPC class
c) Remove dead ECPComponentBuilder::createVrWithData

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCHamiltonians/ECPComponentBuilder_L2.cpp
+++ b/src/QMCHamiltonians/ECPComponentBuilder_L2.cpp
@@ -20,7 +20,7 @@ void ECPComponentBuilder::buildL2(xmlNodePtr cur)
   if (grid_global == 0)
   {
     app_error() << "    Global grid needs to be defined." << std::endl;
-    APP_ABORT("ECPComponentBuilder::buildL2");
+    myComm->barrier_and_abort("ECPComponentBuilder::buildL2");
   }
 
   // read <L2> attributes
@@ -52,12 +52,12 @@ void ECPComponentBuilder::buildL2(xmlNodePtr cur)
   else
   {
     app_error() << "  Unrecognized format \"" << format << "\" in PP file." << std::endl;
-    APP_ABORT("ECPComponentBuilder::buildL2");
+    myComm->barrier_and_abort("ECPComponentBuilder::buildL2");
   }
   if (rcut < 0.0)
   {
     app_error() << "  L2 attribute cutoff is missing or negative.  Cutoff must be a positive real number." << std::endl;
-    APP_ABORT("ECPComponentBuilder::buildL2");
+    myComm->barrier_and_abort("ECPComponentBuilder::buildL2");
   }
   int npts = grid_global->size();
 
--- a/src/QMCHamiltonians/MPC.cpp
+++ b/src/QMCHamiltonians/MPC.cpp
@@ -33,30 +33,27 @@ namespace qmcplusplus
 void MPC::resetTargetParticleSet(ParticleSet& ptcl) {}
 
 MPC::MPC(ParticleSet& ptcl, double cutoff)
-    : Ecut(cutoff),
-      d_aa_ID(ptcl.addTable(ptcl, DTModes::NEED_FULL_TABLE_ON_HOST_AFTER_DONEPBYP)),
-      PtclRef(&ptcl),
-      FirstTime(true)
+    : Ecut(cutoff), d_aa_ID(ptcl.addTable(ptcl, DTModes::NEED_FULL_TABLE_ON_HOST_AFTER_DONEPBYP)), FirstTime(true)
 {
-  initBreakup();
+  initBreakup(ptcl);
 }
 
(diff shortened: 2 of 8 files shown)


ye-luo:
Test this please

src/QMCHamiltonians/MPC.cpp
```diff
@@ -327,9 +315,7 @@ void MPC::initBreakup()
 
 std::unique_ptr<OperatorBase> MPC::makeClone(ParticleSet& qp, TrialWaveFunction& psi)
 {
-  // return new MPC(qp, Ecut);
-  std::unique_ptr<MPC> newMPC = std::make_unique<MPC>(*this);
-  newMPC->resetTargetParticleSet(qp);
```
prckent, line 332: Why no longer needed?

src/QMCHamiltonians/ECPComponentBuilder_L2.cpp
```diff
@@ -20,7 +20,7 @@ void ECPComponentBuilder::buildL2(xmlNodePtr cur)
   if (grid_global == 0)
   {
     app_error() << "    Global grid needs to be defined." << std::endl;
-    APP_ABORT("ECPComponentBuilder::buildL2");
+    myComm->barrier_and_abort("ECPComponentBuilder::buildL2");
```
PDoakORNL, line 23: These seem like they are guaranteed to happen on all nodes so why not throw UniformCommunicateError? I feel like in case of a fatal error the error handling should start with throwing an exception and it should be a UniformCommunicateError if its expected to occur on every rank.

PDoakORNL (changes requested):
Why not use UniformCommunicateError?

src/QMCHamiltonians/MPC.cpp
ye-luo: The current implementation is no-op at line 33
```
resetTargetParticleSet(qp) {}
```
and resetTargetParticleSet should go away eventually.

src/QMCHamiltonians/ECPComponentBuilder_L2.cpp
ye-luo: The definition of all nodes is encoded in myComm. Once an all-node error reaches a level with accessible communicator, it should be handled rather than keep going up to the top level catch which might have a different definition of all nodes (communicator).

PDoakORNL: Sounds like the scope that owns the myComm should also be the error handling scope.  Doing it this way requires myComm to be known by every bit of code that can be called inside an MPI scope this increases coupling, complicates testing ,and injects access to MPI into tons of classes that don't require it.  Hence some many classes unnecessarily inheriting from MPIObjectBase.

The point of the UniformCommunicateError is to remove this inheritance and therefore dependence on the "message" target from as much of the code as possible.

PDoakORNL (changes requested):
As far as using myComm instead of thinking about the scope of error handling I'm not a big fan.  I think the right place to catch these sort  of errors is at the scope matching a particular communicator. i.e.

qmcapp.cpp  209-: 
```
    if (inputs.size() > 1)
      validInput = qmc->parse(inputs[qmcComm->getGroupID()]);
    else
      validInput = qmc->parse(inputs[0]);

    if (!validInput)
      qmcComm->barrier_and_abort("main(). Input invalid.");

    ...
    qmc.reset()
```

with
```
try
{

    if (inputs.size() > 1)
      qmc->parse(inputs[qmcComm->getGroupID()]);
    else
      qmc->parse(inputs[0]);
    ...
    qmc.reset()
}
catch(const UniformCommunicateError& uce)
{
      qmcComm->barrier_and_abort("main(). Input invalid.");

}

src/QMCHamiltonians/ECPComponentBuilder_L2.cpp
ye-luo: > Sounds like the scope that owns the myComm should also be the error handling scope.
It should. QMCPACK is an MPI program. If the MPI communicator is exposed at a given class, the developer should be aware what operations are serial and what operations are parallel calling from the member function of this class and thus error handling must happen at the level where the communicator is exposed.

It is clearly wrong to catch UniformCommunicateError at the top level. Why we catch UniformCommunicateError at QMCMain.cpp instead of qmcapp.cpp? Because in an ensemble job, the communicator in qmcapp was split into several sub-communicators used by QMCMain. If we catch UniformCommunicateError at qmcapp, the run may hang.

Don't get me wrong. I never say you should not use UniformCommunicateError, If a class doesn't need to be exposed with MPI communicators, the commincator exposure should be removed in the first place and then use UniformCommunicateError to convey a uniform error. However, blindly using UniformCommunicateError is simply dangerous.

PDoakORNL (approved):

PR #4687, opened 2023-07-27
Free MKL reserved memory before the SYCL queue destruction.

ye-luo:
## Proposed changes
If we don't proactively poking MKL, memory free() happens when unloading the dynamic library and causes segfault.

PS: Update sunspot recipe.

- Bugfix

- No

sunspot

## Checklist


ye-luo:
Test this please

prckent (approved):
LGTM

Definitely not an issue right now, and likely never, but this change reminds me that the SYCL implementation is not-portable to not Intel platforms because of our hard-coded MKL dependencies.

prckent:
Test this please

PR #4689, opened 2023-07-31
Add tests for orbital rotation for Be

markdewing:
Test case is a Be atom with an STO basis set.  Basis set and MO matrix are taken from https://github.com/QMCPACK/qmcpack/blob/develop/tests/molecules/Be_STO/partial_opt_Be_STO.wfnoj.xml

Python scripts compute reference values for the test cases:
 1. single determinant
 2. multi-determinant code path with one determinant (same values as 1)
 3. multi-determinant code path with two determinants

For ease of using autodiff, the skew-symmetric matrix that goes into the rotation matrix is constructed by a separate script (construct_rot.py) and hard-coded into the wavefunction scripts.

- No

desktop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/tests/CMakeLists.txt
+++ b/src/QMCWaveFunctions/tests/CMakeLists.txt
@@ -71,7 +71,10 @@ set(FILES_TO_COPY
     C_diamond-twist-third.structure.xml
     C_diamond-twist-third.wfj.xml
     cartesian_order.wfnoj.xml
-    dirac_order.wfnoj.xml)
+    dirac_order.wfnoj.xml
+    rot_Be_STO.wfnoj.xml
+    rot_multi_1det_Be_STO.wfnoj.xml
+    rot_multi_2det_Be_STO.wfnoj.xml)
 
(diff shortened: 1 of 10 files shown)


markdewing:
Test this please

prckent (changes requested):
Thanks Mark. 

If I interpret this and the results of the test coverage correctly,  we should now have good confidence that the multidet rotations are OK. In terms of orbital optimization, we only have open the question of whether the correct/good algorithm is being used at the optimization level. Agree?


Small request: Please can you add some comments at the head of each *.py explaining their purpose & use. The idea is to make things clear to our future selves/a new developer. An overall README_rotation_tests.md would also work.

markdewing:
For all-electron, I think that is correct.  For pseudopotentials, multideterminants need to used the older "non-batched" (not referring to the driver) algorithm.  There is an unimplemented code in the path for the "batched" (VirtualParticle) algorithm.

prckent (approved):

prckent:
Test this please

PR #4699, opened 2023-08-14
Fix Nexus GPU jobs for NERSC/Perlmutter

aannabe:
## Proposed changes

`OMP_PROC_BIND` and `OMP_PLACES` environment variables should not be set for GPU jobs. With this change, Nexus GPU jobs are running for me on Perlmutter.

- Bugfix

- No

NERSC/Perlmutter (Nexus tests are passing)

## Checklist


prckent (approved):
Approving since it allows jobs to run!

However I suspect we might need to revisit this for effective multithreaded GPU offload. Did the environment variable change come as a recommendation from NERSC or through experimentation?

prckent:
Test this please

PR #4700, opened 2023-08-16
Add doxygen target using cmake-generated doxyfile

PhilipFackler:
## Proposed changes

This is a pretty small change to add a generated doxygen target (so you can `make qmcpack_doxygen` and it will put the results in the build tree). I left the old doxygen work untouched and simply added a CMakeLists.txt file to the doxygen directory. This could be further customized of course, but it provides an easy way for developers to look at an up-to-date doxygen site locally.

- Documentation changes

- No

Ubuntu 22.04

## Checklist


doxygen/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+find_package(Doxygen)
+if(${DOXYGEN_FOUND})
```
ye-luo, line 2: You can just put
```
if(DOXYGEN_FOUND)
```
using `if(<variable>)` pattern.

PhilipFackler: Done. Thanks for the reminder.

prckent:
Agree this is useful. We will merge after the release to avoid modifying CMakeLists.txt just ahead of a release.

I assume you have found https://docs.qmcpack.org/doxygen/doxy/ , which is run nightly.

A potentially slicker and longer term solution would be to have the doxygen pages included in our readthedocs builds of the manual, but this is much more work and may need some bridging software.

PhilipFackler:
@prckent , part of the reason I wanted this is that the website you referenced (https://docs.qmcpack.org/doxygen/doxy/) seems drastically out of date. And yes, an automated build included with the rtd would be amazing.

prckent:
Can you give an example of what is missing from the current doxygen created files so that I can check the regeneration? The website has a current date, so it is not that the whole regeneration is failing.

quantumsteve:
> Can you give an example of what is missing from the current doxygen created files so that I can check the regeneration? The website has a current date, so it is not that the whole regeneration is failing.

The inheritance diagram for SPOSet contains derived classes such as `SPOSetProxy` that we couldn't find in the QMCPACK repository.

https://docs.qmcpack.org/doxygen/doxy/d4/d45/a04456.html

prckent:
Check now. git repo updating was broken.

quantumsteve:
> Check now. git repo updating was broken.

SPOSet looks Okay.

doxygen/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+    set(DOXYGEN_USE_MDFILE_AS_MAINPAGE ${PROJECT_SOURCE_DIR}/README.md)
+    set(DOXYGEN_GENERATE_TREEVIEW YES)
+    set(DOXYGEN_TEMPLATE_RELATIONS YES)
+    set(DOXYGEN_BUILTIN_STL_SUPPORT YES)
+    set(DOXYGEN_EXTRACT_PRIVATE YES)
+    set(DOXYGEN_SOURCE_BROWSER YES)
+    set(DOXYGEN_INTERACTIVE_SVG YES)
+    set(DOXYGEN_DISTRIBUTE_GROUP_DOC YES)
+
+    doxygen_add_docs(qmcpack_doxygen ${PROJECT_SOURCE_DIR}/src
```
prckent, line 13: Prefer
```
    doxygen_add_docs(qmcpack_doxygen ${PROJECT_SOURCE_DIR}/src)
else()
    message(STATUS "The qmcpack_doxygen target for the QMCPACK developer/API documentation requires (1) doxygen and (2) dot from graphviz.")
endif()
````

File: doxygen/CMakeLists.txt [hunk 3]
```diff
@@ -0,0 +1,15 @@
+find_package(Doxygen)
+if(DOXYGEN_FOUND)
+    set(DOXYGEN_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
+    set(DOXYGEN_USE_MDFILE_AS_MAINPAGE ${PROJECT_SOURCE_DIR}/README.md)
```
prckent, line 4: I suggest to remove this, since it gives a very strange first page.

Can you modify the config here so that it gives an accessible start page like https://docs.qmcpack.org/doxygen/doxy/index.html ?
The  one produced by this PR has a long function(?) list before the classes and files browser.

prckent (changes requested):

PDoakORNL:
I still have interest in this. Will certainly make cleaning up doxygen output locally more convenient.  It really should be part of dev workflow to see if our doxygen output for new work looks good or not.

prckent:
@PhilipFackler Will you be able to come back to this? This looked to be mergable after fixing the first page (DOXYGEN_USE_MDFILE_AS_MAINPAGE) to be more sensible somehow.

doxygen/CMakeLists.txt
PhilipFackler: @prckent I believe the issue was with the navigation tree on the left. I like using the README since it gives links to all the relevant things on the web. I addressed the navigation tree issues (and in my opinion it's more useful than the tree in the current published version). Let me know what you think.

doxygen/CMakeLists.txt
PhilipFackler: I added a message to similar effect

prckent:
(I expect we'll merge this after it has been run independently, after the QMCPACK workshop this week)

prckent:
I have just tested this on Ubuntu24, and it works very well. For simplicity, I propose we merge this  and then I will address the following subsequently:

1. Add a note on running the qmcpack_doxygen target to the CMake output
2.  Add some developer docs to help people find it.
3. Move all the existing files in doxygen/ to an "attic" directory, other than the CMakeLists.txt. None of the other files are needed and they can likely all be deleted.

prckent (approved):

prckent:
Test this please

ye-luo (approved):

ye-luo:
Test this please

PR #4706, opened 2023-08-21
apply_rotation timer

camelto2:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

adding a timer to the Phi->apply_rotation calls. Useful as we are testing some of the performance of orbital optimization

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- New feature

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
macOS

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- No. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #4701, opened 2023-08-17
add mw_ APIs to RotatedSPO

camelto2:
## Proposed changes

In the case where we go straight from an optimization into a VMC/DMC in the same run, or with WF evaluation during optimization we want to be able to utilize mw_ APIs and  offload when using RotatedSPOs. Currently, since RotatedSPOs inherits from SPOSet and doesn't have any mw_ APIs defined, it will just fall back to the single walker APIs. This PR adds the resource management and mw_ APIs to RotatedSPOs, which simply call its underlying SPOSet's resource and mw_ APIs. 

I made a unit test to check that these mw_ APIs have the correct behavior by introducing two "dummy" SPOSets, one that has mw_ APIs defined and one that doesn't. A RotatedSPO for the dummy SPO which doesn't have any mw_ specialization does RotatedSPOs::mw_XXX which calls underlying SPO::mw_XXX. Since it isn't defined, it defaults to loop over single walker APIs. But in the case where the underlying SPOSet does have mw_ defined, we want it to have the appropriate behavior. The dummy test SPOSets I created set different values in different APIs to ensure that the appropriate code paths are being followed for mw_evaluateValue. I assumed it would be sufficient to test only that one, but can add more if needed. 

Also, if there is a better way to test this, let me know and I can make some changes. 

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- New feature

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
macOS

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- Yes. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #4710, opened 2023-08-24
SplineC2C/R2R rotation with BLAS

camelto2:

## Proposed changes
 change naive matrix multiplication in SplineR2R::applyRotation to a blas call. Speeds the applyRotation call up by quite a bit. 

note that these changes should be covered by the unit test, just changing how the matrix multiplication is done

Update: also added the corresponding change needed for SplineC2C. 

- New feature

- Other (please describe):

- No

macOS

## Checklist


markdewing:
There may be some subtleties related to precision.  If a single precision spline is used, this change reduces the rotation matrix to single precision before the multiplication.

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
```diff
@@ -120,21 +121,12 @@ void SplineR2R<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
...
-      for (auto k = 0; k < OrbitalSetSize; k++)
-      {
-        const auto index = i * Nsplines + k;
-        newval += (*coef_copy_)[index] * rot_mat[k][j];
-      }
-      spl_coefs[cur_elem] = newval;
-    }
-  }
+       rot_mat_padded[i * Nsplines + j] = rot_mat.data()[i * OrbitalSetSize + j];
+  BLAS::gemm('N', 'N', Nsplines, BasisSetSize, Nsplines, ST(1.0), rot_mat_padded.data(), Nsplines, (*coef_copy_).data(), Nsplines, ST(0.0), spl_coefs, Nsplines);
```
ye-luo, line 128: I feel we don't need `rot_mat_padded`.
```
BLAS::gemm('N', 'N', OrbitalSetSize, BasisSetSize, OrbitalSetSize, ST(1.0), rot_mat.data(), OrbitalSetSize, (*coef_copy_).data(), Nsplines, ST(0.0), spl_coefs, Nsplines);
```

camelto2:
> Let's be careful changes in derived classes aren't lost as we move to templated classes in #4684

Thanks for pointing this out. I had two PRs recently (#4706, #4701) that touched RotatedSPOs that doesn't look like it has been captured in #4684. I can make a PR into @williamfgc's branch. And if this gets merged once I sort out the review comments can do the same for this if that works

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
markdewing: There's still a precision issue, unless BLAS::gemm can handle matrices with different precision types.

camelto2: I was having trouble getting the BLAS call to work without padding. I am trying to figure it out now

I agree there is still a precision issue. I wasn't sure the best way to handle that either. the spline coeffs are ST type and the rotation is RealType. The BLAS::gemm doesn't like mixed precision types, so I had to do some conversion somewhere. If I changed coef_copy_ back to RealType, and the rotation is RealType, the problem then lies in spl_coefs. It is ST*, and I can't just cast it to RealType* in the case of ST=float. 
Making another temporary copy would be costly memory wise

markdewing: Are you intending to work with single or double precision splines?  I ran into some problems with rotation and single precision splines early on, and since then have only used double precision splines.
The code might need to be split differently based spline precision - call gemm for double precision, and an explicit loop for single precision.   That would at least speed up the double precision case.

The precision issue also relates to the history list vs. global rotation method.  With the history list, the rotations get accumulated in the existing coefficient storage.  Single precision coefficients fail at this accumulation of small changes.  With the global rotation method this might not be an issue, since this accumulation of rotations happens in the rotation matrix.  It might work to compute the rotation matrix in double precision, then convert to single precision, and then do the matrix multiplication.  But that would need to be tested, since the matrix multiplication step might also be sensitive to precision.

camelto2: I would prefer double only, but some of the unit tests use float so I was trying to get it to work in general

markdewing: I would be okay with different code paths for float vs double (use the loop for float and gemm for double).  Using "if constexpr" to test the type should make it straightforward.

williamfgc:
@camelto2 @quantumsteve thanks for bringing this up. Your changes can be accommodated into the refactoring effort after merged into develop, we just won't touch it for now. We rebase develop constantly.

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
camelto2: @markdewing, added a split depending on float vs. double. this should avoid the loss in precision.

@ye-luo still trying to figure out how to avoid adding the padded rotation matrix. I thought it should be as simple as fiddling with the LDx parameters, but having trouble getting it to work

markdewing: A code path split by precision should also avoid the need for the extra padded rotation matrix.

camelto2: @ye-luo I think I have the blas call correct now to avoid needing to construct the padded rotation matrix.

jptowns: > A code path split by precision should also avoid the need for the extra padded rotation matrix.

- Do we need single precision spline coefs if there is a problem for rotations (which we ultimately expect to be as common as Jastrow optimization eventually)?
- Wouldn't it be better to modify the unit tests to conform to the production code? 

BLAS-ification is good future-proofing, and probably needed at some level for GPU compatibility anyways.

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
```diff
@@ -120,20 +121,28 @@ void SplineR2R<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
...
-    {
-      const auto cur_elem = Nsplines * i + j;
-      auto newval{0.};
-      for (auto k = 0; k < OrbitalSetSize; k++)
+    //Here, ST should be equal to ValueType, which will be double for R2R. Using BLAS to make things faster
+    BLAS::gemm('N', 'N', OrbitalSetSize, BasisSetSize, OrbitalSetSize, ST(1.0), rot_mat.data(), OrbitalSetSize,
+               (*coef_copy_).data(), Nsplines, ST(0.0), spl_coefs, Nsplines);
+  }
+  else
+  {
```
jptowns, line 132: This is tough: apparently everything I did had hidden type mis-matches because ST is not guaranteed to be equivalent to ValueType. Seems to me the underlying problem is the spline stuff is kinda divorced from the rest of the code. If it's true that single precision may be problematic for rotations, then why not ditch single precision splines and keep the production code simple? Or is there some way to enforce double precision splines if rotation is added?

markdewing: My guess is single precision splines are still of interest for size reasons.

markdewing (approved):

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
camelto2: > I feel we don't need `rot_mat_padded`.
> 
> ```
> BLAS::gemm('N', 'N', OrbitalSetSize, BasisSetSize, OrbitalSetSize, ST(1.0), rot_mat.data(), OrbitalSetSize, (*coef_copy_).data(), Nsplines, ST(0.0), spl_coefs, Nsplines);
> ```

I should have read this in detail sooner. I didn't realize you had already told me how to it. Took me a bit, but I stumbled onto doing correctly

camelto2:
Figured out how to do C2C correctly using the BLAS call. added that here as well

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
```diff
@@ -120,20 +121,28 @@ void SplineR2R<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
...
+    //Here, ST is float but ValueType is double for R2R. Due to issues with type conversions, just doing naive matrix multiplication in this case to not lose precision on rot_mat
+    for (auto i = 0; i < BasisSetSize; i++)
```
quantumsteve, line 134:
```suggestion
    for (IndexType i = 0; i < BasisSetSize; i++)
```

I realize this is an existing issue, but using `auto` here and in the other loops makes it harder to later change `IndexType`.

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
```diff
@@ -120,20 +121,28 @@ void SplineR2R<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
...
+      for (auto j = 0; j < OrbitalSetSize; j++)
       {
-        const auto index = i * Nsplines + k;
-        newval += (*coef_copy_)[index] * rot_mat[k][j];
+        const auto cur_elem = Nsplines * i + j;
+        auto newval{0.};
```
quantumsteve, line 138:
```suggestion
        FullPrecValueType newval{0.};
```

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
camelto2: done

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
camelto2: fixed

markdewing (approved):

markdewing:
Test this please

jptowns (approved):

src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
```diff
@@ -120,27 +121,41 @@ void SplineC2C<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
...
-      ST newval_r{0.};
-      ST newval_i{0.};
-      for (auto k = 0; k < OrbitalSetSize; k++)
+  if constexpr (std::is_same_v<ST, RealType>)
+  {
+    //if ST is double, go ahead and use blas to make things faster
+    //Note that Nsplines needs to be divided by 2 since spl_coefs and coef_copy_ are stored as reals.
+    //Also casting them as ValueType so they are complex to do the correct gemm
+    BLAS::gemm('N', 'N', OrbitalSetSize, basis_set_size, OrbitalSetSize, ValueType(1.0, 0.0), rot_mat.data(),
+               OrbitalSetSize, (ValueType*)(*coef_copy_).data(), Nsplines / 2, ValueType(0.0, 0.0),
```
ye-luo, line 130: Prefer `coef_copy_->data()` to `(*coef_copy_).data()`. `()` reduces readability.

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
```diff
@@ -120,20 +121,28 @@ void SplineR2R<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
     std::copy_n(spl_coefs, coefs_tot_size, coef_copy_->begin());
   }
 
-  // Apply rotation the dumb way b/c I can't get BLAS::gemm to work...
-  for (auto i = 0; i < BasisSetSize; i++)
+
+  if constexpr (std::is_same_v<ST, ValueType>)
```
ye-luo, line 125: Prefer `std::is_same_v<ST, RealType>`.

src/QMCWaveFunctions/BsplineFactory/SplineC2C.h
```diff
@@ -64,7 +64,7 @@ class SplineC2C : public BsplineSet
   std::shared_ptr<MultiBspline<ST>> SplineInst;
 
   ///Copy of original splines for orbital rotation
-  std::shared_ptr<std::vector<RealType>> coef_copy_;
+  std::shared_ptr<std::vector<ST>> coef_copy_;
```
ye-luo, line 67: This is a valid change. coef_copy should keep the same precision as the original coefs.

src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
```diff
@@ -120,27 +121,41 @@ void SplineC2C<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
...
+               (ValueType*)spl_coefs, Nsplines / 2);
+  }
+  else
+  {
+    // if ST is float, RealType is double and ValueType is std::complex<double> for C2C
+    // Just use naive matrix multiplication in order to avoid losing precision on rotation matrix
```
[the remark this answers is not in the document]
camelto2: @markdewing was mentioning that he had run into problems with precision when using the float splines. Maybe he can chime in with more details

src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
camelto2: will fix

src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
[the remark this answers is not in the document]
markdewing: There were cases where splines with float precision had trouble with orbital rotation optimization, and those problems went away when I switched to double precision splines. I haven't investigated further.
The important point for these code paths is that the type of the spline (ST) and the main code type (RealType) have to match for BLAS calls to work without additional copies.  For full precision builds, that means double precision splines.  For mixed precision builds, I would guess the single precision splines would follow the BLAS code path.

src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp
camelto2: done

src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
camelto2: done

src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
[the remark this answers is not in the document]
camelto2: How it stands currently is that coefs_copy_ and spl_coefs are ST, whereas rot_mat is passed in as ValueType. There isn't a mixed precision blas call, so in order to use blas we would have to make a copy of rot_mat of ST type. That would allow BLAS to be used always, but you are losing precision when ST is float since it is passed in as ValueType (double or std::complex<double>). I'm not entirely sure how much that precision matters, since the output spline coeffs are ST anyway. We are doing an extra copy, but it is only Norb^2 so shouldn't be a problem.

How it is currently written avoids the extra copy of rot_mat, but only benefits from BLAS when ST == RealType. 

How common are runs where ST != RealType?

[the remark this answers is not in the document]
markdewing: The NiO performance tests have the splines set with precision="single", so I think it's still an important case in general.  I'm not sure how important it will be for cases involving rotation.  Though my guess is someone will want to try it due to memory pressure on the size of the spline coefficients.

prckent:
Agreeing with Ye here - let us focus on getting the feature working.

PR #4714, opened 2023-08-25
[WIP] Optimized Soa spherical tensor

anbenali:

## Proposed changes

Explicitly unrolling the  functions in the Spherical harmonics evaluations based on  Sloan, P., Efficient Spherical Harmonic Evaluation, Journal of Computer Graphics Techniques, vol. 2, no. 2, 84-90, 2013. Gets 1.6 to 6X speedup depending on the angular momentum. 

Total Gain for Fe(CO)6 using bfd_vtz :1.8X

Total Gain for Au single Atom using ccecp_aug-cc-pcv5z : 2.4 X

Total Gain for H2O using ccecp_cc-pcvtz:  1.4X

Tests conducted on laptop.

- no

- Personal Laptops
- Cooley [ALCF]
## Checklist


prckent (commented):
* Please post some of the data you have on the performance improvements with details of the system (basis etc.) and hardware. I do believe the result -- it is more than unrolling.
* Is this code a cut & paste from the cited article or do you have a generator code? If the latter we should capture it, just as we have been capturing Mark's python codes for optimization and wavefunction testing.
* I hope we can get test coverage near 100% for these foundational numeric functions.

prckent:
Test this please

ye-luo:
leave some idea. We can technically generate constants using constexpr functions to avoid storing them in the source.
```
#include <cmath>
#include <iostream>

constexpr double foo(double a)
{
  double res = 0;
  for(int i=0;i<2;i++)
    res+= ::sin(a+i);
  return res;
}

int main()
{
  constexpr double b = foo(2); # can be evaluated at compile time even if it involves math function calls in it.
  std::cout << "b = " <<b <<std::endl;
  return b;
}
```

anbenali:
> * Please post some of the data you have on the performance improvements with details of the system (basis etc.) and hardware. I do believe the result -- it is more than unrolling.

I added the data.  Should I create tests for these or just the numbers as is are enough?

>     * Is this code a cut & paste from the cited article or do you have a generator code? If the latter we should capture it, just as we have been capturing Mark's python codes for optimization and wavefunction testing.
Yes There is a code generator. Kevin (@kgasperich ) will supply it.


>     * I hope we can get test coverage near 100% for these foundational numeric functions.
There is an existing test already covering this function but it is not clear if it is being called.  Once the deadlines are past we will "clean" it.

prckent:
Generator code should be added in case "we" end up redoing this.

prckent:
Ping. What is the status/plan for this code? 

To be merged this will need at least the generator code added, as previously requested, and -- similar to how Mark Dewing did this for other generated code -- suitably clear comment guards around the generated code. ( "DO NOT EDIT AUTOMATICALLY GENERATED BY XYZ.PY" ). Automagically generated unit tests would also be ideal -- we ought to be able to get ~100% coverage for these functions.

anbenali:
Putting a comment here to push it up as a possible additional speedup

prckent:
Main request is for the generator code, which exists per the comment made in 2023(!). Should be easy to add to this PR then remove WIP.

prckent:
#6181 reminded me of this WIP PR and the request for the generator code.  I was able to quickly find a few implementations of the algorithms from Peter Pike Sloan's paper ( https://jcgt.org/published/0002/02/06/ ), but identifying which (if any) is a good starting point would take some research. Alternatively, perhaps we have fresh thoughts after 3 years?

shivupa:
OK I was able to reverse engineer the generator (using Claude Opus 5). I opened a PR https://github.com/anbenali/qmcpack/pull/22 with the generator and tests. The only code diff is the added "generated code" warning.

 I think following that we can merge develop + resolve conflicts? This is obviously quite a bit better than the existing code. It would be nice to have and #6181 can be rebased on this

Issue #4654, opened 2023-06-29
Segfault in propagator initialization in complex AFQMC


smaintz-nv:
**Describe the bug**
When running an AFQMC complex dataset on 1 GPU, I get a segfault during Propagator initialization:
[backtrace.txt](https://github.com/QMCPACK/qmcpack/files/11907697/backtrace.txt)

**To Reproduce**
1.  develop@9a6e1d3
2. `FFTW_HOME=/usr/local/fftw CC=/usr/local/openmpi/bin/mpicc CFLAGS='-mtune=generic -march=x86-64 -O2 -pipe' CXX=/usr/local/openmpi/bin/mpicxx CXXFLAGS='-mtune=generic -march=x86-64 -O2 -pipe' F77=/usr/local/openmpi/bin/mpifort F90=/usr/local/openmpi/bin/mpifort FC=/usr/local/openmpi/bin/mpifort FCFLAGS='-mtune=generic -march=x86-64 -O2 -pipe' FFLAGS='-mtune=generic -march=x86-64 -O2 -pipe' cmake -DCMAKE_INSTALL_PREFIX=/usr/local/qmcpack/sm70 -DENABLE_CUDA=ON -DCMAKE_BUILD_TYPE=DEBUG -DQMC_MIXED_PRECISION=1 -DHDF5_ROOT=/opt/hdf5 -DENABLE_PHDF5=1 -DQMC_MIXED_PRECISION=1 -DENABLE_TIMERS=1 -DQMC_BUILD_STATIC=0 -DBUILD_AFQMC=ON -DBUILD_AFQMC_WITH_NCCL=ON -DCMAKE_EXE_LINKER_FLAGS=-lnccl -DQMC_COMPLEX=ON -DCUDA_HOST_COMPILER=`which gcc` -DCUDA_PROPAGATE_HOST_FLAGS=OFF -DENABLE_OFFLOAD=ON -DUSE_OBJECT_TARGET=ON -DCMAKE_CUDA_ARCHITECTURES=70 /var/tmp/qmcpack`
3. `/usr/local/qmcpack/sm70/bin/qmcpack_complex afqmc_1e-5.xml`
4. none

**System:**
 - NVIDIA DGX-1 (V100)
 - containerized environment based on Ubuntu 22.04, llvm 16, openmpi 4.1.5, ucx 1.13.1, CUDA 11.8
 - have not yet tried other systems

**Additional context**
[...]

PR #4717, opened 2023-08-29
fix type mismatch of multi array in AFQMC

correaa:
## Proposed changes

Multi is more picky now for implicit conversions.
There is no impact at runtime, since it was unreachable code.

- Bugfix

- No

## Checklist


markdewing (approved):

correaa:
> Good to hear from you Alfredo. Any chance this helps with #4654 ? Can you reproduce the reported problem at all? I am not sure if that is a new problem introduced by some of our recent "small" changes or a old one.

yes, let's see how easy it is to reproduce.
I don't think it is related because the code I changed here was unreachable; it was a compilation issue.

PR #4722, opened 2023-09-08
Fix parameters for real-complex dot product

PhilipFackler:
## Proposed changes
Template parameters did not allow for a real type other than the underlying type of the complex type.

- Bugfix

- No

Ubuntu 22.04

## Checklist

### DIFF
--- a/src/Containers/OhmmsPETE/TinyVectorOps.h
+++ b/src/Containers/OhmmsPETE/TinyVectorOps.h
@@ -355,11 +355,11 @@ struct OTDot<TinyVector<T1, 4>, TinyVector<T2, 4>>
 };
 
 /** specialization for real-complex TinyVector */
-template<class T1>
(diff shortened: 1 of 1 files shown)


prckent (commented):
Wondering where/why this is needed? This potentially introduces new functionality and not a fix. Potentially needing this to compile some code is an indication of a bug, and one that could be tricky to track down.

 The merged real-complex implementation should not require any new functionality...

williamfgc:
@PhilipFackler I agree with @prckent , our changes should be bottom-up, and try to prevent top-down as in the PR (only if required like the day we remove `QMC_COMPLEX` globally).  The issue with mixed precision in on our [effort](https://github.com/williamfgc/qmcpack/actions/runs/6104333538/job/16566208379?pr=29) is related to `DistanceTable` still depending on `ValueType`. We can follow up offline.

PR #4732, opened 2023-09-15
Fix checkMatrix usage in some tests

camelto2:

## Proposed changes

While trying to get a new feature tested that I'm working on, I noticed that a few unit tests have incorrect usage of the "checkMatrix". Those tests were just calling checkMatrix(matA, matB) and not getting the result and checking to see if the matrices agree or not. 

This means that some of those features could actually be failing and we have been missing it since the checks weren't correct. I think I caught all instances here.

- Bugfix

- No

macOS

## Checklist

- Yes. This PR adds tests to cover any new code, or to catch a bug that is being fixed _(Kind of, it is fixing existing tests)_

### DIFF
--- a/src/Utilities/for_testing/checkMatrix.hpp
+++ b/src/Utilities/for_testing/checkMatrix.hpp
@@ -17,20 +17,21 @@
 #include <string>
 #include <complex>
 #include <type_traits>
+#include <limits>
 #include "type_traits/complex_help.hpp"
 namespace qmcplusplus
 {
 
 template<typename T, IsComplex<T> = true>
-bool approxEquality(T val_a, T val_b)
+bool approxEquality(T val_a, T val_b, double eps)
 {
-  return val_a == ComplexApprox(val_b);
+  return val_a == ComplexApprox(val_b).epsilon(eps);
 }
 
 template<typename T, IsReal<T> = true>
-bool approxEquality(T val_a, T val_b)
+bool approxEquality(T val_a, T val_b, double eps)
 {
-  return val_a == Approx(val_b);
+  return val_a == Approx(val_b).epsilon(eps);
 }
 
 struct CheckMatrixResult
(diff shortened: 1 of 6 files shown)


camelto2:
Note that there are failing tests only from the files where I fixed the checkMatrix call to actually see if the matrices agree or not. Seems like there could be lingering bugs

ye-luo:
Test this please

ye-luo:
@camelto2 remaining failure are from the added code?

camelto2:
Looks like it would still be tolerance actually. I think to add a tolerance argument into checkMatrix and that would solve the mixed precision issue

camelto2:
Ok, so the failures were due to the fact that the checkMatrix function uses Catch2 Approx, and was always using the default epsilon. For mixed precision applyRotation in the splines, the laplacian was failing tests due to that default tolerance. 

I added some functionality which allows us to pass in a desired tolerance for the element by element comparison in checkMatrix. Also added some testing of that inside the unit test for checkMatrix.

src/Utilities/for_testing/checkMatrix.hpp
```diff
@@ -50,9 +51,13 @@ struct CheckMatrixResult
...
  *  \param[in] b_mat     - the matrix to check
  *  \param[in] check_all - if true continue to check matrix elements after failure
+ *  \param[in] eps       - add a tolerance for Catch Approx checks. Default to same as in Approx. 
  */
 template<class M1, class M2>
-CheckMatrixResult checkMatrix(M1& a_mat, M2& b_mat, const bool check_all = false)
+CheckMatrixResult checkMatrix(M1& a_mat,
+                              M2& b_mat,
+                              const bool check_all              = false,
+                              const double eps = std::numeric_limits<float>::epsilon() * 100)
```
ye-luo, line 60: I don't feel right with this hidden choice. Could you use std::optional and default to `std::nullopt_t`

camelto2: Think I have this working with optional.

ye-luo (approved):

ye-luo:
Test this please

PR #3341, opened 2021-08-04
Correctly handle NULL pivot in getr[f|i]Batched

jakurzak:
Right now, hipblas?getrfBatched() and hipblas?getriBatched() routines do not allow for passing NULL for the pivot vector.
I.e., right now, hipBLAS does not follow cuBLAS semantics.
The issue is being resolved, i.e., cuBLAS semantics are being implemented.
In the meantime, the problem is resolved in the CUDA2HIP wrappers by calling appropriate rocBLAS routines.

PR #4742, opened 2023-09-20
Update batched matrix inversion and update frontier recipe

### LINKED ISSUE #4741: Spin off identity kernels needed
I spotted the issue on Frontier when using rocm 5.7 due to the HIP_PATH set in modules provided by OLCF.
the hip compiler(llvm/bin/clang) in rocm 5.7.0 picks up `/opt/rocm-5.7.0/hip` instead of `/opt/rocm-5.7.0` and we lost the implicit `-I /opt/rocm-5.7.0/include` and it ended up with
```
/ccs/home/yeluo/opt/qmcpack/src/Platforms/ROCm/hipBLAS.hpp:18:10: fatal error: 'hipblas/hipblas.h' file not found
#include <hipblas/hipblas.h>
         ^~~~~~~~~~~~~~~~~~~
```

It is a CMake and ROCm interacting issue.
We mark
https://github.com/QMCPACK/qmcpack/blob/develop/src/Platforms/ROCm/hipBLAS.hip.cpp
as HIP source code in https://github.com/QMCPACK/qmcpack/blob/053f38a3d27996f9e9a67bccd2c5bb51c80832c7/src/Platforms/ROCm/CMakeLists.txt#L16
Although we set dependency on `roc::hipblas` `roc::rocsolver` targets
https://github.com/QMCPACK/qmcpack/blob/053f38a3d27996f9e9a67bccd2c5bb51c80832c7/src/Platforms/ROCm/CMakeLists.txt#L18
These targets are strictly CXX libraries and doesn't define include/library path for HIP language.

The solution I can think of is to split all the identity kernels in a separate file and only set HIP language on that file.

ye-luo:
## Proposed changes
Resolves #4741 by deleting the non-pivoting code path that needs kernels.
The non-pivoting case was needed likely by the legacy CUDA which was deleted.

- No

epyc-server

## Checklist


ye-luo:
@jakurzak I was reading https://github.com/QMCPACK/qmcpack/pull/3341#issuecomment-892965849. What is the current behavior of hipblas?getrfBatched() and hipblas?getriBatched()? Are they unconditionally hitting the non-pivoting case or they behave like cuBLAS selecting code path depending on whether the pointer is nullptr? In the former case, the current code is good. In the latter case, we may further simplify the code in QMCPACK.

ye-luo:
Test this please

prckent (commented):
The problem with asserts is that they are only active in debug builds. We don't have any ROCm debug builds in testing. Throw an exception instead?

ye-luo:
> The problem with asserts is that they are only active in debug builds. We don't have any ROCm debug builds in testing. Throw an exception instead?

Agreed. It is pure C++ now and thus exception is better. I was concerned of C++ and HIP interaction.

ye-luo:
Test this please.

prckent (approved):
Simpler.

rtd build for nexus docs is sick but this PR does not touch any docs. Hopefully a transient problem.

Issue #4666, opened 2023-07-11
Batched performance-C-molecule-C12* assertion failures


prckent:
**Describe the bug**

Debug CPU builds fail assertion check in VMC batched. Interestingly only hit by these performance tests, not any deterministic tests.

```
        147 - performance-C-molecule-C12-e72-ae-batched_driver-w256-1-4 (Subprocess aborted)
        149 - performance-C-molecule-C12-e72-ae-J3-batched_driver-w256-1-4 (Subprocess aborted)
        155 - performance-C-molecule-C12-e48-pp-batched_driver-w256-1-4 (Subprocess aborted)
        157 - performance-C-molecule-C12-e48-pp-J3-batched_driver-w256-1-4 (Subprocess aborted)
```

qmcpack/src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp:670: virtual void qmcplusplus::DiracDeterminantBatched<>::mw_evaluateGL(const RefVectorWithLe\
ader<qmcplusplus::WaveFunctionComponent> &, const RefVectorWithLeader<qmcplusplus::ParticleSet> &, const RefVector<ParticleSet::ParticleGradient> &, const RefVector<ParticleSet::ParticleLaplacian> &, bool) cons\
t [DET_ENGINE = qmcplusplus::MatrixUpdateOMPTarget<double, double>]: Assertion `dpsiM_from_device == det.dpsiM' failed.

**To Reproduce**

Develop version
```
cmake -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DQMC_MPI=0 -DCMAKE_BUILD_TYPE=Debug -DENABLE_SANITIZER=asan -DQMC_DATA=/scratch/pk7/QMC_DATA_FULL -DQMC_PERFORMANCE_NIO_MAX_ATOMS=48 -DQMC_PERFORMANCE_C_MOLECULE_MAX_ATOMS=16 -DQMC_PERFORMANCE_C_GRAPHITE_MAX_ATOMS=16 ../qmcpack/
```

prckent:
Thanks @shivupa . Do the molecule-C12 tests pass for you? I still get the failure.

PR #4752, opened 2023-09-29
Fix assertion failure

ye-luo:
## Proposed changes
Addresses #4666.
mw_evaluate_notranspose and evaluate_notranspose don't produce bitwise identical results. 
The problem started by https://github.com/QMCPACK/qmcpack/pull/4407 which introduce a batched evaluation and caused bitwise matrix value checking failure. I printed diff and found the value is at the epsilon of float or double. So no concern.

The intention of the assertion was to catch any missing D2H transfer earlier and there is no need of recomputing. So the solution is removing the re-computation.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent:
Thanks. An important source of worry removed. Did you find this by bisection or inspection?

prckent (approved):

prckent:
Closes #4666

ye-luo:
> Thanks. An important source of worry removed. Did you find this by bisection or inspection?

After I printed the delta values and found nothing to worry, I had a rough idea about the cause and tested before/after merging #4407.

PR #4758, opened 2023-10-08
Rerun matrix inversion when failure happens (non-batched inversion case)

ye-luo:
## Proposed changes
One failure mode of 512 atom running on Summit.
cuSolver was showing non-deterministic failure. Re-doing inversion seems to rescue the run.
Add additional scanning of NaN.

- New feature

- No

Summit.

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
+++ b/src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
@@ -64,7 +64,49 @@ void DiracDeterminant<DU_TYPE>::invertPsiM(const ValueMatrix& logdetT, ValueMatr
 {
   ScopedTimer local_timer(InverseTimer);
   if (matrix_inverter_kind_ == DetMatInvertor::ACCEL)
-    updateEng.invert_transpose(logdetT, invMat, log_value_);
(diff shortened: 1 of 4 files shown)


prckent (approved):
Suggestions for later updates:

- The batched case - urgently need to report status of 128NiO batched
- Instead of excepting on the first found NaN, perhaps scan the whole matrix and give the status of all the bad elements, then except? Will tell us if we have bad rows/cols or single elements etc.
- (Minor)The magic retry number 5 should be a named const

prckent:
Test this please

ye-luo:
Test this please

ye-luo:
> Suggestions for later updates:
> 
>     * The batched case - urgently need to report status of 128NiO batched

Will update the too tight NaN check.

>     * Instead of excepting on the first found NaN, perhaps scan the whole matrix and give the status of all the bad elements (or first 200+total), then except? Will tell us if we have bad rows/cols or single elements etc.

Right now we scan all the diagonal elements. Scanning all the elements seems too expensive.

>     * (Minor)The magic retry number 5 should be a named const

prckent (approved):

PR #4759, opened 2023-10-09
Relax NaN check in TWF::calcRatioGrad

ye-luo:
## Proposed changes
When the ratio of calcRatioGrad returns 0, gradients can be NaN. The accept/reject algorithm naturally rejects such moves and thus the NaN check should be skipped. I updated the comment of calcRatioGrad to remind its caller checking the ratio value before accepting a move and fixed its use cases in NLPP T-move related routines.

- Bugfix

- No

Summit

## Checklist


ye-luo:
Test this please

prckent (approved):

PR #4760, opened 2023-10-09
Consolidate QMCHamiltonian code paths.

ye-luo:
## Proposed changes
Rename `updateNonKinetic` to `updateComponent` and adjust `updateKinetic`. Maximize code reuse.

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCHamiltonians/QMCHamiltonian.cpp
+++ b/src/QMCHamiltonians/QMCHamiltonian.cpp
@@ -543,25 +543,13 @@ QMCHamiltonian::FullPrecRealType QMCHamiltonian::evaluate(ParticleSet& P)
   for (int i = 0; i < H.size(); ++i)
   {
     ScopedTimer h_timer(my_timers_[i]);
-    const auto LocalEnergyComponent = H[i]->evaluate(P);
-    if (qmcplusplus::isnan(LocalEnergyComponent))
-    {
-      std::ostringstream msg;
-      msg << "QMCHamiltonian::evaluate component " << H[i]->getName() << " returns NaN." << std::endl;
-      P.print(msg);
-      throw std::runtime_error(msg.str());
-    }
-    LocalEnergy += LocalEnergyComponent;
-    H[i]->setObservables(Observables);
+    H[i]->evaluate(P);
+    updateComponent(*H[i], *this, P);
 #if !defined(REMOVE_TRACEMANAGER)
     H[i]->collectScalarTraces();
 #endif
-    H[i]->setParticlePropertyList(P.PropertyList, myIndex);
   }
-  KineticEnergy                      = H[0]->getValue();
-  P.PropertyList[WP::LOCALENERGY]    = LocalEnergy;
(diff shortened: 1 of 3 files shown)


PDoakORNL (commented):
Looking at this now.

src/QMCHamiltonians/QMCHamiltonian.cpp
```diff
@@ -572,34 +560,22 @@ QMCHamiltonian::FullPrecRealType QMCHamiltonian::evaluateDeterministic(ParticleS
...
 #if !defined(REMOVE_TRACEMANAGER)
     H[i]->collectScalarTraces();
 #endif
-    H[i]->setParticlePropertyList(P.PropertyList, myIndex);
   }
-  KineticEnergy                      = H[0]->getValue();
-  P.PropertyList[WP::LOCALENERGY]    = LocalEnergy;
-  P.PropertyList[WP::LOCALPOTENTIAL] = LocalEnergy - KineticEnergy;
-  // auxHevaluate(P);
+  updateKinetic(*this, P);
```
PDoakORNL, line 569: I wish there was a unit test here and elsewhere to check for instance that the LocalEnergy value is indeed the same, but this refactor looks ok by instpection.

PDoakORNL (changes requested):
QMCHamiltonian could really use a unit test to either confirm the local energies the methods return are correct/consistent with those coming via the property lists. This would then be simple to approve.  But as it stands this is a refactor that is missing the unit test it should depend on.

src/QMCHamiltonians/QMCHamiltonian.cpp
ye-luo: Did you mean testing
`P.PropertyList[WP::LOCALENERGY]   == ham.getLocalEnergy()` ?

ye-luo: I added checks in one of the unit test.

src/QMCHamiltonians/tests/test_QMCHamiltonian.cpp
```diff
@@ -203,6 +204,8 @@ TEST_CASE("integrateListeners", "[hamiltonian]")
 
   QMCHamiltonian::mw_evaluate(ham_list, twf_list, p_list);
```
PDoakORNL, line 205: this call also returns a std::vector of local energy which should match. 

And the sum of the per particle component energies and the sum of local energies per particle sum should match this returned value as well. (This is a test I should have added so you can just drop an issue for me.)

ye-luo: > this call also returns a std::vector of local energy which should match.

I added checks for this.

> And the sum of the per particle component energies and the sum of local energies per particle sum should match this returned value as well. (This is a test I should have added so you can just drop an issue for me.)

Added https://github.com/QMCPACK/qmcpack/issues/4769

ye-luo:
Test this please

ye-luo:
Test this please

PDoakORNL (approved):
LGTM

PR #4763, opened 2023-10-10
[BUG] Fix wrong PhaseDiff and protect NaN for DMCBatched

ye-luo:
## Proposed changes
Found a severe bug in the batched DMC driver. It was introduced in 534fffc9d213935df04725305d89061a7aec1fd0
PhaseDiff is constantly 0 due to `std::imag(std::arg(ratios[iw]))`, the return of `std::arg` is always a real number. It doesn't affect complex build but breaks fixed node approximation in the real build. The insensitive `short-diamondC_2x1x1_pp-vmcbatch_dmcbatch` indicates lack of tests.

I would like to eventually remove TWF::PhaseDiff which is an unnecessary state to carry with. It can always be recomputed from ratio and the computation is cheap.

Also protect another source of NaN in DMCBatched.

- Bugfix

- No

epyc-server, frontier

## Checklist

### DIFF
--- a/src/QMCDrivers/DMC/DMCBatched.cpp
+++ b/src/QMCDrivers/DMC/DMCBatched.cpp
@@ -33,6 +33,7 @@ namespace qmcplusplus
 {
 using std::placeholders::_1;
 using WP = WalkerProperties::Indexes;
+using PsiValueType = TrialWaveFunction::PsiValueType;
 
 /** Constructor maintains proper ownership of input parameters
  *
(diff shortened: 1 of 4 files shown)


ye-luo:
Test this please

ye-luo:
Test this please

ye-luo:
I guess when tau is small, the chance of crossing node boundary is small. I did the following inspection.
1. I injected some printout with tests  `short-diamondC_2x1x1_pp-vmcbatch_dmcbatch`. Before this fix, no reject was triggered by phase change. After this fix, I saw rejections.
2. The corresponding deterministic tests, I didn't see any rejection regardless of the fix.
3. I don't see noticeable change of acceptance ratio in multiple runs I had. It means the rejection by sign change is very rare.
4. Probably a large time step can hit the rejection more frequently.

prckent (approved):
Well spotted. Agree that ideally phasediff would go as it simplifies the code. 

It is a puzzle that this was not more of a problem. Progress on #4764 needed in case there are other occurrences.

Issue #4767, opened 2023-10-11
NaN check failure with clang in offload build in multi slater determinant test


markdewing:
Built using Clang for offload (`-D ENABLE_OFFLOAD=1`, the setting of `ENABLE_CUDA` does not matter)
Tested LLVM versions are 16, 17 and 18.
Reproduces in release and debug builds.

The unit test test_wavefunction_determinant fails: 
```
-------------------------------------------------------------------------------
LiH multi Slater dets precomputed_table_method
-------------------------------------------------------------------------------
/home/mdewing/nvme/physics/qmcpack/main/qmcpack/src/QMCWaveFunctions/tests/test_multi_slater_determinant.cpp:360
...............................................................................

/home/mdewing/nvme/physics/qmcpack/main/qmcpack/src/QMCWaveFunctions/tests/test_multi_slater_determinant.cpp:360: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  NaN check in TWF::mw_calcRatioGrad found
    particle 1 grads[1] is NaN.
```
It may depend on the CUDA version.  With CUDA 11.2.2 the test fails, but in a different way (values don't match).  CUDA 11.4, 11.6 and 12.2 show the NaN failure.

The problem seems to be an assignment of a TinyVector after a reduction, here:
https://github.com/QMCPACK/qmcpack/blob/bb8c8872ff710260c28a1937ceefb7d3e06737ca/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp#L795
Adding the print statements
```
      ratioGradRef_list_ptr[iw] = ratioGradRef_local;
      printf("ratiograd1 local  %lu %g %g %g \n",iw,ratioGradRef_local[0], ratioGradRef_local[1], ratioGradRef_local[2]);
      printf("ratiograd1 device %lu %g %g %g \n",iw,ratioGradRef_list_ptr[iw][0], ratioGradRef_list_ptr[iw][1], ratioGradRef_list_ptr[iw][2]);
```

Produces the output
```
ratiograd1 local  1 -0.785575 0.785575 -3.95485 
ratiograd1 local  0 -0.477015 -0.477015 -0.965852 
ratiograd1 device 1 -0.785575 0 -3.95485 
ratiograd1 device 0 -0.477015 0 -0.965852 
```
The assignment fails for the second component of the vector.
(Side note on cuda version: the build with cuda 11.2.2 shows the correct assignment behavior here.)

[...]

PR #4768, opened 2023-10-11
Fix NaN check failure in offload build with clang

markdewing:
Addresses https://github.com/QMCPACK/qmcpack/issues/4767 by assigning each component of the TinyVector individually.

Also adds `ratioGradRef_list_ptr` to the mapping list. (Code works without it, but it seems like it ought to be there along with `curRatio_list_ptr`)

- Bugfix

- No

local server, JLSE machine with A40.

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
+++ b/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
@@ -781,7 +781,7 @@ void MultiDiracDeterminant::mw_evaluateDetsAndGradsForPtclMove(
 
 
     PRAGMA_OFFLOAD("omp target teams distribute is_device_ptr(psiV_list_devptr, psiMinv_temp_list_devptr) \
-		                    map(always, from:curRatio_list_ptr[:nw], ratioGradRef_list_ptr[:nw])")
+		                    map(always, from:curRatio_list_ptr[:nw])")
     for (size_t iw = 0; iw < nw; iw++)
     {
       GradType ratioGradRef_local(0);
--- a/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
+++ b/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
@@ -781,7 +781,7 @@ void MultiDiracDeterminant::mw_evaluateDetsAndGradsForPtclMove(
 
 
     PRAGMA_OFFLOAD("omp target teams distribute is_device_ptr(psiV_list_devptr, psiMinv_temp_list_devptr) \
-		                    map(always, from:curRatio_list_ptr[:nw])")
+		                    map(always, from:curRatio_list_ptr[:nw], ratioGradRef_list_ptr[:nw])")
     for (size_t iw = 0; iw < nw; iw++)
     {
       GradType ratioGradRef_local(0);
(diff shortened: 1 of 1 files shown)


markdewing:
There's a `to` map clause later
https://github.com/QMCPACK/qmcpack/blob/bb8c8872ff710260c28a1937ceefb7d3e06737ca/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp#L838
And it's not clear to me when a data transfer will be done, and when it won't.  If a data transfer occurs here, it seems like the device version would be overwritten by the host version (which would be incorrect since the updated value was never transferred from the device to the host).  But there probably isn't a data transfer here, under semantics I don't understand.

I should clarify my confusion by turning it into a more general question.
The `map(to...)` seems to be needed for two purposes:
1. Tell the compiler to use the device version inside offload region (sometimes the compiler can figure this out automatically, and sometimes it can't.  The LIBOMPTARGET_INFO output is useful for determining what the compiler thinks of all the variables in the offload region.)
2. Transfer data from host to device.  Clearly if this is the first time the data is used, then a transfer ought to happen.  But if it's inside a data enter/exit region, does a data transfer happen?

ye-luo:
> There's a `to` map clause later
> 
> https://github.com/QMCPACK/qmcpack/blob/bb8c8872ff710260c28a1937ceefb7d3e06737ca/src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp#L838
> 
> And it's not clear to me when a data transfer will be done, and when it won't. If a data transfer occurs here, it seems like the device version would be overwritten by the host version (which would be incorrect since the updated value was never transferred from the device to the host). But there probably isn't a data transfer here, under semantics I don't understand.
> 
> I should clarify my confusion by turning it into a more general question. The `map(to...)` seems to be needed for two purposes:
> 
>     1. Tell the compiler to use the device version inside offload region (sometimes the compiler can figure this out automatically, and sometimes it can't.  The LIBOMPTARGET_INFO output is useful for determining what the compiler thinks of all the variables in the offload region.)
> 
>     2. Transfer data from host to device.  Clearly if this is the first time the data is used, then a transfer ought to happen.  But if it's inside a data enter/exit region, does a data transfer happen?

When the reference count is non-zero, no transfer. Since `ratioGradRef_list` is dual space, the underlying memory has been enter/exit data mapped thus non-zero refcount.

That `map(to` is simply wrong, it should be removed.

markdewing:
I found the part of the OpenMP 5.2 specification that answers my question.  Section 5.8.3 (map clause), item 3b under the 'Semantics' section:

    If the reference count of the corresponding list item is one or if the always
    map-type-modifier is specified, and if the map-type is to or tofrom, the corresponding list
    item is updated as if the list item appeared in a to clause on a target update directive.

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
```diff
@@ -792,7 +792,14 @@ void MultiDiracDeterminant::mw_evaluateDetsAndGradsForPtclMove(
         psiV_temp_list_ptr[iw][i] = psiV_list_devptr[iw][J];
         ratioGradRef_local += psiMinv_temp_list_devptr[iw][i * psiMinv_cols + WorkingIndex] * dpsiV_list_ptr[iw][J];
       }
-      ratioGradRef_list_ptr[iw] = ratioGradRef_local;
+
+      // Workaround for https://github.com/QMCPACK/qmcpack/issues/4767
+      // Compiler-generated assignment fails for the second component.
+      // Workaround is to assign each component individually.
+      //ratioGradRef_list_ptr[iw] = ratioGradRef_local;
+      ratioGradRef_list_ptr[iw][0] = ratioGradRef_local[0];
```
PDoakORNL, line 800: This fix would be ok if we had no way to actually fix TinyVector, say it was a type from a closed source library.

This should be checked for in TinyVectors unit tests.  And an special assignment method for omptarget would make sense. It should just default to the normal assignment operator if the compiler figures this out properly.  And this needs to be documented in TinyVector so the next person to assign to one in a target section has a fighting chance.  They're not going to see this comment.  The documentation of the special assignment for TinyVector should end up in the doxygen.

PDoakORNL (changes requested):
Not surprised about this MultiDiracDeterminant has no unit tests.
But I suggest this is actually dealt with at the TinyVector level.

src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
markdewing: The underlying issue also affects complex numbers.  I'm trying to make standalone reproducer, but I'm not having much success.

markdewing:
Have a better workaround.

PR #4773, opened 2023-10-13
Workaround for assignment-after-reduction offload bug

markdewing:
Loop index over nw must be 32 bits in size, otherwise the assignment following the reduction will miss the second item.
Bug affects offload with NVidia.
See https://github.com/QMCPACK/qmcpack/issues/4767

This PR represents the minimal changes needed to fix the real and complex offload builds for the multideterminant tests.
More changes could be made to make the type of `nw` consistent (A static analyzer would complain if run against the current code).

- Bugfix

- No

local server (A2000), JLSE A40

## Checklist


ye-luo (approved):
Good fix. We probably better to change all the GPU offload related loop to uint32_t. I think all the GPUs naturally handle 32 bit integer with 64 bit addressing.

markdewing:
Test this please

Issue #4774, opened 2023-10-13
Use uint32_t as index type for offload loops


ye-luo:
See https://github.com/QMCPACK/qmcpack/pull/4773#pullrequestreview-1677034614

PR #4777, opened 2023-10-16
Use 32bit integer for loop indices in offload region.

ye-luo:
## Proposed changes
Close https://github.com/QMCPACK/qmcpack/issues/4774
Use int or uint32_t in offload loop indices.

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
@@ -707,7 +707,7 @@ void SplineC2ROMPTarget<ST>::mw_evaluateVGL(const RefVectorWithLeader<SPOSet>& s
                                             const RefVector<ValueVector>& d2psi_v_list) const
 {
   assert(this == &sa_list.getLeader());
-  auto& phi_leader = sa_list.getCastedLeader<SplineC2ROMPTarget<ST>>();
+  auto& phi_leader         = sa_list.getCastedLeader<SplineC2ROMPTarget<ST>>();
   auto& mw_mem             = phi_leader.mw_mem_handle_.getResource();
   auto& mw_pos_copy        = mw_mem.mw_pos_copy;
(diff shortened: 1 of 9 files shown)


src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
```diff
@@ -866,7 +866,7 @@ void SplineC2ROMPTarget<ST>::mw_evaluateVGLandDetRatioGrads(const RefVectorWithL
             omptarget::min(last_cplx + omptarget::min(nComplexBands_local, last_cplx), requested_orb_size);
         ValueType ratio(0), grad_x(0), grad_y(0), grad_z(0);
         PRAGMA_OFFLOAD("omp parallel for reduction(+: ratio, grad_x, grad_y, grad_z)")
-        for (size_t j = first_real; j < last_real; j++)
+        for (int j = first_real; j < last_real; j++)
```
PDoakORNL, line 869: Why int for just a couple?

PDoakORNL (commented):
Question about consistency

src/QMCWaveFunctions/BsplineFactory/SplineC2ROMPTarget.cpp
ye-luo: A lot of places already has `int`. There is no need to change. I don't feel the need of 100% consistency. `int` for signed and `unit32_t` for unsigned. 64bit to 32bit is critical. signed vs unsigned is not.

PDoakORNL (approved):
This looks fine. I wish it were totally consistent.

I'd strongly urge that any integer variable including for indexes, that are going to be used to do more than simple indexing into an array i.e. have any signed math or interconversion with signed types just be signed from the start. The loss of one bit is worth avoiding these sort of bugs. 

An intersting detail we get here is if expression like `i * lda + j` are ever greater than the max uint32_t value they will wrap around since this is unsigned integer artithmetic.

ye-luo:
Test this please

PR #4780, opened 2023-10-18
Offload AO evaluateV and VP related valueonly rourtines

kgasperich:
## Proposed changes

This is a step toward offloading all of the LCAO evaluation.
This PR offloads the AO evaluation for virtual particles (i.e. for NLPP evaluation) as well as the corresponding AO->MO gemm.

- added [`SoaAtomicBasisSet::mw_evaluateV`](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L682-L852) (AO evaluateV w/ OMP offload)
  - added [`SoaCartesianTensor::batched_evaluateV`](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/Numerics/SoaCartesianTensor.h#L85-L116) and static [`SoaCartesianTensor::evaluate_bare_impl`](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/Numerics/SoaCartesianTensor.h#L182-L285) (not yet added to codegen script, can be added later after VGL is also done)
  - added [`SoaSphericalTensor::batched_evaluateV`](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/Numerics/SoaSphericalTensor.h#L73-L108) and [`SoaSphericalTensor::evaluate_bare_impl`](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/Numerics/SoaSphericalTensor.h#L214-L315)
  - added [`MultiQuinticSpline1D::batched_evaluate`](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h#L155-L215), which required new data member [`LogGridLight.LogDelta`](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h#L43) (not sure if it's better to use `double` here or `T` like `{lower,upper}_bound` and `OneOverLogDelta`; [elsewhere in the class](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h#L52) a similar quantity is `double` instead of `T`, so that's why I made it a double, but I'm open to changing it if anyone has a strong opinion about it).
    - This class should eventually get a mw memory resource for the spline coeffs.
  - added [`MultiFunctorAdapter::batched_evaluate`](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/QMCWaveFunctions/LCAO/MultiFunctorAdapter.h#L62-L91); this is all done on the host and requires D2H transfer of nuc-elec distances and H2D transfer of radial wf component values.  This should be fixed eventually, but in the interest of time I'd been told not to worry about it for now.
- changed [`SoaLocalizedBasisSet::mw_evaluateValueVPs`](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/QMCWaveFunctions/LCAO/SoaLocalizedBasisSet.cpp#L251-L312) to call new offload AO evaluation
  - [calculation of translation vectors and displacements](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/QMCWaveFunctions/LCAO/SoaLocalizedBasisSet.cpp#L277-L291C6) is still done on the CPU.  I think I need an extra level of indirection for the flattened walker/vp iteration if that is to be done on GPU (could probably be done with just a cumulative sum of VPs per walker).  I could also use some advice about the best way to access the distance tables on the GPU (I assume I can get them from [`mw_r_dr`](https://github.com/kgasperich/qmcpack/blob/382779d2e0f4a62fd99721548aa6105e353139ff/src/Particle/SoaDistanceTableABOMPTarget.h#L46) if I get the distance table from the VP_list leader, but I'm not sure if it's safe to assume that will always be a `SoaDistanceTableABOMPTarget` or if it will sometimes be just a `SoaDistanceTableAB`)

Also, I needed to make [this change](https://github.com/kgasperich/qmcpack/commit/8076a49904091145c987ccd17b4075df379b5d56) to get around a segfault I was seeing when running on sunspot.

- New feature

- No

Sunspot (can provide modules and config/build details)
I've been using the LCAO diamond [here](https://github.com/kgasperich/qmcpack/tree/offload-vp-ao/tests/solids/diamondC_1x1x1-Gaussian_pp) as a test case (with a slightly modified input in order to use the batched VMC driver and add more walkers)
## Checklist


kgasperich:
just saw that the complex build is failing; I think I was maybe overzealous in reusing template parameters; I'll go through and fix those (I assume it's something like a ValueType used as coordinates)

ye-luo:
Please keep the GEMM change out of this PR.

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
```diff
@@ -162,8 +162,8 @@ void LCAOrbitalSet::evaluateValue(const ParticleSet& P, int iat, ValueVector& ps
 }
 
 /** Find a better place for other user classes, Matrix should be padded as well */
-template<typename T, unsigned D>
-inline void Product_ABt(const VectorSoaContainer<T, D>& A, const Matrix<T>& B, VectorSoaContainer<T, D>& C)
+template<typename T, unsigned D, typename Alloc>
```
ye-luo, line 165: Is this GEMM related change?

src/Numerics/SoaCartesianTensor.h
```diff
@@ -140,6 +179,111 @@ SoaCartesianTensor<T>::SoaCartesianTensor(const int l_max, bool addsign) : Lmax(
 }
 
 
+PRAGMA_OFFLOAD("omp declare target")
+template<class T>
+void SoaCartesianTensor<T>::evaluate_bare_impl(T x, T y, T z, T* restrict XYZ, size_t Lmax_)
```
ye-luo, line 184: Could you make `evaluate_bare` calling `evaluate_bare_impl` to minimize code duplication?

src/Numerics/SoaSphericalTensor.h
```diff
@@ -169,6 +211,109 @@ inline SoaSphericalTensor<T>::SoaSphericalTensor(const int l_max, bool addsign)
     }
 }
 
+PRAGMA_OFFLOAD("omp declare target")
+template<typename T>
+inline void SoaSphericalTensor<T>::evaluate_bare_impl(const T x,
```
ye-luo, line 216: Could you make `evaluate_bare` calling `evaluate_bare_impl` to minimize code duplication?

src/Particle/VirtualParticleSet.h
```diff
@@ -95,6 +95,15 @@ class VirtualParticleSet : public ParticleSet
...
                  int iat     = -1);
 
+  inline size_t getTotalNum() const { return TotalNum; }
+  /**Extract list of Distance Tables
+    */
+  static const RefVectorWithLeader<const DistanceTableAB> extractDTRefList_vp(
+      const RefVectorWithLeader<const VirtualParticleSet>& vp_list,
+      int id);
+
+  static const std::vector<QMCTraits::PosType> extractCoordsRefList_vp(
```
ye-luo, line 105: Please document this function.

src/Particle/VirtualParticleSet.h
```diff
@@ -95,6 +95,15 @@ class VirtualParticleSet : public ParticleSet
                  bool sphere = false,
                  int iat     = -1);
 
+  inline size_t getTotalNum() const { return TotalNum; }
+  /**Extract list of Distance Tables
+    */
+  static const RefVectorWithLeader<const DistanceTableAB> extractDTRefList_vp(
```
ye-luo, line 101: Could you remove `_vp`?

src/QMCWaveFunctions/LCAO/MultiFunctorAdapter.h
```diff
@@ -57,6 +59,37 @@ struct MultiFunctorAdapter
...
+      {
+        for (size_t i = 0, n = Rnl.size(); i < n; ++i)
+          u_ptr[ir * nRnl + i] = 0.0;
+      }
+      else
+      {
+        for (size_t i = 0, n = Rnl.size(); i < n; ++i)
+          u_ptr[ir * nRnl + i] = Rnl[i]->f(r_ptr[ir]);
+      }
+    }
```
ye-luo, line 88: More of my taste. Remove all `{}` on lines 77-88.

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
kgasperich: yes, I can remove it for now along with the other changes to have dual-space coeffs (this function; added typedef to SpoSet.h; changed type of C and C_copy in LCAO).

src/Numerics/SoaSphericalTensor.h
kgasperich: I'm not sure how to handle it for SoaSphericalTensor.
[`evaluateVGL` calls `evaluate_bare`](https://github.com/kgasperich/qmcpack/blob/5821946cf2a6fff89c589c21fbbb66e1a84af9b7/src/Numerics/SoaSphericalTensor.h#L415-L416) and passes the scratch space of the object, but `evaluate_bare_impl` is static, so it shouldn't modify `cYlm`.
This is what I get when I try to have `evaluate_bare` call `evaluate_bare_impl`
```
27: AddressSanitizer:DEADLYSIGNAL
27: =================================================================
27: ==1264599==ERROR: AddressSanitizer: SEGV on unknown address 0x03e800134bd7 (pc 0x561133b7a46b bp 0x7ffff530a780 sp 0x7ffff530a720 T0)
27: ==1264599==The signal is caused by a READ memory access.
27:     #0 0x561133b7a46b in qmcplusplus::SoaSphericalTensor<double>::evaluate_bare_impl(double, double, double, double*, unsigned long, double const*, double const*) (/home/keving/work/qmcpack/qmcpack-ao-resource/build_asan/src/QMCWaveFunctions/tests/test_wavefunction_sposet+0x17bf46b) (BuildId: 34e365f3ed46d03bce9509e4c7975c5a04bb5060)
27:     #1 0x561133b7d9a9 in qmcplusplus::SoaSphericalTensor<double>::evaluateVGL(double, double, double) (/home/keving/work/qmcpack/qmcpack-ao-resource/build_asan/src/QMCWaveFunctions/tests/test_wavefunction_sposet+0x17c29a9) (BuildId: 34e365f3ed46d03bce9509e4c7975c5a04bb5060)
```

I think there's a similar issue with [`SoaCartesianTensor<T>::evaluateV(T, T, T)` ](https://github.com/kgasperich/qmcpack/blob/5821946cf2a6fff89c589c21fbbb66e1a84af9b7/src/Numerics/SoaCartesianTensor.h#L66-L83)

kgasperich: I would've thought that the static function could modify the data since it was being passed as an argument rather than accessing it via `this`, so maybe the error comes from something else?

The change I made to get that error was to just have `evaluate_bare` call `evaluate_bare_impl`

```
template<typename T>
inline void SoaSphericalTensor<T>::evaluate_bare(T x, T y, T z, T* restrict Ylm) const
{
  evaluate_bare_impl(x, y, z, Ylm, Lmax, FactorL.data(), FactorLM.data());
}
```

maybe it's a `const` related issue?

src/QMCWaveFunctions/LCAO/MultiFunctorAdapter.h
kgasperich: fixed

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
kgasperich: fixed

src/Particle/VirtualParticleSet.h
kgasperich: done

src/Particle/VirtualParticleSet.h
kgasperich: done

src/Numerics/SoaCartesianTensor.h
kgasperich: fixed

src/Numerics/SoaCartesianTensor.h
```diff
@@ -56,27 +60,61 @@ struct SoaCartesianTensor
...
   ///compute Ylm
   inline void evaluateV(T x, T y, T z) { evaluateV(x, y, z, cXYZ.data(0)); }
 
+  /**
+   * @brief evaluate for multiple electrons and multiple pbc images
+   * 
+   * @param [in] xyz electron positions [Nelec, Npbc, 3(x,y,z)]
+   * @param [out] XYZ Cartesian tensor elements [Nelec, Npbc, Nlm]
+  */
+  inline void batched_evaluateV(OffloadArray3D& xyz, OffloadArray3D& XYZ) const
```
ye-luo, line 90: put const on `in` parameters.

src/Numerics/SoaCartesianTensor.h
```diff
@@ -56,27 +60,61 @@ struct SoaCartesianTensor
...
+    assert(xyz.size(2) == 3);         // x,y,z
+
+    assert(XYZ.size(0) == nElec);
+    assert(XYZ.size(1) == Npbc);
+    const size_t Nlm = XYZ.size(2);
+
+    size_t nR = nElec * Npbc; // total number of positions to evaluate
+
+    auto* xyz_devptr     = xyz.device_data();
+    auto* XYZ_devptr     = XYZ.device_data();
```
ye-luo, line 103: Avoid explicit device pointer. Use `data()` and `map` instead of `is_device_ptr`.

src/Numerics/SoaCartesianTensor.h
```diff
@@ -56,27 +60,61 @@ struct SoaCartesianTensor
...
+    auto* NormFactor_ptr = NormFactor.data();
+
+    PRAGMA_OFFLOAD("omp target teams distribute parallel for map(to:NormFactor_ptr[:Nlm]) \
```
ye-luo, line 106: Is `NormFactor` constant? If so, put it in dual space.

src/Numerics/SoaSphericalTensor.h
```diff
@@ -55,21 +59,58 @@ struct SoaSphericalTensor
...
+    auto* xyz_devptr     = xyz.device_data();
+    auto* Ylm_devptr     = Ylm.device_data();
+    auto* flm_ptr        = FactorLM.data();
+    auto* fl_ptr         = FactorL.data();
+    auto* NormFactor_ptr = NormFactor.data();
+
+
+    PRAGMA_OFFLOAD("omp target teams distribute parallel for \
+                    map(always, to:flm_ptr[:Nlm], fl_ptr[:Lmax+1], NormFactor_ptr[:Nlm]) \
+                    is_device_ptr(xyz_devptr, Ylm_devptr)")
```
ye-luo, line 99: Are `FactorLM` and `FactorL` constants? If so, put them in dual space and transfer ahead.

src/Numerics/SoaSphericalTensor.h
```diff
@@ -169,8 +210,15 @@ inline SoaSphericalTensor<T>::SoaSphericalTensor(const int l_max, bool addsign)
...
 }
 
+PRAGMA_OFFLOAD("omp declare target")
 template<typename T>
-inline void SoaSphericalTensor<T>::evaluate_bare(T x, T y, T z, T* restrict Ylm) const
+inline void SoaSphericalTensor<T>::evaluate_bare(const T x,
+                                                 const T y,
+                                                 const T z,
+                                                 T* restrict Ylm,
+                                                 const int Lmax_,
```
ye-luo, line 219: Do not put underscore. `XXX_` is reserved for class private member variables. Use `lmax` instead.

src/Numerics/SoaSphericalTensor.h
```diff
@@ -169,8 +210,15 @@ inline SoaSphericalTensor<T>::SoaSphericalTensor(const int l_max, bool addsign)
...
+                                                 const T* FactorL_ptr,
+                                                 const T* FactorLM_ptr)
```
ye-luo, line 221: Use lower case for variables. `factorL` and `factorLM`.

src/Particle/SoaDistanceTableABOMPTarget.h
```diff
@@ -329,7 +332,7 @@ class SoaDistanceTableABOMPTarget : public DTD_BConds<T, D, SC>, public Distance
       if (!(modes_ & DTModes::MW_EVALUATE_RESULT_NO_TRANSFER_TO_HOST))
       {
         PRAGMA_OFFLOAD(
-            "omp target update from(r_dr_ptr[:mw_r_dr.size()]) depend(inout:r_dr_ptr[:mw_r_dr.size()]) nowait")
+            "omp target update from(r_dr_ptr[:num_r_dr]) depend(inout:r_dr_ptr[:num_r_dr]) nowait")
```
ye-luo, line 335: Revert unrelated changes unrelated to this PR.

src/Particle/VirtualParticleSet.h
ye-luo: This function doesn't make sense at all. VP doesn't have active particle coordinates. Something is wrong in the caller side.

src/QMCWaveFunctions/LCAO/MultiFunctorAdapter.h
```diff
@@ -57,6 +59,31 @@ struct MultiFunctorAdapter
       u[i] = Rnl[i]->f(r);
   }
 
+  inline void batched_evaluate(OffloadArray2D& r, OffloadArray3D& u, RealType Rmax) const
```
ye-luo, line 62: Missing documentation. Cannot understand what are r and u. Need to document u r and u and mention the expected data locality about whether they are up-to-date on host or device or both.

src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h
```diff
@@ -137,6 +152,62 @@ class MultiQuinticSpline1D
...
+    auto* u_devptr = u.device_data();
+
+    auto* coeff_devptr       = coeffs->device_data();
+    auto* first_deriv_devptr = first_deriv.device_data();
+    const size_t nCols       = coeffs->cols();
+    // const size_t coefsize = coeffs->size();
+
+
+    PRAGMA_OFFLOAD("omp target teams distribute parallel for \
+                    is_device_ptr(r_devptr, u_devptr, coeff_devptr, first_deriv_devptr)")
```
ye-luo, line 180: do not use `device_ptr` unless it is absolutely necessary.

src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h
```diff
@@ -47,11 +51,20 @@ struct LogGridLight
...
     double dlog_ratio = log_ratio / static_cast<double>(n - 1);
     OneOverLogDelta   = 1.0 / dlog_ratio;
+    LogDelta          = dlog_ratio;
     r_values.resize(n);
     for (int i = 0; i < n; i++)
       r_values[i] = ri * std::exp(i * dlog_ratio);
   }
 
+  PRAGMA_OFFLOAD("omp declare target")
+  static inline T getCL(T r, int& loc, double OneOverLogDelta_, T lower_bound_, double dlog_ratio_)
```
ye-luo, line 61: Avoid using `_` it violates coding convention. Upper case is also reserved for type names. So `OneOverLogDelta_` is bad.
`OneOverLogDelta_` => `one_over_log_delta`
`lower_bound_` => `lower_bound`
`dlog_ratio_` => `log_delta`

src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h
```diff
@@ -137,6 +152,62 @@ class MultiQuinticSpline1D
...
+
+    double OneOverLogDelta = myGrid.OneOverLogDelta;
+    T lower_bound          = myGrid.lower_bound;
+    T dlog_ratio           = myGrid.LogDelta;
+
+    auto* r_devptr = r.device_data();
+    auto* u_devptr = u.device_data();
+
+    auto* coeff_devptr       = coeffs->device_data();
+    auto* first_deriv_devptr = first_deriv.device_data();
```
ye-luo, line 174: Avoid using device pointers.

src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h
```diff
@@ -274,6 +345,8 @@ class MultiQuinticSpline1D
         out[(i * 6 + 5) * ncols + ispline] = static_cast<T>(F[i]);
       }
     }
+    first_deriv.updateTo();
+    coeffs->updateTo(); // FIXME: this is overkill, probably better to just do once after all splines added
```
ye-luo, line 349: Both updateTo are overkill. Update comment please.

src/QMCWaveFunctions/LCAO/MultiFunctorAdapter.h
kgasperich: I added documentation.
r is [n_elec, n_pbc]; u is [n_elec, n_pbc, n_Rnl]

n_elec and n_pbc aren't really important on their own, and it wouldn't change anything if they were flattened to one dimension (this is the same in MultiQuinticSpline1D and Soa{Cartesian,Spherical}Tensor), but outside of this function the n_elec, n_pbc need to be treated separately.

This could be written in a more general way where r is 1D and u is 2D, but seems like too much work for now.  If I need to add one for 1D r and 2D u, I can have them both call some _impl function where I just pass array dims and pointers to the start of u/r.

src/Particle/VirtualParticleSet.h
kgasperich: I updated it and gave it a better name

```
const std::vector<QMCTraits::PosType> VirtualParticleSet::extractVPCoords(
    const RefVectorWithLeader<const VirtualParticleSet>& vp_list)
{
  std::vector<QMCTraits::PosType> coords_list;
  for (const VirtualParticleSet& vp : vp_list)
    for (int iat = 0; iat < vp.getTotalNum(); iat++)
      coords_list.push_back(vp.R[iat]);

  return coords_list;
}
```

I'm not sure the returned vector needs to be const, and I don't think std::vector<PosType> is a great choice, but this part needs to be cleaned up anyway.

I think the ideal case would be to just directly access the coordinates through the distance table mw_mem, but I'm not sure how to do that.
Could I get the DistanceTableAB from the leader and then get [`mw_r_dr`](https://github.com/kgasperich/qmcpack/blob/65a2d8ffd5b9cd3b055750f8770c9ab2e18e0295/src/Particle/SoaDistanceTableABOMPTarget.h#L46) from that distance table? (after casting it to a `SoaDistanceTableABOMPTarget`?)

kgasperich:
@ye-luo @anbenali 
 I'm having trouble tracking down a bug in the detRatio unit tests.  If you have any insight I'd appreciate it.
https://github.com/QMCPACK/qmcpack/actions/runs/6619602610?pr=4780

I return the MW ratios in `ratios_list` and the single-walker (SW) ratios in `ratios_ref_0` and `ratios_ref_1`.
I check that the MW and SW ratios are the same, and then I check that the SW and MW values are correct (I guess this is redundant and I only need to compare them to each other and then compare either the MW or SW values to the reference values)

The tests fail in several different ways:
- the [MW ratios](https://github.com/QMCPACK/qmcpack/pull/4780/files#diff-de74b3b8139247baa607b38540a969400a9018096598f601cf69290cba118c6dR187-R188) are all nan, and the [SW ratios](https://github.com/QMCPACK/qmcpack/pull/4780/files#diff-de74b3b8139247baa607b38540a969400a9018096598f601cf69290cba118c6dR192-R193) have incorrect values
-  the MW ratios are 0 and the SW ratios are correct
- MW and SW ratios agree with each other, but not with the reference values


[GCC9-NoMPI-Debug-Real](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980457837?pr=4780#step:6:235) MW: nan, SW: wrong value
[GCC9-NoMPI-NoOMP-Real](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980458186?pr=4780#step:6:225)  MW: nan, SW: wrong value
[GCC9-NoMPI-NoOMP-Complex](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980458514?pr=4780#step:6:201)  MW: nan, SW: wrong value

[GCC9-MPI-Gcov-Real](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980458956?pr=4780#step:6:253) Pass
[GCC9-MPI-Gcov-Complex](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980459103?pr=4780#step:6:229) Pass

[GCC11-NoMPI-Werror-Real](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980459242?pr=4780#step:6:243) Pass
[GCC11-NoMPI-Werror-Complex](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980459377?pr=4780#step:6:219)  MW: nan, SW: wrong value
[GCC11-NoMPI-Werror-Real-Mixed](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980459581?pr=4780#step:6:189)  MW: nan, SW: wrong value
[GCC11-NoMPI-Werror-Complex-Mixed](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980459760?pr=4780#step:6:169)  MW: wrong value, SW: wrong value (MW and SW agree)

[Clang14-NoMPI-ASan-Real](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980459915?pr=4780#step:6:235) Pass
[Clang14-NoMPI-ASan-Complex](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980460060?pr=4780#step:6:211) Pass
[Clang14-NoMPI-UBSan-Real](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980460192?pr=4780#step:6:235) Pass

[Clang16-NoMPI-Offload-Real](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980460324?pr=4780#step:6:115)  MW: 0, SW: correct

[macOS-GCC11-NoMPI-Real](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980457338?pr=4780#step:7:248)  MW: nan, SW: wrong value

[GCC11-NoMPI-Debug-Real](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980457539?pr=4780#step:7:251)  MW: nan, SW: wrong value
[GCC11-NoMPI-NoOMP-Real](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980458018?pr=4780#step:7:231)  MW: nan, SW: wrong value
[GCC11-NoMPI-NoOMP-Complex](https://github.com/QMCPACK/qmcpack/actions/runs/6619602610/job/17980458362?pr=4780#step:7:207) Pass

The test failure appears for some builds, but not others.  I don't see anything obvious from asan/msan/ubsan, and I see failures even when only using one OpenMP thread.

In testing on my machine, I've also seen cases where the tests will pass when called via ctest, but fail when I run the test executable directly from the test directory.
The way this is failing looks like some bad memory access, but sanitizers don't indicate that.
Using ctest:
```
keving@keving-p1g5:~/work/qmcpack/qmcpack-ao-resource/build_ubsan$ ctest -V --output-on-failure -R tmp
UpdateCTestConfiguration  from :/home/keving/work/qmcpack/qmcpack-ao-resource/build_ubsan/DartConfiguration.tcl
Parse Config file:/home/keving/work/qmcpack/qmcpack-ao-resource/build_ubsan/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/keving/work/qmcpack/qmcpack-ao-resource/build_ubsan/DartConfiguration.tcl
Parse Config file:/home/keving/work/qmcpack/qmcpack-ao-resource/build_ubsan/DartConfiguration.tcl
Test project /home/keving/work/qmcpack/qmcpack-ao-resource/build_ubsan
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 30
    Start 30: deterministic-unit_test_wavefunction_tmp

30: Test command: /home/keving/work/qmcpack/qmcpack-ao-resource/build_ubsan/src/QMCWaveFunctions/tests/test_wavefunction_tmp
30: Working Directory: /home/keving/work/qmcpack/qmcpack-ao-resource/build_ubsan/src/QMCWaveFunctions/tests
30: Environment variables: 
30:  OMP_NUM_THREADS=1
30: Test timeout computed to be: 3600
30: QMCPACK printout is suppressed. Use --turn-on-printout to see all the printout.
30: ===============================================================================
30: All tests passed (23 assertions in 1 test case)
30: 
1/1 Test #30: deterministic-unit_test_wavefunction_tmp ...   Passed    0.04 sec

The following tests passed:
        deterministic-unit_test_wavefunction_tmp

100% tests passed, 0 tests failed out of 1

Label Time Summary:
deterministic      =   0.04 sec*proc (1 test)
quality_unknown    =   0.04 sec*proc (1 test)
unit               =   0.04 sec*proc (1 test)

Total Test time (real) =   0.09 sec
```
From test directory:
```
~/work/qmcpack/qmcpack-ao-resource/build_ubsan$ cd src/QMCWaveFunctions/tests; OMP_NUM_THREADS=1 ./test_wavefunction_tmp --turn-on-printout; cd -
==================================================
--- Memory usage report : Before running tests ---
==================================================
Available memory on node 0, free + buffers :    6622 MiB
Memory footprint by rank 0 on node 0       :      39 MiB
==================================================
  Adding ion0 ParticleSet to the pool
  ParticleSet 'ion0' contains 4 particles :  C(4)

    C                 0                 0                 0
    C       1.686580575       1.686580575       1.686580575
    C        3.37316115        3.37316115                 0
    C       5.059741726       5.059741726       1.686580575


  Adding elec ParticleSet to the pool
  LCAO: SoaAtomicBasisSet<MultiGTO,1>
Reading BasisSet from HDF5 file:C_Diamond.h5
<input node="atomicBasisSet" name="bfd-vtz" expandYlm="pyscf" angular="spherical" elementType="C" normalized="no" type="Numerical" expM="1" />
  AO BasisSet for C
   Spherical Harmonics contain (-1)^m factor
   Angular momentum m expanded as -l, ... ,l, with the exception of L=1 (1,-1,0)
   R(n,l,m,s) 0 0 0 0
   R(n,l,m,s) 1 0 0 0
   R(n,l,m,s) 2 0 0 0
   R(n,l,m,s) 3 1 0 0
   R(n,l,m,s) 4 1 0 0
   R(n,l,m,s) 5 1 0 0
   R(n,l,m,s) 6 2 0 0
   R(n,l,m,s) 7 2 0 0
   R(n,l,m,s) 8 3 0 0
Expanding Ylm as L=1 as (1,-1,0) and L>1 as -l,-l+1,...,l-1,l
   Adding 1 spherical orbitals
   Adding 1 spherical orbitals
   Adding 1 spherical orbitals
   Adding 3 spherical orbitals
   Adding 3 spherical orbitals
   Adding 3 spherical orbitals
   Adding 5 spherical orbitals
   Adding 5 spherical orbitals
   Adding 7 spherical orbitals
   Maximum Angular Momentum  = 3
   Number of Radial functors = 9
   Basis size                = 29

   Using Identity for the LCOrbitalSet 

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test_wavefunction_tmp is a Catch v2.13.9 host application.
Run with -? for options

-------------------------------------------------------------------------------
LCAO DiamondC_2x1x1
-------------------------------------------------------------------------------
/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:32
...............................................................................

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:195: FAILED:
  CHECK( std::real(ratios_list[0][ivp]) == Approx(std::real(ratios_ref_0[ivp])) )
with expansion:
  nan == Approx( -0.1345122899 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:195: FAILED:
  CHECK( std::real(ratios_list[0][ivp]) == Approx(std::real(ratios_ref_0[ivp])) )
with expansion:
  nan == Approx( 0.0094266907 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:195: FAILED:
  CHECK( std::real(ratios_list[0][ivp]) == Approx(std::real(ratios_ref_0[ivp])) )
with expansion:
  nan == Approx( -0.182138386 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:195: FAILED:
  CHECK( std::real(ratios_list[0][ivp]) == Approx(std::real(ratios_ref_0[ivp])) )
with expansion:
  nan == Approx( -0.2170559822 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:197: FAILED:
  CHECK( std::real(ratios_list[1][ivp]) == Approx(std::real(ratios_ref_1[ivp])) )
with expansion:
  nan == Approx( 0.0426385616 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:197: FAILED:
  CHECK( std::real(ratios_list[1][ivp]) == Approx(std::real(ratios_ref_1[ivp])) )
with expansion:
  nan == Approx( 0.9259574691 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:197: FAILED:
  CHECK( std::real(ratios_list[1][ivp]) == Approx(std::real(ratios_ref_1[ivp])) )
with expansion:
  nan == Approx( 0.8030272173 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:221: FAILED:
  CHECK( std::real(ratios_list[0][0]) == Approx(-0.11554491049855) )
with expansion:
  nan == Approx( -0.1155449105 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:222: FAILED:
  CHECK( std::real(ratios_list[0][1]) == Approx(0.19155774810121) )
with expansion:
  nan == Approx( 0.1915577481 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:223: FAILED:
  CHECK( std::real(ratios_list[0][2]) == Approx(-0.16063724839636) )
with expansion:
  nan == Approx( -0.1606372484 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:224: FAILED:
  CHECK( std::real(ratios_list[0][3]) == Approx(-0.37105113615831) )
with expansion:
  nan == Approx( -0.3710511362 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:225: FAILED:
  CHECK( std::real(ratios_list[1][0]) == Approx(-0.12705469585615) )
with expansion:
  nan == Approx( -0.1270546959 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:226: FAILED:
  CHECK( std::real(ratios_list[1][1]) == Approx(0.67930890998428) )
with expansion:
  nan == Approx( 0.67930891 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:227: FAILED:
  CHECK( std::real(ratios_list[1][2]) == Approx(0.83583922544552) )
with expansion:
  nan == Approx( 0.8358392254 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:229: FAILED:
  CHECK( std::real(ratios_ref_0[0]) == Approx(-0.11554491049855) )
with expansion:
  -0.1345122899 == Approx( -0.1155449105 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:230: FAILED:
  CHECK( std::real(ratios_ref_0[1]) == Approx(0.19155774810121) )
with expansion:
  0.0094266907 == Approx( 0.1915577481 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:231: FAILED:
  CHECK( std::real(ratios_ref_0[2]) == Approx(-0.16063724839636) )
with expansion:
  -0.182138386 == Approx( -0.1606372484 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:232: FAILED:
  CHECK( std::real(ratios_ref_0[3]) == Approx(-0.37105113615831) )
with expansion:
  -0.2170559822 == Approx( -0.3710511362 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:233: FAILED:
  CHECK( std::real(ratios_ref_1[0]) == Approx(-0.12705469585615) )
with expansion:
  0.0426385616 == Approx( -0.1270546959 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:234: FAILED:
  CHECK( std::real(ratios_ref_1[1]) == Approx(0.67930890998428) )
with expansion:
  0.9259574691 == Approx( 0.67930891 )

/home/keving/work/qmcpack/qmcpack-ao-resource/src/QMCWaveFunctions/tests/test_LCAO_diamondC_2x1x1.cpp:235: FAILED:
  CHECK( std::real(ratios_ref_1[2]) == Approx(0.83583922544552) )
with expansion:
  0.8030272173 == Approx( 0.8358392254 )

===============================================================================
test cases:  1 |  1 failed
assertions: 23 | 2 passed | 21 failed
```

kgasperich:
@ye-luo I added more output for debugging [here](https://github.com/kgasperich/qmcpack/commit/b5d1cd20862bd87c6c4f5fe4b4067bfd31796f08), and it looks like the electron displacements are different for full vs mixed precision (the output is cleaned a bit to remove duplicate/redundant information)

[Here are the two functions in SoaLocalizedBasisSet](https://github.com/kgasperich/qmcpack/blob/b5d1cd20862bd87c6c4f5fe4b4067bfd31796f08/src/QMCWaveFunctions/LCAO/SoaLocalizedBasisSet.cpp#L253-L336), and [this is where the single-walker one is called from](https://github.com/kgasperich/qmcpack/blob/b5d1cd20862bd87c6c4f5fe4b4067bfd31796f08/src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp#L610-L618).

Could this be some issue with how the distance is pulled from the distance table? In the single-walker function, we use the VP index as if it is an active particle index 
```
  const auto& displ   = (P.getActivePtcl() == iat) ? d_table.getTempDispls() : d_table.getDisplRow(iat);
```
but in the MW function we always just use `getDisplRow`
```
const auto& displ = dt_list[iw].getDisplRow(iat);
```
Although it doesn't seem like this can be the issue, because I assume there is only one active particle index, so that should only affect (at most) one of the VP positions in the single-walker function.

All of the positions are shifted by (-0.37255, -0.37255, -0.37255) for the mixed precision build, so maybe it has something to do with something like a minimum image convention being applied? (but the lattice vectors are 
```
6.7463223   6.7463223   0.0
0.0         3.37316115  3.37316115
3.37316115  0.0         3.37316115
```
so it's not immediately obvious to me where the 0.37255 comes from


```
FULL                                                       |   MIXED
mw_displ -1.832550647e-06 -1.000001833 -1.832550647e-06    |   mw_displ  -0.3725489974 -1.372549057 -0.3725489974
mw_displ -0.1000018326    -1.200001833 -0.3000018326       |   mw_displ  -0.4725489914 -1.572549105 -0.6725490093
mw_displ -0.2000018326    -1.400001833 -0.6000018326       |   mw_displ  -0.5725489855 -1.772548914 -0.9725490212
mw_displ -0.3000018326    -1.600001833 -0.9000018326       |   mw_displ  -0.6725490093 -1.972548962 -1.272548914
mw_displ -1.832550647e-06 -1.100001833 -1.832550647e-06    |   mw_displ  -0.3725489974 -1.472548962 -0.3725489974
mw_displ -0.2000018326    -1.400001833 -0.4000018326       |   mw_displ  -0.5725489855 -1.772549152 -0.7725490332
mw_displ -0.4000018326    -1.700001833 -0.8000018326       |   mw_displ  -0.7725490332 -2.072549105 -1.172549009
                                                           |
sw_displ -1.832550647e-06 -1.000001833 -1.832550647e-06    |   sw_displ  -0.3725489974 -1.372549057 -0.3725489974
sw_displ -0.1000018326    -1.200001833 -0.3000018326       |   sw_displ  -0.4725489914 -1.572549105 -0.6725490093
sw_displ -0.2000018326    -1.400001833 -0.6000018326       |   sw_displ  -0.5725489855 -1.772548914 -0.9725490212
sw_displ -0.3000018326    -1.600001833 -0.9000018326       |   sw_displ  -0.6725490093 -1.972548962 -1.272548914
sw_displ -1.832550647e-06 -1.100001833 -1.832550647e-06    |   sw_displ  -0.3725489974 -1.472548962 -0.3725489974
sw_displ -0.2000018326    -1.400001833 -0.4000018326       |   sw_displ  -0.5725489855 -1.772549152 -0.7725490332
sw_displ -0.4000018326    -1.700001833 -0.8000018326       |   sw_displ  -0.7725490332 -2.072549105 -1.172549009
```

kgasperich:
@ye-luo @anbenali the test seems to be okay now after Ye's fixes!  I'll clean up the tests and add a non-trivial complex one.

anbenali:
Fantastic. Let's wrap this up then. Thanks a lot for all the work

kgasperich:
@anbenali @ye-luo (and anyone else interested)
There's still an issue with the device/host data transfer somewhere, but I haven't found it yet.
I'll put together a summary of all of the data movement that should be happening, and maybe you can see if I'm doing something wrong in any of those places

kgasperich:
@ye-luo @anbenali 
I think the issue is somewhere in [`SoaLocalizedBasisSet<COT, ORBT>::mw_evaluateValueVPs`](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaLocalizedBasisSet.cpp#L253-L312) (or it's callees).
First I [construct the translation vectors (for phase computation) and elec-ion displacements on the host](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaLocalizedBasisSet.cpp#L268-L295).
Both of these objects are `Vector<RealType, OffloadPinnedAllocator<RealType>>`, which I assume means I can do the following:
- assign values to `Tv_list` on host
- call `updateTo()` (values are now correct on device)
- use `map(to: Tv_list_ptr[:num_elements])` in a pragma with `omp target ...` (where `auto* Tv_list_ptr = Tv_list.data()`)
  - This is the part where I think I might not be understanding something: 
  I'm assuming that the use of `OffloadPinnedAllocator` means that the host pointer `Tv_list_ptr` (and the `num_elements * sizeof(RealType)` following bytes) are already mapped to some address on the device, and that when I use `Tv_list_ptr` inside the offload region, it will be pointing to the correct data on the device (and this data should be up to date from when I called `updateTo()`).
  I rely on this in several places (for pbc translation vectors, phase factors, radial spline coeffs/derivs, and constant factors in the SoaSphericalTensor evaluation, all described below)

After constructing `TV_list` and `displ_list_tr` on the host, I call `updateTo()` to update the device data. (The Tv_list is only used for complex builds, so I only update that `#if defined(QMC_COMPLEX)`)

Then in an offload region I set all of the VP AO values to zero:
```
  // set AO data to zero on device
  auto* vp_basis_v_ptr = vp_basis_v.data();
  PRAGMA_OFFLOAD("omp target teams distribute parallel for collapse(2) map(to:vp_basis_v_ptr[:nVPs*BasisSetSize]) ")
  for (size_t i_vp = 0; i_vp < nVPs; i_vp++)
    for (size_t ib = 0; ib < BasisSetSize; ++ib)
      vp_basis_v_ptr[ib + i_vp * BasisSetSize] = 0;
```

At this point, the VP_AO values should be zero on the device, and the Tv_list and displacements should have the correct values on the device.
Then I call [`SoaAtomicBasisSet::mw_evaluateV`](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L708-L844) for each ion/center.
Here I use several arrays with `OffloadPinnedAllocator` for scratch space (from the [MW mem struct](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L879-L884))
First I get some pointers to data, then I [compute the phase contribution](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L762-L783) that comes from the translation vectors (`Tv_list` from SoaLocalizedBasisSet::mw_evaluateValueVPS, which I'm assuming are correct on the device)

Then I compute the [distances and displacements for each electron at each periodic image](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L786-L805).
This uses the member data [`periodic_image_displacements`](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L54), which is first calculated by the [LCAO builder](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/LCAOrbitalBuilder.cpp#L971-L1086), which calls the [SoaLocalizedBasisSet::setPBCParams](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaLocalizedBasisSet.cpp#L103-L113) ([xml](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/LCAOrbitalBuilder.cpp#L367), [h5](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/LCAOrbitalBuilder.cpp#L367)), and that calls the [SoaAtomicBasisSet::setPBCParams](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L112-L127), which assigns on the host and then calls `updateTo()`
```
  void setPBCParams(const TinyVector<int, 3>& pbc_images,
                    const TinyVector<double, 3> supertwist,
                    const OffloadVector& PeriodicImagePhaseFactors,
                    const OffloadArray2D& PeriodicImageDisplacements)
  {
    PBCImages                    = pbc_images;
    periodic_image_phase_factors = PeriodicImagePhaseFactors;
    periodic_image_displacements = PeriodicImageDisplacements;
    SuperTwist                   = supertwist;


    periodic_image_phase_factors.updateTo();
    periodic_image_displacements.updateTo();
  }
```

At this point, the displacement and distance arrays should be correct on the device.
Then I call the [radial](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L810) and [angular](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L816) batched evaluation functions.

These both assume that some relevant data is already on the device.
For the radial [MultiQuinticSpline1D::batched_evaluate](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h#L155-L214), I assume that the spline [coeffs and first derivatives are already correct on the device](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h#L353-L354).

For the angular [SoaSphericalTensor::batched_evaluateV](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/Numerics/SoaSphericalTensor.h#L78-L107), I assume that [NormFactor, FactorL, and FactorLM are correct on the device](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/Numerics/SoaSphericalTensor.h#L211-L214)

After the radial and angular evaluation, I loop over [elecs, basis functions, and PBC images](https://github.com/kgasperich/qmcpack/blob/8bce6a6478fd25827882be2cb3b63a3ecaa9bf7e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L826-L843) to calculate the AO values (which were set to zero before entering this function).

In the docker image for the offload test, if I print the `phase_fac_ptr[i_xyz]` in this region (in the innermost loop), it's always zero, but if I add `periodic_image_phase_factors.updateTo()` just before this offload region, then it has the correct value.

kgasperich:
Could this this have something to do with the fact that the phase factors, coeffs, normfactors, etc are all class members (and not part of a MW resource collection)?  I could see this maybe happening with the SoaAtomic data members (depending on how these things are built, and if some things only happen for the leader and not the full MW collection).
But I don't see why it would happen with the SoaSphericalTensor, because the assignment and data transfer is in the constructor, so if the thing exists then I'd expect it to have the correct data already on the device (unless there's something happening where the constructor that does the `updateTo` is only called once, and then the rest of the instances are all copies of that, so the copy constructor might only copy the host-side NormFactor data, and that new OffloadVector won't have been updated on the device yet).

kgasperich:
@ye-luo @anbenali I changed it to always map those arrays, and the offload test passes now.
I'll open an issue for adding the MW resources.

ye-luo:
Test this please

src/Numerics/SoaCartesianTensor.h
```diff
@@ -56,27 +60,61 @@ struct SoaCartesianTensor
...
+    assert(XYZ.size(1) == Npbc);
+    const size_t Nlm = XYZ.size(2);
+
+    size_t nR = nElec * Npbc; // total number of positions to evaluate
+
+    auto* xyz_ptr        = xyz.data();
+    auto* XYZ_ptr        = XYZ.data();
+    auto* NormFactor_ptr = NormFactor.data();
+
+    PRAGMA_OFFLOAD("omp target teams distribute parallel for \
```
ye-luo, line 106: Could you add a FIXME saying the "always" should be removed after making the transfer only once ahead of time?

src/Numerics/SoaSphericalTensor.h
```diff
@@ -55,21 +59,58 @@ struct SoaSphericalTensor
...
+    size_t nR = nElec * Npbc; // total number of positions to evaluate
+
+    auto* xyz_ptr        = xyz.data();
+    auto* Ylm_ptr        = Ylm.data();
+    auto* FactorLM_ptr   = FactorLM.data();
+    auto* FactorL_ptr    = FactorL.data();
+    auto* NormFactor_ptr = NormFactor.data();
+
+
+    PRAGMA_OFFLOAD("omp target teams distribute parallel for \
```
ye-luo, line 97: Could you add a FIXME saying the "always" should be removed after making the transfer only once ahead of time?

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
```diff
@@ -596,6 +597,9 @@ void LCAOrbitalSet::evaluateDetRatios(const VirtualParticleSet& VP,
   Vector<ValueType> vTemp(Temp.data(0), BasisSetSize);
   Vector<ValueType> invTemp(Temp.data(1), BasisSetSize);
 
+  if (Identity)
+    BLAS::copy(psiinv.size(), psiinv.data(), invTemp.data());
```
ye-luo, line 601: Can you use `std::copy_n` instead?

src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h
```diff
@@ -137,6 +152,68 @@ class MultiQuinticSpline1D
...
+    auto* u_ptr = u.data();
+
+    auto* coeff_ptr       = coeffs->data();
+    auto* first_deriv_ptr = first_deriv.data();
+    const size_t nCols    = coeffs->cols();
+    const size_t coefsize = coeffs->size();
+
+
+    PRAGMA_OFFLOAD("omp target teams distribute parallel for \
+                    map(always, to:coeff_ptr[:coefsize], first_deriv_ptr[:num_splines_]) \
```
ye-luo, line 185: Could you add a FIXME saying the "always" should be removed after making the transfer only once ahead of time?

src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h
```diff
@@ -665,6 +684,166 @@ struct SoaAtomicBasisSet
...
+    auto* LM_ptr        = LM.data();
+    auto* NL_ptr        = NL.data();
+    auto* psi_ptr       = psi.data();
+
+    auto* ylm_ptr = ylm_v.data();
+    auto* rnl_ptr = rnl_v.data();
+    {
+      ScopedTimer local_timer(psi_timer_);
+      PRAGMA_OFFLOAD("omp target teams distribute parallel for collapse(2) \
+          map(always, to:phase_fac_ptr[:Nxyz], LM_ptr[:BasisSetSize], NL_ptr[:BasisSetSize]) \
```
ye-luo, line 830: Could you add a FIXME saying the "always" should be removed after making the transfer only once ahead of time?

src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h
```diff
@@ -696,13 +875,13 @@ struct SoaAtomicBasisSet
...
-    OffloadArray4D ylm_vgl;     // [5][Nelec][PBC][NYlm]
-    OffloadArray4D rnl_vgl;     // [5][Nelec][PBC][NRnl]
-    OffloadArray3D ylm_v;       // [Nelec][PBC][NYlm]
-    OffloadArray3D rnl_v;       // [Nelec][PBC][NRnl]
-    OffloadMatrix dr_pbc;       // [PBC][xyz]        translation vector for each image
+    OffloadArray4D ylm_vgl; // [5][Nelec][PBC][NYlm]
+    OffloadArray4D rnl_vgl; // [5][Nelec][PBC][NRnl]
+    OffloadArray3D ylm_v;   // [Nelec][PBC][NYlm]
+    OffloadArray3D rnl_v;   // [Nelec][PBC][NRnl]
+    // OffloadArray2D dr_pbc;      // [PBC][xyz]        translation vector for each image
```
ye-luo, line 882: Remove?

src/QMCWaveFunctions/tests/CMakeLists.txt
```diff
@@ -100,6 +104,9 @@ set(TRIALWF_SRC
     test_lattice_gaussian.cpp
     test_TWFGrads.cpp)
 
+set(TMP_SRC
```
ye-luo, line 107: Please put it in the sposet unit test?

ye-luo (commented):
Probably still need to revisit data transfers but not a concern of this PR.

src/Numerics/SoaCartesianTensor.h
kgasperich: fixed

src/Numerics/SoaSphericalTensor.h
kgasperich: fixed

src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h
kgasperich: fixed

src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h
kgasperich: fixed

src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h
kgasperich: fixed

src/QMCWaveFunctions/tests/CMakeLists.txt
kgasperich: fixed

ye-luo:
Test this please

ye-luo (commented):
LGTM. Need non-ANL review/approval.

kgasperich:
@ye-luo 
https://github.com/QMCPACK/qmcpack/actions/runs/6659201270/job/18097775023#step:9:2104

Could this be caused by the fact that I'm only mapping part of the array here?
https://github.com/kgasperich/qmcpack/blob/5f4518efd2d034ad3a19630ffeb226f89abf523e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L792
and here
https://github.com/kgasperich/qmcpack/blob/5f4518efd2d034ad3a19630ffeb226f89abf523e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L772

ye-luo:
> @ye-luo https://github.com/QMCPACK/qmcpack/actions/runs/6659201270/job/18097775023#step:9:2104
> 
> Could this be caused by the fact that I'm only mapping part of the array here? https://github.com/kgasperich/qmcpack/blob/5f4518efd2d034ad3a19630ffeb226f89abf523e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L792 and here https://github.com/kgasperich/qmcpack/blob/5f4518efd2d034ad3a19630ffeb226f89abf523e/src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h#L772

I will take a look.

ye-luo:
Test this please

prckent (approved):
Once all the initial functionality is there, please remove the commented out code in a subsequent PR.

prckent:
Test this please

PR #4789, opened 2023-10-20
Self-healing overlap estimator (old/closed)

jtkrogel:
## Proposed changes

Implements the self-healing overlap estimator as detailed in https://doi.org/10.1103/PhysRevLett.104.193001.

Clarification needed on how to access certain data values for multideterminant wavefunctions.  

Apart from two data access issues, the code is functional and verified correct in VMC with and without Jastrow.

- New feature

- No

Laptop


jtkrogel:
Questions:

1) How to access the size of the multideterminant wavefunction (number of dets) at estimator initialization time?

2) How to access the current value of the first determinant in the MD expansion (access to all others is secured already)?

@ye-luo @anbenali do you know how to do this?  Comments from others welcome also.

src/QMCWaveFunctions/WaveFunctionComponent.h
```diff
@@ -482,6 +482,10 @@ class WaveFunctionComponent : public QMCTraits
   */
   virtual void evaluateDerivativesWF(ParticleSet& P, const opt_variables_type& optvars, Vector<ValueType>& dlogpsi);
 
+
+  virtual void evaluateDerivativesWF_local(Vector<ValueType>& dlogpsi);
```
ye-luo, line 486: Full documentation please.

ye-luo:
> How to access the size of the multideterminant wavefunction (number of dets) at estimator initialization time?
What is the number of dets? Confused. It can be 2 (up/down). Or you mean the number of linear expansion terms?

> How to access the current value of the first determinant in the MD expansion (access to all others is secured already)?
Did you mean the coefficient of the reference determinant pair?

jtkrogel:
I mean the number of terms in the linear expansion (including the first).

The MD is a sum of determinants: Psi(R) = sum_n c_n*Phi_n(R).  What I need is either the value Phi_0(R) or the ratio Phi_0(R)/Psi(R).

I've already obtained the values Phi_n(R)/Psi(R) as d/dc_n log(Psi) using the parameter derivative code.  However, as written in QMCPACK, this excludes the first term (n=0).

src/QMCWaveFunctions/WaveFunctionComponent.h
jtkrogel: Done.

jtkrogel:
Code is functional now.  Ready for review steps toward merging.

PDoakORNL:
Unit tests?
Add valid input section to EstimatorManagerInputTest.cpp
Valid and invalid input test.
Verification Accumulation method produces expected results from known arguments.

prckent:
I agree with Peter here. Some discussion of the testing plan is needed. Ideally there would be set of unit tests to avoid the headaches of a trickier set of deterministic tests. Ideally a unit test would be added here to get some coverage. Further improvements could be broken into subsequent PRs, but the plan should be stated.

src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.h
```diff
@@ -186,6 +187,12 @@ class MultiSlaterDetTableMethod : public WaveFunctionComponent, public Optimizab
 
   void evaluateDerivativesWF(ParticleSet& P, const opt_variables_type& optvars, Vector<ValueType>& dlogpsi) override;
 
+  void evaluateDerivativesWF_local(Vector<ValueType>& dlogpsi) override;
```
ye-luo, line 190: Probably not needed anymore.

src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.cpp
```diff
@@ -1002,6 +1002,40 @@ void MultiSlaterDetTableMethod::evaluateDerivativesWF(ParticleSet& P,
...
+  // CI only for now
+  assert(!csf_data_);
+  ratios.resize(C->size());
+  ratios[0] = psiinv;
+  for (size_t i = 1; i < C->size(); i++)
+  {
+    ValueType cdet = psiinv * detValues0[(*C2node)[det_id][i]];
+    // assume that evaluateLog has been called in opt routine before
+    for (size_t id = 0; id < Dets.size(); id++)
+      if (id != det_id)
```
ye-luo, line 1025: single out `det0` makes the code much harder to read and it is unnecessary.
```
ValueType cdet = psiinv;
for (size_t id = 0; id < Dets.size(); id++)
  cdet *= Dets[id]->getRatiosToRefDet()[(*C2node)[id][i]];
```

src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.h
```diff
@@ -186,6 +187,12 @@ class MultiSlaterDetTableMethod : public WaveFunctionComponent, public Optimizab
...
+
+  void detRatios(Vector<ValueType>& ratios) override;
```
ye-luo, line 192: Please document what this function do. Also longer and more descriptive name is preferred. This function should probably be a const member.

ye-luo (commented):
Lack of tests. No documentation. We should really avoid taking code as if it is dead on arrival.

src/Estimators/EstimatorManagerNew.cpp
```diff
@@ -101,6 +102,7 @@ EstimatorManagerNew::EstimatorManagerNew(Communicate* c,
     if (!(createEstimator<SpinDensityInput>(est_input, pset.getLattice(), pset.getSpeciesSet()) ||
           createEstimator<MomentumDistributionInput>(est_input, pset.getTotalNum(), pset.getTwist(),
                                                      pset.getLattice()) ||
+          createEstimator<SelfHealingOverlapInput>(est_input,twf) ||
```
ye-luo, line 105: @PDoakORNL This feature seems requiring MSD, how to implement a trap? I would like to avoid implementing the trap in the constructor but right before calling the constructor.

src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.h
jtkrogel: Correct.  Will remove.

src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.h
```diff
@@ -186,6 +187,12 @@ class MultiSlaterDetTableMethod : public WaveFunctionComponent, public Optimizab
...
+
+  std::vector<ValueType>& detCoefficients() override;
```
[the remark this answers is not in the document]
jtkrogel: Later implementations using control variates need access to the actual coefficient values.  At this stage, the coefficient count is a necessary by-product.

src/QMCWaveFunctions/WaveFunctionComponent.h
```diff
@@ -527,6 +536,19 @@ class WaveFunctionComponent : public QMCTraits
...
+  {
+    APP_ABORT("detRatios is not implemented. Perhaps wavefunctions is not a multideterminant?");
+  }
+
+  virtual std::vector<ValueType>& detCoefficients()
+  {
+    APP_ABORT("detCoefficients is not implemented. Perhaps wavefunctions is not a multideterminant?");
+    std::vector<ValueType>* Ctmp = new std::vector<ValueType>();
+    return *Ctmp;
+  }
```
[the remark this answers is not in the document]
jtkrogel: There is no similar functionality available, so please recommend a replacement.  All attempts to perform dynamic_casts fail at link time.

src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.cpp
jtkrogel: How is it unnecessary?  The rest of the existing multi-det code is already written this way, so it seemed like a design "feature".

jtkrogel:
A discussion on testing is warranted.  

Unlike other estimators, this one can be tested for correctness with a statistical test based on any system since the mean of the estimator in VMC is just the inputted coefficient values.  A potential wrinkle is that the statistical distribution the coefficients obey has not been characterized, so perhaps this is not the first test to make.

Due to time constraints (my code development time budget is measured in hours per quarter) I prefer adding deterministic tests over unit tests.

src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.h
jtkrogel: I will note it is already the length of similar functions nearby.  The function above it is essentially `derivativesWF` since `evaluate` is superfluous.

jtkrogel: However, state your preference on length and I will create a corresponding name.

src/Estimators/EstimatorManagerNew.cpp
jtkrogel: The estimator actually doesn't require MSD per se, as actually any wavefunction that involves a linear combination will do.  In the interest of future-proofing, I would prefer to limit how much hard coding is done to forever bind this estimator to a single, now extant class.

jtkrogel:
For interest, here is a VMC test of the estimator performed on the LiH MSD system already packaged in `qmcpack/tests`:

sh_overlap_coeffs2

Inputted coefficients are shown in blue.  In this case, the original number of determinants (1487) is reduced to 185 by a magnitude cutoff while the estimator is performing a measurement across the full basis.

jtkrogel:
I will add in-code documentation.  Documentation in the manual will be provided once the first paper is submitted.

src/QMCWaveFunctions/WaveFunctionComponent.h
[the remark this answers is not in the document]
ye-luo: Please use `findMSD()` from #4809, it does the dynamic_cast for you and the linking issue should have been resolved.

src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.h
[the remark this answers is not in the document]
ye-luo: If accessing linear coefficients is needed later, I would prefer
```
auto& getLinearExpansionCoefs() const {return *C;}
```

src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.cpp
ye-luo: Not sure which place you were referring to. In other places, det_id depends on an argument or something but is hard-coded in your case. Thus, better to write the code without singling out `det_id=0`.

src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.h
ye-luo: I saw you were calculating `d log(Psi)/ d c_i`
I would recommend `calcLinearExpansionDeriv`, for the following reasons:
1. functions should start with a verb. See developer guidelines.
2. In MSD, each linear expansion term is a multiplication of one or more determinants, most of the time two. So never call things determinants which are not determinants.

src/Estimators/EstimatorManagerNew.cpp
ye-luo: If there is a second WF with linear combination, we should introduce another accessor base class for all the WF with the same feature and dynamic_cast to this base class. Again, we still need checks before constructing the estimator object. I'd like to know how to inject such checks, current `createEstimator` doesn't facilitate that. @PDoakORNL any input, I'm thinking of not using `createEstimator` and just expand `createEstimator` calls explicitly.

PDoakORNL: Ideally you should be resolving that sort of defective in the input classes. If you can't figure out such a thing out without tearing apart the simulation time twf in estimator specific way then if makes sense for the estimators constructor to throw an exception which is going to lead to a fatal error. I don't think estimator specific error checks should be anywhere except in a particular estimator class. I don't see any issue with a constructor throwing for such an issue.

ye-luo: I usually prefer not to throw from a constructor because it may leave half baked stuff and cannot call destructor. probably it is OK here since anyway it will be a fatal exception that leads to abort.

PDoakORNL (changes requested):
Tests I think we can and should do at least a couple or three unit tests.

* Handles lack of MSD properly.
* Accumulate For twf with set coefficients accumulates the values anticipated. i.e. no off by one errors. Update the coefficients once, accumulate again is the collected data correct?
* HDF5 output correct.

No research needs to be done to do these. More man power may be needed.

src/QMCWaveFunctions/Fermion/MultiSlaterDetTableMethod.h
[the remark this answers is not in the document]
ye-luo: I have changed this as I was trying out findMSD.

src/Estimators/SelfHealingOverlap.cpp
```diff
@@ -0,0 +1,162 @@
...
+                                    const RefVector<QMCHamiltonian>& hams,
+                                    RandomBase<FullPrecRealType>& rng)
+{
+  for (int iw = 0; iw < walkers.size(); ++iw)
+  {
+    MCPWalker& walker      = walkers[iw];
+    ParticleSet& pset      = psets[iw];
+    TrialWaveFunction& psi = wfns[iw];
+    RealType weight        = walker.Weight;
+    auto& wcs              = psi.getOrbitals();
```
ye-luo, line 88: Not for this PR but we should change the name of this function. It should not be call 'orbitals.'

src/QMCWaveFunctions/WaveFunctionComponent.h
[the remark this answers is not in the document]
ye-luo: I have added the use of `findMSD` and there is no more linking issue.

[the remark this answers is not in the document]
ye-luo: Please remember to pull before continuing making changes.

src/Estimators/EstimatorManagerNew.cpp
PDoakORNL: I don't support throwing nonfatal exceptions from a constructor either but in general our convention is to not throw unless its fatal. This keeps the code in the one of the two acceptable places to have it.

jtkrogel:
Closing in favor of fresh PR

PR #4797, opened 2023-10-23
Workaround for assignment-after-reduction bug in mixed precision

markdewing:
Workaround for #4767 in mixed precision builds.

Assign the x,y and z gradient components individually.
Otherwise in mixed precision builds, where these are floats, the z component does not get copied.

As noted in the comments, the workaround of reducing the size of the `iw` index variable to 16 bits also works, but a limit of 32K walkers for a loop is uncomfortably small.

This bug should not affect complex reductions, as complex number only have two components.

- Bugfix

- No

local server

## Checklist


PDoakORNL (approved):
This seems ok to me.  We have reported this as an issue to the openmp target folks?

ye-luo:
Does this workaround help particularly mixed precision complex?

ye-luo:
1) your fix worked on complex mixed precision build with cuda 12.2
2) I tried cuda 12.3 and found it working without this PR and #4773.

ye-luo:
Test this please

ye-luo:
Will merge tomorrow.

PR #4798, opened 2023-10-24
Multi return by constructed object

correaa:
## Proposed changes

Refactor due to change in multi.
(The function seems to be dead code anyway.)

- No

Ubuntu 22.04

## Checklist


ye-luo (approved):

ye-luo:
Test this please

src/QMCTools/ppconvert/src/common/MatrixOps.h
```diff
@@ -51,7 +51,7 @@ static auto decompose(Matrix&& A, Permutation&& P, double tol = std::numeric_lim
 			);
 		}
 	}
-	return std::move(ret);
+	return A({0, N}, {0, N});
```
PDoakORNL, line 54: I don't think this is a constructor, struggling through the multi source it seems like its an operator() on something, I'm not completely sure what comes out of that call and is returned I think its view onto an altered A.  A is a variable and not a type here. (please follow naming conventions this is exactly the situation that is less clear)

At least in NLPP.cc its returned to a reference so I guess it must be a different view on the matrix A that was passed in. So A must be replaced by the LU matrix and that is pretty surprising semantics to me.  I'd expect a decompose I called  like
```
LU = decompose(A, P);
```
to leave A unaltered.

I think this function had the same issue before, I hope this function is dead code.

PDoakORNL (commented):
I feel like its quite bold to touch this tool at all.

src/QMCTools/ppconvert/src/common/MatrixOps.h
correaa: Of course, it is not a constructor; it is creating a newly constructed value. It is a (square) subarray of the original A that has been modified; most of the time, it is the whole array. Yes, we can argue forever on the interface. The interface was designed as an evolution to `decompose(A, P)` where `A` is an in-out parameter but a poor one. In perspective, the first argument, `A`, should accept only square arrays, and the function should just `return std::forward<Matrix>(A);`. But, well, it is dead code anyway. The change was made for it to compile with some new restrictions that I putting in subarrays which are non-copyable and I am possibly making them non-movable too.

PR #4807, opened 2023-10-30
1rdm with spinor support

camelto2:

## Proposed changes

This PR adds the spinor support into the 1rdm. If spinor=true in the particleset, it will generate spin moves similar to how the spatial moves are generated. 

- New feature

- No

macOS

## Checklist


PDoakORNL (changes requested):
**New code requires test support**

Many PR's are coming in with none, you should be writing tests as you develop not as an afterthought later.

jtkrogel:
#4789 will have a test added prior to merge following the discussion on that PR.

For this one, I am firmly requesting a merge of functionality first, followed by an agreed upon and fulfilled PR including the test(s) by Cody.  

The functionality here is part of an active CMS research project and it is important that this functionality makes it into develop and remains current there so that other bugs we are encountering (blocks to research) can be clearly identified in develop and fixed.

src/Estimators/OneBodyDensityMatrices.h
```diff
@@ -276,12 +300,16 @@ class OneBodyDensityMatrices : public OperatorEstBase
 
   //  basis set updates
   void updateBasis(const Position& r, ParticleSet& pset_target);
+  //  basis set updates with spin
```
PDoakORNL, line 303: use `///` or `/** ... */` otherwise method documentation won't be produced by doxygen.

src/Estimators/OneBodyDensityMatrices.cpp
```diff
@@ -354,6 +361,82 @@ inline void OneBodyDensityMatrices::generateDensitySamples(bool save,
   rhocur_ = rho;
 }
 
+inline void OneBodyDensityMatrices::generateDensitySamplesWithSpin(bool save,
```
PDoakORNL, line 364: generateDensitySamples has a unit test this can as well.

src/Estimators/OneBodyDensityMatrices.h
```diff
@@ -234,11 +241,19 @@ class OneBodyDensityMatrices : public OperatorEstBase
    *   *
    */
   void generateDensitySamples(bool save, int steps, RandomBase<FullPrecReal>& rng, ParticleSet& pset_target);
+
+  //same as above, but with spin variables included
```
PDoakORNL, line 245: '///'

PDoakORNL (changes requested):
A big part of keeping code current is being able to tell whether changes made to the code break it or not. This requires tests. All that really needs to be done here is to look at the current testing and figure out how to add cases for pset_targets that are spinors. This probably requires adding a mock particle set factory function to Particle/tests/MinimalParticlePool.h, which makes me wonder how much serious testing any of the spinor code has had.

camelto2:
> A big part of keeping code current is being able to tell whether changes made to the code break it or not. This requires tests. All that really needs to be done here is to look at the current testing and figure out how to add cases for pset_targets that are spinors. This probably requires adding a mock particle set factory function to Particle/tests/MinimalParticlePool.h, which makes me wonder how much serious testing any of the spinor code has had.

The spinor code has had plenty of testing. We didn't use minimal particle pool. Most of the tests build particle sets by hand, similar to how many tests in the code have been written in the past

PDoakORNL:
This is probably high enough level to make switch over to the mock pools worth it. Which I think I agreed to help with in the past. And probably didn't which helped us get here.  One requirement would be a very small wavefunction file.

prckent (changes requested):
The discussion here rightly points out that we need a written up policy that we can point to and apply uniformly.

There has been a lot of discussion in various parts of Slack about this.

The requirement I am going to recommend here and that we apply in the future is simply that there is at least testing of some flavor in this PR, i.e. >=1 test, that executes a significant (>50%) fraction of lines in the code. Test coverage should not decrease as a result of the PR. Many of the reasons for this have been discussed already, but importantly it would allow someone other than the original code author to make a change somewhat safely. The change might be something a simple as formatting or updating error handling. This level of testing also helps guard against memory leaks and corruption creeping in. The test could be a deterministic integration test or (preferred) a set of unit tests. As already discussed, setting up unit tests can be a pain for complex objects. It is the responsibility of the implementer to go through that pain however and not create subsequent pain for others -- we have plenty of historical lessons is this area, hence the vigorous discussion. The rest of us will be trying to minimize the pain to create unit tests by docs/tutorials/hackathons/better design. This should get easier with time since we are going to be deleting lots of legacy and refactoring further, which will disentangle more code. => So for the time being this PR needs a test added. Based on the discussion of working QMC runs, a deterministic test is straightforward and will not take much time. Subsequent PRs will need to add docs, examples etc., but I don't think this was ever in question.

camelto2:
I would like to make it clear that I absolutely understand the need for testing, and that I never intended to make a PR without a test. I only made a PR without a test because I was requested to do so. For most of my development tasks, I do start my code development with a unit test before doing any of the implementation. Sometimes, depending on what I'm doing, it is easier to just get something working and worry about a test later before trying to merge if I just trying to see how or if something would work. It may not be the best practice, but in this case that is what I did. I added the functionality quickly so that Jaron could start to play around with it without it being made a permanent feature in the code. I haven't had time to work on making the tests due to other constraints, so I was planning to just keep my branch up to date with develop and merge eventually whenever I actually got around to writing a test. However, I was asked to go ahead and make a PR now, so I did.

src/Estimators/OneBodyDensityMatrices.h
```diff
@@ -145,6 +147,11 @@ class OneBodyDensityMatrices : public OperatorEstBase
   /// current density
   Real rhocur_ = 0.0;
 
+  //spin related variables
+  Real spcur_;
+  Real dspcur_;
+  bool is_spinor_;
```
ye-luo, line 153: This can be made const.

ye-luo (commented):
Could you also address the CI failure?

jtkrogel:
Yes, I requested the PR.  Code development ideals and research needs are at odds, creating a tug-of-war over which of our own feet to shoot.  Hopefully with some future discussion we can sufficiently reduce the productivity drains on both sides of this issue, but I expect it will take some effort and creativity.

prckent:
(Going through open PRs and checking what is needed to get them merged)
It seems that the ingredients to get this merged are present but the PR has never passed in CI.

src/Estimators/OneBodyDensityMatrices.h
camelto2: done

camelto2:
I'm having trouble reproducing some of the failures I'm seeing in some of the test. One of them is bus error, another a segfault. I haven't been able to reproduce those locally so I'm not sure how to go about debugging it

src/Estimators/OneBodyDensityMatrices.h
camelto2: addressed these

camelto2:
> Have you tried to decode the error found by the asan complex build? This shows that there is a memory related error in the estimator unit test.

I think I have the fix for that now. I wasn't able to get that failure to trigger on my machines, but I think I tracked it down anyway.

There are still tests failing from "maybe unitinitialized" which I've been trying to fix. My simple fixes aren't getting rid of the warning, and it seems to be happening due to the -O2, -O3 optimization flags. When I build with -O1, I do not trigger that warning, so i'm still trying to sort that out

camelto2:
> > Have you tried to decode the error found by the asan complex build? This shows that there is a memory related error in the estimator unit test.
> 
> I think I have the fix for that now. I wasn't able to get that failure to trigger on my machines, but I think I tracked it down anyway.
> 
> There are still tests failing from "maybe unitinitialized" which I've been trying to fix. My simple fixes aren't getting rid of the warning, and it seems to be happening due to the -O2, -O3 optimization flags. When I build with -O1, I do not trigger that warning, so i'm still trying to sort that out

should all be fixed now

src/Estimators/OneBodyDensityMatrices.h
camelto2: fixed

src/Estimators/OneBodyDensityMatrices.cpp
camelto2: fixed

camelto2:
Pinging this again, I think this is now ready to go

ye-luo (approved):
LGTM. Ping @prckent @PDoakORNL

prckent (approved):
Moving in the right direction.

More generally, presumably you have also thought about the random sampling implemented here:
I do wonder if the random sampling is an unnecessary complication. Do we have data to justify? The dimensionality of the integral is not that large so random sampling won't automatically have a large benefit.

prckent:
Test this please

PDoakORNL (approved):

src/Estimators/OneBodyDensityMatrices.h
```diff
@@ -145,6 +147,11 @@ class OneBodyDensityMatrices : public OperatorEstBase
   /// current density
   Real rhocur_ = 0.0;
 
+  ///spin related variables
+  Real spcur_;
+  Real dspcur_;
```
ye-luo, line 152: FYI. The above two uninitialized variables caused failure https://github.com/QMCPACK/qmcpack/actions/runs/8022503383/job/21922019273?pr=4928
Fixed by https://github.com/QMCPACK/qmcpack/pull/4928/commits/1aa51484276f2ea8af92d4c979336fca9abf06b0
Unfortunately compilers didn't warn us in this case.

PR #4809, opened 2023-10-31
Add TrialWaveFunction::findMSD() query function

ye-luo:
## Proposed changes
The query function is introduced for refactoring better #4789.

When working on this, I noticed a linking headache caused by libqmcwfs and libqmcwfs_omptarget as two separate libraries with cyclic dependency. The solution is fairly simple. Just make qmcwfs_omptarget an object target and libqmcwfs contains all the object files of qmcwfs_omptarget. This restored the old library organization, there is only `libqmcwfs.a` at linking.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/tests/test_multi_slater_determinant.cpp
+++ b/src/QMCWaveFunctions/tests/test_multi_slater_determinant.cpp
@@ -96,6 +96,7 @@ void test_LiH_msd(const std::string& spo_xml_string,
   elec_.update();
 
   auto& twf(*twf_ptr);
+  CHECK(twf.findMSD());
   twf.setMassTerm(elec_);
   twf.evaluateLog(elec_);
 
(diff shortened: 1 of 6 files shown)


ye-luo:
Test this please

src/QMCWaveFunctions/tests/test_multi_slater_determinant.cpp
```diff
@@ -96,6 +96,7 @@ void test_LiH_msd(const std::string& spo_xml_string,
   elec_.update();
 
   auto& twf(*twf_ptr);
+  CHECK(twf.findMSD());
```
PDoakORNL, line 99: It would be nice to see that given a twf with more than one wfc that we get the MSD as expected with no troubles.  But this is better than nothing.  We have coverage of both sides of the findMSD just not over the loop. This would mostly be to cover future refactoring and possible regression.

PDoakORNL (approved):
This looks good to me.

src/QMCWaveFunctions/tests/test_multi_slater_determinant.cpp
ye-luo: A second thought. Probably it is better to return a vector of reference wrapper and let the user decide how to handle if there is more than 1. In real simulations, I cannot imaging having two fermionic components in a single TWF.

PR #4815, opened 2023-11-03
Fix builds without OpenMP

prckent:
## Proposed changes

Needed to get builds with AppleClang working. There was an unprotected omp include and calls.

- Bugfix

- No

Apple m1 laptop

## Checklist


ye-luo:
That file should not be compiled in the first place. See #4817

PR #4816, opened 2023-11-03
Backend Bug and Build Fixes for RotatedSPOs when QMC_COMPLEX=1

rcclay:

## Proposed changes
Currently, RotatedSPOs has guards to make sure it is executed only in the real build.  Attempts to run calculations when removing these guards caused build errors due to type inconsistencies in the data structure definitions and in the math kernels.  I have fixed these issues and removed the guards, which allows a lot of the math kernels and RotatedSPOs API's to be tested (and pass) in the QMC_COMPLEX build.  This is still guarded at the SPOSet builder level, so users can't hit it, but at least it will be tested.  

More specifically:
1.) Removed QMC_COMPLEX guard for RotatedSPOs, so it will always be built and executed.  
2.) Promoted all "param" types to ValueType, whereas before there was a mixture of ValueType and RealType.  
3.) Generalized exponentiate_antisym_matrix and log_antisym_matrix calls to a.) execute the correct expressions for complex, and b.) exhibit correct behavior for real and complex matrices.  
4.) Did the same for constructing the antisymmetric/antihermitian matrices.  
5.) Added additional tests for log and exp matrix functions to test the complex generalizations.  

There were also some "bug" fixes.  exponentiate_antisym_matrix had a row-major/column major issues which came out in the wash for the real build, but not for complex matrices.  

What I did NOT do:
1.) Parameter handling between real and complex.  Unit tests that loaded parameters from the h5 file had dimensionality issues, likely due to the 2x difference in numbers between real and complex.  There's a guard for this now.
2.) Fix multideterminant API's.  There's some baked in real vs. complex type issues at a deep level...  fixing that will be a future PR.
3.) Check derivatives and values of complex wave functions.  This will be done in a future PR advertising a working complex orb opt capability.  For now, I only care that the real cases pass in the complex build.  

- Bugfix
- Build related changes
- Documentation changes

- No

Intel Xeon, RHEL8
## Checklist


rcclay:
Going to regroup with GCC.  Was using intel oneAPI, and I wonder if it let some stuff slide.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -286,7 +287,7 @@ void RotatedSPOs::readVariationalParameters(hdf_archive& hin)
 
 void RotatedSPOs::buildOptVariables(const size_t nel)
 {
-#if !defined(QMC_COMPLEX)
+  //#if !defined(QMC_COMPLEX)
```
camelto2, line 290: this should just be deleted, not commented out

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -311,12 +312,12 @@ void RotatedSPOs::buildOptVariables(const size_t nel)
 
     buildOptVariables(created_m_act_rot_inds, created_full_rot_inds);
   }
-#endif
+  //#endif
```
camelto2, line 315: delete, not comment

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -311,12 +312,12 @@ void RotatedSPOs::buildOptVariables(const size_t nel)
...
 }
 
 void RotatedSPOs::buildOptVariables(const RotationIndices& rotations, const RotationIndices& full_rotations)
 {
-#if !defined(QMC_COMPLEX)
+  //#if !defined(QMC_COMPLEX)
```
camelto2, line 320: delete, not comment out

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -390,15 +391,15 @@ void RotatedSPOs::buildOptVariables(const RotationIndices& rotations, const Rota
...
   if (params_supplied)
   {
-    std::vector<RealType> param(m_act_rot_inds.size());
+    std::vector<ValueType> param(m_act_rot_inds.size());
     for (int i = 0; i < m_act_rot_inds.size(); i++)
       param[i] = myVars[i];
     apply_rotation(param, false);
   }
-#endif
+  //#endif
```
camelto2, line 399: delete, not comment

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -511,8 +512,11 @@ void RotatedSPOs::exponentiate_antisym_matrix(ValueMatrix& mat)
   {
     for (int j = i; j < n; ++j)
     {
-      mat_h[i + n * j] = std::complex<RealType>(0, -1.0 * mat[j][i]);
-      mat_h[j + n * i] = std::complex<RealType>(0, 1.0 * mat[j][i]);
+      //This two liner does several things.  The first is it unpacks row-major mat into column major format.
+      //Second, it builds the hermitian matrix -i*mat.  Third, it relies on the hermiticity of -i*mat to
+      //fill the entire -i*mat matrix (in column major form) by iterating over the upper diagonal only.
+      mat_h[i + n * j] = std::complex<RealType>(std::imag(mat[i][j]), -1.0 * std::real(mat[i][j]));
+      mat_h[j + n * i] = std::complex<RealType>(-std::imag(mat[i][j]), 1.0 * std::real(mat[i][j]));
```
camelto2, line 519: I need to think through everything going on here. It looks like there is a relative transpose happening between what you're doing here vs. previously. Is this now consistent with what we've been working out in our notes?

src/QMCWaveFunctions/RotatedSPOs.h
```diff
@@ -94,19 +104,23 @@ class RotatedSPOs : public SPOSet, public OptimizableObject
...
+  // matrices of the form A=e^K, where K is antisymmetric/antihermitian.
+  static void log_antisym_matrix(const Matrix<RealType>& mat, Matrix<RealType>& output);
+  static void log_antisym_matrix(const Matrix<std::complex<RealType>>& mat, Matrix<std::complex<RealType>>& output);
 
   //A particular SPOSet used for Orbitals
   std::unique_ptr<SPOSet> Phi;
 
   /// Set the rotation parameters (usually from input file)
-  void setRotationParameters(const std::vector<RealType>& param_list);
+  void setRotationParameters(const std::vector<ValueType>& param_list);
```
camelto2, line 123: Just leaving a note, based on our previous discussion this will likely need to end up being changed back to RealType since it interfaces with SPOSetBuilder, and SPOSetBuilder explicitly passes this in as RealType. However, this can be fixed when we fully enable complex

src/QMCWaveFunctions/RotatedSPOs.cpp
camelto2: I assume by keeping it as mat[j][i] in the arguments, it produces the wrong matrix exponentials and logs in the tests?

rcclay: Actually, lets block this for now.  As it is written now, it passes the unit tests which were compared against scipy.linalg, whereas before it didn't (matrix was conjugated when it shouldn't have been).  However, lets make sure this doesn't break the real code finite difference tests of the derivative vs. parameter variations.

rcclay: Just checked it.  All good.

src/QMCWaveFunctions/RotatedSPOs.cpp
rcclay: Done

src/QMCWaveFunctions/RotatedSPOs.cpp
rcclay: Done

src/QMCWaveFunctions/RotatedSPOs.h
rcclay: Agreed.  For now, this is coded to conform to current complex handling of optimizable types.  When we decide to do something about it, I will fix this and the parameter dependent unit tests.

src/QMCWaveFunctions/tests/test_RotatedSPOs.cpp
```diff
@@ -395,12 +397,34 @@ TEST_CASE("RotatedSPOs exponentiate matrix", "[wavefunction]")
 
   CheckMatrixResult check_matrix_result3 = checkMatrix(m3, expected_m3, true);
   CHECKED_ELSE(check_matrix_result3.result) { FAIL(check_matrix_result3.result_message); }
+#ifdef QMC_COMPLEX
+  //Going to test exponentiating a complex antihermitian matrix.
+  using cmplx_t                           = std::complex<RealType>;
+  std::vector<cmplx_t> m3_input_data_cplx = {cmplx_t(0, 0),       cmplx_t(0.3, 0.1),   cmplx_t(0.1, -0.3),
```
markdewing, line 403: It would be nice if this example (and the log) were added to `gen_matrix_ops.py`.  However, that script needs some changes to make it easier to add new examples, so maybe not for this PR.

rcclay: I actually hacked gen_matrix_ops.py to generate these reference values, but it was kinda nasty and ad hoc and I was ashamed to add it.

markdewing: That happens.  With the reorganized code in #4818, making a cleaner version should be easier.    Creating a new function `create3x3_complex_matrix` that uses numpy (instead of Sympy) should work.  (And some of output code may need updating, depending on whether you want to just print the values vs.  outputting them in a format ready to use in the unit tests)

rcclay: Oh that's a LOT better.  I can either add it after #4818 is merged prior to this merge, or I can do a follow up PR after they are both in.

markdewing: If there's no urgency to get this PR in, wait for #4818, and then update this PR.

rcclay: Added reference value generation to gen_matrix_ops.py.

src/QMCWaveFunctions/CMakeLists.txt
```diff
@@ -37,11 +37,12 @@ set(WFBASE_SRCS
...
     HarmonicOscillator/SHOSetBuilder.cpp
     ExampleHeBuilder.cpp
-    ExampleHeComponent.cpp)
+    ExampleHeComponent.cpp
+    RotatedSPOs.cpp)
 
-if(NOT QMC_COMPLEX)
-  set(WFBASE_SRCS ${WFBASE_SRCS} RotatedSPOs.cpp)
-endif(NOT QMC_COMPLEX)
+#if(NOT QMC_COMPLEX)
```
markdewing, line 43: Commented-out code should be removed.

rcclay:
After looking into the H5 parameter packing/unpacking, found a type issue I had missed.  This allows us to eliminate the QMC_COMPLEX guards on the parameter handling, so essentially the entire unit test passes now.

rcclay:
Let's make this contingent on upcoming PR to make optimizable parameters real.

src/QMCWaveFunctions/tests/CMakeLists.txt
```diff
@@ -149,11 +150,11 @@ target_include_directories(sposets_for_testing PUBLIC "${CMAKE_CURRENT_SOURCE_DI
...
-if(NOT QMC_COMPLEX)
-  if(NOT ENABLE_CUDA)
-    set(SPOSET_SRC test_RotatedSPOs.cpp ${SPOSET_SRC})
-  endif()
-endif()
+#if(NOT QMC_COMPLEX)
+#  if(NOT ENABLE_CUDA)
+#    set(SPOSET_SRC test_RotatedSPOs.cpp ${SPOSET_SRC})
+#  endif()
+#endif()
```
camelto2, line 157: delete

rcclay:
OK!  @jptowns fixed the complex VariableSet -> real VariableSet issues.  If this passes the tests, it should be good to go.

ye-luo:
Note that this fix build RotatedSPO in the complex code. At the current stage only real part of complex parameters are optimized.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -626,6 +631,69 @@ void RotatedSPOs::log_antisym_matrix(const ValueMatrix& mat, ValueMatrix& output
     }
 }
 
+void RotatedSPOs::log_antisym_matrix(const Matrix<std::complex<RealType>>& mat, Matrix<std::complex<RealType>>& output)
```
ye-luo, line 634: Could you reduce duplication?

rcclay: This function was a pain...  in contrast to exponentiate_antisym_matrix and its LAPACK::heev call, this function relies on geev, which has a different argument list if its real or complex.  I could make one function if I could use #ifdef QMC_COMPLEX or if constexpr.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -626,6 +631,69 @@ void RotatedSPOs::log_antisym_matrix(const ValueMatrix& mat, ValueMatrix& output
...
+  for (int i = 0; i < n; ++i)
+  {
+    for (int j = 0; j < n; ++j)
+    {
+      auto tmp          = (i == j) ? std::log(eval[i]) : std::complex<RealType>(0.0, 0.0);
+      mat_cd[i + j * n] = tmp;
+    }
+  }
+
+  RealType one(1.0);
```
PDoakORNL, line 684: Since RealType is real guaranteed here why not just use the literals?

Still it is less awful than
`RealType cone(1.0);`
Which combines this idiom usually only used with types that can be real or complex with a variable name that I once assumed meant `complex 1` in QMCPACK

PDoakORNL (changes requested):
In additional to only optimizing the real part, the other issue here is that you can't read in for the parameters to start with.  I kind of feel like this could have should have waited for the complex/real templating but we have some communication problems.  That said other than need to document what I think is basically a temporary hack in `setRotationalParameters` I don't see any reason this can't be merged if its a useful intermediate step.

src/QMCWaveFunctions/RotatedSPOs.cpp
rcclay: Done

rcclay:
@PDoakORNL This is meant to be an intermediate step towards supporting full complex (complex optimizable parameters) orbital optimization.  Trying to split these up into more readable chunks, but even this one is kinda beefy.

PDoakORNL (dismissed):

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -22,21 +22,23 @@ namespace qmcplusplus
...
 }
 
 RotatedSPOs::~RotatedSPOs() {}
 
 
 void RotatedSPOs::setRotationParameters(const std::vector<RealType>& param_list)
 {
-  params          = param_list;
-  params_supplied = true;
+  params_.resize(param_list.size());
```
camelto2, line 38: Just leaving a note that this will need to be converted as a proper unpacking into ValueType in the coming PRs.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -22,21 +22,23 @@ namespace qmcplusplus
...
+  std::transform(param_list.begin(), param_list.end(), params_.begin(), [](RealType val) { return ValueType(val); });
```
camelto2, line 39: Just leaving a note that this will need to be converted to a proper unpacking into ValueType in a subsequent PR. Beyond the scope of what is being done here though

prckent:
Test this please

prckent:
(May show CI failures due to problems with the CI on ornl-sulfur V100; we are debugging)

prckent:
Test this please

ye-luo (changes requested):
Please reduce the code duplication as discussed in
https://github.com/QMCPACK/qmcpack/pull/4816/files#r1395759349

rcclay:
@ye-luo I have one function for log_antisym_matrix, with complex vs. type divergences handled by #ifdef qmc_complex directives.

ye-luo:
Test this please

ye-luo (dismissed):
LGTM.

prckent:
Trouble with complex offload in CI

ye-luo:
Test this please

ye-luo (approved):

ye-luo:
> Trouble with complex offload in CI

Caused by feature not yet ready. https://github.com/QMCPACK/qmcpack/issues/4843

PR #4819, opened 2023-11-07
Convert VariableSet to RealType only

camelto2:
## Proposed changes

While working to enable complex orbital rotations, @rcclay, @jptowns and I realized there were some inconsistencies in how we deal with optimizable parameters and their types. 

The optimizer is hard coded deal with only real parameters since it takes the real part of the parameters stored in VariableSet.  Generally, to minimize a **complex analytic** function you only need to consider d/dz, but general functions that aren't complex analytic need to be formulated with d/dz and d/dz*.  For those types of functions, it is completely equivalent to work with just d/d(re(z)) and d/d(im(z)) though. The local energy and cost function that we minimize are not complex analytic, so we need to work with both d/d(re(z)) and d/d(im(z)). So, to make the optimizer properly deal with complex parameters, can keep the optimizer as is if we just treat the real and imaginary part of complex numbers as two independent real parameters. So the optimizer wouldn't need to be touched.

The interface between the optimizer and optimizable wave function components is VariableSet, and it currently stored everything as ValueType. Since the optimizer only ever needs to deal with reals (treating the real/imaguinary components as 2 reals), the VariableSet should only store reals. Currently, no existing wave function component actually supports optimizing a complex parameter, so changing VariableSet to be RealType only is not a breaking change. 

If any future wave function component wants to support optimizing a complex parameter, it needs to simply register the complex parameter as two reals into VariableSet, and then the optimizer will work as it is currently written. We are going to submit a future PR enabling this functionality for RotatedSPOs. 

This PR simply makes the change in VariableSet to remove ValueType, and change the various places in the code where it was used. The existing unit test for VariableSet is modified to remove the checks on ValueType, so these changes are already being tested with existing code. If there are any additional tests that should be added, let me know. 

- No. Note that complex MSD CI coefficients optimization is broken as it is.

macOS

## Checklist

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
@@ -271,7 +271,7 @@ void QMCCostFunctionBase::reportParametersH5()
     if (ci_size > 0)
     {
       CI_Opt = true;
-      newh5 = RootName + ".opt.h5";
+      newh5  = RootName + ".opt.h5";
       *msg_stream << "  <Ci Coeffs saved in opt_coeffs=\"" << newh5 << "\">" << std::endl;
       hdf_archive hout;
       hout.create(newh5, H5F_ACC_TRUNC);
--- a/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
+++ b/src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
@@ -251,7 +251,7 @@ void QMCCostFunctionBase::reportParametersH5()
   if (!myComm->rank())
   {
     int ci_size = 0;
-    std::vector<opt_variables_type::value_type> CIcoeff;
+    std::vector<opt_variables_type::real_type> CIcoeff;
     for (int i = 0; i < OptVariables.size(); i++)
     {
       std::array<char, 128> Coeff;
--- a/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
+++ b/src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
@@ -561,10 +561,10 @@ std::unique_ptr<MultiSlaterDetTableMethod> SlaterDetBuilder::createMSDFast(
     Optimizable = CI_Optimizable = true;
     if (csf_data_ptr)
       for (int i = 1; i < csf_data_ptr->coeffs.size(); i++)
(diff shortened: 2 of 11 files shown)


ye-luo:
This change also get the logic of mixing real/complex parameters in TWF straight. Real parameters like in Jastrow factors remain real and complex parameters are casted as two real parameters.

src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
```diff
@@ -251,7 +251,7 @@ void QMCCostFunctionBase::reportParametersH5()
   if (!myComm->rank())
   {
     int ci_size = 0;
-    std::vector<opt_variables_type::value_type> CIcoeff;
+    std::vector<opt_variables_type::real_type> CIcoeff;
```
ye-luo, line 254: This is part is not so right in math but that is the current state of the code. Optimized CI coefficients should remain in complex. So optimizing complex CI coefficients needs fix following the complexification of RotatedSPO. Could you open an issue tracking this.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -561,10 +561,10 @@ std::unique_ptr<MultiSlaterDetTableMethod> SlaterDetBuilder::createMSDFast(
     Optimizable = CI_Optimizable = true;
     if (csf_data_ptr)
       for (int i = 1; i < csf_data_ptr->coeffs.size(); i++)
-        myVars.insert(CItags[i], csf_data_ptr->coeffs[i], true, optimize::LINEAR_P);
+        myVars.insert(CItags[i], std::real(csf_data_ptr->coeffs[i]), true, optimize::LINEAR_P);
     else
       for (int i = 1; i < C.size(); i++)
-        myVars.insert(CItags[i], C[i], true, optimize::LINEAR_P);
+        myVars.insert(CItags[i], std::real(C[i]), true, optimize::LINEAR_P);
```
ye-luo, line 567: Will need a fix for the complex CI coefficients optimization later.

ye-luo (approved):

ye-luo:
Test this please

src/QMCDrivers/WFOpt/QMCCostFunctionBase.cpp
camelto2: done

ye-luo: https://github.com/QMCPACK/qmcpack/issues/4820

PR #4821, opened 2023-11-08
Fix Incorrect Expressions for Hamiltonian and Overlap Matrices with Complex Wavefunctions

rcclay:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes
For the linear method, mainline QMCPACK builds the Umrigar/Toulouse Hamiltonian and Overlap matrices from d logpsi/dc, E_L, and dE_L/dc, where derivatives are taken w.r.t. variational parameters c.  Before the code actually builds the matrices, it takes the real part of all components.  For real wave functions, this is fine, but is obviously incorrect for complex wave functions, the first smoking gun being the lack of complex conjugation in the wave function overlap matrices.

For what follows notationally, refer to:  https://pubs.aip.org/aip/jcp/article/126/8/084102/309065 

If one rederives the Umrigar/Toulouse 2007 expressions under the assumptions of:
1.) Complex wave function
2.) Real local energy
3.) Real variational parameters (enforced by PR#4819, which implies real dE_L/dc derivatives), 

We arrive at the following expressions for the Hamiltonian and Overlap matrices respectively.  
h_ij
overlap

This PR implements the above expressions by changing Dsaved type from Real to ValueType, adding in appropriate complex conjugation, and deferring taking the real part until we finally build the Hamiltonian and overlap matrices.  

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Bugfix

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
Intel Xeon, RHEL8, Gnu compilers.  
## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- No. Code added or changed in the PR has been clang-formatted
- No. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #4822, opened 2023-11-08
Use throw and add a specialization in Communicate

ye-luo:
## Proposed changes
Prepare for #4821
- New feature

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent (approved):
Consistent with the other implementations, so no objections here. Obviously the temp space memory overhead of size = reduced data is not ideal, but it we are not hitting problems, there are other more important things to work on.

PR #4824, opened 2023-11-08
Move the role of LCAO::setOrbitalSetSize to constructors

ye-luo:
## Proposed changes
On the path to eventually remove SPOSet::setOrbitalSetSize. The size should be recorded in the constructor and not be allowed to change at a later point. Simplify the initialization logic.

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/LCAO/CuspCorrectionConstruction.cpp
+++ b/src/QMCWaveFunctions/LCAO/CuspCorrectionConstruction.cpp
@@ -580,11 +580,11 @@ void applyCuspCorrection(const Matrix<CuspCorrectionParameters>& info,
 
   ScopedTimer cuspApplyTimerWrapper(cuspApplyTimer);
 
-  LCAOrbitalSet phi("phi", std::unique_ptr<LCAOrbitalSet::basis_type>(lcao.myBasisSet->makeClone()));
-  phi.setOrbitalSetSize(lcao.getOrbitalSetSize());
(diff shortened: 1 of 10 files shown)


PDoakORNL (commented):
This looks like a good step to me.
 
src/QMCWaveFunctions/LCAO/LCAOSpinorBuilder.cpp:57 behavior seems special enough to me to deserve some testing coverage.

While the LCAOSpinorBuilder in general seems to have coverage its not straightforward enough that I could find where a unit test of this behavior would go. I think most of the coverage is coming from fairly integration tests like `test_spo_collection_input_LCAO_xml.cpp` Not bad to have but doesn't make adding a test for new code paths easy.

ye-luo:
> This looks like a good step to me.
> 
> src/QMCWaveFunctions/LCAO/LCAOSpinorBuilder.cpp:57 behavior seems special enough to me to deserve some testing coverage.
> 
> While the LCAOSpinorBuilder in general seems to have coverage its not straightforward enough that I could find where a unit test of this behavior would go. I think most of the coverage is coming from fairly integration tests like `test_spo_collection_input_LCAO_xml.cpp` Not bad to have but doesn't make adding a test for new code paths easy.

I modified one of the test in `test_MO_spinor.cpp`

ye-luo:
Test this please

PDoakORNL (approved):
LGTM

PDoakORNL:
test this please

prckent (approved):
Thanks for reviewing Peter.

This strategy will remove a significant source of danger, both in the user input and internally.

ye-luo:
Test this please

PR #4825, opened 2023-11-09
Add ValueAlias for ValueType reconstruction

ye-luo:
## Proposed changes
Add a helper class for Value type reconstruction.

- New feature

- No

laptop

## Checklist

### DIFF
--- a/src/type_traits/complex_help.hpp
+++ b/src/type_traits/complex_help.hpp
@@ -29,14 +29,21 @@ using IsComplex = std::enable_if_t<IsComplex_t<T>::value, bool>;
 template<typename T>
 using IsReal = std::enable_if_t<std::is_floating_point<T>::value, bool>;
 
-template <typename T, typename = bool>
-struct RealAlias_impl {};
+template<typename T, typename = bool>
+struct RealAlias_impl
+{};
 
-template <typename T>
-struct RealAlias_impl<T, IsReal<T>> { using value_type = T; };
+template<typename T>
+struct RealAlias_impl<T, IsReal<T>>
+{
+  using value_type = T;
+};
 
-template <typename T>
-struct RealAlias_impl<T, IsComplex<T>> { using value_type = typename T::value_type; };
+template<typename T>
+struct RealAlias_impl<T, IsComplex<T>>
+{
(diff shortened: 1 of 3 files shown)


PDoakORNL (approved):
I don't see a problem with this but I guess I don't understand why its needed. 

I always have assumed that classes and functions would be templated on Value since that basically determines everything except what the full precisions is.
If a derived type is complex definite and you have an indeterminant type  I always assumed its type would be determined via
```
std::complex<RealAlias<T>> var;
// or
std::complex<FullPrecReal> var_full_prec;
```

What am I not seeing here?

ye-luo:
> I don't see a problem with this but I guess I don't understand why its needed.
> 
> I always have assumed that classes and functions would be templated on Value since that basically determines everything except what the full precisions is. If a derived type is complex definite and you have an indeterminant type I always assumed its type would be determined via
> 
> ```
> std::complex<RealAlias<T>> var;
> // or
> std::complex<FullPrecReal> var_full_prec;
> ```
> 
> What am I not seeing here?

Value is not complex definite. QTBase and QTFull relies on ifdef. To have full template, I tries not ot rely on QTFull but promote precision but to derive FullPrecValue based on Value using ValueAlias.

PDoakORNL:
This is why I thought we would need `Value` and `FullPrecValue` or `FullPrecReal` for value/precision templating. I'm unclear on how if you have only a Real or a Value type coming into a function how you can discover whether it is reduced or full precision and what the full precision is without some global type alias.

This is why `WaveFunctionTypes.hpp` looks like it does, but I'll wait and see how this is used.

ye-luo:
`WaveFunctionTypes<VALUE, VALUE_FP>` needs to consistently put value types. If one use `WaveFunctionTypes<VALUE, REAL_FP>` it breaks. With added reconstruct capability, it can be changed to `WaveFunctionTypes<VALUE, FP_PREC>` and internally derive `VALUE_FP` as `ValueAlias<FP_PREC, VALUE>`.

PDoakORNL:
I get you now.  I agree that `<VALUE, FP_PREC>` is better. And now I see what you are after here.

prckent:
Test this please

PR #4826, opened 2023-11-09
Change findMSD return type to RefVector.

ye-luo:
## Proposed changes
No more returns the first found one but all. More general.

- Bugfix

- No

epyc-server

## Checklist


PDoakORNL (commented):
functionality is equivalent but this is less clear if the only possible states for the returned value are nothing or one element.

ye-luo:
> functionality is equivalent but this is less clear if the only possible states for the returned value are nothing or one element.

Prefer the user to handle size 0,1 and stop >=2.

PDoakORNL (approved):
LGTM now

PR #4827, opened 2023-11-10
Move splinebound calculation to Numerics.

ye-luo:
## Proposed changes
Intend to use this function in other spline classes.

- No

epyc-server

## Checklist

### DIFF
--- a/src/spline2/MultiBsplineEval_helper.hpp
+++ b/src/spline2/MultiBsplineEval_helper.hpp
@@ -48,9 +48,9 @@ inline void computeLocationAndFractional(
 
   T tx, ty, tz;
 
-  qmcplusplus::getSplineBound(x * spline_m->x_grid.delta_inv, tx, ix, spline_m->x_grid.num - 1);
-  qmcplusplus::getSplineBound(y * spline_m->y_grid.delta_inv, ty, iy, spline_m->y_grid.num - 1);
-  qmcplusplus::getSplineBound(z * spline_m->z_grid.delta_inv, tz, iz, spline_m->z_grid.num - 1);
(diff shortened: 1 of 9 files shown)


prckent (changes requested):
Sensible to move to Numerics for wider use. I don't have a better more generic name for the function, although we should try to think of one.

1. Can you add some consts to this functions arguments? The argument ordering is non-standard with inputs first and last. If it is going to be used more widely making its use clearer will help.
2. Is the workaround for the Intel OneAPI compiler bug still needed?

ye-luo:
1. changed the function pattern to make output last.
2. No idea for the moment. Not affecting host but used to impact when used in offload regions.

prckent (approved):

prckent:
Test this please

src/spline2/MultiBsplineEval_helper.hpp
```diff
@@ -20,74 +20,37 @@
...
 /** define computeLocationAndFractional: common to any implementation
  * compute the location of the spline grid point and residual coordinates
  * also it precomputes auxiliary array a, b and c
  */
 template<typename T>
-inline void computeLocationAndFractional(const typename qmcplusplus::bspline_traits<T, 3>::SplineType* restrict spline_m,
-                            T x, T y, T z,
-                            int& ix, int& iy, int& iz,
-                            T a[4], T b[4], T c[4])
+inline void computeLocationAndFractional(
```
PDoakORNL, line 33: is this what clang-format produced?  If this is hand formatting please protect it with format on and off tags.

ye-luo: Done by clang-format

PR #4833, opened 2023-11-13
Add traits for STO cartesian LCAO basis type

markdewing:
Previously would silently always create STO's with spherical angular part.  The inconsistent wavefunction construction would lead to bad results for the energy (could get lower than true ground state energy during optimization.)

I don't know if this combination makes sense scientifically or computationally, but if not, there should be an error message.   Note that this combination can also be reached using STO's by specifying 'transform="yes"'. (That's what lead me to find the issue - a difference between the results of transform and no transform).

Add basic test to create the different types of LCAO basis sets.  It doesn't actually check that the results are constructed as desired. For now, this can be checked by looking at the printed output of the test.

- Bugfix

- No

desktop

## Checklist


ye-luo:
Do we have a test case of STO, Cartesian, transform="no"? If there is one, we should expand it to cover transform=yes case.

markdewing:
Are you asking if there is existing short/long test using cartesian STO?  I don't think there's any existing test with this combination.

ye-luo:
> Are you asking if there is existing short/long test using cartesian STO? I don't think there's any existing test with this combination.

Any test deterministic/short/long tests? I'm wondering if you find the issue when running something or you just noticed it as you read the code.

src/QMCWaveFunctions/tests/test_LCAOrbitalBuilder.cpp
```diff
@@ -0,0 +1,156 @@
...
+  okay = doc.parseFromString(wf_xml_gto_sph);
+  REQUIRE(okay);
+
+  LCAOrbitalBuilder lcaob_gto_sph = LCAOrbitalBuilder(elec, ions, c, doc.getRoot());
+
+
+  // Radial basis: STO    Angular part: Cartesian
+  auto wf_xml_sto_cart = R"(
+    <tmp>
+     <basisset keyword="STO" transform="no">
```
ye-luo, line 121: Currently `transform` is read from the `sposet_collection` level instead of `basisset`.

src/QMCWaveFunctions/tests/test_LCAOrbitalBuilder.cpp
```diff
@@ -0,0 +1,156 @@
...
+          </basisGroup>
+        </atomicBasisSet>
+      </basisset>
+    </tmp>
+  )";
+
+  okay = doc.parseFromString(wf_xml_sto_cart);
+  REQUIRE(okay);
+
+  LCAOrbitalBuilder lcaob_sto_cart = LCAOrbitalBuilder(elec, ions, c, doc.getRoot());
```
ye-luo, line 134: It is less clear to me if this way of writing involves both constructor and copy constructor.
Prefer
```
LCAOrbitalBuilder lcaob_sto_cart(elec, ions, c, doc.getRoot());
```

src/QMCWaveFunctions/tests/test_LCAOrbitalBuilder.cpp
```diff
@@ -0,0 +1,156 @@
...
+
```
ye-luo, line 135: To make these test solid. Please add a basisset accessor to LCAOrbitalBuilder and test the actual basisset class via dynamic_cast.

ye-luo (changes requested):
It is good to catch the missing case but unit test needs improvement to make actual checks. See my comments in source.

src/QMCWaveFunctions/tests/test_LCAOrbitalBuilder.cpp
```diff
@@ -0,0 +1,186 @@
...
+// File created by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+
+#include "catch.hpp"
+
+#include "Configuration.h"
+#include "Message/Communicate.h"
+
+#include "LCAO/LCAOrbitalBuilder.cpp"
```
ye-luo, line 18: Including a cpp file?

markdewing: D'oh!  How'd that sneak in there????

src/QMCWaveFunctions/tests/test_LCAOrbitalBuilder.cpp
ye-luo: I was wrong. `transform` is read as an `basisset` attribute.

ye-luo (approved):
LGTM

ye-luo:
Test this please

prckent (approved):
I'll note that while this improves both the design and testing situation, I think we are still short on numerical correctness unit tests in LCAO.

PR #4847, opened 2023-11-27
ompBLAS gemm

kgasperich:
## Proposed changes

added simple gemm to ompBLAS
doesn't handle conjugate transpose (trans == 'H') for complex matrices

- New feature

- No

## Checklist


src/Platforms/OMPTarget/ompBLAS.cpp
```diff
@@ -23,6 +23,141 @@ namespace qmcplusplus
...
+                         T* const C,
+                         const int ldc)
+{
+  if (transa == 'T' && transb == 'N') //A(ji) * B(jk) -> C(ik)
+  {
+    PRAGMA_OFFLOAD("omp target teams distribute collapse(2) num_teams(M * N) is_device_ptr(A, B, C)")
+    for (size_t m = 0; m < M; m++)
+      for (size_t n = 0; n < N; n++)
+      {
+        C[n * ldc + m] *= beta;
```
ye-luo, line 48: When beta is `0` the result should be `0`. However in practice, if `C[n * ldc + m]` is `NaN`, the result is still `NaN`.  So change to
```
C[n * ldc + m] = beta == T(0) ? T(0) : C[n * ldc + m] * beta;
```

src/Platforms/tests/OMPTarget/test_ompBLAS.cpp
```diff
@@ -22,6 +22,98 @@
...
+      C[j][i] = D[j][i] = T(0);
+
+  A.updateTo();
+  B.updateTo();
+  C.updateTo();
+
+  T alpha(1);
+  T beta(0);
+
+  // in Fortran, B[M][N] is viewed as B^T
```
ye-luo, line 65: // B[M][N] (in Fortran) is viewed as B^T (in C)

src/Platforms/tests/OMPTarget/test_ompBLAS.cpp
```diff
@@ -22,6 +22,98 @@
...
+  // transa/transb == 'N'/'N', the actual calculation is A[K,M] * B[N,K] = C[M,N]
+  // transa/transb == 'N'/'T', the actual calculation is A[K,M] * B[K,N] = C[M,N]
+  // transa/transb == 'T'/'N', the actual calculation is A[M,K] * B[N,K] = C[M,N]
+  // transa/transb == 'T'/'T', the actual calculation is A[M,K] * B[K,N] = C[M,N]
```
ye-luo, line 69: Should be `C[N, M]`? I think you mean A, B, C in C convention here.

src/Platforms/tests/OMPTarget/test_ompBLAS.cpp
```diff
@@ -22,6 +22,98 @@
...
+  ompBLAS::gemm(handle, transa, transb, M, N, K, alpha, A.device_data(), a0, B.device_data(), b0, beta, C.device_data(),
```
ye-luo, line 70: Should be `a1`, `b1`?

src/Platforms/tests/OMPTarget/test_ompBLAS.cpp
```diff
@@ -22,6 +22,98 @@
...
 namespace qmcplusplus
 {
+template<typename T>
+void test_gemm(const int M, const int N, const int K, const char transa, const char transb)
+{
+  const int a0 = transa == 'T' ? K : M;
+  const int a1 = transa == 'T' ? M : K;
+
+  const int b0 = transb == 'T' ? N : K;
+  const int b1 = transb == 'T' ? K : N;
```
ye-luo, line 32: I feel `a0` and `a1`, `b0` and `b1` values needs to be swapped.

prckent (commented):
1. It would be useful to know the intended use case/driver for this development.
2. Are you using any particular source code as a reference?
3. We should have very good test coverage of the mjaority, if not all of the options.

prckent:
Test this please

src/Platforms/OMPTarget/ompBLAS.cpp
kgasperich: ok, I fixed this (I suppose if beta==0, we shouldn't care about whether C is initialized to something sensible anyway)

src/Platforms/tests/OMPTarget/test_ompBLAS.cpp
kgasperich: I was trying to be consistent with the comment from [`test_gemv`](https://github.com/QMCPACK/qmcpack/blob/e95609ab8cf645c1397b442ca6e84cad9bc0dc52/src/Platforms/tests/OMPTarget/test_ompBLAS.cpp#L61-L63), which I interpret as describing a row-major matrix (B) passed to a function (gemv) that assumes the input is column-major.
>  // in Fortran, B[M][N] is viewed as B^T
  // when trans == 'T', the actual calculation is B * A[N] = C[M]
  // when trans == 'N', the actual calculation is B^T * A[M] = C[N]

So for the comments in `test_gemm`, I intended to label everything as if the matrices are assumed to be in row-major layout, even though the gemm function here treats them as column-major.

If the labeling is row-major, then `A[K,M] * B[N,K] = C[M,N]` is really `A^T . B^T = C`, even though the transA/transB args are both `N`.

kgasperich:
@prckent 
> 1. It would be useful to know the intended use case/driver for this development.

The intention here is just to have the ability to perform a gemm on a GPU (specifically for the AO->MO transformations in the LCAO code).  This is meant to be a temporary quick fix because it would take more time to add some layer of abstraction that will dispatch a gemm call to some library with more optimized BLAS routines.
>   2. Are you using any particular source code as a reference?

I didn't use any code in particular as a reference, but my intent was to implement something that would have the same behavior as described by e.g. [netlib](https://netlib.org/lapack/explore-html/dd/d09/group__gemm_ga1e899f8453bcbfde78e91a86a2dab984.html#ga1e899f8453bcbfde78e91a86a2dab984) or [oneMKL](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-fortran/2024-0/gemm-001.html#GUID-97718E5C-6E0A-44F0-B2B1-A551F0F164B2)
>   3. We should have very good test coverage of the mjaority, if not all of the options.

The included tests should cover all four cases of (transA, transB) in {`'N'`,`'T'`}^2 as well as all four combinations of single/double real/complex (following the example of ompBLAS::gemv, I've omitted the conjugate transpose (`'C'`) option for transA/transB, but if there's a use case for that I can add it).
It would probably be possible to generate some pathological test inputs that would result in false positives, but I don't think that's the case here

src/Platforms/OMPTarget/ompBLAS.cpp
```diff
@@ -23,6 +23,141 @@ namespace qmcplusplus
...
+                         const int K,
+                         const T alpha,
+                         const T* const A,
+                         const int lda,
+                         const T* const B,
+                         const int ldb,
+                         const T beta,
+                         T* const C,
+                         const int ldc)
+{
```
ye-luo, line 41: Please throw an error if trans=C.

ye-luo:
Test this please

ye-luo (approved):

PR #4848, opened 2023-11-28
Complex Orbital Optimization Implementation

jptowns:

## Proposed changes

Implements complex orbital optimization. If QMC_COMPLEX=1, then the rotation parameters can be complex valued, and will work with complex-valued wave functions.

- New feature

- No

Mac M1 laptop gcc, linux workstation gcc, Intel Xeon cluster w/ Intel classic compiler + mkl
## Checklist


prckent:
Please use a more parsable title - e.g. the release notes are written based on the PR titles.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -107,36 +114,63 @@ void RotatedSPOs::extractParamsFromAntiSymmetricMatrix(const RotationIndices& ro
...
+    delta_param[i] = ComplexType(active[loc0] - myVars[i], active[loc1] - myVars[i + N]);
+    myVars[i]      = active[loc0];
+    myVars[i + N]  = active[loc1];
+  }
+#else
+  for (int i = 0; i < N; i++)
   {
     int loc        = myVars.where(i);
-    delta_param[i] = active[loc] - myVars[i];
+    delta_param[i] = (active[loc] - myVars[i]);
```
ye-luo, line 144: Unnecessary `()`

src/QMCWaveFunctions/tests/test_RotatedSPOs.cpp
```diff
@@ -639,8 +639,12 @@ TEST_CASE("RotatedSPOs hcpBe", "[wavefunction]")
   rot_spo->resetParametersExclusive(opt_vars);
 
   using ValueType = QMCTraits::ValueType;
-  Vector<ValueType> dlogpsi(1);
-  Vector<ValueType> dhpsioverpsi(1);
+  size_t dim      = 1;
+#ifdef QMC_COMPLEX 
+  dim = 2;  // real and imag stored separately
+#endif
```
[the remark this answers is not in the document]
ye-luo: Probably you can add constexpr

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -107,36 +114,63 @@ void RotatedSPOs::extractParamsFromAntiSymmetricMatrix(const RotationIndices& ro
...
 
   std::vector<ValueType> old_param(psize);
   std::vector<ValueType> new_param(psize);
 
-  for (int i = 0; i < m_act_rot_inds_.size(); i++)
+#ifdef QMC_COMPLEX
+  for (int i = 0; i < N; i++)
+  {
+    int loc0       = myVars.where(i);
+    int loc1       = myVars.where(i + N);
```
ye-luo, line 135: I feel alternating real and imaginary parts straightforward why choosing all real + all imaginary?

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -171,11 +205,21 @@ void RotatedSPOs::writeVariationalParameters(hdf_archive& hout)
...
+      }
+#else
+    Matrix<RealType> tmp(rows, cols);
     for (size_t i = 0; i < rows; i++)
       for (size_t j = 0; j < cols; j++)
         tmp(i, j) = history_params_[i][j];
-
+#endif
     std::string rot_hist_name = std::string("rotation_history_") + SPOSet::getName();
     hout.write(tmp, rot_hist_name);
```
ye-luo, line 224: I think IO should remain in ValueType.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -35,8 +35,15 @@ RotatedSPOs::~RotatedSPOs() {}
 
 void RotatedSPOs::setRotationParameters(const std::vector<RealType>& param_list)
```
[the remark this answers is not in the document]
camelto2: In this case, the only part of the code that calls setRoatationParameters is from SPOSetBuilder, and that reads the optimization parameters into a vector of reals. rather than add additional functionality to the builder to handle real and complex types, this function should just take a list of reals as an argument. 

However, we had discussed (and subsequently forgot to change) params to use a better name like kappas or something, since those are the kappa parameters defined in the original orb opt papers. Then there wouldn't be any confusion about "params" and their types

src/QMCWaveFunctions/RotatedSPOs.cpp
camelto2: This was just a choice that was made so that the we could leave the real parts of the code as is in some places and then add the complex part after with an offset. We can change it if necessary

src/QMCWaveFunctions/RotatedSPOs.cpp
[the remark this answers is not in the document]
ye-luo: I think it is the first time we need to input complex number from xml. I feel the  caller side/builder need to handle the conversion. Not critical we can leave the code as it is now.

src/QMCWaveFunctions/RotatedSPOs.cpp
ye-luo: > we could leave the real parts of the code as is in some places

Could you point me? offset +N adds difficulty writing code with iterators. There is no performance concern.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -361,24 +421,60 @@ void RotatedSPOs::buildOptVariables(const RotationIndices& rotations, const Rota
...
+    if (params_supplied_)
+    {
+      myVars.insert(sstr.str(), std::imag(params_[i]));
+    }
+    else
+    {
+      myVars.insert(sstr.str(), 0.0);
+    }
+  }
+#endif
```
ye-luo, line 443: I feel this duplication can be much simplified if we put real and imag next to each other.

src/QMCWaveFunctions/RotatedSPOs.cpp
camelto2: Doing it this way made it so that the parameters written by both global vs. history method were consistent with the what is written to "name_value_lists" in the vp.h5 which is written by VariableSet, which is now strictly reals. Was easier to deal with as we were debugging and testing derivatives, but can change if necessary

ye-luo: `Matrix<Complex>` will be stored as (nrows, ncols, 2). it is a lot safer than (ncrows, ncols*2). Once the ordering of real/complex in myVars are adjusted, this place will converge to what I was asking for.

src/QMCWaveFunctions/RotatedSPOs.cpp
```diff
@@ -107,36 +111,45 @@ void RotatedSPOs::extractParamsFromAntiSymmetricMatrix(const RotationIndices& ro
...
   }
 
   std::vector<ValueType> old_param(psize);
   std::vector<ValueType> new_param(psize);
 
-  for (int i = 0; i < m_act_rot_inds_.size(); i++)
+  //cast ValueType to RealType for delta param
+  //allows us to work with both real and complex since
+  //active and myVars are stored as only reals
+  auto* delta_param_data_real = (RealType*)delta_param.data();
```
markdewing, line 131: There might be a better suffix than "_real"?  When I read it, my immediate thought  is that this vector is just the real part.   Maybe "_alias"?   Or put the cast and array reference on a single line to avoid the name altogether? (At the cost of complicating the left-hand side of the assignment)

Alternately, this pattern of copying data using a trick to use a single loop for both the real and complex cases is repeated several places. Would it be worthwhile to make  helper function to copy complex numbers to/from a real array?  (Maybe not for this particular loop, but the other loops where it appears)

camelto2: fixed

markdewing:
Test this please

markdewing (approved):

PR #4878, opened 2023-12-15
LCAO simple ompBLAS gemm

kgasperich:

## Proposed changes
Adding call to simple device-side gemm in LCAO routines to avoid data transfer

- Bugfix
- New feature

- Yes
- No

## Checklist


src/QMCWaveFunctions/LCAO/LCAOrbitalBuilder.cpp
```diff
@@ -572,6 +572,7 @@ std::unique_ptr<SPOSet> LCAOrbitalBuilder::createSPOSetFromXML(xmlNodePtr cur)
   }
 #endif
 
+  sposet->finalizeConstruction();
```
kgasperich, line 575: the intention here was to make sure the coefs are correct on the device when the LCAOrbitalSet is created (rather than putting C->updateTo() before each AO->MO gemm), but maybe this is not the right way to do it.  If I understand it correctly, all of the LCAOrbitalSets in the RefVectorWithLeader<SPOSet> that we use in the mw_ functions have a shared pointer to the same coefs, so it is probably better to just do the H2D update once from the leader whenever that RefVectorWithLeader<SPOSet> is created.

This is also where [some tests](https://github.com/QMCPACK/qmcpack/actions/runs/7544922523/job/20539381951?pr=4878#step:6:2032) are failing, but I haven't tracked down exactly what's going wrong there yet.

src/QMCWaveFunctions/LCAO/LCAOrbitalSetWithCorrection.h
```diff
@@ -62,6 +62,12 @@ class LCAOrbitalSetWithCorrection : public SPOSet
                             ValueMatrix& logdet,
                             GradMatrix& dlogdet,
                             ValueMatrix& d2logdet) final;
+  /** update C on device
+   */
+  void finalizeConstruction() override
```
kgasperich, line 67: not sure if this is even worth adding

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
```diff
@@ -586,10 +681,55 @@ void LCAOrbitalSet::mw_evaluateDetRatios(const RefVectorWithLeader<SPOSet>& spo_
...
+    ratios_buffer_ptr[ivp] = 0;
+    for (size_t iorb = 0; iorb < requested_orb_size; iorb++)
+      ratios_buffer_ptr[ivp] += vp_phi_v_ptr[ivp * requested_orb_size + iorb] * invRow_deviceptr_list_ptr[ivp][iorb];
+  }
+
+  size_t index = 0;
+  for (size_t iw = 0; iw < nw; iw++)
+    for (size_t iat = 0; iat < vp_list[iw].getTotalNum(); iat++)
+      ratios_list[iw][iat] = ratios_buffer[index++];
+#else
```
kgasperich, line 727: There's something going wrong here, but I haven't been able to track it down yet.
We get device pointers to the psi inv rows (`invRow_ptr_list`), one per walker.
Each walker has a variable number of VPs, so we need to do nVP[iw] dot products with the correct row of psi_inv (or a single gemv per walker).

As an initial attempt to get this working, I create another list of device pointers to the psi_inv rows (`invRow_deviceptr_list`), but the size of this list is nVPs rather than nwalkers (for each walker, I repeat the psi_inv pointer nVP[iw] times so that I can just do dot products with the same indexing into the `invRow_deviceptr_list` and the first dimensino of the splattered `vp_phi_v`).

The latest issue I'm seeing seems to be a problem with the D2H copy of `ratios_buffer` at the end of the offload region. I added some debug printing [here](https://github.com/kgasperich/qmcpack/compare/lcao-ompblas-simple..kgasperich:qmcpack:lcao-ompblas-debug?expand=1) to make sure that the address where the bad copy is happening is actually the ratios_buffer.

```
ratios_buffer_ptr:    0x5582abf8c500                                                                                                         
ratios_buffer_devptr: 0x7fd80de23800                                                                                                         
"PluginInterface" error: Faliure to copy data from device to host. Pointers: host = 0x00005582abf8c500, device = 0x00007fd80de23800, size = 56: Error in cuMemcpyDtoHAsync: an illegal memory access was encountered
```

In this test, ratios_buffer should contain 7 doubles, so `size=56` makes sense if it's in bytes (but not if it's trying to copy `56*sizeof(double)` bytes)

src/QMCWaveFunctions/LCAO/LCAOrbitalSetWithCorrection.h
```diff
@@ -62,6 +62,12 @@ class LCAOrbitalSetWithCorrection : public SPOSet
...
+  {
+    lcao.finalizeConstruction();
+  }
```
kgasperich, line 70: not sure if this is even worth adding

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.h
```diff
@@ -50,6 +50,10 @@ struct LCAOrbitalSet : public SPOSet
 
   LCAOrbitalSet(const LCAOrbitalSet& in);
 
+#if defined(ENABLE_OFFLOAD)
+  bool isOMPoffload() const override { return true; }
+#endif
+
```
kgasperich, line 56: I'm not sure if this alone (`#if defined(ENABLE_OFFLOAD)`) is sufficient for identifying when to do the `LCAOrbitalSet::mw_evaluate*` on the device for LCAOs.  I suspect I might need to add more logic there, but I was planning to take a closer look at that after the other bugs are fixed.

I think the `isOMPoffload()` here only affects whether the determinant passes host or device ptrs to the inverse psi rows, but I use the same logic in `LCAOrbitalSet::mw_evaluate*` for choosing whether to do the AO->MO gemm on host or device.

prckent:
I would still like to see the plumbing for rocblas, cublas etc. here with this as a fallback. Wouldn't the former just be faster in terms of getting to working accelerated code? If there is some quirk that makes this fallback option preferred it would be best to put the rationale the PR comments. We need to minimize the amount of "standard numeric library" code that we write ourselves, so if we (you) are writing some we should be very clear about why.

ye-luo:
@prckent agreed. For the moment, get the calculation on the GPU and sort out data movement takes precedence.

prckent:
Fair enough. Then the key thing is to make sure this has reasonable testing and that any parts of the API that are not implemented have aborts (hopefully none are needed, but we need to be pragmatic about the amount of effort.)

src/QMCWaveFunctions/LCAO/LCAOrbitalSetWithCorrection.h
ye-luo: No harm

ye-luo:
Test this please

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
```diff
@@ -484,6 +529,8 @@ void LCAOrbitalSet::mw_evaluateVGLImplGEMM(const RefVectorWithLeader<SPOSet>& sp
                  requested_orb_size);
     }
   }
+#endif
+  // phi_vgl_v correct on device if ENABLE_OFFLOAD
```
ye-luo, line 533: ??

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
```diff
@@ -161,8 +168,8 @@ void LCAOrbitalSet::evaluateValue(const ParticleSet& P, int iat, ValueVector& ps
 }
 
 /** Find a better place for other user classes, Matrix should be padded as well */
-template<typename T, unsigned D>
-inline void Product_ABt(const VectorSoaContainer<T, D>& A, const Matrix<T>& B, VectorSoaContainer<T, D>& C)
+template<typename T, unsigned D, typename Alloc>
+inline void Product_ABt(const VectorSoaContainer<T, D>& A, const Matrix<T, Alloc>& B, VectorSoaContainer<T, D>& C)
```
ye-luo, line 172: Do you remember why this is changed?

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
```diff
@@ -586,10 +701,55 @@ void LCAOrbitalSet::mw_evaluateDetRatios(const RefVectorWithLeader<SPOSet>& spo_
...
+#if defined(ENABLE_OFFLOAD)
+  const size_t nw             = vp_list.size();
+  auto& invRow_deviceptr_list = spo_leader.mw_mem_handle_.getResource().invRow_deviceptr_list;
+  auto& ratios_buffer         = spo_leader.mw_mem_handle_.getResource().ratios_buffer;
+
+  invRow_deviceptr_list.resize(nVPs);
+  ratios_buffer.resize(nVPs);
+
+  std::vector<size_t> nVP_mw(nw, 0);  // number of VPs per walker
+  std::vector<size_t> iVP0_mw(nw, 0); // first VP idx of each walker
```
ye-luo, line 713: Can we simply avoid `nVP_mw` and `iVP0_mw`? Feel the code potentially be simplified by modifying the loop body at line 721

ye-luo (commented):
~~It seems that mw_evaluateVGLandDetRatioGrads still needs offloading.~~ implemented.

ye-luo:
Test this please

src/QMCWaveFunctions/LCAO/LCAOrbitalBuilder.cpp
ye-luo: Fixed

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
ye-luo: Figured.

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
ye-luo: changed.

ye-luo:
Test this please

ye-luo:
@kgasperich please have a final pass to see if there is anything you would like to change before merging.

ye-luo:
I changed offload to a runtime option. This means all the ifdef still needs to be resolved. For now, non-offload is routed to the single walker fallback this is still not optimal but should work good enough.

ye-luo:
Test this please

src/QMCWaveFunctions/LCAO/LCAOrbitalSet.cpp
```diff
@@ -632,12 +791,69 @@ void LCAOrbitalSet::mw_evaluateVGLandDetRatioGrads(const RefVectorWithLeader<SPO
...
+  rg_buffer.resize(4, nw);
+
+  for (size_t iw = 0; iw < nw; iw++)
+    invRow_deviceptr_list[iw] = invRow_ptr_list[iw];
+
+  auto* invRow_deviceptr_list_ptr = invRow_deviceptr_list.data();
+  auto* phi_vgl_v_ptr             = phi_vgl_v.data();
+  auto* rg_buffer_ptr             = rg_buffer.data();
+  const size_t phi_vgl_stride     = nw * norb_requested;
+  std::cout << "check " << phi_vgl_stride << "  " << norb_requested << "  " << rg_buffer.size() << std::endl;
```
kgasperich, line 821: reminder to remove before merge

ye-luo: My bad...

ye-luo:
Test this please

prckent (approved):

PR #4883, opened 2023-12-22
Disable target tasks in DTs.

ye-luo:
## Proposed changes
Using async target tasks causes unintended congestion. A task launched by crowd A (thread A) may run on thread B and slows down the execution of crowd B and amplify thread imbalance. Before we have a better understanding of target tasks, disable them and rely on multi-threaded offload for performance.

- Other (please describe): Performance tuning.

- No

Polaris

## Checklist


prckent (approved):
On which architectures and with which runtimes did you observe the congestion and verify the improvement?

ye-luo:
> On which architectures and with which runtimes did you observe the congestion and verify the improvement?

I noticed slowdown when comparing llvm 17 results to llvm 15 results. In llvm 15, `nowait` is no-op. After removing `nowait`, the previous result is recovered.

PR #4901, opened 2024-01-16
[WIP] Add cusp correction based on atomic orbitals

mdewing:
This implements the cusp correction scheme of Manten and Lüchow  ( J. Chem. Phys. 115, 5362–5366 (2001) https://doi.org/10.1063/1.1394757 ).

The existing cusp correction scheme in QMCPACK is based on correcting molecular orbitals.   This does not interact well with orbital rotation (one would need to recompute the correction after every rotation).   Correcting individual atomic basis functions would be a better fit with orbital rotation.

This PR uses a script ("compute_cusp_correction.py") to compute the parameters and write them to an HDF file.  The QMCPACK code reads those parameters and evaluates corrected orbitals. 

This is a minimal implementation to prototype the idea.

The design should be improved for actual use.  The GaussianCombo class should not be modified. Most likely a wrapper class (GaussianComboWithCusp?) should be created that evaluates the cusp function for r < r_c, and calls GaussianCombo otherwise.  The one tricky part is that not all the orbitals needs to modified - only the 1s- and 2s-type orbitals. The storage of cusp parameters in the HDF file may need to change to reflect the design.

It has been tested with C_2 and cc-pv{d,t,q,5}z basis sets, and works with orbital rotation.

Testing with H2O yielded some problems:

- H - the second derivative of the cusp doesn't intersect the second derivative the GTO unless r_c is enlarged to 3.0-4.0.  The paper uses molecules containing hydrogen, but makes no mention of special handling for it.
- O - H2O showed a larger variance than expected, so I tried an O atom.  The O atom causes trouble for the optimizer - the energy and variance decrease for a few steps, then start increasing (sometimes dramatically).  The problem is more severe with J3 included, but can be reproduced with just J1 and J2.   This problem can also occur with the MO cusp, so it is not exclusive to the cusp correction scheme in this PR.  Increasing minwalkers to larger value (0.8) seems to help, but this needs more investigation.  (Note there has been no orbital rotation involved here - just optimizing the Jastrows)

  
- New feature

- No

desktop

## Checklist

### DIFF
--- a/src/Numerics/GaussianBasisSet.h
+++ b/src/Numerics/GaussianBasisSet.h
@@ -86,7 +86,7 @@ struct GaussianCombo
   };
 
   bool hasShortRangeCusp = false;
-  real_type src_rcut;
+  real_type src_rcut     = 0.0;
   real_type src_alpha;
   real_type src_a;
   real_type src_c;
@@ -98,7 +98,7 @@ struct GaussianCombo
   real_type src_d3;
   real_type src_d4;
   real_type src_d5;
-  real_type src_delta;
+  real_type src_delta = 0.0;
 
   // The short-ranged cusp is intended to be used with transform="yes" and
   // so it only implements the function value and first derivative at the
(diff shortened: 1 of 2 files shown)


prckent:
Worth noting that the cusp correction scheme needs a proper visit in the "modern era". Feedback from potential users would increase the priority.

PR #4902, opened 2024-01-17
Fix T-move in batched DMC driver.

ye-luo:
## Proposed changes

T-move info was only set to all the existing walkers (both live and dead) at the beginning of the DMC driver. If branching creates new walkers, these walkers will be running as locality approximation. The fix is to set the handler before each DMC step. A better solution will be moving such meta data out of each walker and keeping them to the driver. This will be done once I get around tho restructure T-move related codes.

- Bugfix

- No

bora

## Checklist


ye-luo:
Test this please

prckent (approved):
LGTM

Thanks for tracking this down. Obviously some refactoring needed. I suspect that this will result in reduced variance in the production runs that otherwise appear healthy. We can talk about improved checks, testing, validation once we have "solved" the current instability issues.

PR #4911, opened 2024-01-31
Fix bug in TrialWaveFunction mw_evalGrad for spinor wave functions

### LINKED ISSUE #4910: Incorrect total energy and energy components for batched VMC with drift and spinors
**Describe the bug**

**TLDR** Batched + Drift shows a bias with spinor and normal code path.  Normal code path exhibits bug when doing a completely spin polarized calculation, i.e. only up electrons. 

In the process of doing some testing of the spinor code, we were trying to set up the exact same problem between a spinor wave function and a normal calculation without spin-orbit before moving to production VMC/DMC with spin-orbit. We ran QE with normal settings as well as noncolin = .true. both using AREP potentials. This should produce equivalent DFT calculations, just one using normal orbitals and one using spinors. These agree in energy at the DFT level and they should agree as well at the QMC level. 

The first test was checking that the VMC energy agreed between spinor and non spinor without jastrows. Those total energies agreed with the batched code, so everything looked reasonable. We then moved to a wave function that had jastrows. Running the spinor and non spinor case in the batched code both with the same jastrows had errors in the total energy. That was alarming, so we did the same run with the legacy code and it turned out that the energies agreed. So it seems there is a bug in the batched drivers somewhere. Then we reran the same tests, using batched/legacy and with/without jastrows but this time running without drift in the VMC drivers. If drift is not included, the spinor calculations seem to agree with the normal non spinor calculations. 

An example is shown below just showing energies and kinetic

w/ Jastrows:
```
> 
> vmc_legacy/spinor/drift  series 0
>   LocalEnergy           =          -6.98049 +/-          0.00026
>   Kinetic               =            2.6990 +/-           0.0015
> 
> vmc_legacy/spinor/nodrift  series 0
>   LocalEnergy           =         -6.980795 +/-         0.000088
>   Kinetic               =           2.69787 +/-          0.00067
>   
>   vmc_legacy/nospinor/drift  series 0
>   LocalEnergy           =          -6.98052 +/-          0.00028
>   Kinetic               =            2.7020 +/-           0.0017
> 
> vmc_legacy/nospinor/nodrift  series 0
>   LocalEnergy           =          -6.98065 +/-          0.00027
>   Kinetic               =            2.7036 +/-           0.002
>   
>   vmc_batched/spinor/drift  series 0
>   LocalEnergy           =          -6.98252 +/-          0.00049
>   Kinetic               =            2.6403 +/-           0.0024
> 
> vmc_batched/spinor/nodrift  series 0
>   LocalEnergy           =          -6.98059 +/-          0.00087
>   Kinetic               =            2.7011 +/-           0.0028
> 
> vmc_batched/nospinor/drift  series 0
>   LocalEnergy           =          -6.98050 +/-          0.00023
>   Kinetic               =            2.7016 +/-           0.0017
> 
> vmc_batched/nospinor/nodrift  series 0
>   LocalEnergy           =         -6.980612 +/-         0.000090
>   Kinetic               =           2.69838 +/-          0.00078
>
```

So currently, **with spinors and batched and jastrows and drift** has noticeable error, particularly with the kinetic energy. If I remove the jastrows, everything actually agrees for the spinor case.  
 

 
**To Reproduce**
1. last commit: 5bf03ee0409fca60acd214db8140222c66409745
2. normal make build, just using -DQMC_COMPLEX=1
3. running qmcpack as normal, i.e. qmcpack_complex qmc.in.xml 

**Expected behavior**
Energy should agree between batched drift, batched no drift, legacy drift, legacy no drift, regardless of whether there is a jastrow or not. 

**System:**
 - my M1 mac

**Additional context**
Working on adding files for reproducibility

camelto2:

## Proposed changes
This PR fixes a bug I found when doing some testing of the spinor code using the batched drivers. Essentially, I saw errors in the energy and kinetic energy relative to legacy only if there is a jastrow present and only if drift was on. Ended up tracking it down to the fact that WaveFunctionComponents that do not use the spins do not touch the spin gradient component, which they shouldn't. However,  the problem was that in the TWF::mw_evalGrad function, it was passing in the previous wave function components TWFGrad type. Since the Jastrow mw_evalGrad wasn't setting the spin component to zero, then the grads += grads_z ends up accumulating the first slater determinants value over and over again. So for 3 WF components, the spin gradient was 3x too large. 

~~Fixed by simply moving the TWFGrad initialization into the loop.~~
set zero before calling the single walker fallback.

Closes #4910 

- Bugfix

- No

M1 mac

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/TrialWaveFunction.cpp
+++ b/src/QMCWaveFunctions/TrialWaveFunction.cpp
@@ -564,9 +564,9 @@ void TrialWaveFunction::mw_evalGrad(const RefVectorWithLeader<TrialWaveFunction>
   const int num_wfc             = wf_leader.Z.size();
   auto& wavefunction_components = wf_leader.Z;
 
+  TWFGrads<CT> grads_z(num_wf);
   for (int i = 0; i < num_wfc; i++)
   {
-    TWFGrads<CT> grads_z(num_wf);
     ScopedTimer localtimer(wf_leader.WFC_timers_[VGL_TIMER + TIMER_SKIP * i]);
     const auto wfc_list(extractWFCRefList(wf_list, i));
     wavefunction_components[i]->mw_evalGrad(wfc_list, p_list, iat, grads_z);
--- a/src/QMCWaveFunctions/TrialWaveFunction.cpp
+++ b/src/QMCWaveFunctions/TrialWaveFunction.cpp
@@ -564,9 +564,9 @@ void TrialWaveFunction::mw_evalGrad(const RefVectorWithLeader<TrialWaveFunction>
   const int num_wfc             = wf_leader.Z.size();
   auto& wavefunction_components = wf_leader.Z;
 
-  TWFGrads<CT> grads_z(num_wf);
   for (int i = 0; i < num_wfc; i++)
   {
+    TWFGrads<CT> grads_z(num_wf);
     ScopedTimer localtimer(wf_leader.WFC_timers_[VGL_TIMER + TIMER_SKIP * i]);
     const auto wfc_list(extractWFCRefList(wf_list, i));
     wavefunction_components[i]->mw_evalGrad(wfc_list, p_list, iat, grads_z);
--- a/src/QMCWaveFunctions/WaveFunctionComponent.cpp
+++ b/src/QMCWaveFunctions/WaveFunctionComponent.cpp
@@ -95,7 +95,10 @@ void WaveFunctionComponent::mw_evalGradWithSpin(const RefVectorWithLeader<WaveFu
 {
   assert(this == &wfc_list.getLeader());
   for (int iw = 0; iw < wfc_list.size(); iw++)
-    grad_now[iw] = wfc_list[iw].evalGradWithSpin(p_list[iw], iat, spingrad_now[iw]);
+  {
+    spingrad_now[iw] = 0;
+    grad_now[iw]     = wfc_list[iw].evalGradWithSpin(p_list[iw], iat, spingrad_now[iw]);
+  }
 }
 
 void WaveFunctionComponent::mw_calcRatio(const RefVectorWithLeader<WaveFunctionComponent>& wfc_list,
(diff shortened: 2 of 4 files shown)


rcclay (approved):

prckent (approved):
LGTM

prckent:
Test this please

src/QMCWaveFunctions/TrialWaveFunction.cpp
```diff
@@ -564,9 +564,9 @@ void TrialWaveFunction::mw_evalGrad(const RefVectorWithLeader<TrialWaveFunction>
   const int num_wfc             = wf_leader.Z.size();
   auto& wavefunction_components = wf_leader.Z;
 
-  TWFGrads<CT> grads_z(num_wf);
   for (int i = 0; i < num_wfc; i++)
   {
+    TWFGrads<CT> grads_z(num_wf);
     ScopedTimer localtimer(wf_leader.WFC_timers_[VGL_TIMER + TIMER_SKIP * i]);
     const auto wfc_list(extractWFCRefList(wf_list, i));
     wavefunction_components[i]->mw_evalGrad(wfc_list, p_list, iat, grads_z);
```
ye-luo, line 572: Need a bit change. mw_evalGrad is not expected to depends on previous value of grads_z. This is different from single walker API. At
https://github.com/QMCPACK/qmcpack/blob/27ff296dda38c127f120f6356c268bd5bb268e7a/src/QMCWaveFunctions/WaveFunctionComponent.cpp#L98
Set `spingrad_now[iw]` to zero before calling `evalGradWithSpin`. `grad_now[iw]` has been directly overwritten.

ye-luo (changes requested):
See my comment above.

src/QMCWaveFunctions/TrialWaveFunction.cpp
camelto2: reverted TrialWaveFunction back and changed in WaveFunctionComponent.cpp

ye-luo (approved):

ye-luo:
Test this please

PR #4918, opened 2024-02-09
Add mw_ WithSpin APIs to SlaterDet

camelto2:

## Proposed changes

While I was profiling some runs on spinor calcualtions, I noticed that a few spinor calculations were going through single walker APIs where I expected they would use mw_. Turns out I forgot to add mw_ WithSpin APIs to both SlaterDet as well as DiracDeterminant so everuthing was always falling back to single walker. They were only added to DiracDeterminantBatched, which isn't being used in CPU builds since I am not building with ENABLE_OFFLOAD. 

This PR adds the appropriate APIs to SlaterDet first. Will add something to DiracDeterminant in the future, but falling back to the single walker should be fine for now. 

Additionally, when I was looking at unit tests, I noticed there isn't actually a unit test specfically for SlaterDet, or at least I didn't find anything.  I made a simple unit test to make sure that the mw_ APIs get used correctly by creating some dummy DiracDeterminants for testing. Before adding the new mw_ WithSpin APIs this test appropriately fails. After adding the WithSpin APIs, it passes. 

- New feature

- No

m1 mac

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/SlaterDet.cpp
+++ b/src/QMCWaveFunctions/Fermion/SlaterDet.cpp
@@ -94,6 +94,17 @@ void SlaterDet::mw_ratioGradWithSpin(const RefVectorWithLeader<WaveFunctionCompo
                                      spingrad_now);
 }
 
+
+void SlaterDet::mw_evalGradWithSpin(const RefVectorWithLeader<WaveFunctionComponent>& wfc_list,
+                                    const RefVectorWithLeader<ParticleSet>& p_list,
+                                    int iat,
+                                    std::vector<GradType>& grad_now,
+                                    std::vector<ComplexType>& spingrad_now) const override
+{
+  const int det_id = getDetID(iat);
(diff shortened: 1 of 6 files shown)


ye-luo:
Test this please

ye-luo (approved):

camelto2:
some tests were failing because I moved an implementation from the .h to the .cpp and forgot to remove the override keyword. Fixed now

ye-luo:
Test this please

ye-luo (approved):

camelto2:
Failing test for CUDA offload builds were using MatrixDelayedUpdateCUDA, which has a throw because mw_evalGradWithSpin has not been implemented yet. These were only added to MatrixDelayedUpdateOMPTarget. I added an ifdef guard on the failing test to not run that one if we have ENABLE_CUDA. 

Need to remove once MatrixDelayedUpdateCUDA is updated.

ye-luo:
Could you add a github issue documenting the TODO?

ye-luo:
Test this please

ye-luo (approved):

PR #4927, opened 2024-02-15
Shorten timeout on AMD CI runs.

ye-luo:
## Proposed changes
CI tests may hang on AMD GPU runs. Add a 180sec timeout and allow repeat at most twice. Better than current 1h limit.

- Other (please describe): CI settings

- No

None

## Checklist


ye-luo:
Test this please

prckent (approved):
Good idea. We can keep an eye on this and refine further. Stopping running any nightlies on nitrogen could also fix the problem.

ye-luo:
> Good idea. We can keep an eye on this and refine further. Stopping running any nightlies on nitrogen could also fix the problem.

IMO Current hangs are not due to other runs but AMD software stack issues.

PR #4933, opened 2024-02-20
Fast, efficient implementation of SOC 

camelto2:

## Proposed changes

This PR fixes a longstanding problem with SOC. Traditionally, due to the nature of the TrialWaveFunction which doesn't really exploit the fact that we only really ever use Jastrows * Determinants (or multidets), I had to do the spin integration numerically. This was quite inefficient, since we were essentially doing 8x the work since the default number of spin quadrature points was 8. 

The spin integration can be done exactly if you can exploit the fact that the wavefunction is a exp(J) * Det(spinors), which can be further simplified into the ratio jastrows * inverse transpose slater matrix * spinors.  Since the spinors are phi_up * exp(is) + phi_dn * exp(-is) you can pull the spin integration inside and do those terms exactly. That removes the need for the spin quadrature all together. 

This is done by utilizing the TWFFastForceWrapper which gives convenient access to the appropriate quantities. A simple test problem with 16 walkers, 1000 blocks, 1 thread on my laptop shows the expected speedup. 

Traditional slow evaluation: 103.8466 s
New implementation: 12.3576 s
speedup: 8.4

Right now, this only works with legacy I think. Need to look at mw_ APIs in the future. Had to add some new features to the TWFFastDerivWrapper and the SpinorSet, but everything is unit tested. 

- New feature

- No

M1 mac

## Checklist


ye-luo:
Is it possible to move this optimization to NLPP batched code path?

src/QMCHamiltonians/SOECPComponent.cpp
```diff
@@ -192,6 +211,64 @@ SOECPComponent::RealType SOECPComponent::calculateProjector(RealType r, const Po
...
+  //calculate radial potentials
+  for (int ip = 0; ip < nchannel_; ip++)
+    vrad_[ip] = sopp_m_[ip]->splint(r);
+
+  for (int iq = 0; iq < nknot_; iq++)
+  {
+    W.makeMove(iel, deltaV_[iq]);
+    psi.getRowMSpinDecomposed(W, iel, up_row, dn_row);
+    RealType jratio = psi.evaluateJastrowRatio(W, iel);
+    W.rejectMove(iel);
```
ye-luo, line 248: To me, any NLPP code paths that require accept and reject are big no no. It can always be reformulated without need of accept/reject.

camelto2: Yea I agree, but I wanted to get it working first.

camelto2:
> Is it possible to move this optimization to NLPP batched code path where NLPP ratios of dets (without Jastrow) are batched? I mean specialize
> 
> https://github.com/QMCPACK/qmcpack/blob/a82117fe3bc030e90ea564f2fae407c76e0ef3b9/src/QMCWaveFunctions/SPOSet.h#L205

evaluateDetRatios is used to give you psi(q)/psi(r). But I need it in a slightly different form in order to remove the spin integration and get what I want. Will send you some notes to take a look

src/QMCHamiltonians/SOECPComponent.cpp
camelto2: Actually, on second thought we aren't doing anything costly like updating the inverse or anything like that by accepting/rejecting moves. the TWFwrapper being used here just grabs the jastrow ratio at each quadrature point and grabs the SPOSet values at the quadrature points. So I'm not sure there is much benefit from moving to using VirtualParticles

rcclay (approved):
LGTM.  This is a very straightforward generalization of what was already done in NonLocalECPotential for the fast forces.  There are probably opportunities for optimization (better use of batching, multiwalker interfaces, etc.), but those should probably be left for a future, more targeted PR.

prckent:
Test this please

prckent:
FYI, the more modern compilers are complaining e.g.
```
src/QMCHamiltonians/tests/test_SOECPotential.cpp:258:54: error: non-const lvalue reference to type 'Real' (aka 'float') cannot bind to a value of unrelated type 'Return_t' (aka 'double')
  testing::TestSOECPotential::evalFast(so_ecp, elec, value)***
```

src/QMCHamiltonians/ECPotentialBuilder.cpp
```diff
@@ -73,6 +74,7 @@ bool ECPotentialBuilder::put(xmlNodePtr cur)
   pAttrib.add(pbc, "pbc", {"yes", "no"});
   pAttrib.add(forces, "forces", {"no", "yes"});
   pAttrib.add(physicalSO, "physicalSO", {"yes", "no"});
+  pAttrib.add(fast_so, "fastSO", {"yes", "no"});
```
ye-luo, line 77: "fastSO" is hard to understand what is fast. Could you find a more understandable name?

camelto2:
> FYI, the more modern compilers are complaining e.g.
> 
> ```
> src/QMCHamiltonians/tests/test_SOECPotential.cpp:258:54: error: non-const lvalue reference to type 'Real' (aka 'float') cannot bind to a value of unrelated type 'Return_t' (aka 'double')
>   testing::TestSOECPotential::evalFast(so_ecp, elec, value)***
> ```

This should be fixed now. There was a type mismatch in my unit test

src/QMCHamiltonians/ECPotentialBuilder.cpp
camelto2: Switched to spin_integrator. Can choose between "exact" and "simpson" which will use the old simpsos rule to integrate

camelto2:
Not sure I understand why linux (Clang14-NoMPI-UBSan-Real) is failing. Looks like test_parser, which is something I didn't touch

rcclay (approved):

ye-luo:
Test this please

ye-luo (changes requested):
Temporary blocking merge. Will review this PR in 2 days.

rcclay:
What's the status of this?

prckent:
Yes, this is a huge improvement and good to get in. Need to identify what is need for batched code execution in followup work.

camelto2:
> A huge improvement and good to get in. Need to identify what is needed for batched code execution in followup work.

I have been working on the batched implementation. Should have it submitted in the next few days

prckent:
(Hadn't noticed the "Unchanged files with check annotations" beta feature before. Would be good to work the matrix of options in subsequent PRs to get coverage on these features since we historically have trouble with this sort of thing.)

ye-luo:
> (Hadn't noticed the "Unchanged files with check annotations" beta feature before. Would be good to work the matrix of options in subsequent PRs to get coverage on these features since we historically have trouble with this sort of thing.)

Where is this feature?

src/QMCHamiltonians/SOECPotential.cpp
```diff
@@ -101,6 +108,81 @@ void SOECPotential::evaluateImpl(ParticleSet& P, bool keep_grid)
...
+  if (!keep_grid)
+    for (int ipp = 0; ipp < ppset_.size(); ipp++)
+      if (ppset_[ipp])
+        ppset_[ipp]->rotateQuadratureGrid(generateRandomRotationMatrix(*my_rng_));
+
+  std::vector<ValueMatrix> mats_minv;
+  std::vector<ValueMatrix> mats_b;
+  std::vector<ValueMatrix> mats_b_gs;
+  std::vector<ValueMatrix> mats_m;
+  std::vector<ValueMatrix> mats_m_gs;
```
ye-luo, line 123: A lot of scratch space needed in this formulation.

ye-luo (commented):
The saving comes from the removal of numerical integration over spinors. Excellent!
However, replying on TWFFastDerivWrapper requires SO being aware of the fine structure of TWF in the SO class. It seems cumbersome to me. Also it doesn't work with multdets and need a separate code path.
I experimented not relying on TWFFastDerivWrapper
https://github.com/camelto2/qmcpack/compare/fast_soc...ye-luo:qmcpack:soc-simplified?expand=1
which allows further expansion to multidet without loss of generality.
I will work with @camelto2 offline to sort things out before finalizing this PR.

prckent:
> Where is this feature?
Click your "More static member functions." commit and scroll down.

rcclay:
> The saving comes from the removal of numerical integration over spinors. Excellent! However, replying on TWFFastDerivWrapper requires SO being aware of the fine structure of TWF in the SO class. It seems cumbersome to me. Also it doesn't work with multdets and need a separate code path. I experimented not relying on TWFFastDerivWrapper https://github.com/camelto2/qmcpack/compare/fast_soc...ye-luo:qmcpack:soc-simplified?expand=1 which allows further expansion to multidet without loss of generality. I will work with @camelto2 offline to sort things out before finalizing this PR.

If speeding up the SOECP evaluation was all we needed or intended to do with this, I agree 100% with Ye's proposed solution.  However, next step for this and NonLocalECP is to speed up the derivative evaluation w.r.t. orbital rotation parameters--which is where doing it this way pulls ahead of the proposed alternative method.

ye-luo:
> > The saving comes from the removal of numerical integration over spinors. Excellent! However, replying on TWFFastDerivWrapper requires SO being aware of the fine structure of TWF in the SO class. It seems cumbersome to me. Also it doesn't work with multdets and need a separate code path. I experimented not relying on TWFFastDerivWrapper https://github.com/camelto2/qmcpack/compare/fast_soc...ye-luo:qmcpack:soc-simplified?expand=1 which allows further expansion to multidet without loss of generality. I will work with @camelto2 offline to sort things out before finalizing this PR.
> 
> If speeding up the SOECP evaluation was all we needed or intended to do with this, I agree 100% with Ye's proposed solution. However, next step for this and NonLocalECP is to speed up the derivative evaluation w.r.t. orbital rotation parameters--which is where doing it this way pulls ahead of the proposed alternative method.

Parameter derivative calculation and SO evaluation (part of Hamitonian) are in separate driver stages and code paths. 
They don't necessarily need to overlap in terms of algorithms. I do feel calculating parameter derivatives have a different pattern and probably need this formalism but I don't think much code can be reused from this PR. We will see. For SO evaluation, we do want an implementation that is portable to GPU otherwise users got a huge performance hit when turning on SO.

rcclay:
@ye-luo Fair point.  In that case, I don't see any problems with the solution you've laid out.

prckent:
Just checking on the status of this. I assume the rework is being done with the limited time available. Any gotchas?

camelto2:
Just no time. I reworked it using Ye's approach on my end, and got the unit test passing for the SOECPotential. Just need to finish adapting the unit test for the changes to the SpinorSet. Will hopefully get to it today

camelto2:
Sorry for taking so long, this should be up to date with Ye's recommendations

PDoakORNL:
There is a great deal of per walker data stashed in SpinorSet and the force wrapper. In fact none of this really looks like it took into account the batched design. These should really have been crowd scope memory resources i.e setup through resource sets (which will also allow use to track memory use much more easily). Evaluations should be over these whole blocks and not broken up by walker unless necessary. Certainly the end goal should be a single kernel launch to deal with the force calculations over a block of walkers. 

The way this is setup right now assuming a Wrapper object per walker and a SpinsorSet object per walker than a bunch of workspace or additional state can be stashed in really isn't great. extracting pointers from each walker and its inheritance pile is not a good design and was just a hack to avoid refactoring too much at once.

prckent:
Thanks Cody. I am hopeful that this gets merged after the CI passes. The real build needs fixing and there are sanitizer failures.

Peter: I think we should see where we end up with e.g. real world memory usage and performance for observables before tackling design issues here.

src/QMCHamiltonians/ECPotentialBuilder.cpp
```diff
@@ -166,10 +168,16 @@ bool ECPotentialBuilder::put(xmlNodePtr cur)
     }
     app_log() << "\n  Using SOECP potential \n"
               << "    Maximum grid on a sphere for SOECPotential: " << nknot_max << std::endl;
-    app_log() << "    Maximum grid for Simpson's rule for spin integral: " << sknot_max << std::endl;
+    if (spin_integrator == "exact")
+    {
+      app_log() << "    Using fast SOECP evaluation. Spin integration is exact" << std::endl;
+      apot->setFastEvaluation(true);
```
ye-luo, line 174: Do you need fast and slow code paths both active in a single object?
If not, Can this option be passed via constructor?

src/QMCHamiltonians/SOECPComponent.h
```diff
@@ -116,6 +118,7 @@ class SOECPComponent : public QMCTraits
 
   RealType calculateProjector(RealType r, const PosType& dr, RealType sold);
 
+  RealType evaluateOneFast(ParticleSet& W, const int iat, const TrialWaveFunction& psi, const int iel, const RealType r, const PosType& dr);
```
ye-luo, line 121: Could you please rename it? "Fast" is not really helpful. Please also add documentation with a brief description of the algorithm.

src/QMCHamiltonians/SOECPotential.h
```diff
@@ -79,6 +79,10 @@ class SOECPotential : public OperatorBase
   //return a shared resource to a collection
   void releaseResource(ResourceCollection& collection, const RefVectorWithLeader<OperatorBase>& o_list) const override;
 
+  //triggers using fast evaluation using TWFFastDerivWrapper for normal evalualations and derivatives.
+  void setFastEvaluation(bool fast) { use_fast_evaluation_ = fast; }
+  TWFFastDerivWrapper psi_wrapper_;
```
ye-luo, line 84: psi_wrapper_ is no longer used right? Please remove.

src/QMCHamiltonians/SOECPotential.h
```diff
@@ -107,6 +111,9 @@ class SOECPotential : public OperatorBase
   //multi walker resource
   ResourceHandle<SOECPotentialMultiWalkerResource> mw_res_handle_;
 
+  //flag to use fast evaluation
+  bool use_fast_evaluation_;
```
ye-luo, line 115: If we set its value via constructor, please add const.

src/QMCWaveFunctions/SPOSet.h
```diff
@@ -207,6 +207,12 @@ class SPOSet : public QMCTraits
                                  const ValueVector& psiinv,
                                  std::vector<ValueType>& ratios);
 
+  virtual void evaluateDetSpinorRatios(const VirtualParticleSet& VP,
```
ye-luo, line 210: Please add a brief documentation.

src/QMCWaveFunctions/tests/test_einset_spinor.cpp
```diff
@@ -11,6 +11,7 @@
 //////////////////////////////////////////////////////////////////////////////////////
 
 
+#include "QMCTools/QMCFiniteSize/FSUtilities.h"
```
ye-luo, line 14: QMCWaveFunctions is at a lower level than QMCTools. Cannot have this include.

src/QMCWaveFunctions/SpinorSet.h
```diff
@@ -15,6 +15,7 @@
 
 #include "QMCWaveFunctions/SPOSet.h"
 #include <ResourceHandle.h>
+#include <sys/wait.h>
```
ye-luo, line 18: Should not include this.

src/QMCWaveFunctions/SPOSet.cpp
```diff
@@ -57,6 +57,15 @@ void SPOSet::evaluateDetRatios(const VirtualParticleSet& VP,
   }
 }
 
+void SPOSet::evaluateDetSpinorRatios(const VirtualParticleSet& VP,
+                             ValueVector& psi,
+                             const std::pair<ValueVector, ValueVector>& spinor_multiplier,
+                             const ValueVector& invrow,
+                             std::vector<ValueType>& ratios)
+{
+  throw std::runtime_error("Need specialization of SPOSet::evaluateDetSpinorRatios");
```
ye-luo, line 66: Please expand the error message with ClassName.

src/QMCHamiltonians/SOECPotential.cpp
```diff
@@ -15,6 +15,7 @@
 #include "NLPPJob.h"
 #include <ResourceCollection.h>
 #include <optional>
+#include "type_traits/ConvertToReal.h"
```
ye-luo, line 18: Doesn't seem needed anymore.

src/QMCWaveFunctions/tests/test_einset_spinor.cpp
camelto2: Don't know how that ended up there

src/QMCWaveFunctions/SpinorSet.h
camelto2: Not sure how that got added either. Thanks for catching this. I think my neovim setup may have added this when I accidentally typed something and triggered some snippet to be added. Sometimes this is useful, and sometimes it isn't. I'll be more careful, and try to make sure my setup isn't stuff like this.

src/QMCHamiltonians/SOECPotential.cpp
camelto2: fixed

src/QMCWaveFunctions/SPOSet.cpp
camelto2: fixed

src/QMCWaveFunctions/SpinorSet.h
camelto2: fixed

src/QMCWaveFunctions/tests/test_einset_spinor.cpp
camelto2: fixed

src/QMCWaveFunctions/SPOSet.h
camelto2: done

src/QMCHamiltonians/SOECPotential.h
camelto2: fixed

src/QMCHamiltonians/SOECPComponent.h
camelto2: done

src/QMCHamiltonians/ECPotentialBuilder.cpp
camelto2: will fix this next, want to make sure tests pass first.

src/QMCHamiltonians/SOECPotential.h
camelto2: see above

ye-luo:
Test this please

src/QMCHamiltonians/ECPotentialBuilder.cpp
ye-luo: I fixed the complex build. When you remove this setting, please make sure
https://github.com/QMCPACK/qmcpack/blob/592182424c27eda9caf049865da84aca11505607/src/QMCHamiltonians/SOECPComponent.cpp#L67
being reactive to the fast algorithm selection and set `total_knots_` properly.

ye-luo:
Test this please

src/QMCHamiltonians/ECPotentialBuilder.cpp
camelto2: This looks to be taken care of now from what I can tell. Should fix the failing test as well

ye-luo (approved):

prckent (approved):

PR #4939, opened 2024-03-07
Fix hybridrep complex

ye-luo:
## Proposed changes
Noticed a missing safety check in the complex code path of hybridrep.

- Bugfix

- No

Polaris

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/HybridRepReal.h
+++ b/src/QMCWaveFunctions/BsplineFactory/HybridRepReal.h
@@ -128,7 +128,7 @@ public:
                          const ValueVector& psiinv,
                          std::vector<ValueType>& ratios) override
   {
-    if (VP.isOnSphere() && HYBRIDBASE::is_batched_safe(VP))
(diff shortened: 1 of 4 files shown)


prckent:
Test this please

prckent (commented):
Surely a fix to both complex and real? Or is the real path protected elsewhere? If so, where/how?

src/QMCWaveFunctions/BsplineFactory/HybridRepReal.h
```diff
@@ -128,7 +128,7 @@ class HybridRepReal : public SPLINEBASE, private HybridRepCenterOrbitals<typenam
                          const ValueVector& psiinv,
                          std::vector<ValueType>& ratios) override
   {
-    if (VP.isOnSphere() && HYBRIDBASE::is_batched_safe(VP))
```
ye-luo, line 131: The real case has the safety check already.

prckent (approved):

PR #4940, opened 2024-03-13
[WIP] dump walkers periodically for post-processing

Paul-St-Young:
## Proposed changes

Repurpose unused input "record_configs" to set frequency of dumping walkers for post-processing.

*work in progress*: I'm stuck at step 1 "reuse config.h5 instead of deleting and re-creating it at every checkpoint".
serial execution runs through, but an mpi execution would get stuck. Any suggestions?

Example input of a simple harmonic oscillator [tb38_msho.zip](https://github.com/QMCPACK/qmcpack/files/14594402/tb38_msho.zip)

- New feature

- No

Intel workstation

## Checklist

### DIFF
--- a/src/Particle/HDFWalkerOutput.cpp
+++ b/src/Particle/HDFWalkerOutput.cpp
@@ -79,7 +79,7 @@ HDFWalkerOutput::~HDFWalkerOutput() = default;
  *  - walker_partition (int array)
  *  - walkers (nw,np,3)
  */
-bool HDFWalkerOutput::dump(const WalkerConfigurations& W, int nblock)
+bool HDFWalkerOutput::dump(const WalkerConfigurations& W, int nblock, const bool identify_block)
 {
   std::filesystem::path FileName = myComm->getName();
   FileName.concat(hdf::config_ext);
@@ -93,13 +93,16 @@ bool HDFWalkerOutput::dump(const WalkerConfigurations& W, int nblock)
 
   //try to use collective
   hdf_archive dump_file(myComm, true);
-  dump_file.create(FileName);
+  bool exists = dump_file.open(FileName);
+  if (!exists) // create new config.h5
+    dump_file.create(FileName);
+
   HDFVersion cur_version;
   dump_file.write(cur_version.version, hdf::version);
   dump_file.push(hdf::main_state);
   dump_file.write(nblock, "block");
 
-  write_configuration(W, dump_file, nblock);
+  write_configuration(W, dump_file, nblock, identify_block);
   dump_file.close();
 
   currentConfigNumber++;
@@ -107,8 +110,18 @@ bool HDFWalkerOutput::dump(const WalkerConfigurations& W, int nblock)
   return true;
 }
 
-void HDFWalkerOutput::write_configuration(const WalkerConfigurations& W, hdf_archive& hout, int nblock)
+void HDFWalkerOutput::write_configuration(const WalkerConfigurations& W, hdf_archive& hout, int nblock, const bool identify_block)
 {
+  if (identify_block)
+  { // change h5 slab name to record more than one block
+  } else { // remove previous checkpoint
+    std::vector<std::string> names = {"block", hdf::num_walkers, "walker_partition", hdf::walkers};
+    for (auto aname : names)
+    {
+      if (hout.is_dataset(aname)) hout.unlink(aname);
+    }
+  }
+
   const int wb = OHMMS_DIM * number_of_particles_;
   if (nblock > block)
   {
--- a/src/Particle/HDFWalkerOutput.h
+++ b/src/Particle/HDFWalkerOutput.h
@@ -52,7 +52,7 @@ public:
   /** dump configurations
    * @param w walkers
    */
-  bool dump(const WalkerConfigurations& w, int block);
+  bool dump(const WalkerConfigurations& w, int block, const bool identify_block=false);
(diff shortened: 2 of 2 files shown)


prckent:
Hi Paul. Thanks for adding/restoring this ghost feature. As we have discussed this functionality is useful for a variety of uses, and clearly the functionality could be evolved further.

For now, two questions:

1. How much testing have you done that the energies and weights are correct and that, when averaged, they match what is in the scalar files?
2. Does this work with vmc_batched?

src/Particle/HDFWalkerOutput.h
```diff
@@ -52,7 +52,7 @@ class HDFWalkerOutput
   /** dump configurations
    * @param w walkers
    */
-  bool dump(const WalkerConfigurations& w, int block);
+  bool dump(const WalkerConfigurations& w, int block, const bool identify_block=false);
```
ye-luo, line 55: Please document what identify_block=true do.

src/Particle/HDFWalkerOutput.cpp
```diff
@@ -93,22 +93,41 @@ bool HDFWalkerOutput::dump(const WalkerConfigurations& W, int nblock)
...
+    std::stringstream block_str;
+    block_str << nblock;
+    partition_name += block_str.str();
+    dataset_name += block_str.str();
+    weights_name += block_str.str();
+  } else { // remove previous checkpoint
+    std::vector<std::string> names = {"block", hdf::num_walkers, partition_name, dataset_name, weights_name};
+    for (auto aname : names)
+      if (hout.is_dataset(aname)) hout.unlink(aname);
+  }
```
ye-luo, line 129: could you mentioned the data set name used for recording instead of checkpointing.

src/Particle/HDFWalkerOutput.h
```diff
@@ -62,7 +62,7 @@ class HDFWalkerOutput
   std::array<BufferType, 2> RemoteData;
   std::array<std::vector<QMCTraits::FullPrecRealType>, 2> RemoteDataW;
   int block;
-  void write_configuration(const WalkerConfigurations& W, hdf_archive& hout, int block);
+  void write_configuration(const WalkerConfigurations& W, hdf_archive& hout, int block, const bool identify_block);
```
ye-luo, line 65: Please document what identify_block=true do.

ye-luo (commented):
I only did an initial pass. Didn't fully understand the change at the moment. Requested more documentation.

ye-luo:
I also noticed that the added change only affect legacy driving. For my curiosity, what is missing for you to adopt the batched driver?

PDoakORNL:
With regards to batched I want to remove WalkerConfiguration as an object and heavily refactor eliminate the "Walker-->Particle" objects. The Walker object and some of its data possessions which are stuffed into particle set  is stuffed full of values that should be contiguous vectors in the crowds or even at rank scope but are instead here in an AOS with respect to walkers and mapped into a byte buffer. As it stands all this "dumping walkers" stuff is predicated on that. Pointer arithmetic is done etc.  The less dependency there is is on this the better. I think we would be better off writing an "estimator" that wrote the per walker values for data out.  This exposes legacy walker implementation details that we don't want to carry forward either, to the extent that some still exist that is because refactoring ParticleSet has been blocked for several years now.

If you want to dump walkers it should be a crowd or population level operation so it can be synchronized properly and done efficiently. If you want to use parallel hdf5 the appropriate level would be at MCPopulation.  Ideally it would be done from a separate io thread while real work is being done by the main crowd threads.

Paul-St-Young:
@prckent  apologies for the delayed response. To answer your questions:
> How much testing have you done that the energies and weights are correct and that, when averaged, they match what is in the scalar files?
I did not test the energies and weights. The same `dump` function used for a normal checkpoint is called, so I just assumed it would be correct.
> Does this work with vmc_batched?
No, I did not add this feature to the batched drivers.

The current implementation breaks down in an MPI run. I think this is due to problems with parallel hdf5 file writing.
@ye-luo I added some more details of this feature in the PR description. Does this clear up the meaning of `identify_block`?
I will also add some comments in the source code.

@ye-luo @PDoakORNL  I would love some help on getting this feature to work.
It's not clear to me exactly what needs to be done to make parallel HDF5 writing correct.

src/Particle/HDFWalkerOutput.cpp
Paul-St-Young: Dataset name for recording is `walkers{block_idx}`, whereas the name for checkpoint is simply `walkers`.

prckent:
Hi Paul - Is #5019 working for you? If so, we can close this PR.

ye-luo:
@prckent please keep this PR alive. it is needed for different uses.

Paul-St-Young:
@prckent  yes, #5019  does what I need.
It outputs a bit too much information, but it seems straight-forward to modify so I can go from there.
I don't mind if you decide to close this PR.

@ye-luo  would it be easier to add a "replay" flag to the new WalkerLog class?
I think it just needs to output walker coordinates and weights at every step without all the other debugging contents.

ye-luo:
WalkerLog has very different design logics and use cases. For example, it writes one file per MPI rank instead of using scalable I/O. Data are not organized in steps and thus requires extra work to recover a true reply.

ye-luo:
I think a more refined version can be implemented when properly implementing a reply feature or correlated sampling.

PR #4945, opened 2024-03-26
Create routines to solve generalized eigenvalue problem

markdewing:
This is the first in a series of changes to make a uniform interface to get the target eigenvector.  The goal is to make it easier to interface to alternate eigensolvers, such as MAGMA or ARPACK.

---

Extract this code from getLowestEigenvector.  Eventually getLowestEigenvector will call solveGeneralizedEigenvalues.

Create two variants. One uses the lapack routine for the generalized eigenvalue problem (ggev). The other computes S^1 * H, and solves the regular eigenvalue problem.

Add some tests for the eigenvalue solver.

---
Future changes:
* call solveGeneralizedEigenvalues from getLowestEigenvector (and the versions suffixed with _Inv).  The plan is to add a suffix to the function names to differentiate between different underlying solvers.
* remove matrix inversion from `one_shift_run`, since it will be present in solveGeneralizedEigenvalues_Inv

- No

Desktop

## Checklist

- Yes/. Code added or changed in the PR has been clang-formatted


markdewing:
The functions could go here or in Numerics.  I suspect there isn't going to be re-use of the eigenvalue routines elsewhere in the code, so there isn't a strong reason to put them in Numerics.

I don't think there will be platform-specific calls to the eigensolver. (AFAIK the non-symmetric or general eigenvalue solvers haven't been implemented for GPU offload anywhere besides MAGMA).
But more importantly,  the linear method only needs one eigenvector, so switching to an iterative solver (like ARPACK) improves performance considerably.   At that point, fillOverlapHamiltonMatrices becomes the bottleneck.

src/QMCDrivers/WFOpt/LinearMethod.h
```diff
@@ -36,6 +36,18 @@ class LinearMethod
   void getNonLinearRange(int& first, int& last, const QMCCostFunctionBase& optTarget) const;
 
 public:
+  // Use generalized eigenvalue solver
```
prckent, line 39: Small point, but since the interface to eigensolvers is being expanded, now would be a good time to improve the comments on these function definitions - what is A, B etc. ?

prckent (changes requested):
Thanks Mark.

I think no reason to pull this functionality out to Numerics yet - that just risks building extra infrastructure that we don't have a use for. Easy to move if we find extra uses.

src/QMCDrivers/WFOpt/LinearMethod.cpp
```diff
@@ -20,9 +20,116 @@
 #include <vector>
 #include "QMCCostFunctionBase.h"
 #include <CPU/BLAS.hpp>
+#include "Numerics/MatrixOperators.h"
+#include "Numerics/DeterminantOperators.h"
```
ye-luo, line 24: What is this needed for?

src/QMCDrivers/WFOpt/LinearMethod.cpp
```diff
@@ -20,9 +20,116 @@
...
+  assert(A.rows() == A.cols());
+  assert(Nl == eigenvals.size());
+  assert(Nl == eigenvectors.rows());
+  assert(Nl == eigenvectors.cols());
+
+  invert_matrix(B, false);
+
+  Matrix<Real> prdMat(Nl, Nl);
+
+  qmcplusplus::MatrixOperators::product(B, A, prdMat);
```
ye-luo, line 98: `qmcplusplus::` not needed?

src/QMCDrivers/WFOpt/LinearMethod.cpp
```diff
@@ -20,9 +20,116 @@
...
+  LAPACK::geev(&jl, &jr, &Nl, prdMat.data(), &Nl, &alphar[0], &alphai[0], eigenD.data(), &Nl, eigenvectors.data(), &Nl,
+               &work[0], &lwork, &info);
+  lwork = int(work[0]);
+  work.resize(lwork);
+
+  LAPACK::geev(&jl, &jr, &Nl, prdMat.data(), &Nl, &alphar[0], &alphai[0], eigenD.data(), &Nl, eigenvectors.data(), &Nl,
+               &work[0], &lwork, &info);
+  if (info != 0)
+  {
+    APP_ABORT("Invalid Matrix Diagonalization Function!");
```
ye-luo, line 123: No APP_ABORT in newly written code. Throw exception.

src/QMCDrivers/WFOpt/LinearMethod.h
```diff
@@ -36,6 +36,18 @@ class LinearMethod
...
+  void solveGeneralizedEigenvalues(Matrix<Real>& A,
+                                   Matrix<Real>& B,
+                                   std::vector<Real>& eigenvals,
+                                   Matrix<Real>& eigenvectors) const;
+
+  // Solve by explicitly inverting the overlap matrix
+  void solveGeneralizedEigenvalues_Inv(Matrix<Real>& A,
```
ye-luo, line 46: Can these two functions just marked static?

ye-luo (commented):
Preferably these two functions can be in a new class.

src/QMCDrivers/WFOpt/LinearMethod.cpp
markdewing: MatrixOperators::product and invert_matrix

ye-luo (approved):

prckent (approved):

prckent:
Test this please

prckent:
Test this please

PR #4958, opened 2024-04-17
Nexus: Element names need to be capitalized in incorporate_system at fileio.py

kayahans:
## Proposed changes

Element names need to be capitalized to be able to run qdens. 
For a MoS2 calculation, I used the qdens in the following way:
qdens -e 100 -a -v -f xsf -i dmc.g000.twistnum_0.in.xml *.s002.stat.h5

With averaging over multiple twists, the run dies without any error message (only prints "Killed" and terminates).
For a single twist, I get the following error: 
`*** KeyError: 'MO'
`

After debugging, I found that MO should stand for Molybdenum. After converting MO to Mo, qdens script works as expected for a single twist. 

- Bugfix

- No

Bilayer MoS2


prckent:
Can you explain where the uppercase MO is coming from in your example? Ideally Nexus has and enforces a consistent use throughout. The fix here is user friendly so we could run with it, but there must be other cases where capitalization is not either converted or enforced.

kayahans:
I think the bug is because of this: https://github.com/QMCPACK/qmcpack/blob/23c5f3661c2f25f14e3124d54e456e5b0acc6cff/nexus/lib/qmcpack_input.py#L2801

Nexus makes the upper/lower case correction of qmcpack input by keeping a list of parameters that are used in the XML file. Therefore the Molybdenum atom symbol is converted from 'Mo' to 'MO' in the qmcpack input file, despite it was written as 'Mo' in the nexus input script. I think this is a problem specific to Mo atom, because the "MO" is a tag in qmcpack input. This does not interfere with the qmcpack run, but only seems to be a problem when the structure info is read from the xml file (in qdens), where qmcpack does not check capitalization.

prckent (commented):
@jtkrogel Any potential side effects of this update we should be aware of?

(Given @kayahans 's analysis, I think the current fix seems very reasonable and pragmatic)

jtkrogel:
There is already a feature that performs this type of function (atomic symbol processing).  Better to use this rather than duplicate.

I think this loop file fileio.py:
```
        for e in s.elem:
            ne = len(e)
            if ne>1:
                if ne==2 and not e[1].isalpha():
                    e = e[0]
                elif ne>2:
                    e = e[0:2]
                #end if
            #end if
            if is_element(e):
                elem.append(ptable.elements[e].atomic_number)
            else:
                elem.append(0)
            #end if
        #end for
```

Can be replaced with this one:
```
        for e in s.elem:
            is_elem,e = is_element(e,symbol=True)
            if is_elem:
                elem.append(ptable.elements[e].atomic_number)
            else:
                elem.append(0)
            #end if
        #end for
```
If you try this, does it fix the issue?

prckent (approved):
Thanks all

prckent:
Test this please

ye-luo:
Test this please

ye-luo (approved):

PR #4961, opened 2024-04-19
Remove driver scope time and use function scope

ye-luo:
## Proposed changes
driver class scope time didn't work well when a VMC driver is nested inside a WFOpt driver.
A lesson learned is that scoped timer should only be used in a function scope and cannot be in a class scope.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

ye-luo:
Test this please

prckent (approved):

PR #4969, opened 2024-05-04
Set the number of teams as the loop bound.

ye-luo:
## Proposed changes
Very minor change and should not affect correctness.
The main intention is to confirm recent amdgpu driver update.

- Bugfix
- Other (please describe):

- No

epyc
## Checklist


ye-luo:
Test this please

prckent (approved):
Interesting that this works and appears to have the correct driver?

ye-luo:
> Interesting that this works and appears to have the correct driver?

The driver query from CI
```
============================ ROCm System Management Interface ============================
============================== Version of System Component ===============================
Driver version: 6.7.0
==========================================================================================
================================== End of ROCm SMI Log ===================================
```
which is good.

PR #4974, opened 2024-05-07
Two sets of refactoring changes

ye-luo:
## Proposed changes
1. Introduce value_type and size_type in VectorSoaContainer to align with other containers.
2. Pass stream handles by value since they are pointers.

- No

epyc-server

## Checklist

### DIFF
--- a/src/Containers/OhmmsSoA/VectorSoaContainer.h
+++ b/src/Containers/OhmmsSoA/VectorSoaContainer.h
@@ -34,7 +34,8 @@ template<typename T, unsigned D, typename Alloc = aligned_allocator<T>>
 struct VectorSoaContainer
 {
   using AoSElement_t = TinyVector<T, D>;
-  using Element_t    = T;
+  using size_type    = size_t;
+  using value_type   = T;
 
   ///default constructor
   VectorSoaContainer() : nLocal(0), nGhosts(0), nAllocated(0), myData(nullptr) {}
@@ -44,7 +45,7 @@ struct VectorSoaContainer
    * @param n new nLocal
    * @param n_padded new nGhosts
    */
-  VectorSoaContainer(T* ptr, size_t n, size_t n_padded) : nLocal(n), nGhosts(n_padded), nAllocated(0), myData(ptr) {}
+  VectorSoaContainer(T* ptr, size_type n, size_type n_padded) : nLocal(n), nGhosts(n_padded), nAllocated(0), myData(ptr) {}
 
   ///destructor
   ~VectorSoaContainer() { free(); }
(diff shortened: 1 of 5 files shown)


ye-luo:
Test this please

src/Containers/OhmmsSoA/VectorSoaContainer.h
```diff
@@ -34,7 +34,8 @@ template<typename T, unsigned D, typename Alloc = aligned_allocator<T>>
 struct VectorSoaContainer
 {
   using AoSElement_t = TinyVector<T, D>;
-  using Element_t    = T;
+  using size_type    = size_t;
```
PDoakORNL, line 37: why not `Value` and `Size` in line with the type naming rules?
size_type and value_type still read as variables

src/Platforms/CUDA/cuBLAS_missing_functions.hpp
```diff
@@ -47,7 +47,7 @@ using cuBLAS_MFs_handle = cudaStream_t;
  * @param incy increment for the elements of y. It cannot be zero.
  * @param batch_count batch size
  */
-cuBLAS_MFs_status gemv_batched(cuBLAS_MFs_handle& handle,
+cuBLAS_MFs_status gemv_batched(cuBLAS_MFs_handle handle,
```
PDoakORNL, line 50: since we're changing these lines anyway can we remove the useless typedef that only serves to make this files types not match those in the implementation file?

PDoakORNL (changes requested):
We can touch the same lines and actually improve how the code reads.

ye-luo:
> We can touch the same lines and actually improve how the code reads.

Containers follow std::vector pattern to maximize compatibility

src/Containers/OhmmsSoA/VectorSoaContainer.h
ye-luo: Containers follow std::vector pattern to maximize compatibility

src/Platforms/CUDA/cuBLAS_missing_functions.hpp
ye-luo: Will fix it.

src/Containers/OhmmsSoA/VectorSoaContainer.h
PDoakORNL: right sorry

src/Platforms/CUDA/cuBLAS_missing_functions.hpp
ye-luo: Done.

ye-luo:
Test this please

PDoakORNL (approved):

prckent (changes requested):
Better title for the PR please -- I have to write the release notes

Also, is this a bug fix in the wrappers?

ye-luo:
It is refactoring only.

ye-luo:
@prckent good to go?

prckent (approved):

PR #4976, opened 2024-05-07
Add Queue<PlatformKind::CUDA>

ye-luo:
~~Please review after merging #4973 and #4974~~

## Proposed changes
Add a new abstraction class Queue to unify code paths.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Platforms/CUDA/CUDALinearAlgebraHandles.h
+++ b/src/Platforms/CUDA/CUDALinearAlgebraHandles.h
@@ -15,20 +15,20 @@
 #include "ResourceCollection.h"
 #include "CUDA/CUDAruntime.hpp"
 #include "CUDA/cuBLAS.hpp"
+#include "CUDA/QueueCUDA.hpp"
 
 namespace qmcplusplus
 {
 struct CUDALinearAlgebraHandles : public Resource
 {
   // CUDA specific variables
-  cudaStream_t hstream;
+  compute::Queue<PlatformKind::CUDA> queue;
   cublasHandle_t h_cublas;
 
   CUDALinearAlgebraHandles() : Resource("CUDALinearAlgebraHandles")
   {
-    cudaErrorCheck(cudaStreamCreate(&hstream), "cudaStreamCreate failed!");
     cublasErrorCheck(cublasCreate(&h_cublas), "cublasCreate failed!");
-    cublasErrorCheck(cublasSetStream(h_cublas, hstream), "cublasSetStream failed!");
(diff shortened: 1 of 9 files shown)


ye-luo:
Test this please

src/Platforms/CUDA/CUDALinearAlgebraHandles.h
```diff
@@ -15,28 +15,27 @@
...
 #include "CUDA/cuBLAS.hpp"
+#include "CUDA/QueueCUDA.hpp"
 
 namespace qmcplusplus
 {
 struct CUDALinearAlgebraHandles : public Resource
 {
   // CUDA specific variables
-  cudaStream_t hstream;
+  compute::Queue<PlatformKind::CUDA> queue;
```
PDoakORNL, line 25: feels like now might be the time to change the name of this class. the minute the sycl queue shows up the error messages are all going to be sometimes wrong as well.

PDoakORNL (changes requested):
I think propagating better and coding convention following type names from the previous review through here would be worth it.

src/Platforms/CUDA/CUDALinearAlgebraHandles.h
ye-luo: Were you referring to CUDALinearAlgebraHandles? Yes it has to be changed. It is not clear how exactly. I probably will demolish this class. So far it is still CUDA only and should not be renamed in this PR.

PDoakORNL: Yes I was referring to this.

prckent (commented):
(Small initial comment) I prefer that we don't call things Queues if they are not STL Queues. Perhaps an additional prefix/suffix or different name?

ye-luo:
> (Small initial comment) I prefer that we don't call things Queues if they are not STL Queues. Perhaps an additional prefix/suffix or different name?

I added namespace protection for this purpose. It is `qmcpusplus::compute::Queue` in full

PDoakORNL:
I think SYCL calls them queues but I'm not sure the semantics are really the same as std::queue.  MAGMA also calls their streams queues but there I think the semantics are closer to std::queue.

ye-luo:
@PDoakORNL rebased and reduced code change. Ready for final review.

PDoakORNL (approved):

PDoakORNL:
This looks good to me now. I do think we should try to get encourage use of the `compute::` in future code that uses this.

ye-luo:
Test this please

PR #4979, opened 2024-05-09
Add Mixed Precision Coverage Runs in GitHub Actions

prckent:
## Proposed changes

This is an experiment to see if adding two more coverage runs is reliable and practical/fast enough.

It adds the "missing" MPI Mixed precision builds to GitHub Actions CI, and will increase coverage.

If successful over a week+ of usage, we can refine further. If not, this is easy to revert.

- No

None

## Checklist


ye-luo:
Most mixed precision code paths share with full precision. I don't really expect coverage to increase.
However, considering coverage that is slow to run and unreliable, I would prefer adding variants on other builds.

prckent:
Test this please

prckent:
Lets just watch the turnaround and reliability.

Sometime "soon" we should update the compiler versions in CI.

PR #4984, opened 2024-05-10
Nexus: Print warning for the Nexus user before pwscf rsync

aannabe:
## Proposed changes

When a new NSCF is located on a different path than SCF, Nexus will rsync the SCF results into the new path. This is a great feature for submitting multiple NSCF calculations with various k-points. However, this can take quite a while and can give the false sense that Nexus is hanging when, in fact, it is silently rsync'ing the results. This has confused me a few times.

I simply put two print statements warning the user that this is happening. It only executes the first time when the directory does not exist. There is probably a better way to output this - let me know if there is and I will be happy to update.

- No

All Nexus tests pass on my laptop (Linux Mint 21.3), except for 61/69 qdens_radial. 

## Checklist

### DIFF
--- a/nexus/lib/pwscf.py
+++ b/nexus/lib/pwscf.py
@@ -227,6 +227,7 @@ class Pwscf(Simulation):
                 outdir = os.path.join(self.locdir,c.outdir)
                 command = 'rsync -av {0}/* {1}/'.format(result.outdir,outdir)
                 if not os.path.exists(outdir):
+                    print('Running rsync for the {} directory. This might take a while.'.format(outdir))
                     os.makedirs(outdir)
                 #end if
                 sync_record = os.path.join(outdir,'nexus_sync_record')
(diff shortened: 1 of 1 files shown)


prckent (dismissed):
Good idea

nexus/lib/pwscf.py
```diff
@@ -227,6 +227,7 @@ def incorporate_result(self,result_name,result,sim):
                 outdir = os.path.join(self.locdir,c.outdir)
                 command = 'rsync -av {0}/* {1}/'.format(result.outdir,outdir)
                 if not os.path.exists(outdir):
+                    print('Running rsync for the {} directory. This might take a while.'.format(outdir))
                     os.makedirs(outdir)
```
ye-luo, line 231: Please add a paired printout indicating the completion of rysnc.

aannabe: Done. Also moved the earlier print to just before the rsync. Makes more sense this way.

ye-luo: Please add back the printout before copy. They service distinct purpose and there is no harm to be verbose on the status of nexus.

ye-luo:
Test this please

ye-luo (approved):

PR #4985, opened 2024-05-13
Nexus: Fix the qdens-radial test

aannabe:
## Proposed changes

The Nexus `qdens-radial` test is passing for me when running on a NERSC machine, but it's failing on my Linux laptop. Upon inspection, the test is failing due to float comparison with many significant digits, which is a problematic thing to do:

Laptop:
```
Cumulative Value of C Species at Cutoff 0.6 is: 1.1078267486386277+/-0.0016066467832022906
```

NERSC:
```
Cumulative Value of C Species at Cutoff 0.6 is: 1.1078267486386275+/-0.0016066467833404942
```

In practice, we don't need 16 digits after decimal for charge densities. So I changed the `qdens-radial` output to have 8 digits after decimal and updated the reference data for the tests. After this change, all tests are passing for me locally and on NERSC.

- No

NERSC, Laptop

## Checklist

### DIFF
--- a/nexus/bin/qdens-radial
+++ b/nexus/bin/qdens-radial
@@ -499,9 +499,9 @@ class QMCDensityRadialProcessor(QDRBase):
                     r = opt.radii[0]
                 #end if
                 if not show_error:
-                    self.log('{} Value of {} Species at Cutoff {} is: {}\n'.format(cstr,at,r,rdens.tot[at].density[-1]))
+                    self.log('{} Value of {} Species at Cutoff {} is: {:.8f}\n'.format(cstr,at,r,rdens.tot[at].density[-1]))
                 else:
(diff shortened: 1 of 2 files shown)


prckent (approved):
Thanks for investigating this. It has been a problem from time to time in various nightly tests etc.   Your modifications are an indirect way of adding a tolerance to 8 digits. This works for me and I agree with the reasons you give.

jtkrogel:
An alternative (direct) way is to add the absolute tolerance to the equality check:

```
assert(text_eq(out,out_ref,atol=1e-8))
```

aannabe:
Added the tolerance explicitly.

prckent (approved):
Thanks Gani

ye-luo:
Test this please

PR #4986, opened 2024-05-14
Move psiMinv out of delayed update engines to DiracDeterminantBatched

ye-luo:
## Proposed changes
Make delayed update engines holding less data.

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
+++ b/src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
@@ -104,16 +104,7 @@ public:
     }
   };
 
-  const DualMatrix<Value>& get_psiMinv() const { return psiMinv_; }
-  DualMatrix<Value>& get_ref_psiMinv() { return psiMinv_; }
-
 private:
-  /* inverse transpose of psiM(j,i) \f$= \psi_j({\bf r}_i)\f$
-   * Only NumOrbitals x NumOrbitals subblock has meaningful data
-   * The number of rows is equal to NumOrbitals
-   * The number of columns in each row is padded to a multiple of QMC_SIMD_ALIGNMENT
-   */
-  DualMatrix<Value> psiMinv_;
   /// scratch space for rank-1 update
   UnpinnedDualVector<Value> temp;
   /// row of up-to-date Ainv
@@ -195,7 +186,9 @@ private:
    * @param Ainv inverse matrix
    * @param rowchanged the row id corresponding to the proposed electron
    */
-  static void mw_prepareInvRow(const RefVectorWithLeader<This_t>& engines, const int rowchanged)
+  static void mw_prepareInvRow(const RefVectorWithLeader<This_t>& engines,
+                               const RefVector<DualMatrix<Value>>& psiMinv_refs,
+                               const int rowchanged)
   {
     auto& engine_leader              = engines.getLeader();
     auto& cuda_handles               = engine_leader.cuda_handles_.getResource();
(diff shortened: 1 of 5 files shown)


ye-luo:
Test this please

ye-luo:
Test this please

src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
```diff
@@ -208,7 +196,7 @@ class MatrixDelayedUpdateCUDA
     auto& cone_vec                   = mw_mem.cone_vec;
     auto& czero_vec                  = mw_mem.czero_vec;
     auto& prepare_inv_row_buffer_H2D = mw_mem.prepare_inv_row_buffer_H2D;
-    const int norb                   = engine_leader.get_psiMinv().rows();
+    const int norb                   = engine_leader.invRow.size();
```
PDoakORNL, line 199: If feel like getting this value from the passed psiMInv_ref is less of a change and depends on more straight forward input. But It seems like consistency between psiMInv rows and invRow(s) also must be maintained.

src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
```diff
@@ -698,8 +692,8 @@ class MatrixDelayedUpdateCUDA
     auto& queue                = engine_leader.cuda_handles_.getResource().queue;
     auto& h_cublas             = engine_leader.cuda_handles_.getResource().h_cublas;
     auto& updateInv_buffer_H2D = engine_leader.mw_mem_handle_.getResource().updateInv_buffer_H2D;
-    const int norb             = engine_leader.get_psiMinv().rows();
-    const int lda              = engine_leader.get_psiMinv().cols();
+    const int norb             = engine_leader.invRow.size();
```
PDoakORNL, line 695: if instead this is a indirect consistency check, an assert would probably be a better way to handle it.  If there is a chance they may not be consistent some documentation somewhere would be useful.

src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
```diff
@@ -431,18 +420,17 @@ class MatrixDelayedUpdateCUDA
...
   static void mw_evalGrad(const RefVectorWithLeader<This_t>& engines,
+                          const RefVector<DualMatrix<Value>>& psiMinv_refs,
                           const std::vector<const Value*>& dpsiM_row_list,
                           const int rowchanged,
                           std::vector<GT>& grad_now)
   {
     auto& engine_leader = engines.getLeader();
     if (!engine_leader.isSM1())
-      mw_prepareInvRow(engines, rowchanged);
+      mw_prepareInvRow(engines, psiMinv_refs, rowchanged);
```
PDoakORNL, line 433: this seems like a great chance to rename isSM1 since I immediately had to search for what it meant.

PDoakORNL (changes requested):
I just want some clarity on the apparent change of the source of the `norb` dependent/local variables source and the size consistency relation between invRows and psiMinv. 

Other than that looks like a clean refactor.

src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
ye-luo: It is a stopgap for the moment. I will actual be able to store norb as a const via the constructor since DiracDeterminantBatched also have the size ironed at the constructor. This was never possible when MatrixDelayedUpdateCUDA was first introduced.

src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
ye-luo: I also would like to change it. It will happen when I rename MatrixDelayedUpdateCUDA class. Cannot do it on this PR due to complicating following PRs. Use https://github.com/QMCPACK/qmcpack/issues/4996 for tracking.

PDoakORNL (approved):

src/QMCWaveFunctions/Fermion/MatrixDelayedUpdateCUDA.h
PDoakORNL: Sounds like a good approach.

ye-luo:
Test this please

PR #4990, opened 2024-05-14
Split eigenvalue solve from eigenvalue selection and add choice of solvers

markdewing:
Move the eigenvalue selection code to a separate function, so it can be re-used with different solvers.

Add an "ev_solver" parameter to allow switching between the two implemented solution methods.
The choice of eigenvalue solvers is only implemented for the "one shift only" optimizer in the batched driver.

Prior to this PR, the code using the generalized eigensolver existed in the code, but was not connected.   More importantly, these changes are paving the way for a larger selection of eigenvalue solvers.

One change in logic is the eigenvalue problem is solved on one rank and the resulting change in parameters is broadcast to the other ranks. Previously, the eigenvalue problem was solved on each rank separately.  This is looking forward to different solvers which may not produce exactly the same answer when run multiple times.  The amount of data broadcast is small, so the effect on performance should be minimal.
The broadcast is here:
https://github.com/QMCPACK/qmcpack/blob/355302fe9da8dbb73e555b09b4b9f20ade6d5688/src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp#L1693

I can add documentation (and more tests) if the code is okay.

- No

desktop

## Checklist

- Not yet. Documentation has been added (if appropriate)

### DIFF
--- a/src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
+++ b/src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
@@ -576,6 +576,11 @@ void QMCFixedSampleLinearOptimizeBatched::process(xmlNodePtr q)
   m_param.add(OutputMatricesHDF, "output_matrices_hdf", {"no", "yes"});
   m_param.add(FreezeParameters, "freeze_parameters", {"no", "yes"});
 
+  m_param.add(ev_solver_, "ev_solver",
+      {"inverse",  // Inverse + nonsymmetric eigenvalue solver
+       "general"   // General eigenvalue problem solver
+      });
+
   oAttrib.put(q);
   m_param.put(q);
 
@@ -1619,7 +1624,7 @@ bool QMCFixedSampleLinearOptimizeBatched::one_shift_run()
       output_hamiltonian_.output(hamMat);
     }
 
-    if (do_output_matrices_hdf_)
+    if (do_output_matrices_hdf_ && is_manager())
     {
       std::string newh5 = get_root_name() + ".linear_matrices.h5";
       hout.create(newh5, H5F_ACC_TRUNC);
@@ -1632,7 +1637,8 @@ bool QMCFixedSampleLinearOptimizeBatched::one_shift_run()
               << std::endl;
   }
 
-  {
+  // Solve eigenvalue problem on one rank.
+  if (is_manager()) {
     ScopedTimer local(eigenvalue_timer_);
     Timer t_eigen;
     app_log() << std::endl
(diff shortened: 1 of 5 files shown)


src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
```diff
@@ -1649,28 +1655,24 @@ bool QMCFixedSampleLinearOptimizeBatched::one_shift_run()
...
-
-    // transpose the result (why?)
-    for (int i = 0; i < N; i++)
-      for (int j = i + 1; j < N; j++)
-        std::swap(prdMat(i, j), prdMat(j, i));
+    RealType lowestEV;
+    // compute the lowest eigenvalue and the corresponding eigenvector
+    if (ev_solver_ == "general") {
+      app_log() << "  Using generalized eigenvalue solver (ggev)" << std::endl;
+      lowestEV = getLowestEigenvector_Gen(hamMat, invMat, parameterDirections);
```
ye-luo, line 1667: > The code using the generalized eigensolver has been in the code, but has not been connected. 

Is it not connected here?

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
```diff
@@ -576,6 +576,11 @@ void QMCFixedSampleLinearOptimizeBatched::process(xmlNodePtr q)
   m_param.add(OutputMatricesHDF, "output_matrices_hdf", {"no", "yes"});
   m_param.add(FreezeParameters, "freeze_parameters", {"no", "yes"});
 
+  m_param.add(ev_solver_, "ev_solver",
```
ye-luo, line 579: Would like to call it `eigensolver` than `ev_solver`.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
markdewing: In the past the generalized eigensolver was not connected, but it is connected now.

ye-luo: I see what you mean. PR description reflects the code in the branch. So "has not been connected" confused me.

markdewing: I updated the PR description to be more clear.

src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp
ye-luo: Would you like to add the manual entry in this PR or later? If later, I would like to approve the current PR.

markdewing: Either way is fine. If you want to keep this moving, I can add the manual entry later.

ye-luo (approved):
Still need non-ANL approval

prckent (approved):
LGTM (airport review)

prckent:
> Previously, the eigenvalue problem was solved on each rank separately.
I consider the new implementation a bug fix -- one day we would have been caught by the old implementation. DFT codes that have done similar things have run into trouble historically.

prckent:
Test this please

ye-luo:
> > Previously, the eigenvalue problem was solved on each rank separately.
> > I consider the new implementation a bug fix -- one day we would have been caught by the old implementation. DFT codes that have done similar things have run into trouble historically.

@markdewing could you post a link to the line of broadcast after solving on rank 0?

markdewing:
I updated the PR description with a link to the broadcast, also given here:
https://github.com/QMCPACK/qmcpack/blob/355302fe9da8dbb73e555b09b4b9f20ade6d5688/src/QMCDrivers/WFOpt/QMCFixedSampleLinearOptimizeBatched.cpp#L1693

Issue #4992, opened 2024-05-14
Multidet wavefunction problems in batched code


jtkrogel:
The batched code produces gigantic kinetic energy spikes and incorrect acceptance ratios (with and without drift) for VMC for general multideterminant wavefunctions.

This problem was observed for the LiH test problem included with QMCPACK when the coefficients in the MSD file were changed by hand/with a tool.  

The legacy code handles the updated coefficients without issue, as evidenced by successful self-healing DMC iterations converging on exact results.

Included/attached is a short deterministic 1 MPI 1 OMP-thread run to reproduce and hunt down this issue.  For VMC, the batched and legacy drivers were verified to deterministically reproduce each other for the full original wavefunction as well as the updated wavefunction when restriced to MSD with a single determinant.  The included case uses 2-determinants, which is sufficient to show the problem.

The first presenting issue is that the kinetic energy is crazy for some walker steps with the batched code (the spike at step 9 goes up to **400 Ha**):
![batched_vs_legacy_2det_1walker__kinetic](https://github.com/QMCPACK/qmcpack/assets/24547905/84f1786b-59a3-44ba-8ee9-577779ae2cd9)

The second issue is that the acceptance ratio is too low in the batched code with and without drift, which affects quantities like the electron-electron energy:
![batched_vs_legacy_2det_1walker__elecelec](https://github.com/QMCPACK/qmcpack/assets/24547905/67fca810-594d-4f7d-bd36-37b7313b5993)

The legacy driver (and also the batched one for the correctly functioning related cases described above), produces an acceptance ratio near 60%, while for the problem case with the batched code an acceptance ratio closer to 40% is observed.

The very fast fixed seed deterministic test cases are attached (each run finishes in 10ms):
[lih_short_deterministic.zip](https://github.com/QMCPACK/qmcpack/files/15313694/lih_short_deterministic.zip)

**Expected behavior**
The batched code should reproduce the legacy code deterministically, quantity for quantity, for each MC step.

**System:**

Laptop.

jtkrogel:
Here is a plot of the local energy of the longer original 5-determinant run (batched vs legacy):
![Uploading lih_multidet_5rand__batched_vs_legacy.png…]()

prckent:
Of note, VMC without drift gives wrong acceptance ratio => the value of the wavefunction is wrong.

ye-luo:
> when the coefficients in the MSD file were changed by hand/with a tool.

[...]

PR #4993, opened 2024-05-15
ReferenceDeterminant should only be used at evaluation not preparation.

ye-luo:
## Proposed changes
Fixes https://github.com/QMCPACK/qmcpack/issues/4992 when the determinant with the largest coefficient is not the first one after the cutoff filtering.

- Bugfix

- No

epyc-server

## Checklist


jtkrogel (approved):
I've confirmed this PR fixes the bug

prckent (approved):
Interesting this did not show up earlier....

prckent:
Test this please

PR #4995, opened 2024-05-15
Fix Traces functionality

jtkrogel:
## Proposed changes

Traces became broken due to changes in hdf_archive.  These were not caught by the current 1 mpi 1 omp-thread deterministic traces tests.  

This PR adds a 4 mpi 4 omp-thread deterministic test for traces and updates hdf_archive to allow Traces to function again.  

- Bugfix

- No

Laptop

## Checklist

### DIFF
--- a/src/io/hdf/hdf_archive.cpp
+++ b/src/io/hdf/hdf_archive.cpp
@@ -142,8 +142,8 @@ bool hdf_archive::create(const std::filesystem::path& fname, unsigned flags)
   possible_filename_ = fname;
   if (Mode[NOIO])
     return true;
-  if (!(Mode[IS_PARALLEL] || Mode[IS_MASTER]))
-    throw std::runtime_error("Only create file in parallel or by master but not every rank!");
+  //if (!(Mode[IS_PARALLEL] || Mode[IS_MASTER]))
+  //  throw std::runtime_error("Only create file in parallel or by master but not every rank!");
   close();
   file_id = H5Fcreate(fname.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, access_id);
   return file_id != is_closed;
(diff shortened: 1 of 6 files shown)


src/io/hdf/hdf_archive.cpp
```diff
@@ -142,8 +142,8 @@ bool hdf_archive::create(const std::filesystem::path& fname, unsigned flags)
   possible_filename_ = fname;
   if (Mode[NOIO])
     return true;
-  if (!(Mode[IS_PARALLEL] || Mode[IS_MASTER]))
-    throw std::runtime_error("Only create file in parallel or by master but not every rank!");
```
ye-luo, line 146: Why are these lines offensive? Please explain your use case.

ye-luo: I feel the caller side was using hdf_archive incorrectly.

jtkrogel: This is what blocks Traces from functioning in a parallel run.  Traces creates one H5 file per MPI process.  This does not create significant runtime overhead for QMCPACK vs having traces off (measured at about a 5% performance impact on large scale Titan runs).

jtkrogel: See Sec. IV A in https://doi.org/10.1109/CSE.2013.24

ye-luo: hdf_archive should support one file file per rank when it is not aware of MPI. So the issue is incorrect open mode of the file not the hdf_archive itself. Please provide a link where open() is called from the trace code.

jtkrogel: This works.  I've restored the guards in hdf_archive now and the update for the fix is now confined to TraceManager.h

ye-luo:
Test this please

ye-luo (approved):

PR #5006, opened 2024-05-24
Add documentation and test for eigensolver option.

markdewing:
Add a short test and documentation for the `eigensolver` option added in #4990 .

The existing tests cover the default `inverse` option.  The added test is a copy of optm-batch-OneShiftOnly-general.xml, but using the `general` eigensolver.

- No

desktop

## Checklist


ye-luo (approved):

ye-luo:
Test this please

prckent (changes requested):
Do we have a deterministic or unit test for these?

It would be good to test the code path in CI and at least check we handle the option correctly.

markdewing:
Do you want a deterministic test to check that the right code path is being followed?
A unit test for that would be difficult, but a deterministic test that checks for the right strings in the output could work.

prckent:
Test this please

prckent:
`Fatal Error. Aborting at QMCDriverFactory::createQMCDriver : method="linear_batch" is not safe with CPU mixed precision. Please use full precision build instead.` Hmm. Wasn't aware of this one. CMakeLists just needs protecting.

Coverage uploads failed due to rate limits.

markdewing:
The other option is to add "not safe with CPU mixed precision" to the passing regular expression list.  Do you have a preference between this and guarding the CMake?

prckent:
The latter one (WILL FAIL) seems simplest and also checks that the code bails as expected.

prckent:
Test this please

prckent (approved):

prckent:
Test this please

PR #5010, opened 2024-05-28
Useful walker ids

PDoakORNL:
## Proposed changes

Walker ID's in batched version of the drivers should now supply trackable walker ID's

Practically this looks like after a load balance is done:
44: Test command: /usr/local/bin/mpiexec "-n" "4" "--oversubscribe" "/Users/Shared/ornldev/code/qmcpack/build_new/src/QMCDrivers/tests/test_new_drivers_mpi"

from the "MPI WalkerControl population swap walkers" test case

Section ("Load Balance")
44: count_before: { 4, 1, 1, 1,  }  count_after: { 1, 2, 2, 2,  }
44: Walkers Per Rank (Total: 7)
44: rank: 0  walker ids: { 1,  } parent ids: { 0,  }
44: rank: 1  walker ids: { 2, 6,  } parent ids: { 0, 1,  }
44: rank: 3  walker ids: { 4, 8,  } parent ids: { 0, 1,  }
44: rank: 2  walker ids: { 3, 7,  } parent ids: { 0, 1,  }

Section("Load Balance Multiple Copy Optimization")
44: count_before: { 12, 1, 1, 1,  }  count_after: { 3, 4, 4, 4,  }
44: Walkers Per Rank (Total: 15)
44: rank: 0  walker ids: { 1, 5, 9,  } parent ids: { 0, 1, 1,  }
44: rank: 1  walker ids: { 2, 6, 10, 14,  } parent ids: { 0, 1, 6, 6,  }
44: rank: 2  walker ids: { 3, 7, 11, 15,  } parent ids: { 0, 1, 7, 7,  }
44: rank: 3  walker ids: { 4, 8, 12, 16,  } parent ids: { 0, 1, 8, 8,  }

- New feature

- No

OSX laptop

## Checklist


src/QMCDrivers/MCPopulation.h
```diff
@@ -223,14 +230,25 @@ class MCPopulation
...
+   *  this is is just `R` and `spins` + possibly G, L
+   */
   void saveWalkerConfigurations(WalkerConfigurations& walker_configs);
+
+  /** Generator for walker_ids for this MCPopulation
+   *  This should be the single source for `walker_id`'s on this rank.
+   *  These are unique across MCPopulations with rank.
+   *  Defined as
+   *  \f$ walker_id = num_walkers_created_++ * num_ranks_ + rank_ + 1\f$
+   *  So starting from 1, 0 belongs to the golden walker.
```
ye-luo, line 249: The formula of walker_id should be defined at the the top of MCpopulation. The documentation of a function needs to describe what a function does. It is hard to digest now.

The id of the gold walker is debatable. I tend to not include the golden walker at all. It doesn't participate calculation and should not. So it is not part of the population and doesn't deserve an ID. Even it is assigned an ID, the current formula is incorrect considering there is one golden walker per rank. `( 1+ num_walkers_created_++) * num_ranks_ + rank_ ` makes more sense.

It is acceptable but I don't feel comfortable that the proposed id formula produces gaps in ids.

In addition, it seems that the consumer of `nextWalkerID()` seems to know exactly what it does. It breaks encapsulation. Why is it public?

src/QMCDrivers/DMC/WalkerControl.cpp
```diff
@@ -344,17 +352,25 @@ void WalkerControl::swapWalkersSimple(MCPopulation& pop)
...
     ncopy_pairs.push_back(std::make_pair(static_cast<int>(good_walkers[iw]->Multiplicity), iw));
+  }
+  // sort good walkers by the number of copies
   std::sort(ncopy_pairs.begin(), ncopy_pairs.end());
 
   struct job
   {
-    const int walkerID;
+    // Walker_index is just its index not its "walker_id"
+    const int walker_index;
```
ye-luo, line 371: very helpful renaming.

src/QMCDrivers/MCPopulation.cpp
```diff
@@ -64,18 +64,36 @@ void MCPopulation::createWalkers(IndexType num_walkers, const WalkerConfiguratio
 
   outputManager.pause();
 
-  //this part is time consuming, it must be threaded and calls should be thread-safe.
-#pragma omp parallel for
+  // nextWalkerID is not thread safe so we need to get our walker id's here
+  // before we enter a parallel section.
+  std::vector<long> walker_ids(num_walkers_plus_reserve, -1);
+  for (size_t iw = 0; iw < num_walkers; iw++)
+    walker_ids[iw] = nextWalkerID();
```
ye-luo, line 71: `nextWalkerID` is just horrible.
Why not having
```
computeWalkerID(iw) const
{
  return iw * num_ranks_ + rank_ + 1;
}
```
and get rid of `num_walkers_created_` state

ye-luo (commented):
Did my first pass.

prckent:
I think there are only two requirements to satisfy here (i) scheme is performant [e.g. minimal comms] (ii) walkers IDs are unique.  If it helps -- and I expect it does -- i think walker IDs can be completely arbitrary numbers.  Rationale: Postmortem analysis will be done by automation.  Do we have any other actual requirements vs nice to have properties (e.g., sequential, increasing)?

src/QMCDrivers/MCPopulation.h
PDoakORNL: It shouldn't be public, will change.

src/QMCDrivers/MCPopulation.cpp
PDoakORNL: that won't work as spawns can occur later.  There needs to be state to not reuse walker ids which will break the whole intent which is for each walker to have an unique id through its life.

PDoakORNL:
If we use UUID's (sorry I they are only called GUIDs in windows.) We would need no state, but without post processing it would be basically human unreadable.

ye-luo:
Here is my expectation of walker ID. It represents the location(rank, index within the rank) of walker in each step. It is not carried over steps. At initialization or after branching, walker ids are reassigned and the old value is kept as parent walker id. Both current and parent ids should allow us to reconstruct walker traces.

> The only state needed is the number of walkers created by each rank, and this never has to be communicated. 

I don't get why this is even needed in the current scheme. Since walkers are created and annihilated from step to step, I don't think `number of walkers created by each rank` really means anything.

If the intention is making the walker id never change after creation. I would like to see how branching and trace reconstruction is managed. I don't mean implementation but some description before adopting this idea.

Regarding golden walkers. Walkers have configurations (electron positions) and contexts (particleset, TWF, Ham).
During the initialization before hitting QMC drivers, a full context is built and later used to cloning. That is a called a golden walker. To be accurate, it should be called golden walker context. Only driver cloned contexts are used for calculation and thus the gold walker context is not affected by any drivers being created and destroyed.
```
  MCPopulation(int num_ranks,
               int this_rank,
               ParticleSet* elecs,
               TrialWaveFunction* trial_wf,
               QMCHamiltonian* hamiltonian_);
```
Treat `elecs, trial_wf, hamiltonian` as reference as they are owned by particleset, twf, ham pools.
For this reason, the golden walkers are not members of a population.

prckent:
I think we need to discuss the envisioned use cases for this. What myself, Jaron, and others are after is indeed "making the walker ID never change after creation", so that we can e.g. follow the history of presumed "bad" walkers. With a walker's unique ID plus that of its parent (which it would have to store), we could do that. IDs would never be reused so the scripting would be easy. I believe this is what the legacy code attempts/attempted to do as well, not that it been carefully verified in the modern era.

Do you or anyone else have other uses in mind?

ye-luo:
To follow the bad walker, you will need to "backtrace". If IDs are fully unique, you will need to search the parent id in the population at every step which is very bad. If following my ID scheme, ID is an index and there is no need of searching.

ye-luo:
> Maybe some crossed language here: once created, walkers have fixed ID independent of timestep. So if a run aborts due to population explosion (most common case), we just follow the low energy walker back to when it was moved to the bad position (assuming we were logging output). This is usually a walker with a single ID. Perhaps your index == what I was describing as a fixed ID? See my example a few comments back.

In my indexing way, the index is only valid for the current step, the parent index is valid for the previous step.

In your fixed ID way, the ID is unique but look up is painful. At step N, walker with id=5 goes bad and its parent id is X, we backtrace to the step N-1, it has 100 walkers, due to unsorted walker ID, need to search all the 100 walkers to find X in order to read info like coordinates/energy/weight. How about now we have 10^6 walkers...

ye-luo:
When using my index(parent index).
Step 0.
rank 0: walkers 0(0),2(2)
rank 1: walkers 1(1),3(3)
step 1.
rank 0: walkers 0(0),2(2)
rank 1: walkers 1(1),3(3),5(1) # both 1 and 5 are branched from walker 1 of step 0.
step 2.
rank 0: walkers 0(0),2(5) # walker 2 is killed and replaced by branching off walker 5 of step 1.
rank 1: walkers 1(3),3(5) # walker 1 of step 1 is killed and load balancing shuffled walkers.

@prckent Please also clarify what is parent walker ID in your scheme and how to handle killing/branching.

prckent:
>How about now we have 10^6 walkers instead of 100 ... Looking up takes O(N).
The only envisaged use I have for this facility is offline debugging and maybe occasional online debugging. There is zero cost to everyday simulations beyond moving a few extra bytes and one extra piece of state. See what I put in the requirements above. What are you envisaging using this for?

ye-luo:
I'm thinking about a fast way of backtracing a bad walker and access infos like coordinates and energy on the evolution path in the analysis phase. This analysis is usually done offline.

Walker id is the only way of tracing walkers. The walker id from creation is not enough to backtrace a walker to step 0 due to branching and thus the parent walker id is also needed. That is why I'm wondering with your labeling scheme how to reconstruct a backtrace.

Particularly, I need this backtrace construction fast and not disk I/O and memory intensive. When I know a parent walker id X, i need an efficient way to find X at step N-1 and read out coordinates and energy.
1. preferably read only partial data from disk and work only with this partial data. Reading all the walkers of all the MPI ranks of the whole history from disk and storing in the memory is bad. Preferably data can be read per step or a step block.
2. With the partially read data, looking up still needs to be fast. If there are a million record in a step block, a skip forward method should be used to locate X. Linear searching X is bad.

jtkrogel:
Indeed, I intend for both walker and parent ID's to be immutable and carried with a walker upon/following its creation.

Using hash tables in post-processing reduces search times to O(1) following an O(N) construction.  

The walker tracing functionality (#5019) naturally organizes the walker data by rank (one file per rank), obviating the need to track it explicitly.  As Ye and I have discussed offline, a slight generalization of the current functionality could optionally segment the outputted buffer data by MC block to further reduce the construction cost if data only from particular blocks is desired.

PDoakORNL:
Examining in more detail how the driver code actually sequences the transfers and copies I realized we can in fact do all the necessary walker spawning so that rank id's are local, parent ids will allow tracking the originating walker whether it be on or off rank.  See the new walker control tests for how this works in more detail.  Much more of the walker control is now actually unit tested as well.  I'd still consider coverage to be pretty incomplete as the tests are easy caes and only single stage, but I do think they nail down the expected behavior more clearly than in the past.

ye-luo:
@PDoakORNL could you clarify "rank id's are local"? What is rank id you were referring?
My understanding is that walker id is defined in each step and thus meaningful for that step only. At a given step, each walker on every MPI rank has a unique id. parent id is for the previous step. This allows tracing a walker back to the previous step. Is that what you meant?

There are two ways of indexing walkers at a given step.
1. contiguously indexing. rank 0 (walker 0, walker 1...) rank 1(walker 0, walker 1...)
2. more or less the scheme used in the current PR. non contiguously indexing. gid = Np * local_index + RankID. NP is the number of MPI ranks, local_index is the index of a walker on its own MPI rank.

Method 1 eliminates the info about ranks. It is more flexible if rank info is not critical. It can also be more difficult if rank info is needed.
Method 2 requires to know the number of walkers on each rank but more friendly to handling data by ranks.
I'm open to both schemes defining walker IDs.

I don't quite follow you mention of ""parentless" walkers" and also have difficulty to understand your two stage comments.

WalkerConfiguration carries Walkers not ParticleSet. It carries data cross QMC drivers.
MCPopulation is a driver concept. It captures a WalkerConfiguration to start a walker population. With a population created, walker id makes sense. Once a QMC driver ends, the population is technically gone and walker id is not meaningful anymore.

so in a driver
```
create a population based on loaded WalkerConfiguration
assign initial walker id. parallel id is basically undefined.
loop step
   random walking
   branching.
   parent id = walker id. Reassign walker id
end loop step
```

src/QMCDrivers/MCPopulation.h
PDoakORNL: The golden walker actually exists in MCPopulation only as a concept, it isn't actually present as a member and if it were it should be equivalent on all ranks so it doesn't need and or shouldn't have its own walker id.  It never moves so it is not logged.

PDoakORNL:
I mean that on a rank all walkers id's will be walker_id % (num_ranks) = rank + 1.   I'll edit that to make it more clear.

I believe in practice walkers should not be getting moved more than necessary so the ID's should be presevered between steps unless the walker actually changed rank or unless it is a copy from another walker on the rank in which case it gets a new id. In practice I don't think a large % of walker die, are born, or transferred each step.

IDs are never reused and are not based on local index only on number of new walkers born on the rank.  There are no gaps if you look at the sequence of particular ranks IDs with integer division by number of ranks.

**IDs should be unique for a particular walker and there cannot be based on the local walker index**

My understanding of mw_saveWalker ... mw_loadWalker was that it was not supposed to just treat the walkers and walker elements as just slots that could have a completely new identity every step as we serialize in and out of some insane untyped byte buffer. 

But that rather a walker than had not been killed would be a continuation of a particular walker trajectory. Its hard to be certain since this whole save load thing moving state between walkers and particle sets is a little... 😭 But I have always intended that a indexes of walker elements will never be mixed. I.e. Walker[1], ParticleSet[1], TWF[1], Hamiltonian[1]

ye-luo:
Here is an example of the bubble I was referring to.
When starting 3 walkers on rank 0, 1 and 2 walkers on rank 2, 3 for total request 10 walkers.
Rank 0 ids: 0, 4, 8
Rank 1 ids: 1, 5, 9
Rank 2 ids: 2, 6
Rank 3 ids: 3, 7
at this point there is no gap. Now rank 3 creates a walker during branching.
Rank 3 ids: 3, 7, 11
There is no walker  with id 10.

ye-luo:
If the walker id is only intended to generate a unique id, then the scheme in this PR is ok. However, it cannot be used to be more informative indicating MPI rank or index. As long as killed/born/migrating do happen, such scenarios should be taken into account. I don't think how large % of walker affects our criteria for choosing a scheme for id.

ye-luo:
> I believe in practice walkers should not be getting moved more than necessary so the ID's should be presevered between steps unless the walker actually changed rank or unless it is a copy from another walker on the rank in which case it gets a new id.

Now @PDoakORNL is saying walker id is mutable. @prckent @jtkrogel @PDoakORNL we need a discussion.

PDoakORNL:
I should also add that any time I'm think about producing massive amounts of data I consider how much work getting it into a relational database would be. That is after all "what you do" if working with data that won't fit in memory that you want to make structured queries against.

Id use tables like:
Walker
---------------------
WalkerKey (Primary Key)
Block
Walker Id
Parent ID

Step
-----------------
stepkey (primary key)
walker index
step
block
kinetic energy
potential energy
...

Conveniently if you limited your database to just one block per instance you could just user walker ID as the key to the first table and step+walkerID as the key second table.

To track a particular walker select all steps for a particular walker ID say (12). Then I'd select all walkers IDs with a parent ID (12). Then query each of those IDs and their steps.  You would continue this process until you got no more walkerID's connected to your original ID.  Merge all the resulting rows together you could plot a nice graph of the trajectory of some value with respect to the originating walker.  If it every reached multiplicity high enough to get copied you would get  multiple values per step otherwise you would just recover the trajectory even as it moved between ranks.

Conceptually you can accomplish the same thing with a scripting language just hammering on the output files but I'd rather have WalkerIDs that made processing those files into a DB easy.  The issue of primary keys is basically the same as in memory hash.

PDoakORNL:
Walker ID is not mutable.  I walker that migrates is killed and born again.

ye-luo:
> Walker ID is not mutable. I walker that migrates is killed and born again.

Unfortunately killing a physical term not a technical term. Migranting a walker cannot be treated as kill and born.

PDoakORNL:
From my perspective there are no longer any gaps.

The walker IDs come from a N = num_ranks disjoint sequences from the positive integers with each value x = {0,..., num_ranks-1} determines a sequence of walker ids such that members are those integers such that w_id -> P_INT % num_rankss == x.  Each rank uses the set of IDs such the x defining the set has x == rank.  on any particular rank All ID's are used sequentially for active walkers. An active walker is one that will on the next step attempt a move. A walker being overwritten must have either been freshly constructed or have been however briefly in an inactive state and receives the next rank ID.

ye-luo:
In the above [example](https://github.com/QMCPACK/qmcpack/pull/5010#issuecomment-2148617363)  I was talking about gaps between ranks not within a rank. ids spaced by num_ranks were not considered disjoint.

PDoakORNL:
In your example the next time rank two creates a walker it will be 10.

ye-luo:
> In your example the next time rank two creates a walker it will be 10.

At the current step. there is no walker 10. If the simulation ends here, there is no walker 10.

jtkrogel:
Perhaps just post the pseudo-code and the ID's generated by 2-3 adjacent ranks?

docs/developing.rst
```diff
@@ -914,6 +914,21 @@ An example of the second approach is
...
   }
 
+Walker
+------
+Lightweight representation of a walker, used as the element in the WalkerConfigurations container class.  In the batched version of the code the ``Walker::walker_id_`` and ``Walker::parent_id_`` allow logging walker instance trajectories etc. The basic data relations involved in are shown in :numref:`fig1`. If limit ourselves to single section logging the each walkerID will be unique generated using the equation
+.. math::
+  :label: eq_walker_id
+  walker_id = walker_id = num_walkers_created_++ * num_ranks_ + rank_ + 1
+
+where ``num_walkers_created_`` is a member variable of the sole ```MCPopulation`` object on the rank and initially set to 0.  Each walkers ``parent_id`` is set at initiation of the walkers configuration to the walker from home the configuration is assigned.  If that assignment is from previous section's or run's ``WalkerConfigurations`` object then the value of the ``Walker::getWalkerID()`` is multiplied by -1. If the Walker's initial configuration comes from the golden particle set the parent_id will be 0.
```
ye-luo, line 924: "set to 0" each driver run or once the whole lifetime?

docs/developing.rst
```diff
@@ -914,6 +914,21 @@ An example of the second approach is
     }
   }
 
+Walker
+------
+Lightweight representation of a walker, used as the element in the WalkerConfigurations container class.  In the batched version of the code the ``Walker::walker_id_`` and ``Walker::parent_id_`` allow logging walker instance trajectories etc. The basic data relations involved in are shown in :numref:`fig1`. If limit ourselves to single section logging the each walkerID will be unique generated using the equation
+.. math::
+  :label: eq_walker_id
+  walker_id = walker_id = num_walkers_created_++ * num_ranks_ + rank_ + 1
```
ye-luo, line 922: typo?

docs/developing.rst
ye-luo, line 924: What is "home"?

src/Particle/Walker.h
```diff
@@ -84,10 +84,22 @@ class Walker
...
+   *  0 is the value of a default constructed walker.
+   *  Any negative value must have been set by an outside entity and indicates
+   *  an invalid walker ID.
    */
   long walker_id_ = 0;
   /** in legacy the ancients have said only:
    *  id reserved for forward walking
+   *
+   *  In Batched
+   *  If a walker is initialized from the golden particle set it keeps the default constructed 0
```
ye-luo, line 99: We are talking about lightweight walker here. Should not reference "golden particle set"

docs/developing.rst
ye-luo: The above section seems explaining what happens when walkerIDs and parent IDs are initialized.
Could you document what happens during branching?

src/QMCDrivers/DMC/WalkerControl.cpp
```diff
@@ -236,13 +208,15 @@ void WalkerControl::branch(int iter, MCPopulation& pop, bool do_not_branch)
     throw std::runtime_error("Potential bug! Population num_global_walkers mismatched!");
 #endif
 
+  // This defensive code implies that previous code left population walkers in invalid state.
```
ye-luo, line 211: This is not a defensive code.

src/QMCDrivers/DMC/WalkerControl.cpp
```diff
@@ -236,13 +208,15 @@ void WalkerControl::branch(int iter, MCPopulation& pop, bool do_not_branch)
...
   if (!do_not_branch)
     for (UPtr<MCPWalker>& walker : pop.get_walkers())
     {
       walker->Weight       = 1.0;
       walker->Multiplicity = 1.0;
     }
 
+  // This is debugging code, should be removed.
```
ye-luo, line 219: This not debugging code. This was used to avoid unnecessary recompute.

src/QMCDrivers/DMC/WalkerControl.cpp
```diff
@@ -407,14 +386,13 @@ void WalkerControl::swapWalkersSimple(MCPopulation& pop)
...
       newW.push_back(pop.spawnWalker());
-
       // recv the number of copies from the target
       myComm->comm.receive_n(&nsentcopy, 1, plus[ic]);
       job_list.push_back(job(newW.size() - 1, plus[ic]));
       if (plus[ic] != plus[ic + nsentcopy] || minus[ic] != minus[ic + nsentcopy])
         throw std::runtime_error("WalkerControl::swapWalkersSimple send/recv pair checking failed!");
 #ifdef MCWALKERSET_MPI_DEBUG
-      fout << "rank " << minus[ic] << " recvs a walker with " << nsentcopy << " copies from rank " << plus[ic]
+      std::cout << "rank " << minus[ic] << " recvs a walker with " << nsentcopy << " copies from rank " << plus[ic]
```
ye-luo, line 395: Your temporary change?

src/QMCDrivers/DMC/WalkerControl.cpp
```diff
@@ -462,11 +440,15 @@ void WalkerControl::swapWalkersSimple(MCPopulation& pop)
   }
   else
   {
+    // Walker::copyFromBuffer copies the walker_index_ of the sent walker and the local walker does
+    // not have a local id created.
+    // the walker on this rank.  But this replacement allows tracking this walker across the ranks.
```
ye-luo, line 445: Please improve the comment.

src/QMCDrivers/DMC/WalkerControl.cpp
```diff
@@ -475,7 +457,11 @@ void WalkerControl::swapWalkersSimple(MCPopulation& pop)
       {
         ScopedTimer local_timer(my_timers_[WC_recv]);
         myComm->comm.receive_n(awalker.DataSet.data(), byteSize, jobit->target);
-        awalker.copyFromBuffer();
+	auto walker_id = awalker.getWalkerID();
+	awalker.copyFromBuffer();
+	auto parent_id = awalker.getWalkerID();
+	awalker.setWalkerID(walker_id);
+	awalker.setParentID(parent_id);
```
ye-luo, line 464: Could you explain all the setting?
`auto parent_id = awalker.getWalkerID();` why?
Are you sure the parent id also has been wired?

src/QMCDrivers/DMC/WalkerControl.h
```diff
@@ -68,7 +68,22 @@ class WalkerControl : public MPIObjectBase
...
+   *  2. compute curData, collect multiplicity on every rank
+   *  ### fixed population
+   *  1. compute curData, collect weight on every rank
+   *  2. compute multiplicity by comb method
+   *  ---
+   *  3. figure out final distribution, apply walker count ceiling
+   *  4. collect good, bad walkers
+   *  5. communicate walkers
+   *  6. unpack received walkers, apply walker count floor
+   *  7. call MCPopulation to amplify walkers with Multiplicity > 1
```
ye-luo, line 86: Please move this back to its original location. Users calling this API doesn't need to know the implementation detail.

src/QMCDrivers/DMC/WalkerControl.h
```diff
@@ -166,7 +193,15 @@ class WalkerControl : public MPIObjectBase
   const IndexType num_ranks_;
   ///0 is default
   IndexType SwapMode;
-  ///any temporary data includes many ridiculous conversions of integral types to and from fp
+  /** large amount of state data MPI_AllReduce once per step size is LE_MAX + number of ranks.
+   *  partial enum to access elements is the enum above dumped into WalkerControl's namespace
+   *  includes many integral types converted to and from fp
+   *  \todo MPI in the 21st centure allows user defined data types that would allow all this to benefit from actual
+   *        C++ type safety and still can have performant collective operation efficiency.
```
ye-luo, line 200: MPI is C only. I don't think it knows C++ type safety. The derived type in MPI is intended for performance, not for safety. It is not useful in this particular case.

docs/developing.rst
PDoakORNL: If walker is not assigned from a walker config walker, a transfered walker, or an amplified walker, the parent_id is left as the default i.e. 0.

docs/developing.rst
PDoakORNL: yes will fix

src/QMCDrivers/DMC/WalkerControl.cpp
PDoakORNL: Ah makes sense... I will reread

src/QMCDrivers/DMC/WalkerControl.cpp
PDoakORNL: Ok so by my reading through the walker life cycle, branching, and population control walkers must always have Multiplicty == 1.0 after amplification and/or transfer. Multiplicity is set in a small number of significant locations (except for here). 

1.  On construction the default member initializer of `Walker` sets the initial value to 1.0
2.  During warmup = Multiplicity = 1.0
3.  To `static_cast<int>(walker->Weight + rng_())` resulting in a integer value of >= 0
4.  After swapping where on rank multiplicity is set to the remaining number of copies on a rank.
5. After swapping where a new walker has its multiplicity set to the number of copies received + 1 (number of copies received is counted from 0 due to lack of discrimination between counting numbers and indexing numbers).
6.  After amplificiation i.e. `copyHighMultiplicityWalkers` and `killDeadWalkers` the population will contain only walkers  of Multiplicity == 1.
7. Here where all living walkers have their multiplicity set to 1.

Several off these state transforms appear defensive in nature only, i.e. they can only cover up Walkers arriving to this function in an invalid state.

2. Can only erase a state transformation to walker that left Multiplicity in an incorrect state. Fresh walkers have multiplicity == 1 walkers from walker configuration have multiplicity == 1. An assert should be added and this side of the branch is noop.
7. If multiplicity != something is wrong, this simply covers it.

4, 5 are only necessary because of how we construct transfer jobs. Where basically walker have a multiplicity per destination for a short span of code. During this section Walker->Multiplicty is no long a simple value.

Multiplicity is a very important value for correct population control so its update and validity so reasoning about its state changes and validity should be as simple as possible.

PDoakORNL: I'm clarifying the comments, I don't want to lose this insight but it can't really be cleaned up properly until we drop the legacy drivers.

src/QMCDrivers/DMC/WalkerControl.cpp
PDoakORNL: Left some documentiation behind so this can be returned to in the ParticleSet refactoring.

src/QMCDrivers/DMC/WalkerControl.cpp
PDoakORNL: fixed

src/QMCDrivers/DMC/WalkerControl.cpp
PDoakORNL: Added test and discovered the answer was kinda.  It was for transfer but not for reading from WalkerConfigurations

src/QMCDrivers/DMC/WalkerControl.h
PDoakORNL: returned.  I left some information for this call in the header.

src/QMCDrivers/DMC/WalkerControl.h
PDoakORNL: I looks like its intended so everyone doesn't do things like this constantly.

docs/developing.rst
PDoakORNL: done

docs/developing.rst
PDoakORNL: I'm added documentation of branching.  I could not find other discussion of our branching algorithm in the manual.  Am I just missing it?

src/Particle/Walker.h
PDoakORNL: I've cleaned these docs up now.

src/QMCDrivers/DMC/WalkerControl.cpp
PDoakORNL: I have since discovered that it is also not written or read from walker hdf5 file. I'd rather address that in a future PR.

prckent:
Just looking at this after some time away...
- I wasn't expecting this to involve studying all the population control,  improving its testing and documentation, but we were going to have to do that sometime. Hopefully it will be easier to get fixed population working correctly in future :-) 
- Are we happy that with the current status we do actually have useful walker IDs?
- Given the changes here I think it would be wise to do a real DMC run and sanity check nothing has been unintentionally changed. Has anyone done this as part of testing?

PDoakORNL:
I'm working on another rebase and forced push.  This has been sitting for a week I'm not sure why anymore.  But I don't think its in the interest of completely Thrust 3 for me to continue bringing this in line with develop until there is actual interest in it again.  I will not do any work on the walkerconfiguration ouput until this PR goes in, it has enough in it, tests what it adds and adds tests and documentation that should have always been there.

PDoakORNL:
The osx failures are also all with nexus and do not appear to have anything to do with this PR.

ye-luo:
> I'm working on another rebase and forced push. 

If it is not showing conflict, there is no need to aggressively do rebasing. We rarely requires PRs being rebased before merging.

>  This has been sitting for a week I'm not sure why anymore.

Sorry I was busy and didn't pay attention to commits closely. Sometimes it is hard for me to tell if a submitter is still working on it or it is ready for review again. It is not the best use of maintainer time monitoring PR progress. If you need a specific reviewer to respond, better to ping if the PR stalls. For me specifically, feel free to @ me on github or slack me.

Unfortunately, we (submitter and reviewer) all have bandwidth limits and communication helps.

ye-luo:
Just noticed that you sent a new review request. I'm sorry that I missed it.

PDoakORNL:
I'm never sure if I don't rebase if I'm going to end up with history wise.  I guess I should do some research on this.  In this case I don't have dependent branches on this PR so it probably makes no difference.

src/Particle/Walker.h
```diff
@@ -161,10 +178,11 @@ class Walker
   }
 
   Walker(const Walker& a) : Properties(1, WP::NUMPROPERTIES, 1, WP::MAXPROPERTIES) { makeCopy(a); }
-  Walker(const Walker& a, long walker_id, long parent_id)
-      : walker_id_(walker_id), parent_id_(parent_id), Properties(1, WP::NUMPROPERTIES, 1, WP::MAXPROPERTIES)
+  Walker(const Walker& a, long walker_id, long parent_id) : Properties(1, WP::NUMPROPERTIES, 1, WP::MAXPROPERTIES)
   {
     makeCopy(a);
+    walker_id_ = walker_id;
```
PDoakORNL, line 184: Here I make these assignments after makeCopy, putting them above means they get replace by `a`'s walker_id and parent_id which is not the behavior needed or expected.

src/QMCDrivers/DMC/WalkerControl.cpp
```diff
@@ -205,26 +211,9 @@ void WalkerControl::branch(int iter, MCPopulation& pop, bool do_not_branch)
   // ranks sending walkers from other ranks have the lowest walker count now.
   untouched_walkers = std::min(untouched_walkers, walkers.size());
 
-  { // copy good walkers
+  {
     ScopedTimer copywalkers_timer(my_timers_[WC_copyWalkers]);
-    const size_t good_walkers = walkers.size();
```
PDoakORNL, line 210: This is entirely manipulation and calls that should be private to MCPopulation so I refactored it into there.  This also allows unit testing this with MCPopulation which makes sense since code can be tested without the need for the WalkerControl code.

src/QMCDrivers/DMC/WalkerControl.cpp
```diff
@@ -462,11 +461,22 @@ void WalkerControl::swapWalkersSimple(MCPopulation& pop)
   }
   else
   {
+    auto unpackWalker = [](auto& awalker) {
+      auto walker_id = awalker.getWalkerID();
+      // Walker::copyFromBuffer overwrites the walker_id
+      awalker.copyFromBuffer();
```
PDoakORNL, line 467: This lambda is just the repeated code for the synchronous and asynchronous receive below.  Previously the walker_ids and parent_itds were just getting overwritten by the sent walker_id and parent_id.  From the context of all ranks this resulted in duplicate id's all over which is not the behavior this PR is setting up. The copyFromBuffer from the current code is preserved here the new statements only effect parent_id and walker_id which are not used by the current code.

src/QMCDrivers/DMC/WalkerControl.cpp
```diff
@@ -517,7 +529,14 @@ void WalkerControl::swapWalkersSimple(MCPopulation& pop)
...
     TotalMultiplicity += good_walkers[iw]->Multiplicity;
   if (static_cast<int>(TotalMultiplicity) != fair_offset_[rank_num_ + 1] - fair_offset_[rank_num_])
-    throw std::runtime_error("Multiplicity check failed in WalkerControl::swapWalkersSimple!");
+  {
+    std::ostringstream error_msg;
+    error_msg << "Multiplicity check failed in WalkerControl::swapWalkersSimple!\n"
+              << "for rank: " << rank_num_ << " total_multiplicity: " << TotalMultiplicity
+              << "  fair_offset_[rank_num_ + 1] - fair_offset_[rank_num_]: " << fair_offset_[rank_num_ + 1] << " - "
+              << fair_offset_[rank_num_] << '\n';
+    throw std::runtime_error(error_msg.str());
```
PDoakORNL, line 538: This is just a less useless error message.  It's in an 
```
#ifndef NDEBUG
```
section so this has no effect on production science

src/QMCDrivers/MCPopulation.cpp
```diff
@@ -44,8 +44,44 @@ MCPopulation::MCPopulation(int num_ranks,
...
+      // copy the copied from walkers id to parent id.
+      walker_elements.walker.setParentID(walker_elements.walker.getWalkerID());
+      // put the walkers actual id back.
+      walker_elements.walker.setWalkerID(walker_id);
+      // fix the multiplicity of the new walker
+      walker_elements.walker.Multiplicity = 1.0;
+      // keep good walker valid.
+      good_walker.Multiplicity -= 1.0;
+      num_copies--;
+    }
```
PDoakORNL, line 78: This was the code that was in WalkerControl
```
for (size_t iw = 0; iw < good_walkers; iw++)
    {
      size_t num_copies = static_cast<int>(walkers[iw]->Multiplicity);
      while (num_copies > 1)
      {
        auto walker_elements = pop.spawnWalker();
        // save this walkers ID
        // \todo revisit Walker assignment operator after legacy drivers removed.
        // but in the modern scheme walker IDs are permanent after creation, what walker they
        // were copied from is in ParentID.
        long save_id                    = walker_elements.walker.getWalkerID();
        walker_elements.walker          = *walkers[iw];
        walker_elements.walker.setParentID(walker_elements.walker.getWalkerID());
        walker_elements.walker.setWalkerID(save_id);
        num_copies--;
      }
    }
```
The significant diferrence here is that we continue to make the multiplicities valid rather than just ignore them until the end when we set them all 1.0 regardless of what happened to them.  I'm of the opinion if you have a state variable its validity must be maintained through state changes not valid for some states and invalid in others.  That's asking for bugs later.

src/QMCDrivers/MCPopulation.cpp
```diff
@@ -91,18 +149,6 @@ void MCPopulation::createWalkers(IndexType num_walkers, const WalkerConfiguratio
 
   outputManager.resume();
 
-  int num_walkers_created = 0;
```
PDoakORNL, line 94: We don't have to wait until here to have valid walker and parent id's

PDoakORNL (commented):
I've tried to explain every code change in WalkerControl.cpp to ease concerns that any real change has been made in the DMC population control.  Should I back port the unit tests and/or right additional ones to prove the behavior is the same?

ye-luo (commented):
@PDoakORNL Thanks for the additional comments. I'd like to make the following request of changes.
1. Please revert the type change to `Multiplicity`. The purpose of this PR is to handle walker_id. It will be better to handle it separately. I also feel `Multiplicity` should be an unsigned it but all its use needs to be checked or preferably unit tested. So making a separate PR for that topics will be much better.
2. Changes around `walker.h` and `MCPopulations.h/cpp` can be placed on a separate PR. They are much less exposed to the full scheme of id handling. reviewing and merging turnaround should be fast.
3. Then we handle the last but most complicated bits in `WalkerControl`.

Sorry for the additional refactoring but that warranties stable progress.

src/Particle/Walker.h
```diff
@@ -82,38 +82,55 @@ class Walker
...
   int Age = 0;
   ///Weight of the walker
   FullPrecRealType Weight = 1.0;
-  /** Number of copies for branching
+  /** Number of replicas of this walker after branching
    * When Multiplicity = 0, this walker will be destroyed.
+   * PD: It seems to me that this should be an integer.
    */
-  FullPrecRealType Multiplicity = 1.0;
+  int Multiplicity = 1.0;
```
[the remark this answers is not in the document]
PDoakORNL: Will do

ye-luo:
superseded by https://github.com/QMCPACK/qmcpack/pull/5063 and https://github.com/QMCPACK/qmcpack/pull/5089

PR #5012, opened 2024-05-29
Abort if multideterminant expansion is empty

prckent:
## Proposed changes

Avoid a crash if the cutoff is too high and instead print a meaningful error message

- Bugfix
- New feature

- No

nitrogen2, gcc 13.2 nightly config

## Checklist


src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -1221,8 +1229,16 @@ bool SlaterDetBuilder::readDetListH5(xmlNodePtr cur,
   app_log() << "Found " << coeff.size() << " terms in the MSD expansion.\n";
```
ye-luo, line 1229: How about checking `coeff.size()` and error out if 0.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -1221,8 +1229,16 @@ bool SlaterDetBuilder::readDetListH5(xmlNodePtr cur,
...
+  auto empty = false;
+  for (auto grp = 0; grp < nGroups; grp++)
+  {
+    app_log() << "Found " << uniqueConfgs[grp].size() << " unique group " << grp << " determinants.\n";
+    if (uniqueConfgs[grp].size() == 0)
+      empty = true;
+  }
+  if (empty)
+    throw std::runtime_error(
+        "At least one group is empty with either zero determinants input or remaining after cutoff applied.");
```
ye-luo, line 1241: As long as there is one term in the MSD expansion. None of the  `uniqueConfgs[grp].size()` will be zero. Please catch the pathology in the first occurrence instead of exposing confusing implementation details to users.

ye-luo (changes requested):

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
prckent: Fair point. I tested the new code with a too large cutoff (filtering the # dets to zero), and for this case checking coeff.size() works well.

src/QMCWaveFunctions/Fermion/SlaterDetBuilder.cpp
```diff
@@ -967,13 +967,18 @@ bool SlaterDetBuilder::readDetList(xmlNodePtr cur,
     }
 
     app_log() << "Found " << coeff.size() << " terms in the MSD expansion.\n";
+
+    if (coeff.size() == 0)
+      throw std::runtime_error(
```
ye-luo, line 972: It is a uniform error and can be better handled. A lot of I/O fixes needed in this file. #5013

ye-luo (approved):

ye-luo:
Test this please

PR #5019, opened 2024-05-31
Driver-portable walker logging

jtkrogel:
## Proposed changes

This PR introduces a simplified version of the legacy TraceManager that focuses solely on writing per-walker data to HDF5 throughout the run (no streaming/listener functionality).  The new implementation, WalkerLogManager, is compatible with and implemented in both the legacy and batched drivers.

This implementation extends the original to include the ability to write full data for the highest, lowest, and median energy walkers on each rank to disk at every step.  The intention of writing this abbreviated data is to enable inexpensive access to data from problematic walkers (e.g. population explosions) in production runs.

**Capabilities**
- Writing per-walker scalar data (local energy, etc)
- (optional) writing per-walker particle data (positions, spins, gradient, laplacian)
- (optional) writing full data for min/max/median energy walkers

Below is an example generated from a short DMC run of diamond on 6 MPI ranks showing the energies of all walkers (blue), highest energy walkers (green), lowest energy walkers (red), median energy walkers (black), vs the mean energy from dmc.dat (magenta line):

walker_traces

**Design**

- Main classes: WalkerLogManager, WalkerLogCollector, WalkerLogBuffer
- WalkerLogManager
  - Driver-level resource 
  - Responsible for HDF file I/O
  - Finds min/max/median energy walker data
  - Writes walker data buffers from each crowd to HDF at the end of each block
- WalkerLogCollector
  - Crowd-level resource 
  - Collects per-walker data during MC steps and stores it in local data buffers
- WalkerLogBuffer
  - Acts as a dynamic buffer for walker data collected during a block
  - Resource allocation is first-touch

**Tests**

- Legacy and batched driver deterministic tests
- 1-mpi/1-thread and 4-mpi/4-thread versions
- VMC and DMC
- The tests pass if the walker logging data can be used to reconstruct the information outputted to scalar.dat (VMC) or dmc.dat (DMC)

- New feature

- No

Laptop

## Checklist

## Path out of WIP

- [x] Manual entry
- [x] In-line documentation
- [x] Batched driver naming (Peter)
- [x] Input section ownership at driver-level (Peter)
- [x] WalkerLogCollector::collect() folded into Crowd (Peter)
- [x] Use of const in WalkerLogCollector::collect (Ye) 
- [x] Naming conventions/rideTheCamel (Ye)
- [x] WalkerLogBuffer and WalkerLogCollector in separate h/cpp files (Ye) 
- [x] Vacuous class for types, enum for const size and index names (Ye)
- [x] WalkerLogCollector unit test, collect method (Ye)

## Tasks for review round 3

- [x] Make step counter seem like a global data member
- [x] Long names for crowd member functions
- [x] Make more getters()
- [x] Inline default values for WalkerLogState
- [x] Try higher level api for hdf (results in nan)

## Elements for reviewers to review

- WalkerLog* classes
- WalkerLogCollector unit test
- Driver integration


src/Estimators/WalkerTraceManager.h
```diff
@@ -0,0 +1,465 @@
...
+  inline void reset_collect()
+  {
+    if(quantity_pointer!=quantity_info.size())
+      throw std::runtime_error("WalkerTraceBuffer quantity_pointer has not been moved through all quantities prior during collect() call.");
+    first_collect = false;
+    quantity_pointer = 0;
+  }
+
+
+  inline bool same_as(WalkerTraceBuffer<T>& ref)
```
ye-luo, line 119: const function should be marked const.

src/Estimators/WalkerTraceManager.h
```diff
@@ -0,0 +1,465 @@
...
+
+  void check_buffers();
+
+private:
+  void reset_buffers();
+};
+
+
+
+class WalkerTraceManager
```
ye-luo, line 402: This is the third full class in the source file. Please make `WalkerTraceBuffer` and `WalkerTraceCollector` into separate files and properly unit tested.

ye-luo (changes requested):
Please make WalkerTraceBuffer and WalkerTraceCollector into separate files (h and cpp) and properly unit tested in a separate PR. Doing this definitely helps fast turnaround reviewing and merging.

src/Estimators/WalkerTraceManager.h
```diff
@@ -0,0 +1,465 @@
...
+
+struct WalkerQuantityInfo
+{
+  std::string name;
+  size_t dimension;
+  size_t size;
+  size_t unit_size;
+  TinyVector<size_t, WDMAX> shape;
+  size_t buffer_start;
+  size_t buffer_end;
```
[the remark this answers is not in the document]
jtkrogel: Note the WIP todo list above.

src/Estimators/WalkerTraceManager.h
```diff
@@ -0,0 +1,465 @@
...
+#include "hdf/hdf_archive.h"
+
+#include <unordered_set>
+
+
+
+namespace qmcplusplus
+{
+
+const unsigned int WDMAX = 4;
```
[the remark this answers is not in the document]
jtkrogel: How is an enum better?  This constant represents an actual maximum dimension supported by the code.

src/Estimators/WalkerTraceManager.h
```diff
@@ -0,0 +1,465 @@
...
+using WTraceInt          = long;
+using WTraceReal         = OHMMS_PRECISION_FULL;
+using WTraceComp         = std::complex<WTraceReal>;
+#ifndef QMC_COMPLEX
+using WTracePsiVal = WTraceReal;
+#else
+using WTracePsiVal = WTraceComp;
+#endif
```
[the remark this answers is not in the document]
jtkrogel: So you want just these types wrapped in a vacuous class and moved into separate file, correct?

src/Estimators/WalkerTraceManager.h
```diff
@@ -0,0 +1,465 @@
...
+
+  WalkerQuantityInfo(const std::string& name_,size_t unit_size_,size_t buffer_start_,size_t n1=1,size_t n2=0,size_t n3=0,size_t n4=0)
+  {
+    name         = name_;
+    unit_size    = unit_size_;
+    buffer_start = buffer_start_;
+    shape[0]     = n1;
+    shape[1]     = n2;
+    shape[2]     = n3;
+    shape[3]     = n4;
```
ye-luo, line 59: I don't get what are 0, 1, 2, 3. You can probably give a name to each index.
```
enum
{ DIM_0_NAME = 0,
  DIM_1_NAME,
  DIM_2_NAME,
  DIM_3_NAME,
  WDMAX }
```
WDMAX will be set 4 by the compiler.

src/Estimators/WalkerTraceManager.h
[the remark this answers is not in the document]
ye-luo: https://github.com/QMCPACK/qmcpack/pull/5019/files#r1622971831

src/Estimators/WalkerTraceManager.h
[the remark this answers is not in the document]
ye-luo: You can create a WalkerTraceDataTypes class. WalkerTraceDataTypes and WalkerQuantityInfo probabaly don't need their own files but can stay inside WalkerTraceBuffer.h. I do need WalkerTraceBuffer and WalkerTraceCollector in individual files.

src/Estimators/WalkerTraceManager.h
jtkrogel: They don't have names, but I can give them names if you prefer.  They literally are dimension 0, 1, 2, 3 of an array.

src/Estimators/WalkerTraceManager.h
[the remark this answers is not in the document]
jtkrogel: Done

src/Estimators/WalkerTraceManager.h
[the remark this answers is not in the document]
jtkrogel: Done

src/Estimators/WalkerTraceManager.h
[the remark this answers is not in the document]
jtkrogel: Done

src/Estimators/WalkerTraceManager.h
```diff
@@ -0,0 +1,465 @@
...
+      }
+    buffer_end = buffer_start + size*unit_size;
+  }
+};
+
+
+
+
+template<typename T>
+struct WalkerTraceBuffer
```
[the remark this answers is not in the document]
jtkrogel: Done

src/Estimators/WalkerTraceManager.h
```diff
@@ -0,0 +1,465 @@
...
+    label             = "?";
+    verbose           = false;
+    first_collect     = true;
+    walker_data_size  = 0;
+    quantity_pointer  = 0;
+    reset_buffer();
+  }
+
+
+  inline void reset_buffer()
```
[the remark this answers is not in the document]
jtkrogel: Done

src/Estimators/WalkerTraceManager.h
jtkrogel: Done

src/Estimators/WalkerTraceManager.h
jtkrogel: Done

src/Estimators/WalkerTraceManager.h
jtkrogel: Done

jtkrogel:
NB: I will fix the failing unit test (complex estimators) in 30 min.

jtkrogel:
Hold on merge for naming discussion for now.  Will commit a naming change if the decision is made today.

PDoakORNL:
There is a great deal to cover here.  I'm just starting on the review.  Please introduce one piece at a time from the most detailed up.  I'd start with trace buffer and its unit tests.  It needs a review on its own.

prckent (dismissed):
Putting a note to say:

(1) thanks for this, very much needed. e.g. provides the feature that was being worked on in https://github.com/QMCPACK/qmcpack/pull/4940 but also for the batched drivers.

(2) after some Slack discussions, this feature will be renamed "walker logging" to create "walker logs" etc. since that is clearer and easier to understand/find than "walking tracing".

jtkrogel:
Renaming applied.  Ready to proceed.

ye-luo (changes requested):
Please create **a new PR** with WalkerLogBuffer WalkerLogCollector and WalkerLogManager and their unit tests under QMCDrivers. Once that PR is merged, we can update this PR with much less stuff to go through in review.

PDoakORNL:
That would be a big help to me as well.  I have suggestions/changes needed for the lower level classes which would be better to address and test at that level.  I.e. use of lower level hdf5 functions when there is already a way to easily pass containers.

Other issues relate the integration with existing higher level classes which are largely separate.

jtkrogel:
See #5035

src/QMCDrivers/QMCDriverFactory.cpp
```diff
@@ -294,6 +295,11 @@ std::unique_ptr<QMCDriverInterface> QMCDriverFactory::createQMCDriver(xmlNodePtr
       (das.traces_tag == "none" && (das.new_run_type == QMCRunType::VMC || das.new_run_type == QMCRunType::DMC));
   new_driver->requestTraces(allow_traces);
 
+  //add trace information
+  bool allow_walker_logs = das.walkerlogs_tag == "yes" ||
+      (das.walkerlogs_tag == "none" && (das.new_run_type == QMCRunType::VMC || das.new_run_type == QMCRunType::DMC || das.new_run_type == QMCRunType::VMC_BATCH || das.new_run_type == QMCRunType::DMC_BATCH));
+  new_driver->requestWalkerLogs(allow_walker_logs);
```
ye-luo, line 301: Complicated control logic. Prefer not to have another switch. I would like to have the following.
if "walkerlog" appears globally, all the drivers honor it. if not, each driver honor the "walkerlog" within the driver input section.

src/QMCDrivers/Crowd.h
```diff
@@ -83,6 +85,9 @@ class Crowd
     estimator_manager_crowd_.accumulate(mcp_walkers_, walker_elecs_, walker_twfs_, walker_hamiltonians_, rng);
   }
 
+  /// Collect walker log data
+  void collect(int current_step);
```
ye-luo, line 89: Please call it collectStepWalkerLog. We may adjust the name if we expand its use.

src/QMCDrivers/Crowd.h
```diff
@@ -139,6 +144,8 @@ class Crowd
   unsigned long n_accept_          = 0;
   unsigned long n_nonlocal_accept_ = 0;
   /** @} */
+public:
+  WalkerLogCollector wlog_collector_;
```
ye-luo, line 148: Please make it private and add necessary API for accessing it.

src/QMCDrivers/DMC/DMCBatched.cpp
```diff
@@ -429,6 +432,14 @@ bool DMCBatched::run()
   IndexType num_blocks = qmcdriver_input_.get_max_blocks();
 
   estimator_manager_->startDriverRun();
+
+  //start walker log manager
+  wlog_manager_ = std::make_unique<WalkerLogManager>(walker_logs_input, allow_walker_logs, get_root_name(), myComm);
+  std::vector<WalkerLogCollector*> wlog_collectors;
+  for (auto& c: crowds_)
+    wlog_collectors.push_back(&c->wlog_collector_);
+  wlog_manager_->startRun(wlog_collectors);
```
ye-luo, line 441: a vector of raw pointers is a big no. Crowds own collectors and thus, RefVector is a better way of expressing the ownership.
```
RefVector<<WalkerLogCollector> wlog_collectors;
```

ye-luo: Also it is better to make wlog_manager_ accepts wlog_collectors via constructor and save it. Then later writeBuffers doesn't need arguments.

ye-luo: This change can be a bit involving. Defer it to a later PR.

ye-luo:
I have merged in develop and fixed cmake issues. Please pull before making further changes.

src/QMCDrivers/QMCDriverFactory.cpp
prckent: Agreeing with Ye here - we use similar logical for maxcpusecs

jtkrogel: The control logic protects drivers which either don't support walker logging, or shouldn't.

Traces were never enabled for optimizers, for instance.  On another note, I really don't want to be forced to write code to support walker logging in *all* of the legacy drivers.  CSVMC anyone?

src/QMCDrivers/Crowd.h
jtkrogel: Isn't it redundant to repeat the class name in a member function name?

jtkrogel: Seems redundant to include the class name in the name of one of its member functions.  Will do it.

src/QMCDrivers/QMCDriverFactory.cpp
prckent: Fair point

jtkrogel: Further comments welcome.

src/QMCDrivers/DMC/DMCBatched.h
```diff
@@ -53,6 +53,7 @@ class DMCBatched : public QMCDriverNew
     IndexType recalculate_properties_period;
     const size_t steps_per_block;
     IndexType step = -1;
+    IndexType current_step = -1;
```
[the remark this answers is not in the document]
jtkrogel: Sure

src/QMCDrivers/DMC/DMCBatched.cpp
jtkrogel: (Nothing to change here)

src/QMCDrivers/DMC/DMCBatched.h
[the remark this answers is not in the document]
jtkrogel: Done

src/QMCDrivers/Crowd.h
jtkrogel: Sure

jtkrogel: Done

src/QMCDrivers/Crowd.h
jtkrogel: Done.

jtkrogel:
I've made the requested changes, unless there are more comments...

src/QMCDrivers/QMCDriverFactory.cpp
ye-luo: I was talking about the control logic from the input switch perspective not about implementations. The driver support is expected to be sparse. I thought if a driver doesn't support logging, then certain functions are just no-op.

Still hard to understand the intention of
```
  bool allow_walker_logs = das.walkerlogs_tag == "yes" ||
      (das.walkerlogs_tag == "none" && (das.new_run_type == QMCRunType::VMC || das.new_run_type == QMCRunType::DMC || das.new_run_type == QMCRunType::VMC_BATCH || das.new_run_type == QMCRunType::DMC_BATCH));
```
if I set attribute to "none", It seems that I still get logging when the driver is VMC for example.
If I set it to "yes", then CSVMC will crash? I don't see any "protection".

jtkrogel:
I've addressed Peter's comment from #5035 requesting simplification of WalkerLogState.

jtkrogel:
I've also tried Peter's suggestion to use `f.writeEntry(buffer, "data")` in `WalkerLogBuffer::writeHDF`.  Making that replacement resulted in NaN's being written to the HDF files (picked up by integration test).  Therefore I've kept the original implementation.

jtkrogel:
Comments addressed.  Ready to proceed.

src/QMCDrivers/QMCDriverFactory.cpp
jtkrogel: If you want, this could be changed to:

```
bool allow_walker_logs = das.new_run_type == QMCRunType::VMC || das.new_run_type == QMCRunType::DMC || das.new_run_type == QMCRunType::VMC_BATCH || das.new_run_type == QMCRunType::DMC_BATCH);
```

The `das.walkerlogs_tag == "yes"` part just allows one to selectively turn the feature on/off for certain drivers, if desired (user flexibility).

jtkrogel: Slightly better logic (move the gun away from the user's foot if they intentionally point it there):

```
bool allow_walker_logs = das.walkerlogs_tag == "yes" && (das.new_run_type == QMCRunType::VMC || das.new_run_type == QMCRunType::DMC || das.new_run_type == QMCRunType::VMC_BATCH || das.new_run_type == QMCRunType::DMC_BATCH));
```
With `das.walkerlogs_tag="yes"` being the default.

ye-luo: I would like to get rid of `requestWalkerLogs` completely.

Derived class overriding function putWalkerLogs knows what the driver is.
Then its argument also indicates weather logging should be on or off. I don't see the need of additional control.

ye-luo: The current walker logging support is limited to VMC/DMC and its functionality and implementation can be completely different for a non-VMC/DMC driver. So its input is more suited for a driver input section rather than globally. If only some VMC/DMC drivers need to turn on logging, the input should be in the driver input section. Right now, we provide convenience to have the input globally and I believe it is better not to another switch to turn it off locally. To me, that is a bad design logic.

jtkrogel: As for the other protection, setting `allow_walker_logs` correctly to `false` for unsupported drivers is what makes the `WalkerLogManager`/`WalkerLogCollector` operations a no-op.

ye-luo: Then default value of `allow_walker_logs` is already `false` to all the unsupported drivers. Why do we even need to set it.

ye-luo: Too much details. I would like to discuss the user-facing input logic first and then we can address the implementation issue, maybe at a later PR.

Are you OK with the following input logic
1.  if the global `<walkerlog/>` exists, turn on logging in all the supported drivers.
2.  when the global <walkerlog> doesn't exist, legacy drivers do nothing. batched drivers will look for driver scope `<walkerlog/>` and act on it.

ye-luo:
> I've also tried Peter's suggestion to use `f.writeEntry(buffer, "data")` in `WalkerLogBuffer::writeHDF`. Making that replacement resulted in NaN's being written to the HDF files (picked up by integration test). Therefore I've kept the original implementation.

You need h5d_append and we don't have a high-level API in hdf_archive right now. added issue #5038

ye-luo (approved):
Still quite a few fixes are needed but it is not necessary to hold the PR.

ye-luo:
Test this please

src/QMCDrivers/Crowd.h
ye-luo: it is not obfuscating. Making it public, it can be touched potentially anywhere. I just found out that it is not initialized using "state" in the crowd.

src/QMCDrivers/DMC/DMCBatched.cpp
```diff
@@ -429,6 +432,14 @@ bool DMCBatched::run()
   IndexType num_blocks = qmcdriver_input_.get_max_blocks();
 
   estimator_manager_->startDriverRun();
+
+  //start walker log manager
+  wlog_manager_ = std::make_unique<WalkerLogManager>(walker_logs_input, allow_walker_logs, get_root_name(), myComm);
```
PDoakORNL, line 437: The idea of the log manager being a run scope object is promising but then it should not be a member variable.

PDoakORNL (changes requested):
I think we need to hold some sort of line to model good simulation object life cycle control and input handling for the batched code.

src/QMCApp/QMCMain.h
```diff
@@ -75,6 +75,9 @@ class QMCMain : public MPIObjectBase, public QMCAppBase
   ///xml mcwalkerset read-in elements
   std::vector<xmlNodePtr> walker_set_in_;
 
+  ///walkerlogs xml
+  xmlNodePtr walker_logs_xml_;
```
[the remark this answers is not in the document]
ye-luo: I also don't like storing xml, parsing the xml in-place and store the result seems cleaner. We can make some improvements with a later PR.

src/QMCDrivers/DMC/DMCBatched.cpp
```diff
@@ -500,6 +513,7 @@ bool DMCBatched::run()
       if (qmcdriver_input_.get_measure_imbalance())
         measureImbalance("Block " + std::to_string(block));
       endBlock();
+      wlog_manager_->writeBuffers(wlog_collectors);
```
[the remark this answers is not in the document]
ye-luo: We cannot do that. recordBlock is in QMCDriverNew base class but wlog_manager_ only exists in VMC and DMC drivers. We may reconsider the optimal location of  recordBlock.

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -304,6 +309,12 @@ bool VMCBatched::run()
   IndexType num_blocks = qmcdriver_input_.get_max_blocks();
   //start the main estimator
   estimator_manager_->startDriverRun();
+  //start walker log manager
+  wlog_manager_ = std::make_unique<WalkerLogManager>(walker_logs_input, allow_walker_logs, get_root_name(), myComm);
```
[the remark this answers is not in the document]
ye-luo: It should not be constructed in the constructor. I have moved this line to process() in https://github.com/QMCPACK/qmcpack/pull/5039 at the moment. When the global input doesn't exist, we may look it up locally.

src/QMCDrivers/DMC/DMCBatched.cpp
ye-luo: Good point. We may get a chance to make it a run scope object after restructuring certain bits. Will give it a try once #5039 got merged.

jtkrogel:
Hi guys.  I had thought this was going in given Ye's approval.  

If there really is a *short* list of "must have" items still remaining for this PR after all these passes, please make a short, final list below and I will address them. 

I need to get past this point and get back to other work, some of which depends on this code being in place.

jtkrogel:
Hi guys.  I had thought this was going in given Ye's approval.  

If there really is a *short* list of "must have" items still remaining for this PR after all these passes, please make a short, final list below and I will address them. 

I need to get past this point and get back to other work, some of which depends on this code being in place.

ye-luo (approved):
I don’t feel there are blocking issues although many improvements are still needed.

src/QMCHamiltonians/OperatorBase.cpp
```diff
@@ -268,7 +268,6 @@ void OperatorBase::deleteTraceQuantities()
   have_required_traces_ = false;
   request_.reset();
 }
-
```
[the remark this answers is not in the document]
PDoakORNL: Adding an unecessary file to the PR file list for white space change has never been allowed.  I have had to fix such changes many times. Finding these sort of typos takes effort by reviewers and they stack up over time.

src/QMCDrivers/VMC/VMCBatched.cpp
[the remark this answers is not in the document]
PDoakORNL: Process is also an API element that only exists for legacy compatibility, I'd be very suprised if construction actually needed to delayed until then, but lets discuss later.

PDoakORNL:
Reverting the whitespace OperatorBase.cpp change to reduce the number files is the only must have.

PDoakORNL (approved):

ye-luo:
Test this please

PR #5022, opened 2024-06-01
[AFQMC] only use afqmc-device copy overload if semantically correct

correaa:
(e.g. A is not a fancy pointer -- in continuous memory); and if not, there will be other candidates, possibly provided by Multi itself (e.g. through ADL priority)

AFQMC used to provide its own fancy memory specializations.
Some of these function do not work for iterators that are not fancy pointers themselves, giving a hard error.

## Proposed changes

Use decltype to remove the overload if not semantically correct

- No

Ubuntu 22.04 and CUDA 12

## Checklist


ye-luo:
Test this please

ye-luo (approved):

PR #5029, opened 2024-06-04
Nexus: Implementation of GCTA with (spin)-adapted Fermi levels

aannabe:
## Proposed changes

This PR implements the grand-canonical twist averaging using (spin)-adapted Fermi levels [GCTA-(S)AFL] as described in:
https://pubs.acs.org/doi/10.1021/acs.jctc.4c00058

The implementation introduces a new `GCTA` class, which stores most of the functionality and data in the class instance.   To avoid interference with the `Qmcpack` class, deep-copied attributes of the `Qmcpack` object are passed into the `GCTA` instance.

The user-facing keyword is simple: `gcta = ...` in `generate_qmcpack()`, which activates GCTA occupations. The `gcta` keyword supports four possible arguments: `('safl', 'afl', 'nscf', 'scf')`. Currently, `pwscf` code interface has full support. Other codes can be readily supported by introducing functions that can read the eigenvalues and k-point data.

To verify the implementation, I compared the resulting twist occupations to those obtained by the independent workflow used in the above article. The twist occupations of Al and Fe at primitive cells and supercells showed the same GCTA occupations using the old implementation versus the new implementation in Nexus. Also, primitive cell runs of Fe showed the same VMC energy as in the above article.

The tested cases of Al and Fe are both cubic systems. More production tests are needed with non-cubic systems. In general, a drop in twist-averaged energy is expected for dense twist-mesh when compared to the case without the use of `gcta` keyword.

A Nexus example was added that discusses the case of BCC Fe. If needed, more thorough documentation could be added to Nexus docs. This PR also lacks Nexus tests, which we plan to add in the future. All existing Nexus tests are passing.

- New feature

- No

Laptop, Perlmutter

## Checklist
- Documentation and tests will be added in another PR to keep the size managable.


jtkrogel:
I will review this later today.

nexus/lib/qmcpack.py
```diff
@@ -41,16 +41,524 @@
...
+            data        = data,
+            )
+        self.eig_data = res
+    #end def read_eshdf_data
+
+    def unfolded_nelecs(self):
+        '''
+        Returns the number of electrons in the primitive cell
+        '''
+        if self.system.folded_system == None:
```
jtkrogel, line 172: Change to "is None".  Using "==" actually can cause problems for certain types, "is" is safer.

nexus/lib/qmcpack.py
```diff
@@ -41,16 +41,524 @@
...
+        return nelecs
+    #end def unfolded_nelecs
+
+    def unfolded_nkpoints(self):
+        '''
+        Returns the number of unsymmetrized k-points when a supercell is unfolded back to the primitive cell
+        '''
+        kgrid = array(self.system.generation_info.kgrid)
+        nkgrid = np.prod(kgrid)
+        if self.system.folded_system == None:
```
jtkrogel, line 189: "is None"

nexus/lib/qmcpack.py
```diff
@@ -41,16 +41,524 @@
...
+        #end if
+        nkpoints = round(nkgrid * ntile)
+        return nkpoints
+    #end def unfolded_nkpoints
+
+    def prim_kpoints(self):
+        '''
+        Returns the k-points used to build the supercell in unit coordinates
+        '''
+        if self.system.folded_system == None:
```
jtkrogel, line 203: "is None"

jtkrogel (changes requested):
This looks solid.  Bonus points for the tutorial-length example!

Just change the couple of "==None" to "is None" and I will approve.

jtkrogel (approved):
LGTM.

prckent:
Test this please

prckent:
Test this please

PR #5030, opened 2024-06-04
Allow directly select openmp offload code path.

ye-luo:
## Proposed changes
Somehow this change should have been introduced when DiracDeterminantBatched was added.
`<slaterdeterminant gpu="omptarget">` can be used to hit the offload code path in a CUDA+OpenMP offload build. The CUDA one is default.

- New feature

- No

Polaris

## Checklist

### DIFF
--- a/docs/intro_wavefunction.rst
+++ b/docs/intro_wavefunction.rst
@@ -798,7 +798,7 @@ Additional information:
   The best ``delay_rank`` depends on the processor microarchitecture.
   GPU support is under development.
 
-- ``gpu`` This option is only effective when GPU features are built. Use the implementation with GPU acceleration if ``yes``.
+- ``gpu`` This option is only effective when GPU features are built. Default to using GPU. "omptarget", "cuda", "sycl", "cpu" can be set to target a specific implementation, "yes", "no" can be used to toggle on or off GPU acceleration.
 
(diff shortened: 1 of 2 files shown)


ye-luo:
Test this please

prckent (commented):
Can you please update the docs?

How would a user learn which options are supported?

ye-luo:
Test this please

prckent (commented):
Likely for later PRs:

Can you see a place to print the available candidate values in standard output?
e.g. " Using "omptarget" implementation. "cpu", "sycl" also available. "
In this or a later PR it would be good to note+ensure, e.g.  "yes" defaults to the optimal one. I think this would be a good pattern to repeat for all the Wavefunction Components.

ye-luo:
> Can you see a place to print the available candidate values in standard output?
e.g. " Using "omptarget" implementation. "cpu", "sycl" also available. "

What you suggested seems to complicating our printout. Users are not expected to select specific implementation but relying on default. So they don't need to see additional info in the standard output.
In the manual, I'm thinking of documenting the general behavior of `gpu` option in the GPU section and then having each occurrence pointing to it. Probably a light task on a boring day.

prckent (approved):

PR #5031, opened 2024-06-05
Move hdf5 suppressor into file scope.

ye-luo:
## Proposed changes
No need to add it to the "main" of every executable and simplify linking. A unit test that doesn't use hdf5 doesn't need to link hdf5.

- No

epyc-server

## Checklist


prckent:
Test this please

PR #5039, opened 2024-06-11
Improve the integration of WalkerLog classes in batched drivers

ye-luo:
## Proposed changes
Made the following changes.
1. WalkerLogCollector now captures 'state' during construction. No more copy and update. This modification also requires WalkerLogCollector created via WalkerLogManager which provides 'state'.
2. WalkerLogManager acquires a list of references of WalkerLogCollector from crowds at startRun and releases the reference at stopRun. Not using pointers directly any more 
3. wlog_collector_ are uniuqe_ptr in Crowd. I would like to eventually moving away it way from state.logs_active a deeply hidden state.
4. wlog_manager is a function scoped object now.

- No

epyc-server

## Checklist


PDoakORNL (changes requested):
Efficient improvement.  The .release() into the legacy drivers is a nice trick. Think this just needs the noted. documentation addition.

src/QMCDrivers/Crowd.h
```diff
@@ -118,6 +119,8 @@ class Crowd
 
   const MultiWalkerDispatchers& dispatchers_;
 
+  static RefVector<WalkerLogCollector> getWalkerLogCollectorRefs(const UPtrVector<Crowd>& crowds);
```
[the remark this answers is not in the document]
ye-luo: When using a container, it is common practice to test if it is empty before use. Adding the `std::optional` is unnecessary here. I will add comments to make it clear.

PDoakORNL (dismissed):

ye-luo:
@jtkrogel could you review?

src/QMCDrivers/VMC/VMCBatched.cpp
```diff
@@ -449,8 +450,8 @@ bool VMCBatched::run()
 
   print_mem("VMCBatched ends", app_log());
 
+  wlog_manager.stopRun();
```
jtkrogel, line 453: Not really an issue, but is there a reason for this change?

src/QMCDrivers/WalkerLogManager.cpp
```diff
@@ -68,28 +68,25 @@ WalkerLogManager::WalkerLogManager(WalkerLogInput& inp, bool allow_logs, std::st
...
 }
 
 
-void WalkerLogManager::startRun(std::vector<WalkerLogCollector*>& collectors)
+void WalkerLogManager::startRun(RefVector<WalkerLogCollector>&& collectors)
 {
+  collectors_in_run_ = std::move(collectors);
   if (!state.logs_active)
     return; // no-op for driver if logs are inactive
   if (state.verbose)
```
jtkrogel, line 84: A check (+throw) should be made here for non-empty `collectors` vector.  It is a contradiction for `state.logs_active` and `collectors.size()==0` to be true simultaneously.

jtkrogel (changes requested):
Just one addition requested above (logic check).

Overall, the change to a "capture" style for `collectors` makes the WalkerLogManager function API ambiguous due to semi-ownership of the collectors.  Functions accepting collectors or not seems like an arbitrary choice, potentially raising questions as to whether the collectors passed in to a member function should ever differ from the captured set (they shouldn't).  Despite this, I don't think its a blocking issue.

ye-luo:
writebuffer requires the same set of collectors as startRun. Since the code has be changed to capture the set of references at the startRun, writebuffer doesn't need to take the set again as argument any longer.
startRun stopRun and writeBuffer are the only three public APIs. With PR, only startRun takes a set of references to collectors. So less error-prone.
Regarding private functions, I tried to leave the set of references to collectors as an argument if it helps making the function const. Such functions are more like free functions preferably not pulling member variables directly for exmaple checkCollectors. Regarding the other private member functions openFile openHDFFile, I feel there is some design but the implementation is incomplete and thus leave them as they are.

src/QMCDrivers/VMC/VMCBatched.cpp
ye-luo: Driven by https://github.com/QMCPACK/qmcpack/pull/5019/files#r1636930622

src/QMCDrivers/WalkerLogManager.cpp
ye-luo: added

jtkrogel (approved):
Good enough.

PR #5042, opened 2024-06-13
Add syclBLAS gemv_batch and ger_batch

ye-luo:
## Proposed changes
The gemv_batch of oneMKL doesn't meet our requirement that alpha and beta are vectors. Add our version like cuBLAS_missing_functions.
These are makeshift versions lacking tuning.

## What type(s) of changes does this code introduce?
- New feature

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
sunspot

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted

PR #5043, opened 2024-06-14
Enable DiracDeterminantBatched<PlatformKind::SYCL>

ye-luo:
Review after #5042 being merged.
## Proposed changes
@jngkim contributed `QMCWaveFunctions/detail/SYCL/matrix_update_helper.cpp`
Currently there is no batched code path for the matrix inversion and I just connect the OMPTarget based class which runs inversion on host.

- New feature

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent:
Test this please

prckent (approved):
Many thanks Ye and @jngkim 

Any issues found with GPU inversion or is that the next step?

ye-luo:
> Many thanks Ye and @jngkim
> 
> Any issues found with GPU inversion or is that the next step?

I would like to refactor all the GPU inversion classes before adding one for MKL. We currently choose either one-by-one or all batched based on the delayed update class choice. Two issues
1. all batched routine may need a lot of scratch space. We can probably subdivide a batch to blocks and do one block at a time. So enable certain degree of flexibility.
2. The host inverter and gpu inverter classes are integrated in to the DDB separately but we only choose one. I would like to do better C++ and hide the detail of inverters at the DDB level and also improve the APIs.

PR #5063, opened 2024-06-26
MCPopulation now gives walkers useful walker and parent ids

PDoakORNL:
## Proposed changes
segment of #5010
`Walker`
* Fix bug with Walker constructor that ignored walker_id, parent_id arguments.
* Add useful documentation to Walker.h

`MCPopulation`
* Refactor into MCPopulation the fission of high multiplicity walkers into 2 or more walkers.
* Generate walker_ids for walkers for each trajectory they start. (i.e. a life)
* Assign parent id of new walkers in useful way.
* Expand in code documentation.

- Bugfix
- New feature

There is should be no user observable change except due to #5019 where walker_id's will be different but still undependable.

- No

OSX laptop
linux x86 vm

## Checklist

### DIFF
--- a/src/Particle/Walker.h
+++ b/src/Particle/Walker.h
@@ -2,7 +2,7 @@
 // This file is distributed under the University of Illinois/NCSA Open Source License.
 // See LICENSE file in top directory for details.
 //
-// Copyright (c) 2016 Jeongnim Kim and QMCPACK developers.
+// Copyright (c) 2024 QMCPACK developers.
 //
 // File developed by: D. Das, University of Illinois at Urbana-Champaign
 //                    Ken Esler, kpesler@gmail.com, University of Illinois at Urbana-Champaign
@@ -11,6 +11,7 @@
 //                    Jeongnim Kim, jeongnim.kim@gmail.com, University of Illinois at Urbana-Champaign
 //                    Raymond Clay III, j.k.rofling@gmail.com, Lawrence Livermore National Laboratory
 //                    Ye Luo, yeluo@anl.gov, Argonne National Laboratory
+//                    Peter W. Doak, doakpw@ornl.gov, Oak Ridge National Laboratory
 //                    Mark A. Berrill, berrillma@ornl.gov, Oak Ridge National Laboratory
 //                    Jeongnim Kim, jeongnim.kim@intel.com, Intel Corp
 //
@@ -82,38 +83,55 @@ public:
   /** }@ */
 
 private:
-  /** in legacy the ancients have said only:
-   *  id reserved for forward walking
+  /** walker identifier during a QMCSection
+   *
+   *  batched:
+   *  Only MCPopulation should set A living walker will have a WalkerID > 0
+   *  0 is the value of a default constructed walker.
+   *  Any negative value must have been set by an outside entity and indicates
+   *  an invalid walker ID.
    */
   long walker_id_ = 0;
-  /** in legacy the ancients have said only:
-   *  id reserved for forward walking
+  /** walker identifier that provided the initial state of walker
+   *
+   *  batched:
+   *  default constructed = 0;
+   *  parentID > 0 it is the walkerID in this section it was assigned from
+   *  parentID < 0 it is the  walkerID of a walker in a WalkerConfiguration
+   *  used to construct an initial population of walkers.
    */
   long parent_id_ = 0;
 
 public:
-  /** allegedly DMCgeneration
+  /** allegedly DMC generation
    *  PD: I can find no evidence it is ever updated anywhere in the code.
    */
   int Generation = 0;
-  ///Age of this walker age is incremented when a walker is not moved after a sweep
+  ///Age is incremented when a walker is not moved after a sweep
   int Age = 0;
   ///Weight of the walker
   FullPrecRealType Weight = 1.0;
-  /** Number of copies for branching
+  /** Number of replicas of this walker after branching
    * When Multiplicity = 0, this walker will be destroyed.
+   * PD: It seems to me that this should be an integer.
    */
   FullPrecRealType Multiplicity = 1.0;
   /// mark true if this walker is being sent.
   bool SendInProgress;
-  /// if true, this walker is either copied or tranferred from another MPI rank.
+
+  /** if true, this walker is either a copy to lower multiplicity or tranferred from another MPI rank.
+   *  This walker will need distance table, jastrow factors, etc recomputed.
+   *  So this is really a variable tracking the "cache" of the ParticleSet, etc.
+   *  \todo this is a smell and would be nice to address in ParticleSet refactoring.
+   */
   bool wasTouched = true;
 
   /** The configuration vector (3N-dimensional vector to store
      the positions of all the particles for a single walker)*/
   ParticlePos R;
 
-  //Dynamical spin variable.
+  /** Spin configuration vector (size N)
+   *  i.e. Dynamical spin variable. */
   ParticleScalar spins;
 #if !defined(SOA_MEMORY_OPTIMIZED)
   /** \f$ \nabla_i d\log \Psi for the i-th particle */
@@ -161,10 +179,13 @@ public:
   }
 
   Walker(const Walker& a) : Properties(1, WP::NUMPROPERTIES, 1, WP::MAXPROPERTIES) { makeCopy(a); }
-  Walker(const Walker& a, long walker_id, long parent_id)
-      : walker_id_(walker_id), parent_id_(parent_id), Properties(1, WP::NUMPROPERTIES, 1, WP::MAXPROPERTIES)
+  Walker(const Walker& a, long walker_id, long parent_id) : Properties(1, WP::NUMPROPERTIES, 1, WP::MAXPROPERTIES)
   {
     makeCopy(a);
+    // makeCopy replaces walker_id_ and parent_id with a.walker_id_  ...
+    // so these can not be set in the contructor initializer list
+    walker_id_ = walker_id;
+    parent_id_ = parent_id;
   }
 
   /** create a valid walker for n-particles (batched version)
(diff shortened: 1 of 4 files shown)


src/QMCDrivers/MCPopulation.cpp
```diff
@@ -156,27 +210,27 @@ WalkerElementsRef MCPopulation::spawnWalker()
     walkers_.back()->Age                = 0;
     walkers_.back()->Multiplicity       = 1.0;
     walkers_.back()->Weight             = 1.0;
+    // this does count as a walker creation so it gets a new walker id
+    auto walker_id = nextWalkerID();
+    walkers_.back()->setWalkerID(walker_id);
```
ye-luo, line 215: Fuse the above 2 lines in one?

src/QMCDrivers/MCPopulation.cpp
```diff
@@ -293,8 +347,16 @@ void MCPopulation::saveWalkerConfigurations(WalkerConfigurations& walker_configs
   walker_configs.resize(walker_elec_particle_sets_.size(), elec_particle_set_->getTotalNum());
   for (int iw = 0; iw < walker_elec_particle_sets_.size(); iw++)
   {
+    // The semantics of this call are not what would I would expecte.
```
ye-luo, line 350: typo expect

src/QMCDrivers/MCPopulation.cpp
```diff
@@ -293,8 +347,16 @@ void MCPopulation::saveWalkerConfigurations(WalkerConfigurations& walker_configs
...
+    // Are walkers's R's invalid here?
+    // you are not serializing the population walkers but the particle sets
+    // to the walker_configs walkers...
+    // So if you would like them to carry information between sections be careful
+    // you need to copy it in explicitly.
     walker_elec_particle_sets_[iw]->saveWalker(*walker_configs[iw]);
```
ye-luo, line 356: I also don't like this line. Actually I'm wondering if this is a bug.

src/QMCDrivers/MCPopulation.cpp
```diff
@@ -64,18 +100,40 @@ void MCPopulation::createWalkers(IndexType num_walkers, const WalkerConfiguratio
...
+  // this part is time consuming, it must be threaded and calls should be thread-safe.
+  // It would make more sense if it was over crowd threads as the thread locality of the walkers
+  // would at least initially be "optimal" Depending on the number of OMP threads and implementation
+  // this may be equivalent.
+#pragma omp parallel for shared(walker_ids)
   for (size_t iw = 0; iw < num_walkers_plus_reserve; iw++)
   {
-    walkers_[iw]             = std::make_unique<MCPWalker>(elec_particle_set_->getTotalNum());
+    int parent_id = 0;
+    walkers_[iw]             = std::make_unique<MCPWalker>(walker_ids[iw], parent_id, elec_particle_set_->getTotalNum() );
```
ye-luo, line 118: could you do "0 /* parent_id */"?
Comment: shocked to find out `Walker` class has 4 constructors.

src/QMCDrivers/MCPopulation.cpp
```diff
@@ -64,18 +100,40 @@ void MCPopulation::createWalkers(IndexType num_walkers, const WalkerConfiguratio
...
     if (const auto num_existing_walkers = walker_configs.getActiveWalkers())
+    {
+      // Save this walkers id.
+      auto this_walkers_id = walkers_[iw]->getWalkerID();
+      // the walker assignment operator is basically a simple copy so walker_id is overwritten with
+      // something meaningful only in another QMC section.
       *walkers_[iw] = *walker_configs[iw % num_existing_walkers];
+      // An outside section context parent ID is multiplied by -1.
+      walkers_[iw]->setParentID(-1 * walker_configs[iw % num_existing_walkers]->getWalkerID());
+      walkers_[iw]->setWalkerID(this_walkers_id);
```
ye-luo, line 131: really bad code. save, overwritten, overwritten.
Is it possible to move the creation of walkers_[iw] in to the branch and construct using copy constructors?

src/QMCDrivers/MCPopulation.cpp
```diff
@@ -44,8 +45,43 @@ MCPopulation::MCPopulation(int num_ranks,
...
+  for (MCPWalker& good_walker : good_walkers)
+  {
+    int num_copies = static_cast<int>(good_walker.Multiplicity);
+    while (num_copies > 1)
+    {
+      auto walker_elements = spawnWalker();
+      // In the batched version walker ids are set when walkers are born and are unique,
+      // parent ids are set to the walker that provided the initial configuration.
+      // If the amplified walker was a transfer from another rank its copys get its ID
+      // as their parent, Not the walker id the received walker had on its original rank.
```
ye-luo, line 62: If a walker A got transferred as B and then amplified to C.
```
B.parent_id = A.walker_id
C.parent_id = B.walker_id.
```
is this the scheme in use?
I kind of assume one generation (step) between walker_id and parent_id but for C there is 0 step between walker_id and parent_id. There is no issue for tracking but we cannot count the number steps anymore right?
Not a blocking issue but would like to make it clear.

src/QMCDrivers/MCPopulation.cpp
```diff
@@ -156,27 +210,27 @@ WalkerElementsRef MCPopulation::spawnWalker()
...
   else
   {
-    app_warning() << "Spawning walker number " << walkers_.size() + 1
-                  << " outside of reserves, this ideally should never happened." << std::endl;
-    walkers_.push_back(std::make_unique<MCPWalker>(*(walkers_.back())));
+    outputManager.resume();
+    auto walker_id = nextWalkerID();
+    app_warning() << "Spawning walker ID " << walker_id << " this is living walker number " << walkers_.size()
+                  << "which is beyond the allocated walker reserves, ideally this should never happen." << '\n';
+    walkers_.push_back(std::make_unique<MCPWalker>(*(walkers_.back()), walker_id, 0));
```
ye-luo, line 223: put "0 /* parent_id */"

ye-luo (commented):
Thank you for breaking up the initial PR.
I leave a few comments and questions but questions needs to be sorted out but should not be blocking.

src/Particle/Walker.h
```diff
@@ -161,10 +179,13 @@ class Walker
...
   Walker(const Walker& a) : Properties(1, WP::NUMPROPERTIES, 1, WP::MAXPROPERTIES) { makeCopy(a); }
-  Walker(const Walker& a, long walker_id, long parent_id)
-      : walker_id_(walker_id), parent_id_(parent_id), Properties(1, WP::NUMPROPERTIES, 1, WP::MAXPROPERTIES)
+  Walker(const Walker& a, long walker_id, long parent_id) : Properties(1, WP::NUMPROPERTIES, 1, WP::MAXPROPERTIES)
   {
     makeCopy(a);
+    // makeCopy replaces walker_id_ and parent_id with a.walker_id_  ...
+    // so these can not be set in the contructor initializer list
+    walker_id_ = walker_id;
+    parent_id_ = parent_id;
```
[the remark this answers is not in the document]
PDoakORNL: So do you think we should remove parent_id as a constructor argument and use a.getWalkerID().  This captures behavior in the Walker.h source that is actually elsewhere right now.  I could argue the way the the ID's are set is the reponsibility of MCPopulation and the walker is basically just a data structure.  What do you think is going to be easiest to understand?

src/QMCDrivers/MCPopulation.cpp
PDoakORNL: Seems just as clear will do.

src/QMCDrivers/MCPopulation.cpp
PDoakORNL: I'll look into it this is indeed ugly.

src/Particle/Walker.h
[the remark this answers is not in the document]
ye-luo: We can disable the constructor with ids and fully rely on set functions. If both ids are always been manipulated from outside.

src/QMCDrivers/MCPopulation.cpp
PDoakORNL: Yes, this requires a great deal less state than having fission of walkers that started on the rank follow one logic and fission of walkers that were transferred follow another.

PDoakORNL: I think the real place to document this is `WalkerControl` and the manual but its potentially confusing enough that I felt like these notes might help in the future.

src/QMCDrivers/MCPopulation.cpp
PDoakORNL: Working this a little I don't see a way around reworking walker constructor and id semantics and I think that needs to be a separate PR. I think it will be easier since this will ripple to tests, comments, and I think into files in the next PR in this series.

src/Particle/Walker.h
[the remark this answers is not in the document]
ye-luo: That is better. https://github.com/QMCPACK/qmcpack/issues/5068

src/QMCDrivers/MCPopulation.cpp
ye-luo: The current code is correct. No problem updating it later as we clean up Walker class constructors.

src/QMCDrivers/MCPopulation.cpp
ye-luo: I raised this question simply to capture my thoughts as was going through the PR not a direct issue of this PR.
For this function, its behavior has been documented clearly. My question was to the full scheme and I agree with you documentation for this behavior at WalkerControl or develop manual will be more appropriate.

src/QMCDrivers/MCPopulation.cpp
ye-luo: https://github.com/QMCPACK/qmcpack/issues/5069

ye-luo:
Please address the remaining minor issues and this PR will be ready to go.

ye-luo:
Test this please

ye-luo (approved):

PDoakORNL:
I have no ideas on the osx CI failure and can't reproduce it, I can't build with gcc-12 at all on my mac, there are problems just compiling with GCC-12 and the osx headers for me. Homebrew is onto gcc-14 at this point. gcc-14 has some sort of issue with catch. current llvm is fine.

@jptowns already reported this bus error but resolved it by going to gcc-14 but with M1/2 mac and not sure the OSX rev.  But I have compile issues with that gcc in osx 14.5 on my x86 mac...

PR #5064, opened 2024-06-27
Enable ahead-of-time compilation when using oneapi

ye-luo:
## Proposed changes
Enable ahead-of-time reduces initialization time by skipping the JIT compilation.

- Build related changes

- No

Aurora

## Checklist


PDoakORNL (approved):
What is happening without this?  I would have assumed JIT would not be the default and require more work on our side.

ye-luo:
JIT was the default and still is. AOT is used only when you specify the actual GPU architecture. AOT requires intel compute runtime being installed and I cannot use it on my laptop.

ye-luo:
Test this please

PR #5066, opened 2024-06-27
Introduce DiracMatrixInverter abstraction

ye-luo:
## Proposed changes
Introduce DiracMatrixInverter base class to avoid exposing all the inverters via DiracDeteriminantBatched.h

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/DiracMatrixComputeCUDA.hpp
+++ b/src/QMCWaveFunctions/Fermion/DiracMatrixComputeCUDA.hpp
@@ -22,7 +22,9 @@
 #include "type_traits/complex_help.hpp"
 #include "Concurrency/OpenMP.h"
 #include "CPU/SIMD/algorithm.hpp"
-#include "ResourceCollection.h"
+#if defined(ENABLE_OFFLOAD)
+#include "DiracMatrixInverter.hpp"
+#endif
 
 namespace qmcplusplus
 {
@@ -45,8 +47,11 @@ namespace qmcplusplus
  *  We don't need to actively synchronize the old stream because it gets synchronized right after each use.
  *
  */
-template<typename VALUE_FP>
-class DiracMatrixComputeCUDA : public Resource
+template<typename VALUE_FP, typename VALUE = VALUE_FP>
+class DiracMatrixComputeCUDA
+#if defined(ENABLE_OFFLOAD)
+    : public DiracMatrixInverter<VALUE_FP, VALUE>
+#endif
 {
   using FullPrecReal = RealAlias<VALUE_FP>;
   using LogValue     = std::complex<FullPrecReal>;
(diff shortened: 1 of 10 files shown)


ye-luo:
Test this please

src/QMCWaveFunctions/Fermion/DiracMatrixComputeOMPTarget.hpp
```diff
@@ -220,8 +217,8 @@ class DiracMatrixComputeOMPTarget : public Resource
    *  
    *  \todo measure if using the a_mats without a copy to contiguous vector is better.
    */
-  template<typename TMAT, PlatformKind PL>
-  inline void mw_invertTranspose(compute::Queue<PL>& resource_ignored,
+  template<typename TMAT>
+  inline void mw_invertTranspose(HandleResource& resource,
```
PDoakORNL, line 221: This combined with the actual method definition got mw_invert_transpose which turns this into a call where a default constructed HandleResource is stuffed into this call makes me wonder why the resource is in the API here at all.  Looks like with the introduction of the virtual base class this API should drop the resource argument from (mw_invertTranspose)

src/QMCWaveFunctions/Fermion/DiracMatrixComputeCUDA.hpp
```diff
@@ -217,19 +222,27 @@ class DiracMatrixComputeCUDA : public Resource
   }
 
 public:
-  DiracMatrixComputeCUDA() : Resource("DiracMatrixComputeCUDA")
+  DiracMatrixComputeCUDA()
+#if defined(ENABLE_OFFLOAD)
```
PDoakORNL, line 226: I don't see how this class should even be visible to the compiler if ENABLE_OFFLOAD && ENABLE_CUDA aren't defined.  I did write the unit tests so you wouldn't need ENABLE_OFFLOAD defined, because why would you it's not offload code.  But is that even the reason for this?

src/QMCWaveFunctions/Fermion/DiracMatrixComputeOMPTarget.hpp
ye-luo: It is not a resource handle but a queue. I should rename the type name and that is it. I will do a few renaming once my series of PRs are settled.

The reason I keep this argument but don't use it is the fact DiracMatrixComputeOMPTarget is a fallback and runs inversion on host and then got transferred. We won't implement a GPU inverter ourselves. Bacause it is a fallback and performance don't matter, we don't care about the queue being passed it. That is how I got intel case working without calling mkl for the batched inversion which will come later.

PDoakORNL (changes requested):
Ok so we have a formal interface now, the MatrixCompute object is not going to be treated like a resource but specifically. Seems like it could be more clear.

Needs a few changes and an explanation for an ifdef explosion in DiracMatrixComputeCUDA.

src/QMCWaveFunctions/Fermion/DiracMatrixComputeCUDA.hpp
ye-luo: DiracMatrixInverter APIs uses offload allocators and allow me to switch to select the fallback. Then the API won't be compiled in your unit test/benchmark in a CUDA only build. Then I can only activate the base class when offload is on.

src/QMCWaveFunctions/Fermion/DiracMatrixComputeOMPTarget.hpp
PDoakORNL: I'm still not sure why this means mw_invertTranspose needs a dead argument.  I can see why mw_invert_transpose does.  Is it going to become live later?

ye-luo: I see what you mean now. Removed all the use of dummy queues.

src/QMCWaveFunctions/Fermion/DiracMatrixComputeCUDA.hpp
PDoakORNL: Ok so the base class is only useful when offload is on.  It lets ENABLE_CUDA be on or off. But the other way around is sort of broken except for some unit testing, but that's not a new thing.

PDoakORNL (approved):

ye-luo:
Test this please

PR #5072, opened 2024-06-29
Using explicit template for inverter implementation classes

ye-luo:
## Proposed changes
This is the last part of https://github.com/QMCPACK/qmcpack/pull/5061
All the template variants are explicitly compiled. This is now prepared for the final (long term) target with all the build variants fused.
Explicit template exposed that a few functions of DiracMatrixComputeOMPTarget cannot be compiled and hence I tried to keep a minimal DiracMatrixComputeOMPTarget.
Since DiracMatrixComputeOMPTarget is indeed a fallback running on host, I simplified the code in  `DiracDeterminantBatched::mw_invertPsiM`

- No

epyc-server

## Checklist


ye-luo:
Test this please

PDoakORNL (approved):
This looks good to me.  Nice payoff from previous PR's

PR #5075, opened 2024-07-02
Update mpi3

ye-luo:
## Proposed changes
Unblocks NVHPC 24.5 making cpu-only builds

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

prckent:
No concerns, but for reference, which version and source did you use for Boost-mpi3? By eye I did not spot the corresponding hash.

@correaa Any comments?

correaa:
It seems that Ye did the tree update with an untagged version. Do you want make a tag and update the tree?

prckent:
Consistently using tagged versions would be better.  Is it convenient to make one now?

ye-luo:
I did a subtree update. See commit log
```
    Squashed 'external_codes/mpi_wrapper/mpi3/' changes from 7f82048b19..a9e19abead
```
So using `a9e19abead`

prckent (approved):
Origin is clear.

PR #5079, opened 2024-07-04
Fixes  #5069

### LINKED ISSUE #5069: A strange saveWalker call at MCpopulation
https://github.com/QMCPACK/qmcpack/pull/5063#discussion_r1655626388
Needs visiting. Right now `MCpopulation::saveWalkerConfigurations` seems showing two source of truth.
Walker configuration can be source from `walker_elec_particle_sets_` and `walkers_`.
Better clearly define the scope of both objects when data is up-to-date.

ye-luo:
## Proposed changes
MCpopulation::saveWalkerConfigurations should only save data from internal walker configurations not particlesets.
It affects DMC restart.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCDrivers/MCPopulation.cpp
+++ b/src/QMCDrivers/MCPopulation.cpp
@@ -107,14 +107,13 @@ void MCPopulation::createWalkers(IndexType num_walkers, const WalkerConfiguratio
   for (size_t iw = 0; iw < num_walkers; iw++)
     walker_ids[iw] = nextWalkerID();
 
-  // this part is time consuming, it must be threaded and calls should be thread-safe.
-  // It would make more sense if it was over crowd threads as the thread locality of the walkers
-  // would at least initially be "optimal" Depending on the number of OMP threads and implementation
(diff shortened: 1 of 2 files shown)


prckent (approved):
Anything similar lurking in DMC?

prckent:
Test this please

ye-luo:
It seems that I made a fix to VMC but there is actually no change to VMC. DMC is the one got affected. Due to branching, coordinates in ParticletSets may not be the same as Walkers. Using the former one is incorrect.

prckent:
I was wondering about any similar Walker / ParticleSet manipulations that might be lurking in the DMC sources. There are reports of the batched code paths being noisier than legacy, so perhaps there is some similar confusion.

ye-luo:
> I was wondering about any similar Walker / ParticleSet manipulations that might be lurking in the DMC sources. There are reports of the batched code paths being noisier than legacy, so perhaps there is some similar confusion.

Would like to know what the reports are especially those after Jan 2024.

PR #5086, opened 2024-07-11
Fix acforce tests due to output reordering caused by d149ef72.

ye-luo:
## Proposed changes
Only need to update reference column number in the check script.

- No

bora

## Checklist


ye-luo:
Test this please

prckent (approved):

PR #5087, opened 2024-07-11
Simplify evaluateIonDerivs used by force calculation.

ye-luo:
## Proposed changes
Under the assumption that the force is evaluated after energy.
1. There is no need to recompute/return energy during a derivative calculation. Thus name `H::evaluateWithIonDerivs` to `H::evaluateIonDerivs` with the return value removed.
2. No need to keep versions that randomize quadrature grid. Thus remove the non-deterministic variants and remove `Deterministic` in the function names, all the APIs are deterministic.

- Other (please describe): remove unused and simplify existing code paths. 

- No

bora

## Checklist

### DIFF
--- a/src/QMCHamiltonians/ACForce.cpp
+++ b/src/QMCHamiltonians/ACForce.cpp
@@ -110,9 +110,9 @@ ACForce::Return_t ACForce::evaluate(ParticleSet& P)
   //This function returns d/dR of the sum of all observables in the physical hamiltonian.
   //Note that the sign will be flipped based on definition of force = -d/dR.
   if (fastDerivatives_)
-    ham_.evaluateIonDerivsDeterministicFast(P, ions_, psi_, psi_wrapper_, hf_force_, wf_grad_);
+    ham_.evaluateIonDerivsFast(P, ions_, psi_, psi_wrapper_, hf_force_, wf_grad_);
   else
-    ham_.evaluateIonDerivsDeterministic(P, ions_, psi_, hf_force_, pulay_force_, wf_grad_);
+    ham_.evaluateIonDerivs(P, ions_, psi_, hf_force_, pulay_force_, wf_grad_);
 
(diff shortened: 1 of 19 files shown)


ye-luo:
Test this please

prckent:
"Under the assumption that the force is evaluated after energy." Just wondering: Suppose this is done the other way around, incorrectly. This would be a state machine type error. Would it be caught? Requiring an unclear order of calls (state machine) is one of those things that makes the app hard to program with.

ye-luo:
> "Under the assumption that the force is evaluated after energy." Just wondering: Suppose this is done the other way around, incorrectly. This would be a state machine type error. Would it be caught? Requiring an unclear order of calls (state machine) is one of those things that makes the app hard to program with.

The calling sequence is guaranteed by QMC drivers. The only case that can mess up is creating unit tests from scratch without paying attention to the calling sequence. That is basically testing a unused case and the test should be fixed in the first place.

In a few derived class implementations where recomputing energy costs nothing, I added assertion checking against stored values just in case. We don't require implementing in the most efficient way in every case. Developers can still recompute things as needed. If it eats too much performance, our design allows enabling efficient implementations.

Similarly, in H::evaluate we can recompute Pset and TWF without depending on the previous state of Pset and TWF but we won't do it since it is absolutely unnecessary.

rcclay (approved):
LGTM.  A fine simplification of API's and call logic.

ye-luo:
Indeed one debug build failed at an assertion I added but likely due to round error. I will take a closer look.

ye-luo:
Test this please

ye-luo:
checking float point values on the fly causes false positive assertion failure and undesired abort. I move them to unit tests.

rcclay (approved):

prckent:
Thanks for the approval Ray.

PR #5098, opened 2024-07-23
Nexus: Extend GCTA to SOC Calculations

aannabe:
## Proposed changes
Previously the grand-canonical twist-averaging (GCTA) implementation was hard-coded for collinear systems with spin up and down electrons only. This PR extends it to accept noncollinear systems for SOC calculations. It is mostly achieved by ignoring the down spin where necessary, similar to the qmcpack input.

I also fixed a minor bug (line 156), which erroneously asserted that normalization factor for k-points must be integer. This worked fine for even k-mesh grids, but errors out in odd k-mesh grids, as found by @camelto2 .

- New feature
- Bugfix

- No

Tested on OLCF/Andes using fcc Au system (only input generation using Nexus, no QMC runs).

## Checklist


prckent (commented):
What did you test this on and how did you test it?

aannabe:
At this point I tested that it successfully generates the QMC inputs on fcc gold. I also tested that the previous collinear is intact by running `nspin = 1` and `nspin = 2` on gold. Similarly, Al and Fe collinear systems work without issues. Again, these tests were done only at Nexus level and by hand. It would be good to add some GCTA tests in future.

prckent (approved):

prckent:
Test this please

PR #5100, opened 2024-07-24
passing PSPool to EstimatorManagerNew and a some cleanup

PDoakORNL:
## Proposed changes

To construct the EnergyDensityEstimator the particle set pool is required. This PR adds this and cleans up some test code and type aliases.  This supports further pieces of the EnergyDensityEstimator for batched feature merge.

- New feature

- No

ios laptop

## Checklist


ye-luo:
When I look at what is needed to construct `EstimatorManagerNew`
```
https://github.com/QMCPACK/qmcpack/blob/3757cfe96a012bd4e55338f87cf25f751ee42c07/src/QMCDrivers/QMCDriverNew.cpp#L82
```
It needs `comm`, `global_emi`, `population` which are all accessible at the factor.
How about create `EstimatorManagerNew` outside the constructor of `QMCDriverNew` and pass it in?

Pools are accessed by Builders/factories not the final run object. So drivers should not interact with pools.
Think about some estimators may need not just pset pool but other pools. We cannot just add more arguments to the constructor of `QMCDriverNew`.

PDoakORNL:
The pools are need to be visible to estimators as they are currently designed. And estimator lifetimes are scoped inside estimator manager new and estimator manager crowd whose lifetime is scoped within the driver. The pools are main scope.   therefore they need to be passed into be available at construction of the EstimatorManagerNew. The driver factories are a smell necessary due the the legacy design and the desire to not write constructors.

ye-luo:
> The pools are need to be visible to estimators as they are currently designed. And estimator lifetimes are scoped inside estimator manager new and estimator manager crowd whose lifetime is scoped within the driver.

Agree but you can still treat the construction of EstimatorManager separately from drivers. Once constructed, you transfer its ownership to a driver, namely tie its lifetime to a driver.

> The pools are main scope. therefore they need to be passed into be available at construction of the EstimatorManagerNew.

Yes that is why, I said we could build EstimatorManagerNew in the mains cope before constructing a driver and then transfer its ownership.

> The driver factories are a smell necessary due the the legacy design and the desire to not write constructors.

I can only partially agree. If we would like to avoid handling xml in a constructor, we do need a collection of operations before calling a constructor. That is the purpose of a factory/builder. Also constructor complicates error handling.

PDoakORNL:
I tried to do this just to get by review although I disagree with building the estimator manager in the factories and moving it,  and I'm not going to finish since  its going to require some substantial work in `QMCFixedSampleLinearOptimizeBatched` which needs to repeatedly constructs a VMCBatched driver and now would also need to repeatedly make the estimator manager.  And of course it will need to hold on the the pool to do that anyway.  `QMCFixedSampleLinearOptimizeBatched` It can't just be passed an EstimatorManager because the manager is not intended to be used for multiple driver invocations.  And I don't think it should be changed to do so.

The EnergyDensityEstimator expects to be able to retrieve its particle sets by name which requires access to the pool.
Otherwise I would rather just hand particle sets as arguments to the constructor. 
The structure factor estimator also needs access to the particle pool since specifies its psets by name. 

Maybe the golden ion particleset needs to be in population, but its out of scope for the EnergyDensityEstimator feature. I don't think should block getting feature complete in batched. I still think we'd be better off passing the pools in for use at construction time, or alternately having the ions  in population as well.  At anyrate it seems out of scope for getting the EnergyDensityEstimator in.

ye-luo:
Thanks for pointing out `QMCFixedSampleLinearOptimizeBatched`. I saw more ugliness there but I believe it is fixable.

prckent:
Passing comment for @PDoakORNL since I saw the remarks on QMCFixedSampleLinearOptimizeBatched: 99.9% of the time it is a user mistake have any estimators enabled in optimization because they have no plans to use them and they will slow down the run, possibly considerably. Since the specific goal of wavefunction optimization is to optimize wavefunctions via properties computed from the physical Hamiltonian, we could simply not calculate anything else there, particularly if it simplifies the code and as well as keeps people out of trouble. For the 0.1% of times someone wants to check a property such as the density during optimization they can simply put a loop around vmc opt and regular vmc.

PDoakORNL:
I still maintain that the estimator manager is a subordinate object of the Driver that gets all its input from a subnode of the driver input node with exception to the global estimator node (which I don't think is necessary).  Its lifetime by design is inside of the driver lifetime and its visibility needn't extend past the driver.  Why moving this construction process outside of the construction of the object that owns it to many different factories rather than in the shared code of the drivers doesn't seem worth it to follow the rule that the pools aren't visible in the drivers at construction time. Other than the optimizer none of the drivers or estimators keep the reference to the pools.

PDoakORNL:
Also I don't want messing with the optimizer design (which I think is questionable and needs a great deal of work) in the path of getting these last two estimators in.

PDoakORNL:
Pushed beginning of removing PSPool argument to driver constructors and replacing with EstimatorManagerNew.  Doesn't look promising to me.  Could give EMN a move constructor and pull most repeated factory code into a factory_function.  But...

ye-luo:
superseded by #5105

PR #5105, opened 2024-07-30
Work around Pools cant be visible inside driver constructors "rule" for new estimators

PDoakORNL:
## Proposed changes

Alternate to #5100 with help from Ye, this makes changes he wanted to see.

This breaks the closed relationship between the drivers and the QMCDriverInput but does not pass the PSPool to the driver constructor.  Otherwise the goal is the same have the string to global particle set map available to estimators owned by a section driver's estimator manager at construction time. Estimators that require this are coming in pending PRs.

Previously we had single step construction of the EstimatorManagerNew as part of the construction of the driver. We still could (see #5100).

To work around the issue that at driver scope the rule is we cannot access the string -- global pset,twf,ham pools we do the following work around: We construct estimator manager with only some of the arguments it needs to be valid i.e. it can't instantiate any of the estimators specified  in the input because it doesn't get any of the the objects they need to be constructed. Then you call `constructEstimators` to actually bring it into actual validity.  Then it is moved into the Driver class. 

This won't have much impact overall, but I will continue to insist in reviews that objects be valid at construction. This exception should not become the rule.

I had to make a somewhat nasty rebase

- New feature

- Yes
If you were using estimators inside vmcdriver in the batched optimizer driver that is no longer supported.

osx laptop

## Checklist


src/Estimators/OperatorEstBase.h
```diff
@@ -92,6 +92,7 @@ class OperatorEstBase
   virtual void startBlock(int steps) = 0;
 
   std::vector<QMCT::RealType>& get_data() { return data_; }
+  virtual std::size_t getFullDataSize() { return data_.size(); }
```
ye-luo, line 95: const function?

ye-luo (commented):
LGTM.

PDoakORNL:
In retrospect yes I should have merged I think...

ye-luo:
Test this please

ye-luo (approved):

src/QMCDrivers/QMCDriverFactory.cpp
```diff
@@ -261,9 +330,33 @@ std::unique_ptr<QMCDriverInterface> QMCDriverFactory::createQMCDriver(xmlNodePtr
...
+      qmcdriver_input.readXML(cur);
+      vmcdriver_input.readXML(cur);
+    }
+    catch (const std::exception& e)
+    {
+      throw UniformCommunicateError(e.what());
+    }
+
+    if (qmcdriver_input.get_estimator_manager_input())
+      throw UniformCommunicateError("Local Estimator sections are not allowed for Optimizer drivers!");
```
ye-luo, line 351: This error message doesn't reflect the actual implementation. Could you print out a warning saying
```
The batched wavefunction optimization driver ignores local estimator input.
```

At the moment, please make it user-friendly as a warning instead of a stopper.

src/QMCDrivers/QMCDriverFactory.cpp
```diff
@@ -261,9 +330,33 @@ std::unique_ptr<QMCDriverInterface> QMCDriverFactory::createQMCDriver(xmlNodePtr
...
+    if (emi)
+      app_log() << "Batched optimizer IGNORES global estimator input.";
```
ye-luo, line 353:
```
The batched wavefunction optimization driver ignores global estimator input.
```
and printout to warning.

ye-luo:
Test this please

ye-luo (dismissed):

ye-luo:
Test this please

ye-luo (approved):

PR #5114, opened 2024-08-08
introduce RandomNumberControl::getChildren

ye-luo:
## Proposed changes
Intend to eventually make `RandomNumberControl::Children` and access via exclusively `getChildren()`.

- New feature

- No

laptop

## Checklist


src/QMCDrivers/QMCDriverNew.cpp
```diff
@@ -99,9 +99,10 @@ QMCDriverNew::QMCDriverNew(const ProjectData& project_data,
 // To work around the issue, check the local pointer for nullptr before restoring to global storage.
 QMCDriverNew::~QMCDriverNew()
 {
+  auto& rng_children = RandomNumberControl::getChildren();
```
PDoakORNL, line 102: This is still awful because we're tying QMCDriverNew to a static object. I don't think it should know RandomNumberControl at all. But this PR doesn't bake that in any further.

src/QMCDrivers/QMCDriverNew.cpp
```diff
@@ -287,16 +288,10 @@ void QMCDriverNew::createRngsStepContexts(int num_crowds)
...
   Rng.resize(num_crowds);
 
-  if (RandomNumberControl::Children.size() == 0)
-  {
-    app_warning() << "  Initializing global RandomNumberControl! "
-                  << "This message should not be seen in production code but only in unit tests." << std::endl;
-    RandomNumberControl::make_seeds();
-  }
-
+  auto& rng_children = RandomNumberControl::getChildren();
```
PDoakORNL, line 291: Complete havoc if in testing I create two QMCDriverNew derived classes, the sematics of this are bizarre and exist because the optimizer was not really improved for batched. QMCDriver should take some sort of random context in to retain determinism but which does not bleed how many crowds/threads there are into the global scope. The function should return some updated RandomContext perhaps with just its starting offset updated or whatever. If test code or the main was spawning another driver it could then also update the and no dependence on this global object with global OMP thead semantics to some degree assumed. Also just no invisible single rng collection state machine. 

This could also solve the very annoying problem in tests where multiple  tests cases that use RNG can end up broken when different subcollections of tests are ran.  A useful capability of catch which this global mess completely destroys.

PDoakORNL (approved):
I don't see anything wrong with this in itself, but It's moving orthogonally to improving the issue with multiple RNG use in the code.

src/QMCDrivers/QMCDriverNew.cpp
ye-luo: I'm actually working on this. Moving the ownership of RNGs back and forth is really bad and I'm going to make the driver side keeping reference without changing ownership.

src/QMCDrivers/QMCDriverNew.cpp
ye-luo: I noticed the complaint in the source code comment and it is time to change.

ye-luo:
Test this please

PR #5115, opened 2024-08-08
Remove no longer used Communicate::initialize

ye-luo:
## Proposed changes
tidy up codes.

- No

laptop

## Checklist

### DIFF
--- a/src/Message/Communicate.cpp
+++ b/src/Message/Communicate.cpp
@@ -79,9 +79,6 @@ Communicate::Communicate(const Communicate& in_comm, int nparts)
     GroupLeaderComm.reset();
 }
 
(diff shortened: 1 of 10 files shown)


prckent:
Please add another PR that adds an MPI Sandbox build in github actions so that this problem is caught in CI and future post-breakage fixes won't be necessary. Sandbox builds are very fast so this won't slow the CI significantly.

ye-luo:
> Please add another PR that adds an MPI Sandbox build in github actions so that this problem is caught in CI and future post-breakage fixes won't be necessary. Sandbox builds are very fast so this won't slow the CI significantly.

Will do

ye-luo:
Test this please

PR #5118, opened 2024-08-09
PR for new batched EnergyDensityEstimator input class (synthetic)

PDoakORNL:
## Proposed changes

Adds input class for EnergyDensityEstimator. This is one of the more interesting input classes since it delegates two subnodes to composed input classes. I think this is an important aspect of my input class design since it provides a way to cleanly add subfeatures and decompose parsing and input functionality into clear pieces without creating spurious is-a class hierarchies.  This should allow more code reuse that is not copy paste or inheritance based. We use native C++ constructs instead of generic dictionaries so the code can actually reflect the composition semantics and ownership of input subnodes.

- New feature
- Build related changes

- No

sdgx x86 v100

## Checklist


src/Estimators/EnergyDensityInput.h
```diff
@@ -0,0 +1,89 @@
...
+#include "SpaceGridInput.h"
+#include "type_traits/template_types.hpp"
+
+namespace qmcplusplus
+{
+namespace testing
+{
+template<typename T>
+class EnergyDensityTests;
+}
```
ye-luo, line 26: Why do we need these lines? Seems not used.

src/Estimators/EnergyDensityInput.h
```diff
@@ -0,0 +1,89 @@
...
+
+class NEEnergyDensityEstimator;
```
ye-luo, line 28: What does `NE` stands for?

PDoakORNL: New Estimator, it became necessary because of ODR  and the EnergyDensityEstimator in QMCHamiltonian.
While a scheme could probably be devised around it it will require significant build changes.  It's best just not to have two identically named classes in the same applicaiton. I would like to delete the legacy estimators are rename this.

src/Estimators/EnergyDensityInput.h
PDoakORNL: Seems not for this PR. I removed it.

ye-luo (approved):

PDoakORNL:
test this please

ye-luo:
Test this please

PR #5122, opened 2024-08-13
Move Rng out of QMCDriver base class.

ye-luo:
## Proposed changes
Individual derived class should carry the know whether and how many RNGs to carry with.
I'm refactoring legacy drivers here and will later touch the batched drivers.

- No

laptop

## Checklist

### DIFF
--- a/src/QMCDrivers/CorrelatedSampling/CSVMC.h
+++ b/src/QMCDrivers/CorrelatedSampling/CSVMC.h
@@ -51,13 +51,16 @@ private:
   int prevStepsBetweenSamples;
   ///blocks over which normalization factors are accumulated
   int equilBlocks;
+
+  ///driver copy of Random number generators
+  UPtrVector<RandomBase<QMCTraits::FullPrecRealType>> Rng;
+
   /// Copy Constructor (disabled)
   CSVMC(const CSVMC&) = delete;
   /// Copy operator (disabled).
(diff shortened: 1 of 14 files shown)


ye-luo:
Test this please

prckent (approved):
I did wonder if we should move some of the rng handling to the base driver class, but that could be constraining for some future fancy driver.

ye-luo:
I'm still debating within myself. It all comes down to the question what a QMCDriver base class is. Current implementation of QMCDriver is more of a common bits of VMC and DMC than a well abstracted base class. With the help of QMCDriverInterface, we are in a better position to sort out how to improve the abstraction.

On the legacy driver side, let me document the current implementation. VMC/DMC doesn't directly interact with `RandomNumberControl::Children`. Their constructors take a vector of RNGs which can be `RandomNumberControl::Children` or fake RNGs for unit tests. This is a good pattern that we probably will adopt in the batched drivers. CSVMC and RMC instead keep internal copies of RNGs which are copied from and back to `RandomNumberControl::Children` at the beginning and end of the driver run. I don't really feel comfortable with this behavior. WFOpt is a fancy/high-level driver now doesn't carry RNGs any more. All the patterns explored on the legacy side do help us figuring out good abstraction on the batched drivers.

ye-luo:
Test this please

PR #5123, opened 2024-08-13
Move QMCDriveNew::rngs_/step_contexts_ to derived classes.

ye-luo:
## Proposed changes
Following #5122. Make RNGs driver specific object and remove the ownership motion of RNGs in VMC/DMC batched drivers.

- No

laptop

## Checklist

### DIFF
--- a/src/QMCDrivers/DMC/DMCBatched.cpp
+++ b/src/QMCDrivers/DMC/DMCBatched.cpp
@@ -377,9 +377,12 @@ void DMCBatched::runDMCStep(int crowd_id,
 
 void DMCBatched::process(xmlNodePtr node)
 {
+  ScopedTimer local_timer(timers_.startup_timer);
   print_mem("DMCBatched before initialization", app_log());
+
   try
   {
+
     QMCDriverNew::AdjustedWalkerCounts awc =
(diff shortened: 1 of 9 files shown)


PDoakORNL (dismissed):
This looks good to me, especially if we're working toward removing more ominous comments.  These changes remind me that we don't just have state and reproducibility of deterministic tests and produciton runs dependent on the rng coordination across the application and concurrency contexts we also have some bombs with respect to global parameters like "number of threads" the process has and assumptions about the multiplicity of certain objects.

ye-luo:
See my comment https://github.com/QMCPACK/qmcpack/pull/5122#issuecomment-2289225953
I noticed that legacy VMC/DMC drivers are unit tested by passing in fake RNG but we are not doing it in the batched drivers. A follow-up change will be needed to cut the direct interaction between batched drivers and RandomNumberControl::Children in order to enable proper unit tests of batched drivers.

ye-luo:
Changed my mind a bit. If I consider QMCDriverNew only a base class to VMC/DMC or closed related method. The RNG object should stay in QMCDriverNew. I will make a new PR better describing the actual change.

PR #5137, opened 2024-08-21
Move NonLocalTOperator objects from Hamiltonian to Drivers

ye-luo:
Review after #5135
## Proposed changes
1. NonLocalTOperator object is no longer one per Hamiltonian but per QMCUpdateBase or crowd.
2. Hamiltonian no longer needs to keep track of nlpp_ptr which only works when there is only one NLPP. 
3. Make makeNonLocalMovesPbyP virtual and SO NLPP can implement it later
4. NonLocalTOperator class still needs some cleanup around constructor which I keep it intact.
5. DMCContextForSteps inherited from ContextForSteps to include a NonLocalTOperator object.

- New feature

- No

laptop

## Checklist


src/QMCDrivers/QMCDriverNew.h
```diff
@@ -112,6 +111,16 @@ class QMCDriverNew : public QMCDriverInterface, public MPIObjectBase
   //xmlNodePtr walker_logs_xml;
 
 protected:
+  class ContextForSteps
+  {
```
prckent, line 115: Can you add a description?

src/QMCDrivers/DMC/DMCBatched.cpp
```diff
@@ -33,13 +33,25 @@
...
 
+class DMCBatched::DMCContextForSteps : public ContextForSteps
+{
+public:
+  DMCContextForSteps(RandomBase<FullPrecRealType>& random_gen, const NonLocalTOperator& non_local_ops)
+      : ContextForSteps(random_gen), non_local_ops(non_local_ops)
+  {}
+
+  ///non local operator
+  NonLocalTOperator non_local_ops;
```
prckent, line 52: Can this be protected or private?

prckent (changes requested):

src/QMCDrivers/QMCDriverNew.h
ye-luo: will do.

src/QMCDrivers/DMC/DMCBatched.cpp
ye-luo: ContextForSteps and its derived class are intended to be just a struct. Marking it private and adding an access function are unnecessary.

src/QMCDrivers/QMCDriverNew.h
ye-luo: fixed.

ye-luo:
Test this please

prckent (approved):

Issue #5140, opened 2024-08-21
Offload assertion failures with ROCm debug builds


prckent:
**Describe the bug**

Several nightly tests are failing at assertion checks on host-device consistency. These checks only exist in debug builds.

Could be us, could be another ROCm bug.

Build: AMDClang-NoMPI-Offload-CUDA2HIP-Complex-Debug

deterministic-He_ae-opt_vmc-batch-mwalkers-1-16
https://cdash.qmcpack.org/tests/2189533
QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp:697: virtual void qmcplusplus::DiracDeterminantBatched<qmcplusplus::PlatformKind::CUDA, std::complex<double>, std::complex<double>>::mw_evaluateGL(const RefVectorWithLeader<WaveFunctionComponent> &, const RefVectorWithLeader<ParticleSet> &, const RefVector<ParticleSet::ParticleGradient> &, const RefVector<ParticleSet::ParticleLaplacian> &, bool) const [PL = qmcplusplus::PlatformKind::CUDA, VT = std::complex<double>, FPVT = std::complex<double>]: Assertion `dpsiM_from_device == det.dpsiM' failed.

Also shows in AMDClang-NoMPI-Offload-CUDA2HIP-Debug

deterministic-He_ae-opt-batch-mwalkers-1-16 

src/QMCWaveFunctions/Fermion/DiracDeterminantBatched.cpp:697: virtual void qmcplusplus::DiracDeterminantBatched<qmcplusplus::PlatformKind::CUDA, std::complex<double>, std::complex<double>>::mw_evaluateGL(const RefVectorWithLeader<WaveFunctionComponent> &, const RefVectorWithLeader<ParticleSet> &, const RefVector<ParticleSet::ParticleGradient> &, const RefVector<ParticleSet::ParticleLaplacian> &, bool) const [PL = qmcplusplus::PlatformKind::CUDA, VT = std::complex<double>, FPVT = std::complex<double>]: Assertion `dpsiM_from_device == det.dpsiM' failed.

Also in AMDClang-NoMPI-Offload-CUDA2HIP-Debug

**To Reproduce**

nightly test configuration -- ask if not clear on cdash.

**Expected behavior**

Test passes

**System:**
 nitrogen2, MI210.

prckent:
Run with latest ROCm 6.2.0 compilers

PR #5142, opened 2024-08-21
Remove value assertion that caused false positive faiure

ye-luo:
## Proposed changes

closes https://github.com/QMCPACK/qmcpack/issues/5140

assertions are not robust against numerical round error. When evaluating LCAO, the batched and non-batched evaluation may differ slightly due to using gemm and gemv.

- Bugfix

- No

jlse

## Checklist


PDoakORNL (approved):
LGTM

ye-luo:
Test this please

PR #5139, opened 2024-08-21
Energy density estimator itself

PDoakORNL:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

Describe what this PR changes and why.  If it closes an issue, link to it here
with [a supported keyword](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- Bugfix
- New feature
- Code style update (formatting, renaming)
- Refactoring (no functional changes, no api changes)
- Build related changes
- Testing changes (e.g. new unit/integration/performance tests)
- Documentation or build script changes
- Other (please describe):

### Does this introduce a breaking change?

- Yes
- No

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes/No. This PR is up to date with current the current state of 'develop'
- Yes/No. Code added or changed in the PR has been clang-formatted
- Yes/No. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- Yes/No. Documentation has been added (if appropriate)

PR #5145, opened 2024-08-23
Ohmms matrix assignment operator should be able to do a widening assigment

PDoakORNL:
## Proposed changes

Right now the Matrix class can only do a widening or narrowing assignment if the Matrices involved have the same dimensions. I don't think this makes much sense and its quite problematic for how I want to write the hdf5 output from the new estimators.  To do that hdf5 output only widening assignments are made i.e. a Matrix<float> is assigned to a Matrix<double>.

#5139 depends on this PR to fix its CI failures.

However there are number of narrowing assigments in the mixed precision all adhere to this matching size rule. Maybe that's an important constraint?  So I have left this behavior. But I do think that probably it should be changed so all Matrix Matrix assignments have the same semantics.

Important to note that through the magic of PETE all the non matching assignments were going through the same template operator=.  And I've just added a overload to hand assignment from a narrower Matrix. 

#
- New feature

- No

x86 v100
## Checklist

- Yes  This PR adds tests to cover any new code, or to catch a bug that is being fixed

### DIFF
--- a/src/Containers/OhmmsPETE/OhmmsMatrix.h
+++ b/src/Containers/OhmmsPETE/OhmmsMatrix.h
@@ -2,7 +2,7 @@
 // This file is distributed under the University of Illinois/NCSA Open Source License.
 // See LICENSE file in top directory for details.
 //
-// Copyright (c) 2021 QMCPACK developers.
+// Copyright (c) 2024 QMCPACK developers.
 //
 // File developed by: Ken Esler, kpesler@gmail.com, University of Illinois at Urbana-Champaign
 //                    Miguel Morales, moralessilva2@llnl.gov, Lawrence Livermore National Laboratory
@@ -35,6 +35,7 @@ public:
   using size_type     = typename Container_t::size_type;
   using iterator      = typename Container_t::iterator;
   using This_t        = Matrix<T, Alloc>;
+  using Alloc_t       = Alloc;
 
   Matrix() : D1(0), D2(0), TotSize(0) {} // Default Constructor initializes to zero.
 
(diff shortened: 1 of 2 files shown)


PDoakORNL:
The codecov failure looks like its checking the coverage of the `test_Matrix.cpp` which is undesirable.  I'm having trouble even getting the coverage build to link (issue with my own environment and curl).  Should I write an issue for this?  It does seem to be a failure with `compare_gcov.py` but I shouldn't go down that rabbit hole this week.

src/Containers/OhmmsPETE/OhmmsMatrix.h
```diff
@@ -163,14 +164,36 @@ class Matrix
   }
 
   // Assignment Operators
+  /// From another Matrix with matching type.
   inline This_t& operator=(const This_t& rhs)
   {
     resize(rhs.D1, rhs.D2);
+    // I don't understand why it would be desirable to have this compile
+    // but just do the resize and not the assigment, just seems like a surprising foot gun.
```
ye-luo, line 172: A good catch. Users can abuse it for making matching matrices without the need of copying over values. Such use cases should not call operator `=` .
Needs a revisit when we scrutinize the operator `=` overloading.

ye-luo (approved):

ye-luo:
> The codecov failure looks like its checking the coverage of the `test_Matrix.cpp` which is undesirable. I'm having trouble even getting the coverage build to link (issue with my own environment and curl). Should I write an issue for this? It does seem to be a failure with `compare_gcov.py` but I shouldn't go down that rabbit hole this week.

https://app.codecov.io/gh/QMCPACK/qmcpack/pull/5145?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=QMCPACK

It seems indicating so. We have a 'ignore' setting https://github.com/QMCPACK/qmcpack/blob/8e075b47030057ab9c7ce42ede15fdebf7f9121d/codecov.yaml#L18
but I don't know how it interacts with codecov.

ye-luo:
Test this please

prckent:
Thanks Peter. This is a good catch. The assignment operator as resize but no assignment looks like a old design bug; there could be an unintentional use lurking.

PR #5155, opened 2024-09-03
Update OneShiftOnly doc

ye-luo:
## Proposed changes

Clarify how to choose and adjust `minwalkers`

- No

epyc-server

## Checklist

### DIFF
--- a/docs/methods.rst
+++ b/docs/methods.rst
@@ -606,7 +606,7 @@ optimization block:
 <parameter name="MinMethod"> THE METHOD YOU LIKE </parameter>
 
 OneShiftOnly Optimizer
-~~~~~~~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^^^^^^^
 
 The OneShiftOnly optimizer targets a fast optimization by moving parameters more aggressively. It works with OpenMP and GPU and can be considered for large systems.
 This method relies on the effective weight of correlated sampling rather than the cost function value to justify a new set of parameters.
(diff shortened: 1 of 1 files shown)


docs/methods.rst
```diff
@@ -655,10 +655,14 @@ Recommendations:
...
+When optimizing parameters from scratch, We recommended using this optimizer in two sections with a very small ``minwalkers``
+in the first and a large value in the second, such as the following.
 In the very beginning, parameters are far away from optimal values and large changes are proposed by the optimizer.
-Having a small ``minwalkers`` makes it much easier to accept these changes.
-When the energy gradually converges, we can have a large ``minwalkers`` to avoid risky parameter sets.
+Having a small ``minwalkers`` makes it much easier to accept these changes. If optimization becomes unstable, increase ``minwalkers``.
+If optimization gets stuck with proposed parameter sets being constantly rejected, decrease ``minwalkers``.
+When the energy gradually converges, keeping a large ``minwalkers`` is necessary to prevent accepting risky parameter sets.
+Continuing optimization with more parameters from a partially converged wavefunction should also use large ``minwalkers``,
+for example adding three-body Jastrow factor to converged one-body and two-body Jastrow factors.
```
PDoakORNL, line 665: It sounds like this process is expected to take several job submissions, Is it actually made clear elsewhere in the manual how to do this?

prckent: Peter: We need more examples. Another strategy is also to freeze/unfreeze parameters.

This updated text is a good improvement. It needs a little smoothing for language and I think specific examples of small and large values for minwalkers in the text would help with readability. One also has to search to find the acceptable range of values for minwalkers and what it is.

ye-luo: @PDoakORNL It can be done with several jobs, it can also be done within a single job using https://qmcpack.readthedocs.io/en/develop/methods.html#variational-parameter-selection with a bit waste of computation evaluating zeros.

To reach control free optimization, we need to minimize the set of controlling knobs, understand how they behave and then develop a smart scheme. As @PDoakORNL mentioned, examples are important and I also think some FAQs are needed.

PDoakORNL: So this seems like an improvement for sure.  Unless you're planning to write more now I think it could just go in.

prckent:
Test this please

PR #5167, opened 2024-09-16
Improve input file argument handling

### LINKED ISSUE #3124: Input filename handling and errors are incorrect
**Describe the bug**

We require an xml suffix for the xml inputs but don't give correct errors if the inputs don't have this suffix.

`qmcpack thisfiledoesnotexist` reports "no input file given". Add a .xml suffix and the error is correctly reported. Also running `qmcpackpack ensemblefile.txt` with missing .xml inputs in the ensemble gives the correct error, but if the suffixes are not xml the "no input file given" error returns.

Not a big error but might puzzle new users.

```
$ ./qmcpack thisfiledoesnotexist
Rank =    0  Free Memory = 176684 MB
No input file is given.
Usage: qmcpack <input-files> 
$ ./qmcpack thisfiledoesnotexist.xml
Rank =    0  Free Memory = 176686 MB
  Input file(s): thisfiledoesnotexist.xml 

                    QMCPACK 3.11.9
...
  Input XML = thisfiledoesnotexist.xml
I/O warning : failed to load external entity "thisfiledoesnotexist.xml"
ERROR File thisfiledoesnotexist.xml is invalid
Fatal Error. Aborting at main(). Input invalid.
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI COMMUNICATOR 3 DUP FROM 0
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
--------------------------------------------------------------------------
```
**To Reproduce**

run qmcpack directly, any version.

**Expected behavior**

Error and expected suffix should be correctly communicated.

**System:**

nitrogen, but any will do.

**Additional context**

prckent:
## Proposed changes

Closes #3124 

Enforce QMCPACK input files end in .xml vs have xml string in them

Accurate more verbose message when no valid input specified 

(Could still be nicer for new users, e.g. print list of valid options,  print QMCPACK title and version)

Also: Remove long hidden/untested/undocumented clone feature. There are other ways to do what it does, so simplify by removing.

- Bugfix

- No

All deterministic tests including ensemble tests, gcc14 mpi,

## Checklist

### DIFF
--- a/src/QMCApp/qmcapp.cpp
+++ b/src/QMCApp/qmcapp.cpp
@@ -53,7 +53,6 @@ int main(int argc, char** argv)
   {
     //qmc_common  and MPI is initialized
     qmcplusplus::qmc_common.initialize(argc, argv);
-    int clones = 1;
     std::vector<std::string> fgroup1, fgroup2;
     int i = 1;
     while (i < argc)
@@ -61,8 +60,6 @@ int main(int argc, char** argv)
       std::string c(argv[i]);
       if (c[0] == '-')
       {
-        if (c.find("clones") < c.size())
-          clones = atoi(argv[++i]);
         if (c == "-debug")
           ReportEngine::enableOutput();
 
@@ -109,7 +106,7 @@ int main(int argc, char** argv)
       }
       else
       {
-        if (c.find("xml") < c.size())
+        if (c.find(".xml") == c.size() - 4)
           fgroup1.push_back(argv[i]);
         else
         {
@@ -121,7 +118,7 @@ int main(int argc, char** argv)
             getwords(words, fin);
             if (words.size())
             {
-              if (words[0].find("xml") < words[0].size())
+              if (words[0].find(".xml") == words[0].size() - 4)
               {
                 int nc = 1;
                 if (words.size() > 1)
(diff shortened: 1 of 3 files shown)


src/QMCApp/qmcapp.cpp
```diff
@@ -141,18 +138,20 @@ int main(int argc, char** argv)
       ++i;
     }
     int in_files = fgroup1.size();
-    std::vector<std::string> inputs(in_files * clones + fgroup2.size());
+    std::vector<std::string> inputs(in_files + fgroup2.size());
```
ye-luo, line 141: could you replace `in_files` with `fgroup1`? I would also recommend renaming `fgroup1` as `f_names_cmd` and `fgroup2` as `f_names_txt` to help reading the code.

src/QMCApp/qmcapp.cpp
```diff
@@ -121,7 +118,7 @@ int main(int argc, char** argv)
             getwords(words, fin);
             if (words.size())
             {
-              if (words[0].find("xml") < words[0].size())
+              if (words[0].find(".xml") == words[0].size() - 4)
```
ye-luo, line 121: The nc below also behaves very similar to the 'clones' you just removed and thus I think we should also remove the `nc` handling copies. 

Could you also update https://qmcpack.readthedocs.io/en/develop/running.html?highlight=command%20line#running-qmcpack
`<XML input file(s)>` to `<XML input file(s) and/or text file(s)>` and document the desired text file content.

src/QMCApp/qmcapp.cpp
prckent: Done

src/QMCApp/qmcapp.cpp
prckent: Done

ye-luo:
Test this please

ye-luo (approved):

PR #5156, opened 2024-09-04
Refactor Conjugate Gradient part of SR optimization

camelto2:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

This PR refactors part of the new SR optimization introduced in #5017. This cleans up and simplifies the code inside QMCFixedSampleLinearOptimimzeBatched and hopefully makes a new CG interface that is easier to work with and add new features like preconditioning in the future. Also adds a much needed test to this new part of the code

## What type(s) of changes does this code introduce?
_Delete the items that do not apply_

- New feature
- Code style update (formatting, renaming)
- Refactoring (no functional changes, no api changes)

### Does this introduce a breaking change?

- No

## What systems has this change been tested on?
macOS

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- Yes. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)

PR #5017, opened 2024-05-31
Simple implementation of Stochastic Reconfiguration

camelto2:
Please review the [developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
on the wiki of this project that contains help and requirements.

## Proposed changes

Wanted to out this up to get some eyes on it.

This implements a very simple version of stochastic reconfiguration. The benefit is that we avoid the <Psi_i | H | Psi_j > matrix elements which get costly for a large number of parameters. This stochastic reconfiguration approach basically solves -tau * h = S * dp for dp, where tau is a small projection time, h is the vector <Psi_i | H | Psi_0>, and S is the overlap matrix <Psi_i | Psi_j>. The simple approach would be to build S, invert, and solve for dp. A better method, employed here, is to use a conjugate gradient method to solve the linear equation above. This avoids having to explicitly build the entire S matrix, and only builds the S*z matrix-vector product at each iteration of the CG algorithm...essentially this is the "Accelerated" SR approach described in https://doi.org/10.1103/PhysRevB.85.045103, described in the text surrounding Eqn. 7. 

Preliminary tests using this have allowed us to scale up 100,000+ variational parameters, which was simply not possible with the current implementation of one_shift_only and has a huge speedup over one_shift for large parameter counts. 

This is still a WIP because there are a number of things to improve. A few things off the top of my head are

1. Need to add meaningful tests. This was the result of some exploratory coding to see if it could work. Now that our preliminary testing seems to suggest the method works pretty well and is fast, I need to add proper tests.
2. Would like to abstract away the CG solver
3. Probably need a better implementation of the CG solver...I literally implemented whatever was on wikipedia. I have some hard coded CG convergence criteria which may not be sufficient
4. Right now, it either uses a fixed tau to get the parameter update or it can do a line search with correlated sampling. In some cases, the line search is far superior and converges significantly faster. However, if the correlated sampling weight gets small and stays small throughout the optimization then the energy can blow up and tank the optimization. In those cases, you just need to use the tau as something small, but this can result in many optimization iterations. Each iteration is a lot faster, but you may end up needing O(100) or more to get convergence. Also if the tau is too large, then the optimization can go haywire as well.  I think an optimal approach would be to use the line search unless the weights get too small, and just take steps with small tau otherwise. Or some sort of adaptive method which scales the timestep automatically to help accelerate convergence. 
5. probably many other things I'm not thinking of. 

[...]

PR #5169, opened 2024-09-17
Change regularization in SR

camelto2:

## Proposed changes
This PR changes the regularization scheme back to just adding a constant to the diagonal of the overlap matrix. If S becomes ill-conditioned, the CG can be slower to converge. In #5156, I changed the regularization to what was supposed to be a scaling of the diagonal elements which was argued in a paper from Sorella to be a better regularization. Effectively, in matrix notation, this would be S-> S(I+diag(S)*eps). Unfortunately, I implemented S(I+eps * I) which scales all elements of S, not just the diagonal. That operation obviously does not reduce the condition number, and some simple test problems I was playing around with were running slower since the CG step was taking longer. 

Also, doing some simple numerical experiments on diagonally dominant matrices similar to the overlap matrix, it seems that adding a constant to the diagonal has a larger reduction in the condition number than the scaling the diagonal for a fixed value of the scaling constant epsilon. So for now, it seems better to just revert back to how I implemented it in the original PR. 

I also updated the documentation and changed the name of the method as Ye requested in #5017 after it was merged. 

- Bugfix

- No

mac 

## Checklist

### DIFF
--- a/docs/methods.rst
+++ b/docs/methods.rst
@@ -1175,7 +1175,7 @@ Additional information and recommendations:
    the descent engine to do this averaging on its own.
 
 Stochastic Reconfiguration with Conjugate Gradient
-^^^^^^^^^^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 We have implemented a preliminary version of stochastic reconfiguration (:cite:`Sorella2001` and :cite:`Casula2004`),
 currently only available in the batched drivers. The SR optimization reduces the
 computational cost over the linear method by avoiding the need to build the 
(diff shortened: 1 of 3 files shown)


docs/methods.rst
```diff
@@ -1174,7 +1174,7 @@ Additional information and recommendations:
    parameters discussed earlier for descent are useful for setting up
    the descent engine to do this averaging on its own.
 
-Stochastic Reconfiguration
+Stochastic Reconfiguration with Gradient Descent
```
ye-luo, line 1177: Can we have `Conjugate Gradient` instead of `Gradient Descent`?

camelto2: yep thats what I meant to put. Working too fast. Will fix

docs/methods.rst
```diff
@@ -1174,7 +1174,7 @@ Additional information and recommendations:
    parameters discussed earlier for descent are useful for setting up
    the descent engine to do this averaging on its own.
 
-Stochastic Reconfiguration
+Stochastic Reconfiguration with Conjugate Gradient
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ye-luo, line 1178: also need to add enough `^^^` otherwise we get warnings when compiling the docs.

camelto2: again moving too fast

ye-luo (approved):

ye-luo:
Test this please

PR #5184, opened 2024-09-23
Nexus: Handle J3 terms in spin-orbit calculations

aannabe:
## Proposed changes
For spin-orbit calculations, we only have up electron species. Currently in Nexus, J3 is written as usual with up and down electrons, which errors out in qmcpack. This PR loops over the three-body-Jastrow object and deletes terms that have up-down interactions.

- Bugfix

- No

Local machine.

## Checklist


prckent:
Did you investigate the possibility of not creating these terms in the first place?

aannabe:
I haven't investigated that route. Deleting the down terms seemed the easiest way in terms of handling it. Also, J2 was handled similarly to this, but for some reason J3 was left out. If you think it's better to not generate these terms in the first place, I can take a closer look.

prckent:
Thanks -- I was just asking. Probably this is the pragmatic and quickest route and we should go ahead with this (@jtkrogel ?) but clearly it would be cleaner to never create incorrect terms in the first place. That would mean the parts of Nexus that create wavefunction components would inspect the spin and spin-orbit configuration before creating their terms.

camelto2:
This LGTM. I am approving in case you need non-ORNL approval, and I would like to have this for my work.

camelto2 (approved):

ye-luo:
Test this please

PR #5200, opened 2024-10-14
Nexus: Bugfix for SOC J3 terms

aannabe:
## Proposed changes

In #5184, I introduced the handling of J3 terms in Nexus for SOC calculations. At the time I tested it in CrI3, which has two species. For single-element systems such as Ir, the data structure of J3 terms changes (reported by @rcclay). In this edge case, nothing needs to be done when incorporating the previous Jastrow, since there will be only a single 'uu' term in J3.

- Bugfix

- No

NERSC. Nexus tests are passing.

## Checklist


aannabe:
Added an explicit check for the above-described case.

rcclay (approved):
LGTM

ye-luo:
Test this please

PR #5207, opened 2024-10-28
Add batched EnergyDensityEstimator short statistical integration test

PDoakORNL:
This adds an application level short test for the batched EnergyDensityEstimator

However on my system the test fails because after 250 blocks the paticle moves start moving out of the period cell.  The legacy energy estimator doesn't check that the coordinate to cell mappings land in the grid, but I now bound check this access so the test fails.  Even if i don't I segfault eventually because overflow writes are occurring.

These out of bounds particles are pretty rare but it seems certain at least on my test system that they will occur in a run the size of the short run.

This seems to point to a likely bug in the particle moves. I will explore further tomorrow.

This PR though make it possible to run the EnergyDensityEstimator with the full app which was actually impossible without the fix to the virtual write function.

- Bugfix

- No

dgx 2

## Checklist


prckent:
Good catch. In the QMC simulation the particles are allowed to move anywhere, so any code looking to handle periodic images needs to do so and handle the out of bound case appropriately. Unless the fix is trivial, I suggest to just fix the batched code.

prckent:
Would you like this merged? It would at least start appearing in the nightlies.

prckent (approved):
LGTM

prckent:
Test this please

ye-luo (approved):

PR #5208, opened 2024-10-30
Fix TMDLA in batched DMC driver runs

ye-luo:
## Proposed changes
Fix a bug reported by Yasmine@UCL.

- Bugfix

- No

bora

## Checklist


prckent:
Test this please

prckent:
What are your thoughts for a small test that would catch/reproduce this error? A new LCAO Ca atom test or is there an existing test you think we could modify?

prckent (approved):
Thanks for chasing this down Ye.

ye-luo:
> What are your thoughts for a small test that would catch/reproduce this error? A new LCAO Ca atom test or is there an existing test you think we could modify?

I'd like to first try to adjust the existing Na based NLPP unit test.

PR #5210, opened 2024-11-06
Fix Lattice + OpenBC in batched drivers

ye-luo:
## Proposed changes
When using spline orbitals with boundary condition "n", legacy drivers does one more check when a step is proposed. If the new position is located out of the box in any of the direction marked "n", the proposed move is marked invalid. This bug fixes handled this case.

Think about running a single atom VMC using HF with QE WF and then use "n n n". if electrons are allowed outside the box, the wavefunction norm is infinite due to finite electron density everywhere. Confining electrons within the box prevents this bad situation effectively making electron density 0 outside the box.

Another note. makeMoveAndCheck also rejects moves larger than half of the box as long as a lattice is defined. This is not implemented in the batched API. However, this requires me to regenerate reference numbers for `deterministic-diamondC_2x1x1_pp-vmcbatch-dmcbatch-mwalkers-serialized_sdj` tests.

- Bugfix

- No

Bora

## Checklist


ye-luo:
Test this please

ye-luo:
Test this please

prckent (commented):
Thanks so much for tracking this down.

I was tempted to ask about masking the work in the use_drift section of VMCBatches.cpp:136+  and the similar section of DMCBatched.cpp:201+, where we might have invalid moves… but do you agree that updating the function signatures wouldn’t bring enough benefit given it is a rare event?

I will ask if we are certain it is safe to call the functions in these regions with invalid moves?

I think it is safe with the current implementations but the comments/docs for the WaveFunctionComponents and ParticleSet should be updated with a suitable note/warning that they have to handle arbitrary positions.

ye-luo:
Calculating drift (flex_calcRatioGrad) is fine even positions are invalid because all our SPOs can be evaluated at any position. Jastrows are always defined throughout the space as required by LCAO molecule cases. There is no need to mask ` use_drift`.

"handle arbitrary positions" is a vague requirement. Sometimes, we just need WFCs to run as the results will be discarded regardless. I don't think feel the necessity to put extra note on WFC which might cause unnecessary confusion to developers. The handling is at driver level with both ParticleSet and TWF involved, Individual WFC doesn't need to take responsibility.

PaticleSet yes, I added an extra note `mw_makeMove`, "Note that: drivers should reject invalid moves." `PSdispatcher::flex_makeMove` actually requires `std::vector<bool>& are_valid` return.

prckent (approved):

PR #5078, opened 2024-07-03
Tidy EinsplineSetBuilder

ye-luo:
## Proposed changes
1. ReadOrbitalInfo and ReadOrbitalInfo returning bool removed. Use exception
2. Remove codes related to `Format != ESHDF` already inaccessible.
3. Concentrate bcastBandInfoSet calls
4. Fix OccupyBands_ESHDF hang when error occurs and make it const.

## What type(s) of changes does this code introduce?
- Bugfix

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
epyc-server

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'

PR #5211, opened 2024-11-07
Fix printing 0 caused by #5078

ye-luo:
## Proposed changes
#5078 removed setting `NumDistinctOrbitals` inside `EinsplineSetBuilder::OccupyBands_ESHDF` thus, the printing should also be updated.

- Bugfix

- No

epyc-server

## Checklist


ye-luo:
Test this please

PR #5212, opened 2024-11-08
fix out of bounds index calculation for NESpaceGrid

PDoakORNL:
## Proposed changes

Especially in mixed precision when a minimum image position is taken through a dot product to lattice based coordinates the number error can be great enough to put them very slightly outside of the cell, the algorithm use to convert those lattice coordinates to integer indexes into the grid then produces an out of bounds value. 
We know we should never be far off but rather than do a bunch of float operations I just correct the outside by one indexes caused by this numerical noise.

- Bugfix

- No

## Checklist


src/Estimators/NESpaceGrid.cpp
```diff
@@ -588,16 +588,31 @@ void NESpaceGrid<REAL>::accumulate(const ParticlePos& R,
       {
         particles_outside[p] = false;
         Point u              = dot(axinv_, (R[p] - origin_));
+	// numerical accuracy of the dot is not good enough for this to be not produce out of bounds indexes
+	// with REAL = float, u 10^-8 outside of umin_ to  umax_ is not uncommon i.e  p > (1 in 32000)
+	// for the coordinate ranges typical of our simulations. I guess the branch predictor will mostly
+	// mitigate this.
+	auto gmapIndex = [this](int d, const auto& u) { int raw_index = floor((u[d] - this->umin_[d]) * this->odu_[d]);
```
ye-luo, line 595: Is `u[d] - this->umin_[d] ` always positive? I'm wondering if raw_index can be `-5` when the chosen space is located in the center of a lattice.

src/Estimators/NESpaceGrid.cpp
```diff
@@ -588,16 +588,31 @@ void NESpaceGrid<REAL>::accumulate(const ParticlePos& R,
...
+	  if ( raw_index == gmap_[d].size() )
```
prckent, line 596: Thanks for diagnosing this Peter. A classic case of someone mapping a floating point number into an integer index and not taking care to ensure that the resultant integer is within range given the issues for floating point tolerances, rounding etc....

In the code here, are you certain that the mapping is correct for periodic boundary conditions cases? if the case of arriving at a negative index we could map to either the zeroth or nth-1 grid point. Here you are mapping -1 to zero (first choice). Should we instead be adding/removing gmap_[d].size to get within range 0-gmap_[d].size-1? This is my inclination. This would also address Ye's question, if we haven't already ensured that the [d] - this->umin_[d] is positive.

src/Estimators/NESpaceGrid.cpp
PDoakORNL: This code is just a direct port from legacy and it is complicated in how it sets up the grid lattice coordinate system, I think it might take more time than I can give today to trace through it all. So here is the current case:
```xml
    <parameter name='lattice' units='bohr'>
        3.37316115        3.37316115        0.00000000
        0.00000000        3.37316115        3.37316115
        3.37316115        0.00000000        3.37316115
    </parameter>
```
applyMinimumImage guarantees I will get something that is positive definite for my r_particle

With the space grid input like
```xml
           <spacegrid coord="cartesian">
             <origin p1="zero"/>
             <axis p1="a1" scale=".5" label="x" grid="-1 (.1) 1"/>
             <axis p1="a2" scale=".5" label="y" grid="-1 (.1) 1"/>
             <axis p1="a3" scale=".5" label="z" grid="-1 (.1) 1"/>
           </spacegrid>
``` 
This gives us a transformation matrix that looks like
```c++
axinv_ = {0.29645781969236779, 0.29645781969236779, -0.29645781969236779, -0.29645781969236779, 0.29645781969236779, 0.29645781969236779,
    0.29645781969236779, -0.29645781969236779, 0.29645781969236779}
```
Which seems to be based on the hard assumption that if `coord="cartesian"` whatever grid is defined it going to map the unit cell into it.

Which ideally would take all r_particle to u_grid_particle = {[-1,1],[-1,1],[-1,1]} but instead takes us to something that has some small numerical error +/- for floatingpoint operations of fixed precision.

(I find the way the origin argument works here counterintuitive because I think we are really getting a grid whose origin is `{0.5,0.5,0.5}` in the actual lattice coordinates.  But the Energy Density Estimator grid docs do say: "zero -- Cell center" so I guess this is a feature, not what I would do here since both seems a totally valid way to define the lattice.)

then we get to the actual gmapIndex algorithm. `u - umin` where `umin = {-1,-1,-1}` Which gives a a positive definite [0,2) * odu_

The legacy code has very long and complex initialization that appears to guarantee that the relation between umin_, umax_ and odu_ is always such  without floating point error the indexes are always [0, umax[d]-umin[d] * (1 / step[d]) ).  There was no unit testing of this and I certainly have not had the time to test corner cases I doubt it is terribly robust, but perhaps I just don't see that it is perfection outside of the naivete wrt float point error.

All cartesian coordinate are forced to be treated as period 
```
bool isPeriodic() const { return !(coord_form_ == CoordForm::CYLINDRICAL || coord_form_ == CoordForm::SPHERICAL); }
```

A concern here is  minimumImage is only called in the EDE if the lattice is not open so that's a corner case that is not handled.

cartesian grid coordinates must also be symmetric about whatever origin is used.  None of this is checked for more extensively in legacy than in the current code.  In the scope of this PR I don't think its feasible to do more than fix the out of bounds in the cases that can occur when the user sets up their grid correctly.  Otherwise the exception gives them what we can to fix the grid.

src/Estimators/NESpaceGrid.cpp
PDoakORNL: So you could also claim that for nth grid point you can map to zeroth or nth-1, my observation is misses on both sides occurred.

And
```
std::floor(u[d] * odu[d]) - std::floor(umin[d] * odu[d])
```
seems like it needs a different anlysis umin now could always be truncated even if was arbitrarily close to the next integer but below. Whereas u would still be free to be -1 or +1 about the range. I guess then -2 goes to nth-1 and -1 goes to 0;

There are still missing constraints at the floating point level on the grids but that is a lack on input validation which has always been there, any of these schemes is likely to catch them since almost certainly indexes off by more than one will occur.

PDoakORNL:
Test this please

src/Estimators/tests/ValidSpaceGridInput.h
```diff
@@ -30,6 +30,7 @@ struct ValidSpaceGridInput
     SPHERICAL,
     WITH_STEP,
     WITHOUT_STEP,
+    WEIRD_CARTESIAN,
```
ye-luo, line 33: WEIRD is valid or not? I doesn't sound absolutely wrong. Confused.

ye-luo:
Do we have traps for cases mentioned in the manual but not currently implement?

src/Estimators/tests/ValidSpaceGridInput.h
PDoakORNL: Well its not super strange but the scaling is different for one dimension. This is perhaps unexciting enough to just drop.  Initially I was trying to have it cover stranger cases but they all turned out to have issues with respect to mapping into the grid properly.  I guess a stranger case would be if the origin was shifted to properly drop the grid onto something other than +/- r_x / scaling.  Currently you can't for instance do 0 1 as it won't cover the minimum image coordinate range.  For all these bad choices of you now get an exception rather than a crash when a memory operation is enough out a bounds.

PDoakORNL:
The unimplemented features namely VORONOI grids which is in the manual and a chemical potential feature which isn't are trapped by the input parsing.

src/Estimators/tests/ValidSpaceGridInput.h
ye-luo: I see

ye-luo (approved):

ye-luo:
Test this please

src/Estimators/tests/ValidSpaceGridInput.h
ye-luo: CI is busy on another PR. If you have a second, could add a comment
```
    WEIRD_CARTESIAN, # briefly explain the special aspect of this test case and also comment if this input case is valid or not.
```

ye-luo:
Test this please

Issue #5050, opened 2024-06-19
Nexus: Numpy 2.0 incompatibility


prckent:
I noticed in #5049 that the macos CI failed due to the use of Numpy 2.0 which was released on 17 June. 69 test failures. Possibly we can pin Numpy < 2.0. By eye the failures are to do with string handling and numerical precision. Needed updates might be small but backwards compatibility also needs to be considered.

@jtkrogel Can you please look and propose a reasonable way forward?

prckent:
Updated the title since we have the interim workaround.

prckent:
Closed by #5215

PR #5215, opened 2024-11-09
Nexus: Initial NumPy 2 support

prckent:
## Proposed changes

This PR is sufficient to have all the nexus tests pass with numpy2 (tested with 2.1.2). Unchanged behavior with 1.26.4 was also verified.

This PR does need some discussion on the preferred way forward and likely needs some reworking. Docs also need updating in a subsequent PR, plus we should have at least a nightly test configuration with numpy2 etc.

There are two classes of update:

1. Updates to type aliases. Trivial.
2. numpy2 changes some printing of numeric scalars to include their type information. This breaks k-point printing among other similar operations in the existing tests. I enabled a legacy printing option as an ugly workaround for PySCF and pwscf. (https://numpy.org/doc/stable/reference/generated/numpy.set_printoptions.html#numpy-set-printoptions). However any code that does not have test coverage and that does printing could be affected. "Sometime" a better solution will be needed.

Note I missed seeing the official suggested pattern for testing numpy version (
https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#writing-numpy-version-dependent-code ), which could be used to replace the ugly string manipulation I put in a few spots ahead of `set_printoptions(legacy="1.25")`

I looked at this because as noted in https://github.com/QMCPACK/qmcpack/issues/5050 , numpy was out in June and users will likely hit it before long.

- New feature
- Build related changes

- No

sulfur, nightly gcc 14.2 configuration.

## Checklist


jtkrogel:
I think we might be able to consolidate the changes (and make them generally available) by introducing a single function:
```
def portable_numpy(np):
    if np.lib.NumpyVersion(np.__version__) >= '2.0.0b1':
        np.set_printoptions(legacy="1.25")
        np.string_ = np.bytes_
        np.float_  = np.float64
    #end if
#end def portable_numpy
```

For most of Nexus, this could be placed in versions.py and then called at the top of any affected file like this:
```
from versions import portable_numpy
import numpy as np
portable_numpy(np)
```

The function would have to be duplicated (for independence) in testing.py and called there also.

If I'm right, this approach would not require any changes to the body text in Nexus files and also be portable across numpy versions.

prckent:
Thanks for the cleaner suggestion. I reset my original changes and added your function to testing.py... which surprisingly turned out to already be enough due to the wide usage of the testing module.

I removed the [WIP] since this is hopefully enough for now.

ye-luo (approved):

ye-luo:
Test this please

PR #5227, opened 2024-11-13
Drop FindBoost module which is deprecated in CMake 3.30 and raise minimum version to 1.70.

ye-luo:
## Proposed changes
Boost 1.70 was release in 2019.

Instead of the CMake shipped module file, upstream Boost's BoostConfig.cmake package configuration file will be used.
This change requires installing boost but we only need the "headers" library.
    
Installation of boost headers is very simple and quick
```
./bootstrap.sh --prefix=<location of your installation> --with-libraries=headers
./b2 install
```
QMCPACK CMake takes `-DBOOST_ROOT=<location of your installation>` as usual.

- Build related changes

- No

laptop

## Checklist


ye-luo:
Test this please

prckent (changes requested):
Thanks for looking at this. I have been wondering about the boost cmake change for some time since we have been getting warnings on cdash from the nightlies.

Since this PR breaks multiple CI configurations, we can conclude that not only are we not ready for this change but it will also cause headaches for existing users which we can easily avoid: I suggest to simply use the new find_package mechanisms for cmake >=3.30.0 otherwise use the old scheme.

If CMake 3.30.0 needs a specific minimum boost we can use that boost minimum version for CMake 3.30.0, otherwise I also need no reason to bump the minimum version. It just creates work for no obvious benefit.

(Something to discuss for the rainy day list would would be to have boost as an optional dependency for real space builds. Removing a dependency would definitely make things easier for users. Currently optional LMY_ENGINE and AFQMC need it, but we only make a little use elsewhere.)

PDoakORNL:
I'd add I'm using cmake@3.29.2 was using boost@1.86.0 installed months ago and reinstalled after a pull of spack develop today. So this works with older cmake's but maybe not if you have a old spack install of boost even if its far newer that 1.70

prckent:
(Not changing my view here - needs updating to use the new find_package mechanisms for cmake >=3.30.0 otherwise use the old scheme. This is the least disruptive approach.)

ye-luo:
@PDoakORNL boost issue fixed by https://github.com/spack/spack/pull/46281 and https://github.com/spack/spack/pull/46062 Aug/Sep 2024. So only recently.

ye-luo:
For CMake version < 3.30, keep using the boost module.

ye-luo:
Test this please

ye-luo:
Test this please

prckent (approved):

prckent:
Thanks Ye!

prckent:
Test this please

PR #5228, opened 2024-11-14
[Multi] make changes to rely on ADL get instead of std::get

correaa:
## Proposed changes

use modified interface of Multi that doesn't rely in UB by specializing std::get in namespace std.

- No

Ubuntu gitlab CI

## Checklist


correaa:
codecov in the CI is failing, https://github.com/QMCPACK/qmcpack/actions/runs/11831469869/job/32968026644?pr=5228#step:8:46

correaa:
ok, I think I am ready with the changes Multi is updated to release v0.86.0
https://gitlab.com/correaa/boost-multi/-/releases/v0.86.0

ye-luo:
Test this please

ye-luo (approved):

ye-luo:
@correaa there were build failure in the complex builds https://github.com/QMCPACK/qmcpack/actions/runs/11873027958/job/33107327082

correaa:
Do you know a trick to compile the real and complex version simultaneously? Perhaps using Ninja Multiconfig?

prckent:
The build system has no insight into which files are different between real and complex so there is currently no benefit to be gained over two independent builds.

correaa:
The benefit, at the minimum, is that one has a single consistent cmake
command line.

prckent:
@correaa Indeed. Long term we should have a single executable. Unfortunately that work was not completed during ECP and since it is not a major blocker it is not high on the to do list. Clearly it would be a user friendly feature and simplify builds, installs, and testing. If more arguments appear to complete that work it will be moved higher on the list.

correaa:
ok, thanks. Sorry for the noise. Multi got caught in new undefined behavior introduced between C++17 and C++20 regarding overloading in std namespace.

See "Function templates and member functions of templates" in section https://en.cppreference.com/w/cpp/language/extending_std

Making the code compile in C++17 (without C++20 extensions) and C++20 requires this kind of tricks.

ye-luo:
Test this please

ye-luo (approved):

PR #5233, opened 2024-11-18
Improve the xsf format data precision

aannabe:
## Proposed changes

A common way to analyze charge/spin densities in qmcpack is to generate the `*.xsf` data files via `qdens` that reads the data from the native `stat.h5` files. The xsf files can then be further analyzed via `qdens-radial` or other tools. Currently, the data (such as charge density) in the xsf files are written as floats with 8 significant digits (`12.8f`). I think scientific notation is better for this type of data. For example, QE (v7.0) uses scientific notation with 6 significant digits (`12.6E`). In a simple test I did, the integrated charge (norm) has improved significantly using the `12.6E` style for CH4 molecules in a 16^3 Ang box. Here the correct norm is 5.

Old code:
```
Max Density: 0.00069445
Norm: 4.999454429999994
File Size: 111 MB
```

New code:
```
Max Density: 0.0006944463
Norm: 5.000000002221036
File Size: 111 MB
```

The accuracy can be further improved at the expense of a larger file size.

- enhancement

- No

- NERSC
- Local Machine

## Checklist

### DIFF
--- a/nexus/lib/fileio.py
+++ b/nexus/lib/fileio.py
@@ -680,7 +680,7 @@ class XsfFile(StandardFile):
                         if n%ncols==0:
                             c += '\n    '
                         #end if
-                        c += ' {0:12.6E}'.format(v)
+                        c += ' {0:14.8E}'.format(v)
                         n+=1
                     #end for
                     c += '\n   END_DATAGRID_{0}D_{1}\n'.format(d,dgk)
--- a/nexus/lib/fileio.py
+++ b/nexus/lib/fileio.py
@@ -680,7 +680,7 @@ class XsfFile(StandardFile):
                         if n%ncols==0:
                             c += '\n    '
                         #end if
-                        c += ' {0:12.8f}'.format(v)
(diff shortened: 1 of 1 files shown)


aannabe:
Just want to add that the reason I looked at this is to rule out any possible bugs. Having the norm close to the correct value provides a good sanity check.

prckent:
Thanks Gani. Big improvement. Happy to merge this. Are there any arguments against using an even higher precision? Using the same format as QE is one argument, but why not choose bigger files?...

aannabe:
I just tried to improve the accuracy while maintaining the same file size. Slightly increasing further to `14.8E` shows even better norm:
```
Max Density: 0.000694446266
Norm: 5.000000000032804
File Size: 126 MB
```
This is with a [200x200x200] grid (the increase in file size with denser grids will be steeper). But overall the file size is still acceptable I think, and I have now pushed it to `14.8E`.

ye-luo (approved):

ye-luo:
Test this please

PR #5214, opened 2024-11-08
Nexus: add example for specifying estimators

aannabe:
## Proposed changes

As discussed previously in persion, a comprehensive example was needed for Nexus that demonstrates how to specify various estimators. Here, I have included Density, SpinDenstiy, EnergyDensity, and 1RDM observables in a single Nexus workflow. The goal is to add all the other estimators too as they mature.

In addition, I added a `scripts_postproc` directory here that includes some post-processing tools. Currently, I included the 1RDM retrieval and plotting tools adapted from discussions in #1052. For now I think it will be helpful to users, at least as a starting point. Let me know if it's not well-suited for here and I will remove it.

For long run, there should be a centralized place for post-processing tools that analyze the estimators in the stat.h5 file. Whether it should be one large tool that is one-fits-all, or multiple specialized smaller tools could also be discussed. Currently, it seems to be not well handled - for instance, I see that `qdens` breaks if 1RDM or energy density is present in the same stat.h5 file. I will create an issue for this if the fix is not straightforward.

## What type(s) of changes does this code introduce?
- New example

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
Local machine

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'

PR #5240, opened 2024-11-22
Nexus: Set workstation thread binding to none

prckent:
## Proposed changes

The Nexus workstation configuration, e.g. ws16, implicitly assumes codes are built with OpenMPI for MPI runs. When also used with threading, the threads will be wrongly bound for our uses, giving poor performance. The binding is the default behavior of all recent OpenMPI releases and is easily verified in a system with hyperthreading enabled where the per process CPU% will show as 50% instead of the expected 100%, or in a 16 "core" run, 16 cores will not be used. Performance is consequently halved. Added --bind-to none, similar to what has already been done in some of the Supercomputer job definitions.

Open to changing to more optimal settings or moving the location of the setting, but for everyday workstation runs I think this is close enough and simple.

Noticed while testing #5214.

- Bugfix

- No

nitrogen2, GCC14 + OpenMPI, standard nightly configuration.

## Checklist

### DIFF
--- a/nexus/lib/machines.py
+++ b/nexus/lib/machines.py
@@ -978,7 +978,7 @@ class Workstation(Machine):
 
 
     def process_job_options(self,job):
-        job.run_options.add(np='-np '+str(job.processes))
+        job.run_options.add(np='--bind-to none -np '+str(job.processes))
     #end def process_job_options
 
 
--- a/nexus/tests/unit/test_machines.py
+++ b/nexus/tests/unit/test_machines.py
@@ -533,10 +533,10 @@ def test_workstation_scheduling():
     assert(j.threads==2)
     assert(j.local)
     assert(j.processes==2)
-    assert(j.run_options.np=='-np 2')
+    assert(j.run_options.np=='--bind-to none -np 2')
     assert(j.batch_mode==False)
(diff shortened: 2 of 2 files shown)


ye-luo:
Is it possible to test if it is OpenMPI in use and add the binding option if the answer is yes?

ye-luo:
Test this please

prckent:
But why those sulfur configurations?

prckent:
I discovered that some old but not too-old mpiruns don't take the thread binding settings. Rather than make assumptions, a better alternative might be to use environment variables to allow the user to build up an mpirun-type command, similar to how CMake does it. e.g. https://cmake.org/cmake/help/latest/module/FindMPI.html#usage-of-mpiexec . This would allow any MPI to be supported within the workstation class. 

"${MPIEXEC_EXECUTABLE} ${MPIEXEC_NUMPROC_FLAG} ${MPIEXEC_MAX_NUMPROCS} ${MPIEXEC_PREFLAGS} EXECUTABLE ${MPIEXEC_POSTFLAGS} ARGS"

@jtkrogel Do you know if anyone has ever looked at doing this?

prckent:
Being addressed by #5255 to allow+document more flexibility in mpirun options.

PR #5241, opened 2024-11-25
Relocate myVars in SPOSet/WFC base classes to OptimizableObject

ye-luo:
## Proposed changes
myVars are intended for routines of optimizable components interacting with WFOpt. Thus it is cleaner to pull them out of SPOSet/WFC based classes and place it in OptimizableObject class. Namely myVars disappears in derived classes without immediate optimizable contents.

`JeeIOrbitalSoA` and `TwoBodyJastrow` only contain optimizable parameters in their functors but keeps a local myVars for parameter filtering. For now, add local myVars variables. but I'd like to have them eventually going away.

The use of myVars `SlaterDet` is currently no-op and thus removed.

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/SlaterDet.cpp
+++ b/src/QMCWaveFunctions/Fermion/SlaterDet.cpp
@@ -52,14 +52,9 @@ void SlaterDet::extractOptimizableObjectRefs(UniqueOptObjRefs& opt_obj_refs)
 
 void SlaterDet::checkOutVariables(const opt_variables_type& active)
 {
-  myVars.clear();
   if (isOptimizable())
     for (int i = 0; i < Dets.size(); i++)
-    {
       Dets[i]->checkOutVariables(active);
-      myVars.insertFrom(Dets[i]->myVars);
-    }
-  myVars.getIndex(active);
 }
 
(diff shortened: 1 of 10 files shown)


prckent:
Worried about test coverage here - the interface to the optimizers doesn't have enough coverage in CI. e.g. Nothing goes through the SlaterDet::checkOutVariables isOptimizable() code path. Thoughts on getting coverage?

(I checked on codecov directly since the report is not made here yet)

ye-luo:
We only have a RotatedSPO unit test. There are no SD/MSD integration or system tests. That needs to be improved. Personally still prefer integration test to full simulation system tests. Probably can start from the He/H2 with rotated orbitals.

Regarding `isOptimizable` I'm going to remove it as many layers as possible. We are able to figure out its answer using `extractOptimizableObjectRefs` and `isOptimized`. No need to maintain an extra states.

prckent:
Yes, we should avoid whole simulation tests as much as possible. This PR looks safe but I do worry about state machine related logic bugs, which we are definitely prone to. Do you prefer to try adding an MSD-related test now or later?

ye-luo:
Prefer SD+Rotation first but unfortunately not now.

prckent:
Test this please

PR #5246, opened 2024-11-29
Adding job completion message to convertpw4qmcpack

kayahans:
## Proposed changes

Currently, convertpw4qmcpack gives no confirmation to the user if the conversion is successfully complete or not. I know the plan is to incorporate this code into pw2qmcpack, but for the short term this can be a reasonable solution. 

- No

- ORNL-Baseline 

## Checklist


prckent (approved):
Useful improvement. I changed message to be the same style as the QMCPACK completion message (qmcapp.cpp:229). I also prefer we avoid use the term "job" since this is associated with queuing systems.

prckent:
Test this please

PR #5248, opened 2024-12-02
Set recompute period in performance tests to once every QMC section

prckent:
## Proposed changes

Set the recompute periodic explicitly in the performance tests so that full & mixed precision builds do the same number of matrix inversions. (Mixed precision default is every block vs every 10 for full precision builds.) This changes addresses a couple of recent queries where the code has been profiled by users using the default settings.

Due to the few steps per block in the performance tests the existing settings result in a skew towards matrix inversion in profiles, performance analysis etc. In real/performance focused runs the number of steps per block would be larger than the e.g. 5 used for the DMC sections. A side effect is that some of the mixed precision performance tests will now show closer to real-world speedup.

Also added end of lines to a couple of the files for consistency.

- Yes

New mixed precision results should not be compared to old mixed precision results in terms of overall runtime.

gcc14 mpi nightly configuration, sulfur.

## Checklist

### DIFF
--- a/tests/performance/C-graphite/sample/dmc-a16-e64-cpu/C-graphite-S4-dmc.xml
+++ b/tests/performance/C-graphite/sample/dmc-a16-e64-cpu/C-graphite-S4-dmc.xml
@@ -98,6 +98,7 @@
     <parameter name="blocks">                 2 </parameter>
     <parameter name="timestep">             1.0 </parameter>
     <parameter name="usedrift">              no </parameter>
+    <parameter name="blocks_between_recompute">      10 </parameter>
   </qmc>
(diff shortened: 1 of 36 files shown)


ye-luo:
FYI. Just noticed that https://github.com/QMCPACK/qmcpack/pull/4109 didn't touch batched drivers and the default for FP is 0.

prckent:
Will update this PR based on our discussion (one recompute per QMC section) and then check the defaults today.

Good catch on 4109 -- it might explain a few of the oddities we have seen in large runs. Not all of them though.

prckent:
Also updated the NiO README

ye-luo (approved):

ye-luo:
Test this please

PR #5256, opened 2024-12-16
Nexus: Add dependency versions file + update docs

### LINKED ISSUE #5253: Tested Nexus dependency versions needed
**Describe the bug**
It would be great to have information on the latest tested Nexus dependencies where all the tests pass. Breakages in the tests can occur due to newer versions of these dependencies, so fixing the versions would help users start running QMC. A recent example is breakage due to the new numpy version. I now see an issue with the newest spglib version 2.5.0 where 3 tests fail in Nexus on my local machine. All tests pass in 2.4.0. See the attached log files for these different versions.
[spglib_2-4-0.txt](https://github.com/user-attachments/files/18115349/spglib_2-4-0.txt)
[spglib_2-5-0.txt](https://github.com/user-attachments/files/18115351/spglib_2-5-0.txt)

How about we include a simple `requirements.txt` file with versions under `qmcpack/nexus/`? The user can simply run `pip install -r requirements.txt`. Using my current environment where I only fix spglib==2.4.0 and others to be the latest available versions, I get the following for `requirements.txt` (for which all tests pass):
```
numpy==2.2.0
spglib==2.4.0
scipy==1.14.1
matplotlib==3.9.3
h5py==3.12.1
pydot==3.0.3
pycifrw==4.4.6
cif2cell==2.1.0
seekpath==2.1.0
``` 
**To Reproduce**
Using Python 3.13.1 install spglib 2.5.0 vs 2.4.0 and all the other dependencies and run `nxs-test`.

**Expected behavior**
Working versions for dependencies is needed where all tests pass. 

**System:**
Local Linux machine

**Additional context**
Encountered when creating a new Nexus env for a new QMC user.

aannabe:
## Proposed changes
This PR should close #5253. A `requirements.txt` file was added in `qmcpack/nexus/`. I did not add `requirements_minimal.txt` as only numpy is needed for core functionality and adding a file for just a single package seems redundant to me.

Appropriate description has been added to the Nexus documentation. Other smaller issues were also fixed in the docs: python3 support, and `--user` instead of `-user`.

- No

Local Linux Machine

## Checklist

### DIFF
--- a/nexus/requirements.txt
+++ b/nexus/requirements.txt
@@ -1,0 +1,13 @@
+# While Nexus does not have strict version requirements, this requirements file
+# corresponds to versions that have been tested and are known to work.
+# Last updated: 2024/12/16
+
+numpy==2.2.0
+scipy==1.14.1
+matplotlib==3.10.0
+h5py==3.12.1
+pydot==3.0.3
+spglib==2.4.0 # Some functionality is broken in 2.5.0
+pycifrw==4.4.6
+cif2cell==2.1.0
+seekpath==2.1.0
--- a/nexus/requirements_minimal.txt
+++ b/nexus/requirements_minimal.txt
@@ -1,0 +1,5 @@
+# While Nexus does not have strict version requirements, this requirements file
+# corresponds to versions that have been tested and are known to work.
+# Last updated: 2024/12/16
+
+numpy==2.2.0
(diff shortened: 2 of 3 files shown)


prckent:
Please add requirements_minimal. The intent is to advertise to people that the full set of requirements is not needed. e.g. if spglib breaks, or so that someone using a different package manager can see that everything is not absolutely required. If requirements_minimal is not present people will think that all the packages are required, which is not the case.

prckent (approved):
Thanks Gani!

ye-luo (approved):

PR #5260, opened 2024-12-30
Remove 16 rank driver unit tests

prckent:
## Proposed changes

Remove the 16 task driver unit tests, leaving the 1,2,3, and 4 way versions.

The 16 task tests are occasionally unreliable when run on single GPUs (e.g. via sulfur CI, and in bora nightlies) and do not add needed coverage in my opinion.

- No

None. CI should run and we should check the coverage.

## Checklist

### DIFF
--- a/src/QMCDrivers/tests/CMakeLists.txt
+++ b/src/QMCDrivers/tests/CMakeLists.txt
@@ -120,7 +120,7 @@ if(HAVE_MPI)
       qmcutil
       platform_omptarget_LA)
   endif()
-  foreach(NUM_RANKS 1 2 3 4 16)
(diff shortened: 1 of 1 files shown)


prckent:
Test this please

ye-luo (changes requested):
Put this on hold. It is worth checking the cause of the failure. Likely it is caused by insufficient GPU memory when dealing with 48 walkers each with one cusolver handle.

prckent:
Example failures on bora https://cdash.qmcpack.org/testSummary.php?project=1&name=deterministic-unit_test_new_drivers_mpi-r16&date=2024-12-30

ye-luo:
The original issue has been resolved by #5265

PR #5293, opened 2025-01-29
Add checkVector test helper equivalent to checkMatrix, NativePrint for std::unordered_map and std::array

PDoakORNL:
Expanded Utilities/for_testing.

## Proposed changes

See title
Incoming PR's will have tests that use these and provide coverage.

- New feature

- No

## Checklist

### DIFF
--- a/src/Utilities/for_testing/NativeInitializerPrint.hpp
+++ b/src/Utilities/for_testing/NativeInitializerPrint.hpp
@@ -78,6 +78,17 @@ inline std::ostream& operator<<(std::ostream& out, const NativePrint<std::vector
   return out;
 }
 
+template<class T, std::size_t N>
+inline std::ostream& operator<<(std::ostream& out, const NativePrint<std::array<T, N>>& np_arr)
+{
+  out << "{ ";
+  auto arr = np_arr.get_obj();
+  for (int i = 0; i < N; ++i)
+    out << std::setprecision(10) << arr[i] << ", ";
+  out << " }";
+  return out;
+}
+
 template<class T>
 inline std::ostream& operator<<(std::ostream& out, const NativePrint<Vector<T>>& np_vec)
 {
@@ -86,6 +97,30 @@ inline std::ostream& operator<<(std::ostream& out, const NativePrint<Vector<T>>&
   for (T& t : vec)
     out << std::setprecision(10) << t << ", ";
   out << " }";
+  return out;
+}
+
+template<class T>
+inline std::ostream& operator<<(
+    std::ostream& out,
+    const NativePrint<std::unordered_map<std::string, std::vector<Vector<T>>>>& np_crowd_energy)
+{
+  out << "{";
+  auto& crowd_energy = np_crowd_energy.get_obj();
+  for (auto iter = crowd_energy.begin(); iter != crowd_energy.end(); ++iter)
+  {
+    out << "{{\"" << iter->first << "\"}, {";
+    auto& v_walkers = iter->second;
+    for (auto& v_particles : v_walkers)
+    {
+      out << "{";
+      for (const T& t : v_particles)
+        out << std::setprecision(10) << t << ", ";
+      out << "},";
+    }
+    out << " }},\n";
+  }
+  out << "};";
   return out;
 }
 
--- a/src/Utilities/for_testing/NativeInitializerPrint.hpp
+++ b/src/Utilities/for_testing/NativeInitializerPrint.hpp
@@ -2,7 +2,7 @@
 // This file is distributed under the University of Illinois/NCSA Open Source License.
 // See LICENSE file in top directory for details.
 //
-// Copyright (c) 2022 QMCPACK developers.
+// Copyright (c) 2025 QMCPACK developers.
 //
 // File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Lab
 //
--- a/src/Utilities/for_testing/checkVector.hpp
+++ b/src/Utilities/for_testing/checkVector.hpp
@@ -23,9 +23,23 @@
 namespace qmcplusplus
 {
 
+/** return structure from vector check
+ *  For clean use with catch2 CHECKED_ELSE macro
+ *  If you would rather have an empty string carry success semantics I disagree
+ *  and so do others, it also makes it use with CHECKED_ELSE less clear.
+ *  https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f21-to-return-multiple-out-values-prefer-returning-a-struct
+ */
 struct CheckVectorResult
 {
+  /** If matrix check is successful result = true
+   *  if one or more elements fails result = false
+   */
   bool result;
+  /**  always a valid std::string object
+   *   result = true --> default constructured string at this time
+   *   result = false --> result_message element failure information,
+   *   the extent of which is determined by check_all flag to checkVector
+   */
   std::string result_message;
 };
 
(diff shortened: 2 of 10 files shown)


src/Utilities/for_testing/checkVector.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include "type_traits/complex_help.hpp"
+#include "ApproximateEquality.hpp"
+namespace qmcplusplus
+{
+
+struct CheckVectorResult
+{
+  bool result;
+  std::string result_message;
+};
```
ye-luo, line 30: Can we use empty return `error_message` for pass instead of this two element struct? Just return the error_string.

src/Utilities/for_testing/checkVector.hpp
```diff
@@ -0,0 +1,75 @@
...
+  if (a_vec.size() > b_vec.size())
+    return {false, "b_vec is too small for a_vec to be a checkable segment"};
+  std::stringstream error_msg;
+  auto vectorElementError = [&error_msg](int i, auto& a_vec, auto& b_vec) {
+    error_msg << "checkVector found bad element at " << i << "  " << a_vec[i] << " != " << b_vec[i]
+              << '\n';
+  };
+  bool all_elements_match = true;
+  for (int i = 0; i < a_vec.size(); i++)
+    {
```
ye-luo, line 61: Formatting seems off.

src/Utilities/for_testing/NativeInitializerPrint.hpp
```diff
@@ -89,6 +100,30 @@ inline std::ostream& operator<<(std::ostream& out, const NativePrint<Vector<T>>&
   return out;
 }
 
+template<class T>
+inline std::ostream& operator<<(
```
ye-luo, line 104: Could you unit test the these cases?

PDoakORNL: Added no change test cases.

PDoakORNL:
test this please

src/Utilities/for_testing/checkVector.hpp
ye-luo: Did you miss this one?

PDoakORNL: I'm not a big fan of check for null as success semantics and this turns out quite readable with the catch macros for checkMatrix. I want this to have the same semantics. See: test_ConstantSPOSet.cpp,  test_cuBLAS_LU.cpp, test_spline_applyrotation.cpp, etc.

ye-luo: The current CheckXXXResults is not concise.
If results=true, it is unclear what non-empty result_message should be used for or should I care?
If results=false, but results_message is empty. Should I concerned about an empty result_message? Is that a bug?
I don't like undefined behvaviors.
A single string is neat. If empty, pass. If non-empty, error info is given.
Do you have good ways to prevent any UB with CheckXXXResults?

PDoakORNL: A single string is ambiguous it because both a message and a success value (if empty). Where's the UB? This isn't kernel code every byte isn't important. This is a test function so the overhead of explicitly returning a structured strongly type result with a boolean results and a string that may contain additional information seems fine and is compatible with our testing frameworks' macro.

ye-luo: You didn't answer my question.
If a caller receives results=true but the message string is not empty. What action should be taken for the string or we should enforce empty string in the true case.
If a caller receives results=false but the message string is empty. Is this a bug?
That is what I mean UB and the needed action is unclear to me.
results=true/false, string=empty/non-emtpy. There are 4 combinations but you only intended to use 2 cases. That is why I don't see the benefit of struct over a simple string. If empty, pass. If non-empty, message is an error for users.

ye-luo: In addition, do we really need both CheckVectorResult and CheckMatrixResult or we can simply use one CheckResult? Do you have plan to differentiate them?

PDoakORNL: It's nicer for me to have matrix indexing available for the error reporting when its a matrix and vector indices when it is not. This also allows each matrix or vector check to count as one "assertion" rather than say like 24000 assertions which ends up making that metric useless for a particular test. It's proved useful to me when using the unit tests to actually develop and debug as opposed to just check for no change ex post factor. It's code to facilitate using the unit test code as a tool for development, its somewhat flexible but still can be left in place and check correctness simply. I'm not sure what the big deal is I provided multiple files that show useful usage of the extant checkMatrix, I just want to facility for large vectorized data, I want to use the more self documenting idiom of not overloading the semantics of a single return value. I already have code I've used this with I guess I can just dumb the testing down.

ye-luo: It is a good thing to record all the failed assertion and report once. If you intended to record more details of matrix or vector explicitly instead of expressing via the message, then `CheckVectorResult` and `CheckMatrixResult` is necessary. If not, a single `CheckResult` seems sufficient.

Regarding a single vs multi return value, If you insist on keeping both bool and message. Please use ways to protect the [corner cases](https://github.com/QMCPACK/qmcpack/pull/5293/commits/acb66d903110252464256d61de8eb44b5e5b0485#r1938003041) which cause confusion. For example, use the constructor to error out the undesired corner cases. Or you can do
```
class checkResult
{
   std::string err_msg;
   isSuccess() const { return err_msg.empty(); }
   const& std::string getErrorMsg() const { return err_msg; };
}
```
You can still make the return typed but not exposing the `empty/non-empty` logic to the caller.

PDoakORNL: It's good to get a committed review but this is utility code for testing and I think this is style disagreement. For test code I think easy to understand, read, and adapt is good. I have plenty of code ready to  PR that is in application code paths and the effort will be more worthwhile there. I'd like to put more effort in there. 

Primarily having the structured return type defined right over the function in the header is simple and I'm not intending to set a consistent return type used by multiple functions. But as you guessed If I do decide I want to add indexes to the return value or other information it is very straight forward to do so and doesn't require touching the other function or existing users of the function. It's coincidental that the return of both of these checks is the same for now.

Our developers that actually write tests have been using checkMatrix without needing my help or anything past the source for a couple of years now. So I think its ok enough, I don't see any misuse or confusion about its usage.

If someone wants to add a message for the success case of either it breaks nothing. Current usage just doesn't do anything with the message when the check passes. Accessing an empty string is also safe, but i can't see why someone would add a new fail with no message but for this code I don't think it needs to be enforced.
So I don't really see an invariant that needs to be defended here.
its seems nice to be able to look at the test and see
```
return {false|true, "message of some sort"}
```
You can easily see what each path means.

I don't want to throw any exceptions just provide a boolean check and some information that can be reported or used for development. This is intended for use in a catch2 test section.

There is nothing wrong with your class I'd just rather do it this way.

ye-luo: As you mentioned checkMatix is being used by other developers, its intended use needs to be well defined. I don't like loose ends and thus raised my questions.

> If someone wants to add a message for the success case of either it breaks nothing.

If the caller receives a return value of checkMatrixResult, `{true, "actually a real error message"}`, some bugs happened. If the caller only checks true or false, it is very likely to miss the opportunity to catch a bug. The best way of preventing bugs is to design the code bug free. This requires bug free concept.

> i can't see why someone would add a new fail with no message

bugs happen when code behaves not as intended. developers usually don't intentionally introduce bugs... but we are humans.

If you agree with me that, true + message and false + no-message are odd cases and should not happen. Please mention it in the class documentation. That makes the return struct well-defined. Since you don't consider the enforcement that important, I'm OK with not doing it. So please update the documentation to set the expectation.

ye-luo:
@PDoakORNL your last commit breaks compilation. A few more renaming `error_msg` to `result_message` are needed in `checkMatrix.hpp`

ye-luo:
Test this please

ye-luo (approved):

PR #5310, opened 2025-02-10
Align applyRotation reduction datatype in the mixed precision case.

ye-luo:
## Proposed changes
Align datatype.

- Bugfix

- No

laptop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
@@ -148,8 +148,8 @@ void SplineC2C<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
           const auto index = Nsplines * i + 2 * k;
           ST zr            = (*coef_copy_)[index];
           ST zi            = (*coef_copy_)[index + 1];
-          ST wr            = rot_mat[k][j].real();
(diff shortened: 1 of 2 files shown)


src/QMCWaveFunctions/BsplineFactory/SplineC2C.cpp
```diff
@@ -141,8 +141,8 @@ void SplineC2C<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_stored_co
         // cur_elem points to the real componend of the coefficient.
         // Imag component is adjacent in memory.
         const auto cur_elem = Nsplines * i + 2 * j;
-        ST newval_r{0.};
-        ST newval_i{0.};
+        RealType newval_r{0.};
+        RealType newval_i{0.};
```
PDoakORNL, line 145: This method doesn't look to work like the R2R one.  There the rot matrix stays double when st is float. So the multiplication is done at full precision.  Here I believe due to the assignments wr, wi the multiplication is done at lower precision and only the accumulation at full precision. So I think you want to make the type of wr and wi RealType as well.

PDoakORNL (changes requested):
this look more like a precision than alignment issue but I think it has a problem. It should also probably have a related unit test to make sure future tampering doesn't reduce the precision from what is being shot for here.

PDoakORNL (approved):

PR #5314, opened 2025-02-12
Move timers after no-op early return in WalkerLog classes

ye-luo:
## Proposed changes
Don't activate timer when features are inactive.

- Bugfix

- No

laptop

## Checklist


prckent (approved):

ye-luo:
Test this please

PR #5327, opened 2025-02-19
fix mixed precision kpoint sorting

PDoakORNL:
## Proposed changes

To get consistent kpoint sorting across compilers and mixed and full precision simulations should rely on full precision for the CrystalLattice and KContainer should do the kpoint sorting in full precision.  Then to preserve mixing precision performance where applicable reduced precision ksq and kpts_cart access via the get*Working calls added to KContainer are available.

This unfortunately touches a huge number of files since There are many tests that directly define CrystalLattice's and they do this either explicitly in mixed precision or by using the `PtclOnLatticeTraits::ParticleLayout` type alias. 

Some deterministic test values for mixed precision are now enough closer to full precision that they fall outside of the tolerance of the tests.  Rather than drop the tolerance I have updated the values when they are closer to full precision that the previous test targets.  This requires many one line updates through CMakeLists.txt in the deterministics tests and it will take some time to complete.

- Bugfix
- New feature
- Build related changes
- Other (please describe):

- Yes
- No

## Checklist


ye-luo:
I guess all the changes under `src/Particle/Lattice` can be PRed separately.

prckent:
I took a quick look at the failures. Offload compilations obviously need fixing. The remaining ionion test failures look very reasonable and come from only a couple of systems. Importantly these are truly deterministic "simple" function evaluations of the fixed input atomic positions => if you wished you could set the mixed and full precision values identical in these cases and bump the tolerances to suit.

prckent:
Test this please

ye-luo:
I noticed that `PtclOnLatticeTraitsT` template was introduced. Even the old class `PtclOnLatticeTraits` was quite questionable. I'd like to see it going away instead of getting further complicated.

I agree with @PDoakORNL mentioned "CrystalLattice and KContainer should do the kpoint sorting in full precision." So I tried making all the use of "CrystalLattice" in full precision https://github.com/QMCPACK/qmcpack/pull/5350
I think we can break changes into smaller pieces.

prckent:
Can this be closed or are there still pieces to merge?

PR #5341, opened 2025-02-26
explicitly instantiate CrystalLattice simplify initialization

PDoakORNL:
@ye-luo suggestion this works as a separate PR.

Build and link CrystalLattice once as opposed to repeatedly due to including implementation into header.
Add accessors to facilitate encapsulation and full precision lattices.

this supports #5327 

- New feature

- No

sdgx server

## Checklist

### DIFF
--- a/src/Particle/Lattice/CrystalLattice.cpp
+++ b/src/Particle/Lattice/CrystalLattice.cpp
@@ -38,15 +38,6 @@ CrystalLattice<T, D>::CrystalLattice()
   reset();
 }
 
-template<typename T, unsigned D>
-template<typename TT>
-CrystalLattice<T, D>::CrystalLattice(const Tensor<TT, D>& lat)
-{
-  explicitly_defined = true;
-  R                  = lat;
-  reset();
-}
-
 template<class T, unsigned D>
 bool CrystalLattice<T, D>::outOfBound(const TinyVector<T, D>& u) const
 {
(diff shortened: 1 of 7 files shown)


src/Particle/Lattice/CrystalLattice.h
```diff
@@ -23,6 +23,7 @@
 #define OHMMS_CRYSTALLATTICE_H
 #include <limits>
 #include <iostream>
+#include <OhmmsData/Libxml2Doc.h>
```
ye-luo, line 26: Is this really needed?

src/Particle/Lattice/CrystalLattice.cpp
```diff
@@ -38,6 +38,25 @@ CrystalLattice<T, D>::CrystalLattice()
   reset();
 }
 
+template<typename T, unsigned D>
+template<typename TT>
+CrystalLattice<T, D>::CrystalLattice(const Tensor<TT, D>& lat)
```
ye-luo, line 43: Is it beneficial to have this constructor? Right now this is a short cut of `CrystalLattice()` + `set`. Due to the way how we parse xml and setup `CrystalLattice` I feel `CrystalLattice()` + `set` is simple enough and sufficient and prefer not to add another code path.

src/Particle/Lattice/CrystalLattice.h
```diff
@@ -280,10 +311,27 @@ struct CrystalLattice : public LRBreakupParameters<T, D>
 
   //! Print out CrystalLattice Data
   void print(std::ostream&, int level = 2) const;
+
+  // Allow assignment operator between say T=double and T=float
+  template<class TT, unsigned DD>
```
ye-luo, line 316: Should the dimension be just `D`?

PDoakORNL (commented):
Dropped attempt to set ground work to kill .set.  For the need for a fully defined xmlNodePtr and a non specialized friend definition, these are both stepping stones.  The knowledge of xmlNodePtr will disappear once LatticeInput exists, and the alternate precision friend class can disappear once there are no reduced precision lattices in the code.

src/Particle/Lattice/CrystalLattice.h
ye-luo: I didn't mean using forward declare. I cannot find any use of xmlNodePtr under `src/Particle/Lattice` and thus feel the actual consumer of xmlNodePtr should have the include line.

src/Particle/Lattice/CrystalLattice.cpp
ye-luo: I also don't like the two pass construction. I also prefer to have a single constructor taking care of all the cases and in this way, I can make a lot of things const safely.

> currently some test code omits calling reset() after manipulating internals which means many elements of the lattice are invalid and hopefully not relevant to the test while providing potentially fatal example code.

I hate those examples. Internals should not be modified and input argument should be minimal and validated.

I just feel it is not a good time to tackle them. When we really tackle them, we should just make input arguments in full precision only and keep constructors simple.

ye-luo (approved):

ye-luo:
Test this please

prckent:
Test this please

Issue #5348, opened 2025-02-28
KSpace jastrows need more implementation


camelto2:
**Is your feature request related to a problem? Please describe.**
Looks like Kspace jastrows are listed in the manual, but they lack a lot of implementation. e.g. they are not currently compatible with the default pseudopotential evaluation that uses virtual particles because it is not implemented. If you switch to algorithm = 'non-batched' to avoid virtual particle set evaluation, the code cannot optimize the coefficients because evaluateDerivRatios or evaluateDerivatiesWF are not implemented. 

**Describe the solution you'd like**
implement kspace jastrow functionality. 

**Describe alternatives you've considered**

**Additional context**

- [x] Implement evaluateDerivativesWF
- [x] add unit tests for gradients, laplacians, parameter derivatives
- [ ] add APIs for batched NLPP algorithm, i.e. evauateRatios and evaluateDerivRatios
- [ ] mw_implementations

prckent:
Hopefully this was not too much of a surprise. This "oversight" stems from there being no test of k-space Jastrow optimization with the batched drivers, so the code paths were never hit. However, there is also no test of k-space Jastrow optimization even with the legacy code. e.g. There is no coverage of kSpaceJastrow::evaluateDerivatives  . Our historical experience is that there could be surprises in the old functionality.

camelto2:
I'm working now adding the derivatives testing, and then will add the other APIs I need to get this optimizable

camelto2:
Some of this is addressed in #5356

camelto2:
There is a path to doing the calculations, which is how those calculations were done. But it currently requires non-batched NonLocalECP evaluation (algorithm = "non-batched" in the input file). Additionally, all the evaluation routines are not using mw_ implementations. The checklist above is up to date regarding what is missing

prckent:
Still relevant. Presumably an LLM can help with the "missing" implementation.

PR #5353, opened 2025-03-02
Add kspace jastrow batched test

prckent:
## Proposed changes

Add batched version of existing short test.

Also, remove offload checks in CMakeLists.txt

In part motivated by #5348 . The new test passes on CPUs.

- No

Ubuntu24, AOCC+AOCL

## Checklist

### DIFF
--- a/tests/solids/diamondC_1x1x1_pp/CMakeLists.txt
+++ b/tests/solids/diamondC_1x1x1_pp/CMakeLists.txt
@@ -309,23 +309,19 @@ qmc_run_and_check(
   LONG_DIAMOND_DMC_ALLP_SCALARS # DMC
 )
 
-if(NOT ENABLE_OFFLOAD)
-  list(APPEND LONG_DIAMOND_KSPACE_SCALARS "totenergy" "-10.500719 0.001769")
-  list(APPEND LONG_DIAMOND_KSPACE_SCALARS "variance" "0.312264 0.028662")
-  qmc_run_and_check(
-    long-diamondC_1x1x1_pp-vmc_sdj_kspace
-    "${qmcpack_SOURCE_DIR}/tests/solids/diamondC_1x1x1_pp"
-    qmc_long_kspace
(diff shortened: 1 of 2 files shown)


prckent:
Test this please

tests/solids/diamondC_1x1x1_pp/CMakeLists.txt
```diff
@@ -74,35 +74,44 @@ qmc_run_and_check(
...
+  DIAMOND_KSPACE_SCALARS # VMC
+)
+
+qmc_run_and_check(
+  short-diamondC_1x1x1_pp-vmc_sdj_kspace
+  "${qmcpack_SOURCE_DIR}/tests/solids/diamondC_1x1x1_pp"
+  qmc_short_kspace_4_4
+  qmc_short_kspace_4_4.in.xml
+  4
+  4
```
ye-luo, line 97: This test seems being added legacy only. If we only intend to have this test either legacy or batch, I think batch is preferred.

prckent: The batched driver test is added at line 103. The ones above are not new, except I changed them to always run. Since the tests are fast, I think we should keep both tests to reduce the chance for problems.

ye-luo: I see. Do you think it is necessary to add this case to batched as well?

prckent: I can if you would like it added. Since the Jastrows involve no parallelism, I generally don't think they add much. Your call.

ye-luo: Please do it in this PR and we won't rediscover missing coverage.

prckent: Added. Also added samples check.

ye-luo (approved):

PR #5342, opened 2025-02-26
LCAO GPU Optimization V1 reduce device pointer look up (OpenMP target map)

anbenali:
## Proposed changes
First set of optimizations allowing for a 2.2X speed up by reducing device pointer lookup via OpenMP.

The LCAO branch was about 3 to 4X slower than CPU. After investigation, It seemed like it was due to map(to:) not checking if data was already on device. So converted all calls to is_device_ptr which fixed the issue. 

Tested on multiple systems, but more specifically "TOU ASN" molecule: O(3) S(1) C(8) N(5) H(17), with 106 electrons and 1185 basis functions. 

 
## What type(s) of changes does this code introduce?
Code Optimization

### Does this introduce a breaking change?

- No

## Checklist

_Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask.  This is
simply a reminder of what we are going to look for before merging your code._

- Yes. This PR is up to date with current the current state of 'develop'
- Yes. Code added or changed in the PR has been clang-formatted
- No. This PR adds tests to cover any new code, or to catch a bug that is being fixed
- No. Documentation has been added (if appropriate)
throughput_comparison

PR #5357, opened 2025-03-05
LCAO Optimization per Center

anbenali:

## Proposed changes

When having multiple atoms of the same species that share the same SoaAtomicBasisSetSoaAtomicBasisSet parameters, grouping them by species (rather than iterating over each center individually) cut down significantly on work and enables larger and more efficient kernels on the GPU

This change evaluates the orbitals per species instead of looping through all centers. For "TOU ASN" molecule: O(3) S(1) C(8) N(5) H(17), with 106 electrons and 1185 basis functions, The cost of the evaluation of the basis is reduced by a factor 8.5 compared to Develop branc and by a factor 4 compared to PR #5342. 

Only evalvgl ported and final analysis of the scaling curve will be added. 

-  Refactoring (no functional changes, no api changes)
-

- No

Aurora. Endymion
## Checklist


anbenali:
> How would we know this is correct?

fixing the seed and running on 1 thread allows for the same E_L for each step. 
With multiThreads on Aurora, agreement is at 10^-5. 
Ideally a unit test would be better.

prckent:
Test this please

src/Numerics/SoaSphericalTensor.h
```diff
@@ -152,21 +151,68 @@ class SoaSphericalTensor
...
+    auto* xyz_ptr = xyz.device_data();
+    auto* Ylm_vgl_ptr = Ylm_vgl.device_data();
+    auto* factorLM_ptr = factorLM_.device_data();
+    auto* factorL_ptr = factorL_.device_data();
+    auto* factor2L_ptr = factor2L_.device_data();
+    auto* norm_factor_ptr = norm_factor_.device_data();
+
+    PRAGMA_OFFLOAD("omp target teams distribute parallel for collapse(3) \
+                    is_device_ptr(factorLM_ptr, factorL_ptr, norm_factor_ptr, factor2L_ptr, xyz_ptr, Ylm_vgl_ptr)")
+    for (size_t i_c = 0; i_c < num_centers; ++i_c) {
```
kgasperich, line 196: is this actually doing anything differently from the existing function?
The loop treats every image/center/elec individually (in the existing function, this is clearer because we don't nest the loops, but it looks to me like this is doing the same thing)

kgasperich: also, it's not clear to me that the labels here are correct; it's a dense rectangular loop, so the result will still be correct, but if centers and images are in the same combined dimension, then no matter how they are ordered, they should come after the nElec loop (again, it doesn't actually do anything incorrect because of the dense packing, but the actual data doesn't seem to correspond to the indices as used here, so if we ever want to do anything that relies on having those correct, then this will be very confusing for someone looking at it in the future)

src/QMCWaveFunctions/LCAO/MultiQuinticSpline1D.h
```diff
@@ -223,6 +222,92 @@ class MultiQuinticSpline1D
...
+
+    auto* coeff_ptr       = coeffs_.device_data();
+    auto* first_deriv_ptr = first_deriv_.device_data();
+    const size_t nCols    = coeffs_.cols();
+    const size_t coefsize = coeffs_.size();
+    const auto nsplines   = num_splines_;
+
+    PRAGMA_OFFLOAD("omp target teams distribute parallel for collapse(3) \
+                    is_device_ptr(first_deriv_ptr, coeff_ptr, r_ptr, u_ptr, du_ptr, d2u_ptr)")
+    for (size_t i_c = 0; i_c < num_centers; ++i_c)
```
kgasperich, line 257: This looks like the same as the batched_evaluateVGL (i.e. loops over nelec/npbc/ncenter can be flattened)

kgasperich (changes requested):
This looks good, but I have some minor requests related to array dimensions and documentation, and it looks like some of the new functions don't actually do anything different from the old ones, so it's probably better to reuse the old ones.

If you do keep the new functions instead of reusing the old ones, then I'd recommend a different naming scheme for the new radial/angular functions:
The SoaAtomic function went from mw_evaluateVGL to mw_evaluateVGL_batch, but the radial/angular ones went from batched_evaluateVGL to batched_evaluateVGL_multiCenter.
Maybe it would be better to just add `_multiCenter` instead of `_batch` for the SoaAtomic one (I think the reason why the components have `batched_` instead of `mw_` is because they just operate on the packed arrays rather than using anything that's explicitly a multiwalker resource)

src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h
```diff
@@ -736,6 +736,220 @@ class SoaAtomicBasisSet
...
+            const int lm = LM_ptr[ib];
+            VT psi = 0, dpsi_x = 0, dpsi_y = 0, dpsi_z = 0, d2psi = 0;
+
+            // Get center-specific offset
+            const size_t center_offset = basis_offsets_ptr[i_c];
+
+            for (int i_xyz = 0; i_xyz < Nxyz; ++i_xyz)
+            {
+              // Calculate global index accounting for center batching
+              const size_t global_idx = i_xyz + Nxyz * (i_e + i_c * nElec);
```
kgasperich, line 903: the ordering here looks like it is assuming a data layout as [num_centers, nElec, Nxyz], but in the array dims above it looked like [nElec, Nxyx * num_centers]

kgasperich: ok, I took a closer look, and it looks like at least the ordering for dr/Ylm and r/Rnl are consistent, so that's why the results are correct, but it still seems to be inconsistent with the way the dimensions are described elsewhere in the code (e.g. when resizing at the top of this function).

kgasperich: you could change the ordering of the loops and the index labels to match the array dimensions above, or you could change the array dimensions, but then they wouldn't agree with how they're described in the multiwalker mem

src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h
```diff
@@ -736,6 +736,220 @@ class SoaAtomicBasisSet
...
+    auto& r       = atom_bs_leader.mw_mem_handle_.getResource().r;
+
+    // Resize arrays for batched processing
+    const size_t nRnl = RnlID.size();
+    const size_t nYlm = Ylm.size();
+
+    ylm_vgl.resize(5, nElec, Nxyz * num_centers, nYlm);
+    rnl_vgl.resize(3, nElec, Nxyz * num_centers, nRnl);
+    dr.resize(nElec, Nxyz * num_centers, 3);
+    r.resize(nElec, Nxyz * num_centers);
```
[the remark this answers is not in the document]
anbenali: It is the best way I found to flatten data without having to create a new dimension

src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h
ye-luo: For optimal performance, the fastest index should match the inner most loop under `parallel for` for maxing out vector performance. However, your inner most loop is `ib` which doesn't affect global_idx and thus, I don't think it matters much. We can defer kernel tuning in separate steps. Just make sure the code is consistent with documentation in the source code.

anbenali: For now it passes CI tests...

anbenali:
> How would we know this is correct?
@prckent : Actually, there was a unit test #29 (deterministic-unit_test_wavefunction_sposet) which caught a bug and was fixed

prckent:
> > How would we know this is correct?
> Actually, there was a unit test #29 (deterministic-unit_test_wavefunction_sposet) which caught a bug and was fixed
Yes, sometimes the system works!

I would use that test as inspiration for tests with different groupings of different species. e.g. Ethanol as CCHHHHHOH vs CHCHOHHHH vs OCCHHHHHH. Results should be identical and it would a useful shakeout of other parts of the LCAO code.

prckent:
Test this please

kgasperich (changes requested):

anbenali:
> > > How would we know this is correct?
> > > Actually, there was a unit test #29 (deterministic-unit_test_wavefunction_sposet) which caught a bug and was fixed
> 
> Yes, sometimes the system works!
> 
> I would use that test as inspiration for tests with different groupings of different species. e.g. Ethanol as CCHHHHHOH vs CHCHOHHHH vs OCCHHHHHH. Results should be identical and it would a useful shakeout of other parts of the LCAO code.

@prckent That would be great indeed but I cannot produce these tests right now. Keep in mind that the code already detects "shuffles" and reorders everything internally before doing anything. This is valid only for ECPs.

src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h
```diff
@@ -925,6 +1136,171 @@ class SoaAtomicBasisSet
     }
   }
 
+	template<typename LAT, typename VT>
```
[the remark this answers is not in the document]
anbenali: Done

src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h
```diff
@@ -736,6 +736,220 @@ class SoaAtomicBasisSet
    * @param [in] NumCenters total number of centers in system (for indexing into displ_list)
    *  
   */
+  template<typename LAT, typename VT>
```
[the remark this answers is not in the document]
anbenali: Done!! Thanks for requesting this... had forgotten to  push the call in evaluateValueVPS to the multicenter version.. that costs 2x less to evaluate that function!! now we are faster than CPU.

src/QMCWaveFunctions/LCAO/SoaLocalizedBasisSet.cpp
```diff
@@ -316,40 +368,98 @@ void SoaLocalizedBasisSet<COT, ORBT>::mw_evaluateValueVPs(const RefVectorWithLea
...
+    offs_list_s.resize(local_offsets[s].size());
+
+    for (size_t i = 0; i < local_centers[s].size(); i++)
+      c_list_s[i] = local_centers[s][i];
+    for (size_t i = 0; i < local_offsets[s].size(); i++)
+      offs_list_s[i] = local_offsets[s][i];
+
+    c_list_s.updateTo();
+    offs_list_s.updateTo();
+  }
```
ye-luo, line 446: Are `species_centers` and `species_offsets` invariant not throughout the run and can be computed once for all the evaluations?
Also would like to rename `species_offsets` to `species_center_coffsets` for better clarity.

ye-luo (commented):
Could you paste the timer under `DiracDeterminantBatched::spovgl`
before and after?

prckent (commented):
Noting that over the weekend I built a PySCF test for FeCO6 with UHF orbitals with the atoms sorted two different ways. The current development branch passed vs the exact mean-field total energies from PySCF :-) . To avoid adding more multi MiB binaries to the repository, I will explore adding these tests as full workflow tests. This would improving testing of the whole PySCF->converter->VMC LCAO chain.

If I haven't got them PRed in before, I'll run those tests on the final version of this PR.

kgasperich (dismissed):
There are a few things that could be clearer regarding array dimensions, but nothing that should affect the correctness of the code, so in the interest of expediency I think this is ok for now.

prckent:
I see two requests that should still be addressed and that would be quick to address: Ye's comments on mw_evaluateVGL_multiCenter and the timer around DiracDeterminantBatched::spovgl .

prckent:
I see some references to MKL sparse appearing. These need abstracting. To get this PR merged promptly please don't do too much in the single PR. It is best not to mix efforts. This was going to be mergeable with just a couple of trivial updates. I suggest reverting to the status from when I commented and just fixing the "mw_evaluateVGL_multiCenter and the timer around DiracDeterminantBatched::spovgl ."

prckent:
Also, you don't need to keep merging with develop here. We are keeping away from the LCAO code to avoid risks of conflicts. Of course, the longer the branch is up the greater the risk...

ye-luo:
@anbenali could you address the renaming and open an issue for the optimization I suggested if you won't address it in the current PR. I believe this PR can be merged very quickly.

anbenali:
Hi @ye-luo 
I am buried under thr pressure of the GB working on the forces. I won't be able to do anything until the 15 of April.. sorry.  I agree with all the optimizations you suggested but they will really take me too much time to implement before.

src/QMCWaveFunctions/LCAO/SoaAtomicBasisSet.h
```diff
@@ -718,24 +718,235 @@ class SoaAtomicBasisSet
...
+  * @param[in] Tv_list          Phase translation vectors for all centers and walkers
+  *                             (layout: 3 * NumCenters * nWalkers)
+  * @param[in] nWalkers         Number of walkers (P_list.size())
+  * @param[in] nBasTot          Total number of basis functions in the system
+  * @param[in] c_list           List of center indices for this atomic species
+  * @param[in] basis_offsets    Basis function offsets for each center in c_list
+  * @param[in] NumCenters       Total number of centers in the system (for bounds checking)
+ */
+  template<typename LAT, typename VT>
+  inline void mw_evaluateVGL_batch(const RefVectorWithLeader<SoaAtomicBasisSet>& atom_bs_list,
```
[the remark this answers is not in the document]
anbenali: done

ye-luo:
> @ye-luo if you still want it, can you paste the requested timer under DiracDeterminantBatched::spovg? Or we can merge as is and update in a later PR. The other requests have all been addressed so I think that this can be merged.

My request for timings is mostly informational. This PR is good to go.

ye-luo (approved):

ye-luo:
Test this please

Issue #5283, opened 2025-01-15
Segmentation error in Polaris for SOC-QMC calculations


kayahans:
**Describe the bug**
DMC-SOC run produces segmentation error on Polaris, does not give any feedback to the user what might be wrong about the input. QMCPACK output does not give any indication of insufficient memory from the output, and I have tested the same run with reduced meshfactors to see if the job will go through, but none of those jobs succeeded. The same job with more demanding parameters run successfully on Baseline (OLCF) with no issues. 

Here are the last few lines from the output file with meshfactor=1.0 using debug queue (1 node 4 MPI ranks, minimal setup) in Polaris: 

```
=========================================================
  Start VMCBatched
  File Root dmc.s000
=========================================================
==============================================================
--- Memory usage report : VMCBatched before initialization ---
==============================================================
Available memory on node 0, free + buffers :  178134 MiB
Memory footprint by rank 0 on node 0       :   61632 MiB
Device memory allocated via OpenMP offload :       0 MiB
Device memory allocated via CUDA allocator :       0 MiB
Free memory on the default device          :   38648 MiB
==============================================================
VMCBatched Driver running with
             total_walkers     = 512
             walkers_per_rank  = [128(x4)]
             num_crowds        = 8
  on rank 0, walkers_per_crowd = [16(x8)]

                         steps = 1
                        blocks = 5

===================================================================
--- Memory usage report : VMCBatched after initialLogEvaluation ---
===================================================================
Available memory on node 0, free + buffers :  176221 MiB
Memory footprint by rank 0 on node 0       :   62045 MiB
Device memory allocated via OpenMP offload :     127 MiB
Device memory allocated via CUDA allocator :       0 MiB
Free memory on the default device          :   38366 MiB
===================================================================
```

I have tried reducing meshfactor of the splines, but it made no effect to the outcome. 

dmc.err file is the following:

```

Lmod is automatically replacing "nvhpc/23.9" with "gcc-native/12.3".


Due to MODULEPATH changes, the following have been reloaded:
  1) cray-mpich/8.1.28

QMCPACK ERROR Primitive cell ion 0 vs supercell ion 0 atomic number not matching: 0 vs 75
QMCPACK ERROR Primitive cell ion 1 vs supercell ion 1 atomic number not matching: 0 vs 17
QMCPACK ERROR Primitive cell ion 2 vs supercell ion 2 atomic number not matching: 0 vs 17
[...]

PR #5360, opened 2025-03-06
Fix mw_evalGradWithSpin/mw_ratioGradWithSpin fallback.

### LINKED ISSUE #5358: Segmentation error with SOC calculations using GPU
**Describe the bug**
SOC calculations terminate with a segmentation error at the first VMC step with complex GPU executable. Until the VMC step, the code executes without any error. 

CPU executable works fine though: I used a similar input (changes are only in walker distribution over ranks) to run with the CPU executable, it didn't have any issues in VMC or DMC steps. I have accumulated more than 20k steps in DMC with CPU executable using 128 nodes in Aurora. 

The GPU run fails on both Frontier and Aurora. I had posted a similar issue with segmentation error on Polaris #5283 on the same material system, the issues could be related. 

Here are the final lines from the output file in Frontier

Image

Here is the output file:

Image

**To Reproduce**
Use the input files provided in 
1. In Aurora: /lus/flare/projects/PSFMat_2/issue_5358
2. In Frontier: /lustre/orion/mat151/proj-shared/issue_5358

These are very quick runs with small number of walkers and using a single node in debug queues. If you need a more demanding setup, please let me know. I had the same outcome irrespective of the number of nodes. 

Frontier is using the latest recipe for compilation from last month. 

**Expected behavior**
Code should exit cleanly instead of giving a segmentation error. 

**System:**
 - Frontier, Aurora

ye-luo:
## Proposed changes
Fixes #5358 and maybe #5283?
WFC::mw_eval/ratioGradWithSpin falling back to WFC::eval/ratioGradWithSpin causes mixing batched and non-batched API calls. proposed the move using non-batched API and then crash in the mw_accept. To prevent this, default fallback should be WFC::mw_eval/ratioGrad.

With this change, spinor aware classes should implement both WFC::mw_eval/ratioGradWithSpin and WFC::eval/ratioGradWithSpin although the batched implementation can be just a loop.

DummyDiracDetWithoutMW class in the unit test is removed due to not satisfying the above requirement.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
+++ b/src/QMCWaveFunctions/Fermion/DiracDeterminant.cpp
@@ -158,6 +158,21 @@ typename DiracDeterminant<DU_TYPE>::GradType DiracDeterminant<DU_TYPE>::evalGrad
 }
 
 template<typename DU_TYPE>
+void DiracDeterminant<DU_TYPE>::mw_evalGradWithSpin(const RefVectorWithLeader<WaveFunctionComponent>& wfc_list,
+                                                    const RefVectorWithLeader<ParticleSet>& p_list,
+                                                    int iat,
+                                                    std::vector<GradType>& grad_now,
+                                                    std::vector<ComplexType>& spingrad_now) const
+{
+  assert(this == &wfc_list.getLeader());
+  for (int iw = 0; iw < wfc_list.size(); iw++)
+  {
+    spingrad_now[iw] = 0;
+    grad_now[iw]     = wfc_list[iw].evalGradWithSpin(p_list[iw], iat, spingrad_now[iw]);
+  }
+}
+
+template<typename DU_TYPE>
 typename DiracDeterminant<DU_TYPE>::PsiValue DiracDeterminant<DU_TYPE>::ratioGrad(ParticleSet& P,
                                                                                   int iat,
                                                                                   GradType& grad_iat)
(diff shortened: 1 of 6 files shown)


ye-luo:
@kylanpaa please test both https://github.com/QMCPACK/qmcpack/issues/5358 and https://github.com/QMCPACK/qmcpack/issues/5283 and let us know if this fixes your reproducers.

ye-luo:
Test this please

ye-luo:
Test this please

kayahans:
> @kylanpaa please test both #5358 and #5283 and let us know if this fixes your reproducers.

@ye-luo thanks for looking into this. Do you want me or @kylanpaa to test this? I posted #5358 and #5283

ye-luo:
> > @kylanpaa please test both #5358 and #5283 and let us know if this fixes your reproducers.
> 
> @ye-luo thanks for looking into this. Do you want me or @kylanpaa to test this? I posted #5358 and #5283

Ops. This is for you not @kylanpaa

prckent:
@kayahans I'll be happy to merge this if you can confirm the immediate crash is gone.

prckent:
@camelto2 , @ye-luo Why was this crash and mixing of APIs not caught before?

kayahans:
> @kayahans I'll be happy to merge this if you can confirm the immediate crash is gone.

Thanks Paul I will get to it before this afternoon

camelto2:
> @camelto2 , @ye-luo Why was this crash and mixing of APIs not caught before?

This likely wasn't caught because I haven't done many runs on GPUs using SOC. The CPU code has been used in production in the batched code, but I think these are the first attempts to genuinely run the SOC code on GPUs. 
When I did previously run on GPUs, I was mainly only doing some testing of DiracDeterminantBatched::mw_XYZ. DiracDeterminantBatched already had explicit implementations, so it wouldn't hit the fallback and trigger the segfault. I'm curious why the runs in #5358 and #5283 were hitting the fallback in the first place and not doing through DiracDeterminantBatched. 

When I implemented this originally, I knew the fallback needed to be a loop over single walker APIs, so I originally put it in WFC so that any component that didn't explicitly implement their own version would use the corresponding single walker APIs. That avoided having to add explicit implementations that are loops in every class (which is apparently the right thing to do). But I didn't realize that would cause this mismatch between sw and mw APIs and that it would lead to issues.

camelto2 (approved):
LGTM

prckent (approved):

prckent:
Test this please

ye-luo:
If we add tests of  mw_ratioGrad and mw_accept. This issue can be caught.
https://github.com/QMCPACK/qmcpack/blob/29872488bde736ad99047606ed87445f941835fd/src/QMCWaveFunctions/tests/test_TrialWaveFunction.cpp#L427

PR #5371, opened 2025-03-10
Nexus: ORNL Baseline with partition information

kayahans:
## Proposed changes

This PR adds the partition information to recently added ORNL Baseline cluster. Different # of nodes and wall time limitations based on the queues used are enforced. This implementation can also be applicable to other supercomputers defined in machine.py using parent "validate_queue_config" function under "write_job_header" and defining "queue_config" as member variable (similar to Baseline). 

- New feature

- No

MacOS 15.3.1
## Checklist


ye-luo (dismissed):

prckent (changes requested):
Minor point: According to `scontrol show partitions` the maximum schedulable partition is 138 nodes. Please revise your 128 node figure unless you have other info.

prckent:
Thanks Kayahan, this is helpful.

Extra question: @jtkrogel  mentioned to me last week that some of the other machines have logic around projects, resources etc. Can any other machines be upgraded to use the new verification code and avoid code duplication?

kayahans:
> Minor point: According to `scontrol show partitions` the maximum schedulable partition is 138 nodes. Please revise your 128 node figure unless you have other info.

@prckent I think 138 is the accurate number, I used the information at their website: https://docs.cades.olcf.ornl.gov/baseline_user_guide/baseline_user_guide.html#login-nodes which needs to be updated.

kayahans:
> Thanks Kayahan, this is helpful.
> 
> Extra question: @jtkrogel mentioned to me last week that some of the other machines have logic around projects, resources etc. Can any other machines be upgraded to use the new verification code and avoid code duplication?

Yes, other machines can be upgraded to use this verification code. This only does queue based comparison, for project based limitations ideally it would be better to introduce that under `Settings` class in nexus.py.

kayahans:
@prckent I worked on this a bit more to improve the queue validation. Most machines that already have queue configuration in the machine.py file are already decommissioned. I think only Perlmutter can benefit from such refactoring. I made a pass through each supercomputer listed and commented on their status based on the information I can find about them online.

prckent:
Thanks Kayahan, will take a look. Can you confirm all the tests are now working for you?

kayahans:
@prckent with the last update it is passing all the tests. Main reason it didn't pass was the following under write_job_header function in baseline: 
```
        if job.queue is None:
            job.queue = 'batch_cnms'
        #end if
```
This enforces default queue. If `queue_configs` is used, the default queue should also be defined there.

prckent (approved):
Thank you Kayahan

prckent:
Test this please

PR #4849, opened 2023-11-30
Refactor SplineSetReader and HybridRepSetReader

ye-luo:
## Proposed changes
Avoid multi level inheritance.  SplineSetReader is no more a base of HybridRepSetReader.
Make most functions const and kill as much state machines as possible.

## What type(s) of changes does this code introduce?
- Refactoring (no functional changes, no api changes)

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
epyc-server

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'

PR #5381, opened 2025-03-15
Spline dump should only be read by rank 0.

ye-luo:
## Proposed changes
Fixed a bug introduced in https://github.com/QMCPACK/qmcpack/pull/4849/files
It caused reading the same dump file from every MPI rank.

- Bugfix

- No

Aurora

## Checklist


prckent:
Test this please

prckent (approved):

PR #5386, opened 2025-03-19
Nexus: support for stochastic reconfiguration and basic orbital rotation

jtkrogel:
## Proposed changes

Add support for features Cody's project needs:
- stochastic reconfiguration optimizer inputs
- `<rotate_spo>` wrapper around `<sposet>`
- improved k-space jastrow generation

- New feature

- No

Laptop

## Checklist

### DIFF
--- a/nexus/lib/qmcpack_input.py
+++ b/nexus/lib/qmcpack_input.py
@@ -1834,9 +1834,15 @@ class sposet(QIxml):
                   'source','version','precision','tilematrix',
                   'meshfactor']
     elements   = ['occupation','coefficient','coefficients']
-    text       = 'spos'
+    #text       = 'spos'
     identifier = 'name'
 #end class sposet
+
+class rotated_sposet(QIxml):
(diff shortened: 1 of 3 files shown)


nexus/lib/qmcpack_input.py
```diff
@@ -1834,17 +1834,23 @@ class sposet(QIxml):
                   'source','version','precision','tilematrix',
                   'meshfactor']
     elements   = ['occupation','coefficient','coefficients']
-    text       = 'spos'
+    #text       = 'spos'
```
prckent, line 1837: Why did this need to be commented?

jtkrogel: This relates to the body text of a qmcpack input element.  Sposets have no body text by default (i.e. not <sposet> spos </sposet>.  It's an odd holdover that doesn't have any effect I can tell, but is formally incorrect.

prckent (approved):

Issue #5384, opened 2025-03-18
Long range unit test failures


prckent:
**Describe the bug**
e.g. [deterministic-unit_test_longrange](https://cdash.qmcpack.org/testSummary.php?project=1&name=deterministic-unit_test_longrange&date=2025-03-18)
https://cdash.qmcpack.org/tests/18235080

There was some slop in the k-shell counts built into the tests that needs updating and ideally narrowing. At least Mixed debug gcc14 builds are problematic

**To Reproduce**
Run unit tests. Found via nightly testing

**Expected behavior**
 Tests pass

**System:**
nitrogen2 mixed precision gcc14

ye-luo:
`CHECK( (std::is_same<float, double>::value ? handler.MaxKshell == 78 : handler.MaxKshell >= 117 && handler.MaxKshell <= 128 ) )`
The check reference `[117,128]` probably needs adjustments.

prckent:
Fixed in #5389

PR #5382, opened 2025-03-16
Fix LongRange unit test issue and adjust MPC reference value.

ye-luo:
## Proposed changes
I took a close look at the issue raised by https://github.com/QMCPACK/qmcpack/issues/5374
It seems to be `-ffast-math` related when compiling `src/Particle/LongRange/tests/test_StructFact.cpp`.
I found a workaround by doing reduction in double precision in the unit test.
This PR also fixes https://github.com/QMCPACK/qmcpack/issues/5377

## What type(s) of changes does this code introduce?
- Bugfix

### Does this introduce a breaking change?
- No

## What systems has this change been tested on?
epyc-server

## Checklist
- Yes. This PR is up to date with current the current state of 'develop'

PR #5303, opened 2025-02-06
Structure Factor estimator at unit test stage

PDoakORNL:
Proposed changes

This add the actual batched Structure Factor Estimator. It should cover the functionality of the skAll and static structure factort estimators from the legacy code. This is just the estimator and unit tests it does not include integration with the EstimatorManagerNew or processing of structurefactor input at the application level.

That will be added in a following PRs.
What type(s) of changes does this code introduce?

Delete the items that do not apply

    New feature
    Build related changes
    Testing changes (e.g. new unit/integration/performance tests)
    Documentation or build script changes

Does this introduce a breaking change?

    No

What systems has this change been tested on?

sdgx2
Checklist

Update the following with a yes where the items apply. If you're unsure about any of them, don't hesitate to ask. This is
simply a reminder of what we are going to look for before merging your code.

    Yes This PR is up to date with current the current state of 'develop'
    Yes Code added or changed in the PR has been clang-formatted
    Yes This PR adds tests to cover any new code, or to catch a bug that is being fixed
    Yes Documentation has been added (if appropriate)

PR #5389, opened 2025-03-20
kcontainer now encapsulated and provides k elements at working prec

PDoakORNL:
## Proposed changes

The important change is By providing accessors for working precision ksq and kcart in `KContainer` we continue to support low precision operations  for mixed precision and full precision for full precision builds without having to branch at run time.

And KContainer now does its internal kshell sorting in full precision and previous PR's made the Lattice always full precision.  This removes the ambiguity in number of kshells and which kpoints are members of which shells that mixed precision previously suffered from. This unblocks consistent unit testing that is dependent on kshell sorting.  For instance for the Structure Factor estimator. Incidentally the number of kshells reported for mixed precision also becomes consistent for many other unit tests.

In order to have to compiler help direct member access that possibly returns incorrect precision KContainer made an actual class and accessors are also established for other members previously directly accessed.

- Bugfix
- Refactoring (no functional changes＿
- Build related change）
- Other (please describe):

- Yes

## Checklist


prckent:
Happily it appears that this will likely fix #5384 . For the two HEG test failures in the CI I would simply bump the test tolerance:

```
27/895 Test  #143: deterministic-heg_14_gamma-sj-batch-1-1-potential .................................................***Failed    0.07 sec
Tests for series 0
  Testing quantity: LocalPotential
    reference mean value     :  -2.02361764
    reference error bar      :   0.00002000
    computed  mean value     :  -2.02367767
    computed  error bar      :   0.00000000
    pass tolerance           :   0.00006000  (  3.00000000 sigma)
    deviation from reference :  -0.00006003  ( -3.00160500 sigma)
    error bar of deviation   :   0.00002000
    significance probability :   0.99731426  (gaussian statistics)
```

prckent:
Test this please

PDoakORNL:
test this please

prckent (dismissed):
LGTM. (Will still need a non-ORNL merge).

Thanks Peter. I have a couple of minor points, but I don't think they are worth holding up the PR for: first, there is some commented code that would be nice to remove. second, the "Working" data might be more clearly named "WorkingPrec" , or at least a few more comments to this effect added.

src/Particle/LongRange/tests/test_StructFact.cpp
```diff
@@ -48,22 +48,22 @@ TEST_CASE("StructFact", "[lrhandler]")
...
-  CHECK(sk.rhok_r.cols() == simulation_cell.getKLists().numk);
-  CHECK(sk.rhok_i.cols() == simulation_cell.getKLists().numk);
+  CHECK(sk.rhok_r.cols() == simulation_cell.getKLists().getNumK());
+  CHECK(sk.rhok_i.cols() == simulation_cell.getKLists().getNumK());
 
   std::vector<std::complex<double>> rhok_sum_ref{-125.80618630936, 68.199075127271};
 
   for (int i = 0; i < ref.groups(); i++)
   {
-    std::complex<double> rhok_sum, rhok_even_sum;
```
ye-luo, line 64: My fix https://github.com/QMCPACK/qmcpack/pull/5382/files needs to be reinstated .

ye-luo:
There are a lot of changes involved with replacing direct member access with accessor functions. That makes reviewing the actual change difficult. Could you make a list of bullet points highlighting functional changes?
second questions, how much timings are changed in Coulomb terms in performance tests?

PDoakORNL:
Timings should be practically unchanged the cost of initializing the cached low precision values of ksq and kcart is outside of the particle move loop.  There is a slight memory cost in mixed precision for the lowered precision ksq and kcart vectors.

PDoakORNL:
test this please

src/Utilities/tests/for_testing/test_checkVector.cpp
```diff
@@ -10,6 +10,7 @@
 //////////////////////////////////////////////////////////////////////////////////////
 
 #include "catch.hpp"
+#include "OhmmsPETE/TinyVector.h"
```
ye-luo, line 13: I don't feel it is needed here. Can it be removed?

ye-luo (commented):
Please make a separate PR with all the changes under `Utilities`. It should simplify the review process.

src/Utilities/tests/for_testing/test_checkVector.cpp
PDoakORNL: Done

PDoakORNL: This one can't be removed.  Either it has to be here or in checkVector.hpp and I remember a review comment removing it from that file.

PDoakORNL:
this is going to have a clash with #5410 which I need to look at after that merges so don't merge.

src/Particle/LongRange/KContainer.h
```diff
@@ -26,16 +27,57 @@ namespace qmcplusplus
...
+template<typename REAL>
+class KContainerT
 {
+public:
+  using Real               = REAL;
+  using FullPrecReal       = QMCTraits::FullPrecRealType;
+  using PositionFull       = QMCTraits::QTFull::PosType;
+  static constexpr int DIM = OHMMS_DIM;
+  using AppPosition        = typename QMCTraits::PosType;
+  using Position           = typename QMCTypes<Real, DIM>::PosType;
```
ye-luo, line 39: With this change, there are three precision in this class.
`REAL`, `QMCTraits::FullPrecRealType` and `QMCTraits::PosType` (effectively `QMCTraits::RealType`)
This seems defeating the purpose of templates.

ye-luo:
Trying a simple fix https://github.com/QMCPACK/qmcpack/pull/5411
If it works, we can avoid convoluted precisions in KContainer.
I still think making `KContainer` template useful. It affects StructFact which does benefit from lower precision. Using template allows us enabling precision selection at runtime although this feature is considered low priority.

There are a lot of changes in this PR making KContainer comply to our coding standard. That is a very good thing to do. Unfortunately they made code review very hard. I'd recommend making such types of change separately from actual functional changes. It is unclear to me if we can still cherry-pick them or just redo them from scratch.

PDoakORNL:
I think you are being unecessarily restrictive with this and having just had to write an midterm EPM I am accutely aware of how much my performance is effected by our code review process.  @prckent please resolve this I am not initerested in revising this again and I do not think it is excessively difficult to review.

ye-luo:
I did a very careful review of the whole code and I expressed my concern https://github.com/QMCPACK/qmcpack/pull/5389#discussion_r2024110268 of the key functional change in this PR. Regarding your KPI, that is an offline discussion. If all the assessor changes were in a separate PRs, such PRs should have been merged already.

prckent:
I'll take a proper look this afternoon. Dev velocity is a key point and also a concern of mine. I tend to think we should merge this since the tests pass and it is clearly a step in the right direction. No need to keep chopping it up.

ye-luo:
> I'll take a proper look this afternoon. Dev velocity is a key point and also a concern of mine. I tend to think we should merge this since the tests pass and it is clearly a step in the right direction. No need to keep chopping it up.

I don't advice chopping it up. Instead, just drop this PR and add `accessor/private` in the develop. That seems sufficient to get other dependent PRs to proceed.

ye-luo:
Let me explain my technical assessment of this PR. I raised the [concern](https://github.com/QMCPACK/qmcpack/pull/5389#discussion_r2024110268), the key functional change of this PR is flawed and made the code further complicated. To address this flaw, extra effort will be needed to clean up the over complicated precision handling if we merge this PR and I'm afraid it's not worth it. The alternative route will be actually [using an easily understandable fix](https://github.com/QMCPACK/qmcpack/pull/5411) of the k sorting. Then do minimal fix (adding a few access functions) to enable structure factor estimator #5303/#5406. They are quite self contained. This alternative route has one drawback that KContainer is not made template. Basically we need to defer the template change.

My understanding that the urgent work is to enable estimators not making KContainers template. Taking the alternative route allows to make the needed fix and we can focus working on the actual needed code.

prckent (approved):
I am happy to merge this as is. Let me know your preference @PDoakORNL .

ye-luo:
Test this please

PR #5406, opened 2025-03-31
[WIP] Batched Structure Factor estimator with unit tests

PDoakORNL:
## Proposed changes

This adds the actual batched Structure Factor Estimator. It should cover the functionality of the skAll and static structure factort estimators from the legacy code. This is just the estimator and unit tests it does not include integration with the EstimatorManagerNew or processing of structurefactor input at the application level.

- New feature
- Other (please describe):

- No

sdgx2

## Checklist


src/Utilities/for_testing/NativeInitializerPrint.hpp
```diff
@@ -95,7 +95,10 @@ inline std::ostream& operator<<(std::ostream& out, const NativePrint<Vector<T>>&
   out << "{ ";
   auto vec = np_vec.get_obj();
   for (T& t : vec)
-    out << std::setprecision(10) << t << ", ";
+    if constexpr (IsComplex_t<T>::value)
+      out << std::setprecision(10) << '{' << t.real() << ", " << t.imag() << "}, ";
+    else
+      out << std::setprecision(10) << t << ", ";
```
ye-luo, line 101: I don't quite understand this change. `<<` does support `std::complex`.
The added code doesn't compile if `T` is not `std::complex` because IIRC constexpr still checks syntax, and a simple T=double doesn't support `real()/imag()`.

ye-luo: I was wrong. Template is special

"If a constexpr if statement appears inside a [templated entity](https://en.cppreference.com/w/cpp/language/templates#Templated_entity), and if condition is not [value-dependent](https://en.cppreference.com/w/cpp/language/dependent_name#Value-dependent_expressions) after instantiation, the discarded statement is not instantiated when the enclosing template is instantiated."

https://en.cppreference.com/w/cpp/language/if

ye-luo: Why the [default style](https://en.cppreference.com/w/cpp/numeric/complex/operator_ltltgtgt) doesn't work for you?
Could you still add a comment explaining the reason for this specialization?

ye-luo:
@PDoakORNL any idea what causes failing in all the full precision builds?

PDoakORNL:
Will continue this is private

PR #5408, opened 2025-04-01
Replace NULL with nullptr in test_parser.cpp

ye-luo:
## Proposed changes
Got sporadic failure in test_parser.cpp
https://github.com/QMCPACK/qmcpack/actions/runs/14202948158/job/39793929557#step:7:2097
Try this fix.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/Utilities/tests/test_parser.cpp
+++ b/src/Utilities/tests/test_parser.cpp
@@ -123,11 +123,11 @@ TEST_CASE("readLine", "[utilities]")
         REQUIRE(buf == tc.output[i]);
         if (i == tc.output.size() - 1)
         {
-          REQUIRE(out == NULL);
+          REQUIRE(out == nullptr);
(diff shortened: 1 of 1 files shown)


ye-luo:
Test this please

ye-luo:
@PDoakORNL I just noticed that you actually made the same change at https://github.com/QMCPACK/qmcpack/pull/5389/files#diff-563133da645cd1665588b52c6b5bd248545f0cd97a84e6ff77f00f0652f8acfc
If you could upstream your change in a separate PR, my PR won't be needed and we can reduce that pending PR.

prckent:
Worth trying. If https://github.com/QMCPACK/qmcpack/pull/5389 can be merged as is we can simply watch. centos failures of that unit test have been sporadic and a puzzle.

PR #5412, opened 2025-04-02
Suggest walker totals to achieve VMC load balance

prckent:
## Proposed changes

Help the user by suggesting a walker count that would be load balanced when imbalance is detected.

- New feature

- No

sulfur, gcc14. A few by-hand MPI task and OpenMP thread count settings.

## Checklist

### DIFF
--- a/src/QMCDrivers/QMCDriverNew.cpp
+++ b/src/QMCDrivers/QMCDriverNew.cpp
@@ -405,7 +405,8 @@ QMCDriverNew::AdjustedWalkerCounts QMCDriverNew::adjustGlobalWalkerCount(Communi
                   << num_crowds << "). This will result in a loss of efficiency.\n";
 
   if (awc.global_walkers % num_ranks || awc.walkers_per_rank[rank_id] % num_crowds)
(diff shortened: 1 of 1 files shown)


src/QMCDrivers/QMCDriverNew.cpp
```diff
@@ -404,6 +404,11 @@ QMCDriverNew::AdjustedWalkerCounts QMCDriverNew::adjustGlobalWalkerCount(Communi
     app_warning() << "Walkers per rank (" << awc.walkers_per_rank[rank_id] << ") not divisible by number of crowds ("
                   << num_crowds << "). This will result in a loss of efficiency.\n";
 
+  if (awc.global_walkers % num_ranks || awc.walkers_per_rank[rank_id] % num_crowds)
```
ye-luo, line 407: `awc.global_walkers % (num_ranks * num_crowds)` instead?

prckent: I prefer to copy the exact logic used in the lines above for consistency. i.e. If either of those warnings are triggered then this new printout clearly will be. Makes for easier reading.

ye-luo: I found the condition in the if statement wasn't aligned with the warning. I feel it should.

ye-luo (approved):

prckent:
Test this please

PR #5416, opened 2025-04-03
Nexus: adding BES-MS machine 

kayahans:
This PR adds the BES-MS (or-login01.ornl.gov) machine configuration to Nexus and also adds the compilation script. 

- New feature

- No

BES-MS (or-login01.ornl.gov) 

## Checklist


prckent (changes requested):
Look for my email.

kayahans:
Thanks @prckent, I must have fixed the error in test_machines module. 

On a separate issue I found that in BES-MS due to the way openmpi is installed, some of the other test modules fail. Starting with the test_simulation_module.py file in the test_execute function I get the folliowing output from BESMS: 

"This version of Spack (openmpi ~legacylaunchers schedulers=slurm) \nis installed without the mpiexec/mpirun commands to prevent \nunintended performance issues. See https://github.com/spack/spack/pull/10340 \nfor more details.\nIf you understand the potential consequences of a misconfigured mpirun, you can\nuse spack to install 'openmpi+legacylaunchers' to restore the executables.\nOtherwise, use srun to launch your MPI executables."

What I understand from this is this installation does not allow mpirun to be executed directly. My guess is that most of the tests in Nexus rely on being able to execute mpirun directly. I could try to fix it on a separate PR if you think that could be an issue on larger set of supercomputers.

prckent:
Re: MPI launchers. Sounds like you should be using srun for this machine and not mpirun. This error should not be any different from running jobs "by hand". If it is then perhaps you are missing a module load or similar environment setup. You could also ask the sysadmins to install/use the legacylaunchers option, but they likely have their reasons.

kayahans:
@prckent, I unloaded the opnmpi module and then all the test have passed, except for qdens_radial test which had the following error: 

Test name     : qdens_radial
Test sublabel : test_radial_density
Test exception: "AssertionError: Executed system command failed.

              Command:
              ========
              /home/ksu/SOFTWARE/ks_qmcpack/qmcpack/nexus/bin/qdens-radial -s C -r 0.6 /home/ksu/SOFTWARE/ks_qmcpack/qmcpack/nexus/tests/unit/test_qdens_radial_output/test_radial_density/vmc/vmc.avg.s000.SpinDensity_u+d.xsf

              stdout:
              =======


              stderr:
              =======
                File "/home/ksu/SOFTWARE/ks_qmcpack/qmcpack/nexus/bin/qdens-radial", line 534, in <module>
                  qdensradial.process()
                File "/home/ksu/SOFTWARE/ks_qmcpack/qmcpack/nexus/bin/qdens-radial", line 454, in process
                  equiv_atoms = list(cd.info.structure.equivalent_atoms().keys())
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                File "/home/ksu/SOFTWARE/ks_qmcpack/qmcpack/nexus/lib/structure.py", line 5311, in equivalent_atoms
                  for e,sn in zip(self.elem,ds.equivalent_atoms):
                                            ^^^^^^^^^^^^^^^^^^^
              AttributeError: 'obj' object has no attribute 'equivalent_atoms'


              Return code:
              ============
              1

However, this error must be related to spglib version. I used spglib 2.5.0 while at the time the nexus symmetry code was written it must have used spglib 1.9.9. This would need to be addressed in a separate PR.

prckent:
Yes, NEXUS is broken with current releases of spglib. I think even from v2. Likely a small fix to bring it up to date.

Do you want us to merge this as is? Only a limited number of people will be using this machine, so this is up to you.

prckent (approved):

prckent:
Test this please

PR #5423, opened 2025-04-05
Autodetect number of threads

latkam:

## Proposed changes

- autodetect number of threads where suitable to speed up builds and CI time

No

## Checklist


ye-luo (changes requested):
There can be cgroup resource restrictions on the login nodes of HPC intended for fair use of cores/memory/filesystem. Maximizing parallel builds is actually harmful in many cases.

prckent:
Thanks Latkam. Your are right that these builds could be faster, but, as Ye noted, most of these are shared machines where it ranges from being impolite to being against the rules to use all the cores.  Using all the cores &/or switching to ninja is recommended on machines where you can use them. I have benchmarked speedups up to at least 256 cores on systems with high performance filesystems.

PR #5427, opened 2025-04-07
Nexus: QE HDF5 charge density results support

kayahans:
When QE is installed with HDF5 support(default to compile pw2qmcpack.x), instead of the charge-density.dat and spin-polarization.dat files, a single charge_density.hdf5 file is produced. Currently Nexus does not support transferring results between simulation objects with hdf5 extensions. 

I have noticed that while creating these softlinks, Nexus does not actually check if the .dat file exists in the reference directory. I initially tried implementing these checks, but it broke several tests. New code checks if there is a charge-density.hdf5 file in the reference directory, if not then assumes there is a .dat file (which is the current behavior). 

- New feature

- No. Passes all tests in nxs-test.

Python 3.12.8

## Checklist

### DIFF
--- a/nexus/lib/pwscf.py
+++ b/nexus/lib/pwscf.py
@@ -184,8 +184,18 @@ class Pwscf(Simulation):
         if result_name=='charge_density' or result_name=='restart':
             result.locdir   = self.locdir
             result.outdir   = os.path.join(self.locdir,outdir)
-            result.location = os.path.join(self.locdir,outdir,prefix+'.save','charge-density.dat')
-            result.spin_location = os.path.join(self.locdir,outdir,prefix+'.save','spin-polarization.dat')
+            result_save_outdir = os.path.join(self.locdir,outdir,prefix+'.save')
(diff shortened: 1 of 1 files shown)


prckent:
Thanks Kayahan. Can you please confirm that a "meaningful" QE spin polarized workflow is still good with the latest QE? e.g. Check https://github.com/QMCPACK/qmcpack/blob/develop/nexus/examples/qmcpack/rsqmc_quantum_espresso/04_iron_dft_dmc_gcta/iron_ldaU_dmc_gcta.py works, or at least looks to work (no need to run the whole thing). Anything similar would be fine, potentially what you used to develop this. Unfortunately there is no meaningful testing on Nexus workflows yet.

kayahans:
@prckent, I have tested this on a non-collinear, spin-orbit coupling calculations workflow using Nexus. Additionally, I ran the QE part of the workflow you shared and that worked for me as well.

prckent (approved):
LGTM

ye-luo:
Test this please

PR #5428, opened 2025-04-08
[WIP] Add type sanitizer support

prckent:
## Proposed changes

Expand sanitizer support. Basic plumbing to allow experiments https://clang.llvm.org/docs/TypeSanitizer.html

- Build related changes

- No

None

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,9 +164,9 @@ mark_as_advanced(QMC_EXP_THREADING)
 # Sanitizer options
 #--------------------------------------------------------------------
 
-# Add optional sanitizers ASAN, UBSAN, MSAN
-set(VALID_SANITIZERS "asan" "ubsan" "tsan" "msan")
-set(ENABLE_SANITIZER "" CACHE STRING "asan,ubsan,tsan,msan")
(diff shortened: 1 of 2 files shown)


prckent:
Confirming that this "works" with clang 20.1. For an MPI build there is a ~100% failure rate of the deterministic tests, so there is some digging to do. Presumably something widely used is at fault, e.g.  MPI wrappers or XML handling.

prckent:
Test this please

ye-luo (approved):

ye-luo:
> Confirming that this "works" with clang 20.1. For an MPI build there is a ~100% failure rate of the deterministic tests, so there is some digging to do. Presumably something widely used is at fault, e.g. MPI wrappers or XML handling.

The MPI one sounds very much like passing complex array in Fortran. We can start using this sanitizer without enabling MPI first.

prckent:
> We can start using this sanitizer without enabling MPI first.

Agree. I will setup a "clang v20, no MPI" environment to do this systematically. That is not an environment we have in the nightlies. Don't know what happens without MPI yet.

Issue #5292, opened 2025-01-29
Reduce the state machine in parameter derivative routines needed


ye-luo:
The state machine was invented for WFOpt.
`checkInVariables` queries all the optimizable pieces in TWF and then we create a `globalVars`
`checkOutVariables` updates the per piece `myVars` by storing its index in the `globalVars`.
During `evaluateDerivatives(globalVars)`, `myVars` and its indices are used.

Clearly the global indices stored in all the per piece `myVars` are the state machine. A solution is to keep the action of `checkOutVariables` within `evaluateDerivatives` and then it can be paired with arbitrary amount of `globalVars` which extends this machinery to estimators and even any number of estimators enabled in a single run.

Each feature that needs `evaluateDerivatives` calls simply register its own `FeatureVars` by calling `checkInVariables`.

jtkrogel:
Thanks for thinking this through.  I agree the current implementation is a mess.

How much effort do you think this would be?  Would this still allow parameter derivatives to be called at the WavefunctionComponent level?

PDoakORNL:
My main concern is that we have been trying to keep the data dependence of estimators intelligible, thats less of a concern if these are just a context available only to the self healing estimator that can be supplied once per section or once per block. I don't have a ton of time today but aren't the variables basically owned by the BaseCostFunction? Are we talking about making a reference to these available to the SelfHealingEstimator or a different set of them?  When you say evaluate I assume you mean once per measurement i.e. in the fully concurrent accumulate.

PDoakORNL:
If you need to access them inside the accumulate then the data dependence has changed.  At what scope are the derivatives the accumulation depends on mutable?  Or are we talking about a list of activate variables that the estimator uses to call the evaluateDerivatives.  Apologies for not keeping up with the self healing methodology.

ye-luo:
My intention was to resolve the implicit dependency of `myVars` on `globalVars`. Once that is resolved, evaluateDerivatives can be treated like a const function that can be called as many times we need. Thus places which need derivatives may call freely whereas currently only WFOpt can call derivative routines.

Regarding intelligent data sharing between estimators can be viewed as an optimization that eliminates redundant calculation. We can do that once we have the basic implementation working that allows evaluating derivatives in each estimator.

[...]

PR #5435, opened 2025-04-10
Reducing VariableSet features

ye-luo:
## Proposed changes
I'm investigating https://github.com/QMCPACK/qmcpack/issues/5292 and found a few undocumented feature of VariableSet made my refactoring significantly harder.

This PR removes two **per-parameter** control related features.
1. parameter derivative recompute. After my change, all the recompute is treated as true. Note that, I have not seen a case easily benefiting from this feature.
2. activate/disable parameters in WFOpt drivers. Constraints should be implemented inside each WFC who has the full knowledge. Thus driver level feature is not needed.

- Other (please describe): Feature removal

- No

laptop

## Checklist
- [ ] I have read the pull request guidance and develop docs
- [ ] This PR is up to date with current the current state of 'develop'
- [ ] Code added or changed in the PR has been clang-formatted


prckent:
>activate/disable parameters in WFOpt drivers. Constraints should be implemented inside each WFC who has the full knowledge. Thus driver level feature is not needed.

I currently disagree on this. Keep, or convince me otherwise. This feature is needed to write a smarter optimizer that knows about the different behavior of, say, orbital coefficients vs J1 vs J3 coefficients.

ye-luo:
> > activate/disable parameters in WFOpt drivers. Constraints should be implemented inside each WFC who has the full knowledge. Thus driver level feature is not needed.
> 
> I currently disagree on this. Keep, or convince me otherwise. This feature is needed to write a smarter optimizer that knows about the different behavior of, say, orbital coefficients vs J1 vs J3 coefficients.

The removed feature is about activate/disable individual parameters not individual set of parameters.
For example, spline Jastrow has 10 parameters, there was a mechanism to disable 5th parameter.
This feature is removed.
There is no change to [Variational parameter selection](https://qmcpack.readthedocs.io/en/develop/methods.html#variational-parameter-selection) feature

prckent (approved):
After discussion with Ye, I think that this is good cleanup, simplification, and removal of a design issue.

I was concerned about keeping features available to e.g. toggle parameter optimization and implement statistically aware optimization techniques similar to e.g. Leon Otis' existing code and publications, but these kinds of features do not require this route. Test patch coverage is also >90%.

prckent:
Test this please

PR #5449, opened 2025-04-18
Pin host memory for fast/stable H2D and D2H transfers in non-batched delayed update implementation

ye-luo:
Depends on #5452
## Proposed changes
The non-batched delayed update implementation cannot use dual-space memory containers due to the legacy buffer operations when interacting with legacy drivers. For this reason, to enable pinned host memory on the inverse matrix, we need to register/unregister when resizing. All the memory attaching is a mess that I'd like to get rid of when legacy drivers are removed.

I didn't see noticeable performance improvement but it seems to stabilize the SYCL code path on Aurora when running the 512 atom problem at full machine scale. Intel is working on figuring out the issue when not using pinned memory but we don't need to wait for a fix.

I did the investigation in March but wait until the fuse of CUDA and SYCL. In this way, this feature is applicable to all GPU platforms.

- New feature

- No

Aurora, Epyc-server

## Checklist
* * [x] This PR is up to date with current the current state of 'develop'


ye-luo:
Test this please

ye-luo:
Test this please

prckent (commented):
What does this currently do to performance on Polaris (NV A100), and Frontier (ROCm 6.3+, AMD MI250X)?

src/Containers/OhmmsPETE/OhmmsMatrix.h
```diff
@@ -132,6 +132,9 @@ class Matrix
     X.attachReference(other, ref, n * m);
   }
 
+  /// return true if this container is attached to another container
+  inline bool isAttaching() const { return X.isAttaching(); }
```
prckent, line 136: isAttached?

ye-luo: That is better.

prckent:
Happy to merge if you can confirm that this doesn't do anything bad on NV or AMD.

prckent:
Test this please

prckent (approved):

Issue #5468, opened 2025-04-27
Energy density estimator tests near 100% fail rate


prckent:
**Describe the bug**

Looking at the month long statistics for tests named *energydensity*, the fail rate is near 100%
https://cdash.qmcpack.org/testOverview.php?project=QMCPACK&begin=2025-04-01&end=2025-04-27&filtercount=1&showfilters=1&field1=testname&compare1=63&value1=energydensity

This tests are mostly for legacy where the code has not been intentionally changed in months. Possible explanations include unintentional breakage of the tests, of the underlying code, or perhaps the tests &/or code were simply not robust after the last time they were "fixed". Most of the reported numerics are close to passing.

**To Reproduce**

ctest -R energydensity , any recent version

**Expected behavior**

Test should pass; we should have confidence in energy density.

**System:**

Any

**Additional context**

Did not yet try to bisect failure or check historical PR discussions.

prckent:
Looks like we didn't follow-up and fix the underlying problems after fixing check_stats.py:
```
The following tests FAILED:
	431 - short-LiH_ae-vmc_hf_noj_estimator_energydensity_voronoi-4-4-check (Failed)
	547 - short-LiH_pp-vmc_hf_sdj_estimator_energydensity_voronoi-4-4-check (Failed)
	1240 - short-diamondC_1x1x1_pp-vmc-estimator-energydensity-cell-4-4-check (Failed)
	1242 - short-diamondC_1x1x1_pp-vmc-estimator-energydensity-voronoi-4-4-check (Failed)
	1244 - short-diamondC_1x1x1_pp-dmc-estimator-energydensity-cell-4-4-check (Failed)
Errors while running CTest
313cf7ece465007b72d42164a7f3c7ad7e10e767 is the first bad commit
commit 313cf7ece465007b72d42164a7f3c7ad7e10e767
Author: Ye Luo <yeluo@anl.gov>
Date:   Wed May 15 13:17:27 2024 -0500

    check_stats.py properly exits when h5py read fails.

 tests/scripts/check_stats.py | 5 ++---
 1 file changed, 2 insertions(+), 3 deletions(-)
bisect found first bad commit
```

jtkrogel:
[...]

PR #5469, opened 2025-04-27
Add partial deterministic energy density estimator tests

prckent:
## Proposed changes

As part of investigating #5468 , add some single core deterministic tests. There are no existing deterministic energy density estimator tests. The new tests currently only verify that the underlying VMC run is not changed by the estimators; the existing proven check data for the VMC only runs is used. Legacy passes. Batched changes the result but I am uncertain if this estimator was fully plumbed or if it is reasonable to keep the old behavior; to be revisited once @PDoakORNL is back. 

Once we understand the issue with the existing tests &/or implementation, these new tests should be updated with a check on the estimator results following the existing pattern for the short tests defined in the same file.

- No

Ubuntu 24 LTS, GCC 14, double precision.

## Checklist

* * [X] I have read the pull request guidance and develop docs
* * [X] This PR is up to date with current the current state of 'develop'


prckent:
Test this please

ye-luo (approved):

PR #5490, opened 2025-05-05
Batched Structure Factor Estimator and unit tests

PDoakORNL:
## Proposed changes

Batched structure factor estimator and unit tests. 

- New feature
- Other (please describe):

- No

sdgx2

## Checklist


ye-luo:
FYI. You may add `[Do not merge]` in the title or mark this PR as draft. Then your intention can be visible directly at https://github.com/QMCPACK/qmcpack/pulls

PDoakORNL:
Lets institute the MYOB tag for PRs like this.  This means as opposed to [WIP] that a PR is simply using the testing infrastructure for private development and that the submitter requests anyone looking at it does not use it to make speculative judgements of PR actually submitted for review.

prckent:
I agree it would be useful to have a tag on this topic. [Donotmerge] has already been used for this, but we can should look around to see what other larger projects are doing in case there is a better alternative that is clearer. I addition to considering MYOB, consider [FORCIONLY] [TESTONLY] [JUSTFORTESTING] etc.

prckent:
Test this please

prckent (dismissed):
LGTM

Thanks Peter.

Q. Do we know the origin of the "gold" reference data used in the tests? e.g. sfk_e_e_expected? Browsing the repo I don't spot an obvious source.

src/Estimators/StructureFactorEstimator.h
```diff
@@ -0,0 +1,107 @@
...
+  void startBlock(int steps) override;
+
+  UPtr<OperatorEstBase> spawnCrowdClone() const override;
+
+  void registerOperatorEstimator(hdf_archive& file) override;
+  void write(hdf_archive& file) override;
+  void collect(const RefVector<OperatorEstBase>& type_erased_operator_estimators) override;
+
+  long long getNumKPoints() { return num_kpoints_; }
+  auto getKLists() { return ions_.getSimulationCell().getKLists(); };
```
ye-luo, line 61: `const auto&` and const member function? It seems making a copy now.

src/Estimators/StructureFactorEstimator.cpp
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+}
+
+void StructureFactorEstimator::registerOperatorEstimator(hdf_archive& file)
+{
+  hdf_path hdf_name{my_name_};
+  hdf_path path_variables = hdf_name / std::string_view("kpoints");
+  file.push(path_variables, true);
+  file.write(const_cast<std::vector<KPt>&>(ions_.getSimulationCell().getKLists().getKptsCartWorking()), "value");
```
ye-luo, line 95: Could you add a comment about this `const_cast` to have a piece of mind? Probably a restriction from `write` function?

src/Estimators/tests/test_StructureFactorEstimator.h
```diff
@@ -0,0 +1,659 @@
...
+  using Real = StructureFactorEstimator::Real;
+
+public:
+  const Vector<Real>& getSKElecElec(const StructureFactorEstimator& sfe) const { return sfe.getSKElecElec(); }
+  const Vector<std::complex<Real>>& getRhoKElec(const StructureFactorEstimator& sfe) const { return sfe.getRhoKElec(); }
+};
+} // namespace testing
+
+/// Provides the canned test data for testing
+class StructureFactorTests
```
ye-luo, line 31: Just wondering why this test has a header file. Will it be included in other places?

src/Utilities/for_testing/NativeInitializerPrint.hpp
```diff
@@ -95,7 +95,10 @@ inline std::ostream& operator<<(std::ostream& out, const NativePrint<Vector<T>>&
   out << "{ ";
   auto vec = np_vec.get_obj();
   for (T& t : vec)
-    out << std::setprecision(10) << t << ", ";
+    if constexpr (IsComplex_t<T>::value)
+      out << std::setprecision(10) << '{' << t.real() << ", " << t.imag() << "}, ";
+    else
+      out << std::setprecision(10) << t << ", ";
```
ye-luo, line 101: What is your motivation of changing the style of std::complex printout?
I tend to prefer what the standard library does.
If we change the style in this template specialization, should we all the cases in this file?

PDoakORNL: It needs to be correct for the initializer of a Vector of std::complex.  That is the entire point of this output adapter is to output in a format that can be directly pasted into a test program for expected output.  That doesn't work with the default operator<< for std::complex.

src/Estimators/tests/test_StructureFactorEstimator.h
PDoakORNL: The templating of the canned data requires quite a particular order and the .cpp was becoming too much for some of the tooling I use, including clang format.

src/Estimators/StructureFactorEstimator.cpp
PDoakORNL: restriction from `hdf_archive.` That are const_cast all over the code like this but you normally see them in calls to observable helper.  I'd like to delete observable helper which i think is an unnecessary state machine so I don't use it here.

ye-luo (approved):

PDoakORNL:
@PDoakORNL  currently the test data was just produced by the StructureFactorEstimator itself.

```
constexpr bool generate_test_data = false;
```
setting this to true generates new test data.

looking over the QMCHamiltonian tests I think I can add another to compare to the legacy test_SkAllEstimator. 
I will do that in a further PR.

src/Utilities/for_testing/NativeInitializerPrint.hpp
ye-luo: Basically it is a special need. Q: does the extra ‘, ‘ causes problem or you always remove it by hand.

PDoakORNL:
test this please

ye-luo:
test this please

PR #5491, opened 2025-05-05
Minimal pools target

PDoakORNL:
## Proposed changes

This adds a cmake target to ease dependency reasoning and inclusion of minimal testing pools for walker elements in tests. I found it eased test development for estimators.

- Build related changes

- No

sdgx

## Checklist

* * [ x] This PR is up to date with the current state of 'develop'
* * [ x] Code added or changed in the PR has been clang-formatted
* * [ x] This PR adds tests to cover any new code, or to catch a bug that is being fixed
* * [ x] Documentation has been added (if appropriate)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -126,6 +126,10 @@ if(QMC_BUILD_SANDBOX_ONLY)
 
 else() #{{{
 
+  add_library(minimal_testing_pools INTERFACE)
+  target_sources(minimal_testing_pools INTERFACE Particle/tests/MinimalParticlePool.h QMCHamiltonians/tests/MinimalHamiltonianPool.h QMCWaveFunctions/tests/MinimalWaveFunctionPool.h)
+  target_include_directories(minimal_testing_pools INTERFACE Particle/tests QMCHamiltonians/tests QMCWaveFunctions/tests)
+
   add_subdirectory(QMCWaveFunctions)
   add_subdirectory(QMCHamiltonians)
   add_subdirectory(Estimators)
--- a/src/Estimators/tests/CMakeLists.txt
+++ b/src/Estimators/tests/CMakeLists.txt
@@ -2,7 +2,7 @@
 #// This file is distributed under the University of Illinois/NCSA Open Source License.
 #// See LICENSE file in top directory for details.
 #//
-#// Copyright (c) 2024 QMCPACK developers.
+#// Copyright (c) 2025 QMCPACK developers.
 #//
 #// File developed by: Mark Dewing, mdewing@anl.gov, Argonne National Laboratory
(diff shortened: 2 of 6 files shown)


src/CMakeLists.txt
```diff
@@ -126,6 +126,10 @@ if(QMC_BUILD_SANDBOX_ONLY)
 
 else() #{{{
 
+  add_library(minimal_testing_pools INTERFACE)
+  target_sources(minimal_testing_pools INTERFACE Particle/tests/MinimalParticlePool.h QMCHamiltonians/tests/MinimalHamiltonianPool.h QMCWaveFunctions/tests/MinimalWaveFunctionPool.h)
+  target_include_directories(minimal_testing_pools INTERFACE Particle/tests QMCHamiltonians/tests QMCWaveFunctions/tests)
```
ye-luo, line 131: 1. If I understand correctly `minimal_testing_pools` should only be used by targets that use `MinimalParticlePool`, `MinimalWaveFunctionPool` and `MinimalWaveFunctionPool` all three.
Otherwise, it may cause unnecessary and potentially harmful depdendency.
Could you add a comment mentioning that?
2. Since you exposed the `include_directories` in line 131, is explicitly mentioning header files in line 130 even necessary?

ye-luo (commented):
The rest LGTM.

src/CMakeLists.txt
PDoakORNL: With respect to linking the target its not going to actually do anything but add those include directories at this time.  "Linking" a target obviously creates a dependency on the target and its dependencies.  I prefer not to have this create a link dependency to parent particleset, hamiltonian, wavefunction libraries so the consumers of it can continue to link only the needed walkere element libraries.

ye-luo: > It's suggested to explicitly put the headers of a headers only target into the sources. This on its own won't set the include dirs properly for a case like this where the files are not in the directory the target is defined in.

Could you add a comment saying this target is intended for header file only.

> With respect to linking the target its not going to actually do anything but add those include directories at this time. "Linking" a target obviously creates a dependency on the target and its dependencies. I prefer not to have this create a link dependency to parent particleset, hamiltonian, wavefunction libraries so the consumers of it can continue to link only the needed walkere element libraries.

You are right. Actually this is not a complete target and as you pointed out each dependent target still needs explicitly adding target dependency. Could you also mentions this requirement in the comment?

So it will be helpful to have some brief comments mentioning the above restrictions and requirement. Once this is merged, other developers can potentially use it and they need to know such useful info.

ye-luo (approved):
@PDoakORNL if you prefer adding the requested comments in this PR, feel free to do so and I will make another approval. Or you can include it in your next PR in the PR queue.

ye-luo:
Test this please

PR #5495, opened 2025-05-06
fix missing consts for ParticleSet randomize for source

PDoakORNL:
## Proposed changes
randomizeFromSource does not and should not get a nonconst reference to the source particle set.
It has always been "incorrect" to call the addAttribute in this way but it pretty clearly was the standard way of getting an index for an attribute. In 2010 Ken Essler, was either a. unware or b. aware and unwilling to challenge the SpeciesSet API. 

I need to do this to do other work at a reasonable level of quality and the API  _should have already been this way_.
I believe that there is still a potential issue here if a ParticleSet without charge and membersize attributes already set is passed in but at least now it doesn't mutate an input argument that the caller wouldn't expect to be changed.

- Bugfix
- New feature

- No

## Checklist

* * [x ] This PR is up to date with the current state of 'develop'
* * [x ] Code added or changed in the PR has been clang-formatted
* * [x ] This PR adds tests to cover any new code, or to catch a bug that is being fixed
* * [x ] Documentation has been added (if appropriate)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -126,6 +126,24 @@ if(QMC_BUILD_SANDBOX_ONLY)
 
 else() #{{{
 
+  # a target for the minimal testing pools.
+  # This should be used by test programs using any of the minimal
+  # pools
+  #
+  # What it does:
+  # It removes the need to include a relative
+  # path into another modules testing directory to use a minimal pools
+  #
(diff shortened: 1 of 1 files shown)


ye-luo:
Test this please

ye-luo (approved):

PR #5506, opened 2025-05-12
Correct h5 handle in PySCF converter for multi determinant calculations

kayahans:
## Proposed changes

CI data is written to [title].h5 file rather than the [title]_multidet.h5 file in the current implementation. In the current implementation [title]_multidet.h5 file is empty, strongly indicating that there is a simple bug. Without any data in [title]_multidet.h5, later on convert4qmc complains giving unspecific error, but after this simple fix everything seems to work. 

- Bugfix

- No

Unrestricted CASCI calculations on N atom with a small active space. 

## Checklist

* * [X] This PR is up to date with the current state of 'develop'


prckent:
Hi Kayahan. Thanks. Which tests and/or runs are failing for you with PySCF? Looking at cdash I only see failing AFQMC converter issues. Presumably we have a gap in testing that needs fixing to avoid a repeat, or a least an oversight in the tests we currently have.

prckent:
Emailing for more info.

PR #5510, opened 2025-05-13
Split BsplineAllocator into Single/MulitBsplineAllocator

ye-luo:
## Proposed changes
BsplineAllocator was servicing single and multi b-spline allocations. This PR decouple them. It seems that the single b-spline allocator is only used in unit tests. Eventually it should be the one in use in the builders.

- No

laptop

## Checklist


ye-luo:
Test this please

prckent (approved):
Fair enough. A bit more consistent.

prckent:
Editorial: This is not the place to make final plans, but I would like to see the spline codes moved from src/splineXYZ to a place under Numerics, or similar. einspline needs moving as well. They are not top level features like the wavefunctions or drivers. And as regards testing, we should get the splines near 100% - this is easily done and I worry about typos in little used paths (say, antiperiodic in Y), or when backflow gets reactivated. If we are to keep the current set of einspline files we can easily add the original einspline tests, but spline2 needs a bit more coverage imo in addition to a better and more meaningful name.

ye-luo:
The current spline related codes spread in `src/einspline`, `src/spline` and `src/spline2`.
@ylee88  and I are currently working on unwinding the spaghetti code in the builders and SPO classes. Coverage should go up as we make progress.

PR #5520, opened 2025-05-16
fixes for more complex estimators

PDoakORNL:
## Proposed changes

`OperatorEstBase::zero` becomes a virtual method, more complex estimators are better off not packing all their data into the share `data_` buffer so overriding to customize zeroing is necessary.

`OperatorEstBase::collect`
no longer has the side effect of zeroing the `OperatorEstBase` objects passed to it.

refactor split of collect `OperatorEstBase::makeBlockAverages` otherwise unchanged.

typo in test program.

- New feature

- No

sdgx2

## Checklist

### DIFF
--- a/src/Estimators/OperatorEstBase.h
+++ b/src/Estimators/OperatorEstBase.h
@@ -50,7 +50,7 @@ public:
   /** Shallow copy constructor!
    *  This alows us to keep the default copy constructors for derived classes which
    *  is quite useful to the spawnCrowdClone design.
-   *  Data is likely to be quite large and since the OperatorEstBase design is that the children 
+   *  Data is likely to be quite large and since the OperatorEstBase design is that the children
    *  reduce to the parent it is infact undesirable for them to copy the data the parent has.
    *  Initialization of Data (i.e. call to resize) if any is the responsibility of the derived class.
    */
@@ -59,7 +59,7 @@ public:
   virtual ~OperatorEstBase() = default;
 
   /** Accumulate whatever it is you are accumulating with respect to walkers
-   * 
+   *
    *  This method is assumed to be called from the crowd context
    *  It provides parallelism with respect to computational effort of the estimator
    *  without causing a global sync.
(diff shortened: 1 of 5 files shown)


src/Estimators/OperatorEstBase.h
```diff
@@ -81,11 +81,15 @@ class OperatorEstBase
...
-   *  that the crowd operator estimators accumulation data is not being written to.
+   *  that the crowd operator estimators accumulation data is not
+   *  being written to.
+   *  The input operators are not zeroed after collect is called,
+   *  the owner of the operators must handle the accumulated state explicitly.
    *
    *  There could be concurrent operations inside the scope of the collect call.
    */
   virtual void collect(const RefVector<OperatorEstBase>& oebs);
+  virtual void zero(RefVector<OperatorEstBase>& oebs);
```
ye-luo, line 92: Could you move this next to the other `zero()` function and document it?

src/Estimators/tests/test_manager_mpi.cpp
```diff
@@ -23,7 +23,7 @@ namespace testing
 {
 bool EstimatorManagerNewTest::testMakeBlockAverages()
 {
-  if (em.my_comm_->rank() == 1)
+  if (em.my_comm_->rank() == 0)
```
ye-luo, line 26: Could you explain this change? The change doesn't seem to be complex related but more like a logic error?

ye-luo (commented):
LGTM. requested some comments and explanation.

src/Estimators/tests/test_manager_mpi.cpp
PDoakORNL: The test should work for any number of mpi ranks but due to this typo it works for any legal number of ranks except 1.

ye-luo: Aha. Got it.

PDoakORNL: Also in this case in the commit comment by complex I meant complicated estimators.

ye-luo: I see. Maybe adjust title?

PDoakORNL:
added documentation

src/Estimators/OperatorEstBase.h
```diff
@@ -139,7 +148,18 @@ class OperatorEstBase
    */
   virtual void write(hdf_archive& file);
 
+  /** Calls zero on every OperatorEstBase in refvector
+   *
+   *  like collect this is intended to be called with a refvector
+   *  where the OperatorEstBase derived types are all the same.
+   *  Derived types overriding this can assume this.
+   */
+  virtual void zero(RefVector<OperatorEstBase>& oebs);
```
ye-luo, line 157: Since it is supposed to only modify oebs, should this one marked const?

ye-luo (approved):

ye-luo:
Test this please

PR #5526, opened 2025-05-19
Add append support to Hdf archive

PDoakORNL:
## Proposed changes

hdf5 archive lacks support for appending to hdf5 datasets which is present in the hdf_wrappers.  This PR extends this functionality to the hdf_archive level allowing new modules to append to hdf5 datasets without resorting to using Observable helper.  New estimators should use this functionality instread of observable helper. Revives some code from #4540.

- New feature

- No

sdgx2

## Checklist

### DIFF
--- a/src/io/hdf/hdf_archive.cpp
+++ b/src/io/hdf/hdf_archive.cpp
@@ -12,6 +12,8 @@
 
 
 #include "hdf_archive.h"
+#include <H5Ipublic.h>
+#include <stdexcept>
 #ifdef HAVE_MPI
 #include "mpi3/communicator.hpp"
 #endif
@@ -47,6 +49,16 @@ void hdf_archive::close()
   if (file_id != is_closed)
     H5Fclose(file_id);
   file_id = is_closed;
+  if (file_apl_ != H5I_INVALID_HID)
+  {
+    H5Pclose(file_apl_);
+    file_apl_ = H5I_INVALID_HID;
+  }
+  if (file_cpl_ != H5I_INVALID_HID)
+  {
+    H5Pclose(file_cpl_);
+    file_cpl_ = H5I_INVALID_HID;
+  }
 }
 
 void hdf_archive::set_access_plist()
@@ -140,7 +152,7 @@ void hdf_archive::set_access_plist(boost::mpi3::communicator& comm, bool request
 #endif
 
 
-bool hdf_archive::create(const std::filesystem::path& fname, unsigned flags)
+bool hdf_archive::create(const std::filesystem::path& fname, unsigned mode_flags)
 {
   possible_filename_ = fname;
   if (Mode[NOIO])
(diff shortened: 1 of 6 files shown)


src/io/hdf/tests/test_hdf_archive.cpp
```diff
@@ -14,6 +14,7 @@
 #include "catch.hpp"
 
 #include "hdf/hdf_archive.h"
+#include <H5Fpublic.h>
```
ye-luo, line 17: We tried to prevent hdf5 library details from the users of hdf_archive. Is this include really needed?

src/io/hdf/hdf_archive.cpp
```diff
@@ -140,25 +152,37 @@ void hdf_archive::set_access_plist(boost::mpi3::communicator& comm, bool request
...
   possible_filename_ = fname;
   if (Mode[NOIO])
     return true;
   close();
-  file_id = H5Fopen(fname.c_str(), flags, access_id);
+  file_apl_ = H5Pcreate(H5P_FILE_ACCESS);
+  if (file_apl_ == H5I_INVALID_HID)
+    throw std::runtime_error("hdf_archive failed to create file access properties list!");
+  file_id = H5Fopen(fname.c_str(), mode_flags, access_id);
   return file_id != is_closed;
```
ye-luo, line 186: I just learned about `H5I_INVALID_HID`. It seems that our `is_closed` and `H5I_INVALID_HID` are both `-1`. Probably it is time to get ride of `is_closed`? This can be done in separate PRs.

ye-luo (commented):
I'm concerned about how the added h5 properties variables are created and closed. The actual append seems solid. Thanks for the implementation.

src/io/hdf/hdf_pete.h
```diff
@@ -17,6 +17,7 @@
 #include "OhmmsPETE/OhmmsVector.h"
 #include "OhmmsPETE/OhmmsMatrix.h"
 #include "OhmmsPETE/OhmmsArray.h"
+#include "type_traits/container_traits_ohmms.h"
```
ye-luo, line 20: Is this include really needed?

src/io/hdf/hdf_archive.cpp
PDoakORNL: I wanted to minimize differences by using `is_closed` at times when that made (in standalone code fragments) no logical sense. It could be fixed in a future PR, but there is a great deal in io/hdf that needs cleanup.

src/io/hdf/hdf_pete.h
PDoakORNL: I'll move the include to the test for now, but its needed to make the hdf_proxy for PETE objects fully functional.

I think it should be in `hdf_pete.h`, it puts hdf_pete on equal terms with the inclusion pattern for hdf_hyperslab.  Without this there is a runtime failure of h5data_proxy<Vector<T>::read because it resolves to the resize method of the default container_traits::resize.

There's some docs of all this in src/io/hdf/README.md which seems like it would do much more good in the doxygen blocks of the various files.

It seems like maybe this was done to keep ohmms out of AFQMC but it sneaks in through hdf_archive's inclusion of hdf_pete just without the container_traits that do come over for stl containers through inclusion of hdf_hyperslab.

It seems to me if you are going to include hdf_pete you might as well include `container_traits_ohmms.h` because it's broken without it. And that if you don't want Ohmms contamination you can't include hdf_archive.h as it stands.

src/io/hdf/tests/test_hdf_archive.cpp
PDoakORNL: Seemed removable, I'm not sure why I didn't remove it with whatever was getting used from it in the test program.

src/io/hdf/hdf_archive.cpp
```diff
@@ -140,25 +152,37 @@ void hdf_archive::set_access_plist(boost::mpi3::communicator& comm, bool request
...
+  file_cpl_ = H5Pcreate(H5P_FILE_CREATE);
+  if (file_cpl_ == H5I_INVALID_HID)
+    throw std::runtime_error("hdf_archive failed to create file create properties!");
+  file_apl_ = H5Pcreate(H5P_FILE_ACCESS);
+  if (file_apl_ == H5I_INVALID_HID)
+  {
+    H5Pclose(file_cpl_);
+    throw std::runtime_error("hdr_archive failed to create file access properties!");
+  }
+  file_id = H5Fcreate(fname.c_str(), mode_flags, file_cpl_, file_apl_);
```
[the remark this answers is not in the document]
PDoakORNL: I've reduced us to one now.

src/io/hdf/hdf_archive.cpp
ye-luo: For follow up https://github.com/QMCPACK/qmcpack/issues/5528

src/io/hdf/hdf_pete.h
ye-luo: Thank you for investigating this issue. Now I'm in favor of moving
```
type_traits/container_traits.h
type_traits/container_traits_mulit.h
type_traits/container_traits_ohmms.h
```
under `io/hdf`. Initially it was intended serve multiple places but now I see one solution for all needs isn't very appealing.  It is better each library that needs to interact with OhmmsPETE classes at low levels implements its own container_traits.

Also `hdf_pete.h` should include `container_traits_ohmms.h`, `hdf_multi.h` should include `container_traits_mulit.h` and `hdf_dataproxy.h` should include `container_traits.h`.

I will make a PR.

ye-luo (approved):
Very nice cleanup and feature implementation!

ye-luo:
Test this please

PR #5535, opened 2025-05-27
[WIP] consistent init and use of input name and type

PDoakORNL:
What I believe the intention of the stat.h5 output is that if a name is supplied for an estimator than that will be its group name in the h5 file.  If no name is supplied the estimator type will be copied to the name and that will be used.

## Proposed changes
The proposed changes put this into effect with consistent origination those values from the estimator input classes and there storage in operator_est_base.  This makes them available when needed in either the base class write or the derived classes specialized write member functions.

- Bugfix

- No, or at least I hope not.  There could be breakage in python based testing or nexus if this "bug" is relied on, i.e. The "my_name" in estimators being assumed to be the estimator class name.

sdgx2

## Checklist


src/Estimators/MomentumDistribution.cpp
```diff
@@ -22,18 +22,16 @@ namespace qmcplusplus
...
                                            const PosType& twist_in,
-                                           const Lattice& lattice_in,
+                                           Lattice lattice_in,
                                            DataLocality dl)
-    : OperatorEstBase(dl),
+    : OperatorEstBase(dl, mdi.get_name(), mdi.get_type()),
       input_(std::move(mdi)),
       twist(twist_in),
-      lattice(lattice_in),
+      lattice(std::move(lattice_in)),
```
ye-luo, line 30: The old code only pays the price of a copy constructor.
The new code pays the price of both copy and move constructor. In addition, do we really have the move constructor of `Lattice` implemented and tested?
I feel the old code was better.

PDoakORNL: See
https://xania.org/202101/cpp-by-value-args

Move's should be basically free compared to anything that causes an Allocation. But with our own objects there is a bunch of auditing to be done to be clear on that. And anytime something is an rvalue this can save an extra copy you'd have with a const &.

Lattice has the implicitly defined default move constructor.  Which means this is likely equivalent to the const  better in most cases, but really I just wanted to shut up clang tidy.  I think the thing to do for now is remove this clang-tidy check.

ye-luo:
> These are both attributes in the xml of the estimators and this stupid put the estimator "type" in either type or name carries over from the existing input behavior.
> 
> > It seem to me that we need to have the name default to the estimator type if a name is not explicitly given.
> 
> **Is already done.**

yep. I was trying to following your thoughts behind the code change.

src/Estimators/MomentumDistribution.cpp
ye-luo: I see where you are from. That blog explains the case when the caller pass a raw string (const char *) which is not our case. Most of the time our callers pass `std::string` from `get_type()` and thus "const std::string&" wins.
I still saw two occurrence of `Lattice lattice` that needs to be stored.

ye-luo (approved):

ye-luo:
Test this please

PR #5540, opened 2025-06-04
Structure Factor Estimator Integration

PDoakORNL:
## Proposed changes

This integrates the batched StructureFactorEstimator with the EstimatorManagerNew and the application.

There still needs to be proper documentation in the manual and a proper statistical test.  I'd like to do that in a following PR.

- New feature

- No

sdgx2
## Checklist

### DIFF
--- a/CMake/AddRPATH.cmake
+++ b/CMake/AddRPATH.cmake
@@ -1,6 +1,7 @@
 # Add the directory of the first library file to RPATH/RUNPATH if that directory is not part of OS default locations.
-function(AddRPATH LIBRARY_NAME LIBRARY_FILE)
-  if(LIBRARY_FILE)
+function(AddRPATH LIBRARY_NAME)
+  if(ARGC GREATER 1)
+    list(GET ARGV 1 LIBRARY_FILE)
     cmake_path(GET LIBRARY_FILE PARENT_PATH LIBRARY_DIR)
     list(FIND CMAKE_PLATFORM_IMPLICIT_LINK_DIRECTORIES ${LIBRARY_DIR} isSystemDir)
     if("${isSystemDir}" STREQUAL "-1")
--- a/src/AFQMC/Memory/CUDA/cuda_arch.cpp
+++ b/src/AFQMC/Memory/CUDA/cuda_arch.cpp
@@ -112,6 +112,7 @@ void free(void* p, const std::string& message)
       std::cerr << " Error from: " << message << std::endl;
     }
     std::cerr << " Error from calling cudaFree: " << cudaGetErrorString(status) << std::endl;
+    throw std::runtime_error("Error: cudaFree returned error code.");
   }
 }
 
--- a/src/AFQMC/Memory/HIP/hip_arch.cpp
+++ b/src/AFQMC/Memory/HIP/hip_arch.cpp
@@ -118,6 +118,7 @@ void free(void* p, const std::string& message)
       std::cerr << " Error from : " << message << std::endl;
     }
     std::cerr << " Error when calling hipFree: " << hipGetErrorString(status) << std::endl;
+    throw std::runtime_error("Error: hipFree returned error code.");
   }
 }
 
--- a/src/AFQMC/Numerics/tests/test_dense_numerics.cpp
+++ b/src/AFQMC/Numerics/tests/test_dense_numerics.cpp
@@ -620,7 +620,7 @@ void test_dense_gerf_gqr_strided_device(Allocator& alloc)
     array<int, 1, IAllocator> info(iextensions<1u>{2}, IAllocator{alloc});
     array<T, 2, Allocator> TAU({2, 4}, alloc);
 
-    geqrfStrided(4, 3, A.origin(), 4, 12, TAU.origin(), sz, info.origin(), 2);
+    geqrfStrided(4, 3, A.origin(), 4, 12, TAU.origin(), 4, info.origin(), 2);
     gqrStrided(4, 3, 3, A.origin(), 4, 12, TAU.origin(), 4, WORK.origin(), sz, info.origin(), 2);
     for (int i = 0; i < 2; i++)
     {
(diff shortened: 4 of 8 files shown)


prckent (approved):
LGTM

src/Estimators/StructureFactorEstimator.cpp
```diff
@@ -84,7 +91,20 @@ void StructureFactorEstimator::accumulate(const RefVector<MCPWalker>& walkers,
       sfk_e_e_[k] += weight * (rhok_tot_r_[k] * rhok_tot_r_[k] + rhok_tot_i_[k] * rhok_tot_i_[k]);
       rhok_e_[k] += weight * std::complex<Real>{rhok_tot_r_[k], rhok_tot_i_[k]};
     }
+
+    walkers_weight_ += weight;
+  }
+}
+
+const ParticleSet& StructureFactorEstimator::getParticleSet(const PSPool& psetpool, const std::string& psname) const
```
ye-luo, line 99: There is `ParticleSetPool::getParticleSet` although the call pattern is different currently.
Once we refine that https://github.com/QMCPACK/qmcpack/issues/5543.
This function can be removed. So some future work needed.

src/Estimators/StructureFactorEstimator.cpp
```diff
@@ -103,28 +123,54 @@ void StructureFactorEstimator::write(hdf_archive& file)
...
 void StructureFactorEstimator::collect(const RefVector<OperatorEstBase>& type_erased_operator_estimators)
 {
   int num_crowds = type_erased_operator_estimators.size();
   for (OperatorEstBase& crowd_oeb : type_erased_operator_estimators)
   {
-    StructureFactorEstimator& crowd_sfe = dynamic_cast<StructureFactorEstimator&>(crowd_oeb);
+    auto& crowd_sfe = dynamic_cast<StructureFactorEstimator&>(crowd_oeb);
     this->sfk_e_e_ += crowd_sfe.sfk_e_e_;
     this->rhok_e_ += crowd_sfe.rhok_e_;
+    walkers_weight_ += crowd_sfe.walkers_weight_;
```
ye-luo, line 139: I'm confused by the coverage report. Is this collect function covered by test? IIRC, `collect()` is the most important entry to an estimator.

src/Estimators/StructureFactorEstimator.cpp
```diff
@@ -17,6 +17,15 @@
...
 {
 
+StructureFactorEstimator::StructureFactorEstimator(const StructureFactorInput& sfi,
+                                                   const PSPool& pset_pool,
+                                                   DataLocality data_locality)
+    : StructureFactorEstimator(sfi,
+                               getParticleSet(pset_pool, sfi.get_source()),
+                               getParticleSet(pset_pool, sfi.get_target()),
+                               data_locality)
+{}
```
ye-luo, line 27: Likely we can get remove this constructor and just keep the other one once we clean up `getParticleSet`.

ye-luo (changes requested):
The only required change in this PR is replacing `std::cout` with `app_log()` in all the tests touched by this PR. All the rest of my comment was for future improvement and doesn't require immediate action.

src/Estimators/StructureFactorEstimator.cpp
ye-luo: Probably this also needs to be taken care in this PR. Please clarify.

src/Estimators/StructureFactorEstimator.cpp
PDoakORNL: Right now I don't want the Estimators directly handling particle set pools. I don't think read consumers of pools need access to anything but a name object map.

src/Estimators/StructureFactorEstimator.cpp
PDoakORNL: I actually think this constructor is more fundamental that the public one. 

The names of the particleset source and target are in the in the StructureFactorInput which no other object should really be using or acting on. Right now I feel like the best place to put these global pool to simulation object  operations is in this sort of minimal constructor.

But two things about this. 
* The input isn't actually valid if there isn't a matching source and target particleset so ideally before any StructureFactorEstimator exists the "integrated" validity of the `ParticlePoolInput` + `StructureFactorInput` should be checked. We know before anything in the qmc section is constructed that the input is garbage.

* Design wise I think its desirable to construct simulation objects by handing them their default constructed input and the lowest level simulation objects they require as opposed to vague global objects (like the pools). This declares explicitly the object's construction time data dependence in the constructors signature. Much mucking around in its actual procedural guts to understand this is saved and that code is pulled out in a more organized manner in the delagating constructor.I think there is great value in every simulation object having a "most" delegated constructor like this.

* Additionally unit testing is far more convenient being able to directly construct and pass these sort of objects which may be partially mocked in many cases. 

And hopefully it would reduce developers alternately: 
* leaning on all the heavy input and state machine transformations that the entire application global section and object initialization goes through
or 
* Claiming its too hard to write unit tests because of the same.

ye-luo: > I actually think this constructor is more fundamental that the public one.
This is the public one currently. Could you clarify?

I feel the current public one is less fundamental than the private one that was previously public and now private.

ye-luo: I feel we can handle the input in the following way to avoid too much entanglement.
1. input classes only needs to store names of needed particlesets/wavefunctions/hamiltonians that it operates on. We don't want to contaminate them with pools.
2. It is also undesirable to have pools exposed to estimator classes therefore their constructors.
3. Usually between an object and its input is a builder. In this case EstimatorManagerNew is doing the work making estimators based on input. For this reason, EstimatorManagerNew is the best place to bring in pools. I don't see it bad if EstimatorManagerNew needs to read some info from the input object and it actually always read the type info.
4. In addition, error out from EstimatorManagerNew is always better than an estimator constructor from error handling perspective.

Right now global input estimators has delayed action, extra code will be needed for early validation. It is desirable but not necessarily high priority.

src/Estimators/StructureFactorEstimator.cpp
ye-luo: @PDoakORNL any comment?

ye-luo (approved):

ye-luo:
Test this please

src/Estimators/StructureFactorEstimator.cpp
PDoakORNL: @ye-luo  I will take a look at the coverage

PR #5542, opened 2025-06-06
Bump up CUDA version in CI and fix test_afqmc_numerics

ye-luo:
## Proposed changes
Due to recent change on sulfur, CUDA version on the offload CI case needs to be updated.
Fix a wrong stride used in test_afqmc_numerics.

- Bugfix

- No

epyc-server

## Checklist

### DIFF
--- a/src/AFQMC/Memory/CUDA/cuda_arch.cpp
+++ b/src/AFQMC/Memory/CUDA/cuda_arch.cpp
@@ -112,6 +112,7 @@ void free(void* p, const std::string& message)
       std::cerr << " Error from: " << message << std::endl;
     }
     std::cerr << " Error from calling cudaFree: " << cudaGetErrorString(status) << std::endl;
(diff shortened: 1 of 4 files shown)


ye-luo:
Test this please

prckent (approved):

ye-luo:
Here is what I found so far regarding the remaining failure with CUDA 12.9.
The breakage starts with CUDA 12.8. Swapping cublas and cublasLt to 12.6 workaround the issue.
`LD_PRELOAD="/usr/local/cuda-12.6/lib64/libcublas.so /usr/local/cuda-12.6/lib64/libcublasLt.so"`
The segfault came out of inside gemv/gemm calls to cublas. These are not segfault inside kernels but the launch call on the lost. I cannot rule out the possibility of a bug inside cublas.

prckent:
Switched to a pinned version CUDA 12.6

PR #5546, opened 2025-06-12
Fixing JeeI crashing in DMC GPU runs

ye-luo:
## Proposed changes
CPU/GPU doesn't agree exactly on numerics and extra caution is needed. After branching, distance tables recomputed on the GPU introduce inconsistency with the JeeI compact maintained using CPU calculated distances during moves.
An electron-ion pair distance may considered inside the cutoff by GPU but outside by GPU.
This can be fixed by remaking JeeI compact lists after recomputing for resident walkers. Migrant walkers always recompute fully.

- Bugfix

- No

Aurora

## Checklist


ye-luo:
Test this please

prckent (approved):

anbenali:
I am always happy to break things :) 

Big up to Ye for finding this!!
And for reference this bug was triggered on Aurora but not Nvidia... also... it only happened after the April 15th update!! :)

PR #5547, opened 2025-06-12
Add spinor detratios

camelto2:

Please also follow the [GitHub Pull Request guidance](https://qmcpack.readthedocs.io/en/develop/developing.html#github-pull-request-guidance). As much as possible, try to avoid the ["Don't"s](https://qmcpack.readthedocs.io/en/develop/developing.html#don-t) to help maintain a high development velocity.

## Proposed changes

This PR implements some missing features in SpinorSet, namely evaluateDetRatios and evaluateDetSpinorRatios which and the appropriate mw_ implementations which allows for offloaded determinant ratios evaluations  in NonLocalECPComponent and SOECPComponents. 

After doing some test runs of SOC on a new machine, I noticed I was basically getting constant throughput for SOC as I increased the walkers_per_rank. After looking at the timers, it was clear that basically everything in the hamiltonian evaluation was falling back to CPU implementation, so we were getting basically no performance. Part of the problem was that we hardcoded the SpinorSetBuilder with useGPU="no" as a guard and never removed it. The rest was that we should be using mw_ evaluateDetRatios and mw_evaluateDetSpinorRatios, which didn't exist in SpinorSet. This is all now fixed, along with unit tests. 

Below is a figure showing relative throughput on the GPUs compared to the nonSOC case. Just from the change in determinant size alone in going from an up and down determinant of size (N/2) to a single larger determinant of size N, you expect a factor of 4 slowdown from the standard N^3 scaling. Around 300 walkers/rank is around where the throughput peaks, so we are giving enough work to the GPU and not being dominated by transfers, and with the fix we see about a 2.5 slowdown relative to the nonSOC, which is in the right ballpark. 

image

- New feature

- No

AMD MI-300A APUs

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/SpinorSet.cpp
+++ b/src/QMCWaveFunctions/SpinorSet.cpp
@@ -86,6 +86,35 @@ void SpinorSet::evaluateValue(const ParticleSet& P, int iat, ValueVector& psi)
   psi = eis * psi_work_up + emis * psi_work_down;
 }
 
+void SpinorSet::evaluateDetRatios(const VirtualParticleSet& VP,
+                                        ValueVector& psi,
+                                        const ValueVector& psiinv,
+                                        std::vector<ValueType>& ratios)
+{
+  assert(psi.size() == psiinv.size());
+  const size_t norb_requested = psi.size();
+  psi_work_up.resize(norb_requested);
+  psi_work_down.resize(norb_requested);
+
+  for (size_t iat = 0.0; iat < VP.getTotalNum(); iat++)
+  {
+    psi_work_up   = 0.0;
+    psi_work_down = 0.0;
+    spo_up->evaluateValue(VP, iat, psi_work_up);
+    spo_dn->evaluateValue(VP, iat, psi_work_down);
+
+    ParticleSet::Scalar_t s = VP.activeSpin(iat);
+    RealType coss           = std::cos(s);
+    RealType sins           = std::sin(s);
+
+    ValueType eis(coss, sins);
+    ValueType emis(coss, -sins);
+
+    psi = eis * psi_work_up + emis * psi_work_down;
+    ratios[iat] = simd::dot(psi.data(), psiinv.data(), psi.size());
+  }
+}
+
 void SpinorSet::evaluateDetSpinorRatios(const VirtualParticleSet& VP,
                                         ValueVector& psi,
                                         const std::pair<ValueVector, ValueVector>& spinor_multiplier,
--- a/src/QMCWaveFunctions/SpinorSet.cpp
+++ b/src/QMCWaveFunctions/SpinorSet.cpp
@@ -121,7 +121,7 @@ void SpinorSet::mw_evaluateDetRatios(const RefVectorWithLeader<SPOSet>& spo_list
 {
   auto& spo_leader = spo_list.getCastedLeader<SpinorSet>();
   auto& vp_leader  = vp_list.getLeader();
-  IndexType nat = vp_leader.getTotalNum();
+  IndexType nat    = vp_leader.getTotalNum();
   assert(this == &spo_leader);
 
   IndexType nw                    = spo_list.size();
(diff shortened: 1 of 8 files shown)


camelto2:
I expect the current failure is due to some improper initialization of offload stuff in the test. Need to dig into it a bit to see what is going on

camelto2:
BTW, I have another PR ready to go which implements the applyRotation on SplineC2COMPTarget. Will put that up once this gets merged

ye-luo:
Test this please

src/QMCWaveFunctions/SpinorSet.h
```diff
@@ -14,6 +14,7 @@
 #define QMCPLUSPLUS_SPINORSET_H
 
 #include "QMCWaveFunctions/SPOSet.h"
+#include "VirtualParticleSet.h"
```
ye-luo: Please move this include to the cpp file. Also the test file will need this include.

src/QMCWaveFunctions/SpinorSet.cpp
```diff
@@ -86,6 +86,91 @@ void SpinorSet::evaluateValue(const ParticleSet& P, int iat, ValueVector& psi)
...
+
+void SpinorSet::mw_evaluateDetRatios(const RefVectorWithLeader<SPOSet>& spo_list,
+                                     const RefVectorWithLeader<const VirtualParticleSet>& vp_list,
+                                     const RefVector<ValueVector>& psi_list,
+                                     const std::vector<const ValueType*>& invRow_ptr_list,
+                                     std::vector<std::vector<ValueType>>& ratios_list) const
+{
+  auto& spo_leader = spo_list.getCastedLeader<SpinorSet>();
+  auto& vp_leader  = vp_list.getLeader();
+  IndexType nat    = vp_leader.getTotalNum();
```
ye-luo: Prefer `const size_t nat`

src/QMCWaveFunctions/SpinorSet.cpp
```diff
@@ -86,6 +86,91 @@ void SpinorSet::evaluateValue(const ParticleSet& P, int iat, ValueVector& psi)
...
+  assert(this == &spo_leader);
+
+  IndexType nw                    = spo_list.size();
```
ye-luo: Prefer `const size_t nw`

src/QMCWaveFunctions/SpinorSet.cpp
```diff
@@ -97,22 +182,80 @@ void SpinorSet::evaluateDetSpinorRatios(const VirtualParticleSet& VP,
...
+void SpinorSet::mw_evaluateDetSpinorRatios(const RefVectorWithLeader<SPOSet>& spo_list,
+                                           const RefVectorWithLeader<const VirtualParticleSet>& vp_list,
+                                           const RefVector<ValueVector>& psi_list,
+                                           const RefVector<std::pair<ValueVector, ValueVector>>& spinor_multiplier_list,
+                                           const std::vector<const ValueType*>& invRow_ptr_list,
+                                           std::vector<std::vector<ValueType>>& ratios_list) const
+{
+  auto& spo_leader = spo_list.getCastedLeader<SpinorSet>();
+  auto& vp_leader  = vp_list.getLeader();
+  IndexType nat    = vp_leader.getTotalNum();
```
ye-luo: Prefer `const size_t nat`

ye-luo (commented):
Overall LGTM.

src/QMCWaveFunctions/SpinorSet.h
camelto2: actually, this is already included in SPOSet, which is included in the line before. So I think I can just remove this altogether. For some reason, my editor thought it was needed but it doesn't

src/QMCWaveFunctions/SpinorSet.cpp
camelto2: fixed

src/QMCWaveFunctions/SpinorSet.cpp
camelto2: fixed

src/QMCWaveFunctions/SpinorSet.cpp
camelto2: fixed

ye-luo (approved):

ye-luo:
Test this please

PR #5549, opened 2025-06-16
Orbital Rotation for SplineC2COMPTarget

camelto2:

Please also follow the [GitHub Pull Request guidance](https://qmcpack.readthedocs.io/en/develop/developing.html#github-pull-request-guidance). As much as possible, try to avoid the ["Don't"s](https://qmcpack.readthedocs.io/en/develop/developing.html#don-t) to help maintain a high development velocity.

## Proposed changes
Implements orbital rotations for SplineC2COMPTarget. Basically, it keeps a copy of the spline coeffs on the host, and whenever the splines are rotated, it does the rotation on the CPU and offloads the new spline coeffs to the GPU for later use. This is the simplest approach, and should be fine for production calcualtions. For optimizations, where we will do many transfers from the host to device of the new coefficients, this may not be ideal. But it is at least something to build off of. 

- New feature

- No

El Dorado

## Checklist

**NOTE** it doesn't add a new test, it uncomments a previous guard so that applyRotation can be tested

### DIFF
--- a/src/QMCWaveFunctions/BsplineFactory/SplineC2COMPTarget.cpp
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineC2COMPTarget.cpp
@@ -116,12 +116,8 @@ void SplineC2COMPTarget<ST>::applyRotation(const ValueMatrix& rot_mat, bool use_
         spl_coefs[cur_elem + 1] = newval_i;
       }
   }
-
-  PRAGMA_OFFLOAD("omp target update to (spl_coefs[0:spline_ptr->coefs_size])")
-  {
-    spline_ptr->coefs = spl_coefs;
-  }
-
+  // update coefficients on GPU from host
+  SplineInst->finalize();
 }
 
 template<typename ST>
--- a/src/QMCWaveFunctions/BsplineFactory/SplineC2COMPTarget.cpp
+++ b/src/QMCWaveFunctions/BsplineFactory/SplineC2COMPTarget.cpp
@@ -17,6 +17,7 @@
 #include "Platforms/OMPTarget/ompReductionComplex.hpp"
 #include "ApplyPhaseC2C.hpp"
 #include "Concurrency/OpenMP.h"
+#include "CPU/BLAS.hpp"
 
 namespace qmcplusplus
 {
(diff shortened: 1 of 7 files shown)


src/QMCWaveFunctions/BsplineFactory/SplineC2COMPTarget.cpp
```diff
@@ -52,6 +53,77 @@ bool SplineC2COMPTarget<ST>::write_splines(hdf_archive& h5f)
...
+          ST wi            = rot_mat[k][j].imag();
+          newval_r += zr * wr - zi * wi;
+          newval_i += zr * wi + zi * wr;
+        }
+        spl_coefs[cur_elem]     = newval_r;
+        spl_coefs[cur_elem + 1] = newval_i;
+      }
+  }
+
+  PRAGMA_OFFLOAD("omp target update to (spline_ptr[0:1], spl_coefs[0:spline_ptr->coefs_size])")
```
ye-luo: `spline_ptr` should not be "updated to" since only coefficients are touched.

camelto2: Ok, I was mostly looking at how the coeffs initially get offloaded (MultiBSplineOffload::finalize) and used a similar offload call. Will fix

ye-luo: Actually this is still wrong. It is missing `always`.
Let us call this virtual function instead.
https://github.com/QMCPACK/qmcpack/blob/6fcce959c62c8eec1626e67f89e54207971e39f7/src/spline2/MultiBsplineBase.hpp#L95

ye-luo: Here is what we did on the R2R https://github.com/QMCPACK/qmcpack/blob/6fcce959c62c8eec1626e67f89e54207971e39f7/src/QMCWaveFunctions/BsplineFactory/SplineR2R.cpp#L190

camelto2: Looks like we don't have that in the normal SplineC2C.cpp.

camelto2: Added to both SplineC2C and SplineC2COMPTarget

ye-luo:
Test this please

PR #5556, opened 2025-06-20
Move MockGoldWalkerElements

PDoakORNL:
## Proposed changes

Moved the integration testing class MockGoldWalkerElements from `src/Estimators/Tests` to `src/Utilities/for_integration_testing` and they are now available as a target. I'm looking into shifting all the minimal pools here as well.

I plan to do some better documentation of how to get integration tests written with less pain after that.

- New feature
- Other (please describe):

- No

sdgx2
## Checklist

* * [ ] This PR is up to date with the current state of 'develop'


ye-luo (changes requested):
Utilities are for small classes with independent functionality and thus can be placed at the bottom of the target dependency chain. They are usually not physics related. Anything under `Utilities` should not depends on any classes beyond classes inside Utilities. Thus it should not depend on classes from `Particle`, `QMCWaveFunctions` and `QMCHamiltonians`.

To find a location for integration tests, we need to first choose what level of integration. If MockGoldWalkerElements involves integration up to Hamiltonian, it should be placed under `QMCHamiltonians` and then the consumers can be at `Estimators`, `QMCDrivers` and `QMCApps`. Overall, I expect `for_integration_testing` to be distributed across the cmake target or folder hierarchy serving different kinds of integration.

src/integration_testing/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+#//////////////////////////////////////////////////////////////////////////////////////
+#// This file is distributed under the University of Illinois/NCSA Open Source License.
+#// See LICENSE file in top directory for details.
+#//
+#// Copyright (c) 2025 QMCPACK developers.
+#//
+#// File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Laboratory
+#//////////////////////////////////////////////////////////////////////////////////////
+
+add_library(integration_testing MockGoldWalkerElements.cpp)
```
ye-luo: Could you document what level of integration this CMake target serves?

src/integration_testing/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+target_include_directories(integration_testing PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
+target_link_libraries(integration_testing PUBLIC minimal_testing_pools qmcparticle qmcwfs message)
```
ye-luo: missing qmcham

src/integration_testing/CMakeLists.txt
PDoakORNL: I've tried to clarify this a good bit. I've also change the targets name as a result.

src/integration_testing/CMakeLists.txt
PDoakORNL: done

ye-luo (approved):
I'm OK with the current way.

PR #5559, opened 2025-06-23
Refine minimal pool CMake targets for integration tests

PDoakORNL:
## Proposed changes
Previously the `Minimal.*Pools` have been in the respective modules tests directories.  Here they are moved into the `src/integration_testing` directory.  This simplifies the cmake and allows an easy way to provide both individual targets for minimal dependency for tests and a merged target. I think this will simplify documenting how to write integration tests.

- Build related changes
- Other (please describe):

- No

sdgx2
## Checklist


ye-luo (changes requested):
Thumb up on the making the `.h/.cpp` file split.

I like the idea of creating individual `minimal_particle_pool`, `minimal_wavefunction_pool` and `minimal_hamiltonian_pool` but the proposed file relocation made the dependency worse. For example, `minimal_particle_pool` under `src/integration_testing` needs `qmcparticle` target under `src/Paticles` while tests under `src/Particles/tests` needs `minimal_particle_pool` which lives outside `src/Particles`. To me it is nature to keep `minimal_particle_pool`, `minimal_wavefunction_pool` and `minimal_hamiltonian_pool` in `src/Particles/test`, `src/QMCWavefunction/tests`, `src/QMCHamiltonian/tests`.

ye-luo (approved):

Issue #5563, opened 2025-06-30
Problem export CASSCF wavefunctions produce by PySCF


NastaMauger:
**Describe the bug**
Impossible to export CASSCF wavefunctions produce by PySCF

I am trying to export a CAS(2,2) wavefunction using PySCF. However, when I run the following command:
`convert4qmc -orbitals CAS_2_2.h5 -production -addCusp -multidet CAS_2_2.h5`

the code fails with the following error:

```
QMCPACK ERROR HDF5 read failure in hdf_archive::read NbDet  
Fatal Error. Aborting at Unhandled Exception
```

Inspecting the HDF5 file, it appears that the call to:
```
from PyscfToQmcpack import savetoqmcpack  
savetoqmcpack(mol, casscf, title=title)
```
(using my `casscf `object) does not register the multiple configurations that should be present in the wavefunction.

Is this expected behavior? Looking to this [PR](https://github.com/QMCPACK/qmcpack/pull/5005), it should not be an issue. 
Also, using the same setup but omitting the `-multidet` keyword in the `convert4qmc` command works without any issues.

I am using: 

- PySCF: 2.9.0 with Python 3.11.11  
- QMCPACK version 4.1.9 built on Jun 10 2025

  Git branch: develop
  Last git commit: deb1df30348488ad05509040f555ba25cb4d0dcc-dirty
  Last git commit date: Tue May 27 20:22:31 2025 -0400
  Last git commit subject: Merge pull request #5506 from kayahans/bug/multidet
- My H5 file: [CAS_2_2.h5.txt](https://github.com/user-attachments/files/20987577/CAS_2_2.h5.txt)
- My PySCF script: [python.py.txt](https://github.com/user-attachments/files/20987581/python.py.txt)
- I also want to add that I had no problem to export multideterminant wavefunction generated with QP2.

prckent:
Thanks for the report and also noting that single determinant calculations are OK, as are QP2 calculations.

NastaMauger:
There is still an issue with the CASSCF wavefunctions generated by PySCF. As shown in my PR, I haven’t made any modifications to how the HDF5 file is written, so any problem related to the file itself is not caused by my changes.

In my case, I need to use specific MO in my CAS space but in each case I check my MO and they are consistent. So, here’s the issue I’m currently encountering:

[...]

PR #5564, opened 2025-07-01
Correct savetoqmcpack

NastaMauger:
## Proposed changes
The current version of `savetoqmcpack` (PyscfToQmcpack.py) does not properly handle symmetry when using multiconfigurational SCF methods. As a result, it fails if the user enables symmetry in their PySCF calculation.

Related to this issue: https://github.com/QMCPACK/qmcpack/issues/5563

- Bugfix

The current version of `PyscfToQmcpack.py` uses an explicit type comparison against `mcscf.mc1step.CASSCF:`
```
  possible_determinant_source =[mcscf.casci.CASCI, mcscf.mc1step.CASSCF]
  if type(mf) in possible_determinant_source:
``` 
This fails when the user enables symmetry in PySCF, because the returned object is a subclass like `mcscf.mc1step_symm.SymAdaptedCASSCF`. Even though it is still an `mcscf` object, the `type(mf)` check does not account for subclasses.

This PR updates the type check to use `isinstance()`:
`isinstance(mf, (mcscf.casci.CASCI, mcscf.mc1step.CASSCF))`
This ensures that both base types and any subclasses (including symmetry-adapted variants) are correctly recognized.

Added a clear and reliable condition to detect whether the wavefunction is multideterminant, ensuring proper handling of both symmetry and non-symmetry cases. Additionally, user-friendly print statements were included to notify the user when a multideterminant wavefunction has been successfully saved with the explicit name of the file.

- No

Tested with symmetry on and off; multideterminant detection works correctly.
[test.txt](https://github.com/user-attachments/files/20989433/test.txt)
## Checklist

- [x]  I have read the pull request guidance and develop docs
- [x]  This PR is up to date with the current state of 'develop'
- [x] Code added or changed in the PR has been clang-formatted


README_branch.md
```diff
@@ -0,0 +1,14 @@
...
+### `develop` branch
+This branch contains the latest development work and merges from upstream.
+
+### `Batched_Converter` branch
+This branch includes the work related to the Batched Converter feature. It is based on the forked repository from `anbenali`.
+
+### `LCAO_NumCenters_Opt` branch
+This branch contains optimizations for the `LCAO_NumCenters_Opt` feature. It is based on the forked repository from `anbenali`.
+
+
```
ye-luo: could you remove this file?

src/QMCTools/PyscfToQmcpack.py
```diff
@@ -56,7 +56,7 @@ def savetoqmcpack(cell,mf,title="Default",kpts=[],kmesh=[],sp_twist=[],weight=1.
   def get_supercell(cell,kmesh=[]):
     latt_vec = cell.lattice_vectors()
     if len(kmesh)==0:
-        # Guess kmesh
+#Guess kmesh
```
ye-luo: Could you revert all the formatting change? Bugfix PRs should avoid non-functional changes as much as possible.
Formatting changes are non-functional changes and make reviews much harder.

ye-luo: Could you mention the formatting tool you used for python? We may consider formatting all the python files in the repo.

ye-luo (commented):
Welcome and thank you for the fix.

prckent (commented):
Many thanks for investigating this and making the fix. As also noted by Ye, please can you update this to just have the functional change and not the formatting change? It makes it is more difficult to see the changes.

If you have opinions on helpful python formatting we can tackle that separately.

README_branch.md
NastaMauger: Oops — that file wasn’t meant to be there. 
It’s been removed!

src/QMCTools/PyscfToQmcpack.py
NastaMauger: I've reverted the formatting changes.

NastaMauger: In case you are still interested: I used clang-format, which is also compatible with Python.

ye-luo:
Test this please

prckent:
Thanks @NastaMauger . Unless you prefer otherwise, I think we can merge this. OK? As you noted in #5563 , clearly additional work is needed to get the overall support for PySCF multideterminant wavefunctions working correctly.

NastaMauger:
@prckent I agree.
A clear statement should be added to the manual or somewhere similar, indicating that there is a bug preventing the use of CASSCF with complex wavefunctions, so users are aware.

ye-luo (approved):

PR #5566, opened 2025-07-01
running unit tests with asan on broken for openmpi

PDoakORNL:
## Proposed changes

In my development environment every unit test fails asan for a mpi build. This fixes that. It also makes it possible to use the lsan suppression file at all with ctest.

- Bugfix
- New feature
- Build related changes
- Other (please describe):

- No

sdgx2

## Checklist

### DIFF
--- a/CMake/unit_test.cmake
+++ b/CMake/unit_test.cmake
@@ -1,5 +1,7 @@
 include(test_labels)
 
+include(CMakePrintHelpers)
+
 # Runs unit tests
 function(ADD_UNIT_TEST TESTNAME PROCS THREADS TEST_BINARY)
   message(VERBOSE "Adding test ${TESTNAME}")
@@ -18,8 +20,13 @@ function(ADD_UNIT_TEST TESTNAME PROCS THREADS TEST_BINARY)
   endif()
 
   if(TEST_ADDED)
-    set_tests_properties(${TESTNAME} PROPERTIES PROCESSORS ${TOT_PROCS} ENVIRONMENT OMP_NUM_THREADS=${THREADS}
-                                                PROCESSOR_AFFINITY TRUE)
+    set(TEST_ADDITIONAL_ENV
+        $<$<STREQUAL:${ENABLE_SANITIZER},asan>:LSAN_OPTIONS=suppressions=${PROJECT_SOURCE_DIR}/config/sanitizers/lsan.supp>
+    )
+    set_tests_properties(
+      ${TESTNAME} PROPERTIES PROCESSORS ${TOT_PROCS} ENVIRONMENT "OMP_NUM_THREADS=${THREADS};${TEST_ADDITIONAL_ENV}"
+                             PROCESSOR_AFFINITY TRUE)
+    cmake_print_properties(TESTS ${TESTNAME} PROPERTIES ENVIRONMENT)
 
     if(ENABLE_CUDA
        OR ENABLE_ROCM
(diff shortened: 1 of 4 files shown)


CMake/unit_test.cmake
```diff
@@ -18,8 +20,13 @@ function(ADD_UNIT_TEST TESTNAME PROCS THREADS TEST_BINARY)
...
   if(TEST_ADDED)
-    set_tests_properties(${TESTNAME} PROPERTIES PROCESSORS ${TOT_PROCS} ENVIRONMENT OMP_NUM_THREADS=${THREADS}
-                                                PROCESSOR_AFFINITY TRUE)
+    set(TEST_ADDITIONAL_ENV
+        $<$<STREQUAL:${ENABLE_SANITIZER},asan>:LSAN_OPTIONS=suppressions=${PROJECT_SOURCE_DIR}/config/sanitizers/lsan.supp>
+    )
+    set_tests_properties(
+      ${TESTNAME} PROPERTIES PROCESSORS ${TOT_PROCS} ENVIRONMENT "OMP_NUM_THREADS=${THREADS};${TEST_ADDITIONAL_ENV}"
+                             PROCESSOR_AFFINITY TRUE)
+    cmake_print_properties(TESTS ${TESTNAME} PROPERTIES ENVIRONMENT)
```
ye-luo: I think it can be better to have at least a way to override `LSAN_OPTIONS` instead of being hard-coded internally.

at top level cmakelists.txt around asan.
```
  set(LSAN_OPTIONS "suppressions=${PROJECT_SOURCE_DIR}/config/sanitizers/lsan.supp" CACHE string "cached value of environment variable LSAN_OPTIONS used by ctest")
```
using `APPEND` will be cleaner 
```
set_tests_properties(...) # keep it as it was
if("asan" IN_LIST ENABLE_SANITIZER)
  set_property(TEST ${TESTNAME} APPEND PROPERTY ENVIRONMENT LSAN_OPTIONS=${LSAN_OPTIONS}
endif
```
It will be better if we have this also in "RUN_QMC_APP_NO_COPY" covering all ctest runs.

printing property at every call is an overkill. Let's remove.
This can be revealed by simply running ctest with `-VV`

ye-luo (commented):
Very useful. We can potentially enable ASAN on MPI builds in CI without much headache.

CMake/unit_test.cmake
PDoakORNL: I didn't know that I could use append like that.
Agree on the print, I should have removed that.

Currently it is not possible at least with ccmake to enable more than one sanitizer at a time, is this actually possible?

ye-luo: >is this actually possible?

I don't think so. but we won't lose anything by keeping it flexible I assume. Hopefully we can run multiple sanitizer together instead of one at a time.

CMakeLists.txt
```diff
@@ -168,10 +168,12 @@ mark_as_advanced(QMC_EXP_THREADING)
 #--------------------------------------------------------------------
 
 # Add optional sanitizers ASAN, UBSAN, MSAN, TYPESAN
-set(VALID_SANITIZERS "asan" "ubsan" "tsan" "msan" "typesan")
+set(VALID_SANITIZERS "" "asan" "ubsan" "tsan" "msan" "typesan")
```
ye-luo: Empty string should be not a value of "VALID_SANITIZERS". It is can be risky like nullptr. Is there a reason for this change?

PDoakORNL: I was trying to add that as an option so ENABLE_SANITIZERS could be set back to none.  It doesn't work.  The only thing that can be done is force the cache variable back to "" with the command line.

ye-luo (approved):

CMakeLists.txt
ye-luo: If you don't use command line, editing `CMakeCache.txt`
```
//asan,ubsan,tsan,msan,typesan
ENABLE_SANITIZER:STRING=
```
should still bring sanitizer to none and pass validation.

PR #5567, opened 2025-07-01
WalkerLogXXX should provide move constructors

PDoakORNL:
## Proposed changes

I see no documented reason for why move constructors or the rule of five in general should be broken for `WalkerLog*` classes.  And it became a subtle use after return asan issue when refactoring batched drivers. I have addressed this sufficiently to address work needed for testing of variable estimator measurements to be sound and testable. I have also added the the beginning of actual unit testing for WalkerLogManager and a testing data class for WalkerLogInput.

There was a nasty shared state reference which so far looks to have been unused.  This will all get more coverage int he variable esimator measurement period PR.

- Bugfix
- New feature
- Other (please describe):

- No

## Checklist

### DIFF
--- a/src/QMCDrivers/WalkerLogCollector.h
+++ b/src/QMCDrivers/WalkerLogCollector.h
@@ -96,13 +96,20 @@ private:
   Array<WLog::PsiVal, 2> Gtmp;
   /// tmp storage for walker wavefunciton laplacians
   Array<WLog::PsiVal, 1> Ltmp;
-  /// state data set by WalkerLogManager
-  const WalkerLogState& state_;
+  /** This is just a bundle of constructor arguments, if the intention
+   * was dynamic manipulation of object state through a reference
+   * back door. Since it wasn't used this way I turned it to the by
+   * value.
+   * If you want to do a state transform write a function to do it
+   * and document it.
+   */
+  WalkerLogState state_;
 
 public:
   /// constructor. The state should be given by the manager.
   WalkerLogCollector(const WalkerLogState& state);
-
+  WalkerLogCollector(WalkerLogCollector&& other)            = default;
+  WalkerLogCollector& operator=(WalkerLogCollector&& other) = default;
   /// resize buffers to zero rows at beginning of each MC block
   void startBlock();
 
--- a/src/QMCDrivers/WalkerLogCollector.h
+++ b/src/QMCDrivers/WalkerLogCollector.h
@@ -96,12 +96,14 @@ private:
   Array<WLog::PsiVal, 2> Gtmp;
   /// tmp storage for walker wavefunciton laplacians
   Array<WLog::PsiVal, 1> Ltmp;
-  /** This is just a bundle of constructor arguments, if the intention
-   * was dynamic manipulation of object state through a reference
-   * back door. Since it wasn't used this way I turned it to the by
-   * value.
-   * If you want to do a state transform write a function to do it
-   * and document it.
+  /** Hopefully This is just a bundle of constructor arguments.
+   *  It was a reference, so perhaps, the intention was dynamic
+   *  manipulation of a group of objects state from afar.
+   *  Since it hadn't yet been used this way its just a private member
+   *  now. _reviewers_ do not allow it to be made a reference again.
+   *
+   *  If you want to do a state transform write a function, document
+   *  it, call it from a sensible and obvious scope.
    */
   WalkerLogState state_;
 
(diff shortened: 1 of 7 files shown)


src/QMCDrivers/WalkerLogCollector.h
```diff
@@ -96,13 +96,22 @@ class WalkerLogCollector
...
+  /** Hopefully This is just a bundle of constructor arguments.
+   *  It was a reference, so perhaps, the intention was dynamic
+   *  manipulation of a group of objects state from afar.
+   *  Since it hadn't yet been used this way its just a private member
+   *  now. _reviewers_ do not allow it to be made a reference again.
+   *
+   *  If you want to do a state transform write a function, document
+   *  it, call it from a sensible and obvious scope.
+   */
+  WalkerLogState state_;
```
ye-luo, line 108: IIRC. `WalkerLogManager` owns and manipulates the `state_` and `WalkerLogState` only reads it. In this way, per crowd `WalkerLogState` won't go out of sync. For that reason, keeping it a const ref seems fine to me. What is your motivation to make it a copy especially a non-const copy?

ye-luo: Sent out review before my laptop died. Continue reading the added test.

src/QMCDrivers/tests/test_WalkerLogManager.cpp
```diff
@@ -0,0 +1,69 @@
...
+  bool okay = doc.parseFromString(WLInput::getXml(WLInput::valid::DEFAULT));
+  REQUIRE(okay);
+  xmlNodePtr node = doc.getRoot();
+  WalkerLogInput walker_log_input{node};
+  auto make_stuff = [comm](WalkerLogInput& walker_log_input) -> LogAndStuff {
+    WalkerLogManager wlm{walker_log_input, true, "root_name", comm};
+    CollectorHolder ch;
+    ch.setWalkerLogCollector(wlm.makeCollector());
+    return {std::move(wlm), std::move(ch)};
+  };
```
ye-luo, line 52: This factory function shows the tight connection between the manager and collectors as it returns them as a whole structure.

We can write the code as
```
   auto make_stuff = [comm](WalkerLogInput& walker_log_input) -> LogAndStuff {
-    WalkerLogManager wlm{walker_log_input, true, "root_name", comm};
-    CollectorHolder ch;
-    ch.setWalkerLogCollector(wlm.makeCollector());
-    return {std::move(wlm), std::move(ch)};
+    LogAndStuff logger{{walker_log_input, true, "root_name", comm}, {}};
+    logger.ch.setWalkerLogCollector(logger.wlm.makeCollector());
+    return logger;
   };
```
In this way, the hassle of move constructors can be avoided and also reduce the number copy operations.

The intention of introducing move constructor and operator is to bring real benefit. It is not worth maintaining move constructor and operators rigorously in every class but only places we benefit. Hopefully everyone agrees with that.

src/QMCDrivers/WalkerLogCollector.h
PDoakORNL: You cannot move a ref.

ye-luo: Can you make it at least a `const`?

src/QMCDrivers/tests/test_WalkerLogManager.cpp
ye-luo: The above was my opinion. This is not a request for change.
We can keep the code as you changed for now.

src/QMCDrivers/WalkerLogCollector.h
PDoakORNL: So multiple threads spectate on a single memory location to stay "synchronized"? examining the source shows no evidence this anti feature is used and no crowd level resources state should be getting controlled in this manner.  If this needs to be updated then a call can be made to change the behavior of the per crowd loggers.

src/QMCDrivers/tests/test_WalkerLogManager.cpp
PDoakORNL: Is there a point in blocking this other than making the refactoring I need to do to implement effective testing for a requested feature harder?  I think ease of testing refactoring trumps weasling out of providing sane threaded design and basic constructors.  

I'm correcting a design issue that should have never gotten through review the reference used as a back door state update and fixing the lack of clear declaration of constructors.

ye-luo: Probably you missed my other comment, I'm not making a request for change in this area of code. The blocking reason is missing CMake changes in this PR. Your added code is not compiled and tested.

To be honest, the name state made me feel it may carry variable like step counters. If that was the case, I wanted it to be a const ref in collectors and let the manager to operate it.
But the source code `WalkerLogState` really shows it is just parsed input. So copy is fine.

ye-luo:
Test this please

PR #5568, opened 2025-07-02
Restrict asan suppression to only the MPI case.

ye-luo:
## Proposed changes
a https://github.com/QMCPACK/qmcpack/pull/5566 follow-up

- No

epyc-server

## Checklist


prckent (commented):
1. I think it would be an improvement to have a note in the printout about the suppressions being used.
2. Have you run a comprehensive set of tests with this change? e.g. ctest -E 'long|performance' ?

ye-luo:
> 1. I think it would be an improvement to have a note in the printout about the suppressions being used.

Got LSAN_OPTION printed if it is not empty.

>     2. Have you run a comprehensive set of tests with this change? e.g. ctest -E 'long|performance' ?

No. Only ran the whole deterministic category which covers unit tests and some full simulation runs. They pass on my workstation.

PR #5574, opened 2025-07-15
Implement variable period estimator management feature and estimator_measurement_period input parameter

PDoakORNL:
## Proposed changes

Sometimes its desirable to accumulate measurements for an estimator on some sort of periodic schedule. This PR adds an implementation of this feature. Since this is a feature that requires many steps and active estimators to test.  Testing will be delayed.

- New feature
- Other (please describe):

- No

sdgx2, frontier

## Checklist

### DIFF
--- a/src/Estimators/EstimatorManagerNew.cpp
+++ b/src/Estimators/EstimatorManagerNew.cpp
@@ -258,6 +258,10 @@ void EstimatorManagerNew::stopBlock(unsigned long accept, unsigned long reject, 
   makeBlockAverages(accept, reject);
   reduceOperatorEstimators();
   writeOperatorEstimators();
+  // This is an entry point for rank level estimators to do one last
+  // thing before being zerod
+  for (auto& op_est : operator_ests_)
+    op_est->stopBlock();
   zeroOperatorEstimators();
   // intentionally put after all the estimator I/O
   PropertyCache[cpuInd] = block_timer_.elapsed();
--- a/src/QMCDrivers/DMC/DMCContextForSteps.h
+++ b/src/QMCDrivers/DMC/DMCContextForSteps.h
@@ -1,0 +1,24 @@
+//////////////////////////////////////////////////////////////////////////////////////
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2025 QMCPACK developers.
+//
+// File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Laboratory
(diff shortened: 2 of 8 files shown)


src/Estimators/EstimatorManagerNew.cpp
```diff
@@ -258,6 +258,10 @@ void EstimatorManagerNew::stopBlock(unsigned long accept, unsigned long reject,
   makeBlockAverages(accept, reject);
   reduceOperatorEstimators();
   writeOperatorEstimators();
+  // This is an entry point for rank level estimators to do one last
+  // thing before being zerod
+  for (auto& op_est : operator_ests_)
+    op_est->stopBlock();
```
ye-luo, line 264: repeat my earlier comment:
stopBlock got called in EstimatorManagerNew but startBlock is called in EstimatorManagerCrowd. The unpaired call locations indicates underlying design issues.

@PDoakORNL if you have any thoughts, please comment here. We can do the actual code change later if we figures a better way.

src/QMCDrivers/DMC/DMCContextForSteps.h
```diff
@@ -0,0 +1,24 @@
...
+// This file is distributed under the University of Illinois/NCSA Open Source License.
+// See LICENSE file in top directory for details.
+//
+// Copyright (c) 2025 QMCPACK developers.
+//
+// File developed by: Peter Doak, doakpw@ornl.gov, Oak Ridge National Laboratory
+//////////////////////////////////////////////////////////////////////////////////////
+
+#include <NonLocalTOperator.h>
+#include <QMCDriverNew.h>
```
ye-luo, line 11: Is including QMCDriverNew.h an overkill? Probably just need the file defining RandomBase. In a second thought, we probably won't really use this class in a self-sufficient manner and thus doesn't matter much,

ye-luo:
Test this please.

src/Estimators/EstimatorManagerNew.cpp
PDoakORNL: Careful examination shows that there should be a call to startBlock on all the rank level estimators but that since so far it doesn't seem to have created an bugs not calling these methods on the rank level estimator it hasn't made a bug yet.  I will modify EstimatorManagerNew::startBlock to fix this.

src/QMCDrivers/DMC/DMCContextForSteps.h
PDoakORNL: yes no longer necessary

ye-luo:
Notced that the newly introduced option still needs user manual update.

PR #5584, opened 2025-07-29
[AFQMC] Add operator+ to shared memory fancy pointer

correaa:
## Proposed changes

Add operator+ to shm fancy pointer, so it fulfills the pointer concept from the point of view of Multi

- No

Mac, linux

## Checklist

* * [x ] This PR is up to date with the current state of 'develop'


correaa:
Thanks for asking.

Multi works with arbitrary pointer-like objects, as long as they fulfill the "pointer concept". 
(this includes both pre and post addition, p + n and n + p)

After C++20, and with concepts formally in the language, the requirements will eventually become more strict, so preemptively I am applying this simple patch.

(The pre C++17 philosophy is, if it compiles, it works.
But as "requires" keywords becomes more pervasive, more syntactic features and semantic consistency will be demanded of the user-provided types, in the function interfaces, whether these properties are used or not in the function implementation.)

Preparing dependent code will allow me to use C++20 and requires clauses in the library and replace hacky enable_if, etc. 
(don't worry, Multi is still C++17 as long as you need it to be)

Having said that, at this point the custom "GPU pointers" could be replaced by thrust::cuda/hip::pointers and delegate all this details to a library.
A lot of this functionality was reimplemented as Qmcpack_gpu_pointer back when Thrust _looked_ unviable.

PR #5586, opened 2025-08-05
Let CMake to look up NVTX 3

ylee88:
## Proposed changes

Since CUDA 12.9, NVTX is no longer shipped as a shared library. Instead, a header-only version, `NVTX v3`, is encouraged. This PR allows CMake to find `NVTX v3` first, and fall back to the legacy shared library if it is not found. 

- Build related changes

- No

## Checklist


ye-luo (commented):
LGTM.

PR #5594, opened 2025-08-12
Slight adjustment in batched delayed update

ye-luo:
## Proposed changes
No need to set z part to zero when updating Binv.
Also update source code comments.

- Other (please describe):

- No

Aurora

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/Fermion/DelayedUpdateBatched.h
+++ b/src/QMCWaveFunctions/Fermion/DelayedUpdateBatched.h
@@ -645,11 +645,11 @@ public:
     //std::copy_n(Ainv[rowchanged], norb, V[delay_count]);
     compute::BLAS::copy_batched(blas_handle, norb, invRow_mw_ptr, 1, V_row_mw_ptr, 1, nw);
     // handle accepted walkers
-    // the new Binv is [[X Y] [Z sigma]]
+    // the new Binv is [[X y] [z sigma]]
     //BLAS::gemv('T', norb, delay_count + 1, cminusone, V.data(), norb, psiV.data(), 1, czero, p.data(), 1);
     compute::BLAS::gemv_batched(blas_handle, 'T', norb, delay_count, cminusone_vec.device_data(), V_mw_ptr, norb,
                                 phiVGL_mw_ptr, 1, czero_vec.device_data(), p_mw_ptr, 1, n_accepted);
-    // Y
+    // y
     //BLAS::gemv('T', delay_count, delay_count, sigma, Binv.data(), lda_Binv, p.data(), 1, czero, Binv.data() + delay_count,
     //           lda_Binv);
     compute::BLAS::gemv_batched(blas_handle, 'T', delay_count, delay_count, ratio_inv_mw_ptr, Binv_mw_ptr, lda_Binv,
--- a/src/QMCWaveFunctions/detail/CUDA/matrix_update_helper.cu
+++ b/src/QMCWaveFunctions/detail/CUDA/matrix_update_helper.cu
@@ -321,20 +321,16 @@ __global__ void add_delay_list_save_sigma_VGL_kernel(int* const delay_list[],
 
   if (iw < n_accepted)
   {
-    // real accept, settle y and Z
+    // real accept
     int* __restrict__ delay_list_iw = delay_list[iw];
     T* __restrict__ binvrow_iw      = binv[iw] + delay_count * binv_lda;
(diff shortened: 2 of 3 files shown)


src/QMCWaveFunctions/detail/CUDA/matrix_update_helper.cu
```diff
@@ -370,15 +372,17 @@ __global__ void add_delay_list_save_sigma_VGL_kernel(int* const delay_list[],
...
     const int num_blocks_delay_count = (delay_count + COLBS - 1) / COLBS;
     for (int ib = 0; ib < num_blocks_delay_count; ib++)
     {
       const int col_id = ib * COLBS + tid;
       if (col_id < delay_count)
-        binv_iw[delay_count * binv_lda + col_id] = binv_iw[delay_count + binv_lda * col_id] = T(0);
+        binv_iw[delay_count + binv_lda * col_id] = T(0);
     }
 
+    // Set x to zero
```
PDoakORNL: Seems like this copy is a typo, based on the cpu version and old comment x should get set to 1 assuming
```
x = binv_iw[delay_count * binv_lda + delay_count]
```

src/QMCWaveFunctions/detail/CUDA/matrix_update_helper.cu
```diff
@@ -370,15 +372,17 @@ __global__ void add_delay_list_save_sigma_VGL_kernel(int* const delay_list[],
...
 
+    // Set y to zero
     T* __restrict__ binv_iw          = binv[iw];
     const int num_blocks_delay_count = (delay_count + COLBS - 1) / COLBS;
     for (int ib = 0; ib < num_blocks_delay_count; ib++)
     {
       const int col_id = ib * COLBS + tid;
       if (col_id < delay_count)
-        binv_iw[delay_count * binv_lda + col_id] = binv_iw[delay_count + binv_lda * col_id] = T(0);
+        binv_iw[delay_count + binv_lda * col_id] = T(0);
```
PDoakORNL, line 382: Ok is 
```
y = binv_iw[delay_count + binv_lda * col id]
where col_id = ib * COLBS + tid -> ib = [0...num_block_delay count]
```
If so then this comment makes sense although perhaps the comment can over that value

src/QMCWaveFunctions/detail/CUDA/matrix_update_helper.cu
```diff
@@ -370,15 +372,17 @@ __global__ void add_delay_list_save_sigma_VGL_kernel(int* const delay_list[],
         Urow_iw[col_id] = T(0);
```
PDoakORNL, line 372: I guess this is z

src/QMCWaveFunctions/Fermion/DelayedUpdateBatched.h
```diff
@@ -645,11 +645,11 @@ class DelayedUpdateBatched
     //std::copy_n(Ainv[rowchanged], norb, V[delay_count]);
     compute::BLAS::copy_batched(blas_handle, norb, invRow_mw_ptr, 1, V_row_mw_ptr, 1, nw);
     // handle accepted walkers
-    // the new Binv is [[X Y] [Z sigma]]
+    // the new Binv is [[X y] [z sigma]]
```
PDoakORNL, line 648: So is the meaning here that y and z are vectors and X a matrix?

PDoakORNL (dismissed):
Seem like there was a real bug fix in the gpu kernel.  But the notation i.e. Xx Yy Zz is still confusing to me.  I assume there is a relation between the values the helpers are updating based on accept or pseudo accept and the comment in the DelayedUpdateBatched this seems to be the time to clarify this.

src/QMCWaveFunctions/Fermion/DelayedUpdateBatched.h
ye-luo: Upper case is a matrix and lower case is a vector. Names are aligned with the paper I'm working on.

src/QMCWaveFunctions/detail/CUDA/matrix_update_helper.cu
ye-luo: Seems so. Let me fix it.

src/QMCWaveFunctions/detail/CUDA/matrix_update_helper.cu
ye-luo: This is U. Not part of Binv.

PDoakORNL (approved):
LGTM

PR #5600, opened 2025-09-03
Batched implementation of G of r ( g(r), PairCorrEstimator )

PDoakORNL:
## Proposed changes

Port of the PairCorrEstimator from QMCHamiltonian to where it belongs for batched under Estimators.
I added it the back the the batched HEG test and the estimator writes the expected pair correlation bins to the stat.h5.

As far as I can tell even though legacy HEG included this estimator in its input no actual test of the output was every made.  I have carried forward the equivalent functionality here.

- New feature
- Build related changes

- No

sgdx2

## Checklist

### DIFF
--- a/src/Platforms/SYCL/syclBLAS.cpp
+++ b/src/Platforms/SYCL/syclBLAS.cpp
@@ -93,7 +93,7 @@ template sycl::event gemv(sycl::queue& handle,
 
 /** gemv trans = 'T' case. COLS refers to columns of the m x n column-major Fortran matrix A.
  */
-template<typename T, unsigned COLBS>
+template<typename T, unsigned ROWBS, unsigned COLBS>
 sycl::event gemvT_batched_impl(sycl::queue& handle,
                                const int m,
                                const int n,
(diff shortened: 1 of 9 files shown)


prckent (commented):
Thanks Peter. Wasn't expecting the docs. This is now better described than the legacy version which barely exists in the documentation.

For the reviewers, my request was to get this in and that we will explore proper testing later. The state of the legacy estimator also needs to be checked; it is likely to be good but we don't know for sure.

Unfortunately asan found issues below EstimatorManagerNew::stopBlock...

src/Particle/ParticleSetPool.h
```diff
@@ -89,7 +89,7 @@ class ParticleSetPool : public MPIObjectBase
 
   /** get the Pool object
    */
-  inline const PoolType& getPool() const { return myPool; }
+  inline PoolType& getPool() { return myPool; }
```
ye-luo, line 92: Could you keep both the const and non-const versions?

PDoakORNL: No not without changing the way we have in legacy let QMCHamiltonian estimators change the distance tables of the golden particle sets.

PDoakORNL:
The ASAN issue is a bit strange, its flagging on reading something 0 bytes to the right of the allocated space and I can't reproduce that with a new clang yet.  I did add some asserts that seem to prove there is no overflow, but I want to look at it a bit more.

src/Particle/ParticleSetPool.h
ye-luo: Let us leave the code as it is. I will write a separate comment regarding how to address the overall issue in a later PR.

src/Estimators/PairCorrelationEstimator.cpp
```diff
@@ -0,0 +1,329 @@
...
+  {
+    delta_    = input_.get_delta();
+    num_bins_ = static_cast<int>(rmax_ / delta_);
+  }
+  else if (input_.get_explicit_set_nbins() && input_.get_explicit_set_delta())
+  {
+    // Here we must assume that certain checks were done in parsing.
+    delta_    = input_.get_delta();
+    num_bins_ = input_.get_nbins();
+    if (rmax_ != delta_ * num_bins_)
```
ye-luo: https://github.com/QMCPACK/qmcpack/actions/runs/17499986682/job/49710340630?pr=5600
indicates that this check is not numerically robust.

PDoakORNL: That is probably so, but it also indicated some input class logic that wasn't finished.  I'll also address the possible numerical issue here.

PDoakORNL:
I was just trying for minimum change here, since that was the agreed way for this to be done. But `target` is a qmcsection attribute input argument and so it needs to be an argument here as well, or at least logically it would be. There is a long precedent for allowing users to name any particle set anything they want, although legacy hamiltonian construction is inconsistent as to whether it respects target attributes of operators or just gives them what the qmcsection says is the target.  Hamiltonian factory seems to ignore that target attribute (although it reads it) for all operators except `lattice deviation`.

src/Estimators/PairCorrelationEstimator.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+  // legacy behavior is that the input rmax wins these are overwritten by input tags for rmax
+  if (elecs.getLattice().SuperCellEnum)
+  {
+    rmax_   = input_.get_explicit_set_rmax() ? input_.get_rmax() : elecs.getLattice().WignerSeitzRadius;
+    volume_ = elecs.getLattice().Volume;
+  }
+  else // Open BC's
+  {
+    rmax_   = input_.get_rmax();
```
ye-luo, line 41: Do we need an error trap if rmax is not given in the case of open BC? magic number 10 is not a good choice.

PDoakORNL: I'm not really sure what to do here, legacy has no check of the sanity of these arguments at all and just uses the magic number.  So for this PR I think this is out of scope and this just needs to get merged.  An issue asking that the estimator be able to pick more appropriate defaults would be good to get on the books.

prckent: Default magic number is fine here. The code will not crash or give wrong results. Out of scope for this PR.

[ More generally, we don't have any consistent logic for open boundary condition cutoffs (e.g. density estimators). We could have defaults or better defaults, but these don't appear to trip people up since rerunning is easy. _If_ we were to spend any time on these things, I would argue for good automatic defaults for density grids for atoms and molecules and solids before touching other estimators. ]

ye-luo: I agree that it is beyond the scope of this PR. It is a repetitive issue and we need to choose a consistent approach. https://github.com/QMCPACK/qmcpack/issues/5602
I don't agree with "rerunning". It is better to stop users early than inducing wasteful spending.

src/Estimators/PairCorrelationEstimator.cpp
```diff
@@ -0,0 +1,338 @@
...
+      dlist.push_back(elecs.getDistTable(k).get_origin().getName());
+
+  std::set<int> others_sorted;
+  auto& sources = input_.get_sources();
+  for (int i = 0; i < sources.size(); ++i)
+  {
+    int k = 0;
+    while (k < dlist.size())
+    {
+      if (slist[i] == dlist[k])
```
ye-luo, line 108: I noticed the above code seems to be very ill.
`sources.size()` is used to go though elements of `slist`

ye-luo: This issue was spotted when looking at `sources` input entry. Some additional thoughts.
The current default is "e" but we can do better to leave the default value empty.
When consuming the input class in the estimator, we can fill in the implicit `target` set name which is better than a hard coded "e".

Issue #5608, opened 2025-09-11
compilation error "uint_fast32_t is not a type"


correaa:
**Describe the bug**

I am getting a compilation error with QMCPACK, 

```
/builds/correaa/boost-mpi3/qmcpack/src/Utilities/RandomBase.h:28:63: error: 'uint_fast32_t' does not name a type [-Wtemplate-body]
   28 |   using uint_type                                           = uint_fast32_t;
```

**To Reproduce**

https://gitlab.com/correaa/boost-mpi3/-/jobs/11314820453

```
$ cmake -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx -DBUILD_AFQMC=1 -DBUILD_PPCONVERT=1 -DQMC_MIXED_PRECISION=1 -DCMAKE_BUILD_TYPE=Debug -DMPIEXEC_PREFLAGS="--allow-run-as-root;--bind-to;none" ..
$ make
```

compiler is g++-15

**Expected behavior**

It should compile without error

**System:**

Ubuntu 24.10 CI machine

prckent:
I believe fixed in #5609  please reopen if not

PR #5609, opened 2025-09-12
starting with gcc 15 stdlibc++ headers don't include stdint.h

PDoakORNL:
## Proposed changes

For stdlibc++'s coming with gcc >15 the c++ std library headers contaminate the name space less with other headers.  Issue #5608 seems to have run into this where we depend on C unint types defined in stdint.h.

Since we have already partially moved to c++'s namespaced uint types from the `cstdint` header in many locations this PR just finishes this.

- Bugfix

- No

## Checklist

* * [ ] This PR is up to date with the current state of 'develop'

### DIFF
--- a/src/Utilities/PrimeNumberSet.h
+++ b/src/Utilities/PrimeNumberSet.h
@@ -19,6 +19,7 @@
 #define QMCPLUSPLUS_PRIME_NUMBER_SET_H
 #include <cstdint>
 #include <limits>
+#include <vector>
 
 ///dummy declaration
 template<typename UIntType>


src/QMCWaveFunctions/Fermion/MultiDiracDeterminant.2.cpp
```diff
@@ -22,6 +22,9 @@
 #include "OMPTarget/ompBLAS.hpp"
 #include "OMPTarget/ompReductionComplex.hpp"
 #include "OhmmsPETE/ompReductionTinyVector.hpp"
+#include <cstdint>
+
+using uint32_t = std::uint32_t;
```
ye-luo, line 27: Is this required by gcc 15? uint32_t is a type that doesn't require the std namespace.
https://en.cppreference.com/w/cpp/types/integer.html

PDoakORNL: I'm not sure but my reading and that of others is that whether `<cstdint>` adds the sized integers to the global namespace is implementation specific and not portable.

https://stackoverflow.com/questions/13129945/use-types-in-cstdint-with-or-without-namespace.

For this file and MultiSlaterDetTableMethod since uint32_t was used many times and they are implementation files I just defined a compilation unit scope type alias.

ye-luo: OK. Let us stick to it.

prckent:
Thanks Peter

PR #5614, opened 2025-09-18
[Nexus] Remove Deprecated NumPy Aliases `numpy.string_` and `numpy.float_`

Kcorbyerd:
## Proposed changes

The NumPy aliases `numpy.string_` and `numpy.float_` were removed in NumPy 2.0.0. The correct aliases `numpy.bytes_` and `numpy.float64` have been available since at least NumPy 1.13, and therefore can replace the deprecated aliases. 

Further, the function for NumPy compatibility in `testing.py` was removed and the command for altering print settings was placed in the try-except statement to retain desired print formatting but remove the unnecessary function. 

All instances where `numpy.bytes_` were added to strings were altered so the `numpy.bytes_` object was first converted to string to ensure compatibility.

- Bugfix

- No

Checked compatibility with default python environment on Perlmutter, Python 3.6.15, NumPy 1.17.3, as well as the most recent python environment on Perlmutter, Python 3.13.2, NumPy 2.1.3. Found `numpy.bytes_` and `numpy.string_` are equivalent and produce no errors. Also found that `numpy.float_` and `numpy.float64` are equivalent and produce no errors.

## Checklist

### DIFF
--- a/nexus/lib/rmg_input.py
+++ b/nexus/lib/rmg_input.py
@@ -2414,10 +2414,10 @@ def write_double_array(v):
 
 
 rmg_value_types = obj({
-    'string'        : (str  , np.string_),
+    'string'        : (str  , np.bytes_),
     'boolean'       : (bool , np.bool_  ),
     'integer'       : (int  , np.int_   ),
-    'double'        : (float, np.float_ ),
+    'double'        : (float, np.float64 ),
     'integer array' : (tuple,list,np.ndarray),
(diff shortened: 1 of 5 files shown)


prckent:
Many thanks Brock (@Kcorbyerd), this is very much appreciated. Clearly #5215 did not get all of the deprecated code. Did you find these by running the regular test set (which appeared to be clean here) or something else?

PR #5617, opened 2025-09-19
[Nexus] Change Structure.center_molecule() to behave more robustly

Kcorbyerd:
## Proposed changes

A function in `structure.py`, `Structure.center_molecule()`, previously shifted the average atomic coordinates to the 1/2 point in the unit cell, however this could introduce unexpected behavior when working with molecular systems with a large spatial extent. If a molecule had a large number of atoms in a small cluster and also had a long tail (e.g. an indole with a long alkane tail), the center of atomic coordinates would be localized largely in the cluster, while the tail would be significantly further away from that center. Should the molecule be shifted then so that that center is in the center of the unit cell, it is possible that some of the atoms in the tail of the molecule would be in an adjacent unit cell, which in the case of isolated molecular species would alter the properties of the molecule, such as the molecular dipole moment (see [Figure 1 in this paper](https://doi.org/10.1103/PhysRevB.51.4014)).

To fix this, the proposed algorithm now shifts the molecule to the (+x, +y, +z) corner of the unit cell closest to the origin, calculates the smallest gap between the edge of the spatial extent of the molecule and the unit cell edges, and shifts the molecule so that it is equally far away from each side of the unit cell in every direction. 

An example of a modified form of an adamantane core with an alkene tail is shown below, old centering first, then new centering:

old_centering

new_centering

As is clear here, the previous centering method allows for the molecule to exceed the edges of the unit cell, but the new method contains the entire molecule. It should be noted that these were generated using identical inputs, with identically sized unit cells, the only change was the algorithm used.

This is an exaggerated situation, often the differences would be more subtle, however this will catch situations that may occur for the common user, such as modeling a polymer with an elongated functional group that contains a significant number of atoms.

One possible issue with this method is that it is only applicable for orthorhombic, tetragonal, and cubic unit cells. While this does seem to be a significant issue, the vast majority of molecular calculations in periodic code are done in these types of unit cells. If need be, a fix for the remaining types of cell is possible, however this should suffice for essentially every user case.

- Bugfix
- New feature

- No

Tested to work on Perlmutter, Python 3.13.2, NumPy 2.1.3.

## Checklist

### DIFF
--- a/nexus/lib/structure.py
+++ b/nexus/lib/structure.py
@@ -1206,7 +1206,31 @@ class Structure(Sobj):
 
 
     def center_molecule(self):
-        self.slide(self.center-self.pos.mean(0),recenter=False)
+        """Center a molecule in a unit cell, ensuring equal padding on all sides
+        
+        Note
+        ----
+        Only works in orthorhombic, tetragonal, and cubic unit cells!
+        """
+
+        # Find smallest X, Y, and Z coordinate
+        min_xyz = np.min(self.pos, 0)
+
+        # Shift molecule to be in the corner of the +x, +y, +z octant (think first quadrant but in 3D)
+        for i in range(3):
+            self.pos[:, i] = self.pos[:, i] - min_xyz[i]
+
+        # Find the minimum distance between the molecule's new xyz coordinates 
+        # and the maximum xyz extent of the unit cell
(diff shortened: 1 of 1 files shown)


prckent:
Yes, question answered :-) We do have an active test in CI that goes through this code path.

prckent:
Test this please

prckent (approved):
Approving based on the test passing now and the conversation above.

Note that you still have Jaron tagged for review.

PR #5621, opened 2025-09-23
[Nexus] Updated version of #5619 with accepted changes

Kcorbyerd:
## Proposed changes

Update formatting for

`foo==None` -> `foo is None`
`foo!=None` -> `foo is not None`
`not element in list` -> `element not in list`
`except... None` -> `except... pass`
`var = lambda...` -> `def var()`

See #5619 for discussion.

- No

## Checklist

### DIFF
--- a/nexus/lib/pseudopotential.py
+++ b/nexus/lib/pseudopotential.py
@@ -1611,8 +1611,7 @@ class GaussianPP(SemilocalPP):
                 n = int(lines[i]); i+=1
                 terms = []
                 for j in range(n):
-                    coeff,rpow,expon = lines[i].split()
-                    i+=1
+                    coeff,rpow,expon = lines[i].split(); i+=1
                     terms.append((float(coeff),int(rpow),float(expon)))
                 #end for
                 channels.append(terms)
--- a/nexus/lib/pseudopotential.py
+++ b/nexus/lib/pseudopotential.py
@@ -90,7 +90,7 @@ class PseudoFile(DevBase):
         self.element_label = None
         self.filename      = None
         self.location      = None
-        if filepath!=None:
+        if filepath is not None:
             self.filename = os.path.basename(filepath)
             self.location = os.path.abspath(filepath)
             elem_label,symbol,is_elem = pp_elem_label(self.filename)
@@ -376,7 +376,7 @@ class Pseudopotential(DevBase):
         self.core    = None
         self.Zval    = None
         self.Zcore   = None
-        if filepath!=None:
+        if filepath is not None:
             self.read(filepath,format)
         #end if
     #end def __init__
@@ -394,7 +394,7 @@ class Pseudopotential(DevBase):
         if self.requires_format:
             if format is None:
                 self.error('format keyword must be specified to read file {0}\nvalid options are: {1}'.format(filepath,self.formats))
-            elif not format in self.formats:
+            elif format not in self.formats:
                 self.error('incorrect format requested: {0}\nvalid options are: {1}'.format(format,self.formats))
             #end if
         #end if
@@ -411,12 +411,12 @@ class Pseudopotential(DevBase):
         if self.requires_format:
             if format is None:
                 self.error('format keyword must be specified to write file {0}\nvalid options are: {1}'.format(filepath,self.formats))
-            elif not format in self.formats:
+            elif format not in self.formats:
                 self.error('incorrect format requested: {0}\nvalid options are: {1}'.format(format,self.formats))
             #end if
         #end if
         text = self.write_text(format)
-        if filepath!=None:
+        if filepath is not None:
             open(filepath,'w').write(text)
         #end if
         return text
@@ -468,7 +468,7 @@ class SemilocalPP(Pseudopotential):
         self.rcut_L2 = None
         self.components = obj()
         Pseudopotential.__init__(self,filepath,format)
-        if src!=None:
+        if src is not None:
             self.transfer_core_from(src)
         #end if
     #end def __init__
@@ -1002,7 +1002,7 @@ class SemilocalPP(Pseudopotential):
                 if c==self.local:
                     lab = self.local+' loc'
                 #end if
-                if self.name!=None:
+                if self.name is not None:
                     lab = self.name+' '+lab
                 #end if
                 v = self.evaluate_channel(r,c,with_local=with_local,rmin=rmin-1e-12)
@@ -1013,7 +1013,7 @@ class SemilocalPP(Pseudopotential):
                     if c==self.local:
                         v += self.Zval*r
                     #end if
-                elif metric!=None:
+                elif metric is not None:
                     self.error('invalid metric for plotting: {0}\nvalid options are: r2'.format(metric))
                 #end if
                 plot(r,v,color+linestyle,label=lab)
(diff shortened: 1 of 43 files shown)


nexus/lib/pseudopotential.py
```diff
@@ -1611,7 +1611,8 @@ def read_text(self,text,format=None,filepath=None):
                 n = int(lines[i]); i+=1
                 terms = []
                 for j in range(n):
-                    coeff,rpow,expon = lines[i].split(); i+=1
+                    coeff,rpow,expon = lines[i].split()
```
jtkrogel: Please revert this change

nexus/lib/qmcpack.py
```diff
@@ -1201,7 +1201,7 @@ def get_output_files(self):
     def post_analyze(self,analyzer):
         if not self.has_generic_input():
             calctypes = self.input.get_output_info('calctypes')
-            opt_run = calctypes!=None and 'opt' in calctypes
+            opt_run = True if calctypes is not None and 'opt' in calctypes else False
```
jtkrogel: Please revert and just replace `!=` with `is not`

nexus/lib/quantum_package_input.py
```diff
@@ -35,11 +35,11 @@ def read_qp_value_type(value_filepath):
         return 'bool'
     else:
         try:
-            v = int(svalue)
+            _ = int(svalue)
```
jtkrogel, line 38: Please revert

nexus/lib/quantum_package_input.py
```diff
@@ -35,11 +35,11 @@ def read_qp_value_type(value_filepath):
...
             return 'int'
         except:
             try:
-                v = float(svalue)
+                _ = float(svalue)
```
jtkrogel, line 42: Please revert

nexus/lib/structure.py
```diff
@@ -306,7 +306,7 @@ def read_cif(filepath,block=None,grammar='1.1',cell='prim',args_only=False):
 ##end try
 try:
     import spglib
-except:
+except ImportError:
```
jtkrogel: Please revert

nexus/lib/structure.py
```diff
@@ -1944,7 +1944,7 @@ def freeze(self,identifiers=None,radii=None,exterior=False,negate=False,directio
             #end for
         else:
             for index in indices:
-                frozen[index] = directions[i]==False
+                frozen[index] = not directions[i]
```
jtkrogel: Please revert

nexus/lib/superstring.py
```diff
@@ -138,7 +138,7 @@ def sbool(var):
 #end def sbool
 
 def is_bool(var):
-    return var==True or var==False or var in ['True','False']
+    return type(var) is bool or var in ['True', 'False']
```
jtkrogel: Please revert

nexus/lib/superstring.py
```diff
@@ -201,7 +201,7 @@ def string2array(string):
 #end def string2array
 
 def is_string(var):
-    return type(var)==type("s")
+    return type(var) is type("s")
```
jtkrogel: Please revert

nexus/lib/vasp_input.py
```diff
@@ -1515,7 +1515,7 @@ def incorporate_system(self,system,incorp_kpoints=True,coord='cartesian',set_nel
                 self.error('coord must be either direct or cartesian\nyou provided: {0}'.format(coord))
             #end if
             if s.frozen is not None:
-                poscar.dynamic = s.frozen==False
+                poscar.dynamic = not s.frozen
```
jtkrogel: Please revert

nexus/lib/vasp_input.py
```diff
@@ -1872,7 +1872,7 @@ def generate_poscar(structure,coord='cartesian'):
         error('coord must be either direct or cartesian\nyou provided: {0}'.format(coord),'generate_poscar')
     #end if
     if s.frozen is not None:
-        poscar.dynamic = s.frozen==False
+        poscar.dynamic = not s.frozen
```
jtkrogel: Please revert

nexus/lib/quantum_package_input.py
brockdyer03: This change is an additional bit of syntax work. Since the variable `v` is not accessed other than to assess its ability to be a number, e.g. is not returned, the variable should be named `_` to avoid confusion in other situations where `v` may be returned.

This was previously unofficial but widely accepted use, however in [Python 3.10](https://docs.python.org/3/reference/compound_stmts.html#wildcard-patterns) this was added as a wildcard pattern and as such binds no name.

nexus/lib/superstring.py
brockdyer03: I'm happy to revert this, but I'd like to check quickly that you also intend for `1` and `0` to return `True` in this context.

nexus/lib/superstring.py
brockdyer03: This is another case where the `==` sign should be replaced by `is` since the call here `type(var)==type("s")` is comparing singletons, e.g. `int is str` or `str is str`.

In addition, the function of `type("s")` actually can be replaced by `str` since the only thing that is happening in `type("s")` is assessing the type of the string `"s"`. Using `type("a")`, or even `type("int")` would also return `str` since the variable inside the type statement is a `str`.

Realistically however, the syntactically correct thing to do for any situation that would use this function is to remove the function entirely and replace its usage with
```python
if isinstance(var, str):
    # do something
```
as the `isinstance()` method is a builtin method, and thus is significantly faster. All methods provide the same result:

```python
>>> def func1(var): return type(var) == type("s")
>>> def func2(var): return type(var) is type("s")
>>> def func3(var): return type(var) == str
>>> def func4(var): return type(var) is str
>>>
>>> print(func1("a"))
>>> print(func2("a"))
>>> print(func3("a"))
>>> print(func4("a"))
>>> print(isinstance("a", type("s")))
>>> print(isinstance("a", str))

True
True
True
True
True
True
```

See the following timing tests:

```python
>>> %timeit def func(var): return type(var) == type("s") ; func("a")
53.1 ns ± 3.92 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

>>> %timeit def func(var): return type(var) is type("s") ; func("a")
51.5 ns ± 0.858 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

>>> %timeit def func(var): return type(var) == str ; func("a")
52.3 ns ± 0.783 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

>>> %timeit def func(var): return type(var) is str ; func("a")
52.1 ns ± 0.881 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

>>> %timeit isinstance("a", type("s"))
23.1 ns ± 0.354 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

>>> %timeit isinstance("a", str)
18.4 ns ± 0.827 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

brockdyer03: To be clear, I'd make the reversion here and if you agree with the use of `isinstance()` to replace `is_string()` in all files I'd go back and make that change everywhere in a separate PR.

jtkrogel (approved):
Proceed with merge

prckent:
Test this please

PR #5622, opened 2025-09-24
[Nexus] Remove deprecated Matplotlib keyword `warn` from `plotting.py`

Kcorbyerd:
## Proposed changes

The module `plotting.py` contained a series of `try... except` statements to test for matplotlib, however in the innermost nested `try... except` the command `matplotlib.use()` was invoked with a keyword `warn=False` that has been deprecated since at least the release of matplotlib 3.3, and has been defaulted to `False` since matplotlib 3.0. Given that matplotlib <3.0 is for Python 2 mostly, and since Nexus does not support Python 2 anymore, this should be removed.

Previously, this was throwing an exception that was not caught because of the `try... except` statement it was contained in that caught all exceptions, in this case a `TypeError`, even though it should only be catching the correct exception of `ValueError`.

- Bugfix

- No

Perlmutter with Python 3.9.18 using Matplotlib 3.8.2, and Python 3.13.2 and Matplotlib 3.10.1

## Checklist

### DIFF
--- a/nexus/lib/plotting.py
+++ b/nexus/lib/plotting.py
@@ -18,7 +18,7 @@ try:
     gui_envs = ['GTKAgg','TKAgg','Qt4Agg','WXAgg']
     for gui in gui_envs:
         try:
-            matplotlib.use(gui,warn=False, force=True)
+            matplotlib.use(gui, force=True)
             from matplotlib import pyplot
             success = True
             break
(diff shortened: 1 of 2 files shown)


prckent (approved):
LGTM

Confirmed still working locally (macports installed matplotlib 3.10.6)

Issue #5620, opened 2025-09-23
Nexus: Restructure imports to match PEP 8 guidelines


brockdyer03:
## The Issue
In my attempts to work on learning about everything Nexus has to offer and document its capabilities I've come across a few struggles, one in particular was some coding style that doesn't match PEP 8 recommendations (see issue #5618 and #5619). This was a relatively simple thing to change, however an additional struggle I've come across is what I'm calling "nested imports". For example, in `pwscf_input.py`, there is an import line that reads
```python
from developer import DevBase,log,warn,error
```
which one might assume from reading that `developer.py` contains all of those classes and methods. This is incorrect, as in reality the file `developer.py` contains its own import line that says
```python
from generic import obj,object_interface,log,error,warn,message
```
which shows that in actuality, the functions `log`, `warn`, and `error` belong to `generic.py`.

This is very confusing when trying to determine what modules are dependent on other modules and how things are generally structured in Nexus.

There are also a considerable number of star imports, for example in `pwscf_input.py` there is a line that reads
```python
from debug import *
```
which imports all objects (classes, functions, and variables) from `debug.py`. Now, any reference to these is not specified to be from `debug.py`, and rather is just another object that happens to have a valid reference. This greatly obfuscates the code dependencies and makes it extremely difficult to determine what some objects even are.

In addition, there are a large number of cases where a lot of specific functions are imported from a package. Take for example `structure.py`, which has the import
```python
from numpy import abs,all,append,arange,around,array,atleast_2d,ceil,cos,cross,cross,diag,dot,empty,exp,flipud,floor,identity,isclose,logical_not,mgrid,mod,ndarray,ones,pi,round,sign,sin,sqrt,uint64,zeros
```
and so any usage of these NumPy functions doesn't need its prepending `numpy.` or `np.` to be accessed. This can occasionally be convenient with functions like `sin` or `cos`, however it can muddy the namespace with functions like `append`.

## The Solution

To me the solution seems simple in theory, but difficult in practice. Essentially what needs to happen is that all imports need to be organized and set to their absolute origin. Unused imports should be removed, and any star imports should be replaced with their corresponding functions. Situations with large number of specific method imports (e.g. `structure.py`) should have the number of specific methods in the import reduced and replaced with explicit namespace calls.

Kcorbyerd:
[...]

PR #5624, opened 2025-09-24
[Nexus] Import Restructure and Cleanup

Kcorbyerd:
## Proposed changes

This PR is pursuant to Issue #5620 to restructure and clean up the imports in Nexus libraries.

Some important changes are:

1. Removing unused imports
2. Removing star imports - `from ... import *`
3. Removing conflicting namespace imports - `from numpy import append, abs, all, ...`
4. Removing underutilized `from` imports - `from ... import func` (func only used, say, <10 times in a thousand line code)
5. Shifting imports from `generic.py` to be from `developer.py` - `from generic import ...` -> `from developer import ...`
6. Moving some module-level imports to the top of the file
7. Removing duplicate imports

## Rationale

Many of these were discussed in #5620, however some changes were made to the original proposal after discussion. For example, it was decided that any imports from `generic.py` should be changed to originate from `developer.py` as `developer.py` is designed to collect the developer related base classes such as `obj`, as well as generally contain low-level functions such as `warn`, `error`, and `unavailable`. 

Additionally, the previous occurrences of star imports from `debug.py` were determined to likely be holdovers from development that survived deletion before merging, and as such are to be removed. 

Some of these changes are still pending approval, such as the removal of certain conflicting namespace imports and generally reducing the number of specific variables, functions, and classes that currently exist. The purpose here is supposed to be to make sure there is no ambiguity in the namespace for some functions that exist natively in Python (e.g. `abs`, `all`, `append`) as well as clarifying the origin of other functions that are commonly only referenced via their package name (e.g. `np.array`). The difficulty here is making sure that ALL instances of the function or variable are replaced with the correct namespace access prefix (all `array` -> `np.array`) in the appropriate places. If this is deemed to be too challenging to get done, it either will be moved to a different PR or dropped entirely.

- No

None

## Checklist

### DIFF
--- a/nexus/lib/structure.py
+++ b/nexus/lib/structure.py
@@ -121,15 +121,22 @@ import os
 import numpy as np
 from copy import deepcopy
 from random import randint
-from numpy import abs,all,append,arange,around,array,atleast_2d,ceil,cos,cross,cross,diag,dot,empty,exp,flipud,floor,identity,isclose,logical_not,mgrid,mod,ndarray,ones,pi,round,sign,sin,sqrt,uint64,zeros
-from numpy.linalg import inv,det,norm
+from numpy import (
+    cos,
+    cross,
+    dot,
+    pi,
+    sin,
+    sqrt,
+)
+from numpy.linalg import inv, det, norm
 from unit_converter import convert
-from numerics import nearest_neighbors,convex_hull,voronoi_neighbors
-from periodic_table import pt,is_element
-from fileio import XsfFile,PoscarFile
-from generic import obj
-from developer import DevBase,unavailable,error,warn
-from debug import ci,ls,gs
+from numerics import nearest_neighbors, convex_hull, voronoi_neighbors
+from periodic_table import is_element
+from periodic_table import pt as ptable
+from fileio import XsfFile, PoscarFile
+from developer import DevBase, obj, unavailable, error
+import itertools
 
 try:
     from scipy.special import erfc
@@ -144,6 +151,28 @@ except:
 #end try
 
 
+
+# installation instructions for spglib interface
+#
+#  this is bootstrapped off of spglib's ASE Python interface
+#
+#  installation of spglib
+#    go to http://sourceforge.net/projects/spglib/files/
+#    click on Download spglib-1.8.2.tar.gz (952.6 kB)
+#    unpack directory (tar -xzf spglib-1.8.2.tar.gz)
+#    enter ase directory (cd spglib-1.8.2/python/ase/)
+#    build and install (sudo python setup.py install)
+
+#try:
+#    from pyspglib import spglib
+#except:
+#    spglib = unavailable('pyspglib','spglib')
+##end try
+try:
+    import spglib
+except:
+    spglib = unavailable('spglib')
+#end try
 
 
 
@@ -183,7 +212,6 @@ except:
 
 
 cif2cell_unit_dict = dict(angstrom='A',bohr='B',nm='nm')
-
 
 
 def read_cif_celldata(filepath,block=None,grammar='1.1'):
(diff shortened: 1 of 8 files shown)


nexus/lib/pwscf_analyzer.py
```diff
@@ -867,14 +859,14 @@ def plot_bandstructure(self, filename=None, filepath=None, max_min_e = None, sho
...
-                from matplotlib.pyplot import figure,plot,xlabel,ylabel,title,show,ylim,legend,xlim,rcParams,rc,savefig,gca,xticks,axvline, scatter
+                from matplotlib.pyplot import figure,plot,ylabel,show,ylim,xlim,rcParams,savefig,gca,xticks,axvline, scatter
                 params = {'legend.fontsize':14,'figure.facecolor':'white','figure.subplot.hspace':0.,
                               'axes.labelsize':16,'xtick.labelsize':14,'ytick.labelsize':14}
                 rcParams.update(params)
             except(ImportError, RuntimeError):
                 success = False
             if not success:
-                figure,plot,xlabel,ylabel,title,show,ylim,legend,xlim,rcParams,savefig,bar,xticks,subplot,grid,setp,errorbar,loglog,semilogx,semilogy,text = unavailable('matplotlib.pyplot','figure','plot','xlabel','ylabel','title','show','ylim','legend','xlim','rcParams','savefig','bar','xticks','subplot','grid','setp','errorbar','loglog','semilogx','semilogy','text')
+                figure,plot,ylabel,show,ylim,xlim,rcParams,savefig,xticks = unavailable('matplotlib.pyplot','figure','plot','ylabel','show','ylim','xlim','rcParams','savefig','xticks')
```
prckent, line 869: Logic for removing xlabel, semilogx, semilogy etc.?

nexus/lib/structure.py
```diff
@@ -145,6 +152,28 @@
 
 
 
+# installation instructions for spglib interface
```
prckent, line 155: Not for this PR, but I'll note in passing that these instructions are quite out of date and likely not helpful any more. (I see they were copied from elsewhere due to the relocated import). If modern spglib needs some special handling that should be documented in the documentation...

nexus/lib/pwscf_analyzer.py
brockdyer03: Those functions were not accessed anywhere in the code. I would guess that there was a change at some point in the past that removed the part of the code that accessed them and didn't remove the functions from the import.

ye-luo:
One of the biggest mysteries to me is that our CI doesn't fail immediately but only in later PRs.

nexus/lib/testing.py
```diff
@@ -1,3 +1,20 @@
...
+from nexus_base import (
+    nexus_core,
+    nexus_noncore,
+    nexus_core_noncore,
+    nexus_core_defaults,
+    nexus_noncore_defaults,
+    nexus_core_noncore_defaults,
+)
+from pseudopotential import Pseudopotentials
+from simulation import Simulation
```
ye-luo, line 16: This is probably the single biggest issue moving function scope import to global. It reminds me of Fortran module `use`.

ye-luo:
@Kcorbyerd here are the following steps.
1. This PR needs to be temporarily reverted on develop.
2. Then you check out the develop and `git revert` the revert commit.
3. Apply your fixes and make a new PR.

ye-luo:
@jtkrogel 
> It looks like nxs-test was not run at all prior to merging this PR. This has to be guarded against in future.

Hopefully enough safeguard this time in https://github.com/QMCPACK/qmcpack/pull/5645
see failure caught in https://github.com/QMCPACK/qmcpack/pull/5644/checks

Issue #5625, opened 2025-09-26
Nexus: Replace `superstring.py` functionality in-situ


Kcorbyerd:
## The Issue

The file `superstring.py` is a rarely-used piece of legacy code that was written to enable more advanced string parsing in older Python versions. There are a total of 26 functions and 1 variable that are defined:

1. `next_visible_character()`
2. `remove_whitespace()`
3. `shrink_whitespace()`
4. `var2string()`
5. `sbool()`
6. `is_bool()`
7. `is_int()`
8. `is_float()`
9. `is_array()`
10. `string2val()`
11. `string2array()`
12. `is_string()`
13. `stringmap()`
14. `stringbreak()`
15. `find_matching_brace()`
16. `find_matching_pair()`
17. `remove_pair_sections()`
18. `remove_comment_lines()`
19. `remove_empty_lines()`
20. `contains_any()`
21. `contains_all()`
22. `find_matching_cbrace()`
23. `valid_variable_name()`
24. `split_delims()`
25. `find_preprocessor_end()`
26. `find_comment_block_end()`
27. `invalid_variable_name_chars`

Of these, 6 functions are imported at least once in code that is not the `superstring` test suite:

1. `string2val()`
2. `find_matching_pair()`
3. `remove_pair_sections()`
4. `remove_empty_lines()`
5. `valid_variable_name()`
6. `split_delims()`

with only `valid_variable_name()` seeing 2 separate imports.

The majority of these functions are only listed in `superstring.py` and do not exist in the test suite. As it stands, several functions that exist in `superstring.py` have already been redefined in-situ, and thus the goal of this issue is to remove the remaining instances of these functions (primarily in `xmlreader.py`) and replace with in-situ versions.

Removing the legacy code here will reduce unnecessary testing and further streamline ongoing efforts to document and update Nexus functionality.

## The Work

Currently the goals of this issue are:

- [ ] Replace `superstring.py` functions in `xmlreader.py`
- [ ] Replace `superstring.py` functions in `hdfreader.py`
- [ ] Replace `superstring.py` functions in `pseudopotential.py`
- [ ] Remove the `superstring.py` tests
- [ ] Deprecate and remove `superstring.py` entirely

[...]

PR #5627, opened 2025-09-29
[Nexus] Deprecate `superstring.py`

Kcorbyerd:
## Proposed changes

This PR should close Issue #5625 by removing all dependence on `superstring.py`.

The easiest fix was a carryover from #5624 that removed unused imports, but only the unused imports from `superstring.py`.

Next was replacing the remaining instances of the 6 functions in `superstring.py` that were still imported in other files.

### `split_delims()`
```python
def split_delims(s,delims=['.','-','_']):
    sp = s
    for d in delims:
        sp = sp.replace(d,' ')
    #end for
    return sp.split()
#end def split_delims
```

The first function to be removed was `split_delims()`, referenced in `pseudopotential.py`.

The function was used to pull the atomic symbol out of the name of a pseudopotential file, and only ever took the first string from the split:
```python
self.element = split_delims(os.path.split(filepath)[1])[0]
```
so was replaced with a simpler method that yields the same result:
```python
self.element = os.path.split(filepath)[1]
for delim in (".", "-", "_"):
    self.element = self.element.split(delim)[0]
```
Which passes this test:
```python
def read_pp_atom():
    s = "au.pbe_sol-v1.2.uspp.F.UPF"
    for delim in (".", "_", "-"):
        s = s.split(delim)[0]
    assert(s=="au")

    s = "c_pbe_sol-v1.2.uspp.F.UPF"
    for delim in (".", "_", "-"):
        s = s.split(delim)[0]
    assert(s=="c")

    s = "be-pbe_sol-v1.2.uspp.F.UPF"
    for delim in (".", "_", "-"):
        s = s.split(delim)[0]
    assert(s=="be")
#end def read_pp_atom
```

### `valid_variable_name()`
```python
invalid_variable_name_chars=set('!"#$%&\'()*+,-./:;<=>?@[\\]^`{|}-\n\t ')
def valid_variable_name(s):
    return not contains_any(s,invalid_variable_name_chars)
#end def valid_variable_name
```

This function was present in `xmlreader.py` and `hdfreader.py` and tests, as the name implies, for valid variable names. These are defined as names that do not contain any of these following characters: ```!"#$%&\'()*+,-./:;<=>?@[\\]^`{|}-\n\t ```.

This was replaced with a one-line statement utilizing list comprehension:
```python
if not any([i in ('!"#$%&\'()*+,-./:;<=>?@[\\]^`{|}-\n\t ') for i in var]):
```
which passes the test names from `test_valid_variable_name()`:
```python
def test_valid_variable_name():
    valid_var_names = [
        "valid_variable_name",
        "_valid_variable_name",
        "__valid_variable_name",
        "valid_variable_name__",
        "validVariableName",
    ]
    invalid_var_names = [
        "invalid variable name",
        "invalid_variable-name",
        "inval!d_variable_name",
        "inv@lid_variable_name",
        "in>alid_variable_name",
        "invalid_variable_name ",
    ]

    for var in valid_var_names:
        assert(any([i in ('!"#$%&\'()*+,-./:;<=>?@[\\]^`{|}-\n\t ') for i in var]) == False)

    for var in invalid_var_names:
        assert(any([i in ('!"#$%&\'()*+,-./:;<=>?@[\\]^`{|}-\n\t ') for i in var]) == True)
#end def test_valid_variable_name
```

### `remove_empty_lines()`
```python
def remove_empty_lines(s):
    sr=''
    lines = s.splitlines()
    for l in lines:
        if l.strip()!='':
            sr+=l + '\n'
        #end if
    #end for
    return sr
#end def remove_empty_lines
```

This function was present in `xmlreader.py` and simply removes any empty lines in a string.

It was replaced by the one-line list comprehension
```python
"".join([i+"\n" for i in s.splitlines() if i.strip() != ""])
```
that passes the tests from `test_superstring.py`. The original test is 
```python
def test_remove_empty_lines():
    from superstring import remove_empty_lines


    s = '''
        
        This string


        has a

        number of



        empty lines.
        '''

    sref = '''        This string
        has a
        number of
        empty lines.
'''


    assert(remove_empty_lines(s)==sref)

#end def test_remove_empty_lines
```
however after checking that the function works in this version I rewrote it as
```python
def test_remove_empty_lines():

	s = (
        "		\n"
		"This string\n"
		"\n"
		"\n"
		"has a\n"
		"\n"
		"number of\n"
		"\n"
		"\n"
		"\n"
		"empty lines.\n"
		"       "
	)

	sref = (
        "This string\n"
        "has a\n"
        "number of\n"
        "empty lines.\n"
	)

	s = "".join([i+"\n" for i in s.splitlines() if i.strip() != ""])
	assert(s==sref)

#end def test_remove_empty_lines
```
which is a more compact representation and removes ambiguity from the string literals.

### `find_matching_pairs()`
```python
def find_matching_pair(s,pair,start=0,end=-1):
    if end==-1:
        end=len(s)
    #end if

    left  = pair[0]
    right = pair[1]

    llen=len(left)
    rlen=len(right)

    ileft  = s.find(left,start,end)
    iright = -1
    if ileft==-1:
        return ileft,iright
    else:
        i=ileft+llen
        left_scope  = 0
        right_scope = 0
        found_match = False
        failed = False
        while not found_match and i<end:
            nleft  = s.find(left,i,end)
            nright = s.find(right,i,end)
            if nleft!=-1 and nleft<nright:
                right_scope+=1
                i=nleft+llen
            elif nright!=-1:
                found_match = right_scope==left_scope
                right_scope-=1
                i=nright+rlen
            elif nright==-1:
                failed=True
                break
            #end if
        #end while
        if found_match:
            iright = i
        #end if
        if failed:
            ileft,iright=-1,-1
        #end if
    #end if
    return ileft,iright
#end def find_matching_pair
```

This function was present in `xmlreader.py` and was replaced with an equivalent function `find_pair()`:
```python
def find_pair(s: str, pairs: list[str, str], start: int = 0, end: int | None = None):
    if end is None:
        end = len(s)

    left_pair = pairs[0]
    right_pair  = pairs[1]

    start_loc = s.find(left_pair, start, end)
    end_loc   = s.find(right_pair, start_loc + len(left_pair), end)
    if start_loc == -1 or end_loc == -1:
        return start_loc, end_loc

    return start_loc, end_loc+len(right_pair)
```
and passes the original test from `test_superstring.py`
```python
def test_find_pair():

    s = '''
        <simulation>
            <project id="vmc_hf_noj" series="0">
                <application name="qmcapp" role="molecu" class="serial" version="1.0"/>
            </project>
            <qmcsystem>
                <simulationcell>
                ...
                </simulationcell>
                <particleset>
                ...
                </particleset>
                <wavefunction>
                ...
                </wavefunction>
                <hamiltonian>
                ...
                </hamiltonian>
            </qmcsystem>
            <qmc method="vmc" move="pbyp">
                ...
            </qmc>
        </simulation>
        '''

    i1,i2 = find_pair(s,['<','>'])
    assert(s[i1:i2]=='<simulation>')
    i1,i2 = find_pair(s,['<','>'],i2)
    assert(s[i1:i2]=='<project id="vmc_hf_noj" series="0">')
    i1,i2 = find_pair(s,['</','>'])
    assert(s[i1:i2]=='</project>')
    i1,i2 = find_pair(s,['</','>'],i2)
    assert(s[i1:i2]=='</simulationcell>')
    i1,i2 = find_pair(s,['</','>'],i2)
    assert(s[i1:i2]=='</particleset>')
    i1,i2 = find_pair(s,['</','>'],i2)
    assert(s[i1:i2]=='</wavefunction>')
    i1,i2 = find_pair(s,['</','>'],i2)
    assert(s[i1:i2]=='</hamiltonian>')
    i1,i2 = find_pair(s,['</','>'],i2)
    assert(s[i1:i2]=='</qmcsystem>')
    i1,i2 = find_pair(s,['</','>'],i2)
    assert(s[i1:i2]=='</qmc>')
    i1,i2 = find_pair(s,['</','>'],i2)
    assert(s[i1:i2]=='</simulation>')
#end def test_find_pair
```

### `remove_pair_sections()`
```python
def remove_pair_sections(s,pair):
    sc=s
    ir=0
    n=0
    while ir!=-1 and n<10:
        il,ir = find_matching_pair(sc,pair)
        sc=sc.replace(sc[il:ir],'')
    #end while
    return sc
#end def
```

This function was imported into `xmlreader.py` and removes the strings contained between two matching pairs.

It was updated to use the modified `find_pair()` function:
```python
def remove_pair_sections(s: str, pair: list[str, str]):
    sc = s
    ir = 0
    while ir != -1:
        il, ir = find_pair(sc, pair)
        sc = sc.replace(sc[il:ir], "")
    #end while
    return sc
#end def remove_pair_sections
```
This passes the original test `test_remove_pair_sections()` from `test_superstring.py`:
```python
def test_remove_pair_sections():

    s = '''
        <simulation>
           <project id="vmc_hf_noj" series="0">
              <application name="qmcapp" role="molecu" class="serial" version="1.0"/>
           </project>
           <qmcsystem>
              <simulationcell>
                ...
              </simulationcell>
              <particleset>
                ...
              </particleset>
              <wavefunction>
                ...
              </wavefunction>
              <hamiltonian>
                ...
              </hamiltonian>
           </qmcsystem>
           <qmc method="vmc" move="pbyp">
              ...
           </qmc>
        </simulation>
        '''

    s = remove_pair_sections(s,('<hamiltonian>','</hamiltonian>'))

    s_no_h = '''
        <simulation>
           <project id="vmc_hf_noj" series="0">
              <application name="qmcapp" role="molecu" class="serial" version="1.0"/>
           </project>
           <qmcsystem>
              <simulationcell>
                ...
              </simulationcell>
              <particleset>
                ...
              </particleset>
              <wavefunction>
                ...
              </wavefunction>
              
           </qmcsystem>
           <qmc method="vmc" move="pbyp">
              ...
           </qmc>
        </simulation>
        '''
    assert(s==s_no_h)


    s = remove_pair_sections(s,('<qmcsystem>','</qmcsystem>'))

    s_no_sys = '''
        <simulation>
           <project id="vmc_hf_noj" series="0">
              <application name="qmcapp" role="molecu" class="serial" version="1.0"/>
           </project>
           
           <qmc method="vmc" move="pbyp">
              ...
           </qmc>
        </simulation>
        '''
    assert(s==s_no_sys)
    

    s = remove_pair_sections(s,('<simulation>','</simulation>'))

    assert(s.strip()=='')

#end def test_remove_pair_sections
```
### `string2val()`
```python
def string2val(s,delim=None):
    if is_bool(s):
        val = sbool(s)
    elif is_int(s):
        val = int(s)
    elif is_float(s):
        val = float(s)
    elif is_array(s,int,delim):
        val = array(s.split(delim),int)
    elif is_array(s,float,delim):
        val = array(s.split(delim),float)
    else:
        val = s
    #end if
    return val
#end def string2val
```

This function was used in `xmlreader.py` and was replaced with the equivalent `parse_string()`.
```python
def parse_string(s: str, delim: str | None = None):
    if not isinstance(s, str):
        raise TypeError("This function only parses strings!")

    # Check if number
    try:
        return int(s)
    except ValueError:
        try:
            return float(s)
        except ValueError:
            pass
    #end try

    # Check if bool
    if s.lower() in ["true", "false"]:
        return s.lower() == "true" # return True if s=="true" else False
    #end if

    # Check if number array
    try:
        return np.array(s.split(delim), int)
    except ValueError:
        try:
            return np.array(s.split(delim), float)
        except ValueError:
            return s
    #end try
#end def parse_string
```
This function performs the same, but is self-contained and requires no other functions to be defined. It also passes the previous test `test_string2val()` in `test_superstring.py`:
```python
def test_parse_string():

    v = parse_string('False')
    assert(isinstance(v,bool))
    assert(v==False)

    v = parse_string('True')
    assert(isinstance(v,bool))
    assert(v==True)

    v = parse_string('43')
    assert(isinstance(v,int))
    assert(v==43)

    v = parse_string('3.14')
    assert(isinstance(v,float))
    assert(np.abs(v-3.14)<1e-6)

    v = parse_string('1 2 3 4',delim=' ')
    assert(isinstance(v, np.ndarray))
    assert(v.dtype==int)

    v = parse_string('1, 2, 3, 4',delim=',')
    assert(isinstance(v, np.ndarray))
    assert(v.dtype==int)

    v = parse_string('1. 2 3 4',delim=' ')
    assert(isinstance(v, np.ndarray))
    assert(v.dtype==float)

    v = parse_string('1., 2, 3, 4',delim=',')
    assert(isinstance(v, np.ndarray))
    assert(v.dtype==float)
#end def test_parse_string
```

- Unsure

Tested with Python 3.13.1 & NumPy 2.2.2.

## Checklist


prckent:
Test this please

prckent (dismissed):
LGTM. The tests pass and there are a net 768 lines removed. Thanks.

nexus/lib/hdfreader.py
```diff
@@ -307,7 +306,7 @@ def __init__(self,fpath,verbose=False,view=False):
             hcur  = self.hcur[self.ilevel]
             for kr,v in hcur.items():
                 k=cur._escape_name(kr)
-                if valid_variable_name(k):
+                if not any([i in ('!"#$%&\'()*+,-./:;<=>?@[\\]^`{|}-\n\t ') for i in k]):
```
jtkrogel: Since this is repeated multiple places, it should be bound in a function.  A one line function named valid_variable_name would suffice.  Having a named function also helps a lot for anyone else to understand what that code is for.

nexus/lib/pseudopotential.py
```diff
@@ -1606,7 +1607,9 @@ def read_text(self,text,format=None,filepath=None):
             name,type,Zcore,lmax = lines[i].split(); i+=1
             Zcore = int(Zcore)
             lmax  = int(lmax)
-            element = split_delims(name)[0]
+            element = name
```
jtkrogel: Anytime there is repeat code, bind it in a function higher up.  The function can/should be local to pseudopotential.py as long as it is not used elsewhere.

nexus/lib/xmlreader.py
```diff
@@ -261,7 +310,7 @@ def __init__(self,fpath=None,element_joins=None,element_aliases=None,contract_na
             self.xml = remove_pair_sections(self.xml,pair)
         #end while
         #remove empty lines
-        self.xml = remove_empty_lines(self.xml)
+        self.xml = "".join([i+"\n" for i in self.xml.splitlines() if i.strip() != ""])
```
jtkrogel: Same goes here for multiple use, including the existing unit test.

nexus/lib/xmlreader.py
```diff
@@ -285,12 +334,12 @@ def __init__(self,fpath=None,element_joins=None,element_aliases=None,contract_na
 
     # test needed
     def include_files(self):
-        pair = '<include','/>'
-        qpair = '<?','?>'
-        ir=0
-        while ir!=-1:
-            il,ir = find_matching_pair(self.xml,pair,ir)
-            if ir!=-1:
+        pair = ['<include', '/>']
```
jtkrogel: Why a list vs a tuple?  Tuple is immutable, which is the context here.

nexus/lib/xmlreader.py
```diff
@@ -411,7 +460,9 @@ def found_element_start(self,ename,attributes):
             else:
                 k = kraw
             #end if
-            if valid_variable_name(k):
+            
+            # Check for variables containing invalid characters
+            if not any([i in ('!"#$%&\'()*+,-./:;<=>?@[\\]^`{|}-\n\t ') for i in k]):
```
jtkrogel: `valid_variable_name(...)`

tests/scripts/check_stats.py
```diff
@@ -1192,7 +1171,7 @@ def __init__(self,fpath,verbose=False,view=False):
             hcur  = self.hcur[self.ilevel]
             for kr,v in hcur.items():
                 k=cur._escape_name(kr)
-                if valid_variable_name(k):
+                if not any([i in ('!"#$%&\'()*+,-./:;<=>?@[\\]^`{|}-\n\t ') for i in k]):
```
jtkrogel: More `valid_variable_name`.  Due to its broad use, it might be more appropriate to house the function in `developer.py`

tests/scripts/check_stats.py
```diff
@@ -1253,7 +1232,7 @@ def add_group(self,hcur,cur,k,v):
         hcur  = self.hcur[self.ilevel]
         for kr,v in hcur.items():
             k=cur._escape_name(kr)
-            if valid_variable_name(k):
+            if not any([i in ('!"#$%&\'()*+,-./:;<=>?@[\\]^`{|}-\n\t ') for i in k]):
```
jtkrogel: Same

nexus/bin/nxs-test
```diff
@@ -638,26 +638,6 @@ def plotting():
 
 
 
-def superstring():
-    nunit('imports')
```
jtkrogel, line 642: The updated function code is fine.  However, if the code is repeated in a file or across files, it needs to bound as a function at a higher level in the code hierarchy (near top of file if local, e.g. in `developer` if used across multiple files).  In these cases, the unit tests should also be preserved, but moved to the test for the containing file/module.

jtkrogel (changes requested):
Changes needed.  See comments above.

nexus/lib/xmlreader.py
brockdyer03: I didn't particularly think much about list vs tuple, I usually just default to list since I rarely use tuples.

I think in this context a tuple is actually a much better choice than a list, so I'm happy to make that change.

nexus/lib/pseudopotential.py
brockdyer03: As I was checking `pseudopotential.py` I noticed that there actually is already both a function that reads the element label of a pseudopotential, `pp_elem_label()`, and a test for it, so I just replaced the code I wrote with the function that already exists.

nexus/lib/hdfreader.py
```diff
@@ -367,7 +366,7 @@ def add_group(self,hcur,cur,k,v):
         hcur  = self.hcur[self.ilevel]
         for kr,v in hcur.items():
             k=cur._escape_name(kr)
-            if valid_variable_name(k):
+            if not any([i in ('!"#$%&\'()*+,-./:;<=>?@[\\]^`{|}-\n\t ') for i in k]):
```
jtkrogel: Revert change.  `valid_variable_name` just added to `developer.py` should be called here.

nexus/lib/xmlreader.py
```diff
@@ -253,15 +315,15 @@ def __init__(self,fpath=None,element_joins=None,element_aliases=None,contract_na
...
+        pair = ('<!--', '-->')
         self.xml = remove_pair_sections(self.xml,pair)
         #process includes
         while self.xml.find('<include')!=-1:
             self.include_files()
             self.xml = remove_pair_sections(self.xml,pair)
         #end while
         #remove empty lines
-        self.xml = remove_empty_lines(self.xml)
+        self.xml = "".join([i+"\n" for i in self.xml.splitlines() if i.strip() != ""])
```
jtkrogel: Revert.  Should call `remove_empty_lines` function now included higher in this file.

brockdyer03: I also removed the comment above it about removing empty lines since the code now speaks for itself.

jtkrogel (approved):
LGTM

ye-luo:
Test this please

PR #5628, opened 2025-09-29
[Nexus] Update handling of numeric arrays in `qmcpack_input.py`

### LINKED ISSUE #5626: Nexus: Custom is_int() function in qmcpack_input.py Script Will Make Scientific Notation "Int" Valued for Python 3.12.2
**Describe the bug**
I am piloting using QMCPack on my university's HPC by running the "QMCPack Summer School 2025" tutorials. While executing the lih_workflow.py in "session3_statistics_and_workflows/02_lih_hf_vmc" on my HPC, in the  I get the following error when the script tries to read in the *.xml in the vmc step:

"ValueError: could not convert string '3.0139241961e+00' to int64 at row 1, column 1."

I have identified that the issue comes from the "qmcpack_input.py" script (directory: qmcpack-4.1.0/nexus/lib) [Line 1610]. When reading the structure xml document, the "read" function calls to another custom function called "is_int" to check if the value is an int before parsing it as an int type. The float test comes afterwards.

**To Reproduce**

The custom "is_int" uses Python's built in int() function to see if the passed value turns into a int or not. If the variable becomes an int, the test passes. For my system's Python 3.12.2 version, a Python variable assigned a value like "3.0139241961e+00" will automatically get read as a float, which can get turned into an integer with int(). (See custom Python test output attached where I checked this.) Since the tests passes, the "read" function in qmcpack_input.py assumes the value read in from the structure value is an int64 type, and subsequently fails when trying to read in the float value.

**Expected behavior**
I assume that some Python int() versions will give scientific notation a value error so the float condition gets accepted next (i.e., the one used in the summer school virtual box). 

However, in case this is an issue for later Python versions, could an updated read-in filter that can handle scientific notation get created? Also, does anyone know of any short-term fixes in case I run into this for other tests?

**System:**
 - system name: Prime (located at NDSU's CCAST)
 - modules loaded: 1) DefaultModules 9) mkl/2023.2.0  2) openpbs/openpbs/22.05.11  10) openblas/dynamic/0.3.18
  3) gcc/11.2.0                11) lapack/gcc/64/3.10.0  4) tbb/2021.10.0             12) boost/1.88.0
  5) compiler-rt/2023.0.0      13) fftw/3.3.10  6) oclfpga/2023.2.0          14) cmake/3.30.3
 7) compiler/2023.0.0         15) hdf5/1.12.1-intel  8) mpi/2021.10.0             16) libxml2/2.9.10-intel-ab5p
---Loaded and execute in a Python Virtual environment which includes all needed modules; Python 3.12.2; note that
int() is a default Python function

 - other systems where this is reproducible [none; virtual box does not do this]

**Additional context**
I am the first person testing QMCPack on the the Prime HPC system.

[troubleshoot_qe_intReader.zip](https://github.com/user-attachments/files/22564334/troubleshoot_qe_intReader.zip)

brockdyer03:
## Proposed changes

Updated `qmcpack_input.py` to more robustly handle numeric array parsing

Closes #5626 pending testing.

The previous method for handling the parsing of a string to an array involved checking the type of the first string in the array, however as pointed out in Issue #5626 there was a possibility that the type of the potential array of strings was not homogeneous, and therefore even if the first value would be properly parsed as say, an `int`, an error would be raised when trying to parse the entire string as an array as type `int`.

The new method replaces that code with a series of `try... except:` statements that both circumvent the need for explicitly defined `is_int()`, `is_float()`, and `is_array()` functions, the latter of which introduced an ambiguity in namespace in the function `Param.write()`. The original and updated functions are shown below:

### Original
```python
if 'text' in xml:
    token = xml.text.split('\n',1)[0].split(None,1)[0]
    try:
        if is_int(token):
            val = loadtxt(StringIO(xml.text),int)
        elif is_float(token):
            val = loadtxt(StringIO(xml.text),float)
        else:
            val = array(xml.text.split())
        #end if
    except:
        if is_int(token):
            val = array(xml.text.split(),dtype=int)
        elif is_float(token):
            val = array(xml.text.split(),dtype=float)
        else:
            val = array(xml.text.split())
        #end if
    #end try
    if val.size==1:
        val = val.ravel()[0]
    #end if
#end if
```

### Updated
```python
if 'text' in xml:
    try:
        try: # Try to make int array
            val = loadtxt(StringIO(xml.text),int)
        except ValueError:
            try: # Try to make float array
                val = loadtxt(StringIO(xml.text),float)
            except ValueError:
                pass
            #end try
        #end try
    except:
        try: # Try to make int array
            val = array(xml.text.split(),dtype=int)
        except ValueError:
            try: # Try to make float array
                val = array(xml.text.split(),dtype=float)
            except ValueError:
                val = array(xml.text.split())
            #end try
        #end try
    #end try
    if val.size==1:
        val = val.ravel()[0]
    #end if
#end if
```

In addition to this, the function `attribute_to_value()` was redefined to facilitate the removal of the `is_int()`, `is_float()`, and `is_array()` functions, and to remove the possibility of an error arising in the same way as described earlier. The original and updated functions are shown below:

### Original
```python
def attribute_to_value(attr):
    if is_int(attr):
        val = int(attr)
    elif is_float(attr):
        val = float(attr)
    elif is_array(attr,int):
        val = array(attr.split(),int)
        if val.size==9:
            val.shape = 3,3
        #end if
    elif is_array(attr,float):
        val = array(attr.split(),float)
    else:
        val = attr
    #end if
    return val
#end def attribute_to_value
```

### Updated
```python
def attribute_to_value(attr):
    try:
        return int(attr)
    except ValueError:
        try:
            return float(attr)
        except ValueError:
            pass
        #end try
    #end try

    try:
        val = np.array(attr.split(),int)
        if val.size==9:
            val.shape = 3,3
        #end if
        return val
    except ValueError:
        try:
            return np.array(attr.split(),float)
        except ValueError:
            return attr
        #end try
    #end try

#end def attribute_to_value
```

- Bugfix

- No

Checked with Python 3.13.1 & NumPy 2.2.2

## Checklist

### DIFF
--- a/nexus/lib/qmcpack_input.py
+++ b/nexus/lib/qmcpack_input.py
@@ -1575,13 +1575,17 @@ class Param(Names):
         #end if
         if 'text' in xml:
             try:
-                #try: # Try to make int array
-                #    val = loadtxt(StringIO(xml.text),int,converters=int)
-                #except ValueError:
-                #    val = loadtxt(StringIO(xml.text),float)
-                #    #end try
-                ##end try
-                val = loadtxt(StringIO(xml.text),float)
+                try: # Try to make int array
+                    # NumPy >= 1.23.0 uses int(float(value)) here, must specify int converter
+                    if np.lib.NumpyVersion(np.__version__) >= '1.23.0':
+                        val = loadtxt(StringIO(xml.text),int,converters=int)
+                    else:
+                        # int converter however will break NumPy <1.23.0
+                        val = loadtxt(StringIO(xml.text),int)
+                except:
+                    val = loadtxt(StringIO(xml.text),float)
+                    #end try
+                #end try
             except:
                 try: # Try to make int array
                     val = array(xml.text.split(),dtype=int)


nexus/lib/qmcpack_input.py
```diff
@@ -157,58 +157,35 @@
...
     try:
-        float(var)
-        return True
+        return int(attr)
     except ValueError:
-        return False
+        try:
+            return float(attr)
+        except ValueError:
+            pass
```
jtkrogel, line 168: This `pass` allows `val` to be undefined at the end of the code block.  This should never happen.

nexus/lib/qmcpack_input.py
```diff
@@ -1597,23 +1574,28 @@ def read(self,xml):
...
         #end if
         if 'text' in xml:
-            token = xml.text.split('\n',1)[0].split(None,1)[0]
             try:
-                if is_int(token):
-                    val = loadtxt(StringIO(xml.text),int)
-                elif is_float(token):
+                try: # Try to make int array
+                    # NumPy >= 1.23.0 uses int(float(value)) here, must specify int converter
+                    if np.lib.NumpyVersion(np.__version__) >= '1.23.0':
```
jtkrogel, line 1580: I generally don't like local code workaround like this, but perhaps there isn't a better option.

jtkrogel (changes requested):
See first comment above.

nexus/lib/qmcpack_input.py
brockdyer03: If I haven't made any errors, the `pass` statement simply exits out of that `try... except:` block, and the function will move onto the next `try... except:` block, which will end up returning `attr` unless there happens to be an error raised other than `ValueError`, in which case something extremely wrong has happened since the only other error that `np.array()` could raise is `TypeError` for an incorrect `dtype`, but since that is set in the function and cannot be changed that should never happen.

nexus/lib/qmcpack_input.py
brockdyer03: I think there is possibly a way to change this so that it works in any version of NumPy, however it would require creating a dictionary that is always the same length as the length of the data that is being fed into the function, which would be some very involved code.

nexus/lib/qmcpack_input.py
jtkrogel: See this code:
```python
s = '1 #check out my integer!'

try:
  v = int(s)
except ValueError:
    try:
        v = float(s)
    except ValueError:
        pass

print(v)
```

This results in:
```python
NameError: name 'v' is not defined
```

The code previously present in that except statement is there as a catch-all of last resort.  Of course if that call results in failure, nothing can be done and the code should stop via whatever error is raised at that point.

nexus/lib/qmcpack_input.py
jtkrogel: Perhaps this is better?

```python
kw = {}
if np.lib.NumpyVersion(np.__version__) >= '1.23.0':
    kw.update(converters=int)
val = loadtxt(StringIO(xml.text),int,**kw)
```

nexus/lib/qmcpack_input.py
brockdyer03: If I only had the int and float conversions in place I'd agree, but the function will continue to the next set of `try... except:` statements like so:
```python
def func(s):
    try:
        return int(s)
    except ValueError:
        print("Failed int conversion")
        try:
            return float(s)
        except ValueError:
            print("Failed float conversion")
            pass
        #end try
    #end try

    try:
        val = np.array(s.split(),int)
        if val.size==9:
            val.shape = 3,3
        #end if
        return val
    except ValueError:
        print("Failed int array conversion")
        try:
            return np.array(s.split(),float)
        except ValueError:
            print("Failed float array conversion")
            return s
        #end try
    #end try

s = '1 #check out my integer!'
func(s)
```
This yields:
```
Failed int conversion
Failed float conversion
Failed int array conversion
Failed float array conversion

'1 #check out my integer!'
```

nexus/lib/qmcpack_input.py
brockdyer03: I'm not sure if it's better, but it definitely does do the same job. I think I prefer my method because it avoids the creation of a new variable and updating that variable in favor of a simple `if... else:` block, but the method you've suggested does avoid having to have nearly duplicate lines which is a benefit for conciseness. I could be swayed either way, but currently I'm content to stick with what I've got.

prckent: My vote is for @Kcorbyerd 's current method since it is shorter and takes less reasoning to read.

Unfortunately hard version checks on dependencies are going to be needed where functionality &/or the API has changed. The only way we could avoid this would be with strict version checking and supporting a much more limited set of versions.

Independently from this PR, we might discuss some support policy, listing the definitely supported range of versions (python+dependencies required & optional), and having a robot test them. It is an easy prediction that this issue will keep recurring -- APIs are going to keep changing and at some point new python+package features will end up in a PR.

PR #5599, opened 2025-08-29
[WIP] RAII integration for TWFFastDerivWrapper

anbenali:
# \[WIP] RAII: Move `TWFFastDerivWrapper` under `TrialWaveFunction` and add proper MW resource management

Please review the [[developer documentation](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)](https://github.com/QMCPACK/qmcpack/wiki/Development-workflow)
and follow the [[GitHub Pull Request guidance](https://qmcpack.readthedocs.io/en/develop/developing.html#github-pull-request-guidance)](https://qmcpack.readthedocs.io/en/develop/developing.html#github-pull-request-guidance). I’ve tried to avoid the listed “Don’ts”.

## Proposed changes

**Goal:** Make `TrialWaveFunction` the owner/activator of `TWFFastDerivWrapper` and manage its multi-walker (MW) shared buffers via RAII, instead of allocating/freeing per function or from estimator-owned code.

**Why:**

* Previously `TWFFastDerivWrapper` was effectively owned/used by ACForce/estimators, which do not have a dedicated MW resource lock.
* Attempting to `takebackResource()` from nested components under a `ResourceCollectionTeamLock<TrialWaveFunction>` caused runtime errors (mismatched resource contexts).
* Moving ownership to `TWF` restores clean RAII boundaries: resources are lent under the `TWF` team lock and reclaimed at the lock’s destruction; nested components only clear their handles.

**What this PR does (incremental scaffolding):**

* Adds `TWFFastDerivWrapper` ownership to `TrialWaveFunction` (prototype + ctor/init path).
* Adds `TrialWaveFunction::initializeTWFFastDerivWrapper(const ParticleSet&, TWFFastDerivWrapper&)` and uses it to register Slater/Jastrow pieces.
* Ensures nested components **do not** call `takebackResource()`; they clear their `mw_mem_handle_` in `releaseResource()`. The `ResourceCollectionTeamLock<TrialWaveFunction>` drives the actual lifecycle.
* Keeps existing call sites working; functionality is unchanged. This is a structural/ownership refactor to enable later device-only paths for derivative stacks.

> Note: This is **WIP** and intentionally small so we can validate CI and downstream compilation. Follow-up PRs will migrate more call paths to device and expand tests.

## What type(s) of changes does this code introduce?

* Refactoring (no functional changes, no API changes)

## Does this introduce a breaking change?

* No

## Systems tested on

* Linux x86\_64, GCC/Clang (local builds)
[...]

PR #5635, opened 2025-10-08
Remove hamiltonian from test_twf_fastderiv_raii

ye-luo:
## Proposed changes
Unit tests of wavefunction are not allowed to have dependency beyond wavefunction.
Thus the dependency on qmcham needs to be removed. This PR reverts part of the change from #5599
For Hamiltonian related tests, the APIs consumed by Hamiltonian should be tested directly without relying on Hamiltonian objects.

- Other (please describe): revert changes.

- No

laptop

## Checklist

### DIFF
--- a/src/QMCWaveFunctions/tests/CMakeLists.txt
+++ b/src/QMCWaveFunctions/tests/CMakeLists.txt
@@ -186,7 +186,6 @@ foreach(CATEGORY common trialwf sposet jastrow determinant)
     ${UTEST_EXE}
     catch_main
     qmcwfs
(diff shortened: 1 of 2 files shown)


PDoakORNL (commented):
Ok I see the dependency this but should there be an integration test then, I feel like I this could be tested somewhere if this is a requirement of the fastderiv code.

PDoakORNL (approved):
This looks good to me, but we should make sure we have appropriate testing for fastderiv integration at the appropriate level.

PR #5636, opened 2025-10-08
Postpone resizing distance tables from construction to evaluation

ye-luo:
## Proposed changes
distance tables had fixed sizes of target particles. This PR moves the resizing of internal storage to evaluate/mw_evaluate allowing more flexible use of DTs.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/DistanceTable.h
+++ b/src/Particle/DistanceTable.h
@@ -65,12 +65,8 @@ protected:
 
 public:
   ///constructor using source and target ParticleSet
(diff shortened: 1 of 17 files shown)


prckent (approved):
LGTM, but this is quite the test of our CI to avoid logic errors with missed resize calls…
Very minor questions/comments for possible future PRs:
1. I noticed If (newpos) {…} path not hit in get_first_neighbor(). Why not, and can it be?
2. get_first_neighbor() could be a couple of calls to std::min_element (two to avoid the “self term” iat==jat). Probably a bunch of places in the code like this.

PR #5637, opened 2025-10-09
Resize VirtualParticleSet at makeMoves instead of constructor

ye-luo:
## Proposed changes

~~Review after https://github.com/QMCPACK/qmcpack/pull/5636~~
With this change, `VirtualParticleSet` size is no longer fixed at construction. One `VirtualParticleSet` object can be used to serve different quadrature point counts. My next PR will move `VirtualParticleSet` object ownership from `NonLocalECPComponent` to `NonLocalECPotential` and clean up the pointer mess.

- New feature

- No

epyc-server

## Checklist

### DIFF
--- a/src/Particle/VirtualParticleSet.cpp
+++ b/src/Particle/VirtualParticleSet.cpp
@@ -62,7 +62,7 @@ VirtualParticleSet::VirtualParticleSet(const ParticleSet& p, size_t dt_count_lim
   }
 }
 
-void VirtualParticleSet::resizeVP(const size_t nptcl)
+void VirtualParticleSet::resize(const size_t nptcl)
 {
   TotalNum = nptcl;
   R.resize(nptcl);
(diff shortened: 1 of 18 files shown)


src/Particle/VirtualParticleSet.cpp
```diff
@@ -68,6 +62,15 @@ VirtualParticleSet::VirtualParticleSet(const ParticleSet& p, int nptcl, size_t d
   }
 }
 
+void VirtualParticleSet::resizeVP(const size_t nptcl)
```
prckent: Surely, resize() for consistency with other classes? Logic behind added "VP"?

prckent (changes requested):
Looks good overall but see naming question

src/Particle/VirtualParticleSet.cpp
ye-luo: should be called `resize()`. Fixed.

prckent (approved):

PR #5641, opened 2025-10-16
More housekeeping around NLPP

ye-luo:
## Proposed changes
1. Spin off `NonLocalECPotential` neighbor lists used for T-moves to a separate class.
2. Clean up unused neighbor list in SOECPComponent
3. Add buildQuadraturePointDeltaPosAndPartialPotential for maximizing code reuse.

- Refactoring

- No

epyc-server

## Checklist


ye-luo:
Test this please

src/QMCHamiltonians/NonLocalECPComponent.h
```diff
@@ -77,7 +77,7 @@ class NonLocalECPComponent : public QMCTraits
   std::vector<ValueType> wvec;
 
   //Position delta for virtual moves.
-  std::vector<PosType> deltaV;
+  std::vector<PosType> deltaV_;
```
camelto2, line 80: Looks like you're updating the private data members that are being touched in this PR to have the trailing _
It is probably worth updating this class to follow the formatting standards for all its private data members, but that should be a subsequent PR. 

Other class like SOECPComponent already do this, so it would nice to have this one updated as well to help with readability

src/QMCHamiltonians/NonLocalECPotential.cpp
```diff
@@ -489,34 +468,23 @@ void NonLocalECPotential::evaluateIonDerivs(ParticleSet& P,
...
   //We're going to ignore psi and use the internal Psi.
 
-  //loop over all the ions
-  const auto& myTable = P.getDistTableAB(myTableIndex);
-  // clear all the electron and ion neighbor lists
-  for (int iat = 0; iat < NumIons; iat++)
-    IonNeighborElecs.getNeighborList(iat).clear();
-  for (int jel = 0; jel < P.getTotalNum(); jel++)
-    ElecNeighborIons.getNeighborList(jel).clear();
-
```
camelto2, line 499: Don't you need to add the corresponding neighbor_lists.clear()

src/QMCHamiltonians/NonLocalECPotential.cpp
```diff
@@ -489,34 +468,23 @@ void NonLocalECPotential::evaluateIonDerivs(ParticleSet& P,
...
+      const auto& displ = myTable.getDisplRow(jel);
+      for (int iat = 0; iat < PP.size(); iat++)
         if (PP[iat] != nullptr && dist[iat] < PP[iat]->getRmax())
-        {
           value_ +=
               PP[iat]->evaluateOneWithForces(P, vp_ ? makeOptionalRef<VirtualParticleSet>(*vp_) : std::nullopt, ions,
                                              iat, Psi, jel, dist[iat], -displ[iat], forces_[iat], PulayTerm);
-          NeighborIons.push_back(iat);
-          IonNeighborElecs.getNeighborList(iat).push_back(jel);
-        }
```
camelto2, line 519: Doesn't this need to also add the neighbor_lists.addElecIonPair(jel, iat);

src/QMCHamiltonians/NonLocalECPotential.cpp
```diff
@@ -547,28 +513,16 @@ void NonLocalECPotential::evaluateOneBodyOpMatrix(ParticleSet& P,
...
         PPset[ipp]->rotateQuadratureGrid(generateRandomRotationMatrix(*myRNG));
 
-  //loop over all the ions
   const auto& myTable = P.getDistTableAB(myTableIndex);
-  // clear all the electron and ion neighbor lists
-  for (int iat = 0; iat < NumIons; iat++)
-    IonNeighborElecs.getNeighborList(iat).clear();
-  for (int jel = 0; jel < P.getTotalNum(); jel++)
-    ElecNeighborIons.getNeighborList(jel).clear();
-
```
camelto2, line 557: I think you need the neighbor_list.clear() here

src/QMCHamiltonians/NonLocalECPotential.cpp
```diff
@@ -547,28 +513,16 @@ void NonLocalECPotential::evaluateOneBodyOpMatrix(ParticleSet& P,
...
-      auto& NeighborIons = ElecNeighborIons.getNeighborList(jel);
-      for (int iat = 0; iat < NumIons; iat++)
+      const auto& dist  = myTable.getDistRow(jel);
+      const auto& displ = myTable.getDisplRow(jel);
+      for (int iat = 0; iat < PP.size(); iat++)
         if (PP[iat] != nullptr && dist[iat] < PP[iat]->getRmax())
-        {
           PP[iat]->evaluateOneBodyOpMatrixContribution(P, iat, psi, jel, dist[iat], -displ[iat], B);
-          NeighborIons.push_back(iat);
-          IonNeighborElecs.getNeighborList(iat).push_back(jel);
```
camelto2, line 570: I think you need the neighbor_lists.addElecIonPair(jel, iat); here

src/QMCHamiltonians/NonLocalECPotential.cpp
```diff
@@ -585,29 +539,17 @@ void NonLocalECPotential::evaluateOneBodyOpMatrixForceDeriv(ParticleSet& P,
...
         PPset[ipp]->rotateQuadratureGrid(generateRandomRotationMatrix(*myRNG));
 
-  //loop over all the ions
   const auto& myTable = P.getDistTableAB(myTableIndex);
-  // clear all the electron and ion neighbor lists
-  for (int iat = 0; iat < NumIons; iat++)
-    IonNeighborElecs.getNeighborList(iat).clear();
-  for (int jel = 0; jel < P.getTotalNum(); jel++)
-    ElecNeighborIons.getNeighborList(jel).clear();
-
```
camelto2, line 595: Probably need the neighbor_list.clear here

src/QMCHamiltonians/NonLocalECPotential.cpp
```diff
@@ -585,29 +539,17 @@ void NonLocalECPotential::evaluateOneBodyOpMatrixForceDeriv(ParticleSet& P,
...
+      const auto& dist  = myTable.getDistRow(jel);
+      const auto& displ = myTable.getDisplRow(jel);
+      for (int iat = 0; iat < PP.size(); iat++)
         if (PP[iat] != nullptr && dist[iat] < PP[iat]->getRmax())
-        {
           PP[iat]->evaluateOneBodyOpMatrixdRContribution(P, source, iat, iat_source, psi, jel, dist[iat], -displ[iat],
                                                          Bforce);
-          NeighborIons.push_back(iat);
-          IonNeighborElecs.getNeighborList(iat).push_back(jel);
-        }
```
camelto2, line 610: probably need the neighbor_list.addElecIonPair here

src/QMCHamiltonians/SOECPotential.cpp
```diff
@@ -114,29 +106,17 @@ void SOECPotential::evaluateImpl(ParticleSet& P, bool keep_grid)
         ppset_[ipp]->rotateQuadratureGrid(generateRandomRotationMatrix(*my_rng_));
 
   const auto& ble = P.getDistTableAB(my_table_index_);
-  for (int iat = 0; iat < num_ions_; iat++)
-    ion_neighbor_elecs_.getNeighborList(iat).clear();
-  for (int jel = 0; jel < P.getTotalNum(); jel++)
-    elec_neighbor_ions_.getNeighborList(jel).clear();
```
camelto2, line 120: don't we need the neighbor_list.clear() here

src/QMCHamiltonians/SOECPotential.cpp
```diff
@@ -114,29 +106,17 @@ void SOECPotential::evaluateImpl(ParticleSet& P, bool keep_grid)
...
-      {
-        RealType pairpot = 0;
-        if (use_exact_spin_)
-          pairpot = pp_[iat]->evaluateOneExactSpinIntegration(P, *vp_, iat, psi_, jel, dist[iat], -displ[iat]);
-        else
-          pairpot = pp_[iat]->evaluateOne(P, vp_ ? makeOptionalRef<VirtualParticleSet>(*vp_) : std::nullopt, iat, psi_,
-                                          jel, dist[iat], -displ[iat]);
-        value_ += pairpot;
-        NeighborIons.push_back(iat);
-        ion_neighbor_elecs_.getNeighborList(iat).push_back(jel);
```
camelto2, line 138: dont we need the neighbor_list.addElecIonPair here

src/QMCHamiltonians/SOECPotential.cpp
```diff
@@ -212,11 +192,6 @@ void SOECPotential::mw_evaluateImpl(const RefVectorWithLeader<OperatorBase>& o_l
 
     //loop over all the ions
     const auto& ble = P.getDistTableAB(O.my_table_index_);
-    //clear elec and ion neighbor lists
-    for (size_t iat = 0; iat < O.num_ions_; iat++)
-      O.ion_neighbor_elecs_.getNeighborList(iat).clear();
-    for (size_t jel = 0; jel < P.getTotalNum(); jel++)
-      O.elec_neighbor_ions_.getNeighborList(jel).clear();
```
camelto2, line 219: clear here

src/QMCHamiltonians/SOECPotential.cpp
```diff
@@ -225,16 +200,11 @@ void SOECPotential::mw_evaluateImpl(const RefVectorWithLeader<OperatorBase>& o_l
...
-        const auto& displ              = ble.getDisplRow(jel);
-        std::vector<int>& NeighborIons = O.elec_neighbor_ions_.getNeighborList(jel);
-        for (size_t iat = 0; iat < O.num_ions_; iat++)
+        const auto& dist  = ble.getDistRow(jel);
+        const auto& displ = ble.getDisplRow(jel);
+        for (size_t iat = 0; iat < O.pp_.size(); iat++)
           if (O.pp_[iat] != nullptr && dist[iat] < O.pp_[iat]->getRmax())
-          {
-            NeighborIons.push_back(iat);
-            O.ion_neighbor_elecs_.getNeighborList(iat).push_back(jel);
```
camelto2, line 235: addElecIonPair here

camelto2 (commented):
There are a number of places where we used to clear the neighbor lists that you didn't add the corresponding clear() for the new NeighborListForPseudo class. Similar for the addElecIonPair call. you only did it in the evaluateImpl and mw_evaluateImpl call.

src/QMCHamiltonians/NonLocalECPComponent.h
ye-luo: I touched these two because the plan is to remove them as class members.
The reason for not touching the rest is to separate non-functional change from functional change in this PR. Once I finished all the intended functional changes. renaming/formatting PR will follow.

src/QMCHamiltonians/NonLocalECPotential.cpp
camelto2: Actually now that I'm thinking about it more, you're only using the neighbor lists to register these for the Tmoves. So doing the clear and addElecIonPair only in the evaluateImpl and mw_evaluateImpl calls is probably fine. 

All the other calls where the clears were happening were in force related functions, and that happens after the evaluateImpl and mw_evaluateImpl have already been called. So I guess you don't need any of these?

ye-luo: No. neighbor lists' usage is removed in this function `evaluateIonDerivs` which doesn't interact with T-moves at all.

src/QMCHamiltonians/NonLocalECPotential.cpp
ye-luo: No. neighbor lists' usage is removed in this function `evaluateIonDerivs` which doesn't interact with T-moves at all.

src/QMCHamiltonians/NonLocalECPotential.cpp
ye-luo: No. neighbor lists' usage is removed in this function `evaluateOneBodyOpMatrix` which doesn't interact with T-moves at all.

src/QMCHamiltonians/NonLocalECPotential.cpp
ye-luo: No. neighbor lists' usage is removed in this function `evaluateOneBodyOpMatrix` which doesn't interact with T-moves at all.

src/QMCHamiltonians/NonLocalECPotential.cpp
ye-luo: No. neighbor lists' usage is removed in this function `evaluateOneBodyOpMatrixForceDeriv` which doesn't interact with T-moves at all.

src/QMCHamiltonians/NonLocalECPotential.cpp
ye-luo: No. neighbor lists' usage is removed in this function `evaluateOneBodyOpMatrixForceDeriv` which doesn't interact with T-moves at all.

src/QMCHamiltonians/SOECPotential.cpp
ye-luo: We don't have T-move support in SOECPotential. If we worked out the correct formulation in the case of SO case. It can be added back. For the moment, removal is better.

src/QMCHamiltonians/SOECPotential.cpp
ye-luo: The removal is intended.

src/QMCHamiltonians/SOECPotential.cpp
ye-luo: The removal is intended.

src/QMCHamiltonians/SOECPotential.cpp
ye-luo: The removal is intended.

src/QMCHamiltonians/NonLocalECPotential.cpp
ye-luo: > Actually now that I'm thinking about it more, you're only using the neighbor lists to register these for the Tmoves. So doing the clear and addElecIonPair only in the evaluateImpl and mw_evaluateImpl calls is probably fine.
> 
> All the other calls where the clears were happening were in force related functions, and that happens after the evaluateImpl and mw_evaluateImpl have already been called. So I guess you don't need any of these?

Just saw your second comment. You analysis is correct.

camelto2: Right, I realized that after I posted all these reviews. They are only needed for interaction with Tmoves, which none of the Force related calls need

camelto2 (approved):
I realized after I reviewed all those comments that all of the previous neighbor list calls weren't needed anywhere except the evaluateImpl and mw_evaluateImpl call. All force related (and SOECP) don't need them since they don't interact with T-Moves. 

In that case, this LGTM

PR #5645, opened 2025-10-17
[Nexus] Make nxs-test --ctestlist to have minimal import and stop at CMake if it fails

ye-luo:
## Proposed changes
1. Having a local nexus_version requires testing module imports and make "nxs-test --ctestlist" slow to run. Since nexus_version is not consumed, just remove. there remains a dedicated nexus version test run by ctest.
2. Postpone importing of the `testing` module. So "nxs-test --ctestlist" stops before importing numpy. If numpy version is too low, later ctest runs should still fail.
3. Issue fatal error if "nxs-test --ctestlist" fails. "nxs-test --ctestlist" becomes more robust and fast to run due to changes 1 and 2. Thus when it fails, it is better to request user actions.

- No

laptop

## Checklist

### DIFF
--- a/nexus/tests/CMakeLists.txt
+++ b/nexus/tests/CMakeLists.txt
@@ -9,7 +9,11 @@ if(ADD_TEST)
   if(NOT "${NEXUS_TESTLIST_TIMESTAMP}" STREQUAL "${CACHE_NEXUS_TESTLIST_TIMESTAMP}")
     message("Generating Nexus test list for the first time or after a nxs-test time stamp change")
     execute_process(COMMAND ${Python3_EXECUTABLE} ${qmcpack_SOURCE_DIR}/nexus/bin/nxs-test --ctestlist
-                    OUTPUT_VARIABLE NEXUS_TESTLIST)
+                    RESULT_VARIABLE NEXUS_TESTLIST_RESULT OUTPUT_VARIABLE NEXUS_TESTLIST)
+    if(NOT NEXUS_TESTLIST_RESULT EQUAL 0)
+      message(FATAL_ERROR "Command line '${Python3_EXECUTABLE} ${qmcpack_SOURCE_DIR}/nexus/bin/nxs-test --ctestlist' failed!")
+    endif()
+
     set(CACHE_NEXUS_TESTLIST
         ${NEXUS_TESTLIST}
         CACHE INTERNAL "NEXUS_TESTLIST cache variable" FORCE)


nexus/bin/nxs-test
```diff
@@ -85,11 +85,6 @@ def import_from_unit_test_module(module_name,test_name):
 #end def import_from_unit_test_module
 
 
-# import core testing functions (shared between nxs-test and unit tests)
-from testing import *
```
jtkrogel, line 89: Removing `import *`  needs more careful handling than is done here.  There are imported names used below that have erroneously been excluded here.

nexus/bin/nxs-test
```diff
@@ -2225,24 +2218,6 @@ def import_nexus_module(module_name):
 #end def import_nexus_module
 
 
-# Get Nexus version
```
jtkrogel, line 2228: Please revert.  `nxs-test` reporting its version is relevant.  The code removed here cannot raise an exception (never fails).

nexus/bin/nxs-test
```diff
@@ -2257,7 +2232,6 @@ if __name__=='__main__':
     parser = OptionParser(
         usage='usage: %prog [options]',
         add_help_option=False,
-        version='%prog {}.{}.{}'.format(*nexus_version)
```
jtkrogel, line 2260: Same as above.

nexus/bin/nxs-test
jtkrogel: I'm all for adding local imports (inside functions or in `__main__`).

jtkrogel: As the changes in this PR aim to protect against failures due to corruption of `testing.py`, by the same philosophy the imports from `generic` on line 38 need similar protection.

nexus/bin/nxs-test
```diff
@@ -2404,6 +2375,12 @@ if __name__=='__main__':
         os.environ['PYTHONPATH']=options.pythonpath
     #end if
 
+    # import core testing functions (shared between nxs-test and unit tests)
+    from testing import *
```
jtkrogel, line 2379: This import does not cover all the code that needs imports from `testing.py`.  See, e.g., use of `object_diff` on line 1857.

jtkrogel: The `*` import should be replaced with specific names.

nexus/bin/nxs-test
ye-luo: This part causing loading all the modules potentially used by nexus via
```
import_nexus_module('versions')
```
this causes nxs-test being a monster.
CMake should not be the location of checking python module compliance.
So I need to make the only execution of nxs-test, namely `nxs-test --ctestlist` light.

I hesitate to mark  `nxs-test --ctestlist` run fatal error in cmake if it fails easily. There is no need to stop CMake due to subtle nxs-test failure.

nexus/bin/nxs-test
ye-luo: I could not find a consumer of `version`. Did I miss anything?
If `version` is actually needed, we can reintroduce it however its initialization needs to be handled differently without making `nxs-test -ctestlist` heavy.

nexus/bin/nxs-test
ye-luo: Yep. Better  being explicit.

nexus/bin/nxs-test
ye-luo: > As the changes in this PR aim to protect against failures due to corruption of `testing.py`.

That is correct.

>by the same philosophy the imports from generic on line 38 need similar protection.

Will take a look.

jtkrogel: Perhaps for `generic` a local `try-except` will be enough.  When making these changes, perhaps temporarily put a failing line in `generic.py`, e.g. `from nothing import something` to check for robustness.

nexus/bin/nxs-test
jtkrogel: I don't understand this statement.  `versions.py` is light and only has imports from the Python standard library (nothing from the rest of Nexus).

nexus/bin/nxs-test
jtkrogel: Can you show timings or something?  Similar to the comment above, the `versions.py` import should be very cheap and safe.

nexus/bin/nxs-test
jtkrogel: If you want, you can forgo the import of `versions.py` in the following way: 

1) Put `nexus_version = 2,1,0` directly in `testing.py` (duplicate). 
2) Add a test to `test_versions.py` to check that `nexus_version` from `testing.py` matches the one in `versions.py`.

nexus/bin/nxs-test
jtkrogel: See comment in thread above.  A fix to guarantee `testing.py` is lightweight is suggested there.

nexus/bin/nxs-test
ye-luo: I don't like making duplicates. After reading `versions.py`, I found `nexus_version` doesn't interact with the rest of the code in `versions.py`.
How about split `versions.py` in to `nexus_version.py` and `dependency_versions.py`?

nexus/bin/nxs-test
ye-luo: It is unclear to me how to deal with except case. I just moved import of `generic` to local in #5650

nexus/bin/nxs-test
ye-luo: Implemented in #5650

nexus/bin/nxs-test
jtkrogel: Splitting is fine.

ye-luo:
Superseded by #5650 #5651 #5654

